PR #3, opened 2019-03-06
Fix permission checking for performance counters

xiw:
This pull request consists of two patches.

The first patch fixes accesses to `cycle`/`instret`, which should be allowed (e.g., `rdcycle` in M-mode); the current code unconditionally throws an exception.  The cause is that `check_Counteren()` is short-circuited by `is_CSR_defined()`.  Note that the patch keeps `time` hidden as before, since it's accessible as a memory-mapped register.

The second patch fixes checks for U-mode accesses to counters.  According to the specification 3.1.12, "When the CY, TM, IR, or HPMn bit in the `mcounteren` register is clear, attempts to read the `cycle`, `time`, `instret`, or `hpmcountern` register while executing in S-mode or U-mode will cause an illegal instruction exception."  In other words, U-mode accesses to these counters require checks on both `mcounteren` and `scounteren`.  The current code checks `scounteren` only and ignores `mcounteren`.

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -76,9 +76,9 @@ function check_Counteren(csr : csreg, p : Privilege) -> bool =
     (0xC01, Supervisor) => mcounteren.TM() == true,
     (0xC02, Supervisor) => mcounteren.IR() == true,
 
-    (0xC00, User) => scounteren.CY() == true,
-    (0xC01, User) => scounteren.TM() == true,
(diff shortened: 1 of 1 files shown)


pmundkur:
The issue addressed by the first patch is fixed in the rv32 branch, which should soon become master.  I've applied your second patch on that branch, and edited it slightly to handle the case when supervisor mode is not implemented.  Thanks!

xiw:
Awesome - thank you!

PR #5, opened 2019-05-01
Add base address register as extra argument to ext_data_check_addr hook

rmn30:
This will assist with implementing CHERI capability mode.


rmn30:
In the CHERI case for LOAD and STORE we end up reading rs1 twice which is unfortunate and could be a problem with RMEM. We could avoid this by moving the read and add into the handler in both cases and just passing in the offset, but it might be slightly confusing to read in the vanilla RISCV model. If it helps we could rename the hook `computer_memory_addr` or something.

pmundkur:
Moving the read and add into the handler looks preferable.  We need to document the hook anyway, and we could perhaps add an additional clarifying comment in the vanilla model.  would `data_get_addr` be okay as a name? i.e. just changing `check` -> `get`.

PR #8, opened 2019-05-26
Two small README updates.

ccasin:
Two suggested edits to the README based on my attempt to build the project.  One is a link explaining how to install sail via opam at the appropriate place, and the other documents a dependency on `dtc` that seems required to use the ocaml simulator.


pmundkur:
Thanks!

PR #9, opened 2019-05-29
Display correct filename in message when using -T

scottj97:
This untested but trivial change fixes a clear bug in the filename displayed when using -T.


pmundkur:
thanks!

PR #11, opened 2019-05-30
Makefile upgrade

scottj97:
This is a minor change to the Makefile to clean up a few small things.

I had bigger plans to enable incremental recompilation of the C sources (instead of always having to recompile the 4MB+ `riscv_model_RV64.c` every time I touch `riscv_sim.c`) but I realized that sometimes the C sources are compiled with -DRVFI_DII and sometimes not, so we would need to build two sets of `*.o` files, one with -DRVFI_DII and one without, and I decided that's not worth the hassle.


Alasdair:
I think the main reason the generated C is so large is that the assembly syntax mappings are represented in a very naive way (essentially giant if-then-else chains)

PR #20, opened 2019-09-17
Clarify virtual vs physical address in variable names

scottj97:
There are a lot of variables in `riscv_mem.sail` and `riscv_platform.sail` named `addr` and it's not always clear if that is physical or virtual address. These changes rename those variables to `paddr` to reduce ambiguity.

I first made these changes because I wanted access to both vaddr and paddr for tracing purposes, so I first needed to rename the existing variables to distinguish. I'm not planning to upstream my changes to add vaddr, but I feel these variable renames are general enough to upstream here.

### DIFF
--- a/model/riscv_mem.sail
+++ b/model/riscv_mem.sail
@@ -20,14 +20,14 @@ function is_aligned_addr forall 'n. (addr : xlenbits, width : atom('n)) -> bool 
   unsigned(addr) % width == 0
 
 // only used for actual memory regions, to avoid MMIO effects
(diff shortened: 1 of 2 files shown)


rmn30:
Looks fine to me but I've requested Prashanth's review to check he's OK with it. The main downside is the potential for conflicts during merges, but separating it into many separate commits as you have done might help with this.

pmundkur:
looks good to me, thanks, merged.

scottj97:
Thanks Prashanth. Can you clarify the squash policy? I kept these as separate, atomic commits but it looks like you would have preferred one big commit.

Also if you enable "Squash and Merge" in the GitHub repo, then GitHub can do the squashing for you, instead of (misleadingly) showing this PR as unmerged.

pmundkur:
Separate atomic commits is great for review, thanks for submitting it that way.  For merging, I thought it would be worth having your motivational comment in the commit message, in addition to having a single commit, which is why I made a custom merge (which is confusing the github pull system).

PR #22, opened 2019-10-31
Attempt to fix RVFI for instruction fetch exceptions. 

rmn30:
I made this to try to fix a problem with TestRig where instructions that cause a fetch exception (e.g. CHERI PCC error) do not produce an RVFI report and instead enter a loop executing the same instruction. I'm not sure what 'need_instr' was intended to do but does not seem to be what we want. Maybe you have an idea? It's a bit untested until someone (e.g. @PeterRugg) can give it a go.

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -687,35 +687,33 @@ void run_sail(void)
   while (!zhtif_done && (insn_limit == 0 || total_insns < insn_limit)) {
 #ifdef RVFI_DII
     if (rvfi_dii) {
-      if (need_instr) {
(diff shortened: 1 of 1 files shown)


pmundkur:
It looks like need_instr checks if there was a risc-v exception, in order to break out of the loop.  iirc the step boolean return for zstep() was to get step counters in the execution trace to match those of Spike.  I'm not sure if that is what need_instr actually wants.

run_sail() badly needs a cleanup, it's pretty confusing.  Is there a reproducible test-case?  I could use that to do the cleanup.

rmn30:
Peter Rugg tested and it appeared to work.

PR #28, opened 2019-12-05
Added a Dockerfile to enable one-step installation of sail-riscv

jameyhicks:
Use of a docker container enables one step installation of the sail-riscv dependences on Linux, Mac, or Windows.

Assuming docker image has been built and pushed to docker hub, e.g. jameyhicks/sail-riscv:

First, clone sail-riscv:
` $ git clone https://github.com/rems-project/sail-riscv`

Second, perform actions of interest in a docker container with necessary tools installed:
` $ docker run -v $PWD/sail-riscv:/sail-riscv -w /sail-riscv jameyhicks/sail-riscv:latest make csim`

I plan to add usage notes, either to README.md or docker/README.md.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,17 @@
+From ocaml/opam2:ubuntu-lts
+
+RUN sudo apt-get update; sudo apt-get -y install build-essential libgmp-dev z3 m4 pkg-config zlib1g-dev device-tree-compiler; sudo apt-get -y clean
+
+RUN opam switch create ocaml-base-compiler.4.06.1
+RUN opam config env
+
+ENV OPAM_SWITCH_PREFIX='/home/opam/.opam/ocaml-base-compiler.4.06.1'
+ENV CAML_LD_LIBRARY_PATH='/home/opam/.opam/ocaml-base-compiler.4.06.1/lib/stublibs:Updated by package ocaml'
+ENV OCAML_TOPLEVEL_PATH='/home/opam/.opam/ocaml-base-compiler.4.06.1/lib/toplevel'
+ENV MANPATH=':/home/opam/.opam/4.06.1/man:/home/opam/.opam/ocaml-base-compiler.4.06.1/man'
+ENV PATH='/home/opam/.opam/ocaml-base-compiler.4.06.1/bin:/home/opam/.opam/4.06.1/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
+
+RUN which ocaml
+RUN opam config env
+RUN opam repository add rems https://github.com/rems-project/opam-repository.git; opam install sail
+


rmn30:
Hi Jamey, this is a good idea and similar to what we do for our continuous integration in Jenkins. Note that the opam switch command is not really necessary -- sail should work OK with the latest ocaml in the docker image (4.06.1 is really a minimum requirement). Are the ENV directives necessary? Doesn't that image have a default entry point of opam exec ...?

docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+From ocaml/opam2:ubuntu-lts
+
+RUN sudo apt-get update; sudo apt-get -y install build-essential libgmp-dev z3 m4 pkg-config zlib1g-dev device-tree-compiler; sudo apt-get -y clean
+
+RUN opam switch create ocaml-base-compiler.4.06.1
+RUN opam config env
```
rmn30: is this line for debugging purposes?

docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+ENV OPAM_SWITCH_PREFIX='/home/opam/.opam/ocaml-base-compiler.4.06.1'
+ENV CAML_LD_LIBRARY_PATH='/home/opam/.opam/ocaml-base-compiler.4.06.1/lib/stublibs:Updated by package ocaml'
+ENV OCAML_TOPLEVEL_PATH='/home/opam/.opam/ocaml-base-compiler.4.06.1/lib/toplevel'
+ENV MANPATH=':/home/opam/.opam/4.06.1/man:/home/opam/.opam/ocaml-base-compiler.4.06.1/man'
+ENV PATH='/home/opam/.opam/ocaml-base-compiler.4.06.1/bin:/home/opam/.opam/4.06.1/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
+
+RUN which ocaml
```
rmn30: also debugging?

docker/Dockerfile
jameyhicks: Yes, it could be removed. I was working through Nikhil's installation slides for the tutorial that he is doing next week at the Summit and some of those steps are still in the Dockerfile.

docker/Dockerfile
jameyhicks: yes

rmn30:
I tried this today. The following Dockerfile was sufficient to install sail in the container:
```
FROM ocaml/opam2:ubuntu-lts
RUN sudo apt-get update; sudo apt-get -y install build-essential libgmp-dev z3 m4 pkg-config zlib1g-dev device-tree-compiler
RUN opam repository add rems https://github.com/rems-project/opam-repository.git; opam install sail
```
However when I tried to build sail-riscv using a bind-mount as you suggested I ran into problems with permissions. The opam base image has a user `opam` with user and group ID 1000 but that does not match my user and group ID on the host so the container could not write to the sail-riscv directory. You might get away with this if you happen to be user/group ID 1000 on the host (quite likely on a single user system) but I needed another solution. I tried a few ways but I'm not really satisfied with any of them:

1) Run the container with `-u <myhostuid>:<myhostgid>` . This doesn't work because opam immediately dies when it can't open it's log file in the container
2) Run the container with `-u root` . This works but you get warnings from opam about not running as root and it ends up scattering root owned files throught the sail-riscv directory
3) Use `chmod g+rwX -R sail-riscv` on host to make files in the bound mount group writable then run the container with `-u opam:<myhostgid>`. This works but any files the container creates have uid 1000 (some other user on my host system) and my gid which is a bit weird. These files are not group writable either. To set the container umask we'd have to create an entrypoint script.

Do you have any other ideas?

jameyhicks:
I also found that I could remove the other lines from the Dockerfile.

I saw the notes about uid in the opam docker repository, but I did not have
a problem even though the container is using uid 1000 and my uid is 501.
However, I've been testing with Docker for Mac, which has an additional
connection between the container and the host filesystem. It has different
quirks than on a Linux system.

Maybe we can make the log file location world writable inside the container
so that it can be run with the host uid and gid. I'm not very hot on
options 2 and 3.

I'll do some experimentation on both Mac and Linux and let you know how it
goes.

rmn30:
Commit 60776f9 referenced above added a simple Dockefile that makes this PR obsolete. However I think there is more that could be done to improve use of Docker e.g.:

1) Document use of the Dockerfile to build the model. It is partially documented in the above commit message but needs a proper README
2) Add a Dockerfile that builds the model too
3) Integrate with github actions / CI
4) Push built images to Dockerhub

All this could benefit from someone with more expierence in Docker than me.

martinberger:
> makes this PR obsolete. However I think there is more that could be done to improve use of Docker 

@rmn30  Sounds like we should close this pull request. Do we have an official wishlist somewhere? If not, I will create one.

rmn30:
Yes, I think it's fine to close. I'm not aware of any wishlist. You could create a new issue tagged with 'enhancement' and reference this PR?

PR #30, opened 2019-12-23
Generate correct cause for AMO faults

jrtc27:
The spec does not spell this out clearly, but I assume the intent is that AMOs
always give AMO (aliases store) faults and never the load counterparts. I have
opened https://github.com/riscv/riscv-isa-manual/issues/469 to track this.


pmundkur:
Thanks!

PR #32, opened 2020-01-07
Fix parsing long options in the C emulator for RVFI-DII

jrtc27:
idx, passed as longindex to getopt_long, is updated every time a long option is parsed to be the index of that option in the options array, not argv. We should instead use optind as in the non-RVFI-DII case, and we can remove the variable to avoid confusion since it is unused.

I have also included a convenience change so building RVFI behaves like non-RVFI, producing an executable suffixed with the architecture and isolating its intermediate build products.


pmundkur:
Thanks!

PR #33, opened 2020-01-18
Allow extensions to provide their own exception codes/names

jrtc27:


rmn30:
Looks fine to me, but maybe if the functions and ext_exc_type were scattered they would be easier to extend.

jrtc27:
Are you allowed empty scattered unions and functions for the no extension case?

jrtc27:
Yeah you're not, that gives:

```
Error:
Failure hd
```

Somewhere deep inside Sail is an assumption that scattered unions and/or functions have at least one clause.

rmn30:
Sounds like a big that should be fixed to at least five a sensible error. On the mean time there are two options:
1) use a sentinel
2) convert the entire exception type to scattered union
2) is probably preferable but I've not looked at the code to see how much work out would be

jrtc27:
I tried to convert the entire exception to a scattered union but ran into various issues.

The first issue is PTW_Error's PTW_Ext_Error which needs ext_translate_exception to return a valid ExceptionType, and that doesn't make sense when you have no extension. I then went down the rabbit hole of trying to convert *that* to a scattered union rather than nesting, which then encounters AccessType/ext_access_type. That may all be doable and result in better code, but that's far more extensive than I want to be doing.

However, sail-cheri-riscv provides its own replacement cheri_ptw.sail, so sail-riscv's PTW_Ext_Error is completely useless, and I tried just deleting it to avoid the issue. However, that encounters the issue that sync_exception is a struct containing an ExceptionType, and when it's a scattered union, Sail complains with `No val spec found for undefined_ExceptionType`, even if there's an `end ExceptionType` before it (at which point it *should* be able to de-scatter the union and do its normal undefined declaration stuff, but clearly that code is not yet there; such an implementation might not even need the `end` if it's already been parsed once).

It is at this point that I have given up trying to convert ExceptionType to a scattered union. I also believe that 1 is strictly worse than this pull request. With this pull request, you only have a sentinel (unit) when building with no extension. With 1, you would always have this sentinel lying around. I don't see what benefits 1 brings? We could always move the val specs from riscv_types_ext.sail to riscv_types_common.sail if you're concerned about that duplication, but otherwise I think this pull request is the pragmatic solution until someone has more time to do an extensive refactoring.

rmn30:
OK, was just musing without reading the code. Agree that a sentinel is an ugly workaround. Fine with me to merge and think about a matter solution later if we want one. @pmundkur  ?

PR #37, opened 2020-01-30
Invoke recursive make using $(MAKE) as GNU recommends

scottj97:
> Recursive make commands should always use the variable MAKE, not the
explicit command name ‘make’

See:
https://www.gnu.org/software/make/manual/make.html#MAKE-Variable


rmn30:
Thanks for this. We have a fairly homogeneous development environment so things like this can easily slip through. Is this required for Mac?

scottj97:
I don't know about Mac. This wasn't causing any problem for me; I noticed it by inspection.

PR #41, opened 2020-02-13
Report correct value in trace for CSR writes to frm and fflags

### LINKED ISSUE #40: trace shows incorrect value for frm/fflags
When I do a CSRW instruction to `frm` or `fflags`, which are sub-fields of the `fcsr` register, the trace shows the new `fcsr` value but reports it as the new value of `frm` or `fflags`.

For example if `fflags` is 2 and I do a CSRW to `frm` to write a value of 4, the new `fcsr` value is 0x82, but the trace reports that `frm` is now 0x82.

The patch pasted below will fix this, but I wonder if this is the best way.

Since these two subsets of `fcsr` can be accessed separately, would it be better to model them as two separate registers, instead of as bitfields within `fcsr`?

I think that would simplify many things in the code, and I also prefer that the same bits always show up in the trace under the same name. I find it confusing that sometimes `fcsr` is updated, and other times `fflags` is updated, when they refer to the same bits.

Though splitting it into two would have the complication that a single CSRW to `fcsr` would write two different registers and require two calls to print_reg(), which doesn't fit nicely in function `writeCSR` today.

If you think this patch is the best way I will submit it as a PR.

```
diff --git a/model/riscv_fdext_control.sail b/model/riscv_fdext_control.sail
index 7155699..ac3169f 100644
--- a/model/riscv_fdext_control.sail
+++ b/model/riscv_fdext_control.sail
@@ -17,8 +17,8 @@ function clause ext_read_CSR (0x001) = Some (EXTZ (fcsr.FFLAGS()))
 function clause ext_read_CSR (0x002) = Some (EXTZ (fcsr.FRM()))
 function clause ext_read_CSR (0x003) = Some (EXTZ (fcsr.bits()))
 
-function clause ext_write_CSR (0x001, value) = write_fcsr (fcsr.FRM(), value [4..0])
-function clause ext_write_CSR (0x002, value) = write_fcsr (value [2..0], fcsr.FFLAGS())
-function clause ext_write_CSR (0x003, value) = write_fcsr (value [7..5], value [4..0])
+function clause ext_write_CSR (0x001, value) = { write_fcsr (fcsr.FRM(), value [4..0]); Some(EXTZ(fcsr.FFLAGS())) }
+function clause ext_write_CSR (0x002, value) = { write_fcsr (value [2..0], fcsr.FFLAGS()); Some(EXTZ(fcsr.FRM())) }
+function clause ext_write_CSR (0x003, value) = { write_fcsr (value [7..5], value [4..0]); Some(EXTZ(fcsr.bits())) }
 
diff --git a/model/riscv_fdext_regs.sail b/model/riscv_fdext_regs.sail
index 59efd22..b746803 100644
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -273,12 +273,11 @@ bitfield Fcsr : bits(32) = {
 
 register fcsr : Fcsr
 
-val write_fcsr : (bits(3), bits(5)) -> option(xlenbits) effect {rreg, wreg}
+val write_fcsr : (bits(3), bits(5)) -> unit effect {rreg, wreg}
 function write_fcsr (frm, fflags) = {
   fcsr->FRM()    = frm;      /* Note: frm can be an illegal value, 101, 110, 111 */
   fcsr->FFLAGS() = fflags;
   dirty_fd_context();
-  Some (EXTZ (fcsr.bits()))
 }
 
 val write_fflags : (bits(5)) -> unit effect {rreg, wreg}

```

scottj97:
Fixes #40.

### DIFF
--- a/model/riscv_fdext_control.sail
+++ b/model/riscv_fdext_control.sail
@@ -17,8 +17,8 @@ function clause ext_read_CSR (0x001) = Some (EXTZ (fcsr.FFLAGS()))
 function clause ext_read_CSR (0x002) = Some (EXTZ (fcsr.FRM()))
 function clause ext_read_CSR (0x003) = Some (EXTZ (fcsr.bits()))
 
-function clause ext_write_CSR (0x001, value) = write_fcsr (fcsr.FRM(), value [4..0])
-function clause ext_write_CSR (0x002, value) = write_fcsr (value [2..0], fcsr.FFLAGS())
-function clause ext_write_CSR (0x003, value) = write_fcsr (value [7..5], value [4..0])
+function clause ext_write_CSR (0x001, value) = { write_fcsr (fcsr.FRM(), value [4..0]); Some(EXTZ(fcsr.FFLAGS())) }
+function clause ext_write_CSR (0x002, value) = { write_fcsr (value [2..0], fcsr.FFLAGS()); Some(EXTZ(fcsr.FRM())) }
+function clause ext_write_CSR (0x003, value) = { write_fcsr (value [7..5], value [4..0]); Some(EXTZ(fcsr.bits())) }
 
 /* **************************************************************** */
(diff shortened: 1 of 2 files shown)


pmundkur:
Thanks!

PR #43, opened 2020-02-22
RISC-V config integration

rmn30:
These changes integrate [riscv-config](https://github.com/riscv/riscv-config) into the sail-riscv model. It works by taking the RISC-V config yaml, processing it with the riscv-config tool then using a small ocaml program to convert the checked yaml into some sail code that instantiates values of types riscv_isa_config and riscv_platform_config. The model then uses these to implement the expected behaviour.

I've provided configs that match the existing default behaviour of the model. I checked in the checked yaml (generated by riscv-config) so that there is no additional dependency on riscv-config if you are happy with these default configurations.


pmundkur:
Looks generally fine to me.  I guess some of the csr configs still remain to be integrated, e.g. mstatus_config.  Perhaps we should move the legalize_* functions from riscv_sys_regs.sail to riscv_config_types.sail, since that is a more logical place for them.  That should also hide the cruft involved in legalizing mstatus.

rmn30:
I forgot about mstatus, did I miss some others? Difficulty with mstatus is that we can't really support anything other than hardwired MXL == SXL ==  UXL == xlen at present but I think we could handle the other fields.

I tried moving the legalize_* functions but ran into difficulty getting things defined in the right order. Will need some further shuffling and Makefile tweaking.

rmn30:
I just tried resolving the conflicts. They are not too bad but it looks like the master branch did something slightly different when legalising misa: if the new misa has D set but not F then master will drop the whole misa update but this branch will just clear D and proceed with the update. The way I read the spec. the latter is correct:
> If an ISA feature x depends on an ISA feature y, then attempting to enable feature x but disable
> feature y results in both features being disabled. For example, setting “F”=0 and “D”=1 results
> in both “F” and “D” being cleared.

scottj97:
6 of the 13 commits here are merges from master. May I request using `git rebase` instead of repeatedly merging? It helps to maintain a linear commit history and would make review a lot easier. See [The Case for Git Rebase](http://www.darwinweb.net/articles/the-case-for-git-rebase).

rmn30:
This PR is outdated because riscv-config has moved on. Work on riscv-config integration is taking place elsewhere (e.g. https://github.com/plctlab/sail-riscv/tree/plct-rvconfig-dev ) but I don't know of the current status.

martinberger:
> This PR is outdated because riscv-config has moved on.

So should we close this pull request? @ronorton

PR #45, opened 2020-03-30
Fix several pmpcfg issues

scottj97:
* pmp8cfg and pmp9cfg were not writable due to a typo in the source.

* Bits 5 and 6 of each 8-bit field must be masked to 0. (Is there a better way to do this using bitfield? I couldn't figure that out.)


pmundkur:
Thanks, great catches.

PR #48, opened 2020-04-16
Fix mstatus.MPRV fetches

### LINKED ISSUE #47: Instruction fetch erroneously affected by mstatus.MPRV
The `mstatus.MPRV` bit affects the effective privilege level of loads and stores. According to priv spec section 3.1.6.3 "Memory Privilege in mstatus Register":

> The MPRV (Modify PRiVilege) bit modies the privilege level at which loads and stores execute in all privilege modes. When MPRV=0, loads and stores behave as normal, using the translation and protection mechanisms of the current privilege mode. When MPRV=1, load and store memory addresses are translated and protected as though the current privilege mode were set to MPP. Instruction address-translation and protection are unaffected by the setting of MPRV.

But the Sail model seems to apply the modified privilege even to instruction fetch.

I believe the problem is in `pmp_mem_read`, where the `pmpCheck()` privilege level is specified. It should consider fetches differently than loads. Something like this already exists in `riscv_vmem_rv32.sail`:
```
  let effPriv : Privilege = match ac {
    Execute() => cur_privilege,
    _         => effectivePrivilege(mstatus, cur_privilege)
  };
```

scottj97:
Fixes #47

### DIFF
--- a/model/riscv_mem.sail
+++ b/model/riscv_mem.sail
@@ -64,7 +64,7 @@ function pmp_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext_
   if   (~ (plat_enable_pmp ()))
   then checked_mem_read(t, paddr, width, aq, rl, res, meta)
   else {
-    match pmpCheck(paddr, width, t, effectivePrivilege(mstatus, cur_privilege)) {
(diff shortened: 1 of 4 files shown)


pmundkur:
thanks!

PR #54, opened 2020-05-27
Fix mtval when store gets bad PMP

scottj97:
I observed that when a store instruction has invalid PMP permissions and takes an access fault, the `mtval` register was being written with the physical address. It should always be a virtual address.

I renamed some identifiers in the code in the first few commits to make the problem more obvious, then fixed it in the final commit.


pmundkur:
Good catch again, thanks!

scottj97:
I wonder, would it make sense (and is it possible) for vaddr and paddr to be different types, so the compiler can catch mixups like this?

pmundkur:
The address translation already uses different types, see the various `vaddr`,`paddr` definitions in `riscv_vmem_common`.  But it doesn't help in cases like this, where the mtval could genuinely store either as raw bits. It's up to the spec author to be careful.

PR #61, opened 2020-06-17
SV32: allow external specification of PTE extension bits

nwf:
Just because they're not present in the SV32 PTE doesn't mean that
models necessarily need to interpret them as zeros; any constant will do
just fine.  This allows extensions (like CHERI) that have both RV32 and
RV64 versions to define more standard idiomatic interpretation to the
bits within the PTE extension field.

### DIFF
--- a/model/riscv_vmem_sv32.sail
+++ b/model/riscv_vmem_sv32.sail
@@ -24,7 +24,7 @@ function walk32(vaddr, ac, priv, mxr, do_sum, ptb, level, global, ext_ptw) = {
     MemValue(v) => {
       let pte = Mk_SV32_PTE(v);
       let pbits = pte.BITS();
-      let ext_pte : extPte = zeros(); // no reserved bits for extensions
+      let ext_pte : extPte = default_sv32_ext_pte;
       let pattr = Mk_PTE_Bits(pbits);
       let is_global = global | (pattr.G() == 0b1);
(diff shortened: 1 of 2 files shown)


pmundkur:
Looking at this again, I'm not sure this does what you'd like it to do.  It seems like a minor cleanup instead of allowing `init_sv32_ext_pte` to be externally specified.

nwf:
Ah, do I need to put `init_sv32_ext_pte` in a different file so it can be overridden?

nwf:
While here, I moved the `extPte` type definition into `vmem_common` so we don't need to awkwardly redefine it in `cheri_pte`.

model/riscv_vmem_sv32.sail
```diff
@@ -24,7 +24,7 @@ function walk32(vaddr, ac, priv, mxr, do_sum, ptb, level, global, ext_ptw) = {
     MemValue(v) => {
       let pte = Mk_SV32_PTE(v);
       let pbits = pte.BITS();
-      let ext_pte : extPte = zeros(); // no reserved bits for extensions
+      let ext_pte : extPte = init_sv32_ext_pte;
```
pmundkur, line 27: Yes, this would now allow overriding by an extension.   Should we use `default_sv32_ext_pte` though?  I'd also put back the deleted comment but explain that there is a default for extension bits since there are no reserved bits.

PR #62, opened 2020-06-20
Allow extensions to override page fault causes

jrtc27:
Otherwise there's no use for PTW_Ext_Error. This is required for sail-cheri-riscv to be able to make use of its own page table exceptions. The alternative to this is to pass the ext_ptw to translationException, but given PTW_Ext_Error exists it seems this was the intended way, and aligns with how access faults vs page faults are distinguished.

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -14,7 +14,9 @@
 #include "elf.h"
 #include "sail.h"
 #include "rts.h"
+#ifdef SAILCOV
 #include "sail_coverage.h"
+#endif
 #include "riscv_platform.h"
 #include "riscv_platform_impl.h"
 #include "riscv_sail.h"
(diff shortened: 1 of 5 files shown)


jrtc27:
This should be trivial to review. If someone has a a chance to do so over the weekend that would be particularly appreciated, so we can get the CHERI PTE bits properly supported in sail-cheri-riscv and test Flute against it before we have to build a "final" bitstream for FETT.

bauereiss:
This does look trivial to me, but I haven't been following the development of the extension mechanism, so I'll leave it to others ( @pmundkur ?) to check whether this is fine as is or whether there should be more refactoring of the extension error reporting.  For now, I've pushed this to the `ptw-ext-error` branch of this repo, which should at least allow you to update the submodule in sail-cheri-riscv.

pmundkur:
looks good to me.

PR #63, opened 2020-06-23
A kinder, gentler splitting of ext_access_type's ReadWrite

nwf:
This redoes https://github.com/rems-project/sail-riscv/pull/57 without
nearly as much excitement.  We sill want it for CHERI, so that we can
signal from the instruction to the PTW whether we are prepared to load a
capability (or will strip any tags that we load) and whether the store
will (or might) set a tag.

### DIFF
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -189,10 +189,10 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
     /* Get the address, X(rs1) (no offset).
      * Some extensions perform additional checks on address validity.
      */
-    match ext_data_get_addr(rs1, zeros(), ReadWrite(Data), width) {
+    match ext_data_get_addr(rs1, zeros(), ReadWrite(Data, Data), width) {
       Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
       Ext_DataAddr_OK(vaddr) => {
-        match translateAddr(vaddr, ReadWrite(Data)) {
+        match translateAddr(vaddr, ReadWrite(Data, Data)) {
           TR_Failure(e, _) => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
           TR_Address(addr, _) => {
             let eares : MemoryOpResult(unit) = match (width, sizeof(xlen)) {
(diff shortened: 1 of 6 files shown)


model/riscv_pte.sail
```diff
@@ -42,23 +42,32 @@ function to_pte_check(b : bool) -> PTE_Check =
...
+  let update_d : bool = match a {
+                          Read(Data)      => false,
+                          Execute()       => false,
+                          Write(Data)     => true,
+                          ReadWrite(Data) => true
+                        };
+  let update_d = update_d & p.D() == 0b0;
+
+  // accessed bit
+  let update_a = p.A() == 0b0;
```
jrtc27, line 70: This seems unnecessary, and also gets the rarity of ReadWrite data still wrong. I'd just add `, Data` to the existing code (and at some point I'd like to refactor this to match on `Read(_)` etc and call an `ext_update_PTE_Bits` so we can avoid duplicating the code in sail-cheri-riscv, with suitable changes to `checkPTEPermission` etc).

model/riscv_insts_aext.sail
```diff
@@ -189,10 +189,10 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
     /* Get the address, X(rs1) (no offset).
      * Some extensions perform additional checks on address validity.
      */
-    match ext_data_get_addr(rs1, zeros(), ReadWrite(Data), width) {
+    match ext_data_get_addr(rs1, zeros(), ReadWrite(Data,Data), width) {
```
jrtc27, line 192: (and everywhere else) missing space

model/riscv_pte.sail
```diff
@@ -42,23 +42,23 @@ function to_pte_check(b : bool) -> PTE_Check =
...
 
     (_,               Machine)    => internal_error("m-mode mem perm check")
   }
 }
 
 function update_PTE_Bits(p : PTE_Bits, a : AccessType(ext_access_type), ext : extPte) -> option((PTE_Bits, extPte)) = {
-  let update_d = (a == Write(Data) | a == ReadWrite(Data)) & p.D() == 0b0; // dirty-bit
-  let update_a = p.A() == 0b0;                                             // accessed-bit
+  let update_d = (a == Write(Data) | a == ReadWrite(Data, Data)) & p.D() == 0b0; // dirty-bit
+  let update_a = p.A() == 0b0;                                                   // accessed-bit
```
jrtc27, line 61:
```suggestion
  let update_d = (a == Write(_) | a == ReadWrite(_, _)) & p.D() == 0b0; // dirty-bit
  let update_a = p.A() == 0b0;                                          // accessed-bit
```

nwf: Does `_` matching work with `==` like that?  I'll kick it over to being `match` expressions if not.

jrtc27: Oh, probably not...

pmundkur:
looks good to me, thanks.  let me know if you'd like me to pull this or whether i should wait for more changes.

nwf:
I think this PR is ready to go.  I'll have other changes, but they'll depend on this one.

PR #64, opened 2020-06-30
Split ext_ptw into pieces and add documentation

nwf:
It was somewhat annoying that CHERI would have had to track both
successful and erroneous occurrences within the same type.  Break the
failures out to their own form and, while here, attempt to write down my
understanding of the extension's information flow through vmem.


pmundkur:
thanks, especially for the docs!

PR #75, opened 2020-11-16
Initial introduction of zfinx

abukharmeh:
This request introduces ZFinX extension (https://github.com/riscv/riscv-zfinx/blob/master/Zfinx_spec.adoc) to the model
 
Huawei Technologies Research & Development (Uk) Limited, November 2020


model/riscv_insts_fext.sail
```diff
@@ -588,18 +627,23 @@ mapping clause encdec =
...
   let (fflags, rd_val_S) = riscv_f32Sqrt   (rm_3b, rs1_val_S);
 
   write_fflags(fflags);
-  F(rd) = nan_box (rd_val_S);
+
+  if   ( sys_enable_zfinx() & sizeof(xlen) == 64 )
+  then X(rd) = x_nan_box (rd_val_S)
+  else if ( sys_enable_zfinx() & sizeof(xlen) == 32 )
+  then X(rd) = rd_val_S
+  else F(rd) = nan_box (rd_val_S);
```
scottj97, line 641: This paragraph of code is duplicated at least 9 times in this file. I don't know the Sail language that well, but surely there is a way to avoid all this copypasta?

model/riscv_insts_fext.sail
```diff
@@ -447,9 +468,11 @@ mapping clause encdec =
 /* Execution semantics ================================ */
 
 function clause execute (F_MADD_TYPE_S(rs3, rs2, rs1, rm, rd, op)) = {
-  let rs1_val_32b = nan_unbox (F(rs1));
-  let rs2_val_32b = nan_unbox (F(rs2));
-  let rs3_val_32b = nan_unbox (F(rs3));
+
+  let rs1_val_32b = if (sys_enable_zfinx()) then  x_nan_unbox (X(rs1)) else nan_unbox(F(rs1));
```
scottj97, line 472: This if statement appears 27 times in this file.

model/riscv_insts_fext.sail
```diff
@@ -518,7 +550,14 @@ function clause execute (F_BIN_RM_TYPE_S(rs2, rs1, rm, rd, op)) = {
...
   };
   write_fflags(fflags);
-  F(rd) = nan_box (rd_val_32b);
+  
+
+  if   ( sys_enable_zfinx() & sizeof(xlen) == 64 )
+  then X(rd) = x_nan_box (rd_val_32b)
+  else if ( sys_enable_zfinx() & sizeof(xlen) == 32 )
+  then X(rd) = rd_val_32b
+  else F(rd) = nan_box (rd_val_32b);
```
scottj97, line 559: This paragraph of code appears 8 times in this file. Surely there's a better way.

scottj97 (changes requested):
1. This is a lot to review in a single commit. It would be preferable to break it up into many smaller commits.
2. This appears to include changes to support XLEN != FLEN. Is that so? If so that should be a separate commit or even a separate pull request.
3. There is extreme copypasta in the implementation. I'm not that familiar with the Sail language but I hope there is a better way to do this than copy/pasting the same code all over. I've added individual comments on several of them but there are many more cases that I did not comment.

pmundkur:
Thanks for implementing the Sail model of the Zfinx (or ZfinX?) extension, and for your pull request!

Some suggestions to improve your Zfinx model:

-  I agree with Scott's suggestion to reduce the amount of duplication.  One way to do that would be to write functions that implement your requirements for the register reads and writes, and put them in `riscv_fdext_regs.sail`.  That file already contains a function that abstracts handling the write to the `fcsr` register to avoid similar duplication.

- The Zfinx spec says the extension "removes" all floating point load/store instructions (e.g. FLW, FSW) and moves between integer and floating-point registers.  However, your Sail model does not modify these instructions.

  If the intent of the word "remove" in spec is these instructions are now illegal, you can modify the encdec clauses of STORE_FP/LOAD_FP to ensure that Zfinx is disabled (similarly to how they currently handle RV32/RV64).  The same would need to be done for the various versions of FMV.

- The Zfinx spec says that the assembly syntax of the F/D instructions would need to refer to the integer registers instead; this would require adjusting the register naming functions in `riscv_fdext_regs.sail` to handle Zfinx.

- The discovery aspect of the spec doesn't seem to be handled (i.e. adjusting the values of mstatus.FS and misa.F).

- If you have executable test cases for these instructions, it would be great if you could add them under `test/zfinx-tests`.

There are minor syntactic issues like a few cases of broken indentation and trailing whitespace (please remove!) that could be made when incorporating these suggestions.

abukharmeh:
Hello Prashanth, 

Thank you very much for your comments, they were quite helpful.  I am currently working on all the issues listed, and I faced an Issue I would like your input on. 

I am trying to implement a global immutable variable of type bit, that is assigned to the system argument that we used to enable ZFinX, and then use that for mapping statements through out the code. 
**e.g** `let zfinx_en = bool_to_bits(sys_enable_zfinx())`  
However, I discovered that if this is used in the global scope, the value stays uninitialised. If I use the same code to assign a local variable, it gets the correct value..  

Quickly reading through the generated C emulator, it appears that global variables are assigned before emulator arguments are parsed, if this is the case, how is it possible to assign an immutable global variable to a value passed as an argument to the simulator ?

Kind regards,
Ibrahim

pmundkur:
`riscv_platform` has examples of getting access to externally specified platform and system parameters, for e.g., `plat_enable_pmp`.  It looks similar to what you are trying to do with `sys_enable_zfinx`.  Would that work for you?

abukharmeh:
Hello Prashanth, 

Sorry, I think I might not have explained myself correctly, the functions in riscv_platform are very similar to my function, they call a function in riscv_platform.c which would return a Boolean value, which is also exactly the same as my function does, and that works correctly. I tried to find any global assignments dependent on any of these input arguments, but  I cannot find any

I dont want to call that C function on every call to a floating point operation, and I think I would rather assign it to a global immutable variable.  

Looking at the generated C emulator, it appear that all global assignments are checked and assigned before input arguments are parsed, is this assumption correct ?

abukharmeh:
This is what I was thinking of doing, but was wondering what is the reason behind initialising the model before parsing the input arguments ? 

Its not strictly necessary, however since I need that value in many mapping statements, and I would want to avoid calling the function and converting its returned value on every floating point operation, I was thinking it make sense just to have a global variable always assigned the correct value in the correct representations, especially since this value cannot be changed midway through the emulation.

pmundkur:
I see what you mean.  To answer your earlier question about initialization, the simulator does a consistency check to ensure that it loads an elf file that matches the RV32/RV64 architecture it was generated for.  To do this, it needs to initialize the xlen_val value; so the simulator has a preinit_sail() where all such global variables are initialized before the command line arguments are parsed.  You could set the value of `zfinx_en` from C once you know what it should be, after parsing the command line args; but that would make `sys_enable_zfinx()` redundant.  

I'd use the less performant and repetitious way for now.  The way to configure the model will probably change anyway to get these values from a yaml file.

abukharmeh:
I fixed all of the issues that were pointed out earlier, but I am stuck on trying to reduce the repetitions in the code, and I am wondering if you can help me with this ? 

I am trying to use forall to make a single function that would return the correct register depending on if the zfinx is enabled, the following function works correctly, however I am not sure how one would do such implementation using bits(1) or Boolean input as that cannot be size dependent as far as I am concerned, any input would be appreciated ?

`val      f_reg_zfinx : forall 'n, ('n==0). (int('n),bits(5)) -> bits(flen) effect  {escape, rreg}
function f_reg_zfinx(zfinx_en,index) =    F(index) 
val      f_reg_norm :  forall 'n, ('n==1). (int('n),bits(5)) -> bits(xlen) effect  {escape, rreg}
function f_reg_norm(zfinx_en,index) =   X(index) 
overload f_reg = {f_reg_zfinx,f_reg_norm}
`

abukharmeh:
The following issues were addressed in the previous two commits

1. Updates the register names printed in trace for when ZFINX is enabled. (Since instruction are handled with scattered mapping statements, it’s hard to introduce any function that would handle the register name change for ZFINX unless we have an immutable global variable that tells if ZFINX is enabled, however since the model initialise and assign all global variables before parsing the input command line arguments, it is impossible to assign such global variable with command line arguments (https://github.com/rems-project/sail-riscv/pull/75#issuecomment-736002604).

To overcome this issue, I directly assign the flag value from C after argument parsing, this is not the best solution, however, it appears that it’s the only one if I want to avoid duplicating every single mapping statement for floating point operations. 

2. Remove floating point loads, stores and moves.
3. Updates the handling of mstatus.FS and mstatus.F
4. Clear the code a bit (in an effort to reduce code duplications). 
5. Adds the test binaries we used to test the introduced functionality (Please note that the objdump files have the incorrect syntax for the register naming, as they were created using an old toolchain, I will update them in a next commit along with more code refactoring when I get some help with forall).

.gitignore
```diff
@@ -5,3 +5,4 @@
 _build/
 _sbuild/
 *.o
+*.patch
```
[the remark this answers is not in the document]
abukharmeh: Sorry, I agree, this does not belong here, I will remove it.

Makefile
```diff
@@ -1,28 +1,44 @@
-# Select architecture: RV32 or RV64.
-ARCH ?= RV64
+# Select architecture
+ifeq ($(ARCH),RV32)
+  override XLEN := 32
+  override FLEN := 32
```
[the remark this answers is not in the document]
abukharmeh: In the original Makefile, it was only possible to specify either RV32 which would result in XLEN=FLEN=32, or RV64 which would result in  XLEN=FLEN=64.  My modifications allow for backward compatibility, thus if the same value of ARCH was passed to the Makefile it would result in the same behavior, but it also allows for the FLEN and XLEN values to be specified explicitly as 64 or 32 to allow for the various other cases.

Makefile
```diff
@@ -1,28 +1,44 @@
...
+else ifeq ($(ARCH),RV64)
+  override XLEN := 64
+  override FLEN := 64
+else ifneq ($(and $(XLEN),$(FLEN)),)
+else ifndef $(ARCH)
+  override XLEN := 64
+  override FLEN := 64
+else 
+  $(error 'No valid architecture detected')
```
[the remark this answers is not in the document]
abukharmeh: Same as my previous comment on Makefile, though I can change the error to specify that the passed XLEN or FLEN is not supported.....

model/riscv_insts_fext.sail
```diff
@@ -1033,11 +1112,16 @@ function clause execute (F_UN_TYPE_S(rs1, rd, FMV_X_W)) = {
...
+  if   ( sys_enable_zfinx() & sizeof(xlen) == 64 )
+  then X(rd) = x_nan_box (rd_val_S)
+  else if ( sys_enable_zfinx() & sizeof(xlen) == 32 )
+  then X(rd) = rd_val_S
+  else F(rd) = nan_box (rd_val_S);
   RETIRE_SUCCESS
 }
 
-/* AST -> Assembly notation ================================ */
+// /* AST -> Assembly notation ================================ */
```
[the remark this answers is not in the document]
abukharmeh: Sorry, this is likely inserted by VSCode when I was commenting a complete block, I will remove it.

c_emulator/riscv_sail.h
```diff
@@ -32,6 +32,7 @@ void zrvfi_get_exec_packet(sail_bits *rop, unit);
 #endif
 
 extern mach_bits zxlen_val;
+extern int64_t zzzfinx_en;
```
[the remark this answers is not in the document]
abukharmeh: Yes SAIL generates an int64_t  for a declared Boolean value, I was surprised why it does this as well.

I am only externing the automatically generated code to allow to change it from the command line arguments, I saw it done this way in another argument.  (I don't know why we do it this way, and I asked previously what guarantees that the automatically generated vars names would stay consistent when we recompile the code, but I did not get any answer, and it appears to be deterministic for all SAIL versions I tried).

model/riscv_fdext_regs.sail
```diff
@@ -187,40 +187,72 @@ overload to_str = {freg_name_abi}
 
 /* mappings for assembly */
 
-val freg_name : bits(5) <-> string
+val freg_name : bits(6) <-> string
```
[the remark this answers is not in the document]
abukharmeh: I will try to change that to a tuple and delegate to a reg_name like you suggested.
Its just the manual for SAIL does not list every keyword, type, option like other languages, so often I was guessing from reading existing code.

model/riscv_insts_fext.sail
```diff
@@ -285,11 +307,11 @@ union clause ast = LOAD_FP : (bits(12), regidx, regidx, word_width)
 
 /* AST <-> Binary encoding ================================ */
 
-mapping clause encdec = LOAD_FP(imm, rs1, rd, WORD)          if is_RV32F_or_RV64F()
-                    <-> imm @ rs1 @ 0b010 @ rd @ 0b000_0111  if is_RV32F_or_RV64F()
+mapping clause encdec = LOAD_FP(imm, rs1, rd, WORD)          if (is_RV32F_or_RV64F() & (sys_enable_zfinx() == false))
```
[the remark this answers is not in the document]
abukharmeh: I am following conversation about that in Zfinx  repo, and I might change it to return false as you said.

test/zfinx-tests/rv32udzfinx-p-fadd.dump
```diff
@@ -0,0 +1,518 @@
+
```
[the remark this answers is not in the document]
abukharmeh: I modified the original floating point tests to use the correct register names for Zfinx and handle register pairs etc. I will move them to retain the flat directory structure.

abukharmeh:
Hello Jessica, 

Thank you very much for your comments. Sorry for late reply, I was trying to see what will happen with this issue "https://github.com/riscv/riscv-zfinx/issues/8 " before replying. 

Regarding formatting issues (trailing white spaces, empty lines, etc), I am sorry for these, I will go through the PR and make sure I fix them all. 

Regarding other issues, please check my inline replies. 

Kind regards,
Ibrahim.

abukharmeh:
Hi Jessica @jrtc27,
I wonder if you had a chance to look at the formatted code ? 
Thanks

martinberger:
> Hi Jessica @jrtc27, I wonder if you had a chance to look at the formatted code ? Thanks

 @jrtc27 and @scottj97 I would like to merge this in soon, next week ideally. Do you think you will be able to look at Ibrahim's code?

jrtc27:
Definitely not ready for merging, the way X_or_F or whatever is done is both broken and a layering violation, I just haven’t had a chance to write It up yet

martinberger:
> Definitely not ready for merging, 

Thanks. I'll ping Ibrahim

abukharmeh:
Can you give me at least some idea/ suggestion of the issues you see, so I try to sort them out ?

model/riscv_fdext_regs.sail
```diff
@@ -210,6 +245,67 @@ function wF_bits(i: bits(5), data: flenbits) -> unit = {
 
 overload F = {rF_bits, wF_bits, rF, wF}
 
+val handle_illegal : unit -> unit effect {escape, rreg, wreg}
+
+/* ---- Register Read/ Writes */
+
+val rX_or_rF_32 : (bool, bits(5)) -> (Retired, bits(32)) effect {escape, rreg}
```
jrtc27, line 252: The overload is X_or_F so this should be rX_or_F, no second r

model/riscv_fdext_regs.sail
```diff
@@ -210,6 +245,67 @@ function wF_bits(i: bits(5), data: flenbits) -> unit = {
...
+function rX_or_rF_32(zfinx_en, i) = {
+  let rval = if zfinx_en then nan_unbox(zfinx_en, rX(unsigned(i))) else nan_unbox(zfinx_en, rF(unsigned(i)));
```
jrtc27, line 254: Just use X/F, then you don't need the unsigned conversion.

This would also benefit from being split into multiple lines.

model/riscv_fdext_regs.sail
```diff
@@ -210,6 +245,67 @@ function wF_bits(i: bits(5), data: flenbits) -> unit = {
...
+  }
+  else if zfinx_en & sizeof (xlen) == 64 then
+    (RETIRE_SUCCESS, rX(unsigned(i)))
+  else if ~(zfinx_en) & sizeof (flen) == 64 then
+    (RETIRE_SUCCESS, rF(unsigned(i)))
+  else {
+    assert(false);
+    (RETIRE_FAIL, sail_zeros(64))
+  }
+}
```
jrtc27, line 274: So, this function (and the write version, though one argument doesn't apply) is where all the real problems lie.

Firstly, `(RETIRE_FAIL, sail_zeros(64))` should be a huge red flag saying "the types are wrong"; that's how you write it in Verilog, but it's not how you write it in a language with a proper type system, you don't just shove dummy value into things with a separate validity field, you use a proper discriminated union, which means enum in Sail and other modern languages; in this case there's option already in the language that does exactly what you want.

Secondly, the use of RETIRE_SUCCESS/FAIL here is a layering violation. If you look throughout the model, RETIRE_SUCCESS/FAIL are only ever used in places that are exclusively about instruction execution, and within the riscv_insts_* files themselves (ignoring iris, which is a load of files all concatenated together, and has been removed, and riscv_jalr_rmem/seq which are just there so two different implementations of JALR's execute clause can be used). There is never a pattern like this. Functions like this should be free from the semantics of instruction retiring; for example, if someone were to extend the model with a debug module implementation, register reading and writing is something that it can do (via its abstract commands) but there is no notion of retiring there. The correct design is to have this be independent of instruction execution semantics and let the consumers deal with translating that into RETIRE_FAIL on error, which is then consistent with how we deal with things like memory translation errors and extension hooks; again, this points towards the use of an option type.

However, both of those ignore the more important actually-broken behaviour when you look at the bigger picture of how this function is used during instruction execution. Many of the floating-point instructions also update fflags. But write_fflags gets called before F(rd) is assigned. So `fadd.d x3, x6, x8` on RV32 will fault with an illegal instruction exception, as it should, but already have updated fflags. You could "fix" this by swapping the order, but that is a gross hack. The correct thing to do is to push the actual check that the registers are valid into the guards on the encdec clauses or as the very first thing the execute clauses do, and then these functions can just return the value directly, no Retired or option needed. They should probably still have _asserts_ that the input is sane though to catch any caller that forgot to check the registers.

jrtc27 (commented):
I still have more stylistic problems with the way in which the register functions have been implemented, but I don't think it's worth discussing those now as the code needs to be rewritten anyway which may well render them no longer applicable.

jrtc27:
I'll also note that you don't implement the bit of Smstateen that is required for Zfinx, namely bit 1 of xstateen0, that plays the same role as mstatus.FS does for F.

jrtc27:
Well Zfinx depend on Smstateen, so you'll have to wait then, implementing Zfinx without Smstateen violates architectural security properties

tariqkurd-repo:
The Zfinx spec does not require Smstateen, and can you show me where these "architectural security properties" are defined, and a policy that means we can't update the SAIL model as a result?

I agree that Smstateen is highly advised but it's *not* required for merging Zfinx, and is not through public review so it can't be merged into the model.

You are taking exception to the Zfinx spec. The SAIL model implements the (soon to be ratified - in fact it should be this month) Zfinx spec. You raised this issue which Andrew closed, but no changes were made to the Zfinx spec itself as a result. https://github.com/riscv/riscv-zfinx/issues/13

The model matches the spec and will be suitable for merging once Ibrahim has fixed the RV32D bug that you very diligently pointed out, thank you for that. It was certainly a corner case and very well spotted.

martinberger:
> _no_ changes were made to the Zfinx spec itself as a result. 

It sounds like we are very close to a resolution and can merge this soon!

jrtc27:
> The Zfinx spec does not require Smstateen, and can you show me where these "architectural security properties" are defined

"Extensions must not add state that can be written to by unprivileged code without privileged code having explicitly opted into it" is a fundamental architectural design principle, and is the entire motivation for Smstateen's existence.

> and a policy that means we can't update the SAIL model as a result?

Because it introduces a new bug, namely a glaring security hole.

> I agree that Smstateen is highly advised but it's _not_ required for merging Zfinx

Unless you limit Zfinx to M-mode-only (which circumvents the security issue, but is then not a complete implementation of Zfinx), it absolutely is.

> and is not through public review so it can't be merged into the model.

I have no problem with tracking the draft in the model; we had a PR for scalar crypto that only wasn't merged due to issues other than it being a draft.

> You are taking exception to the Zfinx spec. The SAIL model implements the (soon to be ratified - in fact it should be this month) Zfinx spec. You raised this issue which Andrew closed, but no changes were made to the Zfinx spec itself as a result. [riscv/riscv-zfinx#13](https://github.com/riscv/riscv-zfinx/issues/13)

Because Smstateen filled the gap; the issue was closed by him with "JohnH already thought this one all the way through, and I support his line of reasoning", which was "This is already covered by the state-enable proposal". Note Andrew also said "I agree the side-channel matter is a problem".

> The model matches the spec

The spec needs fixing to make sure this dependency is documented.

> and will be suitable for merging once Ibrahim has fixed the RV32D bug that you very diligently pointed out, thank you for that. It was certainly a corner case and very well spotted.

I'll be the judge of that, I suspect there will still be other code quality issues to address in the rewritten code, just not as glaring as that one, as, like I said in my review, there are other stylistic issues (design and implementation, not whitespace) in the current patch that don't make sense to comment on until the code's been rewritten, since the specifics will change.

I get that Zfinx is your spec and you want it to be merged, but I'm not going to be pressured into approving the merge something that breaks the architecture or is not up to the quality needed in a reference model. Sorry but that's just how it is.

martinberger:
> Sorry but that's just how it is.

I think we should carefully distinguish between the approved ZFinx spec (which may need refinement) and how the PR'ed Sail matches the  approved ZFinx spec.

jrtc27:
> > Sorry but that's just how it is.
> 
> I think we should carefully distinguish between the approved ZFinx spec (which may need refinement) and how the PR'ed Sail matches the approved ZFinx spec.

It's not official until it's ratified by the board. If the security hole gets ratified then we, and the community, will have to deal with it and its implications, and the ratification process will have failed RISC-V, but I'm not going to approve something that is known to break the security of the architecture until then. That's very different from a random draft spec that's going to see future changes but isn't a security hole or otherwise breaks parts of the architecture (bad encodings, unwanted instructions, missing instructions, strange semantics, etc).

tariqkurd-repo:
Again, it's a simple separation of concerns. 

Whether the Zfinx spec is sufficient or not has no bearing on whether the SAIL matches it.

We are following the process. If you think that the Zfinx spec must not be ratified in its current form then it's another matter which is not for this forum. this needs to be resolved separately.

If a spec update _is_ required then I have no issue with updating the model afterwards.

But as it stands the model matches the frozen spec.

jrtc27:
> Again, it's a simple separation of concerns.
> 
> Whether the Zfinx spec is sufficient or not has no bearing on whether the SAIL matches it.

Sure. But that doesn't mean we have to merge support for any pre-ratification spec here. To take an extreme example, if a TG decided to put out a draft spec that introduced a "set the current mode to M" instruction and filed a PR here then I'm not going to approve that. Now that's obviously stupid and not going to happen, but you get my point. I take the hard line of "no security holes unless ratified", because I see anything else as negligent.

> We are following the process.

Sure. But this is why I brought the issue up during public review, and it was supposedly accepted as a problem and then resolved by Smstateen. So I'm disappointed that three months later the issue still exists and I'm having to have the same argument all over again, rather than it jut being fixed properly the first time round.

> If you think that the Zfinx spec must not be ratified in its current form then it's another matter which is not for this forum. this needs to be resolved separately.

I don't mind taking this elsewhere. It's probably a better way to loop in Krste/Andrew/John if you feel that's necessary.

> If a spec update _is_ required then I have no issue with updating the model afterwards.
> 
> But as it stands the model matches the frozen spec.

I'm not debating that given Zfinx has no mention of Smstateen (nor does it separately define the subset of the extension relevant to it, as was done to expedite scalar crypto despite reusing parts of bitmanip), just as it stands implementing the non-ratified spec breaks the architecture's security and so should not be merged unless ratified or fixed.

tariqkurd-repo:
Hi @jrtc27 I forwarded your concerns to the arch review committe,and have a reponse for you, from Greg and Andrew:

- It is not appropriate for Unpriv extensions to depend on Priv extensions.  This is generally avoided.  Unpriv extensions are supposed to be pure User-level ISA extensions.

- The Priv Smstateen extension is the place for defining a bit for Zfinx - and related Z\*inx extensions - to address this side-channel security issue.  And that extension DOES define a bit (bit 1) for ALL the Z\*inx extensions.

- As separate extensions, these allow flexibility for the wide range of RISC-V implementations - including ones that choose to implement Zfinx and not Smstateen (i.e. the side-channel security concern is not relevant for some embedded systems).  At an ISA level we don't want to be forcing implementers of Zfinx to have to also implement Smstateen.  Bundling the preceding Unpriv and Priv functionality into one extension would hinder or prevent this.

- Requiring that implementation of Z*inx also then requires implementation of Smstateen, is not something that should ever be required at an ISA level, but at a platform level (for implementations that choose to be compliant with that platform).


I hope that clarifies RISC-V's position on this. The tieing together of extensions belongs at the platform level.

Tariq

model/riscv_fdext_regs.sail
abukharmeh: Hi @jrtc27 

I am wondering if there is a better way of achieving the following snippet 

`
overload operator | = {or_vec,or_two_bits,or_bit_and_bool}
val or_two_bits : (bit,bit) -> bool
function or_two_bits (x,y) = {
  bit_to_bool(x) | bit_to_bool(y)
}

val or_bit_and_bool : (bit,bool) -> bool
function or_bit_and_bool (x,y) = {
  bit_to_bool(x) | y
}`

I am trying to check that non of any number of `bit` (be it even or odd is bitone), and I ended up with this but maybe there is a better way of achieving the same thing ?  Would this be accepted in the model ? 

Regards,
Ibrahim

jrtc27: That snippet is completely wrong, `bit | bit` is a bit, `bool | bool` is a bool and any other combination is incorrect code, because bits and bools are completely different things.

But I also don't understand what that has to do with what you're saying about "any number of bit", nor what you're actually saying.

abukharmeh:
```
mapping clause encdec =
    F_BIN_RM_TYPE_D(rs2, rs1, rm, rd, FADD_D)
    if (is_RV32D() & ( ~(sys_enable_zfinx()) | (rs1[0] | rs2[0] | rd[0]) == bitzero)) | is_RV64D()
<-> 0b000_0001 @ rs2 @ rs1 @ encdec_rounding_mode (rm) @ rd @ 0b101_0011
    if (is_RV32D() & ( ~(sys_enable_zfinx()) | (rs1[0] | rs2[0] | rd[0]) == bitzero)) | is_RV64D()
```
Hi @jrtc27  I am considering adding the following conditions to each RV32/64D encdec clause, does it looks all right ? I am not sure if this would be considered too repetitive ?

jrtc27: That looks correct but yes you should extract the valid register check into a function

abukharmeh: I was thinking that, but I am not sure that would help much, because either I would create a single function that checks a single register which would be almost the same, or I would create a single function that takes all the registers and return a boolean, but then I would have to create another one for fmadd case and does not feel much better.

jrtc27: So either have a `bits(5) -> bool` and a trio of `(bits(5), bits(5)) -> bool`, `(bits(5), bits(5), bits(5)) -> bool` and `(bits(5), bits(5), bits(5), bits(5)) -> bool` functions (you can use an overload for them all) or have a single `forall 'n, 'n > 0. vector('n, dec, bits(5)) -> bool` function.

abukharmeh:
```
val registers_even : forall 'n, 'n > 0. (vector('n, dec, bits(5))) -> bool
function reg_even (registers) = {
  foreach (x from 0 to (length(registers) - 1)) {
    if (registers[x][0] == bitone) then return false
  };
  true
}
```
@jrtc27  I tried this, SAIL happily compiles it, but it result in a linker error ` undefined reference to length_zz5vecz8z5bv5z9`, I will for now pass the length of the vector, and I will report this as a bug to the SAIL compiler, unless you see something wrong with it, PS: I import vector_dec.sail

jrtc27: Yeah ok I see the problem, ran into somewhat similar things before.

You don't quite need an actual parameter though, you can use `implicit('n)` so as far as the callers are concerned there's only the vector argument.

jrtc27: And don't make the function a registers_even function (unless it's a helper for the helper), pull _all_ the logic out into a function otherwise you'll still need to repeat the `~(zfinx) | rv64 | even(...)` mess everywhere. This way it's just an "are these floating point registers valid operands for a double operation" function.

abukharmeh: Thanks, implicit appears to have worked.
Yes, I am planning to put it all in a single function, was trying to see how liquid types work with vectors first before placing the rest of the logic in that function..

martinberger: > use `implicit('n)`

abukharmeh: @jrtc27  Going back to the `proper discriminated union`, if I use option(bits(64)), then I would need to either use Some on all cases where I call X_or_F_D or need to probably check for, and since I already moved the checking to encdec statements, I am not sure this really make sense as it should in theory never happen ? Any suggestion if this is really necessary.

jrtc27: Since it should never happen, make it an assert. The reason I talked about using an option type, rather than just mentioning the trap behaviour issue and ignoring the no-longer-relevant types + layering violation issue, was so that you're aware of all the issues with your original approach and thus better informed for any future work.

abukharmeh: My usage for those dummy values are like this one : 
https://github.com/riscv/sail-riscv/blob/master/model/riscv_insts_fext.sail#L379-L382

martinberger: > here is a layering violation

I think what @jrtc27  means is that you have a return type `... -> option(unit)` (and possibly drop the `escape` effect) and then let the caller decide what to do with the failed call. This is widely used in the RISCV code base, e.g. in https://github.com/riscv/sail-riscv/blob/master/model/riscv_insts_fext.sail#L375-L382 (note that the last `RETIRE_FAIL` there is also a dummy value to keep the type-checker happy). In an even more expressive language, you'd probably use a monad to avoid the inevitable code duplication at call sites, but Sail doesn't have those.

@jrtc27 is that what you have in mind?

abukharmeh: Ok, I already have an assert statement, but I have to use such dummy values for the type checker to be happy !  I will update the PR shortly for other issues.

jrtc27: Which lives in riscv_insts_fext.sail, not riscv_fdext_regs.sail, is under `Execution semantics` and specifically a helper function for execution floating-point load instructions to factor out common code, not a general "read a register" function that can have other uses.

jrtc27: Though, whilst the layering is fine there, it should still be an assert in that function, rather than silently becoming a NOP that doesn't increment minstret

jrtc27: > Ok, I already have an assert statement, but I have to use such dummy values for the type checker to be happy ! I will update the PR shortly for other issues.

You should not need to; if you do you're almost certainly doing something wrong.

martinberger: > it should still be an assert in that function, 

This is a good place for me to ask something I've long wondered about: what's the official RISCV guideline for using `assert` in code?

abukharmeh: > You should not need to; if you do you're almost certainly doing something wrong.

What do you mean, even if I have assertion, I would need to return some value for type checker to be happy ?

jrtc27: > > You should not need to; if you do you're almost certainly doing something wrong.
> 
> What do you mean, even if I have an assertion, I would need to return some value for type checker to be happy ?

If you have an assertion, that case cannot possibly progress past the assertion, therefore you do not need to handle it.

abukharmeh: Yes I know, but type checker wont compile it if I dont have a return value of the correct type, I guess thats why they did it in https://github.com/riscv/sail-riscv/blob/master/model/riscv_insts_fext.sail#L379-L382. It would never get there, but we still need it.

jrtc27: Why are you trying to add an if for something you just asserted can't happen? That's the whole point, you don't handle the condition, you just assume it can't happen because if it did you've already crashed with an assertion failure.

abukharmeh: Because it would be an incomplete if clause, and there there would be no return value. The use case I am trying to have an assertion for is when you try to read a 64 bit (X_or_F_D) and zfinx is not enabled.  it should never reach there because of conditions on encdec, but I think it still need to be there.

jrtc27: As I said, why are you trying to add an if? It serves no purpose other than to make your life harder. You know it will always (or never, depending which way round you write it) be taken. The whole point of asserts is so that there is no if, and having both an assert and an if for the same condition is a serious code smell.

abukharmeh: I guess I should just convert the last case of the if clauses to else, and just have the assertions.

jrtc27: ... yes that's what I've been saying.

Note that there are no "if clause*s*" plural. There's if and there's else; else if is just else (if ....), there is nothing in the grammar for "else if". This is also true in C-based languages.

abukharmeh: https://github.com/riscv/sail-riscv/blob/01720566318e292df12693bba1fe5f09ea0857df/model/riscv_insts_dext.sail#L946-L947, Is there any reason why haveDExt() is used instead of is_RV32D_or_RV64D in these lines ? This seems slightly inconsistent, I think I would assume the conditions for this one should match feq.d etc.

abukharmeh: PS: Now I am working based on the assumption that regex `\w,\w` should not return any results in the model, and every time there is a comma in a function declaration, vector declaration, argument passing etc, there should be a space after it. This assumption seems mostly true with the exception riscv_termination_common.sail !

abukharmeh: PS: Now I am working based on the assumption that regex `\w,\w` should not return any results in the model, and every time there is a comma in a function declaration, vector declaration, argument passing etc, there should be a space after it. This assumption seems mostly true with the exception riscv_termination_common.sail !

abukharmeh: https://github.com/riscv/sail-riscv/blob/01720566318e292df12693bba1fe5f09ea0857df/model/riscv_insts_cdext.sail#L79-L82 
Similarly, the conditions in this file should have been is_RV32D_or_RV64() instead of redefining it across the file, but I wont fix it in this PR as its not directly related to Zfinx

abukharmeh: @jrtc27  I think I am done with all the changes you requested now, could you have a look at it, and let me know what you think ?

martinberger (commented):
LGTM. @abukharmeh  @jrtc27 @tariqkurd-repo @pmundkur  I'd like to merge this PR soon, Ideally today.

jrtc27:
No. It is not yet ready. And I do not appreciate the timeline being forced on me, that’s not appropriate. I will leave a proper review when I get the chance.

tariqkurd-repo:
@martinberger I'm happy, please merge it.

martinberger:
> No. It is not yet ready. And I do not appreciate the timeline being forced on me, that’s not appropriate. I will leave a proper review when I get the chance.

*"To achieve great things, two things are needed; a plan, and not quite enough time."* Attributed to many, including Leonard Bernstein.

When do you expect that Ibrahim can exhale?

tariqkurd-repo:
@jrtc27 we can't afford to spend more time on this. Due to our internal policies it is very time consuming making changes and submitting patches, which need to also go through an internal approval flow. Unless there are actual bugs in the code then we need to merge it now. If the suggestions are stylistic then they can be fixed after the merge. We have been trying to merge this for more than 1 year now, your feedback has been invaluable and has been the only way to make the code reusable, scalable and to reduce repetition (as the SAIL documentation is very limited you are our authority on how to write it) and I thank you for this. But now we need to move on and get it merged. @abukharmeh has a lot of other work to complete this year.

jrtc27:
> @jrtc27 we can't afford to spend more time on this. Due to our internal policies it is very time consuming making changes and submitting patches, which need to also go through an internal approval flow. Unless there are actual bugs in the code then we need to merge it now. If the suggestions are stylistic then they can be fixed after the merge. We have been trying to merge this for more than 1 year now, your feedback has been invaluable and has been the only way to make the code reusable, scalable and to reduce repetition (as the SAIL documentation is very limited you are our authority on how to write it) and I thank you for this. But now we need to move on and get it merged. @abukharmeh has a lot of other work to complete this year.

No. Internal policy issues are not an excuse to force others to approve sub-standard patches. That's your problem.

jrtc27:
And if you can't drive the patch to completion, then hand it off to someone else in the community to finish. I'm not merging bad code.

tariqkurd-repo:
If you think the code is bad after all this effort then I have no words to describe your attitude. @martinberger what you do next is up to you. We will not submit another patch if this is what we have to put up with.

jrtc27:
> If you think the code is bad after all this effort then I have no words to describe your attitude. @martinberger what you do next is up to you. We will not submit another patch if this is what we have to put up with.

I did not say bad, I said sub-standard. There are clear issues still with its style, the main one being the fact that it turns `F(rd) = ...` into an opaque function call without the = rather than using the register assignment style, so it's totally unclear that it's actually assigning to a register.

martinberger:
We have over a dozen specs being ratified this week ... we need to get going ...

jrtc27:
> We have over a dozen specs being ratified this week ... we need to get going ...

Most of which have no Sail PRs

jrtc27:
There is no rush for Zfinx over and above anything else. The style changes will be simple to fix, this can wait until that happens.

jrtc27:
I have reverted the merge, but will just take it upon myself to re-file a fixed PR since nobody else wants to

jrtc27:
(i.e. it will reappear soon, I just want to keep master clean, so don't get too upset with me...)

martinberger:
Thanks for your help.

We need to avoid the impression that the Sail model  is a bottleneck holding RISCV back, because otherwise some communities might not bother.

There is little Sail knowledge outside Cambridge, and not much documentation on Sail, and the expected way in which to write the RISCV model. The average computer architect has not come across liquid types before ... For this reason, it's unavoidable that the code that we get might be less than perfect. Raising a sustainable Sail/RISCV community will take a while.

martinberger:
> There is no rush for Zfinx over and above anything else. The style changes will be simple to fix, this can wait until that happens.

Have you got any recommendations for an order in which to merge the PRs?

martinberger:
> There is no rush for Zfinx over and above anything else. 

Have you got any recommendations for an order in which to merge the PRs?

jrtc27:
> Thanks for your help.
> 
> We need to avoid the impression that the Sail model is a bottleneck holding RISCV back, because otherwise some communities might not bother.
> 
> There is little Sail knowledge outside Cambridge, and not much documentation on Sail, and the expected way in which to write the RISCV model. The average computer architect has not come across liquid types before ... For this reason, it's unavoidable that the code that we get might be less than perfect. Raising a sustainable Sail/RISCV community will take a while.

I realise and understand that, which is why I don't mind taking the time to help people. What I take issue with is the "I don't have time to fix my code so it should just be merged" attitude etc; had it instead be "I don't have time to fix my code can someone please take it over the line for me" it would've been a different story (i.e. what I'm now doing to attempt to keep the peace).

jrtc27:
> > There is no rush for Zfinx over and above anything else. The style changes will be simple to fix, this can wait until that happens.
> 
> Have you got any recommendations for an order in which to merge the PRs?

I don't think there's much of an interdependence, so just whenever things become ready. I suspect the Zb* PRs will soon be merge candidates, those were looking close to ready last I checked, though I haven't closely been following the recent changes. I've just been a bit more swamped with work and life things the past week or so compared with normal so not reviewing as promptly as I'd like.

scottj97:
I don't know who any of you are but thank you @jrtc27 for your efforts at keeping the model clean. I concur that more documentation and examples could help people write Sail code but that's no excuse for merging substandard code. Of the year this PR has been open, 8 months of that was waiting for @abukharmeh to respond to feedback (Feb to Oct), so the long time frame is no excuse for merging prematurely.

The original PR was quite nasty and would have made a mess of this repo. We must have people like @jrtc27 to uphold quality standards or this project will not survive long-term.

model/riscv_insts_fext.sail
```diff
@@ -673,22 +659,22 @@ mapping clause encdec =
 /* Execution semantics ================================ */
 
 function clause execute (F_UN_RM_TYPE_S(rs1, rm, rd, FSQRT_S)) = {
-  let rs1_val_S = nan_unbox (F(rs1));
+  let rs1_val_S = X_or_F_S(sys_enable_zfinx(), rs1);
```
scottj97, line 662: Can't `X_or_F_S()` call `sys_enable_zfinx()` itself? Unfortunate that we have to keep repeating this all over the place.

jrtc27: Yeah that's one of the things I'll be changing

model/riscv_insts_fext.sail
```diff
@@ -350,6 +325,17 @@ function is_RV64F          () -> bool = (haveFExt() & sizeof(xlen) == 64)
 function is_RV32D_or_RV64D () -> bool = (haveDExt() & (sizeof(xlen) == 32 | sizeof(xlen) == 64))
 function is_RV64D          () -> bool = (haveDExt() & sizeof(xlen) == 64)
 
+val supports_RV32D_or_RV64D : forall 'n, 'n > 0. (implicit('n), vector('n, dec, bits(5))) -> bool effect {rreg}
+function supports_RV32D_or_RV64D (reg_vec_len, registers) = {
+  if ~ (sys_enable_zfinx()) then is_RV32D_or_RV64D()
+  else if sys_enable_zfinx() & is_RV64D() then true
```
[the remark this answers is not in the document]
jrtc27: Yep, hadn't even seen that code yet. The register checking should also be a different function, it's not related to whether RV32D or RV64D are supported.

model/riscv_fdext_regs.sail
```diff
@@ -363,6 +456,6 @@ function accrue_fflags(flags) = {
   then {
     fcsr->FFLAGS() = f;
     update_softfloat_fflags(f);
-    dirty_fd_context();
+    if ~ (sys_enable_zfinx()) then dirty_fd_context();
```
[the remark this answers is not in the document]
jrtc27: I think both are ok, I could go either way on it, though I think I prefer this, since there is no FD context on Zfinx and there should probably be an assert in dirty_fd_context that it's not Zfinx (haven't checked if this patch adds it or not).

jrtc27:
One bug I've already found going through the diff and cleaning it up is that this NaN boxes on Zfinx, i.e. always puts 32 ones in the upper half of X registers when writing a single-precision float on RV64, whereas for Zfinx it's different to F and is supposed to sign-extend. It is at least doing the right thing for the other half of that, namely ignoring the upper 32 bits rather than ensuring they are all ones like for RV32D.

model/riscv_insts_fext.sail
jrtc27: (I suspect that's why the register assignment syntax was abandoned, as I could believe Sail not supporting that when you have more than one argument since that doesn't really make sense normally)

abukharmeh: Hi, was waiting for your feedback, then this went kind of downhill :( Thank you for your valuable help.

model/riscv_fdext_regs.sail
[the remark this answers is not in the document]
scottj97: Rename the function to `dirty_fd_context_if_exists()` then? It's not great to keep repeating the conditional like this. Whether those bits exist or not should be defined in exactly one place.

Alasdair:
I wonder if there are any lint-style warnings we could add to Sail itself that might help catch common style issues and pitfalls? I'd be interested in suggestions both from those writing PRs and reviewing them - I'm not involved in either so it can hard for me to have a good perspective on this. If some amount of effort from me could possible save time getting PRs into sail-riscv it might be worthwhile?

martinberger:
@Alasdair There are two distinct issues that it might be worthwhile to think about separately.

- Sail format (e.g. how to use brackets, indentation)
- RISCVism (e.g. how do we access registers, throwing an exception vs using an option type, code-repetition, use of type-refinements)

The former would be solved by a code formatter. The latter is more substantial, but probably not so easy to write linters for.

abukharmeh:
> @Alasdair There are two distinct issues that it might be worthwhile to think about separately.
> 
> * Sail format (e.g. how to use brackets, indentation)
> * RISCVism (e.g. how do we access registers, throwing an exception vs using an option type, code-repetition, use of type-refinements)
> 
> The former would be solved by a code formatter. The latter is more substantial, but probably not so easy to write linters for.

+1 for code formatter, or even some format document, that wasted review time on every single single recent PR in this repo.

martinberger:
I had several undergraduate interns from Cambridge and Imperial this summer, who did quite a bit of Sail. They have gone pack to university, but they might be interested in doing paid Sail work. They'd be ideal for this, but they would probably want to be paid (or at least get some recognition) in some form of shape. 

If there was interest in this, I'll contact them!

@Alasdair @jrtc27 @tariqkurd-repo @PeterSewell @allenjbaum @cetola @jjscheel

abukharmeh:
@jrtc27  Re Nanboxing, the original spec was to always nanbox but not check for it in zfinx, it appears that it got changed to sign extend, and I didn't notice it. If I fix that along and remove the redundant check in supports_RV32_or_RV64D function. Would you be happy to merge this PR  ?  I am not too keen on changing X_or_F functions to remove the status flag from the argument as that would make it less obvious what is happening !

bilalsakhawat-10xe:
Hi!
I started learning sail roughly two months ago. I help myself from the available video lecture by @rsnikhil, related [documentation](https://github.com/rsnikhil/RISCV_ISA_Spec_Tour) and the sail language [manual](https://github.com/rems-project/sail/blob/sail2/manual.pdf). Actually @allenjbaum  pointed me there. From what I have learnt so far, I did some sail support work i.e., the pull requests for Zb* , zmmul and zfh extensions support are under review. I am improving thanks to @jrtc27 (from her reviews). I am interested in helping with above work, documentation and Sail format etc.

jrtc27:
Oh I had already started trying to clean things up a bit more on Wednesday between meetings, finishing that off is something I had planned for the weekend, so assuming you/Tariq/Martin/Scott don't spot anything else that should be changed I imagine we should be able to get it merged next week.

Re removing the flag from X_or_F, reg_or_freg already implicitly checks whether Zfinx is enabled, so I think it's better to be consistent, as well as avoiding all the repetition.

scottj97:
> I am not too keen on changing X_or_F functions to remove the status flag from the arguments as that would make it less obvious what is happening !

The term "X_or_F" makes it abundantly clear what's happening. If someone cares to know how the choice is made between X or F they can look in that function.

Especially if passing `sys_enable_zfinx()` means that register assignment syntax is no longer supported, that's a terrible tradeoff.

abukharmeh:
@Alasdair +1 for some kind of code formatter, it would help everyone trying to submit PRs to the model, and would reduce the amount of reviewing if the style is consistent across all PRs. 

Suggestion for lint warning, it appears that the convention here to enforce spaces around commas, though I am not sure that's it make sense for the compiler to warn about such things.  Perhaps we can have couple of such rules enabled by strict lint argument or something similar.

PR #77, opened 2020-11-18
Bug fix: when writing CSR sip, ignore attempt to change undelegated SSIP

scottj97:
The `sip` register is special in that is a restricted view of `mip`. Only the bits delegated to S-mode by `mideleg` are visible in `sip`.

Writing to `sip` when `ssip` is not delegated should not change `mip.ssip`.

The Sail model was mistakenly writing 0 to `mip.ssip`, instead of leaving the previous value untouched.


pmundkur:
Good catch, thanks!

PR #78, opened 2020-11-30
Ignore undefined bits in mstatus on write

arichardson:


scottj97:
Can you explain what problem this solves?

arichardson:
Most of these fields are WARL/WPRI 3.1.6 Machine Status Registers (mstatus and mstatush), so reads of undefined fields should return zeroes. Right now we allow writing any value and also reading that again.

pmundkur:
Good catch, thanks.

PR #79, opened 2020-12-03
Add preliminary support for Zsn extension

daniellustig:
Zsn is a new supervisor mode extension being developed by the virtual memory task group.  The goal of Zsn is to support contiguous address translation ranges of naturally-aligned power-of-two (NAPOT) sizes between the existing page sizes (4KiB, 2MiB).  The specific region size we're aiming to support for now is 64KiB, but we may extend support to other power-of-two sizes in the future.  Zsn is currently in draft state, with a plan to go to public review hopefully by the end of the year.

See https://github.com/riscv/virtual-memory for more info on Zsn.  See https://github.com/riscv/virtual-memory/blob/main/presentations/2020-12-03-Zsn-Sail.adoc for some commentary on these Sail changes.

The Zsn part itself has not yet been tested.  The virtual memory group is working on creating a suite of tests, but we don't have them ready yet.  There's no urgency to merge this before we're sure it's ready, but I figured I'd still open a PR to allow for feedback on the code in the meantime.  This is my first attempt to write Sail, so feedback is welcome.

### DIFF
--- a/model/riscv_pte.sail
+++ b/model/riscv_pte.sail
@@ -29,6 +29,11 @@ bitfield PTE_Bits : pteAttribs = {
   V : 0
 }
 
+bitfield PTE_Ext : extPte = {
+  C : 9,
+  N : 8
+}
+
 function isPTEPtr(p : pteAttribs, ext : extPte) -> bool = {
   let a = Mk_PTE_Bits(p);
   a.R() == 0b0 & a.W() == 0b0 & a.X() == 0b0
--- a/model/riscv_vmem_sv39.sail
+++ b/model/riscv_vmem_sv39.sail
@@ -1,5 +1,18 @@
 /* Sv39 address translation for RV64. */
 
+val napot_bits39 : (bits(44), nat) -> nat
+function napot_bits39 (v, n) = {
+  let mask = shiftl(v ^ v ^ EXTZ(0b1), n + 1) - 1;
+  let pattern = shiftl(v ^ v ^ EXTZ(0b1), n);
+  if n > 8 then {
+    0
+  } else if (v & mask) == pattern then {
+    n + 1
+  } else {
+    napot_bits39(v, n + 1)
+  }
+}
+
 val walk39 : (vaddr39, AccessType(ext_access_type), Privilege, bool, bool, paddr64, nat, bool, ext_ptw) -> PTW_Result(paddr64, SV39_PTE) effect {rmem, rmemt, rreg, escape}
 function walk39(vaddr, ac, priv, mxr, do_sum, ptb, level, global, ext_ptw) = {
   let va = Mk_SV39_Vaddr(vaddr);
@@ -56,16 +69,30 @@ function walk39(vaddr, ac, priv, mxr, do_sum, ptb, level, global, ext_ptw) = {
                 } else {
                   /* add the appropriate bits of the VPN to the superpage PPN */
                   let ppn = pte.PPNi() | (EXTZ(va.VPNi()) & mask);
+                  let shift : nat = PAGESIZE_BITS + (level * SV39_LEVEL_BITS);
 /*                let res = append(ppn, va.PgOfs());
                   print("walk39: using superpage: pte.ppn=" ^ BitStr(pte.PPNi())
                         ^ " ppn=" ^ BitStr(ppn) ^ " res=" ^ BitStr(res)); */
-                  PTW_Success(append(ppn, va.PgOfs()), pte, pte_addr, level, is_global, ext_ptw)
+                  PTW_Success(append(ppn, va.PgOfs()), pte, pte_addr, level, is_global, ext_ptw, shift)
+                }
+              } else if Mk_PTE_Ext(ext_pte).N() == 0b1 then {
+                let shift = napot_bits39(pte.PPNi(), 0);
+                if shift == 4 then {
+                  let Zsn_ppn_part = (pte.PPNi() >> shift) << shift;
+                  let Zsn_mask_vpn_bits = shiftl(va.VPNi() ^ va.VPNi() ^ EXTZ(0b1), shift) - 1;
+                  let napot_vpn_part = Zsn_mask_vpn_bits & va.VPNi();
+                  let napot_ppn = Zsn_ppn_part | EXTZ(napot_vpn_part);
+/*                let res = append(napot_ppn, va.PgOfs());
+                  print("walk39: NAPOT: pte.ppn=" ^ BitStr(pte.PPNi()) ^ " ppn=" ^ BitStr(pte.PPNi()) ^ " res=" ^ BitStr(res)); */
+                  PTW_Success(append(pte.PPNi(), va.PgOfs()), pte, pte_addr, level, is_global, ext_ptw, PAGESIZE_BITS + shift)
+                } else {
+                  PTW_Failure(PTW_Invalid_PTE(), ext_ptw)
                 }
               } else {
                 /* normal leaf PTE */
 /*              let res = append(pte.PPNi(), va.PgOfs());
                 print("walk39: pte.ppn=" ^ BitStr(pte.PPNi()) ^ " ppn=" ^ BitStr(pte.PPNi()) ^ " res=" ^ BitStr(res)); */
-                PTW_Success(append(pte.PPNi(), va.PgOfs()), pte, pte_addr, level, is_global, ext_ptw)
+                PTW_Success(append(pte.PPNi(), va.PgOfs()), pte, pte_addr, level, is_global, ext_ptw, PAGESIZE_BITS)
               }
             }
           }
@@ -89,9 +116,9 @@ function lookup_TLB39(asid, vaddr) =
     Some(e) => if match_TLB_Entry(e, asid, vaddr) then Some((0, e)) else None()
   }
 
-val add_to_TLB39 : (asid64, vaddr39, paddr64, SV39_PTE, paddr64, nat, bool) -> unit effect {wreg, rreg}
-function add_to_TLB39(asid, vAddr, pAddr, pte, pteAddr, level, global) = {
-  let ent : TLB39_Entry = make_TLB_Entry(asid, global, vAddr, pAddr, pte.bits(), level, pteAddr, SV39_LEVEL_BITS);
+val add_to_TLB39 : (asid64, vaddr39, paddr64, SV39_PTE, paddr64, nat, bool, nat) -> unit effect {wreg, rreg}
+function add_to_TLB39(asid, vAddr, pAddr, pte, pteAddr, level, global, shift) = {
+  let ent : TLB39_Entry = make_TLB_Entry(asid, global, vAddr, pAddr, pte.bits(), level, pteAddr, shift);
   tlb39 = Some(ent)
 }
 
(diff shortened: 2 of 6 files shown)


model/riscv_pte.sail
```diff
@@ -29,6 +29,11 @@ bitfield PTE_Bits : pteAttribs = {
   V : 0
 }
 
+bitfield PTE_Ext : extPte = {
```
jrtc27, line 32: If you're making this defined by the standard then this should be a separate type (logically, part of pteAttribs, but it may be easier to have pteAttribsHi and pteAttribsLo), not extPte. extPte is for non-standard extensions to the model to use (i.e. just sail-cheri-riscv, currently), but this also needs to be done in a way that those extensions continue to be able to work (code changes are fine and may well be necessary, but there must be a way for them to hook into sail-riscv as needed without modifying sail-riscv).

model/riscv_vmem_sv39.sail
```diff
@@ -1,5 +1,18 @@
...
+val napot_bits39 : (bits(44), nat) -> nat
+function napot_bits39 (v, n) = {
+  let mask = shiftl(v ^ v ^ EXTZ(0b1), n + 1) - 1;
+  let pattern = shiftl(v ^ v ^ EXTZ(0b1), n);
+  if n > 8 then {
+    0
+  } else if (v & mask) == pattern then {
+    n + 1
+  } else {
+    napot_bits39(v, n + 1)
```
jrtc27, line 12: The recursion seems unnecessarily complex. What are you actually trying to compute? Would a loop suit you better or can you do away with even that?

model/riscv_vmem_sv48.sail
```diff
@@ -1,5 +1,18 @@
...
+val napot_bits48 : (bits(44), nat) -> nat
+function napot_bits48 (v, n) = {
+  let mask = shiftl(v ^ v ^ EXTZ(0b1), n + 1) - 1;
+  let pattern = shiftl(v ^ v ^ EXTZ(0b1), n);
+  if n > 8 then {
+    0
+  } else if (v & mask) == pattern then {
+    n + 1
+  } else {
+    napot_bits39(v, n + 1)
```
jrtc27, line 12: Wrong function? And how is the implementation of this meant to differ from the Sv48 version? Can they share a common implementation?

model/riscv_pte.sail
daniellustig: Thanks for the feedback @jrtc27.  Is renaming to pteAttribsHi and pteAttribsLo enough to avoid that conflict, or are you suggesting something else also needs to be done?

model/riscv_vmem_sv39.sail
daniellustig: Please forgive my Sail naivety.  This is really just "find last zero" aka "count trailing zeros + 1".  If there's a simpler way to write that, great.

model/riscv_vmem_sv48.sail
daniellustig: Oops, yes, this should be napot_bits48.  Good catch.

In general almost all of the code in Sv32, Sv39, and Sv48 is redundant already.  I agree it would make sense to factor out not just this new napot_bits function but also all of the rest of that common code, but I didn't want to tie that into this Zsn PR or take all of that on at one in my first attempt to write Sail, so I kept it separate for now.

model/riscv_vmem_sv39.sail
jrtc27: There's a `count_leading_zeros` in the standard library (in fact it's a builtin) but seemingly no trailing ones. There's also the following commented-out version (for if it weren't a builtin):
```
function count_leading_zeros x = {
  foreach (i from ('N - 1) to 0 by 1 in dec)
    if [x[i]] == [bitone] then return 'N - i - 1;
  return 'N;
}
```
so I imagine something like the following would work:
```
val count_trailing_zeros = "count_trailing_zeros" : forall 'N , 'N >= 1. bits('N) -> {'n, 0 <= 'n <= 'N . atom('n)}
function count_trailing_zeros x = {
  foreach (i from 0 to ('N - 1) by 1)
    if [x[i]] == [bitone] then return i;
  return 'N;
}
```
Not sure what the square brackets around `x[i]` and `bitone` are for, mind, I'd expect it to work without.

model/riscv_vmem_sv48.sail
jrtc27: Yeah, it's just slightly more awkward with the extra level of translation (but easy enough to do), and Sv32 is rather different due to not having the upper bits in the PTEs. Merging Sv39 and Sv48 together is easy enough to do, but for Sv32 it'd probably be a case of pulling out common functionality but still having separate files, one for Sv32 and one for Sv39+48(+57).

model/riscv_pte.sail
jrtc27: Yes and no. It means `extPte` is free to be defined by extensions again, but there's still the issue of changing what you can do with it; either it overlaps with some/all of pteAttribsHi and things get really weird, or you reduce it to the bits not defined by the standard and then extensions using them break. Either way there are issues for CHERI-RISC-V; the former means composing Zsn and Xcheri doesn't work (but if done correctly you should still be able to support them as mutually exclusive extensions in the model), the latter just breaks it entirely and forces us to change CheriBSD, our four/five different hardware implementations, QEMU, spike and, of course, sail-cheri-riscv, to pick different bits for the 5 bits (and researchers at Microsoft are reserving a 6th bit) we need (and then rinse and repeat as RISC-V gobbles more up in the coming years), which is manageable for a research group but extremely painful when we have downstreams at other universities and companies across the world building things on CHERI-RISC-V. There was no reserved space for non-standard extensions so we picked what we thought was most likely to remain unused (yeah, picking reserved bits not marked for that is a risk, but we had no other option), but here we are. So I don't know what the right answer is beyond being able to convince the virtual memory task group to leave those bits reserved for non-standard extensions.

jrtc27: Also, this was done based on the fact that riscv-privileged says "These reserved bits may also be used to facilitate research experimentation.", which is exactly what we're doing.

model/riscv_vmem_sv48.sail
daniellustig: I'm willing to give this refactoring a shot (on a separate standalone commit).  We're also planning to add Sv57 soon, so it would make sense to refactor this now before duplicating the code yet again.

I do have one question though @jrtc27.  Within the walkN functions, while the code is mostly the same, there are function calls like Mk_SV48_PTE(v) which are specific to the PTE type, and the Sv32, Sv39, Sv48, and Sv57 (assuming I follow the pattern) PTEs are defined as different types in riscv_vmem_common.sail, even though all but Sv32 are actually the same.  If I make the function generic, I don't see an easy way to call the right Mk_SVN_PTE function for the given PTE type.  I can't just factor it out from the top either because the function is called recursively.

Is there an easy way in Sail to write a function like walk in a generic way while still being able to specialize for the right Mk_SVN_PTE(v) function internally?  Similar question elsewhere, e.g., for translate48 and add_to_TLB48.  I couldn't figure out how to pass a function pointer as an argument, if that's even possible in Sail.  Are there other ways around this?  If not, perhaps that's why the code was just duplicated in the first place?

Alternatively, would it be a bad thing to at least simply merge the Sv39, Sv48, and Sv57 PTE types into common "64b PTE" types, so at least we'd get this down to 2x duplication (32b vs. 64b) instead of 4x or more?

model/riscv_vmem_sv39.sail
allenjbaum: This is probably very naive, but can't you just perform count_trailing_zeroes(~x)

model/riscv_vmem_sv48.sail
martinberger: Hello Daniel, what's the state of the refactoring you mentioned?

daniellustig: Hi @martinberger, I haven't had a chance to look at this since my last comment, in part because I didn't (and still don't) have enough Sail expertise to figure out how to address the questions I asked in my last comment.  It still seems like it would be a good idea to refactor the code to remove redundancy between Sv39, Sv48, and (imminently) Sv57 first, but I'm not really sure how to do it.

Also, I should point out that "Zsn" is now called "Svnapot"...the functionality hasn't changed; only the name has changed.

martinberger: @daniellustig How imminent is Sv57?

Maybe it makes sense to  close the PR and resubmit

- if/when Sv57 is added, and
- either the redundancies have been removed, or removing them is not possible?

Note that (for good reasons) Sail has only limited abstraction facilities, so the redundancies might be needed.

daniellustig: Closing the PR is fine with me if that's what you prefer.  Is someone going to look into either removing the redundancies or declaring that we have to live with the redundancy?

FWIW Sv57 is currently out for public review, so it will very likely be ratified soon, even if it's not yet in the Sail model.  https://groups.google.com/a/groups.riscv.org/g/isa-dev/c/1bTAL-PJGYQ

jjscheel: Actually, all Priv instructions like Sv57 have a standing waiver for this year for SAIL work.  But, the advice still stands: do what's right for the project.

martinberger: @daniellustig @jjscheel I don't think it's much work, maybe 100 lines of Sail, so it would be easy to do for somebody who was well-versed in VM and Sail ... Anyway, it sounds like we should not wait until Sv57 is ready and keep it separate.

martinberger: @daniellustig Did you have further thoughts on this? I'm happy to support  Sv57 getting in as soon as possible.

martinberger: @daniellustig I have no strong preference. It would be good to speed up merging important PRs in.

martinberger: @daniellustig I've had some undergraduate summer interns who did a lot of Sail. They are back at university, I can ask them if they wanted to do (paid) Sail work, if you like.

allenjbaum:
adding Mark who may have duget for this

model/riscv_vmem_sv48.sail
daniellustig: Hi @martinberger, sorry for the delayed response (it was the Thanksgiving holiday here in the US).  If they are interested and funding is available, then it would be great to have their help.

martinberger: @daniellustig I don't know about funding, I doubt that the RISCV organisation pays this. But he is interested, I'll connect you in a PM.

jrtc27: Mark Himelstein is the person to talk to about that. I know there is money floating around for some things like this, but don't recall exactly what.

martinberger:
> may have budget for this

@daniellustig @cetola @allenjbaum  Did anything come out of this?

cetola:
@martinberger @allenjbaum We’re looking into internship possibilities, and also have a job [req](https://www.linuxfoundation.org/about/careers/) out for a full time Sail developer.

daniellustig:
Closing this PR for now.  If we can find someone who can properly implement Svnapot in the Sail model, they can either reopen this PR and continue working on it, or they can redo it from scratch, as they prefer.

martinberger:
Thanks @daniellustig I hope we'll the see new version soon.

PR #81, opened 2021-01-27
Prevent CSRW to sstatus from modifying uie & upie if N-ext not present

scottj97:
A CSRW to the `sstatus` shadow register was erroneously allowed to write to the `uie` and `upie` bits of `mstatus`, even when those bits are supposed to be hardwired to 0 when the N-extension (user-mode interrupts) is not present.

By calling `legalize_mstatus()` I avoid duplicating this logic in `legalize_sstatus()`.


jrtc27 (commented):
Looks correct to me

pmundkur:
thanks!

PR #82, opened 2021-01-27
MRET and SRET should increment minstret

scottj97:
The ISA spec says that EBREAK and ECALL do not count as committed instructions (and therefore shouldn't increment `minstret`), but it doesn't say anything like that about MRET or SRET. Spike increments `minstret` for those two instructions.

For some reason the Sail model had MRET and SRET returning RETIRE_FAIL instead of RETIRE_SUCCESS. If I am reading the code correctly, this return code is only used to affect the increment of `minstret`, and therefore this change should have no other effects.

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -717,11 +717,10 @@ mapping clause encdec = MRET()
 
 function clause execute MRET() = {
   if   cur_privilege != Machine
-  then handle_illegal()
+  then { handle_illegal(); RETIRE_FAIL }
   else if ~(ext_check_xret_priv (Machine))
-  then ext_fail_xret_priv ()
-  else set_next_pc(exception_handler(cur_privilege, CTL_MRET(), PC));
-  RETIRE_FAIL
+  then { ext_fail_xret_priv(); RETIRE_FAIL }
+  else { set_next_pc(exception_handler(cur_privilege, CTL_MRET(), PC)); RETIRE_SUCCESS }
 }
 
 mapping clause assembly = MRET() <-> "mret"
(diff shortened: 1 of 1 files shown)


scottj97:
This has a problem; if the SRET takes an illegal opcode exception (e.g. because `mstatus.tsr=1`), it's still incrementing `minstret`.

scottj97:
Updated and squashed my fix so that it only increments `minstret` when the mret/sret was successful.

model/riscv_insts_base.sail
```diff
@@ -717,11 +717,10 @@ mapping clause encdec = MRET()
...
 function clause execute MRET() = {
   if   cur_privilege != Machine
-  then handle_illegal()
+  then { handle_illegal(); RETIRE_FAIL }
   else if ~(ext_check_xret_priv (Machine))
-  then ext_fail_xret_priv ()
-  else set_next_pc(exception_handler(cur_privilege, CTL_MRET(), PC));
-  RETIRE_FAIL
+  then { ext_fail_xret_priv(); RETIRE_FAIL }
+  else { set_next_pc(exception_handler(cur_privilege, CTL_MRET(), PC)); RETIRE_SUCCESS }
```
jrtc27, line 723: Hm, might be better to not make this one a one-liner.

jrtc27 (commented):
This looks correct. I assume MRET/SRET were just written at the same time as ECALL and accidentally ended up inheriting the behaviour of trapping instructions rather than committing ones.

model/riscv_insts_base.sail
scottj97: Done

pmundkur:
great catch, thx!

PR #83, opened 2021-02-12
Add HPM event counters with support for platform-specific customization

pmundkur:
This adds the HPM registers to the Sail model, but does not by default use them to count any events.

Events and the codes for their selection are inherently platform-specific, so minimal support for them is put in the C emulator.  An example of how to use the provided hooks for events is in `c_emulator/platform/{example_event_def.enums,riscv_events_example.c}`.
A new platform can be added by specifying custom versions of these files, and using the `HPM_PLATFORM` variable to pull these files in.

To simplify supporting this platform-specific customization, two registers were introduced in the Sail model that do not correspond to architectural registers.  These capture the bits of the instruction, and whether the instruction retired.  These added registers may be controversial and need discussion.


model/riscv_sys_control.sail
```diff
@@ -106,15 +254,136 @@ function check_Counteren(csr : csreg, p : Privilege) -> bool =
     (0xC00, Supervisor) => mcounteren.CY() == 0b1,
     (0xC01, Supervisor) => mcounteren.TM() == 0b1,
     (0xC02, Supervisor) => mcounteren.IR() == 0b1,
+    (0xC03, Supervisor) => mcounteren.HPM()[0] == bitone,
```
scottj97, line 257: Is there any difference between `0b1` and `bitone`? If not it's probably best not to mix the two styles here.

pmundkur:
> Is there any difference between `0b1` and `bitone`? If not it's probably best not to mix the two styles here.

Yes: 0b1 is a one-element bitvector, bitone is a scalar.  The register fields are bitvectors.

jrtc27:
> Events and the codes for their selection are inherently platform-specific

Only currently, but that is a specification deficiency that Arm learned the hard way. I really dislike how much of this is C for various reasons: the OCaml platform lacks them, theorem provers can't reason about them, and it's just much nicer having things in Sail.

martinberger:
@pmundkur What is the status of this PR, it's been quite for a while.

ptomsich:
GIven that we have had no activity since Oct 4, 2021 (i.e., for more than a year and a half!) on this PR, I am assuming it is abandoned and closing.

If anyone doesn't agree, please re-open w/ an updated patch set.

billmcspadden-riscv:
Re-opening pull request.  I'm working on this.  Attempting to convert the C implementation to Sail (at some level);  all the important bits at least (like incrementing the counter).

I'm working on this in a separate branch on a more recent baseline.

jrtc27:
> Re-opening pull request. I'm working on this. Attempting to convert the C implementation to Sail (at some level); all the important bits at least (like incrementing the counter).
> 
> I'm working on this in a separate branch on a more recent baseline.

Different branch means new PR.

billmcspadden-riscv:
@jrtc27 :  Right.  I re-opened,  thought about it,  and then realized this was wrong.  Closing.

Issue #72, opened 2020-10-29
Range check problem


soberl:
https://github.com/rems-project/sail-riscv/blob/ae6cb1de092e9ea727e2318d76e9b88999bbee59/model/riscv_pmp_control.sail#L60

I think it should be "if      (addr + width <_u lo) | (hi <=_u addr)" instead.
Suppose there are 2 ranges: [0x0, 0x200000) and [0x200000, 0x300000). Giving an address 0x200000 with any size, it should be belong to the 2nd range.

allenjbaum:
This appears to be correct - 
ranges for PMP NAPOT format define regions at base..base+regionsize-1
ranges for PMP range    format define regions at base_lo..base_hi-1
This is a bug in the RISC-V Sail model

scottj97:
It's broken on the low end, too. With your example ranges, consider an access to addr=0x1ffffc width=4. This is outside of the second range, yet `(addr + width <_u lo)` would incorrectly say that it's within that entry.

Correct:
```
                        if      (addr + width <=_u lo) | (hi <=_u addr)
```

soberl:
Yes.
For easier understanding, maybe (assert width > 0):
`if      (addr >= _u hi) | (addr + width <=_u lo) `

scottj97:
I wonder if there's similar corner cases here when accessing the largest-addressed bytes in memory. addr=0xFFFFFFFC width=4, so addr+width wraps around.

Since addr is of type `xlenbits`, this would apply only to RV32, since RV64 physical addresses can't be so large as to wrap. (Right?)

soberl:
See the fix as commented.

scottj97:
@soberl what fix? What comment? #85 is still open?

soberl:
Ah, yes, it's still open. Let me reopen this issue.

> @soberl what fix? What comment? #85 is still open?

rmn30:
I believe this is fixed by https://github.com/rems-project/sail-riscv/pull/85 but someone should check.

scottj97:
I can confirm it is fixed.

PR #85, opened 2021-03-19
PMP address match misses some corner cases

ybc-alkaid:
The PMP address match logic misses some corner cases.
Since an entry's legal address is [lo, hi), if one tries to access hi, the address matching result of this entry should be PMP_NoMatch rather than PMP_PartialMatch.
Another corner case is that one tries to access lo, assume width > 0, then address + width = lo should be PMP_NoMatch rather than PMP_PartialMatch (that address should locate in the previous entry).


scottj97 (approved):

scottj97:
This is a fix for #72

rmn30:
Fix looks good to me. @scottj97 is also correct that there is a problem if `addr + width` overflows. Perhaps that case should be caught in `pmpCheck` and result in an immediate `PMP_Fail`?

pmundkur (approved):
looks good to me

PR #86, opened 2021-03-22
Ignore build library files

PeterRugg:
This prevents git from trying to version the build files in c_emulator/SoftFloat-3e.


rmn30:
Seems reasonable.

PR #88, opened 2021-05-04
macOS build fixes

jrtc27:


rmn30:
We should merge this but please could you add the workaround for ubuntu 18.04 as per https://github.com/CTSRD-CHERI/sail-cheri-riscv/commit/e43f70962492b3cbdbde74a1acb9493a9064e1e7

jrtc27:
@rmn30 could you please commit the `# N.B. GMP does not have pkg-config metadata on Ubuntu 18.04 so default to -lgmp` comment?

PR #89, opened 2021-05-21
Fix internal error caused by invalid LR / SC / AMO widths.

rmn30:
The RISC-V spec. only shows LR / SC / AMOs for word and double widths
even though the encoding would naturally extend to other widths (it is
a little unclear). Previously the Sail model would decode the unused
widths but throw an internal error in execute. With this change the
unused widths will not be decoded and will cause illegal instruction
exceptions instead.

### DIFF
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -20,11 +20,24 @@ function lrsc_width_str(width : word_width) -> string =
     DOUBLE => ".d"
   }
 
+/**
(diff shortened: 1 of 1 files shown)


pmundkur (approved):
lgtm

rmn30:
I have just pushed an additional commit with a fix for lr / sc with unusual aq / rl flags. Comments welcome.

pmundkur (approved):
makes sense

PR #90, opened 2021-06-18
Add an extension point to allow validation of physical memory accesses.

rmn30:
This is required for implementing memory version support for CHERI.

### DIFF
--- a/model/riscv_addr_checks_common.sail
+++ b/model/riscv_addr_checks_common.sail
@@ -20,3 +20,24 @@ union Ext_DataAddr_Check ('a : Type) = {
   Ext_DataAddr_OK : xlenbits,    /* Address to use for the data access */
   Ext_DataAddr_Error : 'a
 }
+
+union Ext_PhysAddr_Check = {
+  Ext_PhysAddr_OK : unit,
+  Ext_PhysAddr_Error : ExceptionType
+}
+
+/*!
+ * Validate a read from physical memory.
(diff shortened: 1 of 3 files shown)


model/riscv_addr_checks_common.sail
```diff
@@ -20,3 +20,24 @@ union Ext_DataAddr_Check ('a : Type) = {
   Ext_DataAddr_OK : xlenbits,    /* Address to use for the data access */
   Ext_DataAddr_Error : 'a
 }
+
+union Ext_PhysAddr_Check = {
```
pmundkur, line 24: Should this be parametrized over the ExceptionType for consistency with the other address check returns?

rmn30: Yes, it probably should and Ext_physAddr_OK should return a replacement address just in case.

rmn30: I had  a go at parameterising `ExtPhysAddr_Check` here:
https://github.com/rems-project/sail-riscv/compare/ext_check_phys_mem_alt?expand=1

To be honest though, I'm not sure there is much much point: in the places where `ext_check_phys_mem_{read,write}` are called the only way to handle a failure is to return a `MemException` unless we change the interface of all the functions in `riscv_mem.sail` and I don't see any need for that. Any extension specific handling could be performed in the check function so there is not much need for the separate `ext_handle_phys_check_error` either.

PR #91, opened 2021-06-24
Allow extensions to set memory privilege level

nwf:
- Move `effectivePrivilege` out to a file that extensions can shadow

- Rename it to `effectiveMemoryPrivilege` for increased clarity

- Make it pure by using its arguments consistently; all information was already
  being passed in as arguments, but, probably as a result of incomplete
  refactorings of yore, global registers were still being used as well.

- Update the call sites and Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ endif
 # Non-instruction sources
 PRELUDE = prelude.sail prelude_mapping.sail $(SAIL_XLEN) $(SAIL_FLEN) prelude_mem_metadata.sail prelude_mem.sail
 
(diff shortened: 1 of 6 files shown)


rmn30:
Looks fine. Would it be worth documenting as a potential extension point? Others often use the prefix ext_ .

jrtc27:
Why not add a new hook instead that gets called with the inputs and default result? That way extensions don't need to copy the existing base logic if they want to reuse it.

nwf:
This looks like it's going to be unnecessary with #92 .

PR #92, opened 2021-06-24
PTW accesses memory with incorrect permission checks 

nwf:
* Presently, the `walk` functions copy the instruction's memory access type when issuing `mem_read`s for PTEs.  That's readily fixed by the attached commit.

* Additionally, @jrtc27 points out that `mem_read` and `mem_write_value` do not allow the PTW to override the influence of `mstatus` and `cur_privilege` on the `Privilege` used to access the PTEs, while the spec says

      PMP checks are also applied to page-table accesses for virtual-address translation, for which the effective privilege mode is S

  (in 3.7 of the latest prerelease of 1.12 as of the time of this writing).  That is going to be a more invasive change to `model/riscv_mem.sail`.

### DIFF
--- a/model/riscv_vmem_sv32.sail
+++ b/model/riscv_vmem_sv32.sail
@@ -12,7 +12,7 @@ function walk32(vaddr, ac, priv, mxr, do_sum, ptb, level, global, ext_ptw) = {
   let pt_ofs : paddr32 = shiftl(EXTZ(shiftr(va.VPNi(), (level * SV32_LEVEL_BITS))[(SV32_LEVEL_BITS - 1) .. 0]),
                                 PTE32_LOG_SIZE);
   let pte_addr = ptb + pt_ofs;
-  match (mem_read(ac, to_phys_addr(pte_addr), 4, false, false, false)) {
+  match (mem_read(Read(Data), to_phys_addr(pte_addr), 4, false, false, false)) {
     MemException(_) => {
 /*    print("walk32(vaddr=" ^ BitStr(vaddr) ^ " level=" ^ string_of_int(level)
             ^ " pt_base=" ^ BitStr(to_phys_addr(ptb))
--- a/model/riscv_vmem_sv32.sail
+++ b/model/riscv_vmem_sv32.sail
@@ -12,7 +12,7 @@ function walk32(vaddr, ac, priv, mxr, do_sum, ptb, level, global, ext_ptw) = {
   let pt_ofs : paddr32 = shiftl(EXTZ(shiftr(va.VPNi(), (level * SV32_LEVEL_BITS))[(SV32_LEVEL_BITS - 1) .. 0]),
                                 PTE32_LOG_SIZE);
   let pte_addr = ptb + pt_ofs;
-  match (mem_read(Read(Data), to_phys_addr(pte_addr), 4, false, false, false)) {
+  match (mem_read_priv(Read(Data), Supervisor, to_phys_addr(pte_addr), 4, false, false, false)) {
     MemException(_) => {
 /*    print("walk32(vaddr=" ^ BitStr(vaddr) ^ " level=" ^ string_of_int(level)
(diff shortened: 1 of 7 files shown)


jrtc27:
(moreover, MPRV is not honoured for _any_ access's PMP checks)

model/riscv_vmem_sv32.sail
```diff
@@ -141,7 +141,7 @@ function translate32(asid, ptb, vAddr, ac, priv, mxr, do_sum, level, ext_ptw) =
                 n_ent.pte = n_pte.bits();
                 write_TLB32(idx, n_ent);
                 /* update page table */
-                match mem_write_value(to_phys_addr(EXTZ(ent.pteAddr)), 4, n_pte.bits(), false, false, false) {
+                match mem_write_value_priv(to_phys_addr(EXTZ(ent.pteAddr)), 4, n_pte.bits(), default_write_acc, Supervisor, false, false, false) {
```
jrtc27, line 144: Hm, the fact that `mem_write_value_priv` is always called with default_write_acc is probably a sign that parameter should be dropped, like how it doesn't exist for `mem_write_value`? Then `mem_write_value` should just call `mem_write_value_meta` directly (rather than constructing `Write(default_write_acc)` to call `effectivePrivilege`, throwing the access type away and `mem_write_value_priv` re-creating the same thing.

nwf: I struggled for a bit with the right shape of things here.  Are you suggesting, then,

```
/* Explicit AccessType, Privilege, and metadata */
val mem_write_value_meta : forall 'n, 0 < 'n <= max_mem_access . (xlenbits, atom('n), bits(8 * 'n), AccessType(ext_access_type), Privilege, mem_meta, bool, bool, bool) -> MemoryOpResult(bool) effect {wmv, wmvt, rreg, wreg, escape}

/* Explicit Privilege, default AccessType and metadata */
val mem_write_value_priv : forall 'n, 0 < 'n <= max_mem_access . (xlenbits, atom('n), bits(8 * 'n), Privilege, bool, bool, bool) -> MemoryOpResult(bool) effect {wmv, wmvt, rreg, wreg, escape}
function mem_write_value_priv (paddr, width, value, priv, aq, rl, con) =
￼  mem_write_value_meta(paddr, width, value, Write(default_write_acc), priv, default_meta, aq, rl, con)

/* Default AccessType, Privilege, and metadata */
function mem_write_value (paddr, width, value, aq, rl, con) = {
￼  let typ = Write(default_write_acc);
￼  let ep = effectivePrivilege(typ, mstatus, cur_privilege);
￼  mem_write_value_meta(paddr, width, value, typ, ep, default_meta, aq, rl, con)
￼}
```

jrtc27: Yeah, that's what I was thinking

nwf:
I've added the global/argument fix from #91 and also split `translateAddr` into a wrapper and `translateAddr_priv` to go along with our more-knobs-exposed `mem_` operations.  I think together this means there's no real need to shim `effectivePrivilege` any more and we can override both places an instruction will dispatch to the memory subsystem.

nwf:
CI failure seems to be `git` and not me?

jrtc27:
weird

jrtc27:
This seems fine, though I'm not sure what translateAddr_priv is for? Is it just there for theoretical completeness, or do you intend to use it for something?

nwf:
I intend to use it for the "as if U" instruction sketch.  I'm happy to pull it out of this PR.  But hold up a second; in the interest of making it easy for CHERI to pull up to a `sail-riscv` including this PR, I've added `mem_read_meta` and `mem_write_value_meta` which sit where you might imagine in the lattice of partial applications.

jrtc27:
Ah right yes that makes sense

rmn30:
Looks fine to me!

PR #97, opened 2021-07-16
adding riscof-plugin for sail-riscv c-emulator

neelgala:
This PR adds the plugin files required to run the SAIL RISC-V C-emulator as a reference model on the new Architectural Testing framework - RISCOF.

### DIFF
--- a/riscof-plugin/sail_cSim/README.md
+++ b/riscof-plugin/sail_cSim/README.md
@@ -1,0 +1,79 @@
+# Riscof Plugin for RISC-V SAIL C-Emulator
+
+This directory provides all the files and artifacts required for running SAIL RISC-V as a Reference model on the 
+[RISCOF](https://github.com/riscv/riscof) framework.
+
+## Getting SAIL RISC-V
+
+The model repository should be cloned from [here](https://github.com/rems-project/sail-riscv).
+
+## Building SAIL RISC-V C-Emulator
+
+The [README.md](../../README.md) at the top level of the sail-riscv repository gives details on 
+building C based executable of the model. You can build both the RV32 and RV64 variants.
+Once built, please add `$SAIL_RISCV/c_emulator` and `$SAIL_RISCV/ocaml_emulator` to your $PATH, 
+where $SAIL_RISCV is the path to the cloned repository.
+
+## Using SAIL RISC-V as a reference model in RISCOF
+
+You will need to edit the `config.ini` file required by RISCOF with the following 
+
+
+```
+# Name of the Plugin. All entries are case sensitive. Including the name.
+[sail_cSim]
+
+#path to the directory where the plugin is present. (required)
+pluginpath=/clone-path/sail-riscv/riscof-plugin/sail_cSim
+
+#path where the `riscv_sim_RV*` binaries are available. (optional)
+#if this value is absent, it is assumed that the binaries are available in the $PATH by default
+PATH=<path-to-binaries>
+
+#number of jobs to spawn in parallel while running tests on sail_cSim (optional)
+jobs=1
+
+#the make command to use while executing the make file. Any of bmake,cmake,pmake,make etc. (optional)
+#Default is make
+make=make
+```
+
+- Export command
+```
+pwd=pwd;export PYTHONPATH=$pwd:$PYTHONPATH;
+```
+
+## About the Plugin
+
+The sail_cSim plugin (`riscof_sail_cSim.py` file) present here has the following four functions:
+
+- \_\_init\_\_: This is the base initialization function. This function will check if the config.ini
+  section corresponding to `sail_cSim` for the following variables and set variables accordingly:
+
+  - PATH: if set in the config.ini file, then the plugin will use this as the sail_cSim executable, else
+    will assume it's available by default in the $PATH
+  - make: if set in config.ini then the plugin uses that command to execute the make file, else
+    default to using the `make` command
+  - jobs: if set in config.ini then the makefile is run with so many jobs in parallel, else defaults
+    to 1
+
+- initialize: This function basically sets up the compiler command for compiling tests. Currently
+  the plugin defaults to using `riscv[64/32]-unknown-elf-gcc` compiler. If you are using an
+  alternate toolchain or executable, you will need to edit this function.
+
+- build: This function will simply deduce the values of the a few compile arguments and check if the
+  required tooling available in the system.
+
+- runTests: This function will first create a Makefile, where each target of the Makefile
+  corresponds to compiling a specific test and then running it on the sail C-emulator. If this plugin is being
+  used to collect coverage of a test via isac, then this function will also append the isac commands
+  for each test in the corresponding Makefile target.
+
+  Note, that the march values for the compiler for each test can be different and are deduced from
+  the test-list provided to this function.
+
+  Once the Makefile is created, this function will execute the makefile targets in parallel
+  depending on the `jobs` value selected in the \_\_init\_\_ function.
+
+This plugin also supports coverage collection using ISAC. 
+
--- a/riscof-plugin/sail_cSim/__init__.py
+++ b/riscof-plugin/sail_cSim/__init__.py
@@ -1,0 +1,2 @@
+from pkgutil import extend_path
+__path__ = extend_path(__path__, __name__)
--- a/riscof-plugin/sail_cSim/env/link.ld
+++ b/riscof-plugin/sail_cSim/env/link.ld
@@ -1,0 +1,18 @@
+OUTPUT_ARCH( "riscv" )
+ENTRY(rvtest_entry_point)
+
+SECTIONS
+{
+  . = 0x80000000;
+  .text.init : { *(.text.init) }
+  . = ALIGN(0x1000);
+  .tohost : { *(.tohost) }
+  . = ALIGN(0x1000);
+  .text : { *(.text) }
+  . = ALIGN(0x1000);
+  .data : { *(.data) }
+  .data.string : { *(.data.string)}
+  .bss : { *(.bss) }
+  _end = .;
+}
+
--- a/riscof-plugin/sail_cSim/env/model_test.h
+++ b/riscof-plugin/sail_cSim/env/model_test.h
@@ -1,0 +1,60 @@
+#ifndef _COMPLIANCE_MODEL_H
+#define _COMPLIANCE_MODEL_H
+#if XLEN == 64
+  #define ALIGNMENT 3
+#else
+  #define ALIGNMENT 2
+#endif
+
+#define RVMODEL_DATA_SECTION \
+        .pushsection .tohost,"aw",@progbits;                            \
+        .align 8; .global tohost; tohost: .dword 0;                     \
+        .align 8; .global fromhost; fromhost: .dword 0;                 \
+        .popsection;                                                    \
+        .align 8; .global begin_regstate; begin_regstate:               \
+        .word 128;                                                      \
+        .align 8; .global end_regstate; end_regstate:                   \
+        .word 4;
+
+//RV_COMPLIANCE_HALT
+#define RVMODEL_HALT                                              \
+  li x1, 1;                                                                   \
+  write_tohost:                                                               \
+    sw x1, tohost, t5;                                                        \
+    j write_tohost;
+
+#define RVMODEL_BOOT
+
+//RV_COMPLIANCE_DATA_BEGIN
+#define RVMODEL_DATA_BEGIN                                              \
+  RVMODEL_DATA_SECTION                                                        \
+  .align ALIGNMENT;\
+  .global begin_signature; begin_signature:
+
+//RV_COMPLIANCE_DATA_END
+#define RVMODEL_DATA_END                                                      \
+  .global end_signature; end_signature:  
+
+//RVTEST_IO_INIT
+#define RVMODEL_IO_INIT
+//RVTEST_IO_WRITE_STR
+#define RVMODEL_IO_WRITE_STR(_R, _STR)
+//RVTEST_IO_CHECK
+#define RVMODEL_IO_CHECK()
+//RVTEST_IO_ASSERT_GPR_EQ
+#define RVMODEL_IO_ASSERT_GPR_EQ(_S, _R, _I)
+//RVTEST_IO_ASSERT_SFPR_EQ
+#define RVMODEL_IO_ASSERT_SFPR_EQ(_F, _R, _I)
+//RVTEST_IO_ASSERT_DFPR_EQ
+#define RVMODEL_IO_ASSERT_DFPR_EQ(_D, _R, _I)
+
+#define RVMODEL_SET_MSW_INT
+
+#define RVMODEL_CLEAR_MSW_INT
+
+#define RVMODEL_CLEAR_MTIMER_INT
+
+#define RVMODEL_CLEAR_MEXT_INT
+
+
+#endif // _COMPLIANCE_MODEL_H
--- a/riscof-plugin/sail_cSim/riscof_sail_cSim.py
+++ b/riscof-plugin/sail_cSim/riscof_sail_cSim.py
@@ -1,0 +1,122 @@
+import os
+import re
+import shutil
+import subprocess
+import shlex
+import logging
+import random
+import string
+from string import Template
+
+import riscof.utils as utils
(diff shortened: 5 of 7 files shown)


riscof-plugin/sail_cSim/riscof_sail_cSim.py
```diff
@@ -0,0 +1,213 @@
+import os
```
jrtc27, line 1: Run a linter like flake8 on this file, it's messy and not PEP-8 compliant

riscof-plugin/sail_cSim/riscof_sail_cSim.py
```diff
@@ -0,0 +1,213 @@
...
+	# capture the suite directory
+        self.suite = suite
+	
+	# capture the working directory.
+        self.work_dir = work_dir
+
+	# set the compile and objdump commands here. Note the march is not hardwired here, 
+	# because it will change for each test.
+        self.objdump_cmd = 'riscv{1}-unknown-elf-objdump -D {0} > {2};'
+        self.compile_cmd = 'riscv{1}-unknown-elf-gcc -march={0} \
```
jrtc27, line 79: Clang/LLVM needs to be as supported as GCC+GNU binutils

riscof-plugin/sail_cSim/riscof_sail_cSim.py
```diff
@@ -0,0 +1,213 @@
...
+
+	# capture the suite directory
+        self.suite = suite
+	
+	# capture the working directory.
+        self.work_dir = work_dir
+
+	# set the compile and objdump commands here. Note the march is not hardwired here, 
+	# because it will change for each test.
+        self.objdump_cmd = 'riscv{1}-unknown-elf-objdump -D {0} > {2};'
```
jrtc27, line 78: You need to shell escape everything (and does your library's add_target correctly escape that again for make?)

riscof-plugin/sail_cSim/README.md
```diff
@@ -0,0 +1,100 @@
+# Riscof Plugin for RISC-V SAIL C-Emulator
```
jrtc27, line 1: Why is all this stuff in a sail_cSim subdirectory? It's already in the sail-riscv repo, so sail is implied. Most (all?) of this should be usable for both the C and OCaml emulators, so the cSim is unnecessary, but even if it weren't, at least use the same naming convention (c_emulator) as sail-riscv.

riscof-plugin/sail_cSim/README.md
```diff
@@ -0,0 +1,100 @@
...
+## Getting SAIL RISC-V
+
+The model repository should be cloned from [here](https://github.com/rems-project/sail-riscv).
+
+## Building SAIL RISC-V C-Emulator
+
+The [README.md](../../README.md) at the top level of the sail-riscv repository gives details on 
+building C based executable of the model. You can build both the RV32 and RV64 variants.
+Once built, please add `$SAIL_RISCV/c_emulator` and `$SAIL_RISCV/ocaml_emulator` to your $PATH, 
+where $SAIL_RISCV is the path to the cloned repository.
```
jrtc27, line 15: None of this makes sense, this is in sail-riscv

riscof-plugin/sail_cSim/__init__.py
```diff
@@ -0,0 +1,2 @@
+from pkgutil import extend_path
+__path__ = extend_path(__path__, __name__)
```
jrtc27, line 2: Newline

riscof-plugin/sail_cSim/env/link.ld
```diff
@@ -0,0 +1,18 @@
+OUTPUT_ARCH( "riscv" )
```
jrtc27, line 1: There are going to be lots of different platforms out there that want this same linker script, or very similar versions to it (QEMU is going to be identical, as is Spike). I don't think this should be duplicated to every platform, there should probably be just one file inside riscof itself.

riscof-plugin/sail_cSim/env/model_test.h
```diff
@@ -0,0 +1,60 @@
+#ifndef _COMPLIANCE_MODEL_H
```
jrtc27, line 1: None of this looks specific to Sail, other than the tohost/fromhost usage that will still be common across Sail and QEMU. A basic common version of this belongs in riscof, not duplicated in every project.

riscof-plugin/sail_cSim/riscof_sail_cSim.py
jrtc27: 213 lines of boilerplate that presumably will need to change over time as riscof evolves doesn't seem like a good design to me. The only thing Sail-specific in here is how to run the emulators; how to use a RISC-V compiler and objdump, how to create a Makefile and how to run this ISAC thing have nothing to do with Sail, that should be in your riscof library. This should only contain things that relate specifically to Sail compared with other platforms.

riscof-plugin/sail_cSim/sail_isa.yaml
```diff
@@ -0,0 +1,38 @@
+hart_ids: [0]
```
jrtc27, line 1: Sail is both RV32 and RV64, so having just a sail_isa.yaml that's RV32 is wrong. Moreover, many aspects of the model are configurable (and it will eventually become even more so), so having this YAML file be static does not seem right either (e.g. RVC can be turned off).

riscof-plugin/sail_cSim/sail_isa.yaml
```diff
@@ -0,0 +1,38 @@
...
+  supported_xlen: [32]
+  misa:
+    reset-val: 0x0
+    rv32:
+      mxl:
+        implemented: false
+      extensions:
+        implemented: false
+      accessible: true
+    rv64:
```
jrtc27, line 14: Why is there an rv64 property in a file describing an RV32 configuration?

riscof-plugin/sail_cSim/sail_platform.yaml
```diff
@@ -0,0 +1,4 @@
+nmi:
+  label: nmi_vector
```
jrtc27, line 2: What are these things? These aren't defined anywhere in the Sail model nor in your PR here. These are just symbols that the old RISC-V ISA tests happen to use, but have no special meaning to the model.

riscof-plugin/sail_cSim/README.md
allenjbaum: Yes, it is in sail-riscv. Is it inaccurate?
The audience for this is people who are running this just to be able to use riscof, and the assumption is that this is where they will start. That assumption won't always be correct, but what matters is that sometimes it will be.

riscof-plugin/sail_cSim/env/link.ld
allenjbaum: There will be even more that have different linker scripts, and we have examples of those.
The linker script is customized to every model - we can make no assumptions what the platform looks like. 
The fact that some models will be the same (you happened to choose software models as opposed to RTL, which makes it more likely) doesn't change the reality that there will be a multitude of linker scripts, and they won't be hosted in one place, but hosted by the model itself,

riscof-plugin/sail_cSim/env/model_test.h
allenjbaum: The point here is exactly to have it in tied to every model. Just because two models share this doesn't mean that all models share this. The majority will not. Sail and QEMU may - but there are lot more models out there than just those two, and we need to make it work for all of them.

riscof-plugin/sail_cSim/README.md
jrtc27: The first section about cloning is completely unnecessary; they are already in the sail-riscv repo, either by browsing to it on GitHub or browsing to a local already-cloned copy.

The second section, fine, give a one-sentence pointer to the README.md at the top level, but "You can build both the RV32 and RV64 variants" seems completely unnecessary (it'd be highly surprising if that _weren't_ the case for the model).

riscof-plugin/sail_cSim/env/link.ld
jrtc27: So why can there not just be a single "software models" linker script in riscof? This is just promoting unnecessary copy/paste development practices.

riscof-plugin/sail_cSim/env/model_test.h
jrtc27: HTIF is a de-facto standard. Other platform-specific mechanisms are not so. That's the difference.

riscof-plugin/sail_cSim/sail_platform.yaml
allenjbaum: These are forward looking; SAIL doesn't currently support interrupts in general- but to be able to test interrupts, that support will be added (in a fairly restricted way). We are working on how that will interface to the tests first, before incorporating it into the Sail csim.

jrtc27: So then don't add it here until it's a thing that exists?

riscof-plugin/sail_cSim/env/link.ld
allenjbaum: Riscof has not idea whether a model is a SW model or not, and doesn't want to. It takes two models and compares them.

jrtc27: It doesn't need to; it can just provide a set of common implementations that platforms can optionally use. I just don't want to see three copies of the same files in sail-riscv, QEMU and Spike, which is what is currently going to happen and is a sign that the current approach could be improved.

jrtc27: Also, have you considered specifying a standard interface for these things rather than requiring platform-specific code in the first place? For example, making use of the semihosting spec? That seems like it would be ideal for some of this.

riscof-plugin/sail_cSim/env/model_test.h
allenjbaum: Yes, but our view is that we need to work with everybody, so anything that might be different has to be supplied - and that includes things that aren't different.  I'm not sure de-facto standard quite meets that bar, but we'll need to discuss that.
I think what you're asking for here (and in other places) is that there be a default that can be overwritten, which has a different set of issues. More discussion...

jrtc27: I think there are two things:
1. Defaults, like linking at 0x80000000 as used by almost all existing platforms
2. Optional common code for de-facto standards, but would still be opt-in, just there for you to use so you can say "I use this common standard method" without having to write your own version or make another copy of the same thing
I don't know if I'd class the HTIF in 1 or 2; in a sense it's 2, but because it's assumed by the existing ISA tests it's effectively an enforced testing standard at this point so may make sense even as a default.

riscof-plugin/sail_cSim/sail_platform.yaml
allenjbaum: Why not? The spec defines the concept of a platform defined NMI vector, so the platform YAML includes it.
But: this is part of a larger discussion of why Sail even needs its own YAML if the intent is to configure it to match some other model. Part of that is to make all plugins have the same format, so everything is interchangeable (e.g. we could have spike as a reference and compare the sail csim to it instead of the other way around. I'm not sure how good an excuse that is, so we will discuss that.

jrtc27: Defining an NMI vector is fine, but not if the thing you're defining it to doesn't exist, or is a property of the tests your running rather than the model (in which case it should be the test runner that knows what the NMI vector for the test should be).

riscof-plugin/sail_cSim/env/link.ld
arichardson: Linker scripts support a `INCLUDE` directive that could be used to avoid lots of copy-pasting. Having a few common .ld files that can be include will avoid lots of divergence and will make it easier to add new features.

Our previous CHERI-MIPS testsuite had various linker very similar scripts with slightly different parameters (e.g. running in cached vs uncached memory). Having all those scripts made the switch from ld.bfd to ld.lld extremely tedious since I had to look at all these very similar scripts and figure out from the git history which differences were actually indented and which of them were just cases of "this commit forgot to update 2 of the scripts but should have done so".

If riscof provides something like a `default-params.ld`
```
__base_addr__ = 0x80000000;
__default_align__ = 0x1000;

OUTPUT_ARCH( "riscv" )
ENTRY(rvtest_entry_point)
```
and a `default-sections.ld`
```
SECTIONS
{
  . = __base_addr__;
  .text.init : { *(.text.init) }
  . = ALIGN(__default_align__);
  .tohost : { *(.tohost) }
  . = ALIGN(__default_align__);
  .text : { *(.text) }
  . = ALIGN(__default_align__);
  .data : { *(.data) }
  .data.string : { *(.data.string)}
  .bss : { *(.bss) }
  _end = .;
}
```
Then all that sail and QEMU need to do would be something like
```
INCLUDE default-params.ld
INCLUDE default-sections.ld
```
(or just provide that as some software-model default linker script)

And to use a different start address you can just use 
```
INCLUDE default-params.ld
/* Override base address */
__base_addr__ = 0xc0000000;
INCLUDE default-sections.ld
```

riscof-plugin/sail_cSim/README.md
allenjbaum: There is a plugin for every model that riscof can run - that is every implementation by every vendor.
We wanted to ensure that if whoever changes a model keep the plugin from breaking, and keeping them in the same directory means they won't get out of sync. We don't want thousands of plugins (assuming they're all different) overloading the riscof repo with hundred or thousands of plugin PRs, so distributing them into the repo whose model is kept is much cleaner.

allenjbaum: Regarding Ocaml: unless and until we know for sure that the plugins are identical, we should probabaly keep it as-is.
If Ocaml is determined to be identical, AND we start using the Ocaml simulator for the arch rest flow, then we can think about a rename. Right now we have no plans to use the Ocaml.

riscof-plugin/sail_cSim/README.md
allenjbaum: We may just build exactly what we want into a docker image that takes sail as is with YAML_config + test suite in, and  reference signatures out, which would move everything you find problematic out of riscv-sail repo entirely, so the PR may disappear

riscof-plugin/sail_cSim/env/model_test.h
allenjbaum: Oddly, you first example is not shared by (my company's) platform. But I don't believe that statement even if we did. It maybe be true for the platforms you've looked at - but there are hundreds of vendors out there building harts with thousands of transistors, and with billions of transistors. And: we are not testing platforms, we are testing harts. We don't know what will be built into PMAs that we can't control and have no visibility into. We have no idea what will be a de-facto standard and have neither resources nor desire to survey them to see if there is one (or two or three) that appear to be at this very moment. That doesn't scale. What scales is letting every vendor specify their own. They are specifying their RTL, they can damn will specify a linker script that works with their RTL. If you want to create some common code and add it to the /doc directory as an example of a link script or whatever, go right ahead, we'd probably accept it. Otherwise, no. It's the vendor's responsibilty.

riscof-plugin/sail_cSim/sail_isa.yaml
allenjbaum: This is a skeleton showing the full schema. The actual YMAL used by a vendor wlll only fill in the parts it needs, but this is here to show what there is to be filled in.

riscof-plugin/sail_cSim/sail_platform.yaml
allenjbaum: The YAML is there to describe what the vendor has implemented according to the RISC-V spec.
There will be tests for this in the not too distant future, and we don't want to make vendors have to go back and modify their YAML as if they were changing their implementation when those tests are added to the ACT repo- because they're not. Only the tests are changing.

neelgala:
I think its best to close this PR for now. Since the arch-test as of now mandates using sail as a reference model, we plan to dispatch the python sail plugin as part of riscof's setup commands itself (i.e. hosted in the riscof-repo). The plugin therefore becomes a headache of the arch-test owners who are going to its be true maintainers and users of this plugin.

However, the only other arch-test requirement would be a cli-based versioning scheme implemented.. which I think is captured in another issue.

(Closing PR)

PR #99, opened 2021-07-30
scalar-crypto: Initial commit of 0.9.4 spec work

ben-marshall:
**Brief:** (PR information duplicated from commit message)

- This commit adds initial support for the scalar cryptography extension to RISC-V, synced with version [v0.9.4](https://github.com/riscv/riscv-crypto/releases/tag/v0.9.4-scalar) of the draft proposals.

- The riscv-crypto repository (https://github.com/riscv/riscv-crypto)  contains all of the work so far on the extension. See the [releases](https://github.com/riscv/riscv-crypto/releases) tab for the latest spec release.

**Details:**

- There are two main classes of functionality being added. The instructions and the entropy source.

- There is a considerable amount of common code (which will eventually also  be used by the vector crypto extension) which lives in  `model/riscv_types_kext.sail`. This consists of things like SBox definitions, block cipher operations and common structs/enums.

- The entropy source lives entirely within  `model/riscv_zkr.sail`. It is a CSR based interface to  an Entropy Source (note this is *not* the same as a "random number  generator") which can be used to _seed_ a DRBG.

**Additional:**

- I'm sorry this is such an enormous PR, and so hard to review. I didn't see a better way to break it down into smaller incremental pieces, but am of course happy to help explain things as needed. I'm also happy maintaining this code (with the help of the repo owners) during the rest of the standardisation process.

- The scalar cryptography extension is unique in that it *borrows* instructions from the Bit-manipulation extension. Since there is not yet any support for Bitmanip within `sail-riscv`, I have only included instructions unique to the scalar crypto extension. As and when Bitmanip support is added, I/we can work to make sure the right instructions are available to both extensions. Details of the relevant instructions are found in the [scalar crypto specification](https://github.com/riscv/riscv-crypto/releases).

- This is a fresh version of an old pull request against rems-project/sail-riscv (See rems-project/sail-riscv#80) which has been cleaned up for the v0.9.4 version of the specification.

Like I say, I'm sure there will be a certain amount of iteration before this can be merged. Please let me know what I can do to help or make things clearer. I also know that the sail-riscv repository is in the the process of being migrated to the `riscv` organisation. I wanted to open this PR now so it doesn't get forgotten or missed. I'm happy re-opening it against a new repo if need be.

Many thanks,
Ben


cetola:
Ben is still working on some fixes, but once complete we will need review. Added @martinberger and @jrtc27.

martinberger:
Just so that I am up-to-date. What is the relationship of the Verilog to the Sail code, have they already been verified against each other, or is the RTL only as a kind of comment?

model/riscv_types_kext.sail
```diff
@@ -0,0 +1,382 @@
...
+ */
+
+/* 
+ * Cryptography extension shared / utility functions
+ * ----------------------------------------------------------------------
+ */
+
+val ror32 : (bits(32), int) -> bits(32)
+function ror32(x, y) = {
+  (x>>to_bits(5,y)) | (x<<to_bits(5,32-y))
```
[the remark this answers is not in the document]
jrtc27: But I believe the following is a nicer way to support rotates (it compiles, possible I messed something up in the calculations though):
```diff
diff --git a/model/prelude.sail b/model/prelude.sail
index e4f9943..52435c8 100644
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -180,6 +180,28 @@ function shift_right_arith32 (v : bits(32), shift : bits(5)) -> bits(32) =
     let v64 : bits(64) = EXTS(v) in
     (v64 >> shift)[31..0]
 
+infix 7 >>>
+infix 7 <<<
+
+val rotate_bits_right : forall 'n 'm, 'm >= 0. (bits('n), bits('m)) -> bits('n)
+function rotate_bits_right (v, n) =
+    (v >> n) | (v << (to_bits(length(n), length(v)) - n))
+
+val rotate_bits_left : forall 'n 'm, 'm >= 0. (bits('n), bits('m)) -> bits('n)
+function rotate_bits_left (v, n) =
+    (v << n) | (v >> (to_bits(length(n), length(v)) - n))
+
+val rotater : forall 'm 'n, 'm >= 'n >= 0. (bits('m), atom('n)) -> bits('m)
+function rotater (v, n) =
+    (v >> n) | (v << (length(v) - n))
+
+val rotatel : forall 'm 'n, 'm >= 'n >= 0. (bits('m), atom('n)) -> bits('m)
+function rotatel (v, n) =
+    (v << n) | (v >> (length(v) - n))
+
+overload operator >>> = {rotate_bits_right, rotater}
+overload operator <<< = {rotate_bits_left, rotatel}
+
 /* helpers for mappings */
 
 val spc : unit <-> string
```
That way it's all polymorphic and also supports operator overloading to mirror shifts. It's also a standard enough function that it just belongs in the prelude rather than something specific to K (I assume bitmanip will want it, for example).

model/riscv_zkr.sail
```diff
@@ -0,0 +1,78 @@
+/*
+ * Contains a basic model for the RISC-V cryptography extension entropy
+ * source. See section 4 of the specifications for details.
+ * - https://github.com/riscv/riscv-crypto/releases
+ */
+
+
+/*
+ * The sentropy CSR
```
jrtc27, line 9: This is what's now seed, right?

model/riscv_types_kext.sail
```diff
@@ -0,0 +1,382 @@
...
+  aes_sbox_inv(x[31..24]) @
+  aes_sbox_inv(x[23..16]) @
+  aes_sbox_inv(x[15.. 8]) @
+  aes_sbox_inv(x[ 7.. 0]) 
+}
+
+/* Easy function to perform an SM4 SBox operation on 1 byte. */
+val sm4_sbox : bits(8) -> bits(8)
+function sm4_sbox(x) = {
+ sbox_lookup(x, sm4_sbox_table)
```
jrtc27, line 231: I see all of 1, 2 and 4 space indents in functions in this file.

Also one-liners don't need curly braces.

ben-marshall:
@martinberger - The example RTL in our riscv/riscv-crypto repository has been a little out of date for a while now if that's the RTL you meant? It's on my list to bring it back in line now things are frozen. We haven't done any formal checking against the Sail code - I didn't know there was a flow for that? If you can point me to it, I'd love to give it a go.

@jrtc27 - Thanks for all of your comments, I realise I should have marked this WiP. Now everything is frozen, I can update the entropy source (since that changed 3 times in three weeks) and fix the editorial issues. Since I seem to be so goodbad at flunking the existing code practices, would it help if I collated your comments into a style-guide doc of sorts to use going forward?

martinberger:
@ben-marshall I am not aware that there is an official flow for for comparing Sail with RTL. It would be great to have one. But I'm not sure we should have RTL in the official Sail model, unless it is proven equivalent.

martinberger:
> would it help if I collated your comments into a style-guide doc of sorts to use going forward?

@ben-marshall That would be lovely. I was thinking about something like this too. In the long run, I think an automatic formatter, like `gofmt` is the way to go.

ben-marshall:
>  In the long run, I think an automatic formatter, like gofmt is the way to go.

I absolutely agree! Okay, as I'm implementing all of my fixes then, I'll try to record them.

> But I'm not sure we should have RTL in the official Sail model, unless it is proven equivalent.

Also agreed. I think keeping RTL out of the Sail model repo completely is sensible.

model/riscv_insts_zks.sail
```diff
@@ -0,0 +1,92 @@
...
+
+/* Assembly string mappings for the SM4 instructions. */
+mapping clause assembly = SM4ED (bs,rs2,rs1,rd) <->
+    "sm4ed" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2) ^ sep() ^ hex_bits_2(bs)
+
+mapping clause assembly = SM4KS (bs,rs2,rs1,rd) <->
+    "sm4ks" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2) ^ sep() ^ hex_bits_2(bs)
+
+/* Execute clause for the sm4 instructions. */
+function clause execute (SM4ED (bs,rs2,rs1,rd)) = {
```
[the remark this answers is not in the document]
ben-marshall: Yeah, it's because the instructions use what's called a "T-Tables" based implementation strategy for SM4 (as with the 32-bit AES instructions) so there is no expectation that the Sail code would correspond to the original algorithm specification. I can add a note the description of the instructions for explaining this certainly. Unfortunately, I've found this is quite common in cryptography, where the "naive" algorithm implementation / specification looks absolutely nothing like the actual real world implementation.

model/riscv_insts_zkn_rv64.sail
```diff
@@ -0,0 +1,166 @@
...
+
+mapping clause assembly = AES64DS (rs2, rs1, rd)
+  <-> "aes64ds" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (AES64KS1I(rcon, rs1, rd)) = {
+  let tmp1 : bits(32) = X(rs1)[63..32];
+  let rc   : bits(32) = aes_decode_rcon(rcon);
+  let tmp2 : bits(32) = if (rcon ==0xA) then tmp1 else ror32(tmp1, 8);
+  let tmp3 : bits(32) = aes_subword_fwd(tmp2);
+  let result : bits(64) = (tmp3 ^ rc) @ (tmp3 ^ rc);
```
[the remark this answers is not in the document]
ben-marshall: I'd have to go back and very carefully study the AES spec and these instructions to remember Barry's explanation of how/why these instructions work as they do. But you are on the right track in suggesting its because they work across all parametrisations of AES. Which again, isn't intuitive from their specification.

model/riscv_insts_zkn_rv64.sail
```diff
@@ -0,0 +1,166 @@
...
+mapping clause assembly = AES64DSM (rs2, rs1, rd)
+  <-> "aes64dsm" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+mapping clause assembly = AES64DS (rs2, rs1, rd)
+  <-> "aes64ds" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (AES64KS1I(rcon, rs1, rd)) = {
+  let tmp1 : bits(32) = X(rs1)[63..32];
+  let rc   : bits(32) = aes_decode_rcon(rcon);
+  let tmp2 : bits(32) = if (rcon ==0xA) then tmp1 else ror32(tmp1, 8);
```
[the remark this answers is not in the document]
ben-marshall: Again, I'd have to go back and remind myself, but different parametrisations of AES have different key-schedules, hence some of the weird looking behaviour here.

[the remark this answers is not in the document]
ben-marshall: I remember now, It's needed for the AES-256 KeySchedule, where you need to be able to apply the SBox _without_ doing the rotation. This doesn't happen in the AES-192 and AES-128 KeySchedule operations.

model/riscv_sys_regs.sail
```diff
@@ -180,6 +180,22 @@ function haveUsrMode() -> bool = misa.U() == 0b1
 function haveNExt()    -> bool = misa.N() == 0b1
 /* see below for F and D extension tests */
 
+/* Cryptography extension support. Note these will need updating once */
```
ben-marshall, line 183: I hope this is a sensible way to put these functions in. The idea is that they will be updated when riscv-config / dynamic extension configuration is implemented in Sail. If not, please let me know where would be a better place.

I tried to replace the `sizeof(xlen) == X` with `cur_Architecture` calls, but it meant the `encdec` clauses had mismatched side-effects for some reason (`{reg}` v.s. `{reg, escape}`) and this is beyond my Sail knowledge to address.

model/riscv_ext_regs.sail
```diff
@@ -83,13 +83,17 @@ function ext_rvfi_init () = {
...
 }
 
-
 /*!
 THIS(csrno, priv, isWrite) allows an extension to block access to csrno,
 at Privilege level priv. It should return true if the access is allowed.
 */
 val ext_check_CSR : (bits(12), Privilege, bool) -> bool
-function ext_check_CSR (csrno, p, isWrite) = true
+scattered function ext_check_CSR
```
ben-marshall, line 91: I understand I may have done something silly here so I'm explicitly asking for feedback. It felt like extensions needed to be able to add clauses to this function ad-hoc (Zkr being one of them), hence I've tried to add this ability in by turning it into a scattered function. See the `riscv_zkr.sail` file for how it get's used to check if access to the `seed` CSR is allowed.

jrtc27: The various ext_* functions are for out-of-tree extensions like CHERI-RISC-V that need to override them with their own hooks, not for standard in-tree extensions.

ben-marshall: _Oh_. So, actually I need to revert this change, and also change the functions which the Zkr implementation uses then, since they all use `ext_*` functions as well. I assumed `ext` meant Extension in the RISC-V sense, not in an out of tree sense.

ben-marshall: Hmm okay. So, `seed` is a weird CSR in that _just_ reading it causes an Illegal Opcode Exception. One has to write to it in order to get the read value, and the written value is thrown away / ignored. This was to avoid having a standardised CSR which has side-effects on a read.

It's not obvious how this can be supported given how the current `execute CSR` function clause is written, unless I'm missing something? I can probably make the modifications as needed, but they look quite invasive to me.

model/riscv_insts_zkn_rv32.sail
```diff
@@ -0,0 +1,170 @@
+
```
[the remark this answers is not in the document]
ben-marshall: Apologies, dumb question time: How do I get the low `XLEN` bits of a 64-bit value? In my naivety I tried
`result[xlen..0]` and `result[sizeof(xlen)..0]`, but no. I'm unaware of another example of this in the architecture.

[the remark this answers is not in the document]
ben-marshall: Meta comment - this feels like it's going to make the Sail code much more counter-intuitive to read.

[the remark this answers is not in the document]
jrtc27: Yeah, but if UXL are ever to be supported then I don't think there's a way to avoid that.

To answer your question, it's `result[(sizeof(xlen) - 1) .. 0]`; see the code to read mcycle/minstret for examples of needing this elsewhere in the model, though it's fairly rare.

model/riscv_sys_regs.sail
```diff
@@ -736,3 +752,71 @@ function legalize_satp32(a : Architecture, o : bits(32), v : bits(32)) -> bits(3
...
+  ES16 <-> 0b10,
+  DEAD <-> 0b11
+}
+
+/*
+ * Entropy Source - Platform access to random bits.
+ * WARNING: This function currently lacks a proper side-effect annotation.
+ *          If you are using theorem prover tool flows, you
+ *          may need to modify or stub out this function for now.
+ * NOTE: This would be better placed in riscv_platform.sail, but that file
```
ben-marshall, line 783: I hope this is an acceptable compromise? I moved the Zkr code to be among the other CSR instructions. By the time I'd fixed the use of `ext_*` functions, there seemed no point it living in a separate file. Happy changing this if there's a better way.

model/riscv_insts_zkn.sail
```diff
@@ -0,0 +1,412 @@
...
+  <-> "aes64es" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+mapping clause assembly = AES64DSM (rs2, rs1, rd)
+  <-> "aes64dsm" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+mapping clause assembly = AES64DS (rs2, rs1, rd)
+  <-> "aes64ds" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (AES64KS1I(rcon, rs1, rd)) = {
+  if(sizeof(xlen)==32) then RETIRE_FAIL else {
```
ben-marshall, line 283: I've cheated / compromised here. This allows it to compile the same source for RV32/64. It _doesn't_ support dynamic XLEN changes, but it does keep the code as readable as possible. Solving how to structure the code for dynamic XLEN changes is clearly a problem to solve, but I don't feel like I can solve it now without making this PR substantially more difficult for people to read and manage.

I hope this is acceptable, but if it's imperative to solve the dynamic XLEN problem now as much as possible, I can do the necessary changes.

jrtc27: You should be able to just `assert(sizeof(xlen) != 32);` at the start of these (also has the advantage that it blows up if you messed up the decode predicates, rather than silently being a NOP that doesn't increment minstret; RETIRE_FAIL only indicates whether or not the instruction should increase minstret, it's up to the instruction to actually raise the relevant exception).

ben-marshall: Oh that's much nicer, I'll do that.

model/prelude.sail
```diff
@@ -248,6 +248,15 @@ function shift_right_arith32 (v : bits(32), shift : bits(5)) -> bits(32) =
     let v64 : bits(64) = EXTS(v) in
     (v64 >> shift)[31..0]
 
+val ror32 : (bits(32), int) -> bits(32)
```
jrtc27, line 251: Please use the polymorphic version I provided rather than writing multiple monomorphised copies.

ben-marshall: Yeah I will, I'm trying to do one thing at a time is all.

ben-marshall: If it makes it easier for you, feel free to ignore my pushes until I say "hey, I think this is ready" again. I realise it's a bit frustrating otherwise!

jrtc27: Oh ok, that would be helpful, yes :)

ben-marshall:
Hi @martinberger and @jrtc27 

I _think_ I've addressed the things you have pointed out / asked for so far. I don't doubt there is room for improvement, but I'm not confident enough to judge what that looks like now.

Some known issues:
- Access control for the entropy source is dependent on the `mseccfg` CSR, which is _just now_ in public review. I've noted the [TODO](https://github.com/riscv/sail-riscv/pull/99/files#diff-e89888d2ba8a243ac03cfe4ecce7acdb820a3135cafc486394d369a2054712e9R205) items where this code will need updating.
- Because the riscv-config (or it's latest iteration) isn't final, the gating of instructions based on whether extensions are implemented or not is done with some stub functions which currently [always return true](https://github.com/riscv/sail-riscv/pull/99/files#diff-e1a3fe7050c3af4c308a5e39bfb7f0f31094022b9892cee9a611bcb57ccb6726R186). They will need changing to support dynamic configuration of extensions and runtime changes of XLEN, depending on what the implementation of that looks like.

Otherwise:
- I've added Jessica's polymorphic rotation implementation and changed the code I've written to use it.
- There are still parts of the scalar crypto extensions (`Zbkb/c/x`) which this PR doesn't add. I'd like to get these instructions merged in first, then I'll turn to the `Zbkb/c/x` instructions. There is some overlap with Bitmanip there which I need to coordinate. I hope submitting things in pieces like this is still okay.

I'm sorry we jumped the gun a bit in asking for a review. Everything you said was still very relevant / would have been necessary if I'd had the PR in a finished state already, so it was all useful. Please let me know anything else you need from me and I'll get on it. I don't want to take up any more of your time than I have, and I'm very grateful for your patience!

Cheers,
Ben

jjscheel:
@martinberger, would appreciate some insight into next steps for this PR for @ben-marshall.  Thanks!

martinberger:
> @martinberger, would appreciate some insight into next steps for this PR for @ben-marshall. Thanks!

I've looked at the current version and it looks good to me.

model/riscv_sys_regs.sail
```diff
@@ -180,6 +180,22 @@ function haveUsrMode() -> bool = misa.U() == 0b1
...
 /* see below for F and D extension tests */
 
+/* Cryptography extension support. Note these will need updating once */
+/* Sail can be dynamically configured with different extension support */
+/* and have dynamic changes of XLEN via S/UXL */
+function haveZkr()    -> bool = true
+function haveZksh()   -> bool = true
+function haveZksed()  -> bool = true
+function haveZknh()   -> bool = true
+function haveZknh32() -> bool = haveZknh() & sizeof(xlen) == 32
```
jrtc27, line 190: These aren't real extension names, and if they were would mean version 32.0 of the Zknh extension; please just inline these as is done for the various other XLEN-specific instructions

model/riscv_sys_regs.sail
jrtc27: Yeah that comes from the `internal_error`. Not sure if that's quite the right way to implement that, might be better as an actual assertion as, as you note, it's a bit annoying. But this is fine, not your problem to solve, and everything else uses static XLEN checks.

model/riscv_sys_regs.sail
jrtc27: This is fine, though can the platform function not just return `bits(16)` and thus you just have `val get_16_random_bits = { ... }` and no need for a wrapper? Then the naming matches other functions in this file, some of which do still use platform functions but don't include a plat_ prefix in their Sail names.

model/riscv_types_kext.sail
```diff
@@ -0,0 +1,355 @@
...
+
+/* 
+ * Cryptography extension shared / utility functions
+ * ----------------------------------------------------------------------
+ */
+
+/* Auxiliary function for performing GF multiplicaiton */
+val xt2 : bits(8) -> bits(8)
+function xt2(x) = {
+  (x << 1) ^ ( match (bit_to_bool(x[7]) ) {
```
jrtc27, line 16: Weird spacing here around the parens (which then also is a bit misleading as it looks like the ones with spaces around them are a pair, but they're not, and that wouldn't make sense), also just use if/else instead of matching on a bool.

model/riscv_types_kext.sail
```diff
@@ -0,0 +1,355 @@
...
+|]
+
+/* Lookup function - takes an index and a list, and retrieves the
+ * x'th element of that list.
+ */
+val sbox_lookup : (bits(8), list(bits(8))) -> bits(8)
+function sbox_lookup(x, table) = {
+  match (x, table) {
+    (0x00, t0::tn) => t0,
+    ( y  , t0::tn) => sbox_lookup(x - 0x01, tn)
```
jrtc27, line 173:
```suggestion
    (   y, t0::tn) => sbox_lookup(x - 0x01, tn)
```

Makefile
```diff
@@ -22,8 +22,11 @@ endif
 SAIL_CHECK_SRCS = riscv_addr_checks_common.sail riscv_addr_checks.sail riscv_misa_ext.sail
 SAIL_DEFAULT_INST = riscv_insts_base.sail riscv_insts_aext.sail riscv_insts_cext.sail riscv_insts_mext.sail riscv_insts_zicsr.sail riscv_insts_next.sail riscv_insts_hints.sail
 SAIL_DEFAULT_INST += riscv_insts_fext.sail riscv_insts_cfext.sail
+SAIL_DEFAULT_INST += riscv_insts_zkn.sail
+SAIL_DEFAULT_INST += riscv_insts_zks.sail
 ifeq ($(ARCH),RV64)
 SAIL_DEFAULT_INST += riscv_insts_dext.sail riscv_insts_cdext.sail
+else
```
jrtc27, line 29:
```suggestion
```

Makefile
```diff
@@ -22,8 +22,11 @@ endif
 SAIL_CHECK_SRCS = riscv_addr_checks_common.sail riscv_addr_checks.sail riscv_misa_ext.sail
 SAIL_DEFAULT_INST = riscv_insts_base.sail riscv_insts_aext.sail riscv_insts_cext.sail riscv_insts_mext.sail riscv_insts_zicsr.sail riscv_insts_next.sail riscv_insts_hints.sail
 SAIL_DEFAULT_INST += riscv_insts_fext.sail riscv_insts_cfext.sail
+SAIL_DEFAULT_INST += riscv_insts_zkn.sail
+SAIL_DEFAULT_INST += riscv_insts_zks.sail
```
jrtc27, line 26: Maybe put these after D just below so D is kept with F?

martinberger (approved):

model/riscv_sys_regs.sail
ben-marshall: There _was_ once a reason for having a separate wrapper function but I think it went away as I learned to Sail. So yes, wrapper removed and `get_16_random_bits` now returns `bits(16)`.

model/riscv_types_kext.sail
ben-marshall: Yeah this was a cursed use of `match`. Changed to `if`.

model/riscv_sys_regs.sail
ben-marshall: Sure thing, I've done this in the latest commits.

ben-marshall:
@jrtc27, @martinberger, @jjscheel - Hopefully I've addressed all of the remaining comments, apologies it took me a while to get round to.

Cheers,
Ben

jrtc27:
@ben-marshall will you be fixing the outstanding issues highlighted above with your changes?

ben-marshall:
Hi @jrtc27 - Yes they are on my todo list. Sorry I should have replied earlier saying so. I'm hoping to get to it on Friday.

Issue #101, opened 2021-08-06
Incorrect SV48_Vaddr bitfield


dylux:
Current implementation of SV48_Vaddr in riscv_vmem_common.sail is an exact copy of SV39_Vaddr, which is incorrect since the VPNi field should have a range of 47 .. 12.
https://github.com/rems-project/sail-riscv/blob/cf099be2a1aff1f326dc855875d5732607e28752/model/riscv_vmem_common.sail#L193-L196

martinberger:
We reviewed it and the fix looks fine, we will merge it in soon.

PR #105, opened 2021-08-21
Fix incorrect SV48_Vaddr bitfield

dylux:
This is a fix for #101.
CC: @cento


jrtc27 (approved):
This looks correct. I assume this was never discovered because no OSes actually use Sv48 due to the lack of hardware implementations that support it, and the ISA tests don't have any coverage of Sv48. It looks like the same is true of the new architecture tests so far, too.

martinberger:
> looks fine to me, and a good reminder for the ISA tests discussion

This looks like a case where suitable boundary value tests could be auto-generated, at least in principle, from the very definition of the SV48_Vaddr datatype.

PR #107, opened 2021-09-27
Add missing MPRV clearing on MRET/SRET

scottj97:
The 1.12 privileged spec [requires](https://github.com/riscv/riscv-isa-manual/blob/d29cb13885d7dfaae9aaa27b45e44733254ead37/src/machine.tex#L689-L690):

> An MRET or SRET instruction that changes the privilege mode to a mode less privileged than M also sets MPRV=0.

In this PR, SRET is implemented a little differently than MRET, because SRET always returns to a mode less privileged than M; whereas MRET might stay in M-mode.

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -456,6 +456,8 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       mstatus->MIE()  = mstatus.MPIE();
       mstatus->MPIE() = 0b1;
       cur_privilege   = privLevel_of_bits(mstatus.MPP());
+      if cur_privilege != Machine
+      then mstatus->MPRV() = 0b0;
       mstatus->MPP()  = privLevel_to_bits(if haveUsrMode() then User else Machine);
       if   cur_privilege != Machine
       then mstatus->MPRV() = 0b0;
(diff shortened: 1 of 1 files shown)


scottj97:
> ~SRET can return to S-mode~ wait I misread, those semantics are _weird_

The goal is that in any mode <M, MPRV will always be 0. So any time you leave M-mode (which is only possible via SRET or MRET), MPRV is cleared.

jrtc27:
Never mind I was just being stupid, I interpreted MPRV as MPP and got very confused. This makes more sense.

model/riscv_sys_control.sail
```diff
@@ -456,6 +456,8 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       mstatus->MIE()  = mstatus.MPIE();
       mstatus->MPIE() = 0b1;
       cur_privilege   = privLevel_of_bits(mstatus.MPP());
+      if cur_privilege != Machine
+      then mstatus->MPRV() = 0b0;
       mstatus->MPP()  = privLevel_to_bits(if haveUsrMode() then User else Machine);
       if   cur_privilege != Machine
```
jrtc27, line 462: This code is already here for MRET

model/riscv_sys_control.sail
```diff
@@ -473,6 +475,7 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       mstatus->SIE()  = mstatus.SPIE();
       mstatus->SPIE() = 0b1;
       cur_privilege   = if mstatus.SPP() == 0b1 then Supervisor else User;
+      mstatus->MPRV() = 0b0;
       mstatus->SPP()  = 0b0;
       if   cur_privilege != Machine
       then mstatus->MPRV() = 0b0;
```
jrtc27, line 481: As is this code (yes technically a redundant check but meh)

scottj97:
Well I'll be! I was using a very outdated tree that did not have this fix, and I didn't see it in the recent commit messages.

It was added in [April 2020](https://github.com/riscv/sail-riscv/commit/555f47a4c94107ea2f350b96692f6bc334cf1516)

PR #116, opened 2021-10-15
Support BitManip Zba, Zbb, Zbc and Zbs extensions

bilalsakhawat-10xe:
This request is for BitManip Zba, Zbb, Zbc and Zbs extensions Support (https://github.com/riscv/riscv-bitmanip/releases/tag/1.0.0) in the Model. 

Kindly review this request and suggest changes. Thanks!


jrtc27:
Don't have time for a full review right now, but at a glance you have bad formatting (inconsistent whitespace, overly-long lines and whitespace where it doesn't belong like after an opening parenthesis) and lack checks that the extension is enabled (see the scalar-crypto PR for an example of how to do this).

bilalsakhawat-10xe:
> > Don't have time for a full review right now, but at a glance you have bad formatting (inconsistent whitespace, overly-long lines and whitespace where it doesn't belong like after an opening parenthesis) and lack checks that the extension is enabled (see the scalar-crypto PR for an example of how to do this).
> 
> Sure, Thanks!

@jrtc27 done.

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+function clause execute (ZBB_RTYPEW(rs2, rs1, rd, op)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  let result : xlenbits = match op {
+    RISCV_ROLW => EXTS((rs1_val[31..0] << rs2_val[4..0]) | (rs1_val[31..0] >> (0x20 - EXTZ(rs2_val[4..0])))),
+    RISCV_RORW => EXTS((rs1_val[31..0] >> rs2_val[4..0]) | (rs1_val[31..0] << (0x20 - EXTZ(rs2_val[4..0]))))
+  };
+  X(rd) = EXTS(result);
+  RETIRE_SUCCESS
+}
```
abukharmeh, line 65: Why don't you factor the bit selection, and rename rs2_val to shimt ? (Again, just another suggestion !)

bilalsakhawat-10xe: No particular reason. I'll update this.

model/riscv_insts_zba.sail
```diff
@@ -0,0 +1,88 @@
...
+  let rs2_val = X(rs2);
+  let result : xlenbits = match op {
+    RISCV_ADDUW    => EXTZ(rs1_val[31..0]) + rs2_val,
+    RISCV_SH1ADDUW => (EXTZ(rs1_val[31..0]) << 1) + rs2_val,
+    RISCV_SH2ADDUW => (EXTZ(rs1_val[31..0]) << 2) + rs2_val,
+    RISCV_SH3ADDUW => (EXTZ(rs1_val[31..0]) << 3) + rs2_val
+  };
+  X(rd) = EXTS(result);
+  RETIRE_SUCCESS
+}
```
abukharmeh, line 57: Also here you could factor out EXTZ(rs1_val[31..0]) to the let statement? (Just another suggestion)

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+  <-> 0b011010111000 @ rs1 @ 0b101 @ rd @ 0b0010011
+
+mapping clause assembly = RISCV_REV8(rs1, rd)
+  <-> "rev8" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1)
+
+function clause execute (RISCV_REV8(rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let result : xlenbits = match sizeof(xlen) {
+    32 => ((rs1_val & 0xff000000) >> 24) | ((rs1_val & 0x00ff0000) >>  8) |
+          ((rs1_val & 0x0000ff00) <<  8) | ((rs1_val & 0x000000ff) << 24),
```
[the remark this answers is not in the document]
abukharmeh: I am slowly reading each instruction implementation and encoding, sorry for the slow review.

abukharmeh:
I just compiled the model because I was curious how some things map to C and noticed that we have a lot more incomplete statements warnings because of some of the added enums. I wonder if it's a good practice to place a wild card (_) at the end of such mappings with an assert statement to remove such warnings and make sure we always have a match. @martinberger  @jrtc27   What do we think?

martinberger:
>  I wonder if it's a good practice to place a wild card (_) at the end of such mappings with an assert(false) statement to remove such warnings and make sure we always have a match. @martinberger @jrtc27 What do we think?

This is a somewhat philosophical question. Once school of (software engineering) thought says you should never have wildcards in pattern matching, since otherwise it's too forget updating pattern matching when the number of cases increases. Clearly `assert(false)` improves this, since forgotten cases will be caught, albeit only at run-time.

jrtc27:
> > I wonder if it's a good practice to place a wild card (_) at the end of such mappings with an assert(false) statement to remove such warnings and make sure we always have a match. @martinberger @jrtc27 What do we think?
> 
> This is a somewhat philosophical question. Once school of (software engineering) thought says you should never have wildcards in pattern matching, since otherwise it's too forget updating pattern matching when the number of cases increases. Clearly `assert(false)` improves this, since forgotten cases will be caught, albeit only at run-time.

Unmatched values are an error already and will abort at run time if hit, all `assert(false)` does is silence the warnings. There are cases when that is the right thing to do, but they are rare, almost always it tells you your types should be improved.

In this case, the problem is that the same enum is used across multiple Zb* extensions; e.g. ZB{A,B,S}_RYTPE all take a brop, which contains the operations for all three, so each of the execute clauses has an incomplete match. This is not a proper use of types; the correct thing to do is split the enums up.

model/riscv_insts_zbc.sail
```diff
@@ -0,0 +1,65 @@
+/* ****************************************************************** */
+/* This file specifies the instructions in the BitManip Zbc extension */
```
jrtc27, line 2: Delete

model/riscv_insts_zbc.sail
```diff
@@ -0,0 +1,65 @@
...
+  <-> "clmul" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (RISCV_CLMUL(rs2, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  result : xlenbits = EXTZ(0x0);
+  foreach (i from 0 to (xlen_val - 1) by 1) {
+    if ((rs2_val >> i)[0] == bitone)
+      then result = result ^ (rs1_val << i);
+  };
```
jrtc27, line 20:
```suggestion
  result : xlenbits = zeros()
  foreach (i from 0 to (xlen_val - 1))
    if rs2_val[i] == bitone then result = result ^ (rs1_val << i);
```

model/riscv_insts_zbc.sail
```diff
@@ -0,0 +1,65 @@
...
+  <-> "clmulh" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (RISCV_CLMULH(rs2, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  result : xlenbits = EXTZ(0x0);
+  foreach (i from 0 to (xlen_val - 1) by 1) {
+    if ((rs2_val >> i)[0] == bitone)
+      then result = result ^ (rs1_val >> (xlen_val - i));
+  };
```
jrtc27, line 41:
```suggestion
  result : xlenbits = zeros()
  foreach (i from 0 to (xlen_val - 1))
    if rs2_val[i] == bitone then result = result ^ (rs1_val >> (sizeof(xlen) - i));
```

model/riscv_insts_zbc.sail
```diff
@@ -0,0 +1,65 @@
...
+  <-> "clmulr" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (RISCV_CLMULR(rs2, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  result : xlenbits = EXTZ(0x0);
+  foreach (i from 0 to (xlen_val - 1) by 1) {
+    if ((rs2_val >> i)[0] == bitone)
+      then result = result ^ (rs1_val >> (xlen_val - i - 1));
+  };
```
jrtc27, line 62:
```suggestion
  result : xlenbits = zeros()
  foreach (i from 0 to (xlen_val - 1))
    if rs2_val[i] == bitone then result = result ^ (rs1_val >> (sizeof(xlen) - i - 1));
```

model/riscv_insts_zbs.sail
```diff
@@ -0,0 +1,94 @@
+/* ****************************************************************** */
+/* This file specifies the instructions in the BitManip Zbs extension */
```
jrtc27, line 2: Delete

model/riscv_insts_zba.sail
```diff
@@ -0,0 +1,88 @@
+/* ****************************************************************** */
+/* This file specifies the instructions in the BitManip Zba extension */
+
+
+/* ****************************************************************** */
+union clause ast = RISCV_SLLIUW : (bits(6), regidx, regidx)
+
+mapping clause encdec = RISCV_SLLIUW(shamt, rs1, rd) if haveZba() & sizeof(xlen) == 64
```
jrtc27, line 8: Guards need to be on both directions of the mapping (repeated throughout)

model/riscv_insts_zbs.sail
```diff
@@ -0,0 +1,94 @@
...
+    RISCV_BEXTI => if sizeof(xlen) == 32
+                     then (rs1_val >> shamt[4..0]) & EXTZ(0x1)
+                     else (rs1_val >> shamt[5..0]) & EXTZ(0x1),
+    RISCV_BINVI => if sizeof(xlen) == 32
+                     then rs1_val ^ (EXTZ(0b1) << shamt[4..0])
+                     else rs1_val ^ (EXTZ(0b1) << shamt[5..0]),
+    RISCV_BSETI => if sizeof(xlen) == 32
+                     then rs1_val | (EXTZ(0b1) << shamt[4..0])
+                     else rs1_val | (EXTZ(0b1) << shamt[5..0])
+  };
```
jrtc27, line 45:
```suggestion
  let rs1_val = X(rs1);
  let mask : xlenbits = if sizeof(xlen) == 32
                        then EXTZ(0b1) << shamt[4..0]
                        else EXTZ(0b1) << shamt;
  let result : xlenbits = match op {
    RISCV_BCLRI => rs1_val & ~(mask),
    RISCV_BEXTI => EXTZ(if sizeof(xlen) == 32
                        then rs1_val[shamt[4..0]]
                        else rs1_val[shamt[5..0]]),
    RISCV_BINVI => rs1_val ^ mask,
    RISCV_BSETI => rs1_val | mask
  };
```

or

```suggestion
  let rs1_val = X(rs1);
  let mask : xlenbits = if sizeof(xlen) == 32
                        then EXTZ(0b1) << shamt[4..0]
                        else EXTZ(0b1) << shamt;
  let result : xlenbits = match op {
    RISCV_BCLRI => rs1_val & ~(mask),
    RISCV_BEXTI => bool_to_bits((rs1_val & mask) != zeros()),
    RISCV_BINVI => rs1_val ^ mask,
    RISCV_BSETI => rs1_val | mask
  };
```

not sure which is better, EXTI doesn't quite fit in with the rest, personally lean towards the latter to make it fit better even though the former is more idiomatic viewed on its own

model/riscv_sys_regs.sail
```diff
@@ -190,6 +190,12 @@ function haveZknh()   -> bool = true
 function haveZkne()   -> bool = true
 function haveZknd()   -> bool = true
 
+/* BitManip extension support. */
+function haveZba()  -> bool = true
```
jrtc27, line 194: Sort above Zk*

model/riscv_types.sail
```diff
@@ -390,6 +390,18 @@ enum amoop = {AMOSWAP, AMOADD, AMOXOR, AMOAND, AMOOR,
               AMOMIN, AMOMAX, AMOMINU, AMOMAXU}   /* AMO ops */
 enum csrop = {CSRRW, CSRRS, CSRRC}                /* CSR ops */
 
+enum brop = {RISCV_SH1ADD, RISCV_SH2ADD, RISCV_SH3ADD, RISCV_BCLR,
```
jrtc27, line 393: As discussed, split these up so there are per-Zb* versions of each enum to avoid incomplete pattern match warnings

model/riscv_insts_zbs.sail
```diff
@@ -0,0 +1,94 @@
...
+    RISCV_BEXT => if sizeof(xlen) == 32
+                    then (rs1_val >> rs2_val[4..0]) & EXTZ(0x1)
+                    else (rs1_val >> rs2_val[5..0]) & EXTZ(0x1),
+    RISCV_BINV => if sizeof(xlen) == 32
+                    then rs1_val ^ (EXTZ(0b1) << rs2_val[4..0])
+                    else rs1_val ^ (EXTZ(0b1) << rs2_val[5..0]),
+    RISCV_BSET => if sizeof(xlen) == 32
+                    then rs1_val | (EXTZ(0b1) << rs2_val[4..0])
+                    else rs1_val | (EXTZ(0b1) << rs2_val[5..0])
+  };
```
jrtc27, line 91:
```suggestion
  let mask : xlenbits = if sizeof(xlen) == 32
                        then EXTZ(0b1) << rs2_val[4..0]
                        else EXTZ(0b1) << rs2_val[5..0];
  let result : xlenbits = match op {
    RISCV_BCLR => rs1_val & ~(mask),
    RISCV_BEXT => EXTZ(bool_to_bits((rs1_val & mask) != zeros())),
    RISCV_BINV => rs1_val ^ mask,
    RISCV_BSET => rs1_val | mask
  };
```
or the equivalent alternative of the other BEXTI suggestion

Makefile
```diff
@@ -27,6 +27,10 @@ SAIL_DEFAULT_INST += riscv_insts_dext.sail riscv_insts_cdext.sail
 endif
 SAIL_DEFAULT_INST += riscv_insts_zkn.sail
 SAIL_DEFAULT_INST += riscv_insts_zks.sail
+SAIL_DEFAULT_INST += riscv_insts_zba.sail
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: Done.

model/riscv_insts_zba.sail
bilalsakhawat-10xe: Done.

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+  <-> 0b011000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011
+
+mapping clause assembly = RISCV_RORI(shamt, rs1, rd)
+  <-> "rori" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ hex_bits_6(shamt)
+
+function clause execute (RISCV_RORI(shamt, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let result : xlenbits = if sizeof(xlen) == 32
+                            then (rs1_val >> shamt[4..0]) | (rs1_val << (0x20 - EXTZ(shamt[4..0])))
+                            else (rs1_val >> shamt[5..0]) | (rs1_val << (0x40 - EXTZ(shamt[5..0])));
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: Rotate right ">>>" operator is not available in Sail prelude.sail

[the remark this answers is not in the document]
jrtc27: It was added by me for the scalar crypto work, rebase and it'll be there

model/riscv_insts_zba.sail
bilalsakhawat-10xe: Done.

model/riscv_insts_zba.sail
```diff
@@ -0,0 +1,88 @@
...
+
+function clause execute (RISCV_SLLIUW(shamt, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let result : xlenbits = EXTZ(rs1_val[31..0]) << shamt;
+  X(rd) = result;
+  RETIRE_SUCCESS
+}
+
+/* ****************************************************************** */
+union clause ast = ZBA_RTYPEW : (regidx, regidx, regidx, bropw)
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: Done.

model/riscv_insts_zbs.sail
bilalsakhawat-10xe: Done.

model/riscv_sys_regs.sail
bilalsakhawat-10xe: Done.

model/riscv_types.sail
bilalsakhawat-10xe: Done.

model/riscv_insts_zbs.sail
bilalsakhawat-10xe: Done.

model/riscv_insts_zba.sail
```diff
@@ -0,0 +1,88 @@
...
+
+function clause execute (ZBA_RTYPEW(rs2, rs1, rd, op)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  let result : xlenbits = match op {
+    RISCV_ADDUW    => EXTZ(rs1_val[31..0]) + rs2_val,
+    RISCV_SH1ADDUW => (EXTZ(rs1_val[31..0]) << 1) + rs2_val,
+    RISCV_SH2ADDUW => (EXTZ(rs1_val[31..0]) << 2) + rs2_val,
+    RISCV_SH3ADDUW => (EXTZ(rs1_val[31..0]) << 3) + rs2_val
+  };
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: Done.

model/riscv_insts_zba.sail
```diff
@@ -0,0 +1,88 @@
...
+  <-> zba_rtype_mnemonic(op) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (ZBA_RTYPE(rs2, rs1, rd, op)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  let result : xlenbits = match op {
+    RISCV_SH1ADD => rs2_val + (rs1_val << 1),
+    RISCV_SH2ADD => rs2_val + (rs1_val << 2),
+    RISCV_SH3ADD => rs2_val + (rs1_val << 3)
+  };
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: Done.

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
bilalsakhawat-10xe: Okay

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+    RISCV_MINU => if (rs1_val <_u rs2_val)
+                    then rs1_val
+                    else rs2_val,
+    RISCV_ROL  => if sizeof(xlen) == 32
+                    then (rs1_val << rs2_val[4..0]) | (rs1_val >> (0x20 - EXTZ(rs2_val[4..0])))
+                    else (rs1_val << rs2_val[5..0]) | (rs1_val >> (0x40 - EXTZ(rs2_val[5..0]))),
+    RISCV_ROR  => if sizeof(xlen) == 32
+                    then (rs1_val >> rs2_val[4..0]) | (rs1_val << (0x20 - EXTZ(rs2_val[4..0])))
+                    else (rs1_val >> rs2_val[5..0]) | (rs1_val << (0x40 - EXTZ(rs2_val[5..0])))
+  };
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: I'll Update for RISCV_ROL and RISCV_ROR, rest was simpler before I think..

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+function clause execute (RISCV_REV8(rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let result : xlenbits = match sizeof(xlen) {
+    32 => ((rs1_val & 0xff000000) >> 24) | ((rs1_val & 0x00ff0000) >>  8) |
+          ((rs1_val & 0x0000ff00) <<  8) | ((rs1_val & 0x000000ff) << 24),
+    64 => ((rs1_val & 0xff00000000000000) >> 56) | ((rs1_val & 0x00ff000000000000) >> 40) |
+          ((rs1_val & 0x0000ff0000000000) >> 24) | ((rs1_val & 0x000000ff00000000) >>  8) |
+          ((rs1_val & 0x00000000ff000000) <<  8) | ((rs1_val & 0x0000000000ff0000) << 24) |
+          ((rs1_val & 0x000000000000ff00) << 40) | ((rs1_val & 0x00000000000000ff) << 56)
+  };
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: I also tried this before. This syntax does not work in sail. It is not able to evaluate [(i+7)..i] like syntax. (Type mismatch). I changed this as suggested by @abukharmeh

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+              then 0x00
+              else 0xFF
+    };
+    result = if (i < (xlen_val - 8))
+               then result << 8
+               else result;
+    rs1_val = if (i < (xlen_val - 8))
+                then rs1_val << 8
+                else rs1_val;
+  };
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: Syntax issue same as above.

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+
+function clause execute (RISCV_CPOP(rs1, rd)) = {
+  let rs1_val = X(rs1);
+  result : xlenbits = EXTZ(0x0);
+  foreach (i from 0 to (xlen_val - 1) by 1) {
+    result = if (rs1_val[i] == bitone)
+               then result + EXTZ(0b1)
+               else result;
+  };
+  X(rd) = result;
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: I just refactor a little my own version.

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+  let rs1_val = X(rs1);
+  result : xlenbits = EXTZ(0x0);
+  done : bits(1) = 0b0;
+  msb : xlenbits = (EXTZ(0b1) << (xlen_val - 1));
+  foreach (i from 0 to (xlen_val - 1) by 1) {
+    if (((rs1_val << i) & msb) != msb) & (done == 0b0)
+      then result = result + EXTZ(0b1)
+      else done = 0b1;
+  };
+  X(rd) = result;
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: Done.

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,365 @@
...
+function clause execute (RISCV_CTZ(rs1, rd)) = {
+  let rs1_val = X(rs1);
+  result : xlenbits = EXTZ(0x0);
+  done : bits(1) = 0b0;
+  foreach (i from 0 to (xlen_val - 1) by 1) {
+    if ((rs1_val >> i)[0] != bitone) & (done == 0b0)
+      then result = result + EXTZ(0b1)
+      else done = 0b1;
+  };
+  X(rd) = result;
```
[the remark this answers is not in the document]
bilalsakhawat-10xe: Done.

model/riscv_insts_zbc.sail
bilalsakhawat-10xe: Done

model/riscv_insts_zbc.sail
bilalsakhawat-10xe: Done.

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
jrtc27: Your version still uses a fixed-width type for result; natural numbers etc are preferred for this kind of thing, it maps more nicely to both the C and theorem-prover backends

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
jrtc27: I deliberately wrote it this way to match how AMOMIN[U] and AMOMAX[U] are implemented, not just because I felt like it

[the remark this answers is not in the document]
jrtc27: Although of course s/min/max for the RISCV_MAX[U] cases, I messed that up

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
jrtc27: It definitely works, I tested it before I suggested it using the following:
```
default Order dec

$include <prelude.sail>

val EXTS : forall 'n 'm, 'm >= 'n. (implicit('m), bits('n)) -> bits('m)
val EXTZ : forall 'n 'm, 'm >= 'n. (implicit('m), bits('n)) -> bits('m)

function EXTS(m, v) = sail_sign_extend(v, m)
function EXTZ(m, v) = sail_zero_extend(v, m)

val zeros_implicit : forall 'n, 'n >= 0 . implicit('n) -> bits('n)
function zeros_implicit (n) = sail_zeros(n)
overload zeros = {zeros_implicit}

val ones : forall 'n, 'n >= 0 . implicit('n) -> bits('n)
function ones (n) = sail_ones (n)

val "decimal_string_of_bits" : forall 'n. bits('n) -> string
val hex_bits_32 : bits(32) <-> string
val hex_bits_32_forwards = "decimal_string_of_bits" : bits(32) -> string
val hex_bits_64 : bits(64) <-> string
val hex_bits_64_forwards = "decimal_string_of_bits" : bits(64) -> string
overload hex_bits = {hex_bits_32, hex_bits_64}

val print = "print_endline" : string -> unit

type xlen : Int = 64
type xlenbits = bits(xlen)

val foo : unit -> unit
function foo() = {
  let rs1_val : xlenbits = EXTZ(0x12345678);
  result : xlenbits = zeros();
  foreach (i from 0 to (sizeof(xlen) - 8) by 8)
    result[(i + 7) .. i] = rs1_val[(sizeof(xlen) - i - 1) .. (sizeof(xlen) - i - 8)];
  print(hex_bits(result))
}

val main : unit -> unit
function main() = {
  foo()
}
```

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
jrtc27: Syntax issue or type error? Because you said type error for the previous comment.

bilalsakhawat-10xe:
@jrtc27 I was doing development on an older branch. That's why some of syntax was not working. I updated every thing you suggested.

jrtc27:
Tagging @kasanovic and @ptomsich

martinberger (approved):

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
allenjbaum: Is it significant that it be done the same way as the AMOs are done? Consistency is nice, but should perhaps the AMOs change instead if the functionality is the same and it's more obvious? Is it because using the intrinsic min/max preferable (though intrinsics maxs/maxu/mins/minu might make this a no-brainer)

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
allenjbaum: That's a subtle point. In general, Is that a good rule-of-thumb when you're doing a reduction kind of op (popcnt, clz, ctz, etc) where the result is log2 size of the source? Or is there a better way to describe when you'd want to do this?

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
jrtc27: I think it's much of a muchness when it comes to which is more obvious; the if/else version is the code people write in languages that lack max/min functions, and the max/min version has max/min and signed/unsigned right there in the implementation. The only ugliness is that using integers rather than bitvectors requires a to_bits on the result.

I agree that min/max_u/s operating on bitvectors would make things better; part of my original thinking that I didn't state was that, by using the min/max version, it's easier to identify the places that can be changed to use min/max_u/s if someone adds those at a later date to clean things up and hide the to_bits messiness as an implementation detail, since you can't instantly determine which ifs are acting as min/max expressions. Having said that, the instruction mnemonics have MIN and MAX in their names so they're still not hard to find, in this specific instance.

model/riscv_insts_zbb.sail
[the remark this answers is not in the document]
jrtc27: I'd say it's specifically the counting part that makes it feel wrong to use bitvectors, as it's not very "pure", it feels very hardware-y. If you were to factor it out as a generic function, the idiomatic interface would be something like `val popcount : forall 'n. bits('n) -> nat` (or `range(0, 'm)` for some properly-defined `'m`, but that gets really awkward to express properly), and then the callers can do what they like with it. This maintains that clear separation between "popcount of a bitvector" and "put that in rd" even though the implementation is inline.

model/riscv_insts_zbb.sail
```diff
@@ -0,0 +1,324 @@
...
+  <-> 0b011000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if haveZbb() & (sizeof(xlen) == 64 | shamt[5] == bitzero)
+
+mapping clause assembly = RISCV_RORI(shamt, rs1, rd)
+  <-> "rori" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ hex_bits_6(shamt)
+
+function clause execute (RISCV_RORI(shamt, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let result : xlenbits = if sizeof(xlen) == 32
+                            then rs1_val >>> shamt[4..0]
+                            else rs1_val >>> shamt;
```
jrtc27, line 30:
```suggestion
                          then rs1_val >>> shamt[4..0]
                          else rs1_val >>> shamt;
```

model/riscv_insts_zbc.sail
```diff
@@ -0,0 +1,59 @@
...
+mapping clause assembly = RISCV_CLMUL(rs2, rs1, rd)
+  <-> "clmul" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (RISCV_CLMUL(rs2, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  result : xlenbits = zeros();
+  foreach (i from 0 to (xlen_val - 1)) {
+    if rs2_val[i] == bitone then result = result ^ (rs1_val << i);
+  };
```
jrtc27, line 16: This doesn't need curly braces

model/riscv_insts_zbc.sail
```diff
@@ -0,0 +1,59 @@
...
+mapping clause assembly = RISCV_CLMULH(rs2, rs1, rd)
+  <-> "clmulh" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (RISCV_CLMULH(rs2, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  result : xlenbits = zeros();
+  foreach (i from 0 to (xlen_val - 1)) {
+    if rs2_val[i] == bitone then result = result ^ (rs1_val >> (xlen_val - i));
+  };
```
jrtc27, line 36: Ditto

model/riscv_insts_zbc.sail
```diff
@@ -0,0 +1,59 @@
...
+mapping clause assembly = RISCV_CLMULR(rs2, rs1, rd)
+  <-> "clmulr" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (RISCV_CLMULR(rs2, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  result : xlenbits = zeros();
+  foreach (i from 0 to (xlen_val - 1)) {
+    if rs2_val[i] == bitone then result = result ^ (rs1_val >> (xlen_val - i - 1));
+  };
```
jrtc27, line 56: Ditto

jrtc27 (commented):
Just a few tiny nits but otherwise look good, thanks, so happy for this to be landed once those are resolved

jrtc27:
That is, so long as the history is squashed, either by you or whoever merges it (one commit for the family is fine; would probably slightly prefer one for each of Zb[abcs] but it's not worth the effort to untangle them at this point)

allenjbaum:
I'm not sure I understand that last comment; you'd prefer one each of what?

jrtc27:
> I'm not sure I understand that last comment; you'd prefer one each of what?
> […](#)

One commit per extension being added is better than all at once, but it's not a huge deal given the size of the the extensions involved and that they're all closely related. We took scalar crypto all as one commit and that's rather larger. Just a suggestion for any future work, and is one of the things suggested in the contributing guide in #131.

bilalsakhawat-10xe:
@jrtc27 I am done with the tiny nits.

jrtc27:
Thanks

PR #121, opened 2021-11-03
Scalar crypto - post code review improvments

ben-marshall:
Hi there

This PR contains updates from @jrtc27's review [here](https://github.com/riscv/sail-riscv/pull/99#pullrequestreview-786873720) which were missing from the original merge of the scalar cryptography extension code.

- [Commit 1](https://github.com/riscv/sail-riscv/commit/a527ad769c9acba9cba70261c3bd73a1d5a0069c) addresses the [inconsistency of `==` whitespace](https://github.com/riscv/sail-riscv/pull/99#discussion_r734560390)
- [Commit 2](https://github.com/riscv/sail-riscv/commit/c4ad3c16a943de9996224a2439b5488ce4fac359) addresses the [inconsistency of ` );`](https://github.com/riscv/sail-riscv/pull/99#discussion_r734563951) whitespace in the `sha*` instructions.
- [Commit 3](https://github.com/riscv/sail-riscv/commit/53611ff431bac23b438f30bbd214cff3c55c6bb6) tries to address [comments](https://github.com/riscv/sail-riscv/pull/99#discussion_r734584700) about the clarity of `aesks1i` by adding code comments, and ensuring that the `aes_decode_rcon` function only accepts valid inputs. It also renames some arguments to the `aesks1i` execute and encdec clauses to match the specification.

About [this question](https://github.com/riscv/sail-riscv/pull/99#discussion_r734562267) asking "please confirm every single change in [commit](https://github.com/riscv/sail-riscv/compare/33aaaa04eaeaa9d2ed69023735ad6ebb93f71775..b83f68b3ade2d5444fd7dfe952e441824bb19dd8) is what you intended?" - Yes, these functional changes were intended and should have been part of the initial PR. Specifically, the changes to when `aesks1i` and `aesks2` are enabled were deliberate, as they should be present for both the `Zkne` and `Zknd` extensions.

About the `aesks1i` comments [here](https://github.com/riscv/sail-riscv/pull/99#discussion_r734584700) - It's been interesting watching people react to these instructions. The more cryptography minded people have reacted quite positively to how they're specified because they map onto some common implementation patterns for the AES KeySchedule _fairly_ well. But, that's only apparent if you're very familiar with those patterns in the first place, which is precious few people. Others have initially been quite perplexed about them for exactly the reasons highlighted. If there's anything _more_ to be done about explaining these instructions better then I'm all ears. Whether this is best going into the Sail code or the specification document, I'm not sure - maybe both. In terms of being future-proof to changes in FIPS-197 parameters, I think we are fairly safe. AES is very stable at this point, and I don't think NIST or anyone in the cryptography community is expecting to see AES parameter changes.

I've deliberately kept the commits separate in the hope it makes the PR easier to map back onto the original review. If you'd like me to squash them down before merging, please let me know. Likewise if there's anything else which could be made clearer.

Cheers,
Ben

### DIFF
--- a/model/riscv_types_kext.sail
+++ b/model/riscv_types_kext.sail
@@ -68,8 +68,15 @@ function aes_mixcolumn_inv(x) = {
   b3 @ b2 @ b1 @ b0 /* Return value */
 }
 
-val aes_decode_rcon : bits(4) -> bits(32)
+/* Turn a round number into a round constant for AES. Note that the
+   AES64KS1I instruction is defined such that the r argument is always
+   in the range 0x0..0xA. Values of rnum outside the range 0x0..0xA
+   do not decode to the AES64KS1I instruction. The 0xA case is used
+   specifically for the AES-256 KeySchedule, and this function is never
+   called in that case. */
+val aes_decode_rcon : bits(4) -> bits(32) effect {escape}
 function aes_decode_rcon(r) = {
+  assert(r <_u 0xA);
   match r {
     0x0 => 0x00000001,
     0x1 => 0x00000002,
@@ -80,13 +87,7 @@ function aes_decode_rcon(r) = {
     0x6 => 0x00000040,
     0x7 => 0x00000080,
     0x8 => 0x0000001b,
-    0x9 => 0x00000036,
-    0xA => 0x00000000,
-    0xB => 0x00000000,
-    0xC => 0x00000000,
-    0xD => 0x00000000,
-    0xE => 0x00000000,
-    0xF => 0x00000000
+    0x9 => 0x00000036
   }
 }
 
--- a/model/riscv_insts_zkn.sail
+++ b/model/riscv_insts_zkn.sail
@@ -190,37 +190,37 @@ mapping clause assembly = SHA512SUM1R (rs2, rs1, rd)
 
 function clause execute (SHA512SIG0H(rs2, rs1, rd)) = {
   X(rd) = EXTS((X(rs1) >>  1) ^ (X(rs1) >>  7) ^ (X(rs1) >>  8) ^
-               (X(rs2) << 31)                  ^ (X(rs2) << 24) );
+               (X(rs2) << 31)                  ^ (X(rs2) << 24));
   RETIRE_SUCCESS
 }
 
 function clause execute (SHA512SIG0L(rs2, rs1, rd)) = {
   X(rd) = EXTS((X(rs1) >>  1) ^ (X(rs1) >>  7) ^ (X(rs1) >>  8) ^
-               (X(rs2) << 31) ^ (X(rs2) << 25) ^ (X(rs2) << 24) );
+               (X(rs2) << 31) ^ (X(rs2) << 25) ^ (X(rs2) << 24));
   RETIRE_SUCCESS
 }
 
 function clause execute (SHA512SIG1H(rs2, rs1, rd)) = {
   X(rd) = EXTS((X(rs1) <<  3) ^ (X(rs1) >>  6) ^ (X(rs1) >> 19) ^
-               (X(rs2) >> 29)                  ^ (X(rs2) << 13) );
+               (X(rs2) >> 29)                  ^ (X(rs2) << 13));
   RETIRE_SUCCESS
 }
 
 function clause execute (SHA512SIG1L(rs2, rs1, rd)) = {
   X(rd) = EXTS((X(rs1) <<  3) ^ (X(rs1) >>  6) ^ (X(rs1) >> 19) ^
-               (X(rs2) >> 29) ^ (X(rs2) << 26) ^ (X(rs2) << 13) );
+               (X(rs2) >> 29) ^ (X(rs2) << 26) ^ (X(rs2) << 13));
   RETIRE_SUCCESS
 }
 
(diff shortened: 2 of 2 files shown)


jrtc27:
Do you have an opinion on:
> I'd probably rather that be specified in a more obvious manner, with aes_decode_rcon only accepting legal inputs instead of silently mapping reserved ones to 0, and AES64KS1I having a very clear special case for 0xA, rather than trying to shoe-horn it into the normal case.

and

> That is, I think this would be a lot easier to understand (and easier to know how to use; I suspect people who know about AES-256 but not Zkn will initially be confused like I was and think it's broken for AES-256 like I did, but maybe that's just me) as something more like:
> ```
> function clause execute (AES64KS1I(rnum, rs1, rd)) = {
>   assert(sizeof(xlen) == 64);
>   let prev     : bits(32) = X(rs1)[63..32];
>   let subwords : bits(32) = aes_subword_fwd(prev);
>   let result   : bits(32) = if (rnum == 0xA) then subwords
>                             else (subwords >>> 8) ^ aes_decode_rcon(rnum);
>   X(rd) = result @ result;
>   RETIRE_SUCCESS
> }
> ```
> I still don't like the duplication of state in rd given the low half is entirely unused (at which point, really, the instructions should just be producing the output in the low 32 bits and zeroes in the others), which I assume is just left over from AES-NI inspiration, but that is probably too big a change at this stage of the spec, unfortunately, so we have to live with that wart (alongside the 0xA sentinel precluding future extension).
> 
> I also took the liberty of renaming some things to what I believe are better names in that example (and rcon -> rnum to match the spec change that came out of one of our previous discussions).

I think that, combined with a brief single sentence note calling out the special 0xA case would go a long way to make things clear for those not intimately familiar with AES.

ben-marshall:
I agree that adding a sentence to explicitly say the `0xA` case is particular for AES-256 is a helpful pointer for people.

In terms of restructuring the `aesks1i` code. I don't have a strong opinion, but I am happy being guided by your experience in trying to understand how the instruction works. If the code you provided is clearer to you (and we assume someone familiar with AES-256 would get it anyway) then that's the better way to write it. I'll make these changes in a commit shortly.

I'm sorry I missed this suggestion initially, Github has an annoying habit of collapsing review comments & threads.

model/riscv_types_kext.sail
```diff
@@ -68,8 +68,14 @@ function aes_mixcolumn_inv(x) = {
   b3 @ b2 @ b1 @ b0 /* Return value */
 }
 
-val aes_decode_rcon : bits(4) -> bits(32)
+/* Turn a round number into a round constant for AES. Note that the
+   AES64KS1I instruction is defined such that the r argument is always
+   in the range 0x0..0xA. Values of rnum outside this range do not decode
+   to the AES64KS1I instruction. The 0xA => 0x0 case is used specifically
```
jrtc27, line 74: This special case should no longer be hit thanks to the revised AES64KS1I implementation, so you can have an assert that it's less than 0xA

martinberger (approved):
OK with me.

model/riscv_types_kext.sail
ben-marshall: Ah right you are! Duly change in the most recent push. :+1:

model/riscv_types_kext.sail
```diff
@@ -68,8 +68,15 @@ function aes_mixcolumn_inv(x) = {
   b3 @ b2 @ b1 @ b0 /* Return value */
 }
 
-val aes_decode_rcon : bits(4) -> bits(32)
+/* Turn a round number into a round constant for AES. Note that the
+   AES64KS1I instruction is defined such that the r argument is always
+   in the range 0x0..0xB. Values of rnum outside the range 0x0..0xA
+   do not decode to the AES64KS1I instruction. The 0xA => 0x0 case is used
```
jrtc27, line 74: > The 0xA => 0x0 case is used

There is no 0xA => 0x0 case, the function is just never called and no XOR is ever performed.

model/riscv_types_kext.sail
```diff
@@ -68,8 +68,15 @@ function aes_mixcolumn_inv(x) = {
   b3 @ b2 @ b1 @ b0 /* Return value */
 }
 
-val aes_decode_rcon : bits(4) -> bits(32)
+/* Turn a round number into a round constant for AES. Note that the
+   AES64KS1I instruction is defined such that the r argument is always
+   in the range 0x0..0xB. Values of rnum outside the range 0x0..0xA
```
jrtc27, line 73: The first one should be 0xA not 0xB?

model/riscv_types_kext.sail
jrtc27: (i.e. just drop the => 0 and I think the sentence reads fine?)

model/riscv_types_kext.sail
ben-marshall: Doh. Right you are

model/riscv_types_kext.sail
ben-marshall: Agreed, just fixed this now.

jrtc27 (approved):
Thanks, looks good to me now

martinberger:
> Thanks, looks good to me now

Thanks!

PR #122, opened 2021-11-03
Add Support for Zmmul

bilalsakhawat-10xe:
RISC-V Zmmul extension includes following instructions:
 - mul
 - mulh
 - mulhsu
 - mulhu
 - mulw (RV64 only)
The opcode encodings and instruction semantics are identical to those in the M-extension.
The ISA string is Zmmul.
8.3 Zmmul Extension, Version 0.1 (riscv unpriv spec)

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -190,6 +190,8 @@ function haveZknh()   -> bool = true
 function haveZkne()   -> bool = true
 function haveZknd()   -> bool = true
 
+function haveZmmul()  -> bool = true
+
 bitfield Mstatush : bits(32) = {
   MBE  : 5,
   SBE  : 4
--- a/model/riscv_insts_mext.sail
+++ b/model/riscv_insts_mext.sail
@@ -85,7 +85,7 @@ mapping clause encdec = MUL(rs2, rs1, rd, high, signed1, signed2)
   <-> 0b0000001 @ rs2 @ rs1 @ encdec_mul_op(high, signed1, signed2) : bits(3) @ rd @ 0b0110011
 
 function clause execute (MUL(rs2, rs1, rd, high, signed1, signed2)) = {
-  if haveMulDiv() then {
+  if haveMulDiv() | haveZmmul() then {
     let rs1_val = X(rs1);
     let rs2_val = X(rs2);
     let rs1_int : int = if signed1 then signed(rs1_val) else unsigned(rs1_val);
@@ -177,7 +177,7 @@ mapping clause encdec = MULW(rs2, rs1, rd)
       if sizeof(xlen) == 64
 
 function clause execute (MULW(rs2, rs1, rd)) = {
-  if haveMulDiv() then {
+  if haveMulDiv() | haveZmmul() then {
     let rs1_val = X(rs1)[31..0];
     let rs2_val = X(rs2)[31..0];
     let rs1_int : int = signed(rs1_val);


scottj97:
Should the new `if haveMulDiv()` qualifiers have been there all along? If so, can you make a separate commit with only that change, then a second commit to add Zmmul?

bilalsakhawat-10xe:
> Should the new `if haveMulDiv()` qualifiers have been there all along? If so, can you make a separate commit with only that change, then a second commit to add Zmmul?

Yes! `haveMulDiv()` qualifier has been there. Sure, I'll make two separate commits.

bilalsakhawat-10xe:
@scottj97 Actually `haveMulDiv()` checks were already there. My bad, I added some extra checks. Now I have removed those extra checks.

scottj97:
It's nice that this is much simplified, but please squash the two commits back into one.

bilalsakhawat-10xe:
When it will merge it will squash merge into the master i.e., will go as one commit. Do I need to squash merge anyway ?

scottj97:
You're assuming the admin who clicks the Merge button chooses the Squash option. I don't think that's a safe assumption.

jrtc27:
Does M not imply Zmmul, and thus only the latter needs checking?

model/riscv_sys_regs.sail
```diff
@@ -180,6 +180,8 @@ function haveUsrMode() -> bool = misa.U() == 0b1
 function haveNExt()    -> bool = misa.N() == 0b1
 /* see below for F and D extension tests */
 
+function haveZmmul()    -> bool = true
```
jrtc27, line 183: As with the bitmanip PR, keep the lists of dummy Z* extension checks sorted, i.e put Zm* below Zk*

bilalsakhawat-10xe: > Does M not imply Zmmul, and thus only the latter needs checking?

Yes, M does not imply Zmmul. You can read this discussion on this [PR](https://github.com/riscv-software-src/riscv-isa-sim/pull/830)

bilalsakhawat-10xe: > As with the bitmanip PR, keep the lists of dummy Z* extension checks sorted, i.e put Zm* below Zk*

Sure :)

bilalsakhawat-10xe:
> You're assuming the admin who clicks the Merge button chooses the Squash option. I don't think that's a safe assumption.

Sure. I'll squash commits.

model/riscv_sys_regs.sail
jrtc27: > > Does M not imply Zmmul, and thus only the latter needs checking?
> 
> Yes, M does not imply Zmmul. You can read the discussion on this [PR](https://github.com/riscv-software-src/riscv-isa-sim/pull/830)

This contradicts https://github.com/riscv/riscv-isa-manual/pull/648#issuecomment-842461775, which is as close to an authoritative source as I can find. Pointing at your own comments on your own Spike pull request is hardly authoritative.

bilalsakhawat-10xe: Let's put it in this way. An implementation can have both extensions as a time. When I said does not imply I meant to say If misa.M() ==1 it does not mean that Zmmul extension is there as well. if Either or both of them are in implementation the Mul instructions will work. That's why there is a `or` between haveMulDiv and haveZmmul.

jrtc27: That doesn't make much sense when Zmmul is defined as a subset of M...

bilalsakhawat-10xe: As far as I understand. The benefit of Zmmul is, as it supports only MUL instructions (Not DIV/REM), it will save hardware in those implementations where DIV/REM instructions are not needed.

scottj97: One difference between M and Zmmul is that the latter is not controlled by misa.M. If you have Zmmul then the multiply instructions are always enabled. If you have M then the multiply instructions are enabled iff misa.M==1. That's why saying "M implies Zmmul" is inaccurate -- because that would mean that misa.M only controls divide and not multiply.

This Sail model doesn't support modifying misa.M so this is moot here -- but that might change in the future so I think it's better to keep them separate here.

jrtc27: Ah indeed that is an important but subtle difference. I believe modifying misa.M should work if sys_enable_writable_misa is true.

scottj97:
@bilalsakhawat-10xe can you please fix the Git commit message. It doesn't make sense to include "Removed extra haveMulDiv() checks" because it doesn't do that anymore.

model/riscv_sys_regs.sail
scottj97: It looks like [this code](https://github.com/riscv/sail-riscv/blob/4508059032ecfdb986a5673108b02900d2c4ccfd/model/riscv_sys_regs.sail#L157-L168) supports modifying bits C/D/F only. And reading it, I see a bug, which I'll file separately.

jrtc27: Oh, right, I skimmed it and thought it was mostly fixing up v, but it's just using v to fix up the existing m.

jrtc27:
Please amend your commit message to include a verb ("Add support for Zmmul", "Implement Zmmul", etc., don't really care exactly what), it's good practice (otherwise it doesn't actually mean anything as it leaves one asking what are you doing to the zmmul support? adding it? removing it? refactoring it? all possibilities without looking at the changes themselves). Otherwise this looks ready.

jrtc27:
This was fine until you pushed that merge commit. Never merge into a pull request branch, always rebase the branch. But in this case there was no need to do either, the changes to master don't affect this PR whatsoever.

bilalsakhawat-10xe:
> This was fine until you pushed that merge commit. Never merge into a pull request branch, always rebase the branch. But in this case there was no need to do either, the changes to master don't affect this PR whatsoever.

Should I rebase it?

bilalsakhawat-10xe:
> > This was fine until you pushed that merge commit. Never merge into a pull request branch, always rebase the branch. But in this case there was no need to do either, the changes to master don't affect this PR whatsoever.
> 
> Should I rebase it now?

jrtc27:
Yes, please rebase in order to lose the merge commit

jrtc27:
(same goes for your bitmanip, and now scalar crypto, PRs)

bilalsakhawat-10xe:
> Yes, please rebase in order to lose the merge commit

Steps I have in mind:
- revert the merge commit
- rebase with master
- squash above two into one i.e., rebase with master 

Is there a good way to do it?

jrtc27:
> > Yes, please rebase in order to lose the merge commit
> 
> Steps I have in mind:
> 
> * revert the merge commit
> * rebase with master
> * squash above two into one i.e., rebase with master
> 
> Is there a good way to do it?

Just `git rebase master`, should just work

jrtc27:
(or `git rebase origin/master` if your local master branch is behind)

scottj97:
I wonder if it's a good idea to haveZmmul==true as the default. We also haveMulDiv==true as the default, and it doesn't really make sense to have both of these at once. (As discussed above w.r.t. misa.M.) It seems the configuration system that the Arch Test SIG is working on should have a configuration knob with three choices: None, M, or Zmmul.

bilalsakhawat-10xe:
> I wonder if it's a good idea to haveZmmul==true as the default. We also haveMulDiv==true as the default, and it doesn't really make sense to have both of these at once. (As discussed above w.r.t. misa.M.) It seems the configuration system that the Arch Test SIG is working on should have a configuration knob with three choices: None, M, or Zmmul.

@jrtc27 you may have a proper answer, why we are using haveZmmul==true as default for now.

bilalsakhawat-10xe:
> > I wonder if it's a good idea to haveZmmul==true as the default. We also haveMulDiv==true as the default, and it doesn't really make sense to have both of these at once. (As discussed above w.r.t. misa.M.) It seems the configuration system that the Arch Test SIG is working on should have a configuration knob with three choices: None, M, or Zmmul.
> 
> @jrtc27 you may have a proper answer, why we are using haveZmmul==true as default for now.

@jrtc27 an input from you will be great.

allenjbaum:
I originally defined it such that Zmmul and M were incompatible, but I was told there weren't. Zmmul is redundant with M, but not disallowed.  So, stupid, but not forbidden.

jrtc27:
> I originally defined it such that Zmmul and M were incompatible, but I was told there weren't. Zmmul is redundant with M, but not disallowed. So, stupid, but not forbidden.

Ok, thanks for the clarification.

Given we have all the Zb* and Zk* extensions on by default with no way to turn them off (currently), it does seem a bit odd to not provide Zmmul. I know it means setting misa.M to 0 no longer turns off MUL*, but does that actually matter? I don't personally mind either way all that much, just so far the philosophy's been to enable as much as we can by default for maximum utility.

allenjbaum:
From the point of view of architectural tests: the fact that either are on doesn't matter.
If M-extension is unimplemented, we won't ever test DIV/REM ops, so it's irrelevant whether they actually correctly execute those ops
Similarly, if both M-extension and Zmmul are unimplemented, we won't test multiply instructions.

scottj97:
> From the point of view of architectural tests: the fact that either are on doesn't matter.

The `misa` CSR is part of the architectural spec too. Don't you want to test that it works properly, disabling or enabling instructions as required?

I don't care what the default is for Zmmul in this Sail model.

allenjbaum:
If MISA.M is writable, we could, sort of, but it is a bit pointless. t boot, if M-ext is implemented, MISA.M will be set, and we will test that it works. IF MISA.M is writable, we can test that it can be cleared, and set again, but that is all. 
If it is cleared, it is still legal for Mul ops to perform a multiply. On the other hand, if Zmmul and M-ext are both implemented, we could text that Mul ops still work with MISA.M cleared, because Zmmul can't be disabled.
 I hadn't thought of that.
 In fact, we probably should run all Zmmul tests that way.  I need to figure out how to do that without having to re-write all the M tests, since we are just re-using them, running if either M-ext  or Zmmul are implemented.

bilalsakhawat-10xe:
I know a way how we can run all Zmmul tests without having to re-write all the M tests. We just need to check if Zmmul is in the ISA string and if yes include mul, mulh, mulhsu, mulhu and mulw in the database.yaml (it contains the list of tests to be run on Ref and DUT). We just need to change 1-2 files in riscof env.

allenjbaum:
I don't think it's that simple. We potentially have to run the tests twice; once if M-extension is supported (and MISA.M is set, which should be the case after reset. Hmm, our tests probably need to check that) and once if Zmmul is supported and either MISA.M is clear or M-ext isn't supported). I think that means those tests need to be replicated. 
I thought we could avoid that, but we can't.

bilalsakhawat-10xe:
> I don't think it's that simple. We potentially have to run the tests twice; once if M-extension is supported (and MISA.M is set, which should be the case after reset. Hmm, our tests probably need to check that) and once if Zmmul is supported and either MISA.M is clear or M-ext isn't supported). I think that means those tests need to be replicated. I thought we could avoid that, but we can't.

Running test twice is not an issue. We can do that for misa.M and zmmul. The issue will be when collecting coverage, we need to make changes in riscv-ctg as well. Even this is a simple task. But this is not a recommended solution to make changes riscof and riscv-ctg for just zmmul. Replicating the tests is the best solution.

allenjbaum:
I talked to Krste. He is suggesting that if an implemention does have both
Zmmul AND M-extension AND misa.M is RW, then vendor is responsible for
running the tests twice with different boot code, the normal one and one
that clears misa.M before running tests. It’s unlikely that anyone will
ever build this specific configuration (why would you ever disable M?) so
making the umplementir do the extra work is fine.

allenjbaum:
I think I've discussed this elsewhere, but the corner case only occurs when there is an implementation that has both M-extension and Zmmul, AND misa.M is read-write, so the M-extension can be disabled (which effectively disables only DIV/REM ops, since Zmmul can't be disabled). 
This is an unlikely enough to be implemented that instead of burdening our test infrastructure with handling this, we will will require the vendor to run tests twice: once normally and once with boot code augmented to clear misa.M. The latter only needs to execute M-extension tests. Both m-extension tests reports should be submitted.

bilalsakhawat-10xe:
So I think, the changes I made in this PR are good enough for now. Later we may add a plusargs in sail '--isa=RVxxx' like we have in spike. Then for 64-bit we can have the cases like:
- RV64IM                                                                        (misa.M == true, Zmmul == false)
- RV64IM_Zmmul                                                           (misa.M == true, Zmmul == true)
- RV64I_Zmmul                                                              (misa.M == false, Zmmul == true)
- And a case where we have neither for them enable    (misa.M == false, Zmmul == false)
depending on the plusargs '--isa' string.

jrtc27:
Well, it's not a plusarg, those are specifically `+foo` and are so called due to the leading `+`, and are pretty specific to the hardware space (and thus spike's fesvr/htif/whatever stuff..). But yes, eventually all the Zfoo should be dynamically configurable via a command-line argument, ideally as an arch string that matches what toolchains use.

jrtc27:
So, if nobody has any strong opinions on this, I suggest we go ahead and merge it with Zmmul defaulting to enabled, given so far we've only disabled extensions by default if they conflict?

bilalsakhawat-10xe:
> Well, it's not a plusarg, those are specifically `+foo` and are so called due to the leading `+`, and are pretty specific to the hardware space (and thus spike's fesvr/htif/whatever stuff..). But yes, eventually all the Zfoo should be dynamically configurable via a command-line argument, ideally as an arch string that matches what toolchains use.

Thanks for correction. I choose the wrong terms actually, I should have said command-line argument.

allenjbaum:
From the point of view of ACTs - it is our intention that we don't execute tests for extensions that don't have support (or are disabled) because results are unpredictable. At reset, all extensions that implemented are enabled, regardless of whether it is possible to disable them. So, defaulting them to enabled seems just fine.

martinberger:
@scottj97 @jrtc27 @allenjbaum @bilalsakhawat-10xe I would like to merge this PR this week. Any objections to that?

allenjbaum:
not from me!

jrtc27:
Nope, long overdue

scottj97 (approved):

martinberger (approved):

martinberger:
Merged.

Thanks @bilalsakhawat-10xe @scottj97 @jrtc27 @allenjbaum

PR #125, opened 2021-11-05
Add support for Scalar Cryptography Zbkb, Zbkc and Zbkx extensions

bilalsakhawat-10xe:
This request is for Scalar Cryptography Zbkb, Zbkc and Zbkx extensions Support (https://github.com/riscv/riscv-crypto/releases/tag/v1.0.0-rc5-scalar) in the Model.

This Pull Request can be merged after #116.

Kindly review this request. Thanks!


model/riscv_insts_zbkb.sail
```diff
@@ -0,0 +1,104 @@
...
+
+mapping clause assembly = RISCV_BREV8(rs1, rd)
+  <-> "brev8" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1)
+
+function clause execute (RISCV_BREV8(rs1, rd)) = {
+  let rs1_val = X(rs1);
+  result : xlenbits = zeros();
+  foreach (i from 0 to (sizeof(xlen) - 8) by 8)
+    foreach (j from i to (i + 7))
+      result[j] = rs1_val[(i + 7) - j];
```
jrtc27, line 101: Doesn't this produce 4 or 8 (depending on XLEN) copies of the reversed rs1_val[7..0]?

This also would probably be better by defining a `val bitvector_reverse : forall 'n. bits('n) -> bits('n)` and `overload reverse = {bitvector_reverse}` (split that way in case anyone wants to add a vector_reverse for something else; it's done like that in the standard library) that you can then use here rather than needing an inner loop.

model/riscv_insts_zbkb.sail
```diff
@@ -0,0 +1,104 @@
...
+mapping clause assembly = ZBKB_RTYPE(rs2, rs1, rd, op)
+  <-> zbkb_rtype_mnemonic(op) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (ZBKB_RTYPE(rs2, rs1, rd, op)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  let result : xlenbits = match op {
+    RISCV_PACK  => if sizeof(xlen) == 32
+                   then rs2_val[15..0] @ rs1_val[15..0]
+                   else rs2_val[31..0] @ rs1_val[31..0],
```
jrtc27, line 24: Just use `sizeof(xlen)/2 - 1` for the upper bound, no if needed

model/riscv_insts_zbkb.sail
```diff
@@ -0,0 +1,104 @@
...
+
+mapping clause encdec = RISCV_ZIP(rs1, rd) if haveZbkb() & sizeof(xlen) == 32
+  <-> 0b000010001111 @ rs1 @ 0b001 @ rd @ 0b0010011 if haveZbkb() & sizeof(xlen) == 32
+
+mapping clause assembly = RISCV_ZIP(rs1, rd)
+  <-> "zip" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1)
+
+function clause execute (RISCV_ZIP(rs1, rd)) = {
+  let rs1_val = X(rs1);
+  result : xlenbits = zeros();
```
jrtc27, line 59: Either consistently hard-code that it's RV32 or write fully general code that happens to only run on RV32 (i.e., either change xlenbits to bits(32) here or change the 15 and 16 below to be `sizeof(xlen)/2` and that `- 1`). If doing the former you will want an `assert(sizeof(xlen) == 32);` at the start of this function to inform the type checker it's dead code on RV64.

For what it's worth, the pseudo-Sail in the the crypto spec takes the "fully general code that happens to only run on RV32" approach, but I think both are ok, doesn't matter unless there's a reason to believe a future extension might want (un)zip on RV64.

model/riscv_insts_zbkb.sail
```diff
@@ -0,0 +1,104 @@
...
+
+mapping clause encdec = RISCV_UNZIP(rs1, rd) if haveZbkb() & sizeof(xlen) == 32
+  <-> 0b000010001111 @ rs1 @ 0b101 @ rd @ 0b0010011 if haveZbkb() & sizeof(xlen) == 32
+
+mapping clause assembly = RISCV_UNZIP(rs1, rd)
+  <-> "unzip" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1)
+
+function clause execute (RISCV_UNZIP(rs1, rd)) = {
+  let rs1_val = X(rs1);
+  result : xlenbits = zeros();
```
jrtc27, line 79: Same as zip

Makefile
```diff
@@ -26,6 +26,7 @@ SAIL_DEFAULT_INST += riscv_insts_dext.sail riscv_insts_cdext.sail
 SAIL_DEFAULT_INST += riscv_insts_zba.sail
 SAIL_DEFAULT_INST += riscv_insts_zbb.sail
 SAIL_DEFAULT_INST += riscv_insts_zbc.sail
+SAIL_DEFAULT_INST += riscv_insts_zbkb.sail
```
jrtc27, line 29: Maybe separate Zbk* out from Zb* like you've done in riscv_sys_regs? It violates the alphabetical ordering, but I think it makes more sense? Not sure, but should probably at least be consistent one way or the other.

model/riscv_insts_zbkx.sail
```diff
@@ -0,0 +1,39 @@
...
+
+mapping clause assembly = RISCV_XPERM8(rs2, rs1, rd)
+  <-> "xperm8" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (RISCV_XPERM8(rs2, rs1, rd)) = {
+  let rs1_val = X(rs1);
+  let rs2_val = X(rs2);
+  result : xlenbits = zeros();
+  foreach (i from 0 to (sizeof(xlen) - 8) by 8) {
+    result[i+7..i] = (rs1_val >> (rs2_val[i+7..i] @ 0b000)[7..0])[7..0];
```
jrtc27, line 15: For a model these would be better written as the
```
let index : bits(8) = rs2_val[i+7..i];
result[i+7...i] = if index < sizeof(xlen)/8
                  then rs1_val[8*index+7..8*index]
                  else zeros()
```
and equivalent for xperm4 versions, much easier to understand

model/riscv_sys_regs.sail
```diff
@@ -186,8 +186,9 @@ function haveZbb()  -> bool = true
 function haveZbc()  -> bool = true
 function haveZbs()  -> bool = true
 
-/* BitManip Cryptography extension support. */
+/* BitManip Cryptography extensions support. */
 function haveZbkb() -> bool = true
+function haveZbkc() -> bool = true
 
 /* Cryptography Crossbar permutation instructions support. */
```
jrtc27, line 193: Why does Zbkx get its own header but Zbkc doesn't? I'd be inclined to just have the one header for all three, we don't break down the exact meaning of Zb[abcs] and Zk* here

model/riscv_insts_zbkb.sail
bilalsakhawat-10xe: Yes it does produce 4 or 8 (depending on XLEN) copies of the reversed rs1_val[7..0].

Sure, I'll update by using bitvector_reverse.

model/riscv_insts_zbkb.sail
bilalsakhawat-10xe: Okay... I'll change this.

model/riscv_insts_zbkb.sail
bilalsakhawat-10xe: sure.

model/riscv_insts_zbkb.sail
bilalsakhawat-10xe: okay

Makefile
bilalsakhawat-10xe: Sure.

model/riscv_insts_zbkb.sail
allenjbaum: To be clear: this should *not* make multiple copies of the same byte; 
it should be reversing each byte individually, e.g.: 
  foreach (j from 0 to  7)
      result[i + j] = rs1_val[i + (7 - j)];
but using a bitvector reverse  function is far more readable, and can be used elsewhere, as pointed out,  in  which case
   foreach (i from 0 to (sizeof(xlen) - 8) by 8)
     result[i..i + 7] = bitreverse[rs1_val[i..i + 7];
(I won't pretend this syntax is exactly correct]

model/riscv_insts_zbkb.sail
allenjbaum: I'm allergic to using /2 vs  >>1; should I not care, or is this an integer vs. bitvector thing and we should be using /2?

model/riscv_insts_zbkx.sail
allenjbaum: Same question for my information: better to use <<3 or *8 ?

model/riscv_insts_zbkb.sail
jrtc27: `>>` isn't defined for integers currently in model/prelude.sail, only for bitvectors, though there's no reason you couldn't do so I guess. But really `/` should be used for things that are logically a division, and `>>` for things that are logically a bit shift (ignoring the fact that the two are equivalent in certain cases), with the compiler left to optimise division to a bit shift as necessary. In this case it'll all be constant-folded anyway as it's a constant divided by a constant.

Also, /2 and >>1 are not the same thing on negative values, they round in different directions.

jrtc27: (i.e. embrace your inner mathematician rather than your inner hardware engineer, we're not in the grotty world of verilog, we're in a nice high-level language that's trying to specify things in as clear and pure a way as we can)

model/riscv_insts_zbkb.sail
jrtc27: Yep, exactly

bilalsakhawat-10xe: @allenjbaum The function I wrote with nested for loops is working perfectly, reversing each byte individually. Yeah but I'll implement bitvector reverse function as suggested by Jessica. 
@jrtc27 A quick question, should I add bitvector reverse function in preludes.sail or riscv_insts_zbkb.sail or it doesn't make any difference where I add the function (overloading)?

model/riscv_insts_zbkx.sail
bilalsakhawat-10xe: > let index : bits(8) = rs2_val[i+7..i];
> result[i+7...i] = if index < sizeof(xlen)/8
>                   then rs1_val[8*index+7..8*index]
>                   else zeros()

overloading for (operator <), can not unify int('n) and bitvector(8, dec)        ( if condition)

model/riscv_insts_zbkb.sail
jrtc27: Prelude, it's a general utility function that should exist in the standard library but doesn't

model/riscv_insts_zbkx.sail
jrtc27: Yeah ok `unsigned(index) < sizeof(xlen)/8`

bilalsakhawat-10xe: No val spec found for (operator /)

bilalsakhawat-10xe: I used 
    index : int = unsigned(rs2_val[i+3..i]);
    result[i+3..i] = if index < quot_round_zero(sizeof(xlen), 8)
                      then rs1_val[8*index+3..8*index]
                      else zeros()
And have following issue:
image

jrtc27: Ok if we don't have division then just do `8 * index < sizeof(xlen)`, that'll probably appease the range checker too

jrtc27: Yeah, ```    let index = unsigned(rs2_val[i+7..i]);
    result[i+7..i] = if 8*index < sizeof(xlen)
                     then rs1_val[8*index+7..8*index]
                     else zeros()
``` works

jrtc27: (And do make sure to use let, without it you have a mutable variable, which should only be done when you actually need the mutability)

bilalsakhawat-10xe: Still the issue in above issue not resolved.

bilalsakhawat-10xe: Resolved! Thanks

model/riscv_insts_zbkb.sail
bilalsakhawat-10xe: Since division is not available I'll hard-code xlenbits to bits(32)

model/riscv_insts_zbkb.sail
bilalsakhawat-10xe: Same division issue will be there while using `sizeof(xlen)/2 - 1` as upper bound

model/riscv_insts_zbkb.sail
jrtc27: There's also `sizeof(xlen_bytes) * 4` as an option

allenjbaum:
More useful, and much simpler:
The grev function that this was originally based on had a parameter P with
log2XLEN bits. It simply mapped result[X] = rs1[X^P]
So if P=5b11111, it reversed all bits in a word.
     if P=5b00111 it swapped all bits in each byte.
     if P=5b11000 it swapped all bytes in a word
You could also reverse
nibbles in a word (P=5B00100),P=6B000100)
nibbles in a byte  (P=5B11100), P=6B111100)
odd&even bits.    (P=5b00001), P=6B000001)
swaphalves         (P=5b10000), P=6B100000)

So a really generalized reverse function would be
(I don't know the correct syntax, but you should get the idea)

function clause execute (GENERAL_REVERSE(rs1, rd),
to_bits(PARAM,xlenbits)) = { //PARAM is 5 or 6 bit bitvector
(to_bits() needed?)
  let rs1_val = X(rs1);
  result : xlenbits = zeros();
  foreach (i from 0 to (sizeof(xlen) -1)
    result[i] = rs1_val]i ^ PARAM]

and BREV8 would simply be

result[i] = GENERAL_REVERSE(rs1_rs1,rd, 7]

abukharmeh:
> when I do ys[i] = xs[index]; I get the following error msg..

I think I got this error when SAIL could not prove that the index is within range. I didn't have a look at the code to see the source of this, but thought mentioning it might help.

jrtc27:
Now that Zb[abcs] are in, could you please rebase this? Given how messy the history has become you may well find it easier to start from a clean history and create (a) new commit(s) than try and unpick what you have, as I suspect you'll have a lot of merge conflicts otherwise.

bilalsakhawat-10xe:
> Now that Zb[abcs] are in, could you please rebase this? Given how messy the history has become you may well find it easier to start from a clean history and create (a) new commit(s) than try and unpick what you have, as I suspect you'll have a lot of merge conflicts otherwise.

Yeah, I'll close this PR and will open a new one.

bilalsakhawat-10xe:
This Pull Request has many merge conflicts. Closing it and opened #135

martinberger:
Thanks.

PR #129, opened 2021-11-19
Add support for Zfh extension

bilalsakhawat-10xe:
“Zfh” Standard Extensions for Half-Precision Floating-Point, Version 0.1

Chapter 15 in the [draft](https://github.com/riscv/riscv-isa-manual/releases/tag/draft-20211118-d42b61f) spec.

FCVT.Q.H and FCVT.H.Q are not implemented. (Quad-Precision is not supported by sail yet)

### DIFF
--- a/model/riscv_insts_fext.sail
+++ b/model/riscv_insts_fext.sail
@@ -76,8 +76,9 @@
 
 /* **************************************************************** */
 /* IMPORTANT!                                                       */
-/* The files 'riscv_insts_fext.sail' and 'riscv_insts_dext.sail'    */
-/* define the F and D extensions, respectively.                     */
+/* The files 'riscv_insts_fext.sail', 'riscv_insts_dext.sail' and   */
+/* 'riscv_insts_zfh.sail' define the F, D and Zfh extensions,       */
+/* respectively.                                                    */
 /* The texts follow each other very closely; please try to maintain */
 /* this correspondence as the files are maintained for bug-fixes,   */
 /* improvements, and version updates.                               */
@@ -324,11 +325,14 @@ function haveSingleFPU() -> bool = haveFExt() | haveZfinx()
 /* Floating-point loads                                               */
 
 /* AST */
-/* FLW and FLD; W/D is encoded in 'word_width' */
+/* FLH, FLW and FLD; H/W/D is encoded in 'word_width' */
 
 union clause ast = LOAD_FP : (bits(12), regidx, regidx, word_width)
 
 /* AST <-> Binary encoding ================================ */
+
+mapping clause encdec = LOAD_FP(imm, rs1, rd, HALF)          if haveZfh()
+                    <-> imm @ rs1 @ 0b001 @ rd @ 0b000_0111  if haveZfh()
 
 mapping clause encdec = LOAD_FP(imm, rs1, rd, WORD)          if haveFExt()
                     <-> imm @ rs1 @ 0b010 @ rd @ 0b000_0111  if haveFExt()
@@ -359,6 +363,13 @@ function process_fload32(rd, addr, value) =
     MemException(e)  => { handle_mem_exception(addr, e); RETIRE_FAIL }
   }
 
+val process_fload16 : (regidx, xlenbits, MemoryOpResult(bits(16)))
+                      -> Retired effect {escape, rreg, wreg}
+function process_fload16(rd, addr, value) =
+  match value {
+    MemValue(result) => { F(rd) = nan_box(result); RETIRE_SUCCESS },
+    MemException(e)  => { handle_mem_exception(addr, e); RETIRE_FAIL }
+  }
 
 function clause execute(LOAD_FP(imm, rs1, rd, width)) = {
   let offset : xlenbits = EXTS(imm);
@@ -375,7 +386,8 @@ function clause execute(LOAD_FP(imm, rs1, rd, width)) = {
           let (aq, rl, res) = (false, false, false);
           match (width, sizeof(flen)) {
             (BYTE, _)   => { handle_illegal(); RETIRE_FAIL },
-            (HALF, _)   => { handle_illegal(); RETIRE_FAIL },
+            (HALF, _)   => 
+               process_fload16(rd, vaddr, mem_read(Read(Data), addr, 2, aq, rl, res)),
             (WORD, _)   =>
                process_fload32(rd, vaddr, mem_read(Read(Data), addr, 4, aq, rl, res)),
             (DOUBLE, 64) =>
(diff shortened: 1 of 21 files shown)


martinberger (commented):
Why do you commit prover snapshots? They are auto-generated.

PeterSewell:
it's often handy, e.g. so people can see the prover definitions directly

scottj97:
> Why do you commit prover snapshots? They are auto-generated.

They already existed in the repo, so he's just keeping them up to date, which is better than letting them fall behind.

Are you recommending these be deleted from the repo?

martinberger:
> it's often handy, e.g. so people can see the prover definitions directly

Yes, but then we need to make sure the generated prover definitions stay up-to-date with the sail model. For example generating prover snapshots in not in the GH Actions CI. So I'm wondering if they are up-to-date even

martinberger:
@PeterSewell How much work do you think it would be to script the building of all the prover snapshot with GitHub Actions, so we can remove the word "not" from https://github.com/riscv/sail-riscv/blob/master/prover_snapshots/README.md ? I imagine there was a reason why the current build process does not automate this.

model/riscv_softfloat_interface.sail
```diff
@@ -112,6 +113,35 @@ function update_softfloat_fflags(flags) = {
 /* **************************************************************** */
 /* ADD/SUB/MUL/DIV                                                  */
 
+
```
jrtc27, line 116: Stray extra line

model/riscv_insts_fext.sail
```diff
@@ -389,6 +418,13 @@ function process_fload32(rd, addr, value) =
     MemException(e)  => { handle_mem_exception(addr, e); RETIRE_FAIL }
   }
 
+val process_fload16 : (regidx, xlenbits, MemoryOpResult(bits(16)))
```
jrtc27, line 421: Really the functions should be the other way round, but that's not your problem, so this is the right thing to do in the situation

model/riscv_insts_zfh.sail
```diff
@@ -0,0 +1,916 @@
+
```
jrtc27, line 1: Should probably have the copyright header from fext given this is heavily derived from it, as well as an equivalent to:
```
/* This file specifies the instructions in the F extension          */
/* (single precision floating point).                               */

/* RISC-V follows IEEE 754-2008 floating point arithmetic standard. */

/* Original version written by Rishiyur S. Nikhil, Sept-Oct 2019    */

/* IMPORTANT!                                                       */
/* The files 'riscv_insts_fext.sail' and 'riscv_insts_dext.sail'    */
/* define the F and D extensions, respectively.                     */
/* The texts follow each other very closely; please try to maintain */
/* this correspondence as the files are maintained for bug-fixes,   */
/* improvements, and version updates.                               */

```
That should probably also be updated across all three to reference Zfh

jrtc27 (commented):
Unfortunately with the recent landing of Zfinx you picked a bad time to submit this pull request, and will need to rebase on top of that and apply the same changes to the new Zfh code. I think it should be relatively straightforward, if tedious. The only thing noteworthy that comes to mind is nan_(un)box should probably be made overloaded calling \_S and \_H versions (with the \_S version being what is currently there) rather than having nan_(un)box and nan_(un)box_H as is done here.

Glancing at the implementation I'm reminded of how non-conforming the F/D code is (blame Bluespec's internal code style for that, which I suspect is derived from treating everything like it's Haskell), but I agree with your choice here to copy the F/D code style (I assume you just copied F and swept through converting all the F-specificity to Zfh-specificity) rather than reformat it for Zfh and increase the divergence between them.

model/riscv_insts_zfh.sail
jrtc27: (Normally wouldn't mandate that block of documentation but F and D, and now Zfh, are a bit special in how they're implemented and we want to ensure they stay as similar as possible)

bilalsakhawat-10xe:
@jrtc27 I refactor the code for Zfh Extension. I think, modifications from Zfh to ZFinx_Zfh should be in a separate Pull Request. If you agree, I just need to add copyright header in the `model/riscv_insts_zfh.sail` and this PR is good to go.

jrtc27:
Deferring Zhinx seems ok to me, it's true that requesting it is moving the goalposts for this PR. Had you been slightly later and copied the F or D file with Z[fd]inx support then you would've been 95% of the way there already for free, but alas that's not what happened and so instead it's a bunch of work to repeat all the changes to the Zfh code.

I quickly skimmed the PR and noticed that the stray line comment from 8 days ago also still applies. I also found the "RV64 only" comment for the L[U] conversions mentions F not Zfh.

I'll try and take a more thorough look tomorrow.

martinberger:
> I'll try and take a more thorough look tomorrow.

@jrtc27 Did you manage to take a look?

model/riscv_insts_fext.sail
```diff
@@ -359,6 +363,13 @@ function process_fload32(rd, addr, value) =
...
+                      -> Retired effect {escape, rreg, wreg}
+function process_fload16(rd, addr, value) =
+  match value {
+    MemValue(result) => { F(rd) = nan_box_H(result); RETIRE_SUCCESS },
```
jrtc27, line 370: This can just be nan_box, same as process_fload32

model/riscv_insts_zfh.sail
```diff
@@ -0,0 +1,934 @@
...
+    Some(rm') => {
+      let rm_3b = encdec_rounding_mode(rm');
+      let (fflags, rd_val_16b) : (bits(5), bits(16)) = match op {
+        FADD_H  => riscv_f16Add (rm_3b, rs1_val_16b, rs2_val_16b),
+        FSUB_H  => riscv_f16Sub (rm_3b, rs1_val_16b, rs2_val_16b),
+        FMUL_H  => riscv_f16Mul (rm_3b, rs1_val_16b, rs2_val_16b),
+        FDIV_H  => riscv_f16Div (rm_3b, rs1_val_16b, rs2_val_16b)
+      };
+      write_fflags(fflags);
+      F(rd) = nan_box_H (rd_val_16b);
```
jrtc27, line 202: These can all be nan_box, for consistency with F (that or those should be changed to nan_box_S, but I don't see the benefit from that, might as well make use of the language's overloading to make the code a bit cleaner to read), though given they'll all need changing in future to `F_or_X_H(rd) = ...` with the NaN boxing implicit I guess this is fine, unlike the one for FP_LOAD which will stay and thus should be fixed (though arguably pushing that into a new F_H/S/D might make things slightly cleaner and more consistent, as a future minor refactor).

model/riscv_insts_zfh.sail
```diff
@@ -0,0 +1,934 @@
...
+<-> 0b000_1010 @ rs2 @ rs1 @ encdec_rounding_mode (rm) @ rd @ 0b101_0011  if haveZfh()
+
+mapping clause encdec =
+    F_BIN_RM_TYPE_H(rs2, rs1, rm, rd, FDIV_H)                             if haveZfh()
+<-> 0b000_1110 @ rs2 @ rs1 @ encdec_rounding_mode (rm) @ rd @ 0b101_0011  if haveZfh()
+
+/* Execution semantics ================================ */
+
+function clause execute (F_BIN_RM_TYPE_H(rs2, rs1, rm, rd, op)) = {
+  let rs1_val_16b = nan_unbox_H (F(rs1));
```
jrtc27, line 189: These will all need to use (a new) F_or_X_H but I guess it's fine to express it this way for now

jrtc27:
> > I'll try and take a more thorough look tomorrow.
> 
> @jrtc27 Did you manage to take a look?

I did not but have now. One minor comment, the rest observations. The commits should also be squashed down into a single commit rather than the 3 there are now (or 4 that it may become).

bilalsakhawat-10xe:
@jrtc27 I have incoporated the changes suggested by you.

jrtc27 (approved):
I think this is now ready

martinberger (approved):
Looks fine to me.

martinberger:
Merged.

Thanks @bilalsakhawat-10xe @jrtc27

PR #130, opened 2021-11-22
Implement support for Zfinx

jrtc27:
NB: Smstateen support is missing in the model so enabling the Zfinx extension provides an architectural covert channel via FCSR if privileged software is not aware of Zfinx's existence.

Full diff between the previous PR and this can be seen at https://gist.github.com/jrtc27/c26d373d866a6fac78ec69bc43f6c31f.

Summary of changes:
* sys_enable_fdext() and sys_enable_zfinx() are now mutually-exclusive, reflecting the architectural conflict, which remove the need for some places to check `sys_enable_fdext() & ~(sys_enable_zfinx())` (this is also now verified in init_sys)
* Added asserts to various F-only and Zfinx-only places for safety
* Sign-extend rather than NaN-box single-precision floats on RV64Zfinx
* Reverted nan_(un)box_do/pass changes back to just nan_(un)box
* Added dirty_fd_context_if_present wrapper
* Fixed have[FD]Ext to not be true for Z[fd]inx, since that does not make sense, instead adding haveSingle/DoubleFPU wrappers that are true for both F/D and their Z*inx counterparts
* Removed the (existing) is_RV32F_or_RV64F/is_RV64F/is_RV32D_or_RV64D/is_RV64D helpers that almost entirely pointless, and made things more awkward for Zfinx; is_RV32F_or_RV64F/is_RV32D_or_RV64D were totally redundant, as they were just haveFExt/haveDExt, and is_RV64F/is_RV64D were hardly used enough to warrant helpers
* Reverted some gratuitous diffs; a handful of let keywords were erroneously dropped, rendering the affected variables mutable, and an EXTS was dropped that I had deliberately added so the code would be correct for RV128
* Dropped the zfinx_en argument from X_or_F to allow register assignment syntax to work and be consistent with reg_or_freg_name
* Reverted back to register assignment syntax
* Renamed X_or_F to F_or_X and reg_or_freg_name to freg_or_reg_name to make it clearer at a glance these are for floating-point (and to reflect the fact that)
* A few whitespace fixes
* A few style improvements

It's quite possible I've missed something from that list, but that's all I recall.

Ibrahim: you are currently still the author of the commit as I just amended yours. If you want me to add myself as Co-authored-by, or swap it so I'm the author and you're in Co-authored-by, then please let me know; I don't care about the credit, but I don't want to attribute code to you that you're not happy with :)

CC @tariqkurd-repo @scottj97 (GitHub only lets repository members be added as reviewers, but please consider yourselves in that list)

### DIFF
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -18,6 +18,9 @@ bool sys_enable_next(unit u)
 bool sys_enable_fdext(unit u)
 { return rv_enable_fdext; }
 
+bool sys_enable_zfinx(unit u)
+{ return rv_enable_zfinx; }
+
 bool sys_enable_writable_misa(unit u)
 { return rv_enable_writable_misa; }
 
--- a/c_emulator/riscv_platform.h
+++ b/c_emulator/riscv_platform.h
@@ -4,6 +4,7 @@
 bool sys_enable_rvc(unit);
 bool sys_enable_next(unit);
 bool sys_enable_fdext(unit);
+bool sys_enable_zfinx(unit);
 bool sys_enable_writable_misa(unit);
 
 bool plat_enable_dirty_update(unit);
--- a/c_emulator/riscv_platform_impl.c
+++ b/c_emulator/riscv_platform_impl.c
@@ -4,6 +4,7 @@
 
 /* Settings of the platform implementation, with common defaults. */
 bool rv_enable_pmp                  = false;
+bool rv_enable_zfinx                = false;
 bool rv_enable_rvc                  = true;
 bool rv_enable_next                 = false;
 bool rv_enable_writable_misa        = true;
--- a/c_emulator/riscv_platform_impl.h
+++ b/c_emulator/riscv_platform_impl.h
@@ -8,6 +8,7 @@
 #define DEFAULT_RSTVEC     0x00001000
 
 extern bool rv_enable_pmp;
+extern bool rv_enable_zfinx;
 extern bool rv_enable_rvc;
 extern bool rv_enable_next;
 extern bool rv_enable_fdext;
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -127,6 +127,7 @@ static struct option options[] = {
   {"trace",                       optional_argument, 0, 'v'},
   {"no-trace",                    optional_argument, 0, 'V'},
   {"inst-limit",                  required_argument, 0, 'l'},
+  {"enable-zfinx",                no_argument,       0, 'x'},
 #ifdef SAILCOV
   {"sailcov-file",                required_argument, 0, 'c'},
 #endif
@@ -236,6 +237,7 @@ char *process_args(int argc, char **argv)
                     "V::"
                     "v::"
                     "l:"
+                    "x"
 #ifdef SAILCOV
                     "c:"
 #endif
@@ -324,6 +326,11 @@ char *process_args(int argc, char **argv)
       break;
     case 'l':
       insn_limit = atoi(optarg);
+      break;
+    case 'x':
+      fprintf(stderr, "enabling Zfinx support.\n");
+      rv_enable_zfinx = true;
+      rv_enable_fdext = false;
       break;
 #ifdef SAILCOV
     case 'c':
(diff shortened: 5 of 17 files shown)


martinberger (approved):
LGTM. Thanks very much for the code-improvements.

martinberger:
> let keywords were dropped, rendering the affected variables mutable

jrtc27:
> > let keywords were dropped, rendering the affected variables mutable
> 
> @jrtc27 Out of curiosity: why does the RISCV style prefer mutable variables where they are not needed? My natural inclination is to use let unless the variable in question is mutated.

I should have been clearer: I reinstated let keywords that were dropped in the original Zfinx PR for no reason. Immutability is very much preferred!

martinberger:
> > > let keywords were dropped, rendering the affected variables mutable
> > 
> > 
> > @jrtc27 Out of curiosity: why does the RISCV style prefer mutable variables where they are not needed? My natural inclination is to use let unless the variable in question is mutated.
> 
> I should have been clearer: I reinstated let keywords that were dropped in the original Zfinx PR for no reason. Immutability is very much preferred!

OK, good, then we agree. I misread this it as being the other way around.

abukharmeh:
Hi @jrtc27, Thank you for the updated PR. https://github.com/riscv/sail-riscv/blob/43e0dd239f64b0c330f807fed15e639c25c9fda6/model/riscv_fdext_control.sail#L80-L82  I think these lines need to be changed as a side effect of the changes done to haveFExt() and haveDExt() functions. I am still reviewing the rest of the PR and trying to see if there is something else.

jrtc27:
> Hi @jrtc27, Thank you for the updated PR.
> 
> https://github.com/riscv/sail-riscv/blob/43e0dd239f64b0c330f807fed15e639c25c9fda6/model/riscv_fdext_control.sail#L80-L82
> 
> I think these lines need to be changed as a side effect of the changes done to haveFExt() and haveDExt() functions. I am still reviewing the rest of the PR and trying to see if there is something else.

Ah yes, those live outside the insts files so I missed them. A quick git grep seems to suggest that's all though I think...

abukharmeh (approved):
It looks good to me, thank you.

abukharmeh:
Oh, please add yourself as co-author, or swap them around, I don't mind either way.

jrtc27:
I would've left it open a few days to see if Scott has any further feedback, don't see what harm that does... I can make mistakes or come up with poor design decisions just like everyone else, so another pair of critical eyes is always appreciated

martinberger:
> so another pair of critical eyes is always appreciated

Yes, but there is a slippery slope here. Code can always be made better.

jrtc27:
> > so another pair of critical eyes is always appreciated
> 
> Yes, but there is a slippery slope here. Code can always be made better.

At least given people time to express their opinions and listen to what they have to say; you can still choose to ignore it if it's too pedantic or you disagree with it. Plenty of open-source projects make forward progress in spite of rigorous code review processes...

martinberger:
> > > so another pair of critical eyes is always appreciated
> > 
> > 
> > Yes, but there is a slippery slope here. Code can always be made better.
> 
> At least give people time to express their opinions and listen to what they have to say; you can still choose to ignore it if it's too pedantic or you disagree with it. Plenty of open-source projects make forward progress in spite of rigorous code review processes...

We don't have a tangible code review process. I would love to evolve one over time. There is nothing from preventing us from making those changes later.

jrtc27:
> > > > so another pair of critical eyes is always appreciated
> > > 
> > > 
> > > Yes, but there is a slippery slope here. Code can always be made better.
> > 
> > 
> > At least give people time to express their opinions and listen to what they have to say; you can still choose to ignore it if it's too pedantic or you disagree with it. Plenty of open-source projects make forward progress in spite of rigorous code review processes...
> 
> There is nothing from preventing us from making those changes later.
> 
> We don't have a tangible code review process. I would love to evolve one over time. In order to get this started, let me express my understanding and expectation: if somebody makes a PR and suggests reviewers, then I expect that the person who made the PR assumes that the PR should be merged when the suggested reviewers approved. That's the heuristic I followed here.

I cannot tag Scott as a reviewer since he is not a member of the repository or organisation. That is why I had

> CC @tariqkurd-repo @scottj97 (GitHub only lets repository members be added as reviewers, but please consider yourselves in that list)

at the end of the description.

martinberger:
> at the end of the description.

OK, I didn't realise that. Rest assured that Tariq would have approved ...

PR #136, opened 2021-12-11
Configurable word size for test signature file output

bilalsakhawat-10xe:
Fix for the issue #111 
Now by default Sail with dump 32-bit (word) length signatures in the signature file.
To make it configurable `signature-granularity=<num-bytes-per-line>` cli is added, just like we have in spike.

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -68,6 +68,7 @@ size_t spike_dtb_len = 0;
 char *sig_file = NULL;
 uint64_t mem_sig_start = 0;
 uint64_t mem_sig_end = 0;
+int signature_granularity = 4;
 
 bool config_print_instr = true;
 bool config_print_reg = true;
@@ -120,6 +121,7 @@ static struct option options[] = {
   {"show-times",                  required_argument, 0, 'p'},
   {"report-arch",                 no_argument,       0, 'a'},
   {"test-signature",              required_argument, 0, 'T'},
+  {"signature-granularity",       required_argument, 0, 'g'},
 #ifdef RVFI_DII
   {"rvfi-dii",                    required_argument, 0, 'r'},
 #endif
@@ -230,6 +232,7 @@ char *process_args(int argc, char **argv)
                     "b:"
                     "t:"
                     "T:"
+                    "g"
                     "h"
 #ifdef RVFI_DII
                     "r:"
@@ -307,6 +310,10 @@ char *process_args(int argc, char **argv)
     case 'T':
       sig_file = strdup(optarg);
       fprintf(stderr, "using %s for test-signature output.\n", sig_file);
+      break;
+    case 'g':
+      signature_granularity = atoi(optarg);
+      fprintf(stderr, "setting signature-granularity to %d bytes\n", signature_granularity);
       break;
     case 'h':
       print_usage(argv[0], 0);
@@ -578,10 +585,10 @@ void write_signature(const char *file)
     fprintf(stderr, "Cannot open file '%s': %s\n", file, strerror(errno));
     return;
   }
-  /* write out words in signature area */
-  for (uint64_t addr = mem_sig_start; addr < mem_sig_end; addr += 4) {
+  /* write out words depending on signature granularity in signature area */
+  for (uint64_t addr = mem_sig_start; addr < mem_sig_end; addr += signature_granularity) {
     /* most-significant byte first */
-    for (int i = 3; i >= 0; i--) {
+    for (int i = signature_granularity - 1; i >= 0; i--) {
       uint8_t byte = (uint8_t) read_mem(addr+i);
       fprintf(f, "%02x", byte);
     }
--- a/ocaml_emulator/riscv_ocaml_sim.ml
+++ b/ocaml_emulator/riscv_ocaml_sim.ml
@@ -11,6 +11,7 @@ let opt_file_arguments = ref ([] : string list)
 let opt_dump_dts = ref false
 let opt_dump_dtb = ref false
 let opt_signature_file = ref (None : string option)
+let signature_granularity_ref = ref 4
 let opt_isa = ref (None : string option)
 
 let report_arch () =
@@ -24,6 +25,9 @@ let set_signature_file s =
   P.config_print_reg := false;
   P.config_print_mem_access := false;
   P.config_print_platform := false
+
+let set_signature_granularity sg =
+  signature_granularity_ref := sg
 
 let options = Arg.align ([("-dump-dts",
                            Arg.Set opt_dump_dts,
@@ -61,6 +65,9 @@ let options = Arg.align ([("-dump-dts",
                           ("-test-signature",
                            Arg.String set_signature_file,
                            " file for signature output (requires ELF signature symbols)");
+                          ("-signature-granularity",
+                           Arg.Int set_signature_granularity,
+                           " test signature granularity (in bytes)");
                           ("-isa",
                            Arg.String (fun s -> opt_isa := Some s),
                            " requested isa");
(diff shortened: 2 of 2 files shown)


abukharmeh:
This raises the question if we want to maintain consistency between c and ocaml emulators @martinberger and what is the expected usage of each of them !

martinberger:
> This raises the question weather we want to maintain consistency between c and ocaml emulators @martinberger and what is the expected usage of each of them !

@abukharmeh Good question. As far as I understand Sail's history, the  purpose of the OCaml emulator was to bootstrap the C emulator, since it's much easier to get OCaml correct than C. The OCaml emulator is quite a lot slower than the C emulator. (@PeterSewell @Alasdair please correct me if I'm wrong)

allenjbaum:
In https://github.com/riscv/sail-riscv/issues/111, jrtc27 commented on Oct 4
Oh ok that thing (which also exists for the OCaml simulator, I might add). Making that configurable isn't hard.

Which implies to me that it is already in the Ocaml simulator - so that shouldn't be an issue.

abukharmeh:
Yes its OCaml emulators, but is not configurable (as far as I can see). If we want to maintain consistency between emulators, then this function should be changed.

https://github.com/riscv/sail-riscv/blob/5549f159b7b7f1fd0e27b37202eee0d103661e9e/ocaml_emulator/riscv_ocaml_sim.ml#L109-L119

allenjbaum:
OK, that's odd since (I interpreted) jrtc27's comment
 " Oh ok that thing (which also exists for the OCaml simulator, I might
add). Making that configurable isn't hard."
to mean that it was already in OCaml, but I can see how it might mean that
the same implementation that is the C-model is also in the OCaml model, and
therefore should also be fixed.

jrtc27:
Ibrahim's interpretation is what I was trying to convey; I specifically called out the OCaml version as also having the test signature dumping feature, since https://github.com/riscv/sail-riscv/issues/111#issuecomment-933673462 only referenced the C version, and wanted to make sure we kept feature parity.

allenjbaum:
Thanks; I naturally interpreted exactly the opposite. It's one of my
superpowers.
And why we need more formal definitions....

allenjbaum:
So, we have a PR for the CSIM support; the question is whether we should
fix both C-sim and OCaml in the same PR or separate ones.
Any opinions?

bilalsakhawat-10xe:
@abukharmeh @jrtc27  I made similar changes to ocaml emulator/simulator. It's my first time writing ocaml syntax so suggestions are welcome.

ocaml_emulator/riscv_ocaml_sim.ml
```diff
@@ -109,13 +116,15 @@ let check_elf () =
...
   while !i < Bytes.length bytes do
-    let s = Printf.sprintf "%02x%02x%02x%02x\n"
-	(int_of_char (Bytes.get bytes (!i+3)))
-	(int_of_char (Bytes.get bytes (!i+2)))
-	(int_of_char (Bytes.get bytes (!i+1)))
-	(int_of_char (Bytes.get bytes (!i+0))) in
-    output_string fl s;
-    i := !i + 4
+    let j = ref 0 in
+    while !j < !signature_granularity_ref do
```
martinberger, line 120: Why not use a `for`-loop here, then you can avoid hand-incrementing the counter `j`

martinberger: Something like

    for j = 0 to !signature_granularity_ref - 1 do
        ...
    done

martinberger: I don't think that OCaml has a `for` loop by increments other than 1 and -1, so the other loop can't be a `for`-loop alas

c_emulator/riscv_sim.c
```diff
@@ -68,6 +68,7 @@ size_t spike_dtb_len = 0;
 char *sig_file = NULL;
 uint64_t mem_sig_start = 0;
 uint64_t mem_sig_end = 0;
+uint64_t signature_granularity = 4;
```
abukharmeh, line 71: Why the size of this is 64 bit, and then you are using int in the for loop ?

abukharmeh: Apart from that, I dont have anything else to comment on, and this looks and works fine to me.

abukharmeh: Oh actually, it might be helpfull to assert that signature_granularity is divisable by 4 !

bilalsakhawat-10xe: sure I'll change the signature_granularity size to int.
About "Δ mem_sig is divisable by signature_granularity" can you please elaborate it a little? like when Δ mem_sig is 
 not divisable by signature_granularity, is the signature granularity is invalid or Δ mem_sig.

ocaml_emulator/riscv_ocaml_sim.ml
bilalsakhawat-10xe: So, I can only change the inner while loop to a for loop, right? I'll do it in the next commit.

c_emulator/riscv_sim.c
abukharmeh: Maybe this is unnecessary, do we know for sure that the addresses are word-aligned or is there any alignment restriction on them? In the last iteration of the loop, if addr < mem_sig_end, then we will increment by signature_granularity, which could make our address outside the signature region?

allenjbaum:
If the signature isn't aligned, then it's not a Sail bug, it's a test bug.
Also the SIGUPD macros are supposed to keep the signature pointer XLEN
aligned

c_emulator/riscv_sim.c
```diff
@@ -578,10 +585,10 @@ void write_signature(const char *file)
     fprintf(stderr, "Cannot open file '%s': %s\n", file, strerror(errno));
     return;
   }
-  /* write out words in signature area */
-  for (uint64_t addr = mem_sig_start; addr < mem_sig_end; addr += 4) {
+  /* write out words depending on signature granularity in signature area */
+  for (uint64_t addr = mem_sig_start; addr < mem_sig_end; addr += signature_granularity) {
     /* most-significant byte first */
-    for (int i = 3; i >= 0; i--) {
+    for (int i = signature_granularity - 1 ; i >= 0; i--) {
```
jrtc27, line 591:
```suggestion
    for (int i = signature_granularity - 1; i >= 0; i--) {
```

ocaml_emulator/riscv_ocaml_sim.ml
```diff
@@ -109,13 +116,13 @@ let check_elf () =
...
-	(int_of_char (Bytes.get bytes (!i+0))) in
-    output_string fl s;
-    i := !i + 4
+    for j = 0 to !signature_granularity_ref - 1 do
+      let s = Printf.sprintf "%02x"
+          (int_of_char (Bytes.get bytes (!i + (!signature_granularity_ref - 1 - j)))) in
+      output_string fl s;
+    done;
+    i := !i + !signature_granularity_ref;
+    output_string fl "\n";
```
jrtc27, line 125: Probably nicer to keep the original order, i.e. swap these two lines back?

ocaml_emulator/riscv_ocaml_sim.ml
```diff
@@ -61,6 +65,9 @@ let options = Arg.align ([("-dump-dts",
                           ("-test-signature",
                            Arg.String set_signature_file,
                            " file for signature output (requires ELF signature symbols)");
+                          ("-signature-granularity",
+                           Arg.Int set_signature_granularity,
+                           " test signature granularity (in Bytes)");
```
jrtc27, line 70:
```suggestion
                           " test signature granularity (in bytes)");
```
unless Bytes specifically is used elsewhere that I can't see (I can see MB but that's different)

ocaml_emulator/riscv_ocaml_sim.ml
```diff
@@ -109,13 +116,13 @@ let check_elf () =
...
-    let s = Printf.sprintf "%02x%02x%02x%02x\n"
-	(int_of_char (Bytes.get bytes (!i+3)))
-	(int_of_char (Bytes.get bytes (!i+2)))
-	(int_of_char (Bytes.get bytes (!i+1)))
-	(int_of_char (Bytes.get bytes (!i+0))) in
-    output_string fl s;
-    i := !i + 4
+    for j = 0 to !signature_granularity_ref - 1 do
+      let s = Printf.sprintf "%02x"
+          (int_of_char (Bytes.get bytes (!i + (!signature_granularity_ref - 1 - j)))) in
```
jrtc27, line 121: You could use `1 to !signature_granularity_ref` for the loop and drop the - 1 here, or just `for j = !signature_granularity_ref - 1 downto 0` like the C version and not have to do the inversion at all

bilalsakhawat-10xe:
@jrtc27 I have made the suggested changes.

jrtc27 (approved):
Thanks

martinberger (approved):
Looks fine.

martinberger:
Merged!

Thanks @bilalsakhawat-10xe @abukharmeh @jrtc27

PR #137, opened 2021-12-18
add support for cmo extension

liweiwei90:
add support for cmo extension:
- add parameter for cache block size
- add cbo.zero to zero the cache block
- add cbo.inval/flush/clean to do envcfg csr and memory access check
- add prefetch* to only do the address-translation

### DIFF
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -82,6 +82,11 @@ mach_bits plat_rom_size(unit u)
   return rv_rom_size;
 }
 
+mach_bits plat_cache_block_size()
+{
+  return rv_cache_block_size;
+}
+
 // Provides entropy for the scalar cryptography extension.
 mach_bits plat_get_16_random_bits()
 {
--- a/c_emulator/riscv_platform_impl.c
+++ b/c_emulator/riscv_platform_impl.c
@@ -20,6 +20,8 @@ uint64_t rv_ram_size = UINT64_C(0x4000000);
 uint64_t rv_rom_base = UINT64_C(0x1000);
 uint64_t rv_rom_size = UINT64_C(0x100);
 
+uint64_t rv_cache_block_size = UINT64_C(0x40);
+
 // Provides entropy for the scalar cryptography extension.
 uint64_t rv_16_random_bits(void)
 {
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -137,6 +137,7 @@ static struct option options[] = {
 #ifdef SAILCOV
     {"sailcov-file",                required_argument, 0, 'c'},
 #endif
+    {"cache-block-size",            required_argument, 0, 'B'},
     {0,                             0,                 0, 0  }
 };
 
@@ -221,6 +222,7 @@ char *process_args(int argc, char **argv)
 {
   int c;
   uint64_t ram_size = 0;
+  uint64_t block_size = 0;
   while (true) {
     c = getopt_long(argc, argv,
                     "a"
@@ -249,6 +251,7 @@ char *process_args(int argc, char **argv)
                     "V::"
                     "v::"
                     "l:"
+                    "B:"
                     "x",
                     options, NULL);
     if (c == -1)
@@ -352,6 +355,17 @@ char *process_args(int argc, char **argv)
       sailcov_file = strdup(optarg);
       break;
 #endif
+    case 'B':
+      block_size = atol(optarg);
+      if ((block_size & (block_size - 1)) == 0) {
+        fprintf(stderr, "setting cache-block-size to %" PRIu64 " B\n",
+                block_size);
+        rv_cache_block_size = block_size;
+      } else {
+        fprintf(stderr, "invalid cache-block-size '%s' provided.\n", optarg);
+        exit(1);
+      }
+      break;
     case '?':
       print_usage(argv[0], 1);
       break;
--- a/model/riscv_addr_checks.sail
+++ b/model/riscv_addr_checks.sail
@@ -116,7 +116,7 @@ type ext_data_addr_error = unit
 
 /* Default data addr is just base register + immediate offset (may be zero).
    Extensions might override and add additional checks. */
-function ext_data_get_addr(base : regidx, offset : xlenbits, acc : AccessType(ext_access_type), width : word_width)
+function ext_data_get_addr(base : regidx, offset : xlenbits, acc : AccessType(ext_access_type), width : xlenbits)
          -> Ext_DataAddr_Check(ext_data_addr_error) =
   let addr = X(base) + offset in
   Ext_DataAddr_OK(addr)
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -127,7 +127,7 @@ function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
     /* Get the address, X(rs1) (no offset).
      * Extensions might perform additional checks on address validity.
      */
-    match ext_data_get_addr(rs1, zeros(), Read(Data), width) {
+    match ext_data_get_addr(rs1, zeros(), Read(Data), EXTZ(size_bits(width))) {
       Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
       Ext_DataAddr_OK(vaddr) => {
         let aligned : bool =
@@ -188,7 +188,7 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
       /* Get the address, X(rs1) (no offset).
        * Extensions might perform additional checks on address validity.
        */
-      match ext_data_get_addr(rs1, zeros(), Write(Data), width) {
+      match ext_data_get_addr(rs1, zeros(), Write(Data), EXTZ(size_bits(width))) {
         Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
         Ext_DataAddr_OK(vaddr) => {
           let aligned : bool =
@@ -278,7 +278,7 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
     /* Get the address, X(rs1) (no offset).
      * Some extensions perform additional checks on address validity.
      */
-    match ext_data_get_addr(rs1, zeros(), ReadWrite(Data, Data), width) {
+    match ext_data_get_addr(rs1, zeros(), ReadWrite(Data, Data), EXTZ(size_bits(width))) {
       Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
       Ext_DataAddr_OK(vaddr) => {
         match translateAddr(vaddr, ReadWrite(Data, Data)) {
--- a/model/riscv_insts_zicbom.sail
+++ b/model/riscv_insts_zicbom.sail
@@ -25,6 +25,7 @@
 /*    Aril Computer Corp., for contributions by Scott Johnson                            */
 /*    Philipp Tomsich                                                                    */
 /*    VRULL GmbH, for contributions by its employees                                     */
+/*    PLCT, for contributions by its employees                                           */
 /*                                                                                       */
 /*  All rights reserved.                                                                 */
 /*                                                                                       */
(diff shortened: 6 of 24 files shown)


model/riscv_insts_cmoext.sail
```diff
@@ -0,0 +1,245 @@
...
+  PREFETCH_I <-> "prefetch.i",
+  PREFETCH_R <-> "prefetch.r",
+  PREFETCH_W <-> "prefetch.w"
+}
+
+mapping clause assembly = RISCV_PREFETCH(op, rs1, offset)
+  <-> pre_mnemonic(op) ^ spc() ^ hex_bits_12(offset @ 0b00000) ^ "(" ^ reg_name(rs1) ^ ")"
+
+function clause execute(RISCV_PREFETCH(preop, rs1, offset)) = {
+  /* do nothing other than translate_address for prefetch.i/r/w */
```
jrtc27, line 211: Why do we even bother doing that? They're just hints, they can be entirely nop'ed out in the model like any other hint. Which is already the case, given they're ORI instructions, and base.sail gets included before cmoext.sail so the real ORIs will take precedence and this code is entirely dead.

Especially since this is actually wrong, due to translateAddr having the side-effect of (if enabled instead of trapping for software emulation) atomically updating accessed and dirty bits as needed, which is specifically not performed for prefetch.

model/riscv_insts_cmoext.sail
```diff
@@ -0,0 +1,245 @@
...
+        Ext_DataAddr_Error(e)  => RETIRE_SUCCESS,
+        Ext_DataAddr_OK(vaddr) =>
+          match translateAddr(vaddr, Read(Data)) {
+            TR_Failure(e, _)     => RETIRE_SUCCESS,
+            TR_Address(paddr, _) => RETIRE_SUCCESS
+          }
+      }
+    }
+  }
+}
```
jrtc27, line 245: Make sure every line has a terminating newline character (required by POSIX for a file to be a valid text file)

model/riscv_sys_control.sail
```diff
@@ -89,6 +89,7 @@ function is_CSR_defined (csr : csreg, p : Privilege) -> bool =
     0x304 => p == Machine, // mie
     0x305 => p == Machine, // mtvec
     0x306 => p == Machine & haveUsrMode(), // mcounteren
+    0x30a => p == Machine, //menvcfg
```
jrtc27, line 92: Space after the `//`

model/riscv_sys_regs.sail
```diff
@@ -813,3 +813,18 @@ function read_seed_csr() -> xlenbits = {
...
+  CBCFE : 6,
+  CBIE  : 5 .. 4
+}
+register senvcfg : Envcfg
+register menvcfg : Envcfg
+
+function legalize_envcfg(c : Envcfg, v : xlenbits) -> Envcfg = {
+  let c = update_CBZE(c, [v[7]]);
+  let c = update_CBCFE(c, [v[6]]);
+  let c = update_CBIE(c, v[5 .. 4]);
```
jrtc27, line 828: Doesn't this need to legalise 10?

model/riscv_types.sail
```diff
@@ -389,6 +389,8 @@ enum sopw = {RISCV_SLLIW, RISCV_SRLIW,
 enum amoop = {AMOSWAP, AMOADD, AMOXOR, AMOAND, AMOOR,
               AMOMIN, AMOMAX, AMOMINU, AMOMAXU}   /* AMO ops */
 enum csrop = {CSRRW, CSRRS, CSRRC}                /* CSR ops */
+enum cbop = {CBO_CLEAN, CBO_FLUSH, CBO_INVAL, CBO_ZERO}  /* cbo ops*/
+enum preop = {PREFETCH_I, PREFETCH_R, PREFETCH_W} /* prefetch ops*/
```
jrtc27, line 393: Follow the convention below; put them in a separate block and include the extension names

ocaml_emulator/platform_impl.ml
```diff
@@ -57,6 +57,8 @@ let rom_base   = 0x00001000L;;  (* Spike::DEFAULT_RSTVEC *)
 
 let dram_size_ref = ref (Int64.(shift_left 64L 20))
 
+let cache_block_size_ref = ref (Int64.(64L))
```
jrtc27, line 60: Ok so this one defaults to 64 as I'd expect, unlike the C model

c_emulator/riscv_sim.c
```diff
@@ -337,6 +340,16 @@ char *process_args(int argc, char **argv)
       sailcov_file = strdup(optarg);
       break;
 #endif
+    case 'B':
+      block_size = atol(optarg);
+      if (((block_size & 7) == 0) && (block_size < 4096)) {
```
jrtc27, line 345: I can't obviously find either of these restrictions in the spec; where have these come from? 4096 seems pretty unproblematic, that's a whole page, things just get awkward when you cover multiple pages, but the PMP already makes that an issue.

model/riscv_insts_cmoext.sail
```diff
@@ -0,0 +1,245 @@
...
+  TransTrap : ExceptionType,
+  TransSuccess : unit
+}
+
+val process_clean_inval : (xlenbits, xlenbits, bool, bool) -> Retired effect {escape, rmem, rmemt, rreg, wmv, wmvt, wreg}
+function process_clean_inval(vaddr, width, clean, inval) = {
+  let load: translation_check = match ext_data_get_addr(to_bits(5, 0), vaddr, Read(Data), BYTE) {
+    Ext_DataAddr_Error(e)  => TransTrap(E_Reserved_14()),
+    Ext_DataAddr_OK(vaddr) =>
+      match translateAddr(vaddr, Read(Data)) {
```
jrtc27, line 78: How does invalidate interact with the dirty bit of page table entries? Just translating a read won't dirty the page, but surely an invalidate needs to dirty the page if actually invalidating rather than just cleaning/flushing? Though whether it's "only dirty if the cache line is dirty" or "always dirty even if there's nothing to write back" I don't know. I can only see discussion of accessed and dirty bits for prefetch and zeroing, not management instruction.

jrtc27: (The answer to that affects whether or not you can clean this up by just hoisting all three load, store and fetch calculations and matching on them all at the same time, which would flatten this function somewhat)

jrtc27 (commented):
Various initial feedback; generally only commented on the first instance of something even when it applies multiple times, so please check for other instances of things I've said, especially when it comes to code formatting

liweiwei90:
> Various initial feedback; generally only commented on the first instance of something even when it applies multiple times, so please check for other instances of things I've said, especially when it comes to code formatting

OK, I'll try to seperate the three extensions.

c_emulator/riscv_platform_impl.c
```diff
@@ -20,6 +20,8 @@ uint64_t rv_ram_size = UINT64_C(0x4000000);
 uint64_t rv_rom_base = UINT64_C(0x1000);
 uint64_t rv_rom_size = UINT64_C(0x100);
 
+uint64_t rv_cache_block_size = UINT64_C(0x10);
```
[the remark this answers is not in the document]
liweiwei90: OK, I'll change this to 64.

model/riscv_insts_cmoext.sail
```diff
@@ -0,0 +1,245 @@
...
+union clause ast = RISCV_CBO : (cbop, regidx)
+
+mapping encdec_cbop : cbop <-> bits(12) = {
+  CBO_CLEAN <-> 0b000000000001,
+  CBO_FLUSH <-> 0b000000000010,
+  CBO_INVAL <-> 0b000000000000,
+  CBO_ZERO  <-> 0b000000000100
+}
+
+mapping clause encdec = RISCV_CBO(cbop, rs1)
```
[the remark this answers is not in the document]
liweiwei90: OK. I'll add this.

allenjbaum:
Thanks for an incredibly quick and thorough review.
But, at one point you say there shouldn't be random spaces, but the
guidelines said you can add them to align things vertically to make them
easier to understand.
Which would be fine if in this case he added a couple more extra spaces to
make the expressions actually line up.
Do I have that correct?

allenjbaum:
I am concerned that jrtc27 has pointed out things that are bugs in the
code, yet there are no tests that cover those.
Shouldn't there be?

jrtc27:
> Thanks for an incredibly quick and thorough review. But, at one point you say there shouldn't be random spaces, but the guidelines said you can add them to align things vertically to make them easier to understand. Which would be fine if in this case he added a couple more extra spaces to make the expressions actually line up. Do I have that correct?
> […](#)

Yes, that's right; I took an educated guess based on the low number of additional spaces that the intent was not to line things up

dkruckemyer-ventana:
What are the next steps to take here?

model/riscv_insts_zicbom.sail
```diff
@@ -0,0 +1,81 @@
...
+  CBO_FLUSH <-> "cbo.flush",
+  CBO_INVAL <-> "cbo.inval"
+}
+
+mapping clause assembly = RISCV_ZICBOM(cbop, rs1)
+  <-> cbop_mnemonic(cbop) ^ spc() ^ reg_name(rs1)
+
+val process_clean_inval : (xlenbits, xlenbits, bool, bool) -> Retired effect {escape, rmem, rmemt, rreg, wmv, wmvt, wreg}
+function process_clean_inval(vaddr, width, clean, inval) = {
+  match ext_data_get_addr(zeros(), vaddr, Read(Data), BYTE) {
```
jrtc27, line 28: As I'm pretty sure I said before, ext_data_get_addr must take the base register that yielded that address in the first argument, not zeroes. That is, you need to pass rs1 (the index) through to here.

model/riscv_insts_zicboz.sail
```diff
@@ -0,0 +1,57 @@
...
+  ~(((p != Machine) & (menvcfg.CBZE() == 0b0)) | ((p == User) & (senvcfg.CBZE() == 0b0)))
+}
+
+val process_cbo_zero : (xlenbits, xlenbits) -> Retired effect {eamem, escape, rmem, rmemt, rreg, wmv, wmvt, wreg}
+function process_cbo_zero(vaddr, width) = {
+  offset : xlenbits = zeros();
+  failed : bool = false;
+  while (offset <_u width) & (failed == false) do {
+    let addr = vaddr + offset;
+    match ext_data_get_addr(zeros(), addr, Write(Data), BYTE) {
```
jrtc27, line 20: As with Zicbom, this must take rs1 as the first argument

c_emulator/riscv_platform.c
```diff
@@ -48,6 +48,12 @@ mach_bits plat_rom_base(unit u)
 mach_bits plat_rom_size(unit u)
 { return rv_rom_size; }
 
+mach_bits plat_zicbom_cache_block_size()
```
arichardson, line 51: Why are there two separate configurable block sizes? I assume this will always be the cache block size so if both extensions are supported the values must be the same?
`plat_cache_block_size()`. Also this function signature is missing the `unit u` argument.

c_emulator/riscv_sim.c
```diff
@@ -133,6 +133,8 @@ static struct option options[] = {
 #ifdef SAILCOV
   {"sailcov-file",                required_argument, 0, 'c'},
 #endif
+  {"zicbom-cache-block-size",            required_argument, 0, 'M'},
```
arichardson, line 136: I don't think there should be multiple options here but rather a single `--cache-block-size` option.

As the spec says: `In the initial set of CMO extensions, the
size of a cache block shall be uniform throughout the system.` so it really doesn't make sense to have different sizes of Zicboz and Zicbom.

arichardson:
The description of this PR says prefetch is supported, but I don't see any Zicbop support in the code? Maybe that got lost while splitting up the files?

ptomsich:
What is the status on this PR?
I see plenty of unresolved review comments but no recent code changes (last update is 8 months old)... who is driving this?

Given that the Zicbo[mpz] extensions are comparatively self-contained (and we don't have to implement any real cache-management functionality/prefetchers), I'd like to see this pushed over the finish line sooner than later.

liweiwei90:
> The description of this PR says prefetch is supported, but I don't see any Zicbop support in the code? Maybe that got lost while splitting up the files?
Yeah. the file with the support for Zicbop  is deleted in previous updates. It will be dead code  since prefetch instructions reuse the encoding of existed instruction.

liweiwei90:
> What is the status on this PR? I see plenty of unresolved review comments but no recent code changes (last update is 8 months old)... who is driving this?

Sorry, I didn't  maintain this PR  after I replied the comments and update the code to fix the problems several month ago. I think most of the questions have been resolved or replied(may need further discussion). 
> 
> Given that the Zicbo[mpz] extensions are comparatively self-contained (and we don't have to implement any real cache-management functionality/prefetchers), I'd like to see this pushed over the finish line sooner than later.

Yeah. I agree. I'll try to re-maintain and update it sooner.

jrtc27:
At a glance I don't think the extension hooks comments have been properly addressed

liweiwei90:
> At a glance I don't think the extension hooks comments have been properly addressed

Sorry. Do you mean ext_data_get_addr  problem?

jrtc27:
> > At a glance I don't think the extension hooks comments have been properly addressed
> 
> Sorry. Do you mean ext_data_get_addr problem?

Yes. You give it a register index for the base address register used, an offset from that register's value and the size of the memory operation being performed. All of those must be correct. Currently you pass zeros() as the register index, the virtual address as the offset and BYTE as the width. This works in the upstream Sail model because all it does is return `Ext_DataAddr_OK(X(base) + offset)`, but that relies on the specifics of the implementation; your use does not conform to the interface. This would break our work downstream which uses that hook for something actually useful, and needs the register index and access width to be correct.

liweiwei90:
> > > At a glance I don't think the extension hooks comments have been properly addressed
> > 
> > 
> > Sorry. Do you mean ext_data_get_addr problem?
> 
> Yes. You give it a register index for the base address register used, an offset from that register's value and the size of the memory operation being performed. All of those must be correct. Currently you pass zeros() as the register index, the virtual address as the offset and BYTE as the width. This works in the upstream Sail model because all it does is return `Ext_DataAddr_OK(X(base) + offset)`, but that relies on the specifics of the implementation; your use does not conform to the interface. This would break our work downstream which uses that hook for something actually useful, and needs the register index and access width to be correct.

Yeah. I don't find the original meaning of this function as I commented before.  So I used it as an address calculation progress.
For the rs1 and offset, I can adjust it to conform the interface. However, I think cache block size is right size for the memory access. However it seems not a corrent value  for it. So Byte is used currently.

jrtc27:
> > > > At a glance I don't think the extension hooks comments have been properly addressed
> > > 
> > > 
> > > Sorry. Do you mean ext_data_get_addr problem?
> > 
> > 
> > Yes. You give it a register index for the base address register used, an offset from that register's value and the size of the memory operation being performed. All of those must be correct. Currently you pass zeros() as the register index, the virtual address as the offset and BYTE as the width. This works in the upstream Sail model because all it does is return `Ext_DataAddr_OK(X(base) + offset)`, but that relies on the specifics of the implementation; your use does not conform to the interface. This would break our work downstream which uses that hook for something actually useful, and needs the register index and access width to be correct.
> 
> Yeah. I don't find the original meaning of this function as I commented before. So I used it as an address calculation progress. For the rs1 and offset, I can adjust it to conform the interface. However, I think cache block size is right size for the memory access. However it seems not a correct value for it. So Byte is used currently.

We can change it to take the actual number of bytes rather than one of the width enum values.

liweiwei90:
c

> > > > > At a glance I don't think the extension hooks comments have been properly addressed
> > > > 
> > > > 
> > > > Sorry. Do you mean ext_data_get_addr problem?
> > > 
> > > 
> > > Yes. You give it a register index for the base address register used, an offset from that register's value and the size of the memory operation being performed. All of those must be correct. Currently you pass zeros() as the register index, the virtual address as the offset and BYTE as the width. This works in the upstream Sail model because all it does is return `Ext_DataAddr_OK(X(base) + offset)`, but that relies on the specifics of the implementation; your use does not conform to the interface. This would break our work downstream which uses that hook for something actually useful, and needs the register index and access width to be correct.
> > 
> > 
> > Yeah. I don't find the original meaning of this function as I commented before. So I used it as an address calculation progress. For the rs1 and offset, I can adjust it to conform the interface. However, I think cache block size is right size for the memory access. However it seems not a correct value for it. So Byte is used currently.
> 
> We can change it to take the actual number of bytes rather than one of the width enum values.

OK. I'll change it.

ptomsich:
Action items from today:
- @liweiwei90 Update PR w/ new patch set and provide information on testing (e.g., link to the cbo.zero ACT PR)
- @billmcspadden-riscv Find a subject matter expert (from the CMO TG?) to review the updated test set.

liweiwei90:
> Action items from today:
> 
>     * @liweiwei90 Update PR w/ new patch set and provide information on testing (e.g., link to the cbo.zero ACT PR)
>
OK. I'll check the above comments again and mark them as resolved or update the patch if needed. The ACT PR can be found in https://github.com/riscv-non-isa/riscv-arch-test/pull/226

liweiwei90:
Two problems may need further discussion:
1. What is the reasonable size range for cache block size? Whether cache block size can be multiple pages?
2. For pmpcheck, whether xlenbits type can be converted to atom type? How ?

ptomsich:
> Two problems may need further discussion:
> 
> 1. What is the reasonable size range for cache block size? Whether cache block size can be multiple pages?

The specification does not disallow it, so it must be supported by our implementations.

The only requirements (from the specification) are:
- naturally aligned
- power-of-two

model/riscv_insts_aext.sail
```diff
@@ -127,7 +127,7 @@ function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
     /* Get the address, X(rs1) (no offset).
      * Extensions might perform additional checks on address validity.
      */
-    match ext_data_get_addr(rs1, zeros(), Read(Data), width) {
+    match ext_data_get_addr(rs1, zeros(), Read(Data), EXTZ(size_bits(width))) {
```
Timmmm, line 130: Shouldn't these `size_bits` be `word_width_bytes` everywhere?

Also when you rebase you'll have to upload a load of the vector code that calls `ext_data_get_addr()`. Though you can be lazy and just wait if you want, because I am rebasing/updating this...

model/riscv_addr_checks.sail
```diff
@@ -116,7 +116,7 @@ type ext_data_addr_error = unit
 
 /* Default data addr is just base register + immediate offset (may be zero).
    Extensions might override and add additional checks. */
-function ext_data_get_addr(base : regidx, offset : xlenbits, acc : AccessType(ext_access_type), width : word_width)
+function ext_data_get_addr(base : regidx, offset : xlenbits, acc : AccessType(ext_access_type), width : xlenbits)
```
Timmmm, line 119: Is `xlenbits` the right type? Maybe `nat` would be better (or `range(1, 1024)` for performance reasons).

davidharrishmc:
As mentioned back in May, cbo.zero support in Sail is gating the ability to verify the cbo.zero tests that were recently added to riscv-arch-test. Looking forward to this PR being completed.

https://github.com/riscv-non-isa/riscv-arch-test/pull/226

Timmmm:
I've prepared a PR for this, but the people that need to approve me uploading it are on holiday. And also it depends on #350 and an HPM counter PR that I also need to get approval for too. Hopefully not too much longer. #350 is ready to merge I think.

davidharrishmc:
Nudge.

Timmmm:
Coincidentally I prepared a PR today. Just getting approval to send it (shouldn't take long).

Timmmm:
#455 was merged so I'll close this one. Thanks for the initial code @liweiwei90!

PR #140, opened 2021-12-29
Sail support rvv

b224hisl:
We are from RIOS lab, being responsible for providing sail support for RVV(RISCV-VECTOR).
This version of Sail-RVV codes have passed our hand-written assembly tests for LMUL=1, XLEN=64, VLEN=128 and VSEW=32. We will keep maintain this project and while generating more tests with other configs


jrtc27:
This is clearly still utterly broken.

It looks like you were working with a year-old tree and then rebased prior to submission, with your conflict resolution strategy being "replace every single file in the tree (not just ones where they conflict) with our version based on sources from 2020", thereby rolling back the past year of work done upstream on the model.

You need to start again from a clean up-to-date sail-riscv checkout and copy in your V extension implementation manually, unless you still have the old pre-rebase branch lying around that you can rebase again, which would need to be done with proper conflict resolution, not `git checkout HEAD .`.

jrtc27:
Also you do not need to submit a new PR, you can just force push to your branch.

Issue #117, opened 2021-10-16
TOR pmp matching with same hi and lo pmpaddr


stuarth-sfu:
In the case the ranges for a PMP check in TOR mode are the same (ie lo == hi), I am not sure the Sail specification/model will do the right thing.
I believe the correct behaviour here will always be to not match the entry, however the Sail specification would give a partial match.
The effects of this are subtle, because a no match will then go on to check the next entry, wheras a part match will cause the whole check to fail.

Take the following example :
if  pmpaddr[0] == pmpaddr[1] == 0x...10000
the address to check is 0x....FFFF for a 2 byte access
This means the two bytes to check are 0x....FFFF and 0x...10000 neither of these should match according to The privileged architecture (section 3.7.1, 20210921-draft), so the result should be "no match" for that entry

here lo == hi == 0x...10000, addr = 0x....FFFF, addr + width = 0x...10001
In the sail spec this will result in a partial match because the first check of the range values is for > not >=, then the second if condition, (addr + width <=_u lo) | (hi <=_u addr), will not be so it falls into the second condition, (lo <=_u addr) & (addr + width <=_u hi), which is also not met so results in a partial match.

Should the initial misconfiguration check,  hi <_u lo, be checking for hi <=_u lo?

allenjbaum:
This is a confirmed  Sail bug. A partial match is if the first or last byte is entirely within a range, and the other isn't.
if the low and high range limits are the same, then the range size is zero, and it is impossible for either end to be inside the range, because there is no inside....'
I don't think I can tag this as a bug, but it should be.

ptomsich:
@allenjbaum Allen, can you confirm this has been addressed by PR #141 ?

allenjbaum:
Given my comments- I think so, but it’s been quite a while…

PR #141, opened 2022-01-13
TOR pmp matching with same hi and lo pmpaddr

bilalsakhawat-10xe:
Fix for the issue #117.

### DIFF
--- a/model/riscv_pmp_control.sail
+++ b/model/riscv_pmp_control.sail
@@ -122,7 +122,7 @@ enum pmpAddrMatch = {PMP_NoMatch, PMP_PartialMatch, PMP_Match}
 function pmpMatchAddr(addr: xlenbits, width: xlenbits, rng: pmp_addr_range) -> pmpAddrMatch = {
   match rng {
     None()         => PMP_NoMatch,
-    Some((lo, hi)) => if   hi <_u lo   /* to handle mis-configuration */
+    Some((lo, hi)) => if   hi <=_u lo   /* to handle mis-configuration */
                       then PMP_NoMatch
                       else {
                         if      (addr + width <=_u lo) | (hi <=_u addr)


scottj97 (approved):

martinberger (commented):
LGTM

jrtc27:
This seems correct to me. I was trying to think if there's a nicer way to express this such that it's not lumped in with the `hi < lo` case which is more of a mis-configuration than the pointless (but could legitimately occur in real code if it happens to end up with some size variable set to 0 in an edge case) zero-length entry, but I don't think so, because this really is a special case where we want something different to what a standard 1D intersection algorithm would give.

allenjbaum:
The other part of this, which I was not aware of,  is that if a access
straddles 2 PMP regions,
then it should not trap with an illegal access unless one of them doesn't
have the correct permissions for that region.
I thought it always trapped if an access wasn't confined within a region.:

The lowest-numbered PMP entry that matches any byte of an access determines
whether that access succeeds or fails.

The confusion is what "access" means. If the hart splits unaligned
load/store into two separate accesses that does not cross a 4B boundary,
then the above statement applies to each access separately.
This allows misaligned accesses to be emulated byte by byte if not
supported in HW.
Note that it can still trap as a misaligned access, but not an illegal
access if permission checks pass.
You could even get the case where the first half gets a misaligned
exception and the second gets an illegal exception.

The exception (sic) to this is if an access is aligned to its size, AND
straddles 2 regions,
  (only happens with 4B granularity & 8B access or CMO zero operation, I
think)
because that's supposed to be atomic - and even there it could be
implemented as such,
and if the granularity of regions is implemented to be more than a cache
line - it probably can't happen.

xiak95:
I think this can be merged and the issue closed ?

allenjbaum:
Agreed. There are other corner cases we will need to address, but I think this one is solid. 
If we didn't accept it, then this would take priority over any following entries, even though this one would fail for another reason (all bytes of the request would be outside the zero length region). 
The number of corner cases in the PMP definition is pretty astounding.

ptomsich:
I see one approval and two LGTMs (counting Allen's comment), so I am merging this.
Please raise any follow-on issues an new tickets and file new PRs.

PR #153, opened 2022-02-14
Add support for Zhinx extension

bilalsakhawat:
Added support for Zhinx extension ([spec](https://github.com/riscv/riscv-zfinx/blob/main/zfinx-1.0.0.pdf))

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -270,6 +270,15 @@ function wF_bits(i: bits(5), data: flenbits) -> unit = {
 
 overload F = {rF_bits, wF_bits, rF, wF}
 
+val rF_or_X_H : bits(5) -> bits(16) effect {escape, rreg}
+function rF_or_X_H(i) = {
+  assert(sizeof(flen) >= 16);
+  assert(sys_enable_fdext() != sys_enable_zfinx());
+  if   sys_enable_fdext()
+  then nan_unbox(F(i))
+  else X(i)[15..0]
+}
+
 val rF_or_X_S : bits(5) -> bits(32) effect {escape, rreg}
 function rF_or_X_S(i) = {
   assert(sizeof(flen) >= 32);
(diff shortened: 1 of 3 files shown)


abukharmeh:
SPEC link it outdate AFAIK, I think this is latest https://github.com/riscv/riscv-zfinx/blob/main/zfinx-1.0.0.pdf

bilalsakhawat:
Hi! @jrtc27 @martinberger 
Can you please review it.

liweiwei90:
Zfinx_Zfh is used in the old spec. It's called Zhinx in the new spec.

model/riscv_fdext_regs.sail
```diff
@@ -270,6 +270,15 @@ function wF_bits(i: bits(5), data: flenbits) -> unit = {
 
 overload F = {rF_bits, wF_bits, rF, wF}
 
+val rF_or_X_H : bits(5) -> bits(16) effect {escape, rreg}
+function rF_or_X_H(i) = {
+  assert(sizeof(flen) >= 32);
```
jrtc27, line 275: As far as this function's concerned it only needs to be 16 (the assert is so that the function always type-checks even if F registers are too small for nan_unbox of them to be able to yield a 16-bit value). Arguably that's not necessary for this or for the _S version due to the architecture not having a way to have floating-point with FLEN < 32 (Zfh requires F) but it's there for completeness in case it's ever needed, and definitely is needed in the _D case for F-without-D systems.

model/riscv_fdext_regs.sail
```diff
@@ -293,6 +302,15 @@ function rF_or_X_D(i) = {
   }
 }
 
+val wF_or_X_H : (bits(5), bits(16)) -> unit effect {escape, wreg}
+function wF_or_X_H(i, data) = {
+  assert(sizeof(flen) >= 32);
```
jrtc27, line 307: Ditto

jrtc27:
(i.e. please change the `assert(sizeof(flen) >= 32);` lines to `assert(sizeof(flen) >= 16);`)

bilalsakhawat:
> (i.e. please change the `assert(sizeof(flen) >= 32);` lines to `assert(sizeof(flen) >= 16);`)

Done. Thanks!!

PR #157, opened 2022-03-29
Added missing effects for RVFI builds

marnovandermaas:
This is required for building sail-riscv with RVFI using the current version of the sail compiler after this commit: https://github.com/rems-project/sail/commit/adcd56e8baa4f6424f4115af2c8f88036b7d400f
The reason is that -O is only used for the C emulator in sail-riscv's Makefile and this flag no longer implies no effects.

This is an example TypeError that I got when compiling without these changes:
```
Type error:
[model/rvfi_dii.sail]:264:0-275:1
264 |function rvfi_zero_exec_packet () = {
    |^------------------------------------
275 |}
    |^
    | Effects do not match: {wreg} declared and {rreg, wreg} found
    |
```

You can replicate the error using the following command:
`ARCH=RV32 make c_emulator/riscv_rvfi_RV32`

### DIFF
--- a/model/riscv_vmem_sv32.sail
+++ b/model/riscv_vmem_sv32.sail
@@ -74,7 +74,11 @@
 
 function to_phys_addr(a : paddr32) -> xlenbits = a[31..0]
 
+$ifdef RVFI_DII
+val walk32 : (vaddr32, AccessType(ext_access_type), Privilege, bool, bool, paddr32, nat, bool, ext_ptw) -> PTW_Result(paddr32, SV32_PTE) effect {rmem, rmemt, rreg, wreg, escape}
+$else
 val walk32 : (vaddr32, AccessType(ext_access_type), Privilege, bool, bool, paddr32, nat, bool, ext_ptw) -> PTW_Result(paddr32, SV32_PTE) effect {rmem, rmemt, rreg, escape}
+$endif
 function walk32(vaddr, ac, priv, mxr, do_sum, ptb, level, global, ext_ptw) = {
   let va = Mk_SV32_Vaddr(vaddr);
   let pt_ofs : paddr32 = shiftl(EXTZ(shiftr(va.VPNi(), (level * SV32_LEVEL_BITS))[(SV32_LEVEL_BITS - 1) .. 0]),
--- a/model/riscv_mem.sail
+++ b/model/riscv_mem.sail
@@ -154,7 +154,7 @@ function pmp_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext_
 /* Atomic accesses can be done to MMIO regions, e.g. in kernel access to device registers. */
 
 $ifdef RVFI_DII
-val rvfi_read : forall 'n, 'n > 0. (xlenbits, atom('n), MemoryOpResult((bits(8 * 'n), mem_meta))) -> unit effect {wreg}
+val rvfi_read : forall 'n, 'n > 0. (xlenbits, atom('n), MemoryOpResult((bits(8 * 'n), mem_meta))) -> unit effect {escape, wreg}
 function rvfi_read (addr, width, result) = {
   rvfi_mem_data->rvfi_mem_addr() = EXTZ(addr);
   rvfi_mem_data_present = true;
@@ -228,7 +228,7 @@ function mem_write_ea (addr, width, aq, rl, con) = {
 }
 
 $ifdef RVFI_DII
-val rvfi_write : forall 'n, 0 < 'n <= max_mem_access . (xlenbits, atom('n), bits(8 * 'n), mem_meta) -> unit effect {wreg}
+val rvfi_write : forall 'n, 0 < 'n <= max_mem_access . (xlenbits, atom('n), bits(8 * 'n), mem_meta) -> unit effect {escape, wreg}
 function rvfi_write (addr, width, value, meta) = {
   rvfi_mem_data->rvfi_mem_addr() = EXTZ(addr);
   rvfi_mem_data_present = true;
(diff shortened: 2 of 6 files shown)


martinberger (commented):
Looks fine.

martinberger:
> This is required for building sail-riscv with version 0.14 of the sail compiler.

Given that 0.14 is from January 2021, I'm surprised that this didn't cause a problem before.

marnovandermaas:
TestRIG uses the `c_emulator/riscv_rvfi_$(ARCH)` target in sail-riscv, so maybe this is why it remained unnoticed?

model/riscv_vmem_sv32.sail
```diff
@@ -74,7 +74,7 @@
 
 function to_phys_addr(a : paddr32) -> xlenbits = a[31..0]
 
-val walk32 : (vaddr32, AccessType(ext_access_type), Privilege, bool, bool, paddr32, nat, bool, ext_ptw) -> PTW_Result(paddr32, SV32_PTE) effect {rmem, rmemt, rreg, escape}
+val walk32 : (vaddr32, AccessType(ext_access_type), Privilege, bool, bool, paddr32, nat, bool, ext_ptw) -> PTW_Result(paddr32, SV32_PTE) effect {rmem, rmemt, rreg, wreg, escape}
```
jrtc27, line 77: This is dodgy. Architecturally page table walks do not write to registers. It's only because as a side-effect they stash state for RVFI to later report.

jrtc27:
test/run_tests.sh builds the RV32 and RV64 RVFI-DII binaries in CI so I don't know why it's only broken for you

jrtc27:
(using the 0.14 Sail from opam)

model/riscv_vmem_sv32.sail
marnovandermaas: Similar is true for the walk39 and walk48 in that case

jrtc27: Yes, I only commented once on it

marnovandermaas:
TestRIG CI is actually using the current version of sail in the repo: https://github.com/rems-project/sail
So not version 0.14 (I will update the pull request message to correct it).
I'm assuming you want sail-riscv to be able to be built with the current version of sail as well right? Or do you only start caring once it has been tagged and released?

model/riscv_vmem_sv32.sail
PeterRugg: Hmm, that's a pain: is there a way to guard the effect with an ifdef of whether rvfi is enabled?

jrtc27:
> TestRIG CI is actually using the current version of sail in the repo: https://github.com/rems-project/sail
> So not version 0.14 (I will update the pull request message to correct it).
> I'm assuming you want sail-riscv to be able to be built with the current version of sail as well right? Or do you only start caring once it has been tagged and released?

Ok, then I assume this is https://github.com/rems-project/sail/commit/adcd56e8baa4f6424f4115af2c8f88036b7d400f; note that -O is only used for the C emulator in sail-riscv's Makefile.

Please update the commit message to be correct.

model/riscv_vmem_sv32.sail
PeterRugg: Although it looks to me like the first function we call `mem_read_priv` has `wreg`, so (assuming you have to inherit the effects of callees), there is currently a bug here.

martinberger: Isn't there a danger that `$ifdef`-ed signatures spread throughout the code base?

jrtc27:
Please squash commits into one reworded commit rather than piling on new ones

marnovandermaas:
Let me know if I can do anything else before this is ready to be up-streamed.

jrtc27 (approved):

marnovandermaas:
@rmn30 would you be able to take a look and merge this PR?

rmn30:
I don't have a problem with the changes but note a few things:
1) The way effects are handled has changed in the recently released [Sail 0.15](https://github.com/rems-project/sail/releases/tag/0.15) and explicit effect annotations are now deprecated. Declarations of external functions should now be declared as either `monadic` or `pure` so we should add these in prelude.sail. Unfortunately adding these is not backwards compatible so upgrading Sail will probably require a flag day. Maybe we should start a sail 0.15 branch? Meanwhile in Sail 0.14 there is a `-no_effects` option that behaves more like Sail 0.15 so it might make more sense to use that?
2) I'm not sure who has the authority to merge PRs in this repo. I have the github permission but I'm not sure whether I should use it. Perhaps @martinberger has an opinion?

bacam:
The closest equivalent to 0.15's behaviour in 0.14 is the `-infer_effects` option.

rmn30:
I guess there is probably no harm in merging this PR but given the above I would be more inclined to try out `-infer-effects` and start preparing for Sail 0.15.

martinberger:
>  Perhaps @martinberger has an opinion?

The question of how to guide the coevolution of Sail and the RISCV model is interesting and important.

I think we should strive to keep both in sync, if this is possible at all. This probably needs a bit of structure.

- Sail team should ideally inform RISCV of anticipated breaking changes.
- It would be ideal to think about tool support.
- How far behind the bleeding edge should the RISCV golden model be?
- Who is responsible for adapting older	Sail?
- How do we validate updated Sail?

I think	Robert's suggestion to have a bleeding edge branch of the model	that is	used to	update the model is a good starting point. I think this needs a bit of though and planning, otherwise we might end up in a mess!

marnovandermaas:
> I guess there is probably no harm in merging this PR but given the above I would be more inclined to try out `-infer-effects` and start preparing for Sail 0.15.

Ok, I'll close this PR because infer effects seems like a better option. For the further discussion on how to handle the upgrade to Sail 0.15 it's probably better to open a separate issue.

PR #132, opened 2021-11-25
Sail Model for RISC-V P Extension

njru8cjo:
Hello, we’re the team from Programming Langauge Lab, NTHU (National Tsing-Hua University), Taiwan, Academia Sinica(Taiwan), and Andes Technology Corporation. Our work entitled "Sail Specification for RISC-V P-Extension" related to [RISC-V P extension](https://github.com/riscv/riscv-p-spec/blob/master/P-ext-proposal.adoc)  will be presented in RISC-V Summit, Dec, 2021. The link is [here](https://riscvsummit2021.sched.com/event/nfFe?iframe=no).  

This PR specifies sail models for all instructions in the [P-Extension Proposal](https://github.com/riscv/riscv-p-spec/blob/master/P-ext-proposal.adoc).

PR #158, opened 2022-03-31
Update with Sail Model for RISC-V P extension 

umcann123:
Hello, we are the team from PR #132. We reorganized 327 RISC-V P extension instructions into 5 files and moved all auxiliary into riscv_pext_insts_prelude.sail. We have an introduction about how we group instructions [here](https://hackmd.io/vFSNmkJoQb-R3Kck8GmrBg?view). All instructions passed coverage test of [RISC-V ISAC](https://github.com/riscv-software-src/riscv-isac) and [RISC-V ctg](https://github.com/riscv-software-src/riscv-ctg) and all the coverage is 100% and here is the [report](https://github.com/umcann123/Utility/blob/main/coverage.html). Kindly review this request and suggest changes. Thanks!


allenjbaum:
could you point me at the coverage yaml file used to generate tests?

allenjbaum:
I'd like to see a link to the .cgf files, not just their names.
What repo are they located in? It's not where I would expect it:
https://github.com/riscv/riscv-p-spec

umcann123:
@allenjbaum 
Here is [rv32ip.cgf](https://github.com/riscv-software-src/riscv-ctg/blob/master/sample_cgfs/rv32ip.cgf)
and [rv64ip.cgf](https://github.com/riscv-software-src/riscv-ctg/blob/master/sample_cgfs/rv64ip.cgf)

billmcspadden-riscv:
There has been a counter-proposal for the definition of the P extension.  As such,  this PR will not be merged until a frozen spec has been reached.

martinberger:
> this PR will not be merged until a frozen spec has been reached.

Maybe it makes sense to close the PR for now, until the situation with P is clear. It can always be reopened, if that makes sense.

PR #172, opened 2022-07-19
Minimal updates for Coq proof assistant output

bacam:
This small set of changes to the Coq support files allows recent versions of Coq to build them (by replacing omega with lia), adds a missing axiom for a platform built-in, and improves the namespacing a little.

(Note that the Coq output won't build against the library from the last Sail release due to a bug in that library, but will build using the current repository version of Sail.)


bacam:
These changes shouldn't affect anything other than the Coq output, and I doubt anyone else is terribly interested in reviewing them, so if there are no objections I'll merge it in a few days time.

PR #177, opened 2022-08-17
Adjust sed arguments for compatibility with BSD sed

bacam:
We don't need the generated backup files, but it appears to be the only
way to be compatible with both GNU and BSD sed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -278,10 +278,12 @@ rvfi_preserve_fns=-c_preserve rvfi_set_instr_packet \
   -c_preserve print_instr_packet \
   -c_preserve print_rvfi_exec
 
+# sed -i isn't posix compliant, unfortunately
 generated_definitions/c/riscv_rvfi_model_$(ARCH).c: $(SAIL_RVFI_SRCS) model/main.sail Makefile
 	mkdir -p generated_definitions/c
 	$(SAIL) $(rvfi_preserve_fns) $(SAIL_FLAGS) -O -Oconstant_fold -memo_z3 -c -c_include riscv_prelude.h -c_include riscv_platform.h -c_no_main $(SAIL_RVFI_SRCS) model/main.sail -o $(basename $@)
-	sed -i -e '/^[[:space:]]*$$/d' $@
+	sed -e '/^[[:space:]]*$$/d' $@ > $@.new
+	mv $@.new $@
 
 c_emulator/riscv_rvfi_$(ARCH): generated_definitions/c/riscv_rvfi_model_$(ARCH).c $(C_INCS) $(C_SRCS) $(SOFTFLOAT_LIBS) Makefile
 	gcc -g $(C_WARNINGS) $(C_FLAGS) $< -DRVFI_DII $(C_SRCS) $(SAIL_LIB_DIR)/*.c $(C_LIBS) -o $@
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -276,7 +276,7 @@ rvfi_preserve_fns=-c_preserve rvfi_set_instr_packet \
 generated_definitions/c/riscv_rvfi_model_$(ARCH).c: $(SAIL_RVFI_SRCS) model/main.sail Makefile
 	mkdir -p generated_definitions/c
 	$(SAIL) $(rvfi_preserve_fns) $(SAIL_FLAGS) -O -Oconstant_fold -memo_z3 -c -c_include riscv_prelude.h -c_include riscv_platform.h -c_no_main $(SAIL_RVFI_SRCS) model/main.sail -o $(basename $@)
-	sed -i -e '/^[[:space:]]*$$/d' $@
+	sed -i.bak -e '/^[[:space:]]*$$/d' $@
```
arichardson, line 279:
```suggestion
	sed -i '' -e '/^[[:space:]]*$$/d' $@
```
`If a zero-length extension is given, no backup will be saved.`

arichardson (changes requested):

Makefile
Alasdair: I would double check if `-i ''` works with GNU sed, I think it may not `-i''` would without the space I think

bacam: They're mutually incompatible, unfortunately.

scottj97:
If you don't need the generated backup files, why not just remove the `-i` completely?

abukharmeh:
> If you don't need the generated backup files, why not just remove the `-i` completely?

image

scottj97:
Right, sorry. Brain fart.

Makefile
rmn30: Perhaps you could use `perl` to do the same thing? I guess most systems will have it and it might be more portable?

jrtc27:
```
sed -e '/^[[:space:]]*$$/d' $@ > $@.new
mv $@.new $@
```
is a common portable way to get around the -i importability

rmn30: On second thoughts using sed and having the backups is probably best.

bacam: @jrtc27 's method is the only true POSIX way... in any case, this discussion has convinced me that a comment is necessary to avoid future pain.

allenjbaum: I agree.

ptomsich:
While it looks good to me, we have 5 reviewers already assigned ... anyone willing to *approve*?

ptomsich (changes requested):
Please implement Jessica's (completely POSIX-compliant) suggestion.

bacam:
I've redone it in the redirect and move style, on top of the current HEAD.

bacam:
It would be useful if someone with a BSDish system could test it.

ptomsich:
@jrtc27 Could you test on one of your BSD systems?

ptomsich (approved):

bacam:
Rebased because the last one became messed up somehow (maybe I did it against the wrong branch?).  Also tested it on a CheriBSD machine.

jrtc27 (approved):

PR #179, opened 2022-08-22
Fix build without -fcommon

arichardson:
The sail C code generator will emit definions for these structs. The
duplicate definition linker error were worked around by adding -fcommon
in https://github.com/riscv/sail-riscv/commit/ffea7a39c32a210a446379aeda0eabcec4918ed6. This commit fixes the
underlying issue by declaring the variables as `extern`.

### DIFF
--- a/c_emulator/riscv_sail.h
+++ b/c_emulator/riscv_sail.h
@@ -7,7 +7,7 @@ typedef int unit;
 typedef uint64_t mach_bits;
 
 struct zMisa {mach_bits zMisa_chunk_0;};
-struct zMisa zmisa;
+extern struct zMisa zmisa;
 
 void model_init(void);
 void model_fini(void);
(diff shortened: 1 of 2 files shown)


c_emulator/riscv_sail.h
```diff
@@ -67,6 +67,7 @@ extern mach_bits zsepc, zstval;
 extern mach_bits zfloat_result, zfloat_fflags;
 
 struct zMcause {mach_bits zMcause_chunk_0;};
-struct zMcause zmcause, zscause;
```
jrtc27, line 70: Why split?

arichardson: Multiple declarations on one line are generally considered bad style, so I thought I'd split it. Can of course change it back to what it was.

jrtc27 (approved):

arichardson:
Would it be possible for someone with write access to merge this trivial fix?

PR #185, opened 2022-10-12
Fix packw sign-extension

janweinstock:
Currently, the result of packw is zero-extended, but according to spec, packw should sign-extend its result to 64bit.

This sign-extension is also in accordance with the other simulators:
Spike: https://github.com/riscv-software-src/riscv-isa-sim/blob/86d9fe49eda1fff863a43e682015216a25cc72f3/riscv/insns/packw.h#L10
QEMU: https://gitlab.com/qemu-project/qemu/-/blob/master/target/riscv/insn_trans/trans_rvb.c.inc#L540

Signed-off-by: Jan Henrik Weinstock <jan@mwa.re>

### DIFF
--- a/model/riscv_insts_zbkb.sail
+++ b/model/riscv_insts_zbkb.sail
@@ -40,7 +40,7 @@ function clause execute (ZBKB_PACKW(rs2, rs1, rd)) = {
   let rs1_val = X(rs1);
   let rs2_val = X(rs2);
   let result : bits(32) = rs2_val[15..0] @ rs1_val[15..0];
-  X(rd) = EXTZ(result);
+  X(rd) = EXTS(result);
   RETIRE_SUCCESS
 }
 


marcfedorow:
That is fun how spike struggled of the same issues (i.e. [this](https://github.com/riscv-software-src/riscv-isa-sim/pull/685)) and for somewhat reason people decided to move to another simulator, and it has the same (type and amount of) bugs in it.
In spike they at least merged pull requests fast enough.

allenjbaum:
I'm more concerned that the ACTs didn't catch this. Before tests are
merged, they're supposed to be compared between Spike and Sail.
Looking at the tests, I see several that should have sign extended with
ones (e.g. inst_50, inst_58) and would have mismatched if Sail didn't sign
extend and Spike did.
So, I'm confused as to what went wrong here.

jrtc27:
This is clearly a correct fix, can we please just merge this? Again I'm happy to do so but don't want to be wielding my collaborator powers if someone with official authority should be approving changes.

marcfedorow:
I do love how, instead of just accepting a perfectly legit easily checkable one line change, nothing is done in over two months.

jrtc27:
Please stop with such unconstructive comments, this isn't the place for those kinds of remarks.

marcfedorow:
The constructive part is — process of reviewing of this PRs should be somehow changed.
E.g. arch tests rely on SAIL, and SAIL is capable of problems like this that are not fixed fast enough (or anywhere fast).
This is an actual problem, not just inconstructive comments.
If you have some guidelines of what proper actions could be performed to trigger pending reviews, I kindly ask to share them.

scottj97:
> The constructive part is — process of reviewing of this PRs should be somehow changed.

A pull request for a bug fix is not the place to discuss this. Move along.

martinberger (approved):

PR #186, opened 2022-10-12
Fix xperm4 index calculation

janweinstock:
This seems to be a copy&paste issue from xperm8. 

This fix brings `sail_riscv` in line with spike for the xperm4 instruction semantics.

Signed-off-by: Jan Henrik Weinstock <jan@mwa.re>


marcfedorow:
What is the estimation of how long could it take to review this pull request?
Arch-tests rely on SAIL simulator but it does not allow to test zbkx using the stable unmodified version.

PR #190, opened 2022-11-04
Fix handling of mstatus.MIE in M-mode only case.

rmn30:
Previously if neither user or supervisor mode are present dispatchInterrupt ignored mstatus.MIE leading to an infinite interrupt loop.

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -336,23 +336,22 @@ function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
  * handled (if any), and the privilege it should be handled at.
  */
 function dispatchInterrupt(priv : Privilege) -> option((InterruptType, Privilege)) = {
-  /* If we don't have different privilege levels, we don't need to check delegation.
-   * Absence of U-mode implies absence of S-mode.
-   */
-  if (~ (haveUsrMode())) | ((~ (haveSupMode())) & (~ (haveNExt()))) then {
-    assert(priv == Machine, "invalid current privilege");
-    let enabled_pending = mip.bits() & mie.bits();
-    match findPendingInterrupt(enabled_pending) {
-      Some(i) => let r = (i, Machine) in Some(r),
-      None()  => None()
-    }
-  } else {
+  if (haveSupMode() | haveNExt()) then {
     match getPendingSet(priv) {
       None()      => None(),
       Some(ip, p) => match findPendingInterrupt(ip) {
                        None()  => None(),
                        Some(i) => let r = (i, p) in Some(r)
                      }
+    }
+  } else {
+    /* If we don't have different privilege levels, we don't need to check
+     * delegation.
+     */
+    let enabled_pending = mip.bits() & mie.bits();
+    match (mstatus.MIE(), findPendingInterrupt(enabled_pending)) {
+      (0b1, Some(i)) => let r = (i, Machine) in Some(r),
+      (_, _)         => None()
     }
   }
 }


rmn30:
Any objections to this?

martinberger (commented):
It's purely aesthetic, but the additional conditional could be done in the match statement.

      match (mstatus.MIE(), findPendingInterrupt(mip.bits() & mie.bits())) {
         (0b1, Some(i)) => let ...
         (_, _)         => None()

I find this more readable, but *de gustibus non est disputandum* ...

rmn30:
Yes, that is more elegant. I will push an alternative version.

model/riscv_sys_control.sail
```diff
@@ -342,9 +342,9 @@ function dispatchInterrupt(priv : Privilege) -> option((InterruptType, Privilege
   if (~ (haveUsrMode())) | ((~ (haveSupMode())) & (~ (haveNExt()))) then {
     assert(priv == Machine, "invalid current privilege");
```
rmn30, line 343: Looking at this again this assert looks suspect. If `haveUsrMode()` is `true` but `haveSupMode()` and `haveNExt()` are false this could be reachable with `priv == User`. I think the second half of the OR in the condition should go away.

rmn30: A comment from @pmundkur would be ideal.

rmn30: I think I've figured it out: the condition is checking whether it is possible for any interrupts to be delegated to a mode other than Machine mode. This makes sense as it then uses `mip` and `mie` directly to compute the pending interrupts rather than calling `getPendingSet` (which considers delegation). This means we need to fix or remove the assertion. I think probably we should remove it as I can't think of much to put there that isn't just a rehash of the preciding if condition.

On the subject of that condition I think it could also be expressed as `~(haveSupMode()) | (haveUsrMode() & ~haveNExt())` which might more clearly express the intent?

rmn30: > On the subject of that condition I think it could also be expressed as ~(haveSupMode()) | (haveUsrMode() & ~haveNExt()) which might more clearly express the intent?

The above is incorrect but a more concise version might be `~(haveSupMode() | haveNExt())`?

allenjbaum:
That does sound like the correct version to determine that an interrupt
cannot be delegated, irrespective of delegation bits and enabling

rmn30:
OK, I'll revise this again to remove the assertion. I'll also flip the if so we don't need to invert the condition.

scottj97:
Please correct the typo (msatus) in the commit message and PR title.

model/riscv_sys_control.sail
```diff
@@ -336,24 +336,23 @@ function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
...
                        Some(i) => let r = (i, p) in Some(r)
                      }
     }
+  } else {
+    /* If we don't have different privilege levels, we don't need to check
+     * delegation.
+     */
+    let enabled_pending = mip.bits() & mie.bits();
+    match (mstatus.MIE(), findPendingInterrupt(enabled_pending)) {
+      (0b1, Some(i)) => let r = (i, Machine) in Some(r),
```
Timmmm, line 353: This can just be `=> Some((i, Machine)),`

model/riscv_sys_control.sail
```diff
@@ -336,24 +336,23 @@ function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
...
-   */
-  if (~ (haveUsrMode())) | ((~ (haveSupMode())) & (~ (haveNExt()))) then {
-    assert(priv == Machine, "invalid current privilege");
-    let enabled_pending = mip.bits() & mie.bits();
-    match findPendingInterrupt(enabled_pending) {
-      Some(i) => let r = (i, Machine) in Some(r),
-      None()  => None()
-    }
-  } else {
+  if (haveSupMode() | haveNExt()) then {
```
Timmmm, line 339: Don't need brackets around this.

Timmmm (commented):
I'm just going through old PRs. I can update this if you like?

Also this whole check is only a performance optimisation. Maybe it's better just to remove it (it probably doesn't make too much difference because it's all `xlenbits` so GMP won't get involved).

Timmmm:
> Also this whole check is only a performance optimisation. Maybe it's better just to remove it (it probably doesn't make too much difference because it's all xlenbits so GMP won't get involved).

We ended up doing this so the behaviour should be correct now. Please open an issue if it isn't!

Issue #194, opened 2022-11-30
Build warnings in model/riscv_pmp_regs.sail


ptomsich:
Building 4d05aa1698a0003a4f6f99e1380c743711c32052 from source outputs warnings for incomplete matches in model/riscv_pmp_regs.sail:
```
Warning: Possible incomplete pattern match at [model/riscv_pmp_regs.sail]:144:7-12
144 |  then match n {
    |       ^---^
    | 
Most general matched pattern is 0

Warning: Possible incomplete pattern match at [model/riscv_pmp_regs.sail]:150:7-12
150 |  else match n { //  sizeof(xlen) == 64
    |       ^---^
    | 
Most general matched pattern is 0

Warning: Possible incomplete pattern match at [model/riscv_pmp_regs.sail]:170:7-12
170 |  then match n {
    |       ^---^
    | 
Most general matched pattern is 0

Warning: Possible incomplete pattern match at [model/riscv_pmp_regs.sail]:193:7-12
193 |  then match n {
    |       ^---^
    | 
Most general matched pattern is 0

Warning: Possible incomplete pattern match at [model/riscv_pmp_regs.sail]:144:7-12
144 |  then match n {
    |       ^---^
    | 
Most general matched pattern is 0

Warning: Possible incomplete pattern match at [model/riscv_pmp_regs.sail]:150:7-12
150 |  else match n { //  sizeof(xlen) == 64
    |       ^---^
    | 
Most general matched pattern is 0

Warning: Possible incomplete pattern match at [model/riscv_pmp_regs.sail]:170:7-12
170 |  then match n {
    |       ^---^
    | 
Most general matched pattern is 0

Warning: Possible incomplete pattern match at [model/riscv_pmp_regs.sail]:193:7-12
193 |  then match n {
    |       ^---^
    | 
Most general matched pattern is 0
```

ptomsich:
/cc @pmundkur 

Let's cc someone with knowledge of the affected lines (my best guess from looking at a `git blame`)...

jrtc27:
This is a long-standing issue that’s been there for years before the model was adopted as official

Alasdair:
I have a new release of Sail that will hopefully be merged into opam soon: https://github.com/ocaml/opam-repository/pull/22553

It has a re-written pattern completeness checker that should be much more accurate.

PR #197, opened 2022-12-21
Add wildcard cases to matches to suppress Sail warnings.

rmn30:
Sail tries to check for pattern match completeness and issues warnings but this often gives false positives: see discussion at https://github.com/rems-project/sail/issues/191 .  To suppress these we add a wildcard case that raises an internal error.  There is effectively no behaviour change as these would previously have resulted in a match error at runtime and they should be unreachable anyway.

At the same time we change the DOUBLE case in memory access paths to allow for xlen > 64.


arichardson (approved):
Commit message should include
`Fixes: https://github.com/riscv/sail-riscv/issues/194` to autoclose that issue.

model/riscv_pmp_regs.sail
```diff
@@ -187,9 +189,10 @@ function pmpWriteCfgReg(n, v) = {
                 pmp13cfg = pmpWriteCfg(pmp13cfg, v[15..8]);
                 pmp14cfg = pmpWriteCfg(pmp14cfg, v[23..16]);
                 pmp15cfg = pmpWriteCfg(pmp15cfg, v[31..24]);
-              }
+              },
+         _ => internal_error("Unexpected pmp config reg write")
        }
-  else if sizeof(xlen) == 64
+  else if sizeof(xlen) >= 64
```
xiak95, line 195: I cannot see how this matter in the current model since XLEN is never configured >64 and changing it here might give confidance that its correctly implemented else where ?

model/riscv_types.sail
```diff
@@ -128,6 +128,23 @@ function arch_to_bits(a : Architecture) -> arch_xlen =
...
+  Error_not_implemented : string,
+  Error_internal_error  : unit
+}
+
+val not_implemented : forall ('a : Type). string -> 'a effect {escape}
+function not_implemented message = throw(Error_not_implemented(message))
+
+val internal_error : forall ('a : Type). string -> 'a effect {escape}
+function internal_error(s) = {
+    assert (false, s);
```
xiak95, line 144: I am not very famillar with SAIL assert implementation, would this print call stack ? otherwise it might be helpfull to print  __FILE__ and __LINE__  as requested in this issue https://github.com/rems-project/sail/issues/175 ?

model/riscv_pmp_regs.sail
rmn30: It doesn't matter at the moment but RV128 is in the spec. and at some point the model might have to support it.

model/riscv_insts_base.sail
```diff
@@ -394,15 +394,16 @@ function clause execute(LOAD(imm, rs1, rd, is_unsigned, width, aq, rl)) = {
...
+          match (width) {
+            BYTE =>
+              process_load(rd, vaddr, mem_read(Read(Data), paddr, 1, aq, rl, false), is_unsigned),
+            HALF =>
+              process_load(rd, vaddr, mem_read(Read(Data), paddr, 2, aq, rl, false), is_unsigned),
+            WORD =>
+              process_load(rd, vaddr, mem_read(Read(Data), paddr, 4, aq, rl, false), is_unsigned),
+            DOUBLE if sizeof(xlen) >= 64 =>
+              process_load(rd, vaddr, mem_read(Read(Data), paddr, 8, aq, rl, false), is_unsigned),
+            _ => internal_error("Unexpected load width")
```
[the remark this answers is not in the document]
rmn30: That's true. Hopefully it should be pretty obvious what has gone wrong, but I could make the message more specific.

model/riscv_types.sail
rmn30: No, unfortunately it will just print a message and `exit(1)`. You'd need a debugger to get a stack trace.

rmn30: __FILE__ and __LINE__ would have to be passed to this function as arguments to be useful. I think Sail might have equivalent macros but all calls to `internal_error` would need updated.

arichardson:
What is the plan for moving this PR forwards? The warnings printed are rather annoying and make it difficult to see whether changes that I've made have introduced new warnings.

model/riscv_types.sail
rmn30: Given that the `internal_error` function is already in use and cannot usefully display the file and line number or stack trace can we just ensure that the assert messages are sufficiently different and detailed?

rmn30:
I could push a version with more detailed error messages and revised commit message if that would please everyone?

model/riscv_types.sail
martinberger: There is only a small number of calls to internal_error (about 40), so it would be easy to update. Right now, some of them have identical error messages (eg. "invalid physical address in TLB"). That's suboptimal.

martinberger (approved):
Large numbers of Sail warnings are annoying. (We changed the Sail compiler at my previous employer to deal with this). Adding wildcard cases is a quick fix. In normal programming, wildcard cases for pattern matches are often considered an anti-pattern since they tend mask problems. In our case the wildcard case terminates the program, so cannot be ignored and cannot mask a problem. In the long term it would be better for the Sail compiler to be smarter wrt pattern analysis. For now, wildcards as proposed are OK.

arichardson:
See https://github.com/rems-project/sail/issues/191. Rust and OCaml apparently have the same "bug", so it's not unexpected.

martinberger:
If pattern matching has arbitrary pattern guards, the problem becomes undecidable.

rmn30:
I've run into a Sail bug while trying to change the type of `internal_error`. Will report there then get back to this.

martinberger (approved):
Adding line/file is good. Thanks.

rmn30:
The Sail bug is now fixed upstream but we probably shouldn't merge this until it is released on opam.

martinberger:
> The Sail bug is now fixed upstream but we probably shouldn't merge this until it is released on opam.

Agreed. Do we have an idea when opam will push this? It would be good to have this soon!

arichardson:
> The Sail bug is now fixed upstream but we probably shouldn't merge this until it is released on opam.

A related question here: What is the minimum supported sail version for the model? Can we assume that it requires the latest release? In that case we could clean up all the effects annotations. Maybe there should be some version somewhere rather than relying on code failing to build on older versions?

martinberger:
>  *Can we assume that it requires the latest release?*

This is an important issue. Given that we have > 30 extensions coming, I think we should have a clear idea about moving to more recent versions of Sail  (and keeping up in the future).

rmn30:
> Agreed. Do we have an idea when opam will push this? It would be good to have this soon!

@Alasdair has suggested doing a 0.15.1 release soon so hopefully not long (i.e. a week or two).

Regarding the required Sail version, 0.15 has some significant improvements so we should adopt it and fix the warnings. Point releases should only really have bug fixes so should be painless to adopt, although I don't know how strict the Sail team are about this e.g. maintaining stable vs. dev branches. At any rate I will update the commit message to point out the need for potential need for  `opam update && opam upgrade`.

martinberger:
Given the increasing importance of RISCV, I think we should be clear about how the model relates to the evolution of Sail, who's responsible for updating the model etc, and communicate that with an explicit roadmap that is on the model's homepage.

rmn30:
I've just pushed a version with a workaround to avoid the need to update Sail.

arichardson:
Can this be merged? The many lines of warnings printed are rather distracting when working on the model.

arichardson:
ping? The warnings are rather annoying when trying to make changes to the sail model.

xiak95:
Yes, could you please merge this !

martinberger:
Thanks everybody.

PR #200, opened 2023-01-11
Add include required on FreeBSD.

rmn30:


martinberger:
Why is an Internet address manipulation routine needed for the Sail model?

rmn30:
The optional RVFI DII support uses sockets so `sys/socket.h` and `net/inet.h` are already included. Apparently this caused the build to fail on FreeBSD even if RVFI is not enabled.

martinberger (approved):

PR #205, opened 2023-01-19
Increase flexibility of the decode hook (and simplify it)

arichardson:
Previously the decoding hook (`ext_post_decode_hook`) allowed models to override the decoded `ast`. However, this is not sufficient if model extension changes interpretation of fields. Additionally, the assembly printing would always print assembly for the "baseline decode" which resulted in incorrect trace output in the sail-cheri-riscv model.

With the new hook models can implement ext_decode()/ext_decode_compressed() to return an `ast` and for encodings that are not adjusted fall back to the default `encdec`/`encdec_compressed`.

### DIFF
--- a/model/riscv_decode_ext.sail
+++ b/model/riscv_decode_ext.sail
@@ -67,9 +67,12 @@
 /*=======================================================================================*/
 
 /* Extensions may wish to interpose and transform decoded instructions,
- * based on other machine state.  This is supported via a post-decode
- * instruction hook, the default implementation of which is provided below.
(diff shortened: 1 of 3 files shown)


arichardson:
CC: @rmn30 @jrtc27

jrtc27:
This looks suitable to me, and simpler

rmn30:
Yes, this makes sense. The new hook can easily emulate the old behaviour without the disassembly bug. Are we aware of any other extensions apart from CHERI that will need updating?

martinberger (approved):
Looks OK to me.

PR #206, opened 2023-01-20
Add support for the Zicond extension

ptomsich:
This implements the Zicond (conditional integer operations) extension, as of version 1.0-draft-20230120.

The Zicond extension acts as a building block for branchless sequences including conditional-arithmetic, conditional-logic and conditional-select/move.
The following instructions constitute Zicond:
  - czero.eqz rd, rs1, rs2  =>  rd = (rs2 == 0) ? 0 : rs1
  - czero.nez rd, rs1, rs2  =>  rd = (rs2 != 0) ? 0 : rs1

See
  https://github.com/riscv/riscv-zicond/releases/download/v1.0-draft-20230120/riscv-zicond_1.0-draft-20230120.pdf
for the proposed specification and usage details.

### DIFF
--- a/model/riscv_insts_zicond.sail
+++ b/model/riscv_insts_zicond.sail
@@ -16,7 +16,7 @@ mapping clause assembly = ZICOND_RTYPE(rs2, rs1, rd, op)
 function clause execute (ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_EQZ)) = {
   let value = X(rs1);
   let condition = X(rs2);
-  let result : xlenbits = if (condition == zeros()) then zeros()
+  let result : xlenbits = if condition == zeros() then zeros()
 						    else value;
   X(rd) = result;
   RETIRE_SUCCESS
--- a/model/riscv_insts_zicond.sail
+++ b/model/riscv_insts_zicond.sail
@@ -1,0 +1,32 @@
+union clause ast = ZICOND_RTYPE : (regidx, regidx, regidx, zicondop)
+
+mapping clause encdec = ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_EQZ) if haveZicond()
+  <-> 0b0000111 @ rs2 @ rs1 @ 0b101 @ rd @ 0b0110011 if haveZicond()
+mapping clause encdec = ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_NEZ) if haveZicond()
+  <-> 0b0000111 @ rs2 @ rs1 @ 0b111 @ rd @ 0b0110011 if haveZicond()
+
+mapping zicond_mnemonic : zicondop <-> string = {
+  RISCV_CZERO_EQZ <-> "czero.eqz",
+  RISCV_CZERO_NEZ <-> "czero.nez"
+}
+
+mapping clause assembly = ZICOND_RTYPE(rs2, rs1, rd, op)
+  <-> zicond_mnemonic(op) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_EQZ)) = {
+  let value = X(rs1);
+  let condition = X(rs2);
+  let result : xlenbits = if (condition == zeros()) then zeros()
+						    else value;
+  X(rd) = result;
+  RETIRE_SUCCESS
+}
+
+function clause execute (ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_NEZ)) = {
+  let value = X(rs1);
+  let condition = X(rs2);
+  let result : xlenbits = if (condition != zeros()) then zeros()
+						    else value;
+  X(rd) = result;
+  RETIRE_SUCCESS
+}
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -6,7 +6,7 @@
 /*  in the prover_snapshots directory (which include copies of their                     */
 /*  licences), is subject to the BSD two-clause licence below.                           */
 /*                                                                                       */
-/*  Copyright (c) 2017-2021                                                              */
+/*  Copyright (c) 2017-2023                                                              */
 /*    Prashanth Mundkur                                                                  */
 /*    Rishiyur S. Nikhil and Bluespec, Inc.                                              */
 /*    Jon French                                                                         */
@@ -23,6 +23,7 @@
 /*    Microsoft, for contributions by Robert Norton-Wright and Nathaniel Wesley Filardo  */
 /*    Peter Rugg                                                                         */
 /*    Aril Computer Corp., for contributions by Scott Johnson                            */
+/*    VRULL GmbH                                                                         */
 /*                                                                                       */
 /*  All rights reserved.                                                                 */
 /*                                                                                       */
@@ -207,6 +208,9 @@ function haveZkne()   -> bool = true
 function haveZknd()   -> bool = true
 
 function haveZmmul()  -> bool = true
+
+/* ZiCond extension support */
+function haveZicond() -> bool = true
 
 bitfield Mstatush : bits(32) = {
   MBE  : 5,
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -23,7 +23,7 @@
 /*    Microsoft, for contributions by Robert Norton-Wright and Nathaniel Wesley Filardo  */
 /*    Peter Rugg                                                                         */
 /*    Aril Computer Corp., for contributions by Scott Johnson                            */
-/*    VRULL GmbH                                                                         */
+/*    VRULL GmbH, for contributions by Philipp Tomsich                                   */
 /*                                                                                       */
 /*  All rights reserved.                                                                 */
 /*                                                                                       */
@@ -209,7 +209,7 @@ function haveZknd()   -> bool = true
 
 function haveZmmul()  -> bool = true
 
-/* ZiCond extension support */
+/* Zicond extension support */
 function haveZicond() -> bool = true
 
 bitfield Mstatush : bits(32) = {
--- a/model/riscv_types.sail
+++ b/model/riscv_types.sail
@@ -6,7 +6,7 @@
 /*  in the prover_snapshots directory (which include copies of their                     */
 /*  licences), is subject to the BSD two-clause licence below.                           */
 /*                                                                                       */
-/*  Copyright (c) 2017-2021                                                              */
+/*  Copyright (c) 2017-2022                                                              */
 /*    Prashanth Mundkur                                                                  */
 /*    Rishiyur S. Nikhil and Bluespec, Inc.                                              */
 /*    Jon French                                                                         */
@@ -23,6 +23,7 @@
 /*    Microsoft, for contributions by Robert Norton-Wright and Nathaniel Wesley Filardo  */
 /*    Peter Rugg                                                                         */
 /*    Aril Computer Corp., for contributions by Scott Johnson                            */
+/*    VRULL GmbH, for contributions by Philipp Tomsich                                   */
 /*                                                                                       */
 /*  All rights reserved.                                                                 */
 /*                                                                                       */
@@ -408,6 +409,8 @@ enum bropw_zbb = {RISCV_ROLW, RISCV_RORW}
 enum biop_zbs = {RISCV_BCLRI, RISCV_BEXTI, RISCV_BINVI, RISCV_BSETI}
 
 enum extop_zbb = {RISCV_SEXTB, RISCV_SEXTH, RISCV_ZEXTH}
+
+enum zicondop = {RISCV_CZERO_EQZ, RISCV_CZERO_NEZ}
 
 val sep : unit <-> string
 mapping sep : unit <-> string = {
--- a/model/riscv_types.sail
+++ b/model/riscv_types.sail
@@ -6,7 +6,7 @@
 /*  in the prover_snapshots directory (which include copies of their                     */
 /*  licences), is subject to the BSD two-clause licence below.                           */
 /*                                                                                       */
-/*  Copyright (c) 2017-2022                                                              */
+/*  Copyright (c) 2017-2023                                                              */
 /*    Prashanth Mundkur                                                                  */
 /*    Rishiyur S. Nikhil and Bluespec, Inc.                                              */
 /*    Jon French                                                                         */
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,8 @@ SAIL_DEFAULT_INST += riscv_insts_zks.sail
 SAIL_DEFAULT_INST += riscv_insts_zbkb.sail
 SAIL_DEFAULT_INST += riscv_insts_zbkx.sail
 
+SAIL_DEFAULT_INST += riscv_insts_zicond.sail
+
 SAIL_SEQ_INST  = $(SAIL_DEFAULT_INST) riscv_jalr_seq.sail
 SAIL_RMEM_INST = $(SAIL_DEFAULT_INST) riscv_jalr_rmem.sail riscv_insts_rmem.sail
 


model/riscv_insts_zicond.sail
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function clause execute (ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_NEZ)) = {
+  let value = X(rs1);
+  let condition = X(rs2);
+  result : xlenbits = if (condition != zeros()) then zeros()
+                                                else value;
+  X(rd) = result;
+  RETIRE_SUCCESS
+}
```
arichardson, line 30: I know the current style matches other extensions, but wouldn't it be simpler to do something like the following (especially since there are only two instructions and the enum doesn't really allow you to share much code):

```suggestion
union clause ast = RISCV_CZERO_EQZ : (regidx, regidx, regidx)
union clause ast = RISCV_CZERO_NEZ : (regidx, regidx, regidx)

mapping clause encdec = RISCV_CZERO_EQZ(rs2, rs1, rd) if haveZiCond() <-> 0b0000111 @ rs2 @ rs1 @ 0b101 @ rd @ 0b0110011
mapping clause encdec = RISCV_CZERO_NEZ(rs2, rs1, rd) if haveZiCond() <-> 0b0000111 @ rs2 @ rs1 @ 0b111 @ rd @ 0b0110011

mapping clause assembly = RISCV_CZERO_EQZ(rs2, rs1, rd)
  <-> "czero.eqz" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
mapping clause assembly = RISCV_CZERO_NEZ(rs2, rs1, rd)
  <-> "czero.nez" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)

function clause execute (RISCV_CZERO_EQZ(rs2, rs1, rd)) = {
  let value = X(rs1);
  let condition = X(rs2);
  result : xlenbits = if (condition == zeros()) then zeros()
                                                else value;
  X(rd) = result;
  RETIRE_SUCCESS
}

function clause execute (RISCV_CZERO_NEZ(rs2, rs1, rd)) = {
  let value = X(rs1);
  let condition = X(rs2);
  result : xlenbits = if (condition != zeros()) then zeros()
                                                else value;
  X(rd) = result;
  RETIRE_SUCCESS
}
```

martinberger:
Thanks for submitting this extension. What kind of testing have you done?

ptomsich:
> Thanks for submitting this extension. What kind of testing have you done?

Validation has been limited to smoke testing using a handful (of hand-written) tests using the riscv-arch-test framework and a check of the trace output from `c_emulator/riscv_sim_RV64` (the binutils patches are still in my queue to pull from our internal repositories and to send to the list):

E.g., we have the following in our tests:
```
// rd = (rs2 == 0) ? 0 : rs1;    with rd = x4 (expect: 0x0), rs1 = x16 (0x800), rs2 = x17 (0x0)
TEST_RR_OP(czero.eqz, x4, x16, x17, 0x0, 0x800, 0x0, x1, 0, x19)
// rd = (rs2 == 0) ? 0 : rs1;    with rd = x4 (expect: 0x800), rs1 = x16 (0x800), rs2 = x17 (0x3456)
TEST_RR_OP(czero.eqz, x4, x16, x17, 0x800, 0x800, 0x3456, x1, 8, x19)
// rd = (rs2 == 0) ? 0 : rs1;    with rd = x29 (expect: 0x1234), rs1 = x11 (0x1234), rs2 = x29 (0x1)
TEST_RR_OP(czero.eqz, x29, x11, x29, 0x1234, 0x1234, 0x1, x1, 16, x19)
```
And hand-validated the output:
```
[236] [M]: 0x000000008000039C (0x8008081B) addiw a6, a6, 2048
x16 <- 0x0000000000000800


mem[X,0x00000000800003A0] -> 0x0893
mem[X,0x00000000800003A2] -> 0x0000
[237] [M]: 0x00000000800003A0 (0x00000893) addi a7, zero, 0
x17 <- 0x0000000000000000


mem[X,0x00000000800003A4] -> 0x5233
mem[X,0x00000000800003A6] -> 0x0F18
[238] [M]: 0x00000000800003A4 (0x0F185233) czero.eqz tp, a6, a7
x4 <- 0x0000000000000000


mem[X,0x00000000800003A8] -> 0xB023
mem[X,0x00000000800003AA] -> 0x0040
[239] [M]: 0x00000000800003A8 (0x0040B023) sd tp, 0(ra)
mem[0x0000000080002218] <- 0x0000000000000000


mem[X,0x00000000800003AC] -> 0x1837
mem[X,0x00000000800003AE] -> 0x0000
[240] [M]: 0x00000000800003AC (0x00001837) lui a6, 1
x16 <- 0x0000000000001000


mem[X,0x00000000800003B0] -> 0x081B
mem[X,0x00000000800003B2] -> 0x8008
[241] [M]: 0x00000000800003B0 (0x8008081B) addiw a6, a6, 2048
x16 <- 0x0000000000000800


mem[X,0x00000000800003B4] -> 0x38B7
mem[X,0x00000000800003B6] -> 0x0000
[242] [M]: 0x00000000800003B4 (0x000038B7) lui a7, 3
x17 <- 0x0000000000003000


mem[X,0x00000000800003B8] -> 0x889B
mem[X,0x00000000800003BA] -> 0x4568
[243] [M]: 0x00000000800003B8 (0x4568889B) addiw a7, a7, 1110
x17 <- 0x0000000000003456


mem[X,0x00000000800003BC] -> 0x5233
mem[X,0x00000000800003BE] -> 0x0F18
[244] [M]: 0x00000000800003BC (0x0F185233) czero.eqz tp, a6, a7
x4 <- 0x0000000000000800


mem[X,0x00000000800003C0] -> 0xB423
mem[X,0x00000000800003C2] -> 0x0040
[245] [M]: 0x00000000800003C0 (0x0040B423) sd tp, 8(ra)
mem[0x0000000080002220] <- 0x0000000000000800


mem[X,0x00000000800003C4] -> 0x15B7
mem[X,0x00000000800003C6] -> 0x0000
[246] [M]: 0x00000000800003C4 (0x000015B7) lui a1, 1
x11 <- 0x0000000000001000


mem[X,0x00000000800003C8] -> 0x859B
mem[X,0x00000000800003CA] -> 0x2345
[247] [M]: 0x00000000800003C8 (0x2345859B) addiw a1, a1, 564
x11 <- 0x0000000000001234


mem[X,0x00000000800003CC] -> 0x0E93
mem[X,0x00000000800003CE] -> 0x0010
[248] [M]: 0x00000000800003CC (0x00100E93) addi t4, zero, 1
x29 <- 0x0000000000000001


mem[X,0x00000000800003D0] -> 0xDEB3
mem[X,0x00000000800003D2] -> 0x0FD5
[249] [M]: 0x00000000800003D0 (0x0FD5DEB3) czero.eqz t4, a1, t4
x29 <- 0x0000000000001234


mem[X,0x00000000800003D4] -> 0xB823
mem[X,0x00000000800003D6] -> 0x01D0
[250] [M]: 0x00000000800003D4 (0x01D0B823) sd t4, 16(ra)
mem[0x0000000080002228] <- 0x0000000000001234
```

billmcspadden-riscv:
@ptomsich :  Has this extension been ratified?

ptomsich:
> @ptomsich : Has this extension been ratified?

Going up for freeze in this quarter.
This PR is in support of the freeze acceptance criteria.

Note that we really need to support experimental (i.e., prior to freeze) extensions on SAIL — otherwise, Spike and QEMU will remain the de-facto reference for experimental extensions.

model/riscv_insts_zicond.sail
```diff
@@ -0,0 +1,30 @@
+union clause ast = ZICOND_RTYPE : (regidx, regidx, regidx, zicondop)
+
+mapping clause encdec = ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_EQZ) if haveZiCond() <-> 0b0000111 @ rs2 @ rs1 @ 0b101 @ rd @ 0b0110011
```
jrtc27, line 3: Guards need to be on both sides of the <-> to apply to both directions, and wrapping the <-> is generally a good idea for readability (like you do for the assembly below)

model/riscv_insts_zicond.sail
```diff
@@ -0,0 +1,30 @@
...
+  RISCV_CZERO_NEZ <-> "czero.nez"
+}
+
+mapping clause assembly = ZICOND_RTYPE(rs2, rs1, rd, op)
+  <-> zicond_mnemonic(op) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_EQZ)) = {
+  let value = X(rs1);
+  let condition = X(rs2);
+  result : xlenbits = if (condition == zeros()) then zeros()
```
jrtc27, line 17: These don't need to be mutable

model/riscv_sys_regs.sail
```diff
@@ -208,6 +208,9 @@ function haveZknd()   -> bool = true
 
 function haveZmmul()  -> bool = true
 
+/* ZiCond extension support */
+function haveZiCond() -> bool = true
```
jrtc27, line 212: Renamed to Zicond 4 days ago

model/riscv_types.sail
```diff
@@ -6,7 +6,7 @@
 /*  in the prover_snapshots directory (which include copies of their                     */
 /*  licences), is subject to the BSD two-clause licence below.                           */
 /*                                                                                       */
-/*  Copyright (c) 2017-2021                                                              */
+/*  Copyright (c) 2017-2022                                                              */
```
jrtc27, line 9: 2022 or 2023?

model/riscv_types.sail
```diff
@@ -409,6 +410,8 @@ enum biop_zbs = {RISCV_BCLRI, RISCV_BEXTI, RISCV_BINVI, RISCV_BSETI}
 
 enum extop_zbb = {RISCV_SEXTB, RISCV_SEXTH, RISCV_ZEXTH}
 
+enum zicondop = { RISCV_CZERO_EQZ, RISCV_CZERO_NEZ }
```
jrtc27, line 413: Seems style in this file is to not have spaces within {} for enums

martinberger (approved):
Looks fine to me. With or without the small variant suggested by @arichardson, but change copyright to 2023.

model/riscv_types.sail
ptomsich: The change has been lingering on our tree since 2022.
I'll update to 2023 when merging.

model/riscv_insts_zicond.sail
ptomsich: Let's stick with the same style with have across other extensions for the time being.
My gut tells me that we may eventually want to share the implementation by testing the 'op' inside the function (to determine what comparison operator to use).

That said: the suggested style is indeed much nicer.

ptomsich:
> Looks fine to me. With or without the small variant suggested by @arichardson, but change copyright to 2023.

@martinberger What is the next step? Will you take care of the merge or should I request write-access to this repository?

martinberger:
> @martinberger What is the next step? Will you take care of the merge 

We will do this.

martinberger:
> We will do this.

@ptomsich  BTW, @billmcspadden-riscv  and I think it's better, at least for now, if we don't merge pre-ratification extensions into `master` for various reasons. We can revisit this decision any time. Instead, we want to merge Zicond into a dedicated  `zicond-dev` branch. Once Zicond is ratified, we can merge `zicond-dev` into `master`. Is that OK with you?

jrtc27:
In the toolchain world GCC tried that approach and it's been painful. LLVM picked the approach of merging support for pre-ratification extensions but gated by an option to enable them, which has turned out to be the better approach, and there have been efforts to move towards that in the GCC world (maybe it finally happened?). Feature branches are a pain to deal with, especially when people want to test their cores that implement multiple pre-ratification extensions against the model, as then they would have to create their own local branches that merge all of the feature branches they need (which may be non-trivial if there are interactions).

Also, Spike has pre-ratification extensions merged and updated regularly, so if we don't do the same people will just keep using Spike and the Sail model will remain official only in name.

martinberger:
> In the toolchain world GCC tried that approach and it's been painful. LLVM picked the approach of merging support for pre-ratification extensions 

What I don't understand how to handle this if the ratification of a new extension requires modification of older, existing extensions. How can both be co-exist in a single branch, at least without major hassle?

The speed of change in LLVM is dramatically higher than that in the RISCV model.

jrtc27:
> > In the toolchain world GCC tried that approach and it's been painful. LLVM picked the approach of merging support for pre-ratification extensions
> 
> @jrtc27 What I don't understand how to handle this if the ratification of a new extension requires modification of older, existing extensions. How can both be co-exist in a single branch, at least without major hassle?
> 
> The speed of change in LLVM is dramatically higher than that in the RISCV model.

LLVM's policy is "Experimental extensions are expected to either transition to ratified status, or be eventually removed. The decision on whether to accept an experimental extension is currently done on an entirely case by case basis". If an extension that requires extensive modification to existing extensions were to be proposed then either it'd have to be seriously on the path to ratification, at which point the expectation is that the future development work will be minimal and it won't be canned, or it'd be rejected. Most extensions live off on the side, like this one. Even vector doesn't need a huge amount of non-vector support.

martinberger:
> LLVM's policy is "Experimental extensions are expected to either transition to ratified status, or be ...

How is the current, fully approved version divintated? By tagging?

jrtc27:
I don't understand the question?

martinberger:
> I don't understand the question?

What do I do if I want to check out a version that free from experimental stuff?

jrtc27:
You don't, you just don't enable them with the opt-in compiler flag.

allenjbaum:
In general, adding a new extension must not require a modification of an
older existing extension.
It can add a new behavior, but cannot alter existing behavior.
If alteration  is required, then another new extension is defined instead
of the existing extension behavior.
You can see the former in ePMP, which is completely backwards compatible,
but adds new behavior,
Vs. Zc?? (don't recall the name exactly) , which is a new extension that
has all of the existing C-extension without FP load/store, and then new
extensions that re-use those encodings with new functionality

PeterSewell:
On Thu, 26 Jan 2023 at 23:59, Jessica Clarke ***@***.***>
wrote:

> You don't, you just don't enable them with the opt-in compiler flag.
>
For RISC-V (presumably similarly), the switches to enable/disable
extensions have to remain in the integrated model indefinitely, so this
seems sensible to me.   Some extensions could need additional hooks added
to the base model, which one might want to back out if an extension never
got ratified, but that seems like a minor problem.

Peter

martinberger:
It looks like there is a preference for following the LLVM model of hiding extensions waiting for ratification behind a flag, and that the case were that is problematic (and extension requires major modification of existing code) is unlikely to apply (and can be handled separately). So I think we should go for flag approach.

model/riscv_insts_zicond.sail
```diff
@@ -0,0 +1,32 @@
...
+  RISCV_CZERO_NEZ <-> "czero.nez"
+}
+
+mapping clause assembly = ZICOND_RTYPE(rs2, rs1, rd, op)
+  <-> zicond_mnemonic(op) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute (ZICOND_RTYPE(rs2, rs1, rd, RISCV_CZERO_EQZ)) = {
+  let value = X(rs1);
+  let condition = X(rs2);
+  let result : xlenbits = if (condition == zeros()) then zeros()
```
jrtc27, line 19:
```suggestion
  let result : xlenbits = if condition == zeros() then zeros()
```

model/riscv_sys_regs.sail
```diff
@@ -208,6 +209,9 @@ function haveZknd()   -> bool = true
 
 function haveZmmul()  -> bool = true
 
+/* ZiCond extension support */
```
jrtc27, line 212: Missed one

jrtc27 (commented):
Couple of remaining nits, otherwise looks good subject to the experimental extension flag existing for this to use. I do agree it would be slightly nicer to have a match on the op inside a shared function clause, but the code is so short and trivial that the duplication isn't a big deal and shouldn't block this; it's a simple cleanup that can come later.

model/riscv_sys_regs.sail
ptomsich: Done.

martinberger (approved):

ptomsich:
@jrtc27 @martinberger @billmcspadden-riscv This are *approved* already. OK to merge, even while we wait for the flag to enable/disable experimental extensions?

From where I stand, we aren't making things worse if we merge experimental extensions while we still miss support for plenty of non-experimental ones.

jrtc27:
I've said as much somewhere on the mailing lists; we can't block adding extensions on having an -march+-menable-experimental-extensions equivalent. There's some real engineering to do for the latter.

You do need a copyright header on the new file though.

ptomsich:
> I've said as much somewhere on the mailing lists; we can't block adding extensions on having an -march+-menable-experimental-extensions equivalent. There's some real engineering to do for the latter.

ACK.
 
> You do need a copyright header on the new file though.

I don't see any copyright headers on (not an exhaustive list):
- model/riscv_insts_zba.sail
- model/riscv_insts_zbb.sail
- model/riscv_insts_zbc.sail
- model/riscv_insts_zbs.sail
- ...

I'll assume this falls under the top-level LICENSE (although we'll have to update that one as well) and will move this forward (leaving any missing copyright headers to a separate near-future cleanup).

@billmcspadden-riscv Please add the entire "copyright headers" (I find the long-form in many of the files quite distracting) issue to the agenda for the next golden-model meeting.  Also: should be use a CLA process on github to ensure that we have a full list of our contributors (as the LICENSE file gets outdated quickly…)?

jrtc27:
The lack of copyright headers in those files in an oversight. Current policy is to have the headers on all files. You can run `make apply_header` to help.

ptomsich:
> The lack of copyright headers in those files in an oversight. Current policy is to have the headers on all files. You can run `make apply_header` to help.

Thanks for the pointer (with apply_headers).  New PR is up.  I'll merge after your review/approval.

PR #210, opened 2023-02-03
Use not() instead of ~() for boolean negation

arichardson:
This may be more readable and also matches the sail-cheri-riscv model. For now this keeps ~ overloaded to accept bool, but in the future we may want to consider removing it (which is what I did to find all uses modified in this patch).


martinberger:
Why only apply this code-style only to negation? We have a similar overload here:

    overload operator ^ = {xor_vec, int_power, concat_str, xor_bool}

PR #223, opened 2023-02-20
Move ILLEGAL/C_ILLEGAL ast declaration to riscv_insts_begin.sail

arichardson:
This is useful for the sail-cheri-riscv model, where we would like to reuse C_ILLEGAL, but can't right now since it is current defined too late.

It is needed for https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/69

### DIFF
--- a/model/riscv_insts_begin.sail
+++ b/model/riscv_insts_begin.sail
@@ -85,3 +85,12 @@ scattered mapping encdec
 
 val encdec_compressed : ast <-> bits(16) effect {rreg}
 scattered mapping encdec_compressed
+
+/*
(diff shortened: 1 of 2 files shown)


jrtc27 (approved):

rmn30:
I previous version of Sail the `scattered union` clauses were logically gathered together at the declaration and the scattered function / mapping clauses were put at the `end` so this would just work. Maybe @Alasdair could confirm / deny.

That said I have no problem with this change.

Alasdair:
That's how it currently works. It may have gotten a little stricter with scoping over time.

martinberger (approved):

PR #224, opened 2023-02-21
Remove duopod

bacam:
The duopod is a small two instruction Sail specification similar to a couple of RISC-V instructions, intended as an example for experimentation and to use in the Sail manual.  The manual now uses an updated copy in the main Sail repository and I propose to remove it here.  Partly this is because it's not really part of RISC-V, but also because it causes some minor problems with the theorem prover output.


bacam:
I left it as a separate commit because it was already out of date, but perhaps it would be better to squash them given that files are being removed rather than added.

BTW, the CI has failed because github are phasing out Ubuntu 18.04 and are doing temporary brownouts.  I'll rerun it once it's back on, but we ought to update the CI to use a more recent release.

martinberger:
This is a good idea. The golden model should be as small as possible. Teaching material like the Duopod should go elsewhere.

bacam:
That's it squashed and passing CI.

martinberger (approved):

martinberger:
Thanks Brian!

Issue #124, opened 2021-11-05
Feature request: Sail code formatter for our CI


martinberger:
I've made a request for a Sail auto-code formatter [1] on Sail's GitHub. Given that they are an academic group, this is unlikely to be their highest priority. But it would be rather useful for RISC-V. Maybe such a formatter already exists in one of the many commercial teams that use RISC-V and they might be interested in donating it?

1. https://github.com/rems-project/sail/issues/158

trdthg:
Considering the current state of sail-riscv, it is almost impossible to have a complete and sufficiently detailed formatting rules.

for example:

- for '^'
  we can choose to have all elements wrap to a new line, or to have them wrap automatically based on line_width, but neither is currently followed in the code.
  ```

  mapping clause assembly = F_BIN_F_TYPE_D(rs2, rs1, rd, op)
                        <-> f_bin_f_type_mnemonic_D(op)
                            ^ spc() ^ freg_or_reg_name(rd)
                            ^ sep() ^ freg_or_reg_name(rs1)
                            ^ sep() ^ freg_or_reg_name(rs2)
  ``` 

  Similar examples also include vector concat syntax. `bits1 @ bits2`

- for Index
  
  This example would be difficult to implement, especially when you consider that the index itself can be any complex expression.

  ```
  val zvk_sm4_subword : bits(32) -> bits(32)
  function zvk_sm4_subword(x) = {
    sm4_sbox(x[31..24]) @
    sm4_sbox(x[23..16]) @
    sm4_sbox(x[15.. 8]) @
    sm4_sbox(x[ 7.. 0])
  }
  ```

- for List
  For crypt extensions, they have a large number of elements, but the data itself usually is just bit vector literals. Automatic wrap seems to be a good choice.  
  ```
  let sm4_sbox_table : vector(256, bits(8)) = [
  ..., ...
  ]
  ```

  But for other lists, have all elements wrap might be the better option.

For these complex examples above, I can't think of a good way to handle them unless you impose strict limitations in the rules. For instance, when trying to align `let foo = bar` statments, we could have limit that only if bar is an Int.

But for some simpler cases, like adding a trailing comma after the last mapping element, these can be done immediately.

I'm wondering how we can quickly apply these simple formatting rules to sail-model. Perhaps we need to add many Sail configuration options for various keywords, and disable them by default.

jrtc27:
[...]

PR #227, opened 2023-02-24
Vector integer/fixed-point arithmetic & mask instructions

XinlaiWan:
This PR mainly contains vector integer/fixed-point arithmetic instructions (Spec Chapter 11, 12) in `riscv_insts_vext_arith.sail` and vector mask instructions (Spec Chapter 15) in `riscv_insts_vext_mask.sail`. Besides, some arithmetic instructions involving mask registers (Spec 11.4, 11.8, 13.13) are collected in `riscv_insts_vext_vm.sail`, and this PR contains part of them for the help of running mask instruction ISA tests.

The model generated by this code can pass the vector ISA tests generated by the [rvv-atg](https://github.com/hushenwei2000/rvv-atg) for different configurations. (ELEN and VLEN are manually specified in `riscv_sys_control.sail` when compiling the model.) Please send me comments if there are issues in the new code. I'll do my best to get it right soon.

### DIFF
--- a/model/riscv_insts_vext_arith.sail
+++ b/model/riscv_insts_vext_arith.sail
@@ -42,7 +42,7 @@ function clause execute(VVTYPE(funct6, vm, vs2, vs1, vd)) = {
   let VLEN_pow = get_vlen_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if ~(valid_rd_mask(vd, vm)) then {handle_illegal(); return RETIRE_FAIL};
+  if ~(valid_rd_mask(vd, vm)) then { handle_illegal(); return RETIRE_FAIL };
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -106,14 +106,14 @@ function clause execute(VVTYPE(funct6, vm, vs2, vs1, vd)) = {
         VV_VMAXU         => to_bits(SEW, max(unsigned(vs2_val[i]), unsigned(vs1_val[i]))),
         VV_VMAX          => to_bits(SEW, max(signed(vs2_val[i]), signed(vs1_val[i]))),
         VV_VRGATHER      => {
-                              if (vs1 == vd | vs2 == vd) then {handle_illegal(); return RETIRE_FAIL};
+                              if (vs1 == vd | vs2 == vd) then { handle_illegal(); return RETIRE_FAIL };
                               let idx = unsigned(vs1_val[i]);
                               let VLMAX = int_power(2, LMUL_pow + VLEN_pow - SEW_pow);
                               assert(VLMAX <= 'n);
                               if idx < VLMAX then vs2_val[idx] else zeros() 
                             },
         VV_VRGATHEREI16  => {
-                              if (vs1 == vd | vs2 == vd) then {handle_illegal(); return RETIRE_FAIL};
+                              if (vs1 == vd | vs2 == vd) then { handle_illegal(); return RETIRE_FAIL };
                               /* vrgatherei16.vv uses SEW/LMUL for the data in vs2 but EEW=16 and EMUL = (16/SEW)*LMUL for the indices in vs1 */
                               let vs1_new : vector('n, dec, bits(16)) = read_vreg(num_elem, 16, 4 + LMUL_pow - SEW_pow, vs1);
                               let idx = unsigned(vs1_new[i]);
@@ -178,7 +178,7 @@ function clause execute(NVSTYPE(funct6, vm, vs2, vs1, vd)) = {
 
   if  ~(valid_rd_mask(vd, vm)) | ~(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow)) | 
       ~(valid_eew_emul(SEW_widen, LMUL_pow_widen)) 
-  then {handle_illegal(); return RETIRE_FAIL};
+  then { handle_illegal(); return RETIRE_FAIL };
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -245,7 +245,7 @@ function clause execute(NVTYPE(funct6, vm, vs2, vs1, vd)) = {
 
   if  ~(valid_rd_mask(vd, vm)) | ~(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow)) | 
       ~(valid_eew_emul(SEW_widen, LMUL_pow_widen)) 
-  then {handle_illegal(); return RETIRE_FAIL};
+  then { handle_illegal(); return RETIRE_FAIL };
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -305,7 +305,7 @@ function clause execute(MASKTYPEV(vs2, vs1, vd)) = {
   let num_elem      = get_num_elem(LMUL_pow, SEW); /* max(VLMAX,VLEN/SEW)) */
   let real_num_elem = if LMUL_pow >= 0 then num_elem else num_elem / (0 - LMUL_pow); /* VLMAX */
 
-  if vd == vreg_name("v0") then {handle_illegal(); return RETIRE_FAIL};
+  if vd == vreg_name("v0") then { handle_illegal(); return RETIRE_FAIL };
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -408,7 +408,7 @@ function clause execute(VXTYPE(funct6, vm, vs2, rs1, vd)) = {
   let LMUL_pow = get_lmul_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if ~(valid_rd_mask(vd, vm)) then {handle_illegal(); return RETIRE_FAIL};
+  if ~(valid_rd_mask(vd, vm)) then { handle_illegal(); return RETIRE_FAIL };
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -528,7 +528,7 @@ function clause execute(NXSTYPE(funct6, vm, vs2, rs1, vd)) = {
 
   if  ~(valid_rd_mask(vd, vm)) | ~(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow)) | 
       ~(valid_eew_emul(SEW_widen, LMUL_pow_widen)) 
-  then {handle_illegal(); return RETIRE_FAIL};
+  then { handle_illegal(); return RETIRE_FAIL };
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -595,7 +595,7 @@ function clause execute(NXTYPE(funct6, vm, vs2, rs1, vd)) = {
 
   if  ~(valid_rd_mask(vd, vm)) | ~(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow)) | 
       ~(valid_eew_emul(SEW_widen, LMUL_pow_widen)) 
-  then {handle_illegal(); return RETIRE_FAIL};
+  then { handle_illegal(); return RETIRE_FAIL };
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -661,7 +661,7 @@ function clause execute(VXSG(funct6, vm, vs2, rs1, vd)) = {
   let VLEN_pow = get_vlen_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if ~(valid_rd_mask(vd, vm)) then {handle_illegal(); return RETIRE_FAIL};
+  if ~(valid_rd_mask(vd, vm)) then { handle_illegal(); return RETIRE_FAIL };
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -679,7 +679,7 @@ function clause execute(VXSG(funct6, vm, vs2, rs1, vd)) = {
     if mask[i] then {
       result[i] = match funct6 {
         VX_VSLIDEUP    => {
-                            if (vs2 == vd) then {handle_illegal(); return RETIRE_FAIL};
+                            if (vs2 == vd) then { handle_illegal(); return RETIRE_FAIL };
                             if i >= rs1_val then vs2_val[i - rs1_val] else vd_val[i]
                           },
         VX_VSLIDEDOWN  => {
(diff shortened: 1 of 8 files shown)


model/riscv_insts_vext_utils.sail
```diff
@@ -8,6 +8,14 @@ mapping maybe_vmask : string <-> bits(1) = {
   sep() ^ "v0.t"  <-> 0b0
 }
 
+/* Check for valid EEW and EMUL values in vector widening/narrowing instructions */
+val check_eew_emul : (int, int) -> bool effect {rreg}
+function check_eew_emul(EEW, EMUL_pow) = {
+  let ELEN = int_power(2, get_elen_pow());
+  EEW      >= 8  & EEW      <= ELEN &
```
billmcspadden-riscv, line 15: Combine lines 15 and 16.

model/riscv_vlen.sail
```diff
@@ -1,10 +1,10 @@
 register elen : bits(1)
 
-val get_elen : unit -> {|32, 64|} effect {rreg}
+val get_elen_pow : unit -> {|5, 6|} effect {rreg}
 
-function get_elen() = match elen {
-    0b0 => 32,
-    0b1 => 64
+function get_elen_pow() = match elen {
```
billmcspadden-riscv, line 5: How was this function used prior to this change?  Benefit?
(The question is for educational purposes,  not to request a change back.)

billmcspadden-riscv (approved):
This review is from someone no familiar with the Vector spec.

I spent a bit of time during this review going back to the spec to see if I could understand the mapping between the spec and the Sail code.  I think I was able to track the basics.

jrtc27:
Why is there so much churn in the diff. The whole point of splitting it up into multiple PRs is so they can be reviewed separately and once each, but if you're going to go and rewrite a large chunk of it in each new PR that defeats the purpose and points at not having properly designed things in the first place.

model/riscv_vlen.sail
XinlaiWan: Before opening this PR I further read the Vector spec and noticed that the definition of ELEN is "the maximum size in bits of a vector element that any operation can produce or consume". So SEW should not be larger than ELEN which I hadn't considered this condition before. So I added this judgement into the function of getting the SEW value (`get_sew_pow()`). And for the convenience of comparing the value of SEW and ELEN, I modified ELEN to be the power value because SEW also uses the power value.

model/riscv_insts_vext_utils.sail
XinlaiWan: This one is fixed.

XinlaiWan:
> Why is there so much churn in the diff. The whole point of splitting it up into multiple PRs is so they can be reviewed separately and once each, but if you're going to go and rewrite a large chunk of it in each new PR that defeats the purpose and points at not having properly designed things in the first place.

In the file of `riscv_insts_vext_arith.sail` I also noticed that the git diff is a mess. Very sorry for that. This is because when I add new instruction blocks in this file, I adjusted the sequence of some instructions, and because each block has a similar code pattern, the git tool misinterpret my action as rewriting a lot of code. Actually the instruction code blocks that have been reviewed in the previous PR are not modified.

model/riscv_insts_vext_arith.sail
```diff
@@ -155,245 +155,270 @@ mapping vvtype_mnemonic : vvfunct6 <-> string = {
...
+  let vs1_val : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs1);
+  let vs2_val : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+  let vd_val  : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
+  result      : vector('n, dec, bits('m)) = undefined;
 
-  RETIRE_SUCCESS
-} 
+  let tail_ag : agtype = get_vtype_vta();
+  foreach (i from 0 to (num_elem - 1)) {
+    if i < start_element then {
```
vmoya-smd, line 310: Why is vstart supported in vmerge instructions but not in other arithmetic operations? Seems arbitrary as there is zero reason why vmerge could be interruptable in the middle of an operation when arithmetic operations aren't. The spec is quite flexible in the support for vstart in non memory operations so hardware implementations will have ample choice. In general interrupting a LMUL <=1 non-memory operation is probably not supported so vstart may only supported at destination register limits (LMUL > 1 at every step). So you can define in the SAIL model that vmerge can and other not, it's legal, but it's unlikely to match a given hardware implementation and you may want to specify why you made this choice.

vmoya-smd:
I couldn't find the code that checks for illegal cases of overlap between the destination register (vd) and source registers (vs1, vs2, ...). That applies mostly to operations with different SEW for destination and operands (widening, narrowing, comparison operations, ...).

vmoya-smd:
Another topic I already mentioned is the vstart support for non-memory vector operations. The spectrum of options supported by the spec is very large this model should follow some reasoned option, for example following what spike or a major vendor/IP provider supports.
If a vector operation isn't interruptible it's reasonable to don't support vstart!=0 as the vector operation won't need to 'continue'. This can be a bit more complicated because a hardware implementation could break vector operations that are LMUL>1 (write to multiple destination logical registers) into more than one internal operation and allow interrupts. Then vstart could be supported but only at register boundaries (depends on destination SEW).
What I think makes little sense is for some instructions to support vstart!=0 and some other don't (and should trigger illegal instruction) without a clear criteria.

XinlaiWan:
> I couldn't find the code that checks for illegal cases of overlap between the destination register (vd) and source registers (vs1, vs2, ...). That applies mostly to operations with different SEW for destination and operands (widening, narrowing, comparison operations, ...).

Thanks for reminding. I'll go through the code to check this issue.

model/riscv_insts_vext_arith.sail
```diff
@@ -420,77 +445,1759 @@ function clause execute(MOVETYPEV(vs1, vd)) = {
...
+  result      : vector('n, dec, bits('m)) = undefined;
+  mask        : vector('n, dec, bool)     = undefined;
+
+  (result, mask) = init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val);
+
+  assert(VLEN >= 0);
+  foreach (i from 0 to (num_elem - 1)){
+    if mask[i] then{
+      result[i] = match funct6 {
+        MVV_VMACC  => slice(to_bits(VLEN, signed(vs1_val[i]) * signed(vs2_val[i])), 0, SEW) + vd_val[i],
```
bacam, line 1325: This `slice(to_bits(...), ...)` pattern appears several times.  I think you could do it more simply by using `get_slice_int` directly (or perhaps by sticking to bitvectors throughout rather than round-tripping through integers).  In this particular example it would help Sail's translation to some theorem provers because they would currently need to case split on all the possible values for `VLEN`.

allenjbaum:
I am starting to get a grasp on what this is all about.
The non-normative note says:

Implementations are *permitted* to raise illegal instruction exceptions
when attempting to execute a vector instruction with a value of vstart that
the implementation can never produce when executing that same instruction
with the same vtype setting.

and

All vector loads and stores may generate and accept a non-zero vstart
value.


This is more than a little ambiguous (and flexible), but the implications
is that whether or not setting vstart to (some) non-zero value (but not
others)
can cause a trap or not can vary on an opcode-by-opcode value -- and
possible even by different setting of other CSRs (e.g. vtype?)
From an architectural test perspect, this is pretty horrible; we need to
supply Sail with a list of (il)legal vstart values for every opcode (and
CSR value?) combination.
At one level, I understand why it was down; there are a lot of "arithmetic"
(==non-memory?) vector ops that won't be single cycle because HW area gets
out of hand (e.g.  permutations, reductions),
and that didn't want to delay interrupts. This has to be constrained in
some way.
Vector reductions, and bit manips (poccnt, vfirst, vms[b/i/o]), viota,
 are called out specifically, so that's a few that are out of the way.

allenjbaum:
And: much as I dislike WARL fields and CSRs with a passion:
 - defining vstart as a WARL field removes the need for trapping entirely,
 - enables re-using the WARL mechanisms we are shoehorning into Sail,
     while removing the need for a new mechanism to describe all the
possible configuration options, and
 - avoids having to complexity into the Sail code for managing all the
special cases

allenjbaum:
Sigh - this doesn't quite solve the problem, because whether a Vstart value
is illegal or not depends on the specific op it is used with,
While WARL fields can have legality dependent on architectural state, the
opode is arguably not processor state.

model/riscv_insts_vext_arith.sail
XinlaiWan: This pattern is modified to use `get_slice_int` now.

XinlaiWan:
The checking functions for illegal vector register overlap are added into `riscv_insts_vext_utils.sail` now.

And for the `vstart` issue, I have checked all the instructions and fixed several missing ones. For now, most of them deal with `vstart` in the by calling the function `init_masked_result()` in `riscv_insts_vext_util.sail` before calculation. This function is used to distinguish prestart/active/inactive/tail elements following the general rule in the spec. The prestart elements in the destination register group will be masked and keep its original content. For some special instructions that don't follow the general rule or ignore the `vstart` explicitly in the spec, the logic of handling destination elements is written case  by case. 

Please send me comments if there are issues in my new commit.

model/riscv_insts_vext_arith.sail
```diff
@@ -42,6 +42,8 @@ function clause execute(VVTYPE(funct6, vm, vs2, vs1, vd)) = {
   let VLEN_pow = get_vlen_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
+  if ~(valid_rd_mask(vd, vm)) then {handle_illegal(); return RETIRE_FAIL};
```
jrtc27, line 45: First line of this file and the formatting is wrong, with 64 copies of this incorrect sequence in total.

XinlaiWan: This issue is fixed now.

ptomsich: > First line of this file and the formatting is wrong, with 64 copies of this incorrect sequence in total.

This file has large amounts of code duplication.  Can we use an execute function that provides a generic prologue and epilogue, implementing only the specific logic of each instruction differently?

jrtc27:
Why was with merged when it was clearly full of formatting issues just yesterday

martinberger:
> Clearly nobody is doing any serious review of this code if nobody’s spotting that

This is only merged into `vector-dev` not `master`. 

The purpose of `vector-dev`  is, in parts to keep things moving, while the code is not ready for `master`. We have a shortage reviewers, if you know any that have time and knowledge to help reviewing, we'd be delighted.

martinberger:
> Then what’s the point of these pull requests if you’re going to merge anything that gets posted without review?

It was reviewed, and the point of these intermediate PRs into `vector-dev` is, among other things, to build up a community with institutional knowledge about the RV Sail model, code reviews, GitHub PRs etc. We cannot take those for granted.

martinberger:
> and I believe that's my fault. 

No fault involved. There was consensus to go ahead despite known issues.

jrtc27:
> > Then what’s the point of these pull requests if you’re going to merge anything that gets posted without review?
> 
> It was reviewed, and the point of these intermediate PRs into `vector-dev` is, among other things, to build up a community with institutional knowledge about the RV Sail model, code reviews, GitHub PRs etc. We cannot take those for granted.

Not properly given the first line of the main file was incorrectly formatted. I didn’t go any further as I did not have time to go review the whole diff at that point, my intent was to illustrate that no real review had been done and therefore (a) it was not ready to merge (b) somebody needs to be reviewing code other than me, because when I don’t nobody does and code style violations quickly pour in.

martinberger:
>  *code style violations quickly pour in.*

We greatly appreciate your efforts, and it makes the Sail model better.

But we should not sequentialise all issues in code review, b/c if we do we never get anywhere. Instead: parallelise ... Not everybody sees code style violations as highest priority issue that should block more macros stuff.  Right now a main goal of the reviews is more macro: does the proposed code faithfully implement the intention of the vector extension?

PeterSewell:
Alasdair has some auto-formatting work in hand...

jrtc27:
> We (mostly Jessica) are spending far too much time having to spot formatting issues. I am a bit unsure why formatting is so high a priority compared to actual semantic issues, but we would all be much better off if we (or development partners) could spend time developing a "linter" that could be used by code developers.
> […](#)

Code is much harder to read when formatting is messy. This makes it harder to actually understand it and spot the semantic issues. Both also need fixing and code style issues are easier to spot, so when they're plentiful they take up all the review time and drain the amount of energy available to go and find semantic issues, not just make it harder to spot them. Finally, this code is for the official model, and is intended to end up being quoted in the ISA manuals, which means it must be consistently formatted lest the ISA manuals end up becoming an unprofessional mess.

allenjbaum:
Messy formatting is one thing; an extra space here and there is another,
and I suspect that someone looking at code snippets will misunderstand what
the code is doing if those slip in.
Semantic corrections are necessary; formatting is a nice-to-have. Incorrect
semantics is fatal; incorrect formatting is not.
I'd much rather reviewers concentrate on semantic changes, and would much
rather leave that to a lint (manual, or ideally programatic) pass after a
semantic review.

ptomsich:
Extraneous spaces (or tab/space mismatches) will cause additional noise in future diffs, making reviews harder.

I generally agree with Jessica on all counts, but the resolution is not manual style reviews — we must get a formatter for SAIL in place that will allow us to reformat efficiently (similarly to how we run `git clang-format HEAD~` on other projects before submitting/merging).

Note that SAIL already can output the AST (using `-output_sail`) after processing.  This is not pretty or enforcing formatting, yet, but I'd imagine this to be a solid starting point for developing a proper linter.

jrtc27:
I agree having manual style review is a total waste of time. Auto-formatters help, though I'm of they view they're just a convenience for the author and shouldn't be *necessary*, as people should be capable of properly formatting their code (at least for the most part) themselves, and for those that can't I question their discipline when it comes to the actual semantics (my only "auto-formatter" is vim's autoindenting based on rudimentary knowledge of block-level indentation). In my experience there is strong correlation between people who write correct code and people who correctly format their code.

But that's just formatting. There are still all kinds of code style issues that aren't just where you put the whitespace, parentheses and braces, such as how to name variables, what types to use, whether you duplicate code and whether variables are mutable. You can write code that implements the architecture just as correctly but has poor style for all of those, and no auto-formatter/linter is going to help you there. But it does at least get the extremely low-hanging fruit.

martinberger:
We all agree that auto-formatting is desirable. A prototype already seems to exist, but code formatting is orthogonal to Vector and really a question of Sail. Can I suggest to move this discussion to 

1. https://github.com/rems-project/sail/issues/158
2. https://github.com/riscv/sail-riscv/issues/124

allenjbaum:
Thanks, that is my opinion also, anc I didn't know how to express that I
woul d separate that as more programming style than just formatting,
That is important to readability and understandability, so is still
something that seems to need manual review (though /i wish simple
violations could still be caught by a linter).

PR #229, opened 2023-03-07
Add a new recipe to run the Sail tests manually via GH Actions

rpsene:
This PR adds the capability to run the tests manually without depending on a new change to trigger it.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:


jrtc27:
Why? If the model hasn't changed the tests really shouldn't. On the very rare occasion that you need a manual run for some unusual set of circumstances you can just open a dummy PR, but I don't see how this is very useful.

jrtc27:
This certainly shouldn't be duplicating so much code if you really do want to add it for some reason. Reuse the existing code and gate steps that are specific to PRs.

jrtc27 (changes requested):
Unjustified and unnecessarily duplicates code

rpsene:
@jrtc27 thanks for the feedback, PR changed. Initially I was not doing executing all same steps.

.github/workflows/compile.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push, pull_request,workflow_dispatch]
```
jrtc27, line 3:
```suggestion
on: [push, pull_request, workflow_dispatch]
```
...

rpsene: fixed in the original PR.

billmcspadden-riscv (approved):
This looks fine to me.

jrtc27:
The minute notes just say "2 use cases" and that this was approved. That is totally useless for anyone not in the meetings. Please justify somewhere, especially when I have explicitly objected to this change and asked for the motivation a week ago.

PR #234, opened 2023-03-23
Add support for the Zvkned extension

charmitro:
Implements the Zvkned (NIST Suite: Vector AES Block Cipher) extension, as of version [Draft: 20230303
](https://github.com/riscv/riscv-crypto/releases/tag/v20230303).

* vaesef.[vv,vs]
* vaesem.[vv,vs]
* vaesdf.[vv,vs]
* vaesdm.[vv,vs]
* vaeskf1.vi
* vaeskf2.vi
* vaesz.vs

All instructions were tested with VLEN=128 and results were compared with QEMU results of each instruction.

Current revision is rebased with the latest changes of `vector-dev` branch.

_Note_
It's better to merge all of Zvk extension support PRs in the created order, as this is going to create conflicts and it would be easier to resolve them.

### DIFF
--- a/model/riscv_insts_zvkned.sail
+++ b/model/riscv_insts_zvkned.sail
@@ -1,5 +1,5 @@
 /*
- * Vector Cryptography Extension - NIST Suite: Vecttor AES Block Cipher
+ * Vector Cryptography Extension - NIST Suite: Vector AES Block Cipher
  * ----------------------------------------------------------------------
  */
 
--- a/model/riscv_insts_zvkned.sail
+++ b/model/riscv_insts_zvkned.sail
@@ -94,3 +94,81 @@ function clause execute (RISCV_VAESEF(vs2, vd, suffix)) = {
     RETIRE_SUCCESS
   }
 }
+
+/* VAESEM.[VV, VS] */
+
+mapping vaesem_mnemonic : bits(7) <-> string = {
+  0b1010001 <-> "vaesem.vv",
+  0b1010011 <-> "vaesem.vs",
+}
+
+union clause ast = RISCV_VAESEM : (regidx, regidx, string)
+
+mapping clause encdec = RISCV_VAESEM(vs2, vd, suffix)	       if (haveRVV() & haveZvkned())
+ <-> vv_or_vs(suffix) @ vs2 @ 0b00010 @ 0b010 @ vd @ 0b1110111 if (haveRVV() & haveZvkned())
+
+mapping clause assembly = RISCV_VAESEM(vs2, vd, suffix)
+ <-> vaesem_mnemonic(vv_or_vs(suffix)) ^ spc() ^ vreg_name(vd)
+				       ^ sep() ^ vreg_name(vs2)
+
+function clause execute (RISCV_VAESEM(vs2, vd, suffix)) = {
+  let SEW      = get_sew();
+  let LMUL_pow = get_lmul_pow();
+  let LMUL     = if LMUL_pow < 0 then 0 else LMUL_pow;
+  let VLEN     = int_power(2, get_vlen_pow());
+  let num_elem = get_num_elem(LMUL_pow, SEW);
+
+  if (zvk_check_elements(VLEN, num_elem, LMUL, SEW) == false)
+  then {
+    handle_illegal();
+    RETIRE_FAIL
+  } else {
+    let 'n = num_elem;
+    let 'm = SEW;
+    assert('m == 32);
+
+    let vs2_val : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+    let vd_val  : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
+    result      : vector('n, dec, bits('m)) = undefined;
+
+    vd_state : bits(128) = undefined;
+    vs2_key  : bits(128) = undefined;
+
+    eg_len = (unsigned(vl) / 'n);
+    eg_start = (unsigned(vstart) / 'n);
+
+    foreach (i from eg_start to (eg_len - 1)) {
+      assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < 'n);
+      vd_state[31..0]   = vd_val[i*4+0];
+      vd_state[63..32]  = vd_val[i*4+1];
+      vd_state[95..64]  = vd_val[i*4+2];
+      vd_state[127..96] = vd_val[i*4+3];
+
+      if suffix == "vv" then {
+	vs2_key[31..0]   = vs2_val[i*4+0];
+	vs2_key[63..32]  = vs2_val[i*4+1];
+	vs2_key[95..64]  = vs2_val[i*4+2];
+	vs2_key[127..96] = vs2_val[i*4+3];
+      } else {
+	vs2_key[31..0]   = vs2_val[0];
+	vs2_key[63..32]  = vs2_val[1];
+	vs2_key[95..64]  = vs2_val[2];
+	vs2_key[127..96] = vs2_val[3];
+      };
+
+      let sb       : bitvector(128, dec) = aes_subbytes_fwd(vd_state);
+      let sr       : bitvector(128, dec) = aes_shift_rows_fwd(sb);
+      let mix      : bitvector(128, dec) = aes_mixcolumns_fwd(sr);
+      let ark      : bitvector(128, dec) = mix ^ vs2_key;
+
+      result[i*4+0] = ark[31..0];
+      result[i*4+1] = ark[63..32];
+      result[i*4+2] = ark[95..64];
+      result[i*4+3] = ark[127..96];
+    };
+
+    write_single_vreg(num_elem, 'm, vd, result);
+    vstart = EXTZ(0b0);
+    RETIRE_SUCCESS
+  }
+}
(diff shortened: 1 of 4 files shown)


charmitro:
One note, the latest revision of `vector-dev` branch fails to build, therefore this PR and https://github.com/riscv/sail-riscv/pull/235 are failing to build also since they are synchronized.

charmitro:
Updated the PR:
* Correction on iterating based on the specification
* Correctly Load & Store from/to vector registers

charmitro:
Rebased from `vector-dev` branch.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,550 @@
+/*
+ * Vector Cryptography Extension - NIST Suite: Vecttor AES Block Cipher
```
nibrunieAtSi5, line 2:
```suggestion
 * Vector Cryptography Extension - NIST Suite: Vector AES Block Cipher
```

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,550 @@
...
+
+/*
+ * Helper functions.
+ * ----------------------------------------------------------------------
+ */
+
+val	 zvk_check_elements : (int, int, int, int) -> bool
+function zvk_check_elements(VLEN, num_elem, LMUL, SEW) = {
+  ((unsigned(vl)%num_elem) != 0) | ((unsigned(vstart)%num_elem) != 0) | (LMUL*VLEN) < (num_elem*SEW)
+}
```
nibrunieAtSi5, line 14: SAIL newbie question: the vector crypto spec distinguishes illegal cases (`(LMUL*VLEN) < (num_elem*SEW)`) and reserved cases (`((unsigned(vl)%num_elem) != 0) | ((unsigned(vstart)%num_elem) != 0)`) (which can legally raise an illegal exception or not). Does this difference is required in the sail model ?

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,550 @@
...
+  let num_elem = get_num_elem(LMUL_pow, SEW);
+
+  if (zvk_check_elements(VLEN, num_elem, LMUL, SEW) == false)
+  then {
+    handle_illegal();
+    RETIRE_FAIL
+  } else {
+    let 'n = num_elem;
+    let 'm = SEW;
+    assert('m == 32);
```
nibrunieAtSi5, line 60: what is the difference between the `assert` here and the `handle_illegal` before ? (Are they both equivalent to raising the illegal instruction exception or is it assumed SEW should be equal to 32 by that point and else the illegal instruction exception was raised before ?)

allenjbaum:
I would argue that it does if we want to be able to test that DUTs that
support either behavior do so correctly.
This is an Archie rural option, and the Sail model should support ALL
architectural options.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,550 @@
...
+
+    let vs2_val : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+    let vd_val  : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
+    result      : vector('n, dec, bits('m)) = undefined;
+
+    vd_state : bits(128) = undefined;
+    vs2_key  : bits(128) = undefined;
+
```
nibrunieAtSi5, line 68: Would it make sense to define both `vd_state` and `vs2_key` within the scope of the `foreach` since they are only used in the loop body and their live range does not exceed one iteration ?

model/riscv_insts_zvkned.sail
nibrunieAtSi5, line 1: There are reserved encodings for the `.vs` form (for example when `vs2` scalar element groups overlaps the `vd` register group), where would such constraint appear in the SAIL code ?

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,550 @@
...
+ */
+
+/*
+ * Helper functions.
+ * ----------------------------------------------------------------------
+ */
+
+val	 zvk_check_elements : (int, int, int, int) -> bool
+function zvk_check_elements(VLEN, num_elem, LMUL, SEW) = {
+  ((unsigned(vl)%num_elem) != 0) | ((unsigned(vstart)%num_elem) != 0) | (LMUL*VLEN) < (num_elem*SEW)
```
nibrunieAtSi5, line 13: the last term looks suspicious, is it trying to implement `EGW <= VLEN * LMUL` / `EGS <= VLMAX` ?
If so I do not think this is correct, for example I don't understand how the EGS / EGW is injected. Could a comment be added to explain which checks this function is supposed to perform ?

model/riscv_insts_zvkned.sail
charmitro: It is assumed that SEW should be equal to 32.

But mainly fixes compilation errors, ensuring the `vd_val` &  `vs2_val` have SEW=32, allowing us to load `bits(32)` later on, for example:
```
vd_state[31..0]   = vd_val[i*4+0];
```

model/riscv_insts_zvkned.sail
charmitro: This is implementing the following found in the specification, Version v0.4.2, 03 March, 2023.
```
if( ((vl%EGS)<>0) | ((vstart%EGS)<>0) | (LMUL*VLEN < EGW))  then {
      handle_illegal();  // illegal instruction exception
      RETIRE_SUCCESS
    } else {
    . . . 
```

model/riscv_insts_zvkned.sail
charmitro: ACK., good catch!

model/riscv_insts_zvkned.sail
nibrunieAtSi5: How is `EGW` derived from `num_elen * SEW` ?

charmitro: From the Vector Crypto Spec we have,
```
• Element Group Width (EGW) - total number of bits in an element group 
• Effective Element Width (EEW) - number of bits in each element
• Element Group Size (EGS) - number of elements in an element group
For all of the vector crypto instructions in this specification, EEW=SEW.
```

So here we're doing with `EGS * EEW`:
* num_elem = EGS
* SEW = EEW

nibrunieAtSi5: Sorry but I am still missing something, it looks like `num_elem` comes from `get_num_elem` and I do not get how `EGS` appears here:
```
$ git grep -A8 "fun.*get_num_elem" *.sail
model/riscv_vext_regs.sail:function get_num_elem(LMUL_pow, SEW) = {
model/riscv_vext_regs.sail-  let VLEN = int_power(2, get_vlen_pow());
model/riscv_vext_regs.sail-  let LMUL_pow_reg = if LMUL_pow < 0 then 0 else LMUL_pow;
model/riscv_vext_regs.sail-  /* Ignore lmul < 1 so that the entire vreg is read, allowing all masking to
model/riscv_vext_regs.sail-   * be handled in init_masked_result */
model/riscv_vext_regs.sail-  let num_elem = int_power(2, LMUL_pow_reg) * VLEN / SEW;
model/riscv_vext_regs.sail-  assert(num_elem > 0);
model/riscv_vext_regs.sail-  num_elem
model/riscv_vext_regs.sail-}
```

model/riscv_insts_zvkned.sail
nibrunieAtSi5: Thank you

nibrunieAtSi5:
@charmitro , when I tried to build this branch, I get the following error:
```

Type error:
model/riscv_insts_vext_utils.sail:422.23-32:
422 |    16 => (xf[15..15], xf[14..0]),
    |                       ^-------^
    | Tried performing type coercion from bitvector(((14 - 0) + 1), dec) to bitvector(('n * 1), dec) on subrange_bits(xf, 14, 0)
    | Coercion failed because:
    | Failed to prove constraint: ((14 - 0) + 1) == ('n * 1)
make: *** [Makefile:315: generated_definitions/lem/RV64/riscv.lem] Error 1
```

Do you know if this is a known issue ?

charmitro:
> @charmitro , when I tried to build this branch, I get the following error:
> 
> ```
> 
> Type error:
> model/riscv_insts_vext_utils.sail:422.23-32:
> 422 |    16 => (xf[15..15], xf[14..0]),
>     |                       ^-------^
>     | Tried performing type coercion from bitvector(((14 - 0) + 1), dec) to bitvector(('n * 1), dec) on subrange_bits(xf, 14, 0)
>     | Coercion failed because:
>     | Failed to prove constraint: ((14 - 0) + 1) == ('n * 1)
> make: *** [Makefile:315: generated_definitions/lem/RV64/riscv.lem] Error 1
> ```
> 
> Do you know if this is a known issue ?

I can't seem to reproduce it on my side. Could it be a problem in the upstream `vector-dev` branch(I believe at some point it was force-pushed, try resetting it(?)), but on my forked `vector-dev` branch everything seems fine.

model/riscv_insts_zvkned.sail
charmitro: I'll analyse this SAIL internal first and reach back. I'm positive `get_num_elem` returns the correct value, but a double-check is not a problem.

KotorinMinami:
@charmitro May I ask if you suggest that I continue to promote the integration of this extension?

charmitro:
> @charmitro May I ask if you suggest that I continue to promote the integration of this extension?

Hi @KotorinMinami, absolutely! However, keep in mind that this was originally developed with Draft version [v20230303](https://github.com/riscv/riscv-crypto/releases/tag/v20230303), so there may have been significant changes since then.

To ease further development, I performed a minimal rebase with master, which, as far as I know, now includes Vector support (since the `vector-dev` branch has been merged). In any case, feel free to reach out for any help.

 @billmcspadden-riscv Would it be helpful to join the Golden Model weekly meeting as well?

jordancarlin:
Closing now that #752 is merged.

PR #235, opened 2023-03-23
Add support for the Zvknh[ab] extensions

charmitro:
Implements the Zvknh[ab]  (NIST Suite: Vector SHA-2 Secure Hash) extension, as of version [Draft: 20230303
](https://github.com/riscv/riscv-crypto/releases/tag/v20230303)

* vsha2ms.vv
* vsha2c[hl].vv

All instructions were tested with VLEN=128(SHA-256) & VLEN=256(SHA-512) and results were compared with QEMU results of each instruction.

Current revision is rebased with the latest changes of `vector-dev` branch.


charmitro:
Updated the PR:
* Correction on iterating based on the specification
* Correctly Load & Store from/to vector registers

This update is adjusting the implementation to match the Spike implenentation and as a follow-up.

charmitro:
Rebased from `vector-dev` branch.

model/riscv_insts_zvknhab.sail
```diff
@@ -0,0 +1,256 @@
...
+  let num_elem = get_num_elem(LMUL_pow, SEW);
+
+  if (zvk_check_elements(VLEN, num_elem, LMUL, SEW) == false)
+  then {
+    handle_illegal();
+    RETIRE_FAIL
+  } else {
+    let 'n = num_elem;
+    let 'm = SEW;
+    assert('m == 32);
```
nibrunieAtSi5, line 175: is this only supporting SHA-256 ? where is tha SHA-512 (SEW=64) support ?

model/riscv_insts_zvknhab.sail
```diff
@@ -0,0 +1,256 @@
...
+  let VLEN     = int_power(2, get_vlen_pow());
+  let num_elem = get_num_elem(LMUL_pow, SEW);
+
+  if (zvk_check_elements(VLEN, num_elem, LMUL, SEW) == false)
+  then {
+    handle_illegal();
+    RETIRE_FAIL
+  } else {
+    let 'n = num_elem;
+    let 'm = SEW;
```
nibrunieAtSi5, line 88: why no assert of `SEW` this time (only `32` and `64` should be supported depending on Zvknhb vs Zvknha)

model/riscv_insts_zvknhab.sail
charmitro: Addressed.

nibrunieAtSi5: Does this mean that no SHA-512 tests were executed before or am I missing something ?

charmitro: These have only been tested against QEMU(see https://github.com/riscv-software-src/riscv-isa-sim/pull/1303#issuecomment-1514634288).

Missing SEW=64 occurred due to misreading of the Specification. 

FYI, in SAIL we manually hardcode `VLEN`(wouldn't be long until we will be able to configure it via Command Line options).

model/riscv_insts_zvknhab.sail
```diff
@@ -0,0 +1,245 @@
...
+    assert('m == 32 | 'm == 64);
+
+    let vm_val  : vector('n, dec, bool)     = read_vmask(num_elem, 0b1, vreg_name("v0"));
```
nibrunieAtSi5, line 91: Why is the mask extracted here and used in `init_masked_result` while it is not in https://github.com/riscv/sail-riscv/pull/234 ? (both `Zvknh[ab]` and `Zvkned` have no instruction using masks)

charmitro: You're correct, thanks!

jordancarlin:
Support for Zvknh[ab] was merged with #843. Closing this earlier version.

PR #240, opened 2023-03-27
Enhances RISC-V Sail test execution within a container

rpsene:
This commit enhances RISC-V Sail test execution within a container introducing the following updates:

* Improved Dockerfile for better container setup.
* Simplified entrypoint script for streamlined build execution.
* GitHub action that triggers image builds only when the Dockerfile is modified.
* Comprehensive documentation for seamless user experience.

### DIFF
--- a/.github/workflows/build-container-img.yml
+++ b/.github/workflows/build-container-img.yml
@@ -1,0 +1,43 @@
+name: Build and Publish Base Container Image for SAIL
+
+on:
+  push:
+    branches: [main]
+    paths:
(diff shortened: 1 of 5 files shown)


jrtc27:
I'm not sure about this one, I have concerns about not being able to test the model with a newer Sail in a PR, for example

rpsene:
> I'm not sure about this one, I have concerns about not being able to test the model with a newer Sail in a PR, for example

we can either create a base image for each version of SAIL or set a parameter that allow using a desired version.

PR #250, opened 2023-04-08
fix(build): Don't hardcode gcc (especially on darwin)

Et7f3:


jrtc27:
On all my Apple systems gcc is an alias for clang. I've never needed this on macOS (and I use macOS as my daily driver).

arichardson:
I've also never had an issue with this but I think the change makes sense and should be merged (with a slightly different commit message). It will allow building on FreeBSD without having to install the GCC port.

Et7f3:
I use nix which sandbox build and doesn't alias by default. This change can also be useful when cross compiling (only change CC to target-prefix-gcc/target-prefix-clang)

jrtc27:
Does `$(CC)` have a default set?

jrtc27:
Who's cross-compiling the Sail model? I doubt that works well...

Et7f3:
> Does `$(CC)` have a default set?

Yes I haven't CC variable in my environment but when I use in makefile it get defined. Apple make (outside nix) have CC=cc (and this is fine because it is also a symlink to clang). Inside nix the clang package also provide a cc stub so it is fine. so the default of make is perfectly usable.

> Who's cross-compiling the Sail mode

I don't know but this solution have the same maintenance cost but one offer more feature. If anyone need it, this change will help him get cross compiled sail quicker.

@arichardson something like:
> fix(build): Don't hardcode gcc

or

> fix(build): Make the compiler configurable

or

> fix(build): Use a generic compiler

ptomsich:
As discussed in today's meeting...

@billmcspadden-riscv @martinberger @jrtc27 Are we going to move this forward, or do we reject this based on the additional maintenance risks for a feature that we currently don't see a need for?

Timmmm:
This looks fine to me. Any objections if I merge this with this commit message?

```
Don't hard-code GCC in Makefile

This uses the `CC` variable (which defaults to `cc`) so that compilation
works when using Clang. This also allows the compiler to be overridden
via `CC=foo make`.
```

Timmmm (approved):

jrtc27:
> This looks fine to me. Any objections if I merge this with this commit message?
> 
> ```
> Don't hard-code GCC in Makefile
> 
> This uses the `CC` variable (which defaults to `cc`) so that compilation
> works when using Clang. This also allows the compiler to be overridden
> via `CC=foo make`.
> ```

Fine with me. I don't recall what I meant in https://github.com/riscv/sail-riscv/pull/250#issuecomment-1568851428, the diff looks fine as-is and it doesn't look like there's been any change since said comment. Maybe I was just confused about whether, and where, CC was being set by default, or some other misunderstanding...

rsnikhil:
On my Mac (Jan 2023 Apple MacBook Air, Ventura 13.6.7), both 'cc' and 'gcc' invoke clang

arichardson (approved):

PR #251, opened 2023-04-09
Add Zfa support (excl. quad-precision instructions)

ptomsich:
This adds the (proposed) Zfa (FP gap fillers) instructions, as specified in chapter 25 of https://github.com/riscv/riscv-isa-manual/releases.

Tested against Spike (Zfa is merged in Spike as https://github.com/riscv-software-src/riscv-isa-sim/commit/09140c07e13ddd3951c231e1fb7cbe2d3d41f7aa) using the (preliminary) ACT support (see pull-request https://github.com/riscv-non-isa/riscv-arch-test/pull/330).


model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,767 @@
...
+union clause ast = RISCV_FLI_H : (bits(5), regidx)
+
+mapping clause encdec = RISCV_FLI_H(rs1, rd)                if (haveFExt() & haveZfa())
+  <-> 0b111_1010 @ 0b00001 @ rs1 @ 0b000 @ rd @ 0b101_0011  if (haveFExt() & haveZfa())
+
+mapping clause assembly = RISCV_FLI_H(constantidx, rd)
+  <-> "fli.h" ^ spc() ^ freg_name(rd) ^ sep() ^ hex_bits_5(constantidx)
+
+function clause execute (RISCV_FLI_H(constantidx, rd)) = {
+  bits : bits(16) = zeros();
```
jrtc27, line 12: let bits = match

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,767 @@
+/* FLI.H */
+
+union clause ast = RISCV_FLI_H : (bits(5), regidx)
+
+mapping clause encdec = RISCV_FLI_H(rs1, rd)                if (haveFExt() & haveZfa())
```
jrtc27, line 5: haveZfhExt?

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,767 @@
...
+union clause ast = RISCV_FLI_S : (bits(5), regidx)
+
+mapping clause encdec = RISCV_FLI_S(rs1, rd)                if (haveFExt() & haveZfa())
+  <-> 0b111_1000 @ 0b00001 @ rs1 @ 0b000 @ rd @ 0b101_0011  if (haveFExt() & haveZfa())
+
+mapping clause assembly = RISCV_FLI_S(constantidx, rd)
+  <-> "fli.s" ^ spc() ^ freg_name(rd) ^ sep() ^ hex_bits_5(constantidx)
+
+function clause execute (RISCV_FLI_S(constantidx, rd)) = {
+  bits : bits(32) = zeros();
```
jrtc27, line 62: let bits

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,767 @@
...
+union clause ast = RISCV_FLI_D : (bits(5), regidx)
+
+mapping clause encdec = RISCV_FLI_D(rs1, rd)                if (haveDExt() & haveZfa())
+  <-> 0b111_1001 @ 0b00001 @ rs1 @ 0b000 @ rd @ 0b101_0011  if (haveDExt() & haveZfa())
+
+mapping clause assembly = RISCV_FLI_D(constantidx, rd)
+  <-> "fli.d" ^ spc() ^ freg_name(rd) ^ sep() ^ hex_bits_5(constantidx)
+
+function clause execute (RISCV_FLI_D(constantidx, rd)) = {
+  bits : bits(64) = zeros();
```
jrtc27, line 112: let bits

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,767 @@
...
+    0b11000 => { bits = 0x41000000 },  /* 8 */
+    0b11001 => { bits = 0x41800000 },  /* 16 */
+    0b11010 => { bits = 0x43000000 },  /* 2^7 */
+    0b11011 => { bits = 0x43800000 },  /* 2^8 */
+    0b11100 => { bits = 0x47000000 },  /* 2^15 */
+    0b11101 => { bits = 0x47800000 },  /* 2^16 */
+    0b11110 => { bits = 0x7f800000 },  /* +inf */
+    0b11111 => { bits = canonical_NaN_S() },
+  };
+  F(rd) = nan_box(bits);
```
jrtc27, line 97: Auto-boxing accessor

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,767 @@
...
+    0b11000 => { bits = 0x4020000000000000 },  /* 8 */
+    0b11001 => { bits = 0x4030000000000000 },  /* 16 */
+    0b11010 => { bits = 0x4060000000000000 },  /* 2^7 */
+    0b11011 => { bits = 0x4070000000000000 },  /* 2^8 */
+    0b11100 => { bits = 0x40e0000000000000 },  /* 2^15 */
+    0b11101 => { bits = 0x40f0000000000000 },  /* 2^16 */
+    0b11110 => { bits = 0x7ff0000000000000 },  /* +inf */
+    0b11111 => { bits = canonical_NaN_D() },
+  };
+  F(rd) = bits;
```
jrtc27, line 147: Auto-boxing

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,767 @@
...
+  RETIRE_SUCCESS
+}
+
+/* FMINM.H */
+
+union clause ast = RISCV_FMINM_H : (regidx, regidx, regidx)
+
+mapping clause encdec = RISCV_FMINM_H(rs2, rs1, rd)     if (haveZfh() & haveZfa())
```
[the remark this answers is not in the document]
ptomsich: Zfa explicitly requires Zfh, F, D or Q ... and does not support Z[fd]inx.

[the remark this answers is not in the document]
jrtc27: That's an odd choice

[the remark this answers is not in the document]
ptomsich: The spec is currently in public review.

The exclusion of Zfinx and Zdinx has been raised there (refer to https://groups.google.com/a/groups.riscv.org/g/isa-dev/c/SED4ntBkabg/m/Wz7Y-A-7AQAJ?utm_medium=email&utm_source=footer&pli=1 for the related discussion).
If the spec changes, we will add support here.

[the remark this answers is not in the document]
ptomsich: > That's an odd choice

Indeed.

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,767 @@
...
+    0b11000 => { bits = 0x4800 },  /* 8 */
+    0b11001 => { bits = 0x4c00 },  /* 16 */
+    0b11010 => { bits = 0x5800 },  /* 2^7 */
+    0b11011 => { bits = 0x5c00 },  /* 2^8 */
+    0b11100 => { bits = 0x7800 },  /* 2^15 */
+    0b11101 => { bits = 0x7c00 },  /* 2^16 */
+    0b11110 => { bits = 0x7c00 },  /* +inf */
+    0b11111 => { bits = canonical_NaN_H() },
+  };
+  F(rd) = nan_box(bits);
```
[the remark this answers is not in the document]
ptomsich: Does this make sense for Zfa, given that in-X is not supported (I am concerned regarding the readability as part of the specifications)?

Or would you prefer us to add new auto-(un)boxing accessors for F (i.e., F_H, F_S, F_D)?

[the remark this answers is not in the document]
jrtc27: Probably we should have non-X variants, though I wouldn't be opposed to just using the X variants everywhere other than for an X<->F move that doesn't make sense on Zfinx, since there's a good chance the instructions will become defined for Zfinx (and there's no reason they shouldn't be).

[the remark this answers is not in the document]
ptomsich: Added the additional accessors w/ auto-(un)boxing.  We can add _or_X accessors back in, if the spec is changed.

model/riscv_fdext_regs.sail
```diff
@@ -270,30 +270,48 @@ function wF_bits(i: bits(5), data: flenbits) -> unit = {
 
 overload F = {rF_bits, wF_bits, rF, wF}
 
+val rF_H : bits(5) -> bits(16) effect {escape, rreg}
+function rF_H(i) = {
+  assert(sizeof(flen) >= 16);
```
jrtc27, line 275: Should probably also assert !zfinx for completeness (especially to catch things like someone adding Zfinx+Zfa but forgetting to update one of the accesses to the right function and trying to debug what on earth went wrong)

model/riscv_fdext_regs.sail
```diff
@@ -270,30 +270,48 @@ function wF_bits(i: bits(5), data: flenbits) -> unit = {
...
+  nan_unbox(F(i))
+}
+
 val rF_or_X_H : bits(5) -> bits(16) effect {escape, rreg}
 function rF_or_X_H(i) = {
   assert(sizeof(flen) >= 16);
   assert(sys_enable_fdext() != sys_enable_zfinx());
   if   sys_enable_fdext()
-  then nan_unbox(F(i))
+  then rF_H(i)
```
jrtc27, line 284: Hoist and use the overloads? The underlying accessors are kind of ugly to use, especially for writes.

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,770 @@
...
+  RETIRE_SUCCESS
+}
+
+/* FLI.D */
+
+union clause ast = RISCV_FLI_D : (bits(5), regidx)
+
+mapping clause encdec = RISCV_FLI_D(rs1, rd)                if (haveDExt() & haveZfa())
```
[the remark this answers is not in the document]
ptomsich: Looks like the current style is without the extra ( ... ), so we'll adapt the Zfa implementation.

model/riscv_fdext_regs.sail
ptomsich: Given that rF_H() needs to assert sizeof(flen) anyway, I would go as far as rewriting this as (consider pseudocode & ignore the indentation):
```
assert(sys_enable_fdext() != sys_enable_zfinx());
if sys_enable_fdext()
then F_H(i)
else 
  {
     assert(sizeof(flen) >= 16);
     X(i)[15..0]
  }
```

Agreed?

jrtc27: What does that achieve? It makes it look like the assertion doesn't need to be true for F_H...

jrtc27: Although, does FLEN actually have a meaning for Zfinx? Arguably it's a bit dodgy to query for Zfinx...

ptomsich: The same assertion lives inside of F_H(), but we can keep it explicit.  It's a matter of taste and I don't care if we have it twice of only once along each path.  So, let's keep it where it is.

That only leave the use of the overloaded F_H() instead of the direct use of rF_H and wF_H.  Personally, I would use the direct accessor rF_H/wF_H on the rF_or_X_H/wF_or_X_H accessors to keep things explicit — then again, it doesn't make a difference.

So to be clear: the requested change is to use the overload instead of the underlying accessor?

jrtc27: Yes; just use F_H rather than [rw]F_H (and therefore move the relevant definitions so they're visible)

jrtc27: I think the problem actually is just that it’s poorly-named and should be maxflen, given how it’s defined and used

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,770 @@
...
+/*
+ * Implement float64 to int32 conversion without saturation, the
+ * result is supplied modulo 2^32.
+ * Rounding mode for FCVTMOD.W.D is always RTZ(Round toward zero).
+ */
+val      fcvtmod_helper : forall 'm, 'm == 64. (bits('m)) -> (bits(5), xlenbits)
+function fcvtmod_helper(x64) = {
+  let (sign, exp, mant) = fsplit_D(x64);
+  fflags		= fcsr.FFLAGS();
+  result : bits(64)	= undefined;
```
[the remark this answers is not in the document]
ptomsich: Hm. Changing this is not as straightforward as it looked. I'll take another look in the coming days to see if this can be expressed more cleanly with a major rewrite of the function.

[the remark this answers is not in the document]
ptomsich: Mind if we kick this can down the road (and deal with it when we reimplement softfloat in pure-SAIL)?

I looked at it and it will be a significant rewrite of the function as we depend on the checks at the end of the if block as a common path...

model/riscv_insts_zfa.sail
```diff
@@ -0,0 +1,753 @@
...
+/* We need rounding mode to be explicitly specified to RTZ(0b001) */
+mapping clause encdec = RISCV_FCVTMOD_W_D(rs1, rd)          if haveDExt() & haveZfa()
+  <-> 0b110_0001 @ 0b01000 @ rs1 @ 0b001 @ rd @ 0b101_0011  if haveDExt() & haveZfa()
+
+mapping clause assembly = RISCV_FCVTMOD_W_D(rs1, rd)
+  <-> "fcvtmod.w.d" ^ spc() ^ reg_name(rd)
+		    ^ spc() ^ freg_name(rs1)
+
+function clause execute(RISCV_FCVTMOD_W_D(rs1, rd)) = {
+  let (fflags, rd_val) = fcvtmod_helper(F_D(rs1));
```
jrtc27, line 749: IMO it's better to keep the register read separate from the function on the result (i.e. revert the change you just made here since my review); that's what gets done consistently(?) elsewhere

ptomsich: I thought that was the intent of your original comment... oh well, I seem to have overcompensated.
That change has been reverted (it was a separate commit anyway) and a new commit that drops empty lines has been added.

model/riscv_insts_zfa.sail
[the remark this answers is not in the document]
ptomsich: We plan to have a rewritten version available by end of this week.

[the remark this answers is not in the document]
ptomsich: @jrtc27 Could you take a peek at the reeimplementatio and let us know if this is good SAIL style (or rather: what should be changed)?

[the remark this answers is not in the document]
ptomsich: @jrtc27 Any thoughts on this?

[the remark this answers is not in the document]
jrtc27: Managed to miss your previous comment; I'll take a look later tonight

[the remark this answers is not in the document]
ptomsich: Thank you!

model/riscv_insts_zfa.sail
```diff
@@ -685,54 +685,57 @@ function clause execute(RISCV_FLTQ_D(rs2, rs1, rd)) = {
 
 /* FCVTMOD.W.D */
 
+/* Calculate the low 32 bits of the integer value from the binary64
+ * floating-point number and return them together with an indicator
+ * whether the NV (overflow) or NX (inexact) exceptions should be raised.
+ */
+val      fcvtmod_helper_inner : forall 'm, 'm >= 0. (bits(1), int('m), bits(53)) -> (bits(5), bits(32))
+function fcvtmod_helper_inner(sign, true_exp, true_mant) = {
+    /* perform the exponentation on the fixed-point
```
jrtc27, line 694: Capitalise the first word of comments

model/riscv_insts_zfa.sail
```diff
@@ -685,54 +685,57 @@ function clause execute(RISCV_FLTQ_D(rs2, rs1, rd)) = {
...
+       mantissa and extract the integer part */
+    let fixedpoint = to_bits(84, unsigned(true_mant)) << true_exp;
```
jrtc27, line 696: Unnecessary round trips through mathematical integers here; should be able to stay in bitvector land

model/riscv_insts_zfa.sail
jrtc27, line 694: Indentation should be 2 not 4

model/riscv_insts_zfa.sail
```diff
@@ -685,54 +685,57 @@ function clause execute(RISCV_FLTQ_D(rs2, rs1, rd)) = {
...
+    let integer    = unsigned(fixedpoint[83..52]);
+    let fractional = to_bits(52, unsigned(fixedpoint[51..0]));
+
+    /* apply the sign bit */
+    let result = if (sign == 0b1) then negate(integer)
```
jrtc27, line 701: Don't need parens for all these conditions

model/riscv_insts_zfa.sail
```diff
@@ -685,54 +685,57 @@ function clause execute(RISCV_FLTQ_D(rs2, rs1, rd)) = {
 
 /* FCVTMOD.W.D */
 
+/* Calculate the low 32 bits of the integer value from the binary64
+ * floating-point number and return them together with an indicator
+ * whether the NV (overflow) or NX (inexact) exceptions should be raised.
+ */
+val      fcvtmod_helper_inner : forall 'm, 'm >= 0. (bits(1), int('m), bits(53)) -> (bits(5), bits(32))
```
jrtc27, line 692: Does nat rather than `int('m)` where `'m >= 0` work?

model/riscv_insts_zfa.sail
```diff
@@ -685,54 +685,57 @@ function clause execute(RISCV_FLTQ_D(rs2, rs1, rd)) = {
...
+    (flags, to_bits(32, result))
+}
+
 /*
  * Implement float64 to int32 conversion without saturation, the
  * result is supplied modulo 2^32.
- * Rounding mode for FCVTMOD.W.D is always RTZ(Round toward zero).
  */
-val      fcvtmod_helper : forall 'm, 'm == 64. (bits('m)) -> (bits(5), xlenbits)
+val      fcvtmod_helper : forall 'm, 'm == 64. (bits('m)) -> (bits(5), bits(32))
```
jrtc27, line 716: Why not just bits(64) instead of `bits('m)` where `'m == 64`?

model/riscv_insts_zfa.sail
```diff
@@ -685,54 +685,57 @@ function clause execute(RISCV_FLTQ_D(rs2, rs1, rd)) = {
...
-
-    let result32b = to_bits(32, unsigned(result));
-    if (to_bits(64, unsigned(result32b)) != result)
-    then fflags = fflags | nvFlag();
-
-    (fflags, EXTS(result[31..0]))
-  };
+
+  /* detect the non-normal cases */
+  let is_subnorm    = (exp == zeros() & mant != zeros());
```
jrtc27, line 721: Don't need parens

model/riscv_insts_zfa.sail
```diff
@@ -685,54 +685,57 @@ function clause execute(RISCV_FLTQ_D(rs2, rs1, rd)) = {
...
+  /* For normal numbers, the mantissa and exponent use special encoding */
+  let true_mant     = 0b1 @ mant;  /* concatenate with the implicit bit */
+  let true_exp      = unsigned(exp) - 1023;   /* undo the offset-binary */
+
+  /* Detect the trivial cases for normal numbers */
+  let is_too_large  = true_exp >= 84;  /* bits will be 'multiplied' out */
+  let is_too_small  = true_exp <  0;   /* bits will be 'divided' out    */
+
+       if (is_zero)       then (zeros(),  zeros())
+  else if (is_subnorm)    then (nxFlag(), zeros())
```
jrtc27, line 734: Could combine subnorm+too_small and nan_or_inf+too_large? At least I don't think that reordering changes semantics. In fact doesn't is_too_large imply is_nan_or_inf and is_subnorm+!is_zero imply is_too_small, i.e. you just need:
```
if (is_zero) then (zeros(), zeros())
else if (is_too_small) (nxFlag(), zeros())
else if (is_too_large) (nvFlag(), zeros())
else fcvtmod_helper_inner(sign, true_exp, true_mant)
```
unless I'm missing something here when glancing through it?

model/riscv_insts_zfa.sail
```diff
@@ -685,54 +685,57 @@ function clause execute(RISCV_FLTQ_D(rs2, rs1, rd)) = {
...
+  else if (is_nan_or_inf) then (nvFlag(), zeros())
+  else if (is_too_large)  then (nvFlag(), zeros())
+  else if (is_too_small)  then (nxFlag(), zeros())
+  else fcvtmod_helper_inner(sign, true_exp, true_mant)
```
jrtc27, line 738: Hm, is there a reason you chose to outline this rather than inline it? Not a huge fan of the "_inner" suffix, may benefit from a better description if kept outlined, something about the number being "regular" (deliberately avoiding the word normal there...).

model/riscv_insts_zfa.sail
ptomsich: While these can be combined, the interpretation of the exp and mantissa bits are different: e.g., for NaN and Inf, the mantissa bits are not interpreted as a mantissa (and the implicit bit doesn't play a role) ... similarly, exp is not decoded using binary-offset to form an exponent.

Given that this is an executable specification, we took the deliberate decision to keep these apart so the checks for the zero/subnorm case do not refer to decoded quantities that are (strictly speaking) not defined for the special cases.

I would expect a (in the far future — or possibly never) sufficiently optimizing compiler/interpreter to combine these cases when executing the specification.

model/riscv_insts_zfa.sail
ptomsich: This is a double-precision FP bitpattern (i.e. binary64) and unrelated to xlenbits.
It could be `bits(64) -> (bits(5), bits(32))`, though.

model/riscv_insts_zfa.sail
ptomsich: Inlined again in the new revision.

ptomsich:
@jrtc27 If you are happy with this now, please hit 'Approve' and I'll merge it (after squashing some of the history).

The copyright header is also missing on this one, but I'd rather clean this up as part of the 'great update of the copyright year and reapplication of the header'.

ptomsich:
Merging this as-is, as I will move the apply_header change forward in a few minutes.

jrtc27:
This history is a total mess. It should not have been merged as-is. Moreover, past PRs have been merged with with rebase and merge, not an actual merge commit.

jrtc27:
(Please treat this the same way as you would treat any other open-source project like LLVM, GCC, binutils, etc, i.e. clean linear history)

ptomsich:
OK to force-push a squash given that it was just a few minutes ago?

jrtc27:
Probably fine, I doubt anyone's pulled it yet

ptomsich:
Fixed. I certainly prefer merging things on the command line over the web UI, as I actually see what I am doing there...

PR #255, opened 2023-04-17
Sail code for Compressed instructions - ZC V1.0.0-RC5.5

junambi:
Pull request for updating compressed instructions.

### DIFF
--- a/model/riscv_insts_zce_ext.sail
+++ b/model/riscv_insts_zce_ext.sail
@@ -1,0 +1,1021 @@
+/*=======================================================================================*/
+/*  RISCV Sail Model                                                                     */
+/*                                                                                       */
+/*  Copyright (c) 2017-2022                                                              */
+/*    Nambi Ju                                                                           */
+/*                                                                                       */
+/*  Redistribution and use in source and binary forms, with or without                   */
+/*  modification, are permitted provided that the following conditions                   */
+/*  are met:                                                                             */
+/*  1. Redistributions of source code must retain the above copyright                    */
+/*     notice, this list of conditions and the following disclaimer.                     */
+/*  2. Redistributions in binary form must reproduce the above copyright                 */
+/*     notice, this list of conditions and the following disclaimer in                   */
+/*     the documentation and/or other materials provided with the                        */
+/*     distribution.                                                                     */
+/*                                                                                       */
+/*  THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS''                   */
+/*  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED                    */
+/*  TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A                      */
+/*  PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR                  */
+/*  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,                         */
+/*  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT                     */
+/*  LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF                     */
+/*  USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND                  */
+/*  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,                   */
+/*  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT                   */
+/*  OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF                   */
+/*  SUCH DAMAGE.                                                                         */
+/*=======================================================================================*/
+
+/* ****************************************************************** */
+/* This file specifies the instructions in the 'ZCEE' extension.      */
+/* ****************************************************************** */
+
+/* ------------------------------------------------------------------ */
+/* Sail Model for the c.l and c.s encoding group                */
+/* ------------------------------------------------------------------ */
+
+/* ****************************************************************** */
+union clause ast = C_LBU : (bits(2), cregidx, cregidx)
+
+//$[wavedrom FUNCT3 _ _ _ _ C0]
+mapping clause encdec_compressed = C_LBU(ui65, rs1, rd)
+  <-> 0b100 @ 0b000 @ rs1 : cregidx @ ui65 : bits(2) @ rd : cregidx @ 0b00
+
+function clause execute (C_LBU(uimm, rsc, rdc)) = {
+  let immext : xlenbits = EXTZ(uimm);
+  let imm : bits(12) = EXTZ(uimm);
+
+  let rd = creg2reg_idx(rdc);
+  let rs = creg2reg_idx(rsc);
+
+  let rsext = X(rs) + immext;
+  X(rs) = rsext;
+
+   match execute(LOAD(imm, rs, rd, false, WORD, false, false)) {
+     RETIRE_SUCCESS => {
+                         X(rd) = EXTZ(X(rd)[7..0]);
+                         RETIRE_SUCCESS
+                       },
+     RETIRE_FAIL => RETIRE_FAIL,
+   }
+}
+
+mapping clause assembly = C_LBU(uimm, rsc, rdc)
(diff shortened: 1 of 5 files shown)


jrtc27:
Why are there two commits with the exact same message

jrtc27:
I see a bunch of commented out code, including some stub implementations. This also puts everything in a Zce file despite there being no such thing as Zce, rather than the individual extra extensions, and enables Zce unconditionally, despite it conflicting with the standard C extension. Please do not put things up for review until they are actually ready; it's a waste of everyone's time.

model/riscv_insts_zce_ext.sail
```diff
@@ -0,0 +1,1021 @@
...
+  let rsext = X(rs) + immext;
+  X(rs) = rsext;
+
+   match execute(LOAD(imm, rs, rd, false, WORD, false, false)) {
+     RETIRE_SUCCESS => {
+                         X(rd) = EXTZ(X(rd)[7..0]);
+                         RETIRE_SUCCESS
+                       },
+     RETIRE_FAIL => RETIRE_FAIL,
+   }
```
jrtc27, line 62: This is clearly nonsense

model/riscv_insts_zce_ext.sail
```diff
@@ -0,0 +1,1021 @@
...
+
+function clause execute (C_LBU(uimm, rsc, rdc)) = {
+  let immext : xlenbits = EXTZ(uimm);
+  let imm : bits(12) = EXTZ(uimm);
+
+  let rd = creg2reg_idx(rdc);
+  let rs = creg2reg_idx(rsc);
+
+  let rsext = X(rs) + immext;
+  X(rs) = rsext;
```
jrtc27, line 54: I don't think this is the right behaviour for these instructions?

model/riscv_insts_zce_ext.sail
```diff
@@ -0,0 +1,1021 @@
+/*=======================================================================================*/
+/*  RISCV Sail Model                                                                     */
+/*                                                                                       */
+/*  Copyright (c) 2017-2022                                                              */
```
arichardson, line 4: Those years look somewhat wrong. Also should be one file per extension.

model/riscv_insts_zce_ext.sail
arichardson, line 62: The existing C instructions (e.g. C_LW) do this in a simpler way - could just reuse that kind of logic?
```suggestion
mapping clause encdec_compressed = C_LBU(ui65, rs1, rd)
  <-> 0b100 @ 0b000 @ rs1 : cregidx @ ui65 : bits(2) @ rd : cregidx @ 0b00

function clause execute (C_LBU(uimm, rsc, rdc)) = {
  let imm : bits(12) = EXTZ(uimm);
  let rd = creg2reg_idx(rdc);
  let rs = creg2reg_idx(rsc);
  execute(LOAD(imm, rs, rd, false, BYTE, false, false))
```

junambi: Thank you for the comments.

I was recommended to adopt this code, after i saw some issues while compiling the Sail code. Could you please let me why this is nonsense, so, i would be able to understand my mistake and also get by better in Sail.

Also, regarding the previous comment on the name for the Zce, i have a separate extension document with Zc as the the document name, so, used Zce.

model/riscv_insts_zce_ext.sail
junambi: Will change them, this is the only file for this extension.

model/riscv_insts_zce_ext.sail
junambi: I did look at the implementation, but, according to my understanding of the specifications, this required the additional steps that i have implemented, could you please clarify my understanding.

tariqkurd-repo:
@abukharmeh are you able to review this?

model/riscv_insts_zce_ext.sail
tariqkurd-repo: @Alasdair I believe coding this "nonsense" was your recommendation?

jrtc27: I highly doubt he recommended implementing "load byte unsigned" as "load word and mask the top bits off"

model/riscv_insts_zce_ext.sail
```diff
@@ -0,0 +1,1021 @@
...
+  
+  let rs2c_lbyte = X(rs2c)[7..0];
+
+  X(rs1c) = rsext;
+  //X(rs2c) = EXTZ(rs2c_lbyte);
+  
+  X(rs2c) = EXTZ(X(rs2c)[7..0]);
+  //mem(rsext[7..0]) = X(rs2c);
+  
+  match execute(STORE(imm, rs2c, rs1c, WORD, false, false)){
```
abukharmeh, line 148: WORD instead of BYTE ? and the match statements feels redundant ?  rs2c_lbyte declared but not used  ?  Comments redundant ? I dont understand the point of rsext ?

model/riscv_insts_zce_ext.sail
```diff
@@ -0,0 +1,1021 @@
...
+  let rs2c = creg2reg_idx(rs2c);
+  let rs1c = creg2reg_idx(rs1c);
+
+  let rsext = X(rs1c) + immext;
+  //let rs2c_lbyte = X(rs2c)[15..0];
+
+  X(rs2c) = EXTZ(X(rs2c)[15..0]);
+
+  // MemValue(rsext[15..0]) = X(rs2c);
+  match execute(STORE(imm, rs2c, rs1c, WORD, false, false)){
```
abukharmeh, line 176: WORD instead of HALF ? same as the comments on C_SB ?

model/riscv_insts_zce_ext.sail
jrtc27: What additional/different steps do you think you need, and can you point at where in the specification it says you need them? I agree with Alex.

model/riscv_insts_zce_ext.sail
jrtc27: Yes, exactly, one of many implementations in this PR full of nonsense. With the way the current PR stands, finding and pointing out all the problems would likely take longer than just reimplementing it properly from scratch, so offering a full code review is a waste of reviewers' time.

model/riscv_insts_zce_ext.sail
```diff
@@ -0,0 +1,1021 @@
...
+  <-> 0b100 @ 0b111 @ rsd : cregidx @ 0b00 @ rs2 : cregidx @ 0b01
+
+function clause execute (C_MUL(rsd,rs2)) = {
+  let rd = creg2reg_idx(rsd);
+  let rs = creg2reg_idx(rs2);
+  
+  let result_wide = to_bits(2 * sizeof(xlen), signed(X(rd)) * signed(X(rs)));
+  X(rd) = result_wide[(sizeof(xlen) - 1) .. 0];
+  RETIRE_SUCCESS
+}
```
abukharmeh, line 294: Why don't you map this to the existing implementation of mul in a similar way to add c.add and other existing existing compressed instructions ? https://github.com/riscv/sail-riscv/blob/72b2516d10d472ac77482fd959a9401ce3487f60/model/riscv_insts_cext.sail#L619

abukharmeh:
It looks like all sequenced instructions are not implemented.

All instructions starting with CM have encodings that conflicts with existing floating points loads/ stores thus we need to make sure that they are enabled separately from other extensions of Zc*.

Zce is just an alias to enable all the extensions, I think we should have an enable flags for each of the extensions `Zca, Zcf, Zcd, Zcb, Zcmp and Zcmt` and assert the combinations that are not allowed !

junambi:
Ibrahim,

Could you please suggest a time to discuss some of the issues with my understanding of the specifications and implementation.

Looking forward to hearing from you,

Thank you.

model/riscv_insts_zce_ext.sail
junambi: I need to add the zero extended value of immediate to the contents of the source register and then put it into the destination, is my understanding from the spec.

model/riscv_insts_zce_ext.sail
junambi: If i do not add the match and just do an execute, then Sail throws an error, so the match was used.
Regarding the commented out line, that was the previous implementation that i had done, but, now is done with the STORE.

I did look for the BYTE Store, looks like i did not look for it properly.

model/riscv_insts_zce_ext.sail
junambi: Same comments as above

tariqkurd-repo:
Hi @junambi , maybe close this PR and start with one for Zcb first? then we can get that one right and move forward to the more complex ones? Zcb is the simplest by far.

junambi:
Request closed as per post above, will open new one for Zcb.

junambi:
Hi Tariq,

As per suggestion, i have closed the previous PR, split the file into three and created another PR. The following is the detail.

Review of ZCE extensions - Sail Code V 0.5 #258

Please review and comment.

Thank you,
Regards,
Nambi

Sent with [Proton Mail](https://proton.me/) secure email.

------- Original Message -------

PR #260, opened 2023-05-10
Svadu support

adlr:
Draft change for Svadu support. I appreciate any feedback on this change. Thank you!

No tests were broken in the `test/run_tests.sh` suite.


model/riscv_sys_regs.sail
```diff
@@ -567,6 +567,26 @@ function retire_instruction() -> unit = {
...
+  HADE : 61
+  /* Fill in other fields when adding support for them */
+}
+register menvcfg : MEnvCfg
+
+function legalize_menvcfg(c : MEnvCfg, high : bool, value : xlenbits) -> MEnvCfg = {
+  if high == false & sizeof(xlen) == 32 then
+    c
+  else {
+    // If 32bit, then high is true
```
jrtc27, line 580: Comments are C-style

model/riscv_sys_regs.sail
```diff
@@ -567,6 +567,26 @@ function retire_instruction() -> unit = {
...
 }
 
+bitfield MEnvCfg : bits(64) = {
+  HADE : 61
+  /* Fill in other fields when adding support for them */
+}
+register menvcfg : MEnvCfg
+
+function legalize_menvcfg(c : MEnvCfg, high : bool, value : xlenbits) -> MEnvCfg = {
+  if high == false & sizeof(xlen) == 32 then
```
jrtc27, line 577: not(...)

model/riscv_sys_regs.sail
```diff
@@ -567,6 +567,26 @@ function retire_instruction() -> unit = {
...
+    c
+  else {
+    // If 32bit, then high is true
+    let full : bits(64) = match (sizeof(xlen)) {
+      64 => value,
+      32 => value @ 0x0000_0000
+    };
+    let c = update_HADE(c, [full[61]]);
+    c
```
Alasdair, line 586: I think this `let c = ... in c` could just be `update_HADE(c, [full[61])`

model/riscv_sys_regs.sail
```diff
@@ -567,6 +567,26 @@ function retire_instruction() -> unit = {
...
+  /* Fill in other fields when adding support for them */
+}
+register menvcfg : MEnvCfg
+
+function legalize_menvcfg(c : MEnvCfg, high : bool, value : xlenbits) -> MEnvCfg = {
+  if high == false & sizeof(xlen) == 32 then
+    c
+  else {
+    // If 32bit, then high is true
+    let full : bits(64) = match (sizeof(xlen)) {
```
Alasdair, line 581: I think the brackets around sizeof are not-needed

model/riscv_sys_regs.sail
jrtc27, line 586: I feel like there must be a better way to write all this... is this the first *h CSR we've implemented that's not just a stub 0?

model/riscv_vmem_sv32.sail
```diff
@@ -214,6 +210,9 @@ function translate32(asid, ptb, vAddr, ac, priv, mxr, do_sum, level, ext_ptw) =
                   MemException(e) => internal_error(__FILE__, __LINE__, "invalid physical address in TLB")
                 };
                 TR_Address(ent.pAddr | EXTZ(vAddr & ent.vAddrMask), ext_ptw)
+              } else {
```
jrtc27, line 213: Inverting these is a little gratuitous, though perhaps it makes more sense

Alasdair:
I left a couple of comments but mostly looks good to me 👍

model/riscv_vmem_tlb.sail
```diff
@@ -120,3 +120,7 @@ function flush_TLB_Entry(e, asid, addr) = {
                            & not(e.global))
   }
 }
+
+function should_update_PTE_bits() -> bool = {
+  plat_enable_dirty_update() | (plat_enable_svadu() & menvcfg.HADE() == 0b1)
+}
```
jrtc27, line 126: Missing newline

jrtc27:
> I left a couple of comments but mostly looks good to me 👍

Your comments appear etc be missing

model/riscv_vmem_sv48.sail
```diff
@@ -188,13 +188,9 @@ function translate48(asid, ptb, vAddr, ac, priv, mxr, do_sum, level, ext_ptw) =
           TR_Address(pAddr, ext_ptw)
         },
         Some(pbits, ext) =>
-          if not(plat_enable_dirty_update())
-          then {
-            /* pte needs dirty/accessed update but that is not enabled */
-            TR_Failure(PTW_PTE_Update(), ext_ptw)
-          } else {
+          if should_update_PTE_bits() then {
```
arichardson, line 191: Is there any reason to swap branch order here instead of using `not(should_update_PTE_bits())`? Not objecting to this just wondering why.

Alasdair: I think it's slightly nicer to avoid the not

Alasdair:
I'm not very familiar with the RISC-V address translation scheme, but looking a bit deeper at the functionality, I think you also need to update the also implement the lookup case for a TLB miss (the None() branch with the `if not(plat_enable_dirty_update())` at line 232 in `riscv_vmem_sv32.sail` (and similarly for the other sizes). It looks like the 'TLB' is essentially an emulation optimization (which every architectural model ends up with for reasonable performance), so the real architectural behavior is really specified by walk32.

Obviously not a task for yourself, but I don't think we need multiple files for the differently sized translations - unless there is significantly differing behaviour I think it could be a single size-polymorphic function. I'm also not sure we are implementing step 7 of the translation process as written in the priv spec, at least I'm not seeing how that is done in the code? I'm wondering about this because the Svadu extension seems to modify this check slightly to avoid a page fault, but I don't see where that is being done.

model/riscv_sys_regs.sail
adlr: Changed to C++ style

model/riscv_sys_regs.sail
adlr: I can't say I'm anywhere close to a SAIL expert. I'm all ears on suggestions here. Thank you!

model/riscv_vmem_sv32.sail
adlr: Changed back the if/then ordering

model/riscv_vmem_tlb.sail
adlr: Added newline back

model/riscv_vmem_sv48.sail
adlr: I changed it back, figuring that without a clear consensus, I'd minimize the diff.

model/riscv_sys_regs.sail
adlr: I tried this, but got an error. So, leaving the `let c =` in

adlr:
> I'm not very familiar with the RISC-V address translation scheme, but looking a bit deeper at the functionality, I think you also need to update the lookup case for a TLB miss (the None() branch with the `if not(plat_enable_dirty_update())` at line 232 in `riscv_vmem_sv32.sail` (and similarly for the other sizes). It looks like the 'TLB' is essentially an emulation optimization (which every architectural model ends up with for reasonable performance), so the real architectural behavior is really specified by walk32.
> 
> Obviously not a task for yourself, but I don't think we need multiple files for the differently sized translations - unless there is significantly differing behaviour I think it could be a single size-polymorphic function. I'm also not sure we are implementing step 7 of the translation process as written in the priv spec, at least I'm not seeing how that is done in the code? I'm wondering about this because the Svadu extension seems to modify this check slightly to avoid a page fault, but I don't see where that is being done.
> 
> Edit:
> 
> After some more reading it seems like the A/D bits update is being done in update_PTE_bits in riscv_pte.sail. I am wondering if that needs to be updated (specifically the check on line 147) for this paragraph in the Svadu document?
> 
> > The translation of virtual addresses (or guest physical addresses) to physical (or guest physical)
> > addresses is accomplished with the same algorithm as specified in the Supervisor-Level ISA
> > extension (section 4.3.2) and as modified by the hypervisor extension (section 8.5.1), except that
> > step 7 of the translation process, instead of causing a page-fault exception due to A and/or D bits
> > being 0 when required to be 1, continues as follows:

I think we've adjusted the code to handle the additional cases Alasdair raised.

Thanks very much for the reviews!

model/riscv_vmem_tlb.sail
```diff
@@ -120,3 +120,7 @@ function flush_TLB_Entry(e, asid, addr) = {
                            & not(e.global))
   }
 }
+
+function should_update_PTE_bits() -> bool = {
+  plat_enable_dirty_update() | (plat_enable_svadu() & menvcfg.HADE() == 0b1)
```
arichardson, line 125: I wonder if the logic should be changed such that `enable-dirty-update` implicitly enables svadu and sets the initial value of menvcfg.HADE to 1? In that case this function could be simplified to `plat_enable_svadu() & menvcfg.HADE() == 0b1`. I believe the --enable-dirty-update flag was only added because there was no Svadu at the time and some operating systems expect the PTE updates.

jrtc27: No, because then you can't configure the model to be an implementation that does A/D updates but doesn't have Svadu, a privileged spec option which was ratified years ago.

arichardson: That is true, but looking at this PR there are no observable behaviour changes for plat_enable_svadu()/not(plat_enable_svadu()) when plat_enable_dirty_update() is also true? Does this mean we need to also guard write access to menvcfg.HADE by plat_enable_svadu()? Those older implementations should have that bit hardwired to zero (well most likely not have the register at all)?

jrtc27: Yes, HADE shouldn't exist when Svadu isn't enabled, and needs to be legalised as 0

model/riscv_sys_regs.sail
jrtc27: Please show the error; it should work just fine

model/riscv_sys_regs.sail
```diff
@@ -567,6 +567,26 @@ function retire_instruction() -> unit = {
...
+function legalize_menvcfg(c : MEnvCfg, high : bool, value : xlenbits) -> MEnvCfg = {
+  if not(high) & sizeof(xlen) == 32 then
+    c
+  else {
+    /* If 32bit, then high is true */
+    let full : bits(64) = match sizeof(xlen) {
+      64 => value,
+      32 => value @ 0x0000_0000
+    };
+    let c = update_HADE(c, [full[61]]);
```
arichardson, line 585: Looking at the Svadu spec I see: `When menvcfg.HADE is zero, the implementation behaves as though Svadu were not implemented. If Svadu is not implemented, menvcfg.HADE is read-only zero.` so I believe this needs to be guarded by plat_enable_svadu()

model/riscv_vmem_tlb.sail
arichardson: That makes sense and I agree we should have two separate options to allow matching those implementations.

model/riscv_sys_regs.sail
adlr:
```
Type error:
model/riscv_sys_regs.sail:585.4-30:
585 |    update_HADE(c, [full[61]]);
    |    ^------------------------^
    | No overloading for update_HADE, tried:
    | * _update_MEnvCfg_HADE
    |    MEnvCfg is not a subtype of unit
```

Any help appreciated!

bacam: You need to remove the semicolon at the end because you're returning the value from `update_HADE`.

Alasdair: Did you try:
```
update_HADE(c, [full[61]]);
c
```
Because that would result in the error you posted, what I meant was:
```
update_HADE(c, [full[61]])
```
As the last expression in the block

adlr: That did it! Seems obvious now, sorry. Will post a new patch set later, just wanted to quickly share that this did work as you suggested. Thanks!

model/riscv_sys_regs.sail
adlr: Thanks! I needed to move `plat_enable_svadu()` out of `riscv_platform.sail` and added a note similar to the one by `get_16_random_bits` explaining it, but I made this change

model/riscv_sys_regs.sail
```diff
@@ -567,6 +567,38 @@ function retire_instruction() -> unit = {
...
+function legalize_menvcfg(c : MEnvCfg, high : bool, value : xlenbits) -> MEnvCfg = {
+  if not(high) & sizeof(xlen) == 32 then
+    c
+  else {
+    /* If 32bit, then high is true */
+    let full : bits(64) = match sizeof(xlen) {
+      64 => value,
+      32 => value @ 0x0000_0000
+    };
+    if not(plat_enable_svadu())
```
arichardson, line 596: Isn't this inverted? We should be able to toggle HADE if svadu is enabled.

adlr: Good catch. I inverted this and cleaned up the `legalize_menvcfg` function. Thank you!

model/riscv_sys_regs.sail
```diff
@@ -573,6 +573,37 @@ function retire_instruction() -> unit = {
...
+  /* Fill in other fields when adding support for them */
+}
+register menvcfg : MEnvCfg
+
+function legalize_menvcfg_high(c : MEnvCfg, value : bits(32)) -> MEnvCfg = {
+  /* Subtract 32 from bit offsets as we only have high half here */
+  if plat_enable_svadu()
+  then update_HADE(c, [value[61 - 32]])
+  else c
+}
```
arichardson, line 598:
```suggestion
function legalize_menvcfg_high(c : MEnvCfg, value : bits(32)) -> MEnvCfg = {
  let newval = Mk_MEnvCfg(value @ to_bits(32, 0));
  if plat_enable_svadu()
  then update_HADE(c, newval.HADE())
  else c
}
```

How about the following to avoid the subtraction (which will become awkward once we add more fields here.

adlr: Thanks! Made the edits and credited you

ved-rivos:
@adlr - the extension name was changed from Svadu to Svhad. The fields in CSRs continue to be HADE. Just extension name changed.

jordancarlin:
Closing now that #1599 is merged.

PR #263, opened 2023-05-17
Alternate clang format config

arichardson:
As suggested in https://github.com/riscv/sail-riscv/pull/246#issuecomment-1495944131


arichardson:
Closing this since the diff-reducing format is preferable.

PR #265, opened 2023-05-23
Smcntrpmf support in SAIL

adlr:
Please take a look and let me know if you have any suggested changes for this.

`test/run_tests.sh` says, `Passed 712 out of 712`

Thanks!


model/riscv_platform.sail
```diff
@@ -319,7 +319,7 @@ function clint_store(addr, width, data) = {
 val tick_clock : unit -> unit effect {rreg, wreg}
 function tick_clock() = {
   if   mcountinhibit.CY() == 0b0
```
jrtc27, line 321: Why not fold the CSR checks into the functions?

model/riscv_sys_regs.sail
```diff
@@ -560,11 +607,11 @@ register mtime : bits(64)
 register minstret : bits(64)
 register minstret_written : bool
 
-function retire_instruction() -> unit = {
+function retire_instruction(orig_priv : Privilege) -> unit = {
```
jrtc27, line 610: priv is fine, it's clearly referring to the instruction being retired

model/riscv_sys_regs.sail
```diff
@@ -543,6 +543,53 @@ function legalize_mcountinhibit(c : Counterin, v : xlenbits) -> Counterin = {
...
+  match (orig_priv) {
+    Machine    => minstretcfg.MINH() != 0b1,
+    Supervisor => minstretcfg.SINH() != 0b1,
+    User       => minstretcfg.UINH() != 0b1
+    /* TODO: VSINH, VUINH when those modes are defined */
+  }
+}
+
+function legalize_smcntrpmf(c : CountSmcntrpmf, high: bool, value : xlenbits) -> CountSmcntrpmf = {
+  if high == false & sizeof(xlen) == 32 then
```
jrtc27, line 576: We probably should give more thought into a nicer way to handle all these *h CSRs being added

model/riscv_sys_regs.sail
```diff
@@ -560,11 +607,11 @@ register mtime : bits(64)
 register minstret : bits(64)
 register minstret_written : bool
 
-function retire_instruction() -> unit = {
+function retire_instruction(orig_priv : Privilege) -> unit = {
   if   minstret_written == true
   then minstret_written = false
   else if mcountinhibit.IR() == 0b0
```
arichardson, line 613:
```suggestion
  else if should_inc_minstret(priv)
```
and move the check for mcountinhibit.IR() into should_inc_minstret?

jrtc27: Yeah, I said as much in my review for tick_clock

model/riscv_platform.sail
adlr: Done

model/riscv_sys_regs.sail
adlr: I wasn't sure what you mean by this. Are you saying that this function can just look at the current privilege, therefore needs no argument? If you can be more specific I will try to incorporate your suggestion. Thanks very much!

model/riscv_sys_regs.sail
adlr: Now that I have a bit more SAIL experience, I took a stab at doing something better here. Please take a look.

model/riscv_sys_regs.sail
adlr: Done

model/riscv_sys_regs.sail
```diff
@@ -543,6 +543,61 @@ function legalize_mcountinhibit(c : Counterin, v : xlenbits) -> Counterin = {
...
+  c
+}
+
+function legalize_smcntrpmf(c : CountSmcntrpmf, value : xlenbits) -> CountSmcntrpmf = {
+  if sizeof(xlen) == 32 then
+  legalize_smcntrpmf_low(c, value)
+  else {
+    let c = legalize_smcntrpmf_high(c, value[63..32]);
+    legalize_smcntrpmf_low(c, value[31..0])
+  }
```
arichardson, line 598: Does something like the following work? Not checked if the syntax is valid.

```suggestion
  let c = if sizeof(xlen) > 32 then legalize_smcntrpmf_high(c, value[63..32]) else c
  legalize_smcntrpmf_low(c, value[31..0])
```

arichardson (commented):
Generally looks good to me - a few suggestions.

model/riscv_sys_regs.sail
```diff
@@ -543,6 +543,57 @@ function legalize_mcountinhibit(c : Counterin, v : xlenbits) -> Counterin = {
...
+bitfield CountSmcntrpmf : bits(64) = {
+  MINH  : 62,
+  SINH  : 61,
+  UINH  : 60,
+  VSINH : 59,
+  VUINH : 58
+}
+register mcyclecfg : CountSmcntrpmf
+
+function should_inc_mcycle() -> bool = {
```
arichardson, line 555: I wonder if this should take a `priv` argument as well for consistency (since it's not immediately obvious if it's called before or after mode changes take effect).

arichardson (approved):
This looks good to me but someone else will have to give the final approval :)

model/riscv_sys_regs.sail
jrtc27: Agreed

jrtc27 (changes requested):
See review comments

KotorinMinami:
So this PR hasn't been updated for over a year. @adlr @jrtc27 @arichardson Do you mind if I take it over and continue to push for its integration?

arichardson:
> So this PR hasn't been updated for over a year and Smcntpmf have been ratified. @adlr @jrtc27 @arichardson Do you mind if I take it over and continue to push for its integration?

Please do! It would be great to catch up with all recent extensions

adlr:
> > So this PR hasn't been updated for over a year and Smcntpmf have been ratified. @adlr @jrtc27 @arichardson Do you mind if I take it over and continue to push for its integration?
> 
> Please do! It would be great to catch up with all recent extensions

+1 from me as well. Please go for it!

KotorinMinami:
Since #690 has been merged, maybe we need to close the PR or do some other clean-up work?

PR #266, opened 2023-05-24
Update copyright headers

ptomsich:
The LICENCE had gotten out-of-sync (copyright year, missing myself and VRULL as contributors) and some files missed the copyright headers.  This updates the LICENCE file and reruns apply_headers.


LICENCE
```diff
@@ -22,6 +22,8 @@ Copyright (c) 2017-2021
   Microsoft, for contributions by Robert Norton-Wright and Nathaniel Wesley Filardo
   Peter Rugg
   Aril Computer Corp., for contributions by Scott Johnson
+  Philipp Tomsich
+  VRULL GmbH, for contributions by its employees
```
jrtc27, line 26: Are you covered by the latter or not? Also don't know if "its employees" is sufficient here, so far everyone has been explicitly named

ptomsich: Technically not, as control contravenes employment in our geography.
We could make it "employees and contractors", though...

ptomsich:
Rebased onto latest  master (incl. the squashed Zfa).

@jrtc27 OK to merge with the three individual commits (i.e., w/o squashing the first two into a single update of LICENCE)?

jrtc27:
That breakdown makes sense to me, so if the copyright is correct then yes

PR #274, opened 2023-06-16
Allow loading more than one ELF binary

arichardson:
This makes it easier to use a separate M-mode bootloader and kernel payload (e.g. OpenSBI fw_jump). It also makes it easier to test booting systems such as FreeBSD without bundling the kernel with the bootloader.

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -217,7 +217,11 @@ static void read_dtb(const char *path)
   fprintf(stdout, "Read %zd bytes of DTB from %s.\n", dtb_len, path);
 }
 
-char *process_args(int argc, char **argv)
+/**
+ * Parses the command line arguments and returns the argv index for the ELF file
+ * that should be loaded.
+ */
+static int process_args(int argc, char **argv)
 {
   int c;
   uint64_t ram_size = 0;
@@ -375,7 +379,7 @@ char *process_args(int argc, char **argv)
   if (!rvfi_dii)
 #endif
     fprintf(stdout, "Running file %s.\n", argv[optind]);
-  return argv[optind];
+  return optind;
 }
 
 void check_elf(bool is32bit)
@@ -1022,7 +1026,8 @@ int main(int argc, char **argv)
   // Initialize model so that we can check or report its architecture.
   preinit_sail();
 
-  char *file = process_args(argc, argv);
+  int files_start = process_args(argc, argv);
+  char *file = argv[files_start];
   init_logs();
 
   if (gettimeofday(&init_start, NULL) < 0) {
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -142,12 +142,9 @@ static struct option options[] = {
 
 static void print_usage(const char *argv0, int ec)
 {
+  fprintf(stdout, "Usage: %s [options] <elf_file> [<elf_file> ...]\n", argv0);
 #ifdef RVFI_DII
-  fprintf(stdout,
-          "Usage: %s [options] <elf_file>\n       %s [options] -r <port>\n",
-          argv0, argv0);
-#else
-  fprintf(stdout, "Usage: %s [options] <elf_file>\n", argv0);
+  fprintf(stdout, "       %s [options] -r <port>\n", argv0);
 #endif
   struct option *opt = options;
   while (opt->name) {
@@ -218,8 +215,11 @@ static void read_dtb(const char *path)
 }
 
 /**
- * Parses the command line arguments and returns the argv index for the ELF file
- * that should be loaded.
+ * Parses the command line arguments and returns the argv index for the first
+ * ELF file that should be loaded. As getopt transforms the argv array, all
+ * argv values following that index are non-options and can be treated as
+ * additional ELF files that should be loaded into memory (but not scanned
+ * for the magic tohost/{begin,end}_signature symbols).
  */
 static int process_args(int argc, char **argv)
 {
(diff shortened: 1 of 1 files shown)


arichardson:
ping?

c_emulator/riscv_sim.c
```diff
@@ -1088,15 +1089,23 @@ int main(int argc, char **argv)
...
+    fprintf(stdout, "Loading additional ELF file %s.\n", argv[i]);
+    load_elf(argv[i], &is32bit, &entry);
+    check_elf(is32bit);
+  }
 
   /* initialize spike before sail so that we can access the device-tree blob,
    * until we roll our own.
    */
-  init_spike(file, entry, rv_ram_size);
+  init_spike(file0, entry, rv_ram_size);
```
billmcspadden-riscv, line 1108: Does Spike have the ability to load multiple .elf files?   If so,  shouldn't this also be put into a "for" loop?

billmcspadden-riscv (commented):
1. The print_usage() function needs to be written so that it shows that multiple .elf files can be loaded.  Perhaps: "Usage %s [options] <elf_file> [<elf_file> ...]"
2. How was this tested?   Is there a test we can add into the test directory that checks out this functionality?
3. Does this adversely affect the processing of the RVFI arguments?  I'm not quite sure how the RVFI options are to be used.

arichardson:
1. Good point, will update the help message
2. I tested this booting FreeBSD (and CheriBSD) using risc-pk bbl as the first elf and the FreeBSD kernel as the second. The boot takes an extremely long time so I'm not sure this can be added to tests (especially since it doesn't shut down after boot but rather waits for login).
3. It will not have any effect on those arguments.

c_emulator/riscv_sim.c
arichardson: I have never used the init_spike logic, we may need to call `tv_load_elf(s, f);` for each of the files. Maybe @pmundkur can comment on this.

jrtc27:
https://github.com/SRI-CSL/l3riscv hasn't been updated since 2019, it'll have no way of using https://github.com/riscv-software-src/riscv-isa-sim/pull/390. Though to be honest the whole thing should probably be cut in favour of some other tandem verification approach; we have RVFI and we have the various official RISC-V testing frameworks. We don't need some ad-hoc one that's specifically for Sail-vs-Spike and nothing else.

c_emulator/riscv_sim.c
```diff
@@ -217,7 +217,7 @@ static void read_dtb(const char *path)
   fprintf(stdout, "Read %zd bytes of DTB from %s.\n", dtb_len, path);
 }
 
-char *process_args(int argc, char **argv)
+int process_args(int argc, char **argv)
```
ptomsich, line 220: This change refactors the process_args signature (although motivated by the new use-case).
Please split off into a separate, standalone commit (together with the change below that changes the usage of process_args).

Could you also add a comment above the process_args function to document its contract?

c_emulator/riscv_sim.c
```diff
@@ -1022,7 +1022,8 @@ int main(int argc, char **argv)
   // Initialize model so that we can check or report its architecture.
   preinit_sail();
 
-  char *file = process_args(argc, argv);
+  int files_start = process_args(argc, argv);
+  char *file0 = argv[files_start];
```
ptomsich, line 1026: Please use a better name than 'file0' for this variable (and possibly also add a comment).  It is not clear from the name (i.e., without reading the rest of the change) that this is assumed to be the filename of the executable (and that any further filenames should be 'additional' ELF files).

ptomsich (changes requested):
Except for the requested style changes, I am happy with this.

c_emulator/riscv_sim.c
arichardson: Sure, how about 'initial_file' or 'main_file'?

arichardson:
I believe I've addressed all outstanding feedback.

c_emulator/riscv_sim.c
```diff
@@ -1088,15 +1093,29 @@ int main(int argc, char **argv)
...
+    entry = load_sail(initial_elf_file);
 #else
-  uint64_t entry = load_sail(file);
+  uint64_t entry = load_sail(initial_elf_file);
 #endif
+  /* Load any additional ELF files into memory */
+  for (int i = files_start + 1; i < argc; i++) {
+    bool is32bit;
+    uint64_t additional_entry;
+    fprintf(stdout, "Loading additional ELF file %s.\n", argv[i]);
```
jrtc27, line 1104: Hm, we don't print the main file, seems a bit odd to do it for these

c_emulator/riscv_sim.c
```diff
@@ -1088,15 +1093,29 @@ int main(int argc, char **argv)
...
+    /*
+     * We use load_elf() directly instead of load_sail() as we do not want to
+     * override tohost/{begin,end}_signature from the initial ELF file.
```
jrtc27, line 1107: Why not add a bool to load_sail that makes it early return after the loading?

c_emulator/riscv_sim.c
```diff
@@ -1088,15 +1093,29 @@ int main(int argc, char **argv)
...
+     */
+    load_elf(argv[i], &is32bit, &additional_entry);
+    check_elf(is32bit);
+    fprintf(stdout, "Additional ELF %d Entry @ 0x%" PRIx64 "\n",
+            i - files_start, additional_entry);
```
jrtc27, line 1112: This isn't meaningful to print, the Sail model isn't doing anything with it, and likely neither are the other ELF files

c_emulator/riscv_sim.c
arichardson: Done. Thanks for the suggestion, makes it cleaner.

c_emulator/riscv_sim.c
arichardson: Fair enough - I thought it could be helpful to print the S-mode entry point, but I've removed it now. You can always use `readelf` to get that info...

c_emulator/riscv_sim.c
arichardson: We print the entry point for the main file. I can remove this, just thought it might be helpful.

c_emulator/riscv_sim.c
jrtc27: For that specific use case it'll be virtual rather than physical anyway, which isn't much help until you're already there

c_emulator/riscv_sim.c
jrtc27: I'd missed that process_args prints it

arichardson:
ping?

c_emulator/riscv_sim.c
billmcspadden-riscv: We should not slow this down for Spike.

arichardson:
monthly ping.

jrtc27:
This is absurd. This should have been merged months ago.

PR #275, opened 2023-06-16
Allow redirecting trace output to a file

arichardson:
This is useful when booting an operating system with tracing enabled
as it allows showing the printed messages on the terminal and storing
the trace log to a separate file.

Example usage:
`./c_emulator/riscv_sim_RV64 -b os-boot/rv64-64mb.dtb -t /dev/stderr os-boot/linux-rv64-64mb.bbl --trace-output /tmp/linux.log`

This PR also includes a second commit to fix --help for options without a corresponding short option (rebase and merge preferred over squash and merge).

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -48,10 +48,14 @@ const char *RV32ISA = "RV32IMAC";
 #define CSR_MTVAL 0x343
 #define CSR_MIP 0x344
 
+#define OPT_TRACE_OUTPUT 1000
+
 static bool do_dump_dts = false;
 static bool do_show_times = false;
 struct tv_spike_t *s = NULL;
 char *term_log = NULL;
+static const char *trace_log_path = NULL;
+FILE *trace_log = NULL;
 char *dtb_file = NULL;
 unsigned char *dtb = NULL;
 size_t dtb_len = 0;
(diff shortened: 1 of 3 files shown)


arichardson:
ping?

Alasdair:
This looks good to me, I don't think there are any reasons not to merge this. The only problem I could foresee is if you don't use these printing functions and use the ones in the Sail standard library they won't be directed.

c_emulator/riscv_sim.c
billmcspadden-riscv, line 1: Per Alasdair's comments,  we need to have a description as to what is and isn't printed to the file.
Based on his comments,  there will be output that will not go into the log file.

Additionally,  what should happen to output going to stderr?   Should that also go into the log file?

Also,  should output go to *both* the log file and stdout (and stderr)?

arichardson: As discussed, the trace log will contain all categorized trace messages (instructions, register updates, etc.) while the existing prints (e.g. UART) will remain. Ideally the sail model would not use the uncategorized print() at all but for now those will just remain unchanged.

arichardson:
I believe I've addressed all outstanding comments, and I've updated the commit messages to explain which messages are redirected:

```
csim: Allow redirecting trace output to a file using command line flag

This is useful when booting an operating system with tracing enabled
as it allows showing the printed messages on the terminal and storing
the trace log to a separate file. All categorized trace messages are
now redirected to the --trace-output file when passed. If this option is
not gived, the current behaviour of using stdout is retained.

Example usage:
`./c_emulator/riscv_sim_RV64 -b os-boot/rv64-64mb.dtb --trace-output /dev/stderr os-boot/linux-rv64-64mb.bbl --trace-output /tmp/linux.log`
```

I've made these commits logically separate, so please rebase and merge rather than squashing the commits.

arichardson:
ping?

c_emulator/riscv_sim.c
billmcspadden-riscv: This all sounds good to me.

PR #277, opened 2023-06-25
Add Sail asciidoc documentation target to Makefile

Alasdair:
We have an Asciidoctor plugin that allows incorporating Sail code into ISA manuals written using asciidoc. See https://github.com/Alasdair/asciidoctor-sail. This requires generating a json file using Sail that indexes the source code in the repository. This commit simply adds a target for creating that to the Makefile.

Currently requires the development version of Sail, but I am working on making a new opam release of Sail for the documentation support, and there should be no harm in getting the Makefile target here set up in advance, or for those who are already using a development build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,7 @@ SAIL_COQ_SRCS  = $(addprefix model/,$(SAIL_ARCH_SRCS) $(SAIL_SEQ_INST_SRCS) $(SA
 PLATFORM_OCAML_SRCS = $(addprefix ocaml_emulator/,platform.ml platform_impl.ml softfloat.ml riscv_ocaml_sim.ml)
 
 SAIL_FLAGS += -dno_cast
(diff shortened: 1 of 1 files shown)


jrtc27:
Should this not be per-ARCH? Which files are included varies between RV32 and RV64. Even if technically the set of instruction sources is the same.

Alasdair:
Updated to be per-arch. Possibly not that useful because the code is mostly the same, other than the virtual memory code. I think that code could probably be merged so it's parametric and therefore not per-arch, but not sure when I'd have the time to do that.

jrtc27:
Agreed, and you’d need that anyway for mutable S-mode XLEN

billmcspadden-riscv (approved):
Looks good to me.

billmcspadden-riscv:
Merge approved at the tech-golden-model meeting on 2023-06-27

PR #279, opened 2023-06-29
Rename EXTZ and EXTS

Alasdair:
I've been working on the Sail RISC-V documentation stuff a bit this week and I feel like the EXTS and EXTZ names are a bit ugly (and visually too similar!) and they look a bit odd when placed next to the prose descriptions in the spec.

Rename EXTZ to zero_extend and EXTS to sign_extend. Two main reasons for doing this: First, it means that the source more closely aligns with the prose descriptions in the documentation with more readable names. Second, EXTS and EXTZ are visually very close to each other with just the S and Z, they are also following an odd convention where they are ALLCAPS rather than snake_case like other functions in the spec.

I think this convention comes from early Power specs in Sail, which influenced Sail MIPS and CHERI-MIPS, but I don't think it's a very good convention we should be keeping in sail-riscv

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -187,11 +187,11 @@ function get_config_print_reg () = false
 function get_config_print_mem () = false
 function get_config_print_platform () = false
 
-val EXTS : forall 'n 'm, 'm >= 'n. (implicit('m), bits('n)) -> bits('m)
-val EXTZ : forall 'n 'm, 'm >= 'n. (implicit('m), bits('n)) -> bits('m)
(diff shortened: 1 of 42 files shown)


billmcspadden-riscv (approved):
I did a visual inspection to check that there was a 1-to-1 change (EXTZ to zero_extend and EXTS to sign_extend).   I found no errors.

Made sure there were no other changes unrelated to the changes around EXTZ and EXTS; 
there weren't.

Looks good to me.

martinberger (approved):
It's fine. Let's merge!

billmcspadden-riscv:
At the tech-golden-model meeting on 2023-08-01,  it was agreed to merge this PR.

Issue #280, opened 2023-06-30
aes_shift_rows_fwd and aes_shift_rows_inv do not match crypto spec. 


mlawson-tt:
An issue was discovered a few months ago in the vector crypto spec (https://github.com/riscv/riscv-crypto/issues/268) that caused a fix for `aes_shift_rows_fwd` and  `aes_shift_rows_inv` (https://github.com/riscv/riscv-crypto/commit/a19ae2086efcb87b6988f4510bf19cd642b16740), but the changes have not yet been incorporated here.

This section in `aes_shift_rows_fwd`:
```
  let oc0 : bits(32) = ic0[31..24] @ ic1[23..16] @ ic2[15.. 8] @ ic3[ 7.. 0];
  let oc1 : bits(32) = ic1[31..24] @ ic2[23..16] @ ic3[15.. 8] @ ic0[ 7.. 0];
  let oc2 : bits(32) = ic2[31..24] @ ic3[23..16] @ ic0[15.. 8] @ ic1[ 7.. 0];
  let oc3 : bits(32) = ic3[31..24] @ ic0[23..16] @ ic1[15.. 8] @ ic2[ 7.. 0];
```
should be:
```
  let oc0 : bits(32) = ic3[31..24] @ ic2[23..16] @ ic1[15.. 8] @ ic0[ 7.. 0];
  let oc1 : bits(32) = ic0[31..24] @ ic3[23..16] @ ic2[15.. 8] @ ic1[ 7.. 0];
  let oc2 : bits(32) = ic1[31..24] @ ic0[23..16] @ ic3[15.. 8] @ ic2[ 7.. 0];
  let oc3 : bits(32) = ic2[31..24] @ ic1[23..16] @ ic0[15.. 8] @ ic3[ 7.. 0];
```

and this section in `aes_shift_rows_inv`:
```
  let oc0 : bits(32) = ic0[31..24] @ ic3[23..16] @ ic2[15.. 8] @ ic1[ 7.. 0];
  let oc1 : bits(32) = ic1[31..24] @ ic0[23..16] @ ic3[15.. 8] @ ic2[ 7.. 0];
  let oc2 : bits(32) = ic2[31..24] @ ic1[23..16] @ ic0[15.. 8] @ ic3[ 7.. 0];
  let oc3 : bits(32) = ic3[31..24] @ ic2[23..16] @ ic1[15.. 8] @ ic0[ 7.. 0];
```
should be:
```
[...]

PR #281, opened 2023-06-30
Fix aes_shift_rows_fwd and aes_shift_rows_inv functions from latest vector spec.

mlawson-tt:
This PR updates the `aes_shift_rows_fwd` and `aes_shift_rows_inv` functions to match the versions defined in the latest vector crypto spec after this fix: https://github.com/riscv/riscv-crypto/commit/a19ae2086efcb87b6988f4510bf19cd642b16740.

These functions are not used in any scalar instructions and are not currently called anywhere, so no existing functionality should be affected, but this change is necessary for the upcoming vector crypto implementation and ensures the versions in the scalar spec match the vector crypto spec (as the Sail versions are used directly in the scalar spec).

### DIFF
--- a/model/riscv_types_kext.sail
+++ b/model/riscv_types_kext.sail
@@ -351,10 +351,10 @@ function aes_shift_rows_fwd(x) = {
   let ic2 : bits(32) = aes_get_column(x, 2);
   let ic1 : bits(32) = aes_get_column(x, 1);
   let ic0 : bits(32) = aes_get_column(x, 0);
-  let oc0 : bits(32) = ic0[31..24] @ ic1[23..16] @ ic2[15.. 8] @ ic3[ 7.. 0];
-  let oc1 : bits(32) = ic1[31..24] @ ic2[23..16] @ ic3[15.. 8] @ ic0[ 7.. 0];
-  let oc2 : bits(32) = ic2[31..24] @ ic3[23..16] @ ic0[15.. 8] @ ic1[ 7.. 0];
-  let oc3 : bits(32) = ic3[31..24] @ ic0[23..16] @ ic1[15.. 8] @ ic2[ 7.. 0];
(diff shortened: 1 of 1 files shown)


mlawson-tt:
Fixes #280.

jrtc27 (approved):
Ideally we would never have added them in the first place if they're unused (did an old version of the scalar spec want them?..), but since they're here and will be used by vector crypto we should fix them and leave them here for it to use once added rather than introduce the churn of removing them and re-adding them with vector crypto. So these changes make sense to land from my POV, and carry zero risk given they're in dead code (and when vector crypto comes into the tree it'll be responsible for testing these as part of its instruction implementations).

ptomsich:
@charmitro Patch #234 uses these functions, but (from what I understand) Zvkned works correctly with the current implementation.  Could you review and report back?

charmitro (commented):
If `vector-dev` branch isn't to be merged soon with `master`. Would it be wise for this to get cherry-picked or similar to `vector-dev` as it seems that it's only going to be needed for my #234 ?

mlawson-tt:
@charmitro I was actually building your branch to test against anyway, so I can just merge this into your branch if it makes more sense.  I just targeted master because these functions are showing up in the scalar crypto spec now, and that spec includes this file directly, so merging into master would let the scalar spec pick up the change so that the vector spec and scalar spec don't disagree on these functions (even though these functions aren't used in any scalar instructions).  Seems like it'd be nice for these functions to not show up in the scalar spec at all since they're not used, but I imagine it may be a bigger pain to try remove them than to just keep including this file directly.

Timmmm:
Looks like this hasn't been merged yet, and I agree with @jrtc27 we probably shouldn't have these functions if they aren't used. However since they are there we may as well fix them. I'll merge this in a few days if nobody objects.

PR #285, opened 2023-07-15
Add Zacas (atomic compare and swap) extension

ved-rivos:
Add the Zacas extension that introduces the following instructions:

- amocas.w for RV32 and RV64
- amocas.d for RV32 and RV64
- amocas.q for RV64 only

Tested with these tests: https://github.com/riscv-non-isa/riscv-arch-test/tree/dev/riscv-test-suite/rv64i_m/Zacas/src

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -118,6 +118,10 @@ function clause hartSupports(Ext_Zaamo) = config extensions.Zaamo.supported
 enum clause extension = Ext_Zabha
 mapping clause extensionName = Ext_Zabha <-> "zabha"
 function clause hartSupports(Ext_Zabha) = config extensions.Zabha.supported
+// Atomic Compare-and-Swap (CAS) Instructions
+enum clause extension = Ext_Zacas
+mapping clause extensionName = Ext_Zacas <-> "zacas"
+function clause hartSupports(Ext_Zacas) = config extensions.Zacas.supported
 // Load-Reserved/Store-Conditional Instructions
 enum clause extension = Ext_Zalrsc
 mapping clause extensionName = Ext_Zalrsc <-> "zalrsc"
@@ -403,6 +407,7 @@ let extensions_ordered_for_isa_string = [
   // Za
   Ext_Zaamo,
   Ext_Zabha,
+  Ext_Zacas,
   Ext_Zalrsc,
   Ext_Zawrs,
 
--- a/model/riscv_insts_zaamo.sail
+++ b/model/riscv_insts_zaamo.sail
@@ -99,6 +99,9 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
                   // All operations except CAS where loaded != X(rd) write their value back.
                   if op == AMOCAS & loaded != (if width <= xlen_bytes then trunc(X(rd)) else trunc(X_pair(rd))) then {
                     // Compared, but didn't swap.
+                    if width <= xlen_bytes
+                    then X(rd) = sign_extend(loaded)
+                    else X_pair(rd) = sign_extend(loaded);
                     RETIRE_SUCCESS
                   } else {
                     match mem_write_value(addr, width, sign_extend(result), aq & rl, rl, true) {
--- a/model/riscv_insts_zaamo.sail
+++ b/model/riscv_insts_zaamo.sail
@@ -74,7 +74,7 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
           // AMOCAS.D for RV32 and AMOCAS.Q for RV64 operate on a pair of
           // registers. When the first register of a register pair is
           // x0, then both halves of the pair read/write as zero.
-          let rs2_val = X_pair(rs2, width)[width * 8 - 1 .. 0];
+          let rs2_val : bits('width * 8) = if width <= xlen_bytes then trunc(X(rs2)) else trunc(X_pair(rs2));
           match mem_write_ea(addr, width, aq & rl, rl, true) {
             Err(e) => Memory_Exception(vaddr, e),
             Ok(_) => {
@@ -97,13 +97,15 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
                     };
 
                   // All operations except CAS where loaded != X(rd) write their value back.
-                  if op == AMOCAS & loaded != X_pair(rd, width)[width * 8 - 1 .. 0] then {
+                  if op == AMOCAS & loaded != (if width <= xlen_bytes then trunc(X(rd)) else trunc(X_pair(rd))) then {
                     // Compared, but didn't swap.
                     RETIRE_SUCCESS
                   } else {
                     match mem_write_value(addr, width, sign_extend(result), aq & rl, rl, true) {
                       Ok(true)  => {
-                        X_pair(rd, width) = sign_extend(loaded);
+                        if width <= xlen_bytes
+                        then X(rd) = sign_extend(loaded)
+                        else X_pair(rd) = sign_extend(loaded);
                         RETIRE_SUCCESS
                       },
                       Ok(false) => { internal_error(__FILE__, __LINE__, "AMO got false from mem_write_value") },
--- a/model/riscv_insts_zaamo.sail
+++ b/model/riscv_insts_zaamo.sail
@@ -131,7 +131,8 @@ mapping amo_mnemonic : amoop <-> string = {
   AMOMIN  <-> "amomin",
   AMOMAX  <-> "amomax",
   AMOMINU <-> "amominu",
-  AMOMAXU <-> "amomaxu"
+  AMOMAXU <-> "amomaxu",
+  AMOCAS  <-> "amocas",
 }
 
 mapping maybe_aqrl : (bool, bool) <-> string = {
--- a/model/riscv_insts_zaamo.sail
+++ b/model/riscv_insts_zaamo.sail
@@ -52,29 +52,6 @@ mapping clause encdec = AMO(op, aq, rl, rs2, rs1, size, rd)
   <-> encdec_amoop(op) @ bool_bits(aq) @ bool_bits(rl) @ encdec_reg(rs2) @ encdec_reg(rs1) @ size_enc_wide(size) @ encdec_reg(rd) @ 0b0101111
   when amo_encoding_valid(size, op, rs2, rd)
 
-// AMOCAS.D for RV32 and AMOCAS.Q for RV64 operate on a pair of
-// registers. When the first register of a source register pair is
-// x0, then both halves of the pair read as zero.
-
-val rX_amo_bits : forall 'w, 'w in {1, 2, 4, 8, 16} & 'w <= xlen_bytes * 2 . (regidx, int('w)) -> bits('w * 8)
-function rX_amo_bits(i, width_bytes) =
-  if i != zreg then {
-    if width_bytes <= xlen_bytes then X(i)[width_bytes * 8 - 1 .. 0] else X(i + 1) @ X(i)
-  } else zeros()
-
-val wX_amo_bits : forall 'w, 'w in {1, 2, 4, 8, 16} & 'w <= xlen_bytes * 2 . (regidx, int('w), bits('w * 8)) -> unit
-function wX_amo_bits(i, width_bytes, data) =
-  if i != zreg then {
-    if width_bytes <= xlen_bytes then X(i) = sign_extend(data) else {
-      X(i) = data[xlen - 1 .. 0];
-      X(i + 1) = data[xlen * 2 - 1 .. xlen];
-    }
-  }
-
-// X_amo allows accessing parts of registers (writes are sign-extended),
-// or register pairs for AMOCAS.
-overload X_amo = {rX_amo_bits, wX_amo_bits}
-
 /* NOTE: Currently, we only EA if address translation is successful.
    This may need revisiting. */
 function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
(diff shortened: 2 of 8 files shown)


jrtc27:
> Pull requests should be a single set of related changes with a clean commit history free from merge commits.

-- https://github.com/riscv/sail-riscv/blob/master/CONTRIBUTING.md

model/riscv_insts_aext.sail
```diff
@@ -265,89 +265,189 @@ mapping encdec_amoop : amoop <-> bits(5) = {
...
+mapping clause encdec = AMO(op, aq, rl, rs2, rs1, size, rd) if amo_encoding_valid(size, op, rs2, rd)
+  <-> encdec_amoop(op) @ bool_bits(aq) @ bool_bits(rl) @ rs2 @ rs1 @ amo_size_bits(size) @ rd @ 0b0101111 if amo_encoding_valid(size, op, rs2, rd)
 
 /* NOTE: Currently, we only EA if address translation is successful.
    This may need revisiting. */
 function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
   if haveAtomics() then {
     /* Get the address, X(rs1) (no offset).
      * Some extensions perform additional checks on address validity.
+     * For obtaining virtual address, the width is not needed.
```
jrtc27, line 301: No, this breaks things like CHERI that hook ext_data_get_addr. The width *must* be accurate for the access that's about to be performed.

jrtc27 (changes requested):

Timmmm:
I might rebase & update this if that's ok...

ved-rivos:
Thanks @Timmmm - I did the rebase and keep it fresh a few times over last 2 years but it kept going stale. Thanks for helping rebase it again !

Timmmm:
I rebased it and reworked it a fair bit (the code has changed a lot!)

I haven't modelled this bit:

> The memory operation performed by an AMOCAS.W/D/Q, when not successful, has acquire semantics if aq bit is 1 **but does not have release semantics, regardless of rl**.

I don't think we can actually model that with the current code, because we tell the memory operations the release semantics before they've told us whether they succeed or not.

I think the correct way to do this would be to have some kind of `release_if_successful` memory ordering, but we don't have that yet. I think we can add it after #792.

Also I haven't tested this version yet.

ved-rivos:
> Also I haven't tested this version yet.

A set of tests are available [here](https://github.com/riscv-non-isa/riscv-arch-test/tree/dev/riscv-test-suite/rv64i_m/Zacas/src).

model/riscv_insts_zaamo.sail
```diff
@@ -34,33 +49,57 @@ mapping encdec_amoop : amoop <-> bits(5) = {
...
+  AMOCAS  <-> 0b00101,
 }
 
 mapping clause encdec = AMO(op, aq, rl, rs2, rs1, size, rd)
-  <-> encdec_amoop(op) @ bool_bits(aq) @ bool_bits(rl) @ encdec_reg(rs2) @ encdec_reg(rs1) @ 0b0 @ size_enc(size) @ encdec_reg(rd) @ 0b0101111
-  when currentlyEnabled(Ext_Zaamo) & amo_width_valid(size)
+  <-> encdec_amoop(op) @ bool_bits(aq) @ bool_bits(rl) @ encdec_reg(rs2) @ encdec_reg(rs1) @ amo_size_enc(size) @ encdec_reg(rd) @ 0b0101111
+  when amo_encoding_valid(size, op, rs2, rd)
+
+// AMOCAS.W for RV32 and AMOCAS.Q for RV64 operate on a pair of
```
arichardson:
```suggestion
// AMOCAS.D for RV32 and AMOCAS.Q for RV64 operate on a pair of
```
I assume this should be a `d`?

model/riscv_types.sail
```diff
@@ -330,6 +331,32 @@ mapping size_bytes : word_width <-> {1, 2, 4, 8} = {
   DOUBLE <-> 8,
 }
 
+enum amo_word_width = {AMO_BYTE, AMO_HALF, AMO_WORD, AMO_DOUBLE, AMO_QUAD}
```
jordancarlin: Adding `quad` to `word_width` isn’t exclusive to AMOs. I wonder if the approach in #445 would make more sense here.

Timmmm: Hmm yeah I wasn't very happy about this either. I think you're right - on balance that approach is better.

The thing I don't like though is that now you can have `QUAD` in instructions that can't ever actually use `QUAD`, `size_enc(QUAD)` isn't detected at compile time, etc. Maybe it's fine?

The other option is to use integers - Sail has much better support for integer types than enums so we could do e.g.

```
type word_width = {1, 2, 4, 8}
type word_width_wide= {1, 2, 4, 8, 16}

mapping size_enc : word_width <-> bits(2) = {
  1 <-> 0b00,
  2 <-> 0b01,
  4 <-> 0b10,
  8 <-> 0b11,
}

mapping size_enc_wide : word_width_wide <-> bits(3) = {
  1 <-> 0b000,
  2 <-> 0b001,
  4 <-> 0b010,
  8 <-> 0b011,
  16 <-> 0b100,
}
```

This would get rid of the need for a separate `size_bytes` mapping.

Thinking about it even more it would probably make even *more* sense to just store the exponent.

```
type width_exp = range(0, 3)
type width_wide_exp = range(0, 4)
```

I think that would provide the best typing and most elegance, though it might be slightly less intuitive?

model/riscv_insts_zaamo.sail
Timmmm: Good spot, fixed!

model/riscv_types.sail
Timmmm: Ok I made #1113 to do the first version (not exponents), and then have rebased this PR on top of it. I think it's a lot nicer.

model/riscv_extensions.sail
```diff
@@ -118,6 +118,10 @@ function clause hartSupports(Ext_Zaamo) = config extensions.Zaamo.supported
 enum clause extension = Ext_Zabha
 mapping clause extensionName = Ext_Zabha <-> "zabha"
 function clause hartSupports(Ext_Zabha) = config extensions.Zabha.supported
+// Atomic Compare-and-Swap (CAS) Instructions
+enum clause extension = Ext_Zacas
+mapping clause extensionName = Ext_Zacas <-> "zacas"
+function clause hartSupports(Ext_Zacas) = config extensions.Zacas.supported
```
jordancarlin, line 124: Zacas needs to be added to the list later in this file for the device tree.

model/riscv_insts_zaamo.sail
```diff
@@ -9,21 +9,31 @@
 function clause currentlyEnabled(Ext_A) = hartSupports(Ext_A) & misa[A] == 0b1 // TODO: Move currentlyEnabled(Ext_A) to a more central file
 function clause currentlyEnabled(Ext_Zaamo) = hartSupports(Ext_Zaamo) | currentlyEnabled(Ext_A)
 function clause currentlyEnabled(Ext_Zabha) = hartSupports(Ext_Zabha) & currentlyEnabled(Ext_Zaamo)
+function clause currentlyEnabled(Ext_Zacas) = hartSupports(Ext_Zacas) & currentlyEnabled(Ext_Zaamo)
```
jordancarlin, line 12: Zacas needs to be given a value above Zaamo in the termination file for Lean.

model/riscv_insts_zaamo.sail
```diff
@@ -34,33 +44,57 @@ mapping encdec_amoop : amoop <-> bits(5) = {
...
 }
 
 mapping clause encdec = AMO(op, aq, rl, rs2, rs1, size, rd)
-  <-> encdec_amoop(op) @ bool_bits(aq) @ bool_bits(rl) @ encdec_reg(rs2) @ encdec_reg(rs1) @ 0b0 @ size_enc(size) @ encdec_reg(rd) @ 0b0101111
-  when currentlyEnabled(Ext_Zaamo) & amo_width_valid(size)
+  <-> encdec_amoop(op) @ bool_bits(aq) @ bool_bits(rl) @ encdec_reg(rs2) @ encdec_reg(rs1) @ size_enc_wide(size) @ encdec_reg(rd) @ 0b0101111
+  when amo_encoding_valid(size, op, rs2, rd)
+
+// AMOCAS.D for RV32 and AMOCAS.Q for RV64 operate on a pair of
+// register. When the first register of a source register pair is
```
arichardson:
```suggestion
// registers. When the first register of a source register pair is
```

model/riscv_insts_zaamo.sail
```diff
@@ -34,33 +44,57 @@ mapping encdec_amoop : amoop <-> bits(5) = {
...
+  if i != zreg then {
+    if width_bytes <= xlen_bytes then X(i) = sign_extend(data) else {
+      X(i) = data[xlen - 1 .. 0];
+      X(i + 1) = data[xlen * 2 - 1 .. xlen];
+    }
+  }
+
+// X_amo allows accessing parts of registers (writes are sign-extended),
+// or register pairs for AMOCAS.
+overload X_amo = {rX_amo_bits, wX_amo_bits}
```
arichardson: Is this the same behaviour as for Zilsd? In that case I'd suggest using something like X_pair() and moving it into a more central file.

Timmmm: Yeah it is. Good idea.

model/riscv_extensions.sail
Timmmm: Done

model/riscv_insts_zaamo.sail
Timmmm: Done

model/riscv_insts_zaamo.sail
Timmmm: Ok I moved it to `riscv_regs.sail`. I also changed it from accepting/returning `width_bytes` and sign-extending to accepting/returning either XLEN or XLEN*2. This means the sign extension is now in the Zacas code, which I think is nicer (also because Zilsd doesn't care about sign extension).

The downside is that the AMO code now has to slice the values it reads... but maybe that's clearer too.

model/riscv_regs.sail
```diff
@@ -223,6 +223,30 @@ function wX_bits(Regidx(i) : regidx, data : xlenbits) -> unit = {
 
 overload X = {rX_bits, wX_bits, rX, wX}
 
+// X_pair allows accessing a single register or a pair of registers for
+// XLEN*2 accesses. When accessing (x0, x1) then the entire access
+// is read-only zero. This is used for Zacas and Zilsd.
+
+// `width_bytes` is the access width. If it is greater than XLEN then
+// a 2*XLEN bit vector will be returned.
+val rX_pair_bits : forall 'w, 'w <= xlen_bytes * 2 . (regidx, int('w)) -> bits((if 'w <= xlen_bytes then xlen_bytes else xlen_bytes * 2) * 8)
```
jordancarlin: While this might simplify code elsewhere, I'm worried that it will be confusing for a function called `X_pair` to sometimes grab a single register and sometimes grab a pair of registers. I'd be in favor of `X_pair` always returning a pair and pulling the check for whether one or two registers is needed into the calling function so it can decide whether to call `X` or `X_pair`.

Timmmm: Maybe a different name would help... I do feel like it makes the code quite a bit simpler. `X_wide`?

jordancarlin: Not sure. Also, part of the reason to move this into a more generic file was for Zilsd down the line. I think Zilsd will always access a register pair. At the risk of adding more indirection, what about an `X_wide` function that always accesses a register pair and then an `X_amo` function that has the extra bits needed to conditionally access parts of a register or to call `X_pair`?

Timmmm: Yeah but actually `zilsd` `ld` is exactly the same encoding as RV64 `ld`, so with the current definition of `X_pair` they could share exactly the same code. Maybe that is too magical though... It's definitely how I would write it if Sail wasn't meant to be a sort of instructional pseudocode.

arichardson: Passing in the width seems a bit odd, it would make more sense to me to do the "do I need a pair of registers" check inside the amocas function and have the `X_pair` unconditionally read/write the pair of registers.

model/riscv_insts_zaamo.sail
```diff
@@ -78,11 +92,23 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
...
-                    Err(e)    => Memory_Exception(vaddr, e),
+
+                  // All operations except CAS where loaded != X(rd) write their value back.
+                  if op == AMOCAS & loaded != X_pair(rd, width)[width * 8 - 1 .. 0] then {
+                    // Compared, but didn't swap.
+                    RETIRE_SUCCESS
+                  } else {
+                    match mem_write_value(addr, width, sign_extend(result), aq & rl, rl, true) {
+                      Ok(true)  => {
+                        X_pair(rd, width) = sign_extend(loaded);
```
arichardson:
```suggestion
                        if width > xlen_bytes then
                          X_pair(rd) = sign_extend(loaded);
                        else
                          X(rd) = sign_extend(loaded);
```

Something like this?

model/riscv_insts_zaamo.sail
```diff
@@ -34,33 +44,37 @@ mapping encdec_amoop : amoop <-> bits(5) = {
...
+      if not(is_aligned_addr(vaddr, width))
       then Memory_Exception(vaddr, E_SAMO_Addr_Align())
       else match translateAddr(vaddr, ReadWrite(Data, Data)) {
         Err(e, _) => Memory_Exception(vaddr, e),
         Ok(addr, _) => {
-          let rs2_val = X(rs2)[width * 8 - 1 .. 0];
+          // AMOCAS.D for RV32 and AMOCAS.Q for RV64 operate on a pair of
+          // registers. When the first register of a register pair is
+          // x0, then both halves of the pair read/write as zero.
+          let rs2_val = X_pair(rs2, width)[width * 8 - 1 .. 0];
```
arichardson: Wouldn't this be readable as a call to truncate?
```suggestion
          let rs2_val = truncate(if width > xlen_bytes then X_pair(rs2) else X(rs2), width * 8);
```

Timmmm: Ah yeah these suggestions seem reasonable to me. I'll update the code.

model/riscv_regs.sail
Timmmm: Ok I updated the code with @arichardson's suggestions. I think it is quite a lot nicer!

jordancarlin:
Tested this against the Zacas tests in `riscv-arch-test` per https://github.com/riscv/sail-riscv/issues/1133#issuecomment-3058811301 and found a few issues:
- Sail throws a pattern match failure when executing an `amocas` instruction because the `amo_mnemonic` was missing a clause for `AMOCAS`. Seems like the Sail compiler should've caught this one.
- `amocas` instructions are supposed to load the value from memory into rd irrespective of whether or not the swap itself happens. The existing implementation was missing this.

I pushed fixes for both and the `riscv-arch-test` tests pass now.

Timmmm (approved):
Ah thank you for that!

Timmmm:
Happy almost 2nd birthday this PR :cake:

Issue #112, opened 2021-10-04
Bad disassembly for RVC HINTs


jrtc27:
RVC HINTs are disassembled with names like `c.nop.hint.<imm>` rather than the mnemonics used by actual (dis)assemblers (which are generally just whatever the non-HINT version would be in the case of using a zero register, but some exceptions for things like `c.nop`, which ends up taking an additional immediate argument when non-zero, and `c.s{ll,rl,ra}i` with a zero immediate, which end up being the strange `c.s{ll,rl,ra}i64` with no immediate).

See discussion in #31.

ThinkOpenly:
Moving the discussion from #324 here... (I may be showing my ignorance, so bear with me.)

Taking one example, `c.li`...  The definition of the instruction (assembly) is:
```
mapping clause assembly = C_LI(imm, rd)
      if rd != zreg
  <-> "c.li" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_6(imm)
      if rd != zreg
```

The hint equivalent (and I agree it is strange):
```
mapping clause assembly = C_LI_HINT(imm)
  <-> "c.li.hint." ^ hex_bits_6(imm)                 
```

Does it make sense for the latter to be something more like:
```
mapping clause assembly = C_LI(imm, rd)
      if rd == zreg
  <-> "c.li" ^ spc() ^ hex_bits_6(imm)                 
      if rd == zreg
```
?

It uses the correct mnemonic and it's mutually exclusive with the actual instruction due to the condition. Does the presence of `rd` break something?

jrtc27:
> Does it make sense for the latter to be something more like:
> 
> mapping clause assembly = C_LI(imm, rd)
>       if rd == zreg
>   <-> "c.li" ^ spc() ^ hex_bits_6(imm)                 
>       if rd == zreg
> ?

Well the assembly needs to actually include rd...

jrtc27:
Note that LLVM will treat it as `c.li zero, imm`, i.e. the "obvious" mapping. I assume GNU as is similar.

Alasdair:
Is:
```
mapping clause assembly = C_LI_HINT(imm)
  <-> "c.li" ^ spc() ^ "zero" ^ sep() ^ hex_bits_6(imm)
```
what you want?

jrtc27:
Better use `reg_name(zreg)` rather than hard-coding register names everywhere

Timmmm:
Should be fixed in #1062 - please reopen if there are any more issues!

PR #287, opened 2023-07-19
Allow compressed hints to be expanded

PeterRugg:
This leads to different values in RVFI. The RISC-V spec allows this behaviour:
"simple implementations can ... execute a HINT as a regular computational instruction".

Without this feature, such implementations cannot be verified using RVFI.

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -120,6 +120,7 @@ static struct option options[] = {
     {"disable-writable-misa",       no_argument,       0, 'I'},
     {"disable-fdext",               no_argument,       0, 'F'},
     {"mtval-has-illegal-inst-bits", no_argument,       0, 'i'},
+    {"c-hints-expand",              no_argument,       0, 'H'},
     {"device-tree-blob",            required_argument, 0, 'b'},
     {"terminal-log",                required_argument, 0, 't'},
     {"show-times",                  required_argument, 0, 'p'},
@@ -232,6 +233,7 @@ char *process_args(int argc, char **argv)
                     "I"
                     "F"
                     "i"
+                    "H"
                     "s"
                     "p"
                     "z:"
@@ -288,6 +290,10 @@ char *process_args(int argc, char **argv)
     case 'i':
       fprintf(stderr, "enabling storing illegal instruction bits in mtval.\n");
       rv_mtval_has_illegal_inst_bits = true;
+      break;
+    case 'H':
+      fprintf(stderr, "enabling expanding of compressed hints.\n");
+      rv_c_hints_expand = true;
       break;
     case 's':
       do_dump_dts = true;
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -57,6 +57,11 @@ bool plat_mtval_has_illegal_inst_bits(unit u)
   return rv_mtval_has_illegal_inst_bits;
 }
 
+bool plat_c_hints_expand(unit u)
+{
+  return rv_c_hints_expand;
+}
+
 bool plat_enable_pmp(unit u)
 {
   return rv_enable_pmp;
--- a/c_emulator/riscv_platform.h
+++ b/c_emulator/riscv_platform.h
@@ -10,6 +10,7 @@ bool sys_enable_writable_misa(unit);
 bool plat_enable_dirty_update(unit);
 bool plat_enable_misaligned_access(unit);
 bool plat_mtval_has_illegal_inst_bits(unit);
+bool plat_c_hints_expand(unit);
 bool plat_enable_pmp(unit);
 
 mach_bits plat_ram_base(unit);
--- a/c_emulator/riscv_platform_impl.c
+++ b/c_emulator/riscv_platform_impl.c
@@ -13,6 +13,7 @@ bool rv_enable_fdext = true;
 bool rv_enable_dirty_update = false;
 bool rv_enable_misaligned = false;
 bool rv_mtval_has_illegal_inst_bits = false;
+bool rv_c_hints_expand = false;
 
 uint64_t rv_ram_base = UINT64_C(0x80000000);
 uint64_t rv_ram_size = UINT64_C(0x4000000);
--- a/c_emulator/riscv_platform_impl.h
+++ b/c_emulator/riscv_platform_impl.h
@@ -16,6 +16,7 @@ extern bool rv_enable_writable_misa;
 extern bool rv_enable_dirty_update;
 extern bool rv_enable_misaligned;
 extern bool rv_mtval_has_illegal_inst_bits;
+extern bool rv_c_hints_expand;
 
 extern uint64_t rv_ram_base;
 extern uint64_t rv_ram_size;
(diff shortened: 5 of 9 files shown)


c_emulator/riscv_sim.c
```diff
@@ -120,6 +120,7 @@ static struct option options[] = {
     {"disable-writable-misa",       no_argument,       0, 'I'},
     {"disable-fdext",               no_argument,       0, 'F'},
     {"mtval-has-illegal-inst-bits", no_argument,       0, 'i'},
+    {"c_hints_expand",              no_argument,       0, 'H'},
```
charmitro, line 123: Just a thought.

I'm wondering if this needs to be `"c-hints-expand"`. That way, it uses the same naming convention as the other flags.

PeterRugg: Oops, sorry, good catch!

c_emulator/riscv_sim.c
```diff
@@ -120,6 +120,7 @@ static struct option options[] = {
     {"disable-writable-misa",       no_argument,       0, 'I'},
     {"disable-fdext",               no_argument,       0, 'F'},
     {"mtval-has-illegal-inst-bits", no_argument,       0, 'i'},
+    {"c-hints-expand",              no_argument,       0, 'H'},
```
billmcspadden-riscv, line 123: Would rather not have 'H' used as the single character switch,  as it probably should be reserved for hypervisor. I would suggest to only have a long option.

billmcspadden-riscv (commented):
1. Shouldn't the compressed hints always be expanded?   In other words,  why do we need rv_c_hints_expand?
2. The expansion seems correct to me.
3. Testing:  what tests exist?  Shouldn't we have some tests for checking this functionality as part of the PR?

PeterRugg:
@billmcspadden-riscv
1. I didn't think I should disrupt the sail model that much: presumably somebody did this intentionally because they had to jump through hoops to do it. I think it's useful that they get labeled as "HINT"s at the very least, so it's clear from the trace if you're running code that's expecting some extension that uses them. Presumably when HINT encodings actually get used, they'll be defined in this file. That said, it suits my use-case for them to be just decoded as normal without special casing, so I wouldn't mind that at all.
2. Thank you for checking.
3. You're quite right: I can look into adding tests if there's agreement that this PR might get merged.

c_emulator/riscv_sim.c
PeterRugg: Happy to, although if I understand correctly, it will mean some refactoring of this file, as currently all the options have short forms and that's what is selected for in the big switch statement.

PeterRugg:
A few thoughts on testing, as I'm not super clear on what tests would look like:
1. It seems the HINT cases should be included in the RVC tests regardless of whether this PR goes ahead, as I've heard two implementations now (including early sail?) misread the spec and made them illegal. I guess a test case that runs the HINT and checks that there's no trap would be a start. Of course that would mean that a future extension processor that actually used the HINTs for something would fail the tests, which is presumably why they aren't tested currently?
2. Asserting "this instruction didn't do anything" seems tricky, and most instruction tests presumably don't bother checking that nothing changed that wasn't supposed to.
3. As far as I can tell, the tests are all written in terms of architecturally observable behaviour, so this option wouldn't change anything?
4. Surely we don't want to run all the tests again for both values of the "c-expand-hints" option: is there a way of selecting a subset of tests that run for given option combinations?

c_emulator/riscv_sim.c
charmitro: I think you can use a any number beyond the ascii-table.  Just remember to remove the short form of this option from the `print_usage()` function.

For example,
```c
#DEFINE C_HINTS_EXPAND_OPT 1000

{"c-hints-expand",              no_argument,       0, C_HINTS_EXPAND_OPT},

case C_HINTS_EXPAND_OPT:
     ...
```

jrtc27:
The current Sail model hints implementation is a continued mess of people overcomplicating things and doing the wrong thing. See https://github.com/riscv/sail-riscv/issues/112 and follow the chain back to earlier discussions. I fully support completely axing the distinction between hints and non-hints at decode time; it’s unnecessary complexity, confusing for people reading the model and gets the disassembly wrong (using some self-invented nonsense mnemonics rather than what both GNU and LLVM toolchains use). The only special casing should be for when extensions are enabled that give the hints additional meaning.

PeterRugg:
@jrtc27 Fine by me. I'm happy to do a PR to simplify things back again.

jrtc27:
I honestly don’t know why https://github.com/riscv/sail-riscv/pull/31 wasn’t merged and instead the current approach was written and merged. It should just be as I originally proposed.

PeterRugg:
I guess the prose spec really doesn't allow that interpretation: presumably they want to make absolutely sure that non-extended code can't use the HINTs and expect them to be nops.

jrtc27:
That’s how toolchains implement them and cores that don’t give the hints special meaning implement them. That’s how the spec intends them to be implemented without further extensions:

image

I think that’s pretty clear that they really are those instructions without additional hint meaning assigning extensions.

allenjbaum:
Also note that there are some (new Zce) compressed ops that don't expand to
full width ops.

jrtc27:
> Also note that there are some (new Zce) compressed ops that don't expand to full width ops.
> […](#)

That doesn’t change anything. Those need implementing directly, but these don’t.

leesum1:
In the cadd test within riscv-arch-tests, the instruction in inst_5
```asm
inst_5:
// rs1==x0, rs2==x2, rs1_val == (2**(xlen-1)-1), rs1_val == 2147483647
// opcode: c.add; op1:x0; op2:x2; op1val:0x0; op2val:-0xb503
TEST_CR_OP( c.add, x0, x2, 0, 0x0, -0xb503, x8, 20, x4) 
```
is considered an invalid instruction. Has there been any progress regarding hint instructions?

PeterRugg:
Hmm, interesting. IIRC, this change should only affect tracing rather than whether instructions are legal or not.

arichardson:
> In the cadd test within riscv-arch-tests, the instruction in inst_5
> 
> ```assembly
> inst_5:
> // rs1==x0, rs2==x2, rs1_val == (2**(xlen-1)-1), rs1_val == 2147483647
> // opcode: c.add; op1:x0; op2:x2; op1val:0x0; op2val:-0xb503
> TEST_CR_OP( c.add, x0, x2, 0, 0x0, -0xb503, x8, 20, x4) 
> ```
> 
> is considered an invalid instruction. Has there been any progress regarding hint instructions?

This is already a HINT? 
```
mapping clause encdec_compressed = C_ADD_HINT(rs2)
      if rs2 != zreg
  <-> 0b100 @ 0b1 @ 0b00000 @ rs2 : regidx @ 0b10
      if rs2 != zreg

function clause execute (C_ADD_HINT(rs2)) = RETIRE_SUCCESS
```
The PR here just expands it to actually execute the ADD with zero as the destination.

Timmmm:
I would support closing this PR and just deleting the hints. The only reason I can think to keep them is for nicer disassembly, but that seems marginal and if the disassembly isn't correct at the moment anyway then I don't think it makes sense to keep them.

Any objections?

PeterRugg:
Would work fine for me!

jordancarlin:
Closing now that #1062 deleted the hints.

PR #298, opened 2023-08-30
Remove effects

Timmmm:
Since Sail 0.15 (released Nov 2022), effects have had no effect. They now generate a deprecation warning. This commit removes all the effect annotations from the model, thus fixing the compiler warnings.

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -87,7 +87,7 @@ val eq_anything = {ocaml: "(fun (x, y) -> x = y)", interpreter: "eq_anything", l
 
 overload operator == = {eq_string, eq_anything}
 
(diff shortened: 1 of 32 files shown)


arichardson:
Thanks! The amount of warnings while building the model is quite annoying, so I'm very happy to see this fixed.

Timmmm:
> Thanks! The amount of warnings while building the model is quite annoying, so I'm very happy to see this fixed.

Happy enough to review it? :-D

arichardson (approved):
Don't have the right permissions but LGTM

Timmmm:
Ah fair enough. Assumed you would since you're contributor number 7 according to GitHub!

billmcspadden-riscv:
Approved by tech-golden-model on 2023-09-12

Issue #300, opened 2023-08-30
Setting F=0 and D=1 should result in both F and D being cleared


ahadali-10x:
The privileged spec says:
>If an ISA feature x depends on an ISA feature y, then attempting to enable feature x but disable
feature y results in both features being disabled. For example, setting “F”=0 and “D”=1 results
in both “F” and “D” being cleared.

[This code](https://github.com/riscv/sail-riscv/blob/861bd6f65593013381c905d7a54e538e5382bf31/model/riscv_sys_regs.sail#L167-L172) does not behave this way: It will leave F & D alone instead of clearing both of them. So it could end up with e.g. F=1 D=0.

billmcspadden-riscv:
@ahadali-10x ,  the PR has been merged.   Would you please close this issue?

PR #301, opened 2023-08-30
Setting F=0 and D=1 should result in both F and D being cleared

ahadali-10x:
### Changed
- Updated `legalize_misa` function in the `riscv_sys_regs.sail` in the `model/` directory to add condition by setting F=0 and D=1 should clear both F and D.
### Issue Fixed
- Fix for issue #300

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -164,10 +164,10 @@ function legalize_misa(m : Misa, v : xlenbits) -> Misa = {
   else {
     /* Suppress enabling C if C was disabled at boot (i.e. not supported) */
     let m = if not(sys_enable_rvc()) then m else update_C(m, v.C());
-    /* Handle updates for F/D. */
-    if   not(sys_enable_fdext()) | (v.D() == 0b1 & v.F() == 0b0)
+    /* Suppress updates to misa.{f,d} if disabled at boot */
+    if   not(sys_enable_fdext())
     then m
-    else update_D(update_F(m, v.F()), v.D())
+    else update_D(update_F(m, v.F()), v.D() & v.F())
   }
 }
 


Timmmm:
This looks correct to me.

ahadali-10x:
@billmcspadden-riscv @scottj97  Kindly review the changes.

arichardson (approved):

billmcspadden-riscv:
@ahadali-10x : please add the following reference to your commit message:

"Per section 3.1.1 of the Privileged Spec ("Machine ISA Register misa"),  ...."

billmcspadden-riscv (approved):
This looks correct to me.

ahadali-10x:
@billmcspadden-riscv This PR can be merged now.

jrtc27:
Commit message was poorly formatted, subject was too long and contained all the detail, rather than a short subject with a detailed body.

Alasdair:
Commit message looks fine to me. Says precisely what was fixed. Do we have precise rules on what a commit message should be in the style guide?

jrtc27:
No, because it’s standard practice for any well-managed open source project. You can tell it’s not a good commit message because GitHub cuts the subject off; were it correctly written the subject would be short enough to not be cut off. Git’s own manpage for the commit command documents this best practice (though 50 characters is a bit aggressive, IIRC GitHub caps at 80) https://manpages.debian.org/bookworm/git-man/git-commit.1.en.html#DISCUSSION

Timmmm:
Ah yeah I think Jessica is right - we might be looking at different things. The PR title is fine:

> Setting F=0 and D=1 should result in both F and D being cleared

(except that it's a description of the bug, not the fix)

The commit message is very long and not in the normal commit message style:

> Per section 3.1.1 of the Privileged Spec (Machine ISA Register misa): F/D both should be disabled if F=0

Probably should have been something like

> Clear miss F and D bits when F=0 and D=1

With the spec reference in the body.

Relatively minor issue IMO though.

PR #303, opened 2023-09-06
Implement menvcfg

Timmmm:
This implements the m/senvcfg(h) CSRs. This CSR is used to enable/disable extensions and behaviours for lower privilege modes. Currently the only implemented bit is FIOM which affects how fences work.

It also affects how atomic memory accesses work in non-cacheable regions, but the model does not currently support PMAs so that can't easily be implemented.

Note there is a partial implementation of `menvcfg` in #137, so this should simplify that PR a little.


handwritten_support/riscv_extras_sequential.lem
```diff
@@ -153,6 +153,10 @@ val sys_enable_next : unit -> bool
 let sys_enable_next () = true
 declare ocaml target_rep function sys_enable_next = `Platform.enable_next`
 
+val sys_enable_fiom : unit -> bool
```
arichardson, line 156: This is hardcoded to true in the platform implementations, does it make any sense to add it there vs hardcoding it in sail?

Timmmm: It just makes it configurable without changing the Sail code. Though you're right, I will add an emulator flag.

Maybe with a warning because it's supposed to be enabled if you support virtual memory and IIRC you can't disable that currently.

jrtc27: Depends what privileged spec version you implement. We need to support all versions that people are making implementations for, so can't just go with the requirements of the latest version.

allenjbaum:
I think you need to define precisely what you mean by "disabling virtual
memory" means exactly.
By my standards (low as they are), setting SATP.MODE to Bare disables
virtual memory.
Clearing MISA.S also disables virtual memory...

handwritten_support/riscv_extras_sequential.lem
Timmmm: Good point - that's something I've been wondering about actually. What version of the spec does the model target?

If it is "all of them" I would have expected some kind of `has_spec(Priv1_12)` checks but those don't seem to exist. For example the very first thing I checked from the privileged spec changes:

> Changed MRET and SRET to clear mstatus.MPRV when leaving M-mode.

is just done unconditionally in the model:

      if   cur_privilege != Machine
      then mstatus->MPRV() = 0b0;

Another example: the 1.12 spec also introduces `mstatush` which is unconditionally implemented in the model.

Perhaps the intention is to add something `has_spec(Priv1_12)` but it just hasn't been done yet? Or maybe the plan for supporting multiple spec versions hasn't been fully formed... Any insight appreciated!

Anyway, I'll just add an emulator flag for now. :-)

model/riscv_sys_regs.sail
```diff
@@ -150,6 +150,9 @@ val sys_enable_fdext = {c: "sys_enable_fdext", ocaml: "Platform.enable_fdext", _
 val sys_enable_zfinx = {c: "sys_enable_zfinx", ocaml: "Platform.enable_zfinx", _: "sys_enable_zfinx"} : unit -> bool
 /* whether the N extension was enabled at boot */
 val sys_enable_next = {c: "sys_enable_next", ocaml: "Platform.enable_next", _: "sys_enable_next"} : unit -> bool
+/* Whether FIOM bit of menvcfg/senvcfg is implemented. It must be if supervisor mode
+   is implemented or non-bare addressing modes are supported. */
```
Timmmm, line 154: This comment is wrong - should be "It must be supported if supervisor mode is implemented *AND* non-bare addressing modes are supported." I'll update it.

Timmmm:
> I think you need to define precisely what you mean by "disabling virtual memory" means exactly.

Not sure what you are referring to here, but anyway here's the relevant bit of the spec:

> If S-mode is not supported, or **if satp.MODE is read-only zero (always Bare)**, the implementation may make FIOM read-only zero.

Now that I read it again, I think I misread it. The Sail model doesn't let you make `satp.MODE` be read-only zero but you can disable S-mode so that would be a legitimate way to also have FIOM disabled. I'll add an emulator flag.

c_emulator/riscv_sim.c
```diff
@@ -298,6 +299,11 @@ char *process_args(int argc, char **argv)
       fprintf(stderr, "enabling storing illegal instruction bits in mtval.\n");
       rv_mtval_has_illegal_inst_bits = true;
       break;
+    case 'o':
```
arichardson, line 302: It's now possible to use long-only options (see e.g. OPT_TRACE_OUTPUT), I wouldn't use one of the few remaining short options for such a rarely used one.

Timmmm: Ah thanks, I missed that! Looks like the C compiler could do with having all of its flags changed to match the OCaml one.

handwritten_support/riscv_extras_sequential.lem
Timmmm: Flag added.

model/riscv_sys_regs.sail
```diff
@@ -827,3 +830,45 @@ function read_seed_csr() -> xlenbits = {
...
+  // Fence of I/O implies Memory
+  FIOM   : 0,
+}
+
+register menvcfg : Envcfg
+register senvcfg : Envcfg
+
+function legalize_envcfg(o : Envcfg, v : bits(64)) -> Envcfg = {
+  let v = Mk_Envcfg(v);
+  let o = update_FIOM(o, v.FIOM());
```
arichardson, line 860: Does FIOM need to be hardwired to zero if it's not enabled?

arichardson (commented):
This looks good to me, just one question.

allenjbaum:
It appears to be allowed but not required :
    Priv 3.1.18:   If S-mode is not supported, or if satp.MODE is read-only
zero (always Bare), the implementation *may* make FIOM read-only zero.
Otherwise, it looks like it must be RW

model/riscv_sys_regs.sail
Timmmm: Good question! I think you're probably right - I'll double check.

Timmmm: Yeah looks like I totally forgot to actually hook up the config flag to this code, oops! I'll fix it.

Timmmm: Done.

Timmmm:
@arichardson any chance of a final review? Pretty please? :-)

arichardson:
> @arichardson any chance of a final review? Pretty please? :-)

I'm happy with the change, but you'll need it to be approved by one of the maintainers .

arichardson (approved):

billmcspadden-riscv (changes requested):
Reconsider the command line argument name.

c_emulator/riscv_platform.c
```diff
@@ -37,6 +37,11 @@ bool sys_enable_zfinx(unit u)
   return rv_enable_zfinx;
 }
 
+bool sys_enable_fiom(unit u)
+{
+  return rv_enable_fiom;
```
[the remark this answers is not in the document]
Timmmm: Agreed, that is a much better name! I've updated it everywhere.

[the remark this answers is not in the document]
billmcspadden-riscv: I don't see the changes.  What am I missing?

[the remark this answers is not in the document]
Timmmm: I think it's just because this comment stays linked to the first commit and gets marked "Outdated". I added a second commit with the rename. If you go to the "Files changed" tab at the top you should see it.

billmcspadden-riscv (approved):

jrtc27:
Commits were not squashed before merging

Timmmm:
> Commits were not squashed before merging

Sorry I didn't know the policy for this project, but I agree. Rebase merging like this is probably the worst of both worlds.

Personally I would disable `merge commits` and `rebase merging` in the project settings so you have to use `squash merge` (i.e. squash and rebase).

jrtc27:
> > Commits were not squashed before merging
> 
> Sorry I didn't know the policy for this project, but I agree. Rebase merging like this is probably the worst of both worlds.
> 
> Personally I would disable `merge commits` and `rebase merging` in the project settings so you have to use `squash merge` (i.e. squash and rebase).

We use rebase merging in other projects extensively. It works well if people curate their commits in the PR, and allows you to preserve a patch series as a patch series, rather than squashing it into one. Squashing a PR only works when PRs are logically a single change. Both have their uses, people just need to be disciplined when they come to merge to either squash when appropriate or check that the commit series is sensible.

Timmmm:
Doesn't it break `git bisect` though since only the last commit is actually tested by CI? I personally don't see much value in preserving history for changes that were never on `master`. Anyway I will squash in future!

PR #305, opened 2023-09-18
Remove duplicate shift definitions

Timmmm:
SHIFTW and SHIFTIWOP were duplicated. This did not cause any harm except that the disassembly for the SHIFTW version was incorrect. Therefore I removed that version.

The `execute()` functions were identical but the SHIFTW version is slightly neater (only one `[31..0]`) so I applied that to the SHIFTIWOP version.

This should cause no functional changes to the model except that the disassembly will have the extra `w` which is correct.


arichardson (approved):

Alasdair:
I think we should discuss this in the next meeting, or just merge it.

billmcspadden-riscv (approved):

PR #310, opened 2023-09-22
Add convenience targets for dual simulator builds

rpsene:
This commit introduces convenience targets that allow building both simulator binaries with a single command: `make csim_all`.


jrtc27 (changes requested):
I have a much cleaner version of this in sail-cheri-riscv. I’ll extract it out when I get back next week.

rpsene:
@jrtc27 what are the issues you see with the changes I have submitted? I would appreciate if you point the problems instead of just sending your version.

jrtc27:
It’s targeted at just csim and papers over the mess that is how ARCH is dealt with.

rpsene:
@jrtc27 as you have not submited your better version I will move forwad with the version I have submited. This is blocking a bunch of other improvements. 

@billmcspadden-riscv  who else can review this?

PR #312, opened 2023-10-04
Changes for BF16 support

adlr:
This is for the BF16 scalar + vector changes. Includes softfloat support for conversion between FP32 and BF16.


model/riscv_insts_zfbfmin.sail
```diff
@@ -0,0 +1,85 @@
+/* **************************************************************** */
+/* This file specifies the instructions in the Zfh extension        */
```
jrtc27, line 2: No it doesn't

model/riscv_insts_zvfbfmin.sail
```diff
@@ -0,0 +1,70 @@
...
+/*    Authors from RIOS Lab, Tsinghua University:                                  */
+/*      Xinlai Wan <xinlai.w@rioslab.org>                                          */
+/*      Xi Wang <xi.w@rioslab.org>                                                 */
+/*      Yifei Zhu <yifei.z@rioslab.org>                                            */
+/*      Shenwei Hu <shenwei.h@rioslab.org>                                         */
+/*      Kalvin Vu                                                                  */
+/*    Other contributors:                                                          */
+/*      Jessica Clarke <jrtc27@jrtc27.com>                                         */
+/*      Victor Moya <victor.moya@semidynamics.com>                                 */
+/*      Andrew de los Reyes <adlr@rivosinc.com>                                    */
```
jrtc27, line 12: These files are written by you only

model/riscv_sys_regs.sail
```diff
@@ -209,6 +209,12 @@ function haveZknd()   -> bool = true
 
 function haveZmmul()  -> bool = true
 function haveRVV()    -> bool = misa.V() == 0b1
+
+/* Zfbfmin (BF16) extension */
+function haveZfbfmin() -> bool = true
+function haveZvfbfmin() -> bool = { haveRVV() }
+function haveZvfbfwma() -> bool = { haveRVV() }
```
jrtc27, line 216: You don't need the curly braces

jrtc27 (commented):
1. This includes a snapshot of the vector changes, which makes this awkward to review other than by going through each bf16-specific commit one by one
2. Do these really need new softfloat C functions? There is the desire to stop using softfloat and instead have the floating point semantics entirely in Sail itself, so new functions should be pure Sail where possible

Commit messages are also a bit spartan; should mention extension names and versions at least, and ideally be one per extension, not all instructions dumped in one commit (see https://github.com/riscv/sail-riscv/blob/master/CONTRIBUTING.md; it's not a _large_ diff to actually add the instructions but still best to follow that approach).

jordancarlin:
Superseded by #1157, #1158, #1184, and #1185.

PR #315, opened 2023-10-10
Fix lem build error

Alasdair:
This should fix the build error that has been reported, although I think in general typing 'make' just does too much by default.


billmcspadden-riscv (approved):
Per the discussion at the tech-golden-model meeting on 2023-10-10,  this should be approved and merged.
Alasdair's comment about "make just does too much by default" will be handled in a separate issue/PR.

PR #316, opened 2023-10-10
Fix opam warning

Timmmm:
Replace `opam config var` with `opam var` as recommended endlessly by OPAM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,14 +113,14 @@ export SAIL_DIR
 EXPLICIT_COQ_SAIL=yes
 else
 # Use sail from opam package
-SAIL_DIR:=$(shell opam config var sail:share)
+SAIL_DIR:=$(shell opam var sail:share)
 SAIL:=sail
 endif
 SAIL_LIB_DIR:=$(SAIL_DIR)/lib
 export SAIL_LIB_DIR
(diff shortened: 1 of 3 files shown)


arichardson:
For the sail-cheri-riscv model I worked around this by setting OPAMCLI to keep 2.0 opam compatibility: https://github.com/CTSRD-CHERI/sail-cheri-riscv/commit/e13a829be6a288f7e447e9c51ac06abda37bab95

But I'm not sure if we still need compatibility with opam 2.0 or if we can assume at least 2.1

arichardson:
Ubuntu 20.04 and Debian 11 come with opam 2.0, so we probably still want to support that version?

Timmmm:
Ok cancel this; lets use @arichardson's fix instead.

arichardson:
> Ok cancel this; lets use @arichardson's fix instead.

I haven't made this change to the base model yet, I was just suggesting using the same OPAMCLI=2.0 approach here.

Timmmm:
Yep, sorry if I wasn't clear. I meant we should apply your fix here too. (I just haven't got around to actually doing it yet.) Feel free to beat me to it! 😄

PR #317, opened 2023-10-10
Rename string_of_int to dec_str

Timmmm:
And string_of_bits to bits_str. These are the names that Sail uses so it makes sense to use them.


Timmmm:
@billmcspadden-riscv this should be completely uncontroversial and is ready to merge.

arichardson (approved):

PR #321, opened 2023-10-17
Add unratified Zimop and Zcmop extensions

ved-rivos:
Add unratified Zimop and Zcmop extension specified in:
https://github.com/riscv/riscv-isa-manual/blob/main/src/zimop.adoc

### DIFF
--- a/model/riscv_insts_zcmop.sail
+++ b/model/riscv_insts_zcmop.sail
@@ -78,8 +78,8 @@
 /* ****************************************************************** */
 
 union clause ast = ZCMOP : (bits(3))
-mapping clause encdec_compressed = ZCMOP(mop_10_08) if haveZcmop()
-  <-> 0b01100 @ mop_10_08 : bits(3) @ 0b100000 @ 0b01
+mapping clause encdec_compressed = ZCMOP(mop) if haveZcmop()
+  <-> 0b01100 @ mop : bits(3) @ 0b100000 @ 0b01
 
 mapping zcmop_mnemonic : bits(3) <-> string = {
   0b000 <-> "c.mop.1",
--- a/model/riscv_insts_zcmop.sail
+++ b/model/riscv_insts_zcmop.sail
@@ -1,0 +1,100 @@
+/*=======================================================================================*/
+/*  RISCV Sail Model                                                                     */
+/*                                                                                       */
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except for the snapshots of the Lem and Sail libraries                   */
+/*  in the prover_snapshots directory (which include copies of their                     */
+/*  licences), is subject to the BSD two-clause licence below.                           */
+/*                                                                                       */
+/*  Copyright (c) 2017-2023                                                              */
+/*    Ved Shanbhogue                                                                     */
+/*    Prashanth Mundkur                                                                  */
+/*    Rishiyur S. Nikhil and Bluespec, Inc.                                              */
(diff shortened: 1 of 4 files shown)


model/riscv_insts_zcmop.sail
```diff
@@ -0,0 +1,100 @@
...
+/* ********************************************************************* */
+/* This file specifies the compressed insts. in the 'Zcmop' extension.   */
+
+/* These instructions are only legal if misa.C() is true and is checked in
+ * the fetch-execute logic.
+ */
+/* ****************************************************************** */
+
+union clause ast = ZCMOP : (bits(3))
+mapping clause encdec_compressed = ZCMOP(b1008) if haveZcmop()
```
Alasdair, line 81: I'm not sure about the b1008 name, maybe:
```
mapping clause encdec_compressed = ZCMOP(mop) if haveZcmop()
  <-> 0b01100 @ mop : bits(3) @ 0b100000 @ 0b01
```
would be better?

model/riscv_insts_zimop.sail
```diff
@@ -0,0 +1,145 @@
...
+/*=======================================================================================*/
+
+/* ****************************************************************** */
+/* This file specifies the instructions in the Zimop extension.       */
+/* mop.r.0-31: 1-00--0111--sssss100ddddd1110011                       */
+/* mop.rr.0-7: 1-00--1tttttsssss100ddddd1110011                       */
+union clause ast = ZIMOP_MOP_R : (bits(5), regidx, regidx)
+union clause ast = ZIMOP_MOP_RR : (bits(3), regidx, regidx, regidx)
+
+mapping clause encdec = ZIMOP_MOP_R(b30 @ b2726 @ b2120, rs1, rd) if haveZimop()
```
Alasdair, line 79: Similar to my comment on the Zcmop file, maybe instead of `b30`, `b2726`, `b2120` these could be `mop_30`, `mop_27_26`, and `mop_21_20`? And the same for the MOP_RR below

Alasdair (commented):
Seems simple enough. Not sure what happens if another extension comes along and uses some of the MOP space, then I guess we'd need some way to configure which MOPs this extension has?

Maybe `haveZimop` and `haveZcmop` should take the mop bits as an argument (even if they just always return `true` for now) to represent this?

model/riscv_insts_zcmop.sail
ved-rivos: The `haveZimop` and `haveZcmop` would have to be present because a future extension that needs them would need to depend on them. Zicfiss is the first extension planning to use them and the idea is to branch out of the ZCMOP/ZIMOP for the few code points that Zicfiss uses - that would be controlled by `haveZicfiss`.

model/riscv_insts_zimop.sail
ved-rivos: Updated as suggested.

Alasdair:
Looks good to me. Perhaps add to the agenda for next meeting @billmcspadden-riscv ?

model/riscv_insts_zcmop.sail
ved-rivos: updated to use name `mop` as the arguement.

ved-rivos:
@Alasdair @billmcspadden-riscv I need to use these to add the Zicfiss extension that uses the Zimop/Zcmop codepoints. As PR did not get accepted, I had to include the contents of this PR into a PR I will be providing for Zicfiss and will close this PR. Thanks much for the review and feedback.

PR #322, opened 2023-10-18
Adding Zcb implementation

junambi:
Zcb implementation for review and merger.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ else
 endif
 
 SAIL_FLEN := riscv_flen_D.sail
-SAIL_VLEN := riscv_vlen.sail
 
 # Instruction sources, depending on target
 SAIL_CHECK_SRCS = riscv_addr_checks_common.sail riscv_addr_checks.sail riscv_misa_ext.sail
@@ -28,6 +27,8 @@ SAIL_DEFAULT_INST += riscv_insts_zba.sail
 SAIL_DEFAULT_INST += riscv_insts_zbb.sail
 SAIL_DEFAULT_INST += riscv_insts_zbc.sail
 SAIL_DEFAULT_INST += riscv_insts_zbs.sail
+
+SAIL_DEFAULT_INST += riscv_insts_zcb.sail
 
 SAIL_DEFAULT_INST += riscv_insts_zfh.sail
 # Zfa needs to be added after fext, dext and Zfh (as it needs
@@ -42,15 +43,6 @@ SAIL_DEFAULT_INST += riscv_insts_zbkx.sail
 
 SAIL_DEFAULT_INST += riscv_insts_zicond.sail
 
-SAIL_DEFAULT_INST += riscv_insts_vext_utils.sail
-SAIL_DEFAULT_INST += riscv_insts_vext_vset.sail
-SAIL_DEFAULT_INST += riscv_insts_vext_arith.sail
-SAIL_DEFAULT_INST += riscv_insts_vext_fp.sail
-SAIL_DEFAULT_INST += riscv_insts_vext_mem.sail
-SAIL_DEFAULT_INST += riscv_insts_vext_mask.sail
-SAIL_DEFAULT_INST += riscv_insts_vext_vm.sail
-SAIL_DEFAULT_INST += riscv_insts_vext_red.sail
-
 SAIL_SEQ_INST  = $(SAIL_DEFAULT_INST) riscv_jalr_seq.sail
 SAIL_RMEM_INST = $(SAIL_DEFAULT_INST) riscv_jalr_rmem.sail riscv_insts_rmem.sail
 
@@ -59,7 +51,6 @@ SAIL_RMEM_INST_SRCS = riscv_insts_begin.sail $(SAIL_RMEM_INST) riscv_insts_end.s
 
 # System and platform sources
 SAIL_SYS_SRCS =  riscv_csr_map.sail
-SAIL_SYS_SRCS += riscv_vext_control.sail    # helpers for the 'V' extension
 SAIL_SYS_SRCS += riscv_next_regs.sail
 SAIL_SYS_SRCS += riscv_sys_exceptions.sail  # default basic helpers for exception handling
 SAIL_SYS_SRCS += riscv_sync_exception.sail  # define the exception structure used in the model
@@ -79,12 +70,11 @@ SAIL_VM_SRCS += $(SAIL_RV64_VM_SRCS)
 endif
 
 # Non-instruction sources
-PRELUDE = prelude.sail prelude_mapping.sail $(SAIL_XLEN) $(SAIL_FLEN) $(SAIL_VLEN) prelude_mem_metadata.sail prelude_mem.sail
+PRELUDE = prelude.sail prelude_mapping.sail $(SAIL_XLEN) $(SAIL_FLEN) prelude_mem_metadata.sail prelude_mem.sail
 
 SAIL_REGS_SRCS = riscv_reg_type.sail riscv_freg_type.sail riscv_regs.sail riscv_pc_access.sail riscv_sys_regs.sail
 SAIL_REGS_SRCS += riscv_pmp_regs.sail riscv_pmp_control.sail
 SAIL_REGS_SRCS += riscv_ext_regs.sail $(SAIL_CHECK_SRCS)
-SAIL_REGS_SRCS += riscv_vreg_type.sail riscv_vext_regs.sail
 
 SAIL_ARCH_SRCS = $(PRELUDE)
 SAIL_ARCH_SRCS += riscv_types_common.sail riscv_types_ext.sail riscv_types.sail
--- a/Makefile
+++ b/Makefile
@@ -29,9 +29,10 @@ SAIL_DEFAULT_INST += riscv_insts_zbb.sail
 SAIL_DEFAULT_INST += riscv_insts_zbc.sail
 SAIL_DEFAULT_INST += riscv_insts_zbs.sail
 
+SAIL_DEFAULT_INST += riscv_insts_zcb.sail
+
 SAIL_DEFAULT_INST += riscv_insts_zfh.sail
 
-SAIL_DEFAULT_INST += riscv_insts_zcb.sail
 
 # Zfa needs to be added after fext, dext and Zfh (as it needs
 # definitions from those)
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ else
 endif
 
 SAIL_FLEN := riscv_flen_D.sail
+SAIL_VLEN := riscv_vlen.sail
 
 # Instruction sources, depending on target
 SAIL_CHECK_SRCS = riscv_addr_checks_common.sail riscv_addr_checks.sail riscv_misa_ext.sail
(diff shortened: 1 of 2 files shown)


Timmmm:
Following the discussion in Zoom I had a closer look at this and almost all of these instructions map very clearly to existing instructions. Only 2 don't map cleanly and would maybe be clearer reimplemented.

* `c.lbu` -> `lbu`
* `c.lhu` -> `lhu`
* `c.lh` -> `lh`
* `c.sb` -> `sb`
* `c.sh` -> `sh`
* `c.zext.b` -> reimplement?
* `c.sext.b` -> `sext.b`
* `c.zext.h` -> `zext.h`
* `c.sext.h` -> `sext.h`
* `c.not` -> reimplement?
* `c.mul` -> `mul`

For example instead of:

```
function clause execute (C_MUL(rsd,rs2)) = {
  let rd = creg2reg_idx(rsd);
  let rs = creg2reg_idx(rs2);

  let result_wide = to_bits(2 * sizeof(xlen), signed(X(rd)) * signed(X(rs)));
  X(rd) = result_wide[(sizeof(xlen) - 1) .. 0];
  RETIRE_SUCCESS
}
```

Use:

```
function clause execute (C_MUL(rsd,rs2)) = {
  let rd = creg2reg_idx(rsd);
  let rs = creg2reg_idx(rs2);

  execute(MUL((rs, rd, rd, false, true, true))
}
```

Primitive obsession notwithstanding, I think this is easier to understand, matches the existing C instructions and is less error prone. The intention of this extension is just to have shorter decodings, so the implementation *should* match.

For example in this case (which I picked randomly!), reimplementing it introduced a bug, whereas just mapping it would not have. I think that's probably a strong enough argument on its own, given the lack of testing.

jrtc27:
> Following the discussion in Zoom I had a closer look at this and almost all of these instructions map very clearly to existing instructions. Only 2 don't map cleanly and would maybe be clearer reimplemented.
> 
> * `c.lbu` -> `lbu`
> * `c.lhu` -> `lhu`
> * `c.lh` -> `lh`
> * `c.sb` -> `sb`
> * `c.sh` -> `sh`
> * `c.zext.b` -> reimplement?
> * `c.sext.b` -> `sext.b`
> * `c.zext.h` -> `zext.h`
> * `c.sext.h` -> `sext.h`
> * `c.not` -> reimplement?
> * `c.mul` -> `mul`
> 
> For example instead of:
> 
> ```
> function clause execute (C_MUL(rsd,rs2)) = {
>   let rd = creg2reg_idx(rsd);
>   let rs = creg2reg_idx(rs2);
> 
>   let result_wide = to_bits(2 * sizeof(xlen), signed(X(rd)) * signed(X(rs)));
>   X(rd) = result_wide[(sizeof(xlen) - 1) .. 0];
>   RETIRE_SUCCESS
> }
> ```
> 
> Use:
> 
> ```
> function clause execute (C_MUL(rsd,rs2)) = {
>   let rd = creg2reg_idx(rsd);
>   let rs = creg2reg_idx(rs2);
> 
>   execute(MUL((rs, rd, rd, false, true, true))
> }
> ```
> 
> Primitive obsession notwithstanding, I think this is easier to understand, matches the existing C instructions and is less error prone. The intention of this extension is just to have shorter decodings, so the implementation _should_ match.
> 
> For example in this case (which I picked randomly!), reimplementing it introduced a bug, whereas just mapping it would not have. I think that's probably a strong enough argument on its own, given the lack of testing.

Yes. See https://github.com/riscv/sail-riscv/pull/255#discussion_r1169123796.

martinberger:
I just asked Tariq Kurd, the original author of the spec what he thinks we should do:

1. Simply forward to the corresponding non-compressed instruction.
2. Implement the spec faithfully to RISC-V Instruction Set Manual (so not forwarding).

Tariq said we should do (1), that was always the intent - and demonstrates that the instructions are trivial. This is also how the other compressed instructions are implemented, e.g. `c.addi`

abukharmeh:
Same thoughts https://github.com/riscv/sail-riscv/pull/255#discussion_r1170505749

junambi:
c.lbu -> lbu                             :: (Is implemented as part of LOAD), i do not see a separate LBU implementation
    c.lhu -> lhu                             :: (Is implemented as part of LOAD), i do not see a separate LHU implementation
    c.lh -> lh                                 :: (Is implemented as part of LOAD), i do not see a separate LH implementation
    c.sb -> sb                               :: (Is implemented as part of STORE), i do not see a separate SB implementation
    c.sh -> sh                               :: (Is implemented as part of STORE), i do not see a separate SH implementation
    c.zext.b -> reimplement?        :: Is the current implementation fine for this
    c.sext.b -> sext.b                    :: Will redo as suggested reusing SEXT_B
    c.zext.h -> zext.h                    :: Will redo as suggested reusing ZEXT_H
    c.sext.h -> sext.h                    :: Will redo as suggested reusing SEXT_H
    c.not -> reimplement?        :: Is the current implementation fine for this
    c.mul -> mul                           :: Will redo as suggested reusing MUL

Just for my understanding, can one of you please clarify the following,
1. What are the advantages of using the existing instruction implementation, will it not affect the performance of the sail code
2. In HW, having to implement other instructions (except the base) to support compressed instructions, will it not affect the area, power and performance of the system as a whole / is it required for HW to implement many other extensions to support only the base and compressed instructions.

junambi:
Thank you Martin,
Are the other listed implementation in my previous comment, good to implement.

Thank you.

Timmmm:
> In HW, having to implement other instructions (except the base) to support compressed instructions, will it not affect the area, power and performance of the system as a whole / is it required for HW to implement many other extensions to support only the base and compressed instructions.

These instructions are only supported if the instructions they map to are are implemented. That was the bug in your `c.mul` implementation that I mentioned: `c.mul` is not supported if `mul` is not implemented.

junambi:
Yes Tim, i understood that part. 

I was trying to understand, does this not put a limitation on the implementation, as someone could choose to implement only the compressed instructions. 

Is this even allowed/supported.

Thank you.

Timmmm:
> as someone could choose to implement only the compressed instructions.

They can't. The ones that are mapped to optional instructions require those instructions to be implemented. For example here's what the spec says about `c.mul`:

image

Your `c.mul` implementation was missing the check for this, but the `mul` implementation has it:

```
function clause execute (MUL(rs2, rs1, rd, high, signed1, signed2)) = {
  if haveMulDiv() | haveZmmul() then {
```

jrtc27:
... and there's no sensible use case for it, either; if you have the hardware to do a c.mul, you might as well use it to do a mul too. Decode is very little cost compared to the actual multiplier.

junambi:
Modified the implementation and pushed the file, please review and comment.

model/riscv_insts_zcb.sail
```diff
@@ -0,0 +1,292 @@
...
+mapping clause assembly = C_LBU(uimm, rsc, rdc)
+  <-> "c.lbu" ^ spc() ^ creg_name(rdc) ^ sep() ^ creg_name(rsc) ^ sep() ^ hex_bits_5(uimm @ 0b000)
+/* ****************************************************************** */
+union clause ast = C_LHU : (bits(1), cregidx, cregidx)
+
+mapping clause encdec_compressed = C_LHU(ui5, rs1, rd)
+  <-> 0b100 @ 0b001 @ rs1 : cregidx @ 0b0 @ ui5 : bits(1) @ rd : cregidx @ 0b00
+
+function clause execute (C_LHU(uimm, rsc, rdc)) = {
+  let immext : xlenbits = zero_extend(uimm);
```
Timmmm, line 102: Also unused. Please check the other instructions too.

model/riscv_insts_zcb.sail
```diff
@@ -0,0 +1,292 @@
...
+  let imm : bits(12) = zero_extend(uimm);
```
Timmmm, line 103: Also wrong - should be `zero_extend(ui1 @ 0b0);`

Please check the other instructions too.

model/riscv_insts_zcb.sail
```diff
@@ -0,0 +1,292 @@
...
+/* Sail Model for the c.mul encoding group                            */
+/* ------------------------------------------------------------------ */
+
+/* the assembly abstract syntax tree (AST) clause for the CMUL instructions */
+union clause ast = C_MUL : (cregidx,cregidx)
+
+/* the encode/decode mapping between AST elements and 16-bit half words */
+
+mapping clause encdec_compressed = C_MUL(rsd, rs2)
+  <-> 0b100 @ 0b111 @ rsd : cregidx @ 0b00 @ rs2 : cregidx @ 0b01
```
Timmmm, line 260: `@ 0b00 @` should be `@ 0b10 @`

jrtc27:
Given these review comments, are you testing your code before pushing it for review, and if so how? Because basic things like decode patterns should be caught by even the most basic testing.

Also you've still deleted a bunch of V-related changes from Makefile which causes the build to fail in CI. So I don't think you could possibly have tested this code given it won't build.

junambi:
Thank you Tim and Jessica for the comments, i have updated the Makefile and the Zcb model.

Makefile
```diff
@@ -30,6 +30,9 @@ SAIL_DEFAULT_INST += riscv_insts_zbc.sail
 SAIL_DEFAULT_INST += riscv_insts_zbs.sail
 
 SAIL_DEFAULT_INST += riscv_insts_zfh.sail
+
+SAIL_DEFAULT_INST += riscv_insts_zcb.sail
```
arichardson, line 34: It would be good to keep this list sorted alphabetically, i.e. move it up one.

model/riscv_insts_zcb.sail
```diff
@@ -0,0 +1,288 @@
...
+  let rs = creg2reg_idx(rsc);
+  execute(LOAD(imm, rs, rd, true, BYTE, false, false))
+}
+
+mapping clause assembly = C_LBU(uimm, rsc, rdc)
+  <-> "c.lbu" ^ spc() ^ creg_name(rdc) ^ sep() ^ creg_name(rsc) ^ sep() ^ hex_bits_5(uimm @ 0b000)
+/* ****************************************************************** */
+union clause ast = C_LHU : (bits(1), cregidx, cregidx)
+
+mapping clause encdec_compressed = C_LHU(ui1, rs1, rd)
```
arichardson, line 97:
```suggestion
mapping clause encdec_compressed = C_LHU(ui1 @Ob0, rs1, rd)
```

Immediate is not scaled in current code. I assume this means there are no tests?

Timmmm:
> Thank you Tim and Jessica for the comments, i have updated the Makefile and the Zcb model.

Looks like you didn't fix the issues with the immediates though?

jjscheel:
> Given these review comments, are you testing your code before pushing it for review, and if so how? Because basic things like decode patterns should be caught by even the most basic testing.

Hi, @jrtc27, fair question.  I asked @junambi to post his current PR work early because @martinberger had interest in the work and we wanted to make sure that we didn't duplicating previous effort.

Makefile
junambi: This has been updated.

model/riscv_insts_zcb.sail
```diff
@@ -0,0 +1,292 @@
...
+
+/* ****************************************************************** */
+union clause ast = C_LBU : (bits(2), cregidx, cregidx)
+
+mapping clause encdec_compressed = C_LBU(ui65, rs1, rd)
+  <-> 0b100 @ 0b000 @ rs1 : cregidx @ ui65 : bits(2) @ rd : cregidx @ 0b00
+
+function clause execute (C_LBU(uimm, rsc, rdc)) = {
+  let immext : xlenbits = zero_extend(uimm);
+  let imm : bits(12) = zero_extend(uimm);
```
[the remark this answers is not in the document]
jrtc27: Swizzle the bits in decode not execute

jjscheel:
@martinberger, what is the status of your work?  I can't tell if a PR was submitted.

Timmmm:
There isn't a PR. I thought Martin added our implementation to a comment but I can't see it now.  I'll make a new PR.

jjscheel:
Thanks, @Timmmm and @martinberger!

PR #324, opened 2023-10-25
Assembly for hint instructions needs spaces after mnemonics

ThinkOpenly:
I admit I'm not quite sure that these assembly representations really define valid mnemonics.

If they do, however, (and maybe regardless) then they certainly seem to need spaces between the mnemonics and the operands, like the assembly representations everywhere else.

### DIFF
--- a/model/riscv_insts_hints.sail
+++ b/model/riscv_insts_hints.sail
@@ -82,7 +82,7 @@ mapping clause encdec_compressed = C_NOP_HINT(im5 @ im40)
 
 function clause execute C_NOP_HINT(imm) = RETIRE_SUCCESS
 
-mapping clause assembly = C_NOP_HINT(imm) <-> "c.nop.hint." ^ hex_bits_6(imm)
+mapping clause assembly = C_NOP_HINT(imm) <-> "c.nop.hint." ^ spc() ^ hex_bits_6(imm)
 
 /* ****************************************************************** */
 union clause ast = C_ADDI_HINT : (regidx)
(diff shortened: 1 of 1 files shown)


jrtc27:
It's all a bunch of nonsense; see https://github.com/riscv/sail-riscv/issues/112. If you're touching this code, perhaps you can be persuaded to make the code more sensible and mirror the actual assembly for them?

Note though that this made-up syntax puts a dot at the end of what you've deemed the mnemonic, and so they really do think of the immediate operand as being part of the mnemonic. If you're putting a space after the dot then that just looks even weirder. So either leave them as they are or fix them properly?

jrtc27 (changes requested):
(see comment)

jjscheel:
> It's all a bunch of nonsense; see #112. If you're touching this code, perhaps you can be persuaded to make the code more sensible and mirror the actual assembly for them?

@jrtc27, I'm sure @ThinkOpenly would appreciate a more descriptive response that just "nonsense".  While the the link is nice, highlighting the key point would be incredibly helpful not only to Paul but to others trying to learn how to contribute here too.

jrtc27:
> > It's all a bunch of nonsense; see #112. If you're touching this code, perhaps you can be persuaded to make the code more sensible and mirror the actual assembly for them?
> 
> @jrtc27, I'm sure @ThinkOpenly would appreciate a more descriptive response that just "nonsense". While the the link is nice, highlighting the key point would be incredibly helpful not only to Paul but to others trying to learn how to contribute here too.

The link *is* the key point. It’s a paragraph that explains the problem.

ThinkOpenly:
I'm fine moving the discussion to #112. I've just added a comment/question there. I'll close this now.

PR #327, opened 2023-10-26
Add Svadu support

defermelowie:
Add unratified Svadu extension specified in:
https://github.com/riscv/riscv-svadu/blob/main/svadu.adoc

As for testing, I [modified](https://gitlab.kuleuven.be/u0165022/riscv-tests/-/commits/svadu) the dirty test from https://github.com/riscv-software-src/riscv-tests. The new 64-bit test passes however the added 32-bit test [fails](https://github.com/defermelowie/sail-riscv/actions/runs/6682585563/job/18157923770#step:7:930) due to https://github.com/rems-project/linksem/issues/27


arichardson:
Thanks for working on this. However, there is already https://github.com/riscv/sail-riscv/pull/260 (which has been open for 5 months now). It is quite sad to see that the slow review process is causing duplicate work. And TBH this is also making me reconsider whether it's worth my time contributing to sail in the future...

PR #329, opened 2023-10-27
Add a simpler default Makefile target

Alasdair:
Changes the Makefile so it has a `default` target that just builds the C simulator. Previously running `make` ran `make all` which would try to build targets for Isabelle/HOL4/Coq, which probably isn't useful for most users. Furthermore, these targets all try to produce more usable definitions for users of these tools by avoiding doing things like renaming and name-mangling, but this can sometimes make them slightly brittle when underlying libraries change.

Adding a simpler default target for what is most likely the common use case should provide a better experience.

In addition, I have adjusted the run_test script so it only uses the OCaml emulator if `TEST_OCAML` is set. In addition to matching the default Makefile behavior, this is also because the generated OCaml we produce hits some bad performance issues with certain versions of OCaml. I also want to improve the OCaml generation in some other ways, and not having it be the default should give me some more flexibility to do so.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,6 +196,8 @@ endif
 
 .PHONY:
 
+default: c_emulator/riscv_sim_$(ARCH)
+
 all: ocaml_emulator/riscv_ocaml_sim_$(ARCH) c_emulator/riscv_sim_$(ARCH) riscv_isa riscv_coq riscv_hol riscv_rmem
 .PHONY: all
 
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -55,30 +55,33 @@ cd $RISCVDIR
 make clean
 
 printf "Building 32-bit RISCV specification...\n"
-if ARCH=RV32 make ocaml_emulator/riscv_ocaml_sim_RV32 ;
-then
-    green "Building 32-bit RISCV OCaml emulator" "ok"
-else
-    red "Building 32-bit RISCV OCaml emulator" "fail"
-fi
-for test in $DIR/riscv-tests/rv32*.elf; do
-    # skip F/D tests on OCaml for now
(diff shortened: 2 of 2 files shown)


arichardson:
Looks like the ci needs to be updated?

jrtc27:
CI indeed needs to test building the OCaml version otherwise it will rot.

Also, it’s pretty odd for make to not be the same as make all; that’s a very common convention. I would caution against breaking that convention.

Alasdair:
I don't think it's a good convention for this project.

jrtc27:
all does not have to mean "everything under the sun you could ever possibly build". For better or worse, it means "all the *configured* things". There are plenty of projects that turn things off by default for make all, but it's still make all.

I get quite irritated when projects gratuitously deviate from convention, and this would count as that, from my experience with open-source software. The precedent is for this to still be called all.

Alasdair:
Ok, fine I won't bother with this or any subsequent CI improvements I was working on

jrtc27:
I'm not saying throw it away, just have make all be the thing you want to be the default

jrtc27:
(I completely agree that having make all include all the theorem proving targets is not a good experience)

Alasdair:
Ok, perhaps I was being churlish. The question is what should `all` actually do. My thinking right now is it should give you something you can use to run the tests, perhaps it should build both the RV32 and RV64 simulators.

I was then going to have something like a `make everything` target for CI, although in practice it might be better to split it into multiple CI actions to run things in parallel. I'm not worried too much about the OCaml support bit-rotting because I have extensive CI tests on the Sail side, and we'd run it in CI here.

jordancarlin:
@Alasdair are any of these other changes still desired now that #464 has been merged or can this be closed?

PR #330, opened 2023-10-27
Some simple CI housekeeping

Alasdair:
Update the versions of some of the Github actions, as some of the older action versions used parts of the github API that are deprecated and due to be removed.


jrtc27 (approved):
Looks good to me if it all works ok

Alasdair:
Looks like the upload-artifacts action also needs updating for node16 as well. I set this branch as the default branch on my fork, so it should run it there as if it was here.

PR #331, opened 2023-10-27
Fix issue with trace_log in printf

Alasdair:
Currently when fopen fails, the code attempts to print trace_log, which has type `FILE *`. I think it intends to print the path instead. Fixes the clang warning:
```
c_emulator/riscv_sim.c:1050:48: warning: format ‘%s’ expects argument of type ‘char *’, but argument 3 has type ‘FILE *’ [-Wformat=]
 1050 |     fprintf(stderr, "Cannot create trace log '%s': %s\n", trace_log,
      |                                               ~^          ~~~~~~~~~
      |                                                |          |
      |                                                char *     FILE *
```

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -1047,7 +1047,7 @@ void init_logs()
   if (trace_log_path == NULL) {
     trace_log = stdout;
   } else if ((trace_log = fopen(trace_log_path, "w+")) < 0) {
-    fprintf(stderr, "Cannot create trace log '%s': %s\n", trace_log,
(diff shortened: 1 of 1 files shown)


jrtc27 (approved):
Add a Fixes 5725f3fe08a2c1b8588b9a4416d91c297e1e5c00?

Alasdair:
I made the first line a bit more descriptive and added a reference to that commit in the body

arichardson (approved):

arichardson:
Ping?

jrtc27:
Taking matters into my own hands; this is just ridiculous, these kinds of obvious fixes need to not take weeks to get in

PR #335, opened 2023-10-30
Add Zabha extension

ved-rivos:
Specification: https://github.com/riscv/riscv-zabha

Tested with RISC-V arch. tests and checked against spike.

### DIFF
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -21,7 +21,7 @@ function aqrl_str(aq : bool, rl : bool) -> string =
   }
 
 function lrsc_width_str(width : word_width) -> string =
-  match (width) {
+  match width {
     BYTE   => ".b",
     HALF   => ".h",
     WORD   => ".w",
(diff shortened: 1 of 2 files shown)


model/riscv_insts_aext.sail
```diff
@@ -29,23 +29,30 @@ function lrsc_width_str(width : word_width) -> string =
...
  */
-function amo_width_valid(size : word_width) -> bool = {
+function lrsc_width_valid(size : word_width) -> bool = {
   match(size) {
     WORD   => true,
     DOUBLE => sizeof(xlen) >= 64,
     _      => false
   }
 }
+function amo_width_valid(size : word_width) -> bool = {
```
Timmmm, line 42: Can you just add a blank line between these functions? Otherwise this LGTM.

Timmmm (approved):
LGTM

Timmmm:
@billmcspadden-riscv any objection to merging this? I can't see any reason not to.

model/riscv_insts_aext.sail
```diff
@@ -29,23 +29,31 @@ function lrsc_width_str(width : word_width) -> string =
...
+function lrsc_width_valid(size : word_width) -> bool = {
   match(size) {
     WORD   => true,
     DOUBLE => sizeof(xlen) >= 64,
     _      => false
   }
 }
 
+function amo_width_valid(size : word_width) -> bool = {
+  match(size) {
```
jrtc27, line 44: Nit: the style of the existing amo_width_valid was wrong, there should be a space after `match`, and the parens also aren't needed (though that mistake is a little more common in the model). As you're copying this code could you please fix it first so we don't propagate more instances of the same mistake?

ved-rivos: updated.

Timmmm:
I think this is good to go. I'll merge this in a few days if nobody objects.

Alasdair (approved):

arichardson (approved):

PR #342, opened 2023-11-13
Add Ssdtso support

cmuellner:
The second draft of the Ssdtso extension was published recently:
  https://lists.riscv.org/g/tech-arch-review/message/183
This PR adds support for this new extension.

Note, that the Ssdtso specification is in development state (i.e., not frozen or even ratified).


Alasdair:
lgtm

While currently we only have support for RVWMO for sail-riscv in [Isla](https://github.com/rems-project/isla), I don't think there is any technical reason why we couldn't have TSO as well.

cmuellner:
I'm revisiting all Ssdtso patches. While doing so, I've rebased this PR (no changes).

PR #349, opened 2023-11-17
Remove duplicate sep

Timmmm:
This is defined already by Sail now.


Timmmm:
... maybe not by the Sail version used in CI I guess!

Issue #325, opened 2023-10-25
Fail to run 'make' for sail-riscv


xiaomaaaaa:
I'm on `sail -v` = `Sail 0.16 (sail @ opam-v2.1.5)`
When trying to run 'make' for sail-riscv on Centos7 I get the following error:

```
Warning: Incomplete pattern match statement at model/riscv_insts_vext_fp.sail:722.77-82:
722 |                              let (fflags, elem) : (bits_fflags, bits('m)) = match 'm {
    |                                                                             ^---^
    | 
The following expression is unmatched: 64

Warning: Incomplete pattern match statement at model/riscv_insts_vext_fp.sail:731.77-82:
731 |                              let (fflags, elem) : (bits_fflags, bits('m)) = match 'm {
    |                                                                             ^---^
    | 
The following expression is unmatched: 64

Warning: Incomplete pattern match statement at model/riscv_insts_vext_red.sail:213.12-17:
213 |      sum = match funct6 {
    |            ^---^
    | 
The following expression is unmatched: FVV_VFWREDOSUM

echo "declare {isabelle} rename field sync_exception_ext = sync_exception_ext_exception" >> generated_definitions/lem/RV64/riscv_types.lem 
lem -isa -outdir generated_definitions/isabelle/RV64 -lib Sail=/home/mashixiong0318004377/.opam/5.1.0/share/sail/src/lem_interp -lib Sail=/home/mashixiong0318004377/.opam/5.1.0/share/sail/src/gen_lib \
	handwritten_support/0.11/riscv_extras.lem handwritten_support/0.11/mem_metadata.lem handwritten_support/0.11/riscv_extras_fdext.lem \
	generated_definitions/lem/RV64/riscv_types.lem \
	generated_definitions/lem/RV64/riscv.lem
File "generated_definitions/lem/RV64/riscv.lem", line 24483, character 7:
  Syntax error
make: *** [Makefile:330: generated_definitions/isabelle/RV64/Riscv.thy] Error 1
```

and the content of the ‘riscv.lem’ is as follows:

```
let riscv_f16Rsqrte7 rm v =
   let class = (fp_class v  : mword ty16) in      #line 24483
   let b__0 = class in
   if ((eq_vec b__0 (0x0001 : mword ty16))) then
     return ((nvFlag ()  : mword ty5), (0x7E00 : mword ty16))
   else if ((eq_vec b__0 (0x0002 : mword ty16))) then
[...]

PR #353, opened 2023-11-18
Fix lem build in Makefile

Alasdair:
'class' is a reserved keyword in lem, so the use of class as a variable name was causing issues. While ultimately this should be fixed in Sail this can easily be worked around here.

Fortunately the code in questions was always using a pattern like
```
let class = f(x);
let y = match class {
  ...
};
y
```
which can be simplified to
```
match f(x) {
  ...
}
```
So I didn't have to invent a new variable name, and got to make the code simpler in the process!

Fixing this unfortunately causes some other issues to become apparent. The machine words library can't handle the large vlenmax, so we have to create a separate file with a smaller vlenmax for the theorem provers. There were also some other issues, so I switched out the default representation to bitlists until they can be resolved, and cleaned up some legacy cruft in handwritten_support while I was there.

The rmem target has similar issues, but as this is only really useful for rmem maintainers it doesn't make sense to run by default, so I removed it from the default set of targets.

This should fix https://github.com/riscv/sail-riscv/issues/325


Alasdair:
I split the commits so the actual model changes are in the first commit (which are all small refactorings).

The second commit contains all the lem hackery to fix the build

model/riscv_insts_vext_utils.sail
```diff
@@ -976,8 +976,7 @@ function rsqrt7 (v, sub) = {
 
 val riscv_f16Rsqrte7 : (bits_rm, bits_H) -> (bits_fflags, bits_H) effect {rreg}
```
billmcspadden-riscv, line 977: Remove effect throughout.

model/riscv_insts_vext_utils.sail
```diff
@@ -1006,14 +1003,12 @@ function riscv_f32Rsqrte7 (rm, v) = {
     0x0080 => (zeros(5), 0x00000000),
     0x0020 => (zeros(5), rsqrt7(v, true)[31 .. 0]),
     _      => (zeros(5), rsqrt7(v, false)[31 .. 0])
-  };
-  (fflags, result)
+  }
 }
 
 val riscv_f64Rsqrte7 : (bits_rm, bits_D) -> (bits_fflags, bits_D) effect {rreg}
```
billmcspadden-riscv, line 1009: Remove effect

model/riscv_insts_vext_utils.sail
```diff
@@ -988,14 +987,12 @@ function riscv_f16Rsqrte7 (rm, v) = {
     0x0080 => (zeros(5), 0x0000),
     0x0020 => (zeros(5), rsqrt7(v, true)[15 .. 0]),
     _      => (zeros(5), rsqrt7(v, false)[15 .. 0])
-  };
-  (fflags, result)
+  }
 }
 
 val riscv_f32Rsqrte7 : (bits_rm, bits_S) -> (bits_fflags, bits_S) effect {rreg}
```
billmcspadden-riscv, line 993: Remove effect.

model/riscv_vlen256.sail
billmcspadden-riscv, line 1: Why is this part of this PR?  Seems like an orthogonal addition.

billmcspadden-riscv (commented):
My review is mainly questions.  I don't see anything wrong in the code and will approve the PR once the questions are answered.

One exception:  the 'effects' of certain functions need to be removed, yes?

Unable to comment on lem code due to lack of knowledge.

Makefile
```diff
@@ -184,15 +185,7 @@ C_LIBS += $(SAIL_LIB_DIR)/coverage/libsail_coverage.a -lm -lpthread -ldl
 endif
 
 RISCV_EXTRAS_LEM_FILES = riscv_extras.lem mem_metadata.lem riscv_extras_fdext.lem
-# Feature detect if we are on the latest development version of Sail
```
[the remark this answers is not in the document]
Alasdair: It's a check for a version of Sail that is many years old, and no longer works with the model.

model/riscv_insts_vext_utils.sail
Alasdair: These lines are not part of this PR

Alasdair: You are seeing them just because this PR was opened before my other PR removing the effects separately was merged.

jrtc27:
> Merge branch 'master' into class_fix

Please rebase rather than merge to keep the history untangled

model/riscv_vlen256.sail
Alasdair: Ideally we would be able to disable V, but the extension is too intertwined into the base model right now to have a sail-riscv model without V, so having a shorter vlenmax is the best we can do.

Alasdair:
> > Merge branch 'master' into class_fix
> 
> Please rebase rather than merge to keep the history untangled

Sorry, I didn't realise the github update branch button did a merge by default. I'll fix it.

bacam:
For what it's worth, I've got development versions of Lem and Sail to handle the model with vectors and machine words, but I think the switch to bitlists by default is worthwhile because it's a little more robust.  However, this does mean that the `vlenmax` change is unnecessary, so I suggest dropping that.

Alasdair:
Ok, the additional vlen file is gone and it seems to build fine.

Alasdair:
I think more people are running into this issue, would be good to get this merged I think

billmcspadden-riscv:
Will do

PR #359, opened 2023-11-22
Fix the encoding and assembly of `vsetvl` instruction

### LINKED ISSUE #337: Encoding and assembly of `vsetvl` are incorrect
In commit c90cf2e6eff5f, new file `model/riscv_insts_vext_vset.sail` was added and includes definitions for instructions `vsetvli` and `vsetvl`. Therein, the encoding for `vsetvl` is define thus:
```
encdec_vsetop(op) @ ma @ ta @ sew @ lmul @ rs1 @ 0b111 @ rd @ 0b1010111
```
where `encdec_vsetop(VSETVL) = 0b1000`

Per my reading of the "RISC-V "V" Vector Extension" specification, Version 1.0, the encoding for `vsetvl` (p. 25), the encoding should look more like:
```
0b1000000 @ rs2 @ rs1 @ 0b111 @ rd @ 0b1010111
```

Similarly, the assembly for `vsetvl` is defined in Sail as:
```
vsettype_mnemonic(op) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ sew_flag(sew) ^ maybe_lmul_flag(lmul) ^ maybe_ta_flag(ta) ^ maybe_ma_flag(ma)
```

But my reading of the spec shows something more like:
```
vsettype_mnemonic(op) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
```

Lastly, the discriminator to mnemonic mapping in Sail is:
```
mapping vsettype_mnemonic : vsetop <-> string ={
  VSETVLI <-> "vsetvli",
  VSETVL  <-> "vsetvli"
}
```

But, the `VSETVL` mnemonic should obviously be `vsetvl`.

XinlaiWan:
The implementation of `vsetvl` and `vsetvli` are rearranged, and some helper functions are added.

This PR fixed #337 .


XinlaiWan:
> Fixing encodings should not be a 178-line diff. If you want to refactor because you think it’s cleaner, fine, but please separate that from the function change that is fixing the encoding.

The issue is because the encoding and assembly definitions of `vsetvl` and `vsetvli` are incorrectly mixed together, but they actually have a lot of difference so that they cannot have a unified definition in the old way and keep their correctness at the same time. So **in order to** fix the encoding issue, I **have to** split the definitions and executions of `vsetvl` and `vsetvli`, which makes a lot of difference in the code. And although their encoding and assembly are very different, the logic in their execution function to calculate AVL and VLMAX and set `vl` are very similar. That's why I further extract some code to be helper functions. 

I don't think the change of refactoring can be separated from fixing the encoding, because the purpose of this refactoring is to fix the encoding issue, and the issue is fixed by refactoring the code of `vsetvl` and `vsetvli`.

XinlaiWan:
If the commit message of "fix the encoding" causes misunderstanding, maybe I can modify that description to be more reasonable.

jrtc27:
So you're not just changing the encoding and assembly. You're also rewriting their implementations to be functionally different to before, regardless of how the instructions were encoded.

Alasdair:
I would adjust the commit message to explain what you wrote in your comment above, as it provides the reasoning for the change. Maybe also add a pointer to https://github.com/riscv/sail-riscv/issues/337?

In general I think we should be proactive in getting vector issues resolved as they are discovered. I actually think the new version with four simpler instructions w/ helper functions is clearer than the original with the branching logic in each function, so this change looks good to me.

jjscheel:
@vmoya-smd, can you review this proposed patch for us?

model/riscv_insts_vext_vset.sail
```diff
@@ -71,60 +71,54 @@ mapping maybe_ma_flag : string <-> bits(1) = {
...
-    /* Note: Implementations can set vill or trap if the vtype setting is not supported.
-     * TODO: configuration support for both solutions
-     */
-    vtype->bits() = 0b1 @ zeros(sizeof(xlen) - 1); /* set vtype.vill */
-    vl = zeros();
-    print_reg("CSR vtype <- " ^ BitStr(vtype.bits()));
-    print_reg("CSR vl <- " ^ BitStr(vl));
-    return RETIRE_SUCCESS
-  };
+  if SEW_pow_new > LMUL_pow_new + ELEN_pow then { handle_illegal_vtype(); return RETIRE_SUCCESS };
```
vmoya-smd, line 114: I know that given operator precende rules this work but I would prefer parenthesis for clarity to make clear what is compared with.

vmoya-smd:
Other than some personal nitpicking on some comparison, where I would prefer to use parenthesis even it not required I think the changes are good.

The new code is more clear and avoid replication of code compared with just a simple fix of the previous implementation so I think the refactoring is appropriate.

allenjbaum:
I agree. Sail was chosen in large part for its readability, and this
improves it, especially for those not familiar with its typing system and
precedence
- and those are exactly the people we should ensure can understand the code.

jjscheel:
@billmcspadden-riscv, what is our plan to complete this work?

model/riscv_insts_vext_vset.sail
XinlaiWan: Parentheses are added to related lines.

XinlaiWan:
I have resolved the conflicts with the base branch and hope this will make code review easier.

PR #361, opened 2023-11-28
Add unratified Zicfilp extension

ved-rivos:
Specification: https://github.com/riscv/riscv-cfi/blob/main/cfi_forward.adoc
Tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/413


jrtc27:
It should not take 15 commits like that to add this extension. Please condense into a sensible history.

ptomsich:
> It should not take 15 commits like that to add this extension. Please condense into a sensible history.

The part that Jessica didn't make explicit is: we use a "rebase and merge" strategy and don't squash your history when merging, so your history should be considered part of the submission.

ved-rivos:
>  It should not take 15 commits like that to add this extension. Please condense into a sensible history.

Sorry, that the commit history was nonsense. I have squashed some of the commits. Let me know if it makes sense now.

jordancarlin:
Merged in #1238.

PR #362, opened 2023-11-28
Constrain TLB struct bit sizes

Timmmm:
Constraining them to <= 64 bits generates code that doesn't use GMP and is therefore much faster. This massively increases C emulation performance.

### DIFF
--- a/model/riscv_vmem_tlb.sail
+++ b/model/riscv_vmem_tlb.sail
@@ -68,9 +68,13 @@
 /*  SUCH DAMAGE.                                                                         */
 /*=======================================================================================*/
 
-/* idealized generic TLB entry to model fence.vm and also speed up simulation. */
+/* Idealized generic TLB entry to model fence.vm and also speed up simulation.
+ * The type size constraints allow Sail to use uint64 which is much faster than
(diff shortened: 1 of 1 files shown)


Alasdair:
LGTM 👍 

Better bitvector length monomorphisation is on my agenda for the Sail -> SystemVerilog model checking path so maybe in the future this will not be needed. That may be a way off yet, so I'm not saying that as a reason to not merge this.

ved-rivos:
@Timmmm should the `asidlen` be 65536 and not 64

Timmmm:
> @Timmmm should the `asidlen` be 16 and not 64

It wasn't really the intent of this change to impose architectural restrictions; just to make Sail generate better C types, and only 64 is necessary for that. I think I'd prefer to leave it as 64 so the intent is clear and this code can be removed if/when Sail uses "small bitvector optimisation".

Timmmm:
(Or more monomorphisation as Alasdair mentioned.)

arichardson:
I think this is no longer needed? If so can we close the pr?

Timmmm:
Yep.

PR #364, opened 2023-12-01
Makefile: Set OPAMCLI to 2.0

Alasdair:
This removes the warning about `opam config var`. Setting OPAMCLI in this way is the correct thing to do if we want to continue supporting opam 2.0.

If we decide to require opam 2.1+ then all `opam var` invocations should become `opam --cli=2.1 var`, as per the opam CLI versioning spec:

https://github.com/ocaml/opam/wiki/Spec-for-opam-CLI-versioning

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ else ifeq ($(ARCH),64)
   override ARCH := RV64
 endif
 
+# Set OPAMCLI to 2.0 to supress warnings about opam config var
+export OPAMCLI := 2.0
+
 ifeq ($(ARCH),RV32)
   SAIL_XLEN := riscv_xlen32.sail
 else ifeq ($(ARCH),RV64)


Timmmm (approved):

billmcspadden-riscv (approved):
lgtm

Alasdair:
Hmm, is anyone else still seeing these warnings? I'm thinking I maybe tricked myself into thinking this fixed it somehow. I think maybe you actually need to set it in every subshell like:

`$(shell OPAMCLI=2.0 opam config var ...)`

to get rid of all the warnings

bacam:
I'm still getting them.  Experimenting a little, it appears that the `shell` function isn't affected by `export`, at least in GNU make 4.3.  It's probably worth keeping the assignment, though, and using `$(shell OPAMCLI=$(OPAMCLI) ...)` so that there's a consistent setting everywhere, including inside scripts.

On a related note, we should use `:=` for all the uses of `shell` to ensure that they're only run once.  That speeds up all make invocations.

PR #365, opened 2023-12-01
Replace old virtual memory code with new (details follow)

rsnikhil:
Old code had much 'cut-and-paste' replication for RV32 (Sv32) and
  RV64 (Sv39, Sv48), and was scattered over several files.

  New code unifies them into single set of parameterized functions
  that works for RV32/RV64 and Sv32/Sv39/Sv48 (and is ready for Sv57).

  Deleted old files:
    riscv_vmem_common.sail    riscv_pte.sail          riscv_ptw.sail
    riscv_vmem_rv32.sail      riscv_vmem_rv64.sail
    riscv_vmem_sv32.sail      riscv_vmem_sv39.sail    riscv_vmem_sv48.sail

  Added 1 new file, where the main vmem code sits:
    riscv_vmem.sail

  Modified (to be compatible with new code, and to isolate TLB stuff
  into one file, since TLBs are not part of RISC-V Architecture Spec)

    riscv_vmem_tlb.sail

  Added documentation on new vmem code: doc/notes_Virtual_Memory.adoc
    Deleted older vmem files.

### DIFF
--- a/model/riscv_csr_map.sail
+++ b/model/riscv_csr_map.sail
@@ -51,6 +51,8 @@ mapping clause csr_name_map = 0x143  <-> "stval"
 mapping clause csr_name_map = 0x144  <-> "sip"
 /* supervisor protection and translation */
 mapping clause csr_name_map = 0x180  <-> "satp"
+/* supervisor envcfg */
+mapping clause csr_name_map = 0x10A  <-> "senvcfg"
 /* machine information registers */
 mapping clause csr_name_map = 0xF11  <-> "mvendorid"
 mapping clause csr_name_map = 0xF12  <-> "marchid"
@@ -65,6 +67,8 @@ mapping clause csr_name_map = 0x304  <-> "mie"
 mapping clause csr_name_map = 0x305  <-> "mtvec"
 mapping clause csr_name_map = 0x306  <-> "mcounteren"
 mapping clause csr_name_map = 0x320  <-> "mcountinhibit"
+/* machine envcfg */
+mapping clause csr_name_map = 0x30A  <-> "menvcfg"
 /* machine trap handling */
 mapping clause csr_name_map = 0x340  <-> "mscratch"
 mapping clause csr_name_map = 0x341  <-> "mepc"


Makefile
```diff
@@ -68,15 +68,20 @@ SAIL_SYS_SRCS += riscv_softfloat_interface.sail riscv_fdext_regs.sail riscv_fdex
...
-SAIL_RV32_VM_SRCS = riscv_vmem_sv32.sail riscv_vmem_rv32.sail
-SAIL_RV64_VM_SRCS = riscv_vmem_sv39.sail riscv_vmem_sv48.sail riscv_vmem_rv64.sail
-
-SAIL_VM_SRCS = riscv_pte.sail riscv_ptw.sail riscv_vmem_common.sail riscv_vmem_tlb.sail
-ifeq ($(ARCH),RV32)
-SAIL_VM_SRCS += $(SAIL_RV32_VM_SRCS)
-else
-SAIL_VM_SRCS += $(SAIL_RV64_VM_SRCS)
-endif
+# Virtual Memory support: (2023-11) OLD VM CODE replaced by NEW VM CODE
```
arichardson, line 71: I think these comments should be deleted, if anyone is interested in the history they can use git to find the previous versions?

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+// Parameters for VM modes sv32, sv39 and sv48
+
+// All VM modes use the same page size (4KB, with 12-bit index)
+
+let PAGESIZE_BITS : nat = 12
+
+// PRIVATE
+struct SV_Params = {
+       // SATP CSR		     // Sv32  Sv39  Sv48
+       satp_asid_size_b	   : nat,    //	   9	16    16
```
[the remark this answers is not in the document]
rsnikhil: Ok. I had noticed that compiler error when I first tried it, but with this workaround (‘undefined_…’), it works.
I’ve updated the code to use this idiom.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+
+       // PTE			     // Sv32  Sv39  Sv48
+       levels		   : nat,    //	   2	 3     4
+       log_pte_size_B	   : nat,    //	   2	 3     3
+       pte_msbs_lsb_index  : nat,    //	  32	54    54
+       pte_msbs_size_b	   : nat,    //	   0	10    10
+       pte_PPNs_lsb_index  : nat,    //	  10	10    10
+       pte_PPNs_size_b	   : nat,    //	  22	44    44
+       pte_PPN_j_size_b	   : nat     //	  10	 9     9
+}
```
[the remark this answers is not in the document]
rsnikhil: Ok, agreed.  I’ve made these changes.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+// The TLB implementation is in a separate file: riscv_vmem_tlb.sail
+// The code in this file is structured and commented so you can easily
+// ignore TLB functionality at first reading.
+
+// ****************************************************************
+// Useful 64-bit numeric constants
+
+// PRIVATE
+let zero_64b : bits(64) = zero_extend(0b0)
+let one_64b  : bits(64) = zero_extend(0b1)
```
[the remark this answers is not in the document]
rsnikhil: Ok, agreed; I have made the changes and removed those two global defs.

There was one place where ‘one_64b' was needed for type inference.  I’ve defined it adjacent to that need as a local def.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+  // Read this-level PTE from mem
+  let mem_result = mem_read_priv(Read(Data),		  // AccessType
+				 Supervisor,		  // Privilege
+				 pte_phys_addr,
+				 8,			  // atom (8)
+				 false,			  // aq
+				 false,			  // rl
+				 false);		  // res
+
+  let ptw_result : PTW_Result =
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+			 a	   : AccessType(ext_access_type),
+			 ext	   : extPte)
+			-> option(bits(64)) = {
+  // Update 'dirty' bit?
+  let update_d = (D_of_PTE (pte) == bitzero)
+		 & (match a {
+		      Execute() => false,
+		      Read() => false,
+		      Write(_) => true,
+		      ReadWrite(_,_) => true
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+	  TR_Address(pa, ext_ptw) => TR_Address(truncate(pa, sizeof(xlen)), ext_ptw),
+	  TR_Failure(f, ext_ptw)  => TR_Failure(translationException(ac, f), ext_ptw)
+	}
+      }
+    };
+  tr_result
+}
+
+// ****************************************************************
+// Initialize Virtul Memory state
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

model/riscv_vmem_tlb.sail
```diff
@@ -1,124 +1,198 @@
...
+  vMatchMask : bits(64),  // matching mask for superpages
+  vAddrMask  : bits(64),  // selection mask for superpages
+  pte	     : bits(64),  // PTE
+  pteAddr    : bits(64),  // for dirty writeback
+  age	     : bits(64)	  // for replacement policy?
 }
 
+// Single-entry TLB (could enlarge this in future for better simulation speed)
+// PRIVATE
+register tlb : option(TLB_Entry)
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+// PUBLIC: see also riscv_insts_zicsr.sail and other CSR-related files
+register satp : xlenbits
+
+// See riscv_sys_regs.sail for legalize_satp{32,64}().
+// WARNING: those functions legalize Mode but not ASID?
+// PUBLIC: invoked from writeCSR() to fixup WARL fields
+function legalize_satp(a : Architecture,
+		       o : xlenbits,	    // previous value of satp
+		       v : xlenbits)	    // proposed new value of satp
+		      -> xlenbits =	    // new legal value of satp
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+	};
+      if not(valid_va) then
+	TR_Failure(translationException(ac, PTW_Invalid_Addr()), ext_ptw)
+      else {
+	let mxr	   : bool     = mstatus.MXR() == 0b1;
+	let do_sum : bool     = mstatus.SUM() == 0b1;
+	let asid   : bits(16) = satp_to_asid(sv_params, satp);
+	let ptb	   : bits(64) = satp_to_PT_base(sv_params, satp);
+	let init_level = sv_params.levels - 1;
+	assert(init_level >= 0);    // allows init_level to be passed as 'nat' type
```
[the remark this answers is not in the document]
rsnikhil: Indeed, the ‘assert()’ is no longer needed; I’ve deleted it.  I’ve also inlined ‘sv_params.levels - 1’ and removed the ‘let init_level = ..’ line.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+
+// Update PTE bits if needed; return new PTE if updated
+// PRIVATE
+function update_PTE_Bits(sv_params : SV_Params,
+			 pte	   : bits(64),
+			 a	   : AccessType(ext_access_type),
+			 ext	   : extPte)
+			-> option(bits(64)) = {
+  // Update 'dirty' bit?
+  let update_d = (D_of_PTE (pte) == bitzero)
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

(This is contrary to my preference and practice—I prefer _always_ having a space before ‘(‘, but I’m deferring to public convention throughout this code.)

model/riscv_vmem.sail
```diff
@@ -0,0 +1,871 @@
...
+  let mode	: SATPMode = translationMode(effPriv);
+  // PTW extensions (non-standard): initialize the PTW extension state
+  let ext_ptw	: ext_ptw = init_ext_ptw;
+
+  let tr_result : TR_Result(xlenbits, ExceptionType) =
+    if mode == Sbare then
+      TR_Address(vAddr, ext_ptw)
+    else {
+      let (valid_va, sv_params) : (bool, SV_Params)
+      = match (mode) {
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

model/riscv_vmem.sail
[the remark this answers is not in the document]
arichardson: Could you please upload the new version before marking comments as resolved?

model/riscv_vmem_tlb.sail
```diff
@@ -1,124 +1,198 @@
...
+// PUBLIC: invoked in init_vmem() [riscv_vmem.sail]
+function init_TLB() -> unit =
+  tlb = None()
+
+// PUBLIC: invoked in translate_TLB_hit() [riscv_vmem.sail]
+function write_TLB(idx : nat, ent : TLB_Entry) -> unit =
+  tlb = Some(ent)
+
+// PRIVATE
+function make_TLB_Entry(asid	      : bits(16),
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

rsnikhil:
My apologies.  I've re-marked it 'unresolved'.
I also don't know who is supposed to mark it 'resolved' (me? or the
commenter?).

[ I'm new to multi-person/multi-org projects on GitHub, and am not
  familiar with the flows and protocols. I was marking each comment
  resolved as I dealt with it in my local clone, re-testing at each
  step.  I had not yet pushed it to my GitHub fork.]

I've just pushed a new commit (225b9cf) with my changes, to my fork
(https://github.com/rsnikhil/sail-riscv). This started a CI run on my
fork, which reported success.

In the PR in upstream repo (https://github.com/riscv/sail-riscv),
it reports this commit, and reran a CI, which reports success.


Nikhil

model/riscv_vmem.sail
```diff
@@ -0,0 +1,857 @@
...
+
+// Valid xlen-wide values containing virtual addrs must have upper
+// bits equal to the MSB of the virtual address.
+// Virtual address widths depend on the virtual memory mode.
+
+// PRIVATE
+function isValidVAddr(sv_params : SV_Params, vAddr : bits(64)) -> bool = {
+  let index = sv_params.va_size_bits - 1;
+  assert((0 <= index) & (index < 64));
+  let upper_bits = vAddr >> (sv_params.va_size_bits);
```
arichardson, line 257: Not sure if it's any better, but if you use a slice here instead of a shift you can drop the shift from the == ones() case.

rsnikhil: I tried the following using a slice, but it does not type-check:

    function isValidVAddr(sv_params : SV_Params, vAddr : bits(64)) -> bool = {
      let index = sv_params.va_size_bits - 1;
      assert((0 <= index) & (index < 64));
      let upper_bits = vAddr [63 .. sv_params.va_size_bits];
      let va_msb	 = vAddr[index];
      if va_msb == bitzero then
        upper_bits == zeros()
      else
        upper_bits == ones()
    }

I think the problem is that the slice does not have a size known statically to the type-checker; the size depends on the dynamic value sv_params.va_size_bits.  My technique avoids this by making everything 64 bits; the “dynamic” part goes into the shift-amount, which is ok, and does not affect type-checking.

Please let me know if my reasoning is ok, or if there is some other way to do this with a slice?

Alasdair: I think the type system isn't quite clever enough to see that `sv_params.va_size_bits` is constant. It might work if you did something like:
```
function isValidVAddr(sv_params : SV_Params, vAddr : bits(64)) -> bool = {
  let index = sv_params.va_size_bits - 1;
  assert((0 <= index) & (index < 64));
  let upper_bits = vAddr [63 .. index + 1];
  let va_msb	 = vAddr[index];
  if va_msb == bitzero then
    upper_bits == zeros()
  else
    upper_bits == ones()
}
```
but not sure if that is clearer

Alasdair: You could also pattern match on the struct, something like:
```
function isValidVAddr(struct { va_size_bits, _ }: SV_Params, vAddr : bits(64)) -> bool = {
  let index = va_size_bits - 1;
  assert((0 <= index) & (index < 64));
  let upper_bits = vAddr [63 .. va_size_bits];
  let va_msb	 = vAddr[index];
  if va_msb == bitzero then
    upper_bits == zeros()
  else
    upper_bits == ones()
}
```

rsnikhil: I don’t uderstand these last two suggestions.  va_size_bits can be 32, 39, 48 (and in future, 57), i.e., it’s still not a unique static value.  Is the type-checker smart enough to know (a) that it typechecks for each of those cases and (b) that zeros() and ones() needs to extended by a different amount for each of those cases?

rsnikhil: I also have a question about the struct pattern-match syntax in your example:

    struct { va_size_bits, _ }

I would have expected something like this:

    struct { va_size_bits = v, _ }

where 'va_size_bits' is the field name and 'v' is the local identifier bound to its value.  Is your syntax equivalent to:

    struct { va_size_bits = va_size_bits, _ }

where the field name is also used as the bound identifier?

Alasdair: The type system keeps track of the numeric value of arguments at the level of individual variables, so if you bind `va_size_bits` as a variable once as in my last example with a struct pattern, it'll know it is either 32, 39, or 48, and importantly it'll know it's the same between the two usages, and it'll correctly infer the implicit arguments for zeros and ones so they end up the right size. The type of `upper_bits` is dependent on the value of `va_size_bits`, and it'll also know that `index` is whatever `va_size_bits` is minus 1, so the assert is actually not needed, and it can prove all the indexing is in bounds. I had a go on my machine and the following checks:

```
function isValidVAddr(struct { va_size_bits, _ }: SV_Params, vAddr : bits(64)) -> bool = {
  let index      = va_size_bits - 1;
  let upper_bits = vAddr[63 .. va_size_bits];
  let va_msb	 = vAddr[index];
  if va_msb == bitzero then
    upper_bits == zeros()
  else
    upper_bits == ones()
}
```

Since we only keep track of integer constraints for variables, in the code you wrote originally it doesn't know that `sv_params.va_size_bits` is the same for each of the two usages in the function, as that would require tracking constraints at a finer field-level granularity (which maybe we could do in the future, and Rust for example already does this type of field-level granularity checking for lifetimes).

Alasdair: > where 'va_size_bits' is the field name and 'v' is the local identifier bound to its value. Is your syntax equivalent to:
> 
> ```
> struct { va_size_bits = va_size_bits, _ }
> ```
> 
> where the field name is also used as the bound identifier?

Yes exactly, a struct pattern `struct { foo, _ }` is shorthand for `struct { foo = foo }`.

rsnikhil: Ok, understood. Very cool. I’ve incorporated your suggested version.

Timmmm (changes requested):
Sail really needs module support!

model/riscv_vmem.sail
```diff
@@ -0,0 +1,855 @@
...
+       pte_PPNs_lsb_index  = 10,
+       pte_PPNs_size_bits  = 44,
+       pte_PPN_j_size_bits = 9
+}
+*/
+
+// PRIVATE
+val      undefined_SV_Params : unit -> SV_Params
+
+function undefined_SV_Params() = sv32_params
```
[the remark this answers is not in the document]
Alasdair: Currently you need it to work around an issue where Sail tries to figure out how it could do `let x : SV_Params = undefined` even though the code never does this 😞 

I've fixed Sail so it's stricter about what can be undefined in https://github.com/rems-project/sail/commit/ff2647a35c9e65df38d7e8981d54bb64de443c95, but would need a new release before we could rely on that.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,855 @@
...
+function offset_of_va(va : bits(64)) -> bits(64) = {
+    let mask : bits(64) = (zero_extend(0b1) << PAGESIZE_BITS) - 1;
+    va & mask
+}
+
+// Valid xlen-wide values containing virtual addrs must have upper
+// bits equal to the MSB of the virtual address.
+// Virtual address widths depend on the virtual memory mode.
+// PRIVATE
+function is_valid_vAddr(struct { va_size_bits, _ } : SV_Params,
```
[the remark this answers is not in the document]
rsnikhil: Ok, thanks for the simpler way.

Re. “should it be xlenbits”: no, the current strategy for having a single piece of code working for Sv32, Sv39, Sv48 and Sv57 is to do all vmem work at 64 bits, parameterized by SV_Params for each specific case.  Note that even when in a system where xlen==64 (for M-mode), S-mode (SXL) and U-mode (UXL) may be 32 bit.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,855 @@
...
+
+function undefined_SV_Params() = sv32_params
+
+// ****************************************************************
+// Fields of virtual addresses
+
+// PRIVATE: Extract full VPN field from VA
+function vpns_of_va(sv_params : SV_Params,
+                    va        : bits(64)) -> bits(64) = {
+    let mask : bits(64) = (zero_extend(0b1) << sv_params.va_size_bits) - zero_extend(0b1);
```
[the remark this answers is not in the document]
rsnikhil: Ok, thanks for the suggestion.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,855 @@
...
+}
+
+// PRIVATE: extract PPNs of PTE
+function PPNs_of_PTE(sv_params : SV_Params, pte : bits(64)) -> bits(64) = {
+  let mask : bits(64) = (zero_extend(0b1) << sv_params.pte_PPNs_size_bits) - zero_extend(0b1);
+  (pte >> sv_params.pte_PPNs_lsb_index) & mask
+}
+
+// PRIVATE: extract LSBs (including permissions) of PTE
+function D_of_PTE(pte : bits(64)) -> bit = pte[7]
```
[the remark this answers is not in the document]
rsnikhil: The original code had three separate bitfields for PTEs for Sv32, Sv39 and Sv48, much of the PTEs are different for each one.  To unify these (since bitfields are not allowed to be polymorphic at the moment), I resorted to working at 64bits throughout and using selector functions, and using SV_Params to parameterize the differences.

Of course, the bottom 8 bits of PTEs are the same for Sv32, Sv39 and Sv48, and one could use a common bitfield for that part of the PTE, but I’m not sure it would clarify things much.

model/riscv_vmem_tlb.sail
```diff
@@ -1,124 +1,175 @@
...
-/*  PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR                  */
-/*  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,                         */
-/*  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT                     */
-/*  LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF                     */
-/*  USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND                  */
-/*  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,                   */
-/*  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT                   */
-/*  OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF                   */
-/*  SUCH DAMAGE.                                                                         */
+/*  RISCV Sail Model									 */
```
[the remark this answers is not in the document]
rsnikhil: I don’t think I understand the display above.  The above cited excerpt looks like this:

    /*  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT			 */
    /*  OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF			 */
    /*  SUCH DAMAGE.									 */

    // Although a TLB is not part of the RISC-V Architecture
    // specification, we model a simple TLB so that

The “RISCV Sail Model” line is on line 2, not after line 68.

The whitespace looks ok on my Emacs screen.

In VSCode, this whole header section initially looks messed up, because by default it has a “Spaces: 2” setting. On the bottom line, where it says “Spaces: 2”, this needs to be changed to “Indent using spaces” and setting the tab size to 8, to match the way Emacs displays it.

model/riscv_vmem.sail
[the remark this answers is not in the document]
rsnikhil: I tried your suggested version of is_valid_vAddr().  It does not typecheck.

    Type error:
    model/riscv_vmem.sail:254.3-53:
    254 |   sign_extend(vAddr[va_size_bits - 1 .. 0]) == vAddr
        |   ^------------------------------------------------^
        | No overloading for (operator ==), tried:
        | * eq_int
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))
        | * eq_bit
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))
        | * eq_bool
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))
        | * eq_unit
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))
        | * eq_string
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))
        | * eq_bit
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))
        | * eq_bits
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))
        | * eq_string
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))
        | * eq_anything
        |    Cannot solve implicit 'm in sign_extend(subrange_bits(vAddr, sub_atom(va_size_bits, 1), 0))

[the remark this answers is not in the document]
Alasdair: Could do `sign_extend(64, vAddr[va_size_bits - 1 .. 0]) == vAddr`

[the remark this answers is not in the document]
Alasdair: `vAddr == sign_extend(vAddr[va_size_bits - 1 .. 0])` also works, as the type checker handles the arguments to `==` from left to right, and `vAddr` has the simpler type which then makes the sign_extend length obvious. I'm not sure I like that behavior but I don't know if there's any way to fix it.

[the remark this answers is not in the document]
rsnikhil: Ok, yes, this works.  Wow! I would not have suspected there’d be any difference just swapping the arg order.

I’ve made the change (in my clone; will push after some more consolidation).

model/riscv_vmem.sail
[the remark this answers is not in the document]
arichardson: I think using a bitfield for the common bits would be slightly easier to read than these wrapper functions.

model/riscv_vmem.sail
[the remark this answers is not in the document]
rsnikhil: I’ve incorporated this successfully for all the (many) mask computations.

But there are two places where typechecking fails, both involving PAGESIZE_BITS : nat.
One place is the following: Both versions 2 and 3 fail on typecheck ("Could not resolve quantifiers for zero_extend”)

    function offset_of_va(va : bits(64)) -> bits(64) = {
        let mask : bits(64) = (zero_extend(0b1) << PAGESIZE_BITS) - 1;    // 1: original code; typecheck succeeds
        // let mask : bits(64) = zero_extend(ones(PAGESIZE_BITS));           // 2: typecheck fails
        // let mask : bits(64) = zero_extend(64, ones(PAGESIZE_BITS));    // 3: typecheck fails
        va & mask
    }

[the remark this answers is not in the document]
jrtc27: Presumably if you constrain PAGESIZE_BITS's type to not be a nat then it'll be happy, because it's (rightly) upset that you can't zero_extend e.g. a 12356-bit number to 64 bits.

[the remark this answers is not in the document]
rsnikhil: Ah, I see (even though PAGESIZE_BIT is an immutable constant with value 12?).
I changed the declaration of PAGESIZE_BITS from:

    let PAGESIZE_BITS : nat = 12

to this:

    let PAGESIZE_BITS : {12} = 12

and it fixed the problem.
Thanks!

model/riscv_vmem.sail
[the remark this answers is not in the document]
Alasdair: In this case I think it would be nice if the order didn't matter. I think what the type-checker would need to do would be to try to infer the first argument, then upon noticing it can't, continue checking the second argument. The second argument would then give us more information that would then 'invalidate' the failed check of the first argument, so we would re-check it. I'm not sure I'm quite up to implementing that right now though!

model/riscv_vmem.sail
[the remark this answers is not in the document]
jrtc27: A pattern we use in sail-cheri-riscv for this kind of thing is:

```
type PAGESIZE_BITS : Int = 12
let  PAGESIZE_BITS = sizeof(PAGESIZE_BITS)
```

which avoids duplicating the value but instead is rather more longwinded. Perhaps there is (or could be invented) some nice syntax for this kind of case.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,849 @@
...
+function satp_to_PT_base(sv_params : SV_Params, satp_val : xlenbits) -> bits(64) = {
+  // This extend op is a no-op when xlen==64, extends when xlen==32
+  let satp_64b : bits(64) = zero_extend(satp_val);
+  let mask_64b : bits(64) = zero_extend(ones(sv_params.satp_ppn_size_bits));
+  let ppn_64b  : bits(64) = (satp_64b >>  sv_params.satp_ppn_lsb_index) & mask_64b;
+  ppn_64b << PAGESIZE_BITS
+}
+
+// Compute address translation mode from SATP
+// TODO: shouldn't we look at mstatus_UXL if priv is User?
```
defermelowie, line 608: Since the format of `satp` is dependend on SXLEN and not UXLEN, why would this be?
Furthermore, the spec. explicitly states that "When SXLEN=32, the only valid setting for MODE is Sv32" and " When SXLEN=64, three paged virtual-memory schemes are defined: Sv39, Sv48, Sv57".

model/riscv_vmem_tlb.sail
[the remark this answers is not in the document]
Timmmm: I think some spaces have been replaced with hard tabs.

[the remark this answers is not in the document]
Timmmm: PS: I highly recommend setting `Render Whitespace` to `boundary` in VSCode.

image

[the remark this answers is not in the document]
rsnikhil: > I think some spaces have been replaced with hard tabs.

Is that a problem?  CODE_STYLE.md is silent on any policy regarding tabs vs. spaces.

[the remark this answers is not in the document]
rsnikhil: > PS: I highly recommend setting `Render Whitespace` to `boundary` in VSCode.

I rely on Emacs’ ’tabify’ and ‘untabify’ functions to do tabs/spaces properly for a selected region or buffer.

If Emacs and VSCode (and possibly vi) do it differently, I suggest not relying either one as a standard, as long as each tools is able to display it properly.

model/riscv_vmem.sail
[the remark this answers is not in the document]
rsnikhil: Ok, I’ll defer to the majority here and do the common 8 LSBs with a bitfield, although IMHO while bitfields are a nice construct, I don’t think it  improves readability in this particular situation, especially since other fields in the PTE are not done this way.

model/riscv_vmem.sail
rsnikhil: Ok, thanks for the clarification.  I’ll remove the TODO in the comment.

model/riscv_vmem.sail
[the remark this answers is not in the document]
rsnikhil: Ok, I’ve added a comment for this.

Alasdair:
It looks like the current code has tabs as well as spaces, we should make sure all the tabs are gone before we merge.

I thought we had a pre-commit hook to check for no tabs, but it looks like we don't.

rsnikhil:
> It looks like the current code has tabs as well as spaces, we should make sure all the tabs are gone before we merge.
> 
> I thought we had a pre-commit hook to check for no tabs, but it looks like we don't.

Yes it has tabs; I was unaware that no-tabs is the recommended style; CODE_STYLE.md is silent on this (perhaps someone should add this recommendation to CODE_STYLE.md?).

I actually applied “tabify()” in Emacs to convert spaces to tabs, because that is my normal practice in my own projects.

I will apply “untabify()” in Emacs to reverse that.

jrtc27:
CODE_STYLE.md clearly states a two space indent, because that was (and still is) the predominant style in the model.

rsnikhil:
> CODE_STYLE.md clearly states a two space indent, because that was (and still is) the predominant style in the model.

Yes, I read that, and am aware of that, but it does not rule out using tabs to reach the context indent before applying two more spaces.  And there were tabs in other long white spaces (e.g., in the license text).

In any case: I’ve applied ‘untabify()’ to the wholef file to remove all tabs.

jrtc27:
> > CODE_STYLE.md clearly states a two space indent, because that was (and still is) the predominant style in the model.
> 
> Yes, I read that, and am aware of that, but it does not rule out using tabs to reach the context indent before applying two more spaces. And there were tabs in other long white spaces (e.g., in the license text).
> 
> In any case: I’ve applied ‘untabify()’ to the wholef file to remove all tabs.

Tabs are not spaces, and don’t even have a fixed column width. If you convert spaces to tabs then you don’t have a 2 space per block indent, you have N tabs and M spaces.

Alasdair:
We have too much code that tries to align things in a pretty way to use tabs, as it'll just look broken in anyone else's editor if they don't have the same tab width.

The error message formatting code in Sail also just assumes a tab width of 4, so error messages will look off to anyone who has set their tab width differently.

rsnikhil:
> > > CODE_STYLE.md clearly states a two space indent, because that was (and still is) the predominant style in the model.
> > 
> > 
> > Yes, I read that, and am aware of that, but it does not rule out using tabs to reach the context indent before applying two more spaces. And there were tabs in other long white spaces (e.g., in the license text).
> > In any case: I’ve applied ‘untabify()’ to the wholef file to remove all tabs.
> 
> Tabs are not spaces, and don’t even have a fixed column width. If you convert spaces to tabs then you don’t have a 2 space per block indent, you have N tabs and M spaces.

Ok, that’s not how I (and possibly others) interpret the language.  IMO, a phrase “don’t use tabs at all” would clarify it.  It’s a moot point anyway, now, for the files in question, since I’ve converted everything to spaces.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+  let mask : bits(64) = zero_extend(ones(sv_params.va_size_bits));
+  (va & mask) >> PAGESIZE_BITS
+}
+
+// PRIVATE: Extract VPN[level] from  VA
+function vpn_j_of_va(sv_params : SV_Params,
+		     va	       : bits(64),
+		     level     : nat)	    -> bits(64) = {
+  let lsb  : nat       = PAGESIZE_BITS + level * sv_params.vpn_size_bits;
+  let mask : bits(64) = zero_extend(ones(sv_params.vpn_size_bits));
```
Timmmm, line 248: No need to pass the whole `sv_params` struct if you only use one field.

Also this whole approach does seem to give up on Sail's fancy type system here when it seems like this is exactly the place that would benefit from it.

Was it not possible to make some of this use type level parameters? E.g.

```
val vpn_j_of_va : forall 'vpn_bits, 'vpn_bits in {10, 9}. (int('vpn_bits), bits(64), ptw_level) -> bits('vpn_bits)
function vpn_j_of_va(vpn_size_bits, va, level) = {
  let lsb = PAGESIZE_BITS + level * vpn_size_bits;
  let msb = lsb + vpn_size_bits - 1;
  va[msb .. lsb]
}
```

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+  ((va >> lsb) & mask)
+}
+
+// PRIVATE: Extract offset within page from VA
+function offset_of_va(va : bits(64)) -> bits(64) = {
```
Timmmm, line 253:
```
function offset_of_va(va : bits(64)) -> bits(PAGESIZE_BITS) = va[PAGESIZE_BITS - 1 .. 0]
```

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+  G : 5,    // global
+  U : 4,    // User
+  X : 3,    // Execute permission
+  W : 2,    // Write permission
+  R : 1,    // Read permission
+  V : 0	    // Valid
+}
+
+// PRIVATE: check if a PTE is a pointer to next level (non-leaf)
+function pte_is_ptr(pte_lsbs : PTE_LSBs) -> bool = (  pte_lsbs.X() == 0b0)
```
Timmmm, line 305: Extra spaces

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+			      mxr      : bool,
+			      do_sum   : bool,
+			      pte_lsbs : PTE_LSBs,
+			      ext      : extPte,
+			      ext_ptw  : ext_ptw) -> PTE_Check = {
+  let pte_U = pte_lsbs.U();
+  let pte_R = pte_lsbs.R();
+  let pte_W = pte_lsbs.W();
+  let pte_X = pte_lsbs.X();
+  let b : bool =
```
Timmmm, line 338: `b` is not a very descriptive name. How about `success`?

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+}
+
+// PUBLIC
+overload to_str = {ptw_error_to_str}
+
+// hook for (non-standard) extensions to customize errors reported by page-table
+// walks during address translation; it typically works in conjunction
+// with any customization to check_PTE_permission().
+
+// PRIVATE
```
Timmmm, line 431: Btw I wonder if we should adopt the "leading underscore = private" convention, given that Sail doesn't support public/private natively yet.

Maybe after modules are added.

(No changes needed here; just a thought).

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+		 mxr,
+		 do_sum,
+		 pt_base,
+		 level,
+		 global,
+		 ext_ptw) = {
+  let vpn_j	 = vpn_j_of_va(sv_params, va, level);
+  let pte_offset = vpn_j << sv_params.log_pte_size_bytes;
+  let pte_addr	 = pt_base + pte_offset;
+  let pte_phys_addr : xlenbits = pte_addr[(sizeof(xlen) - 1) .. 0];
```
Timmmm, line 489: Is this right? Physical addresses are 34 bits with SV32.

Maybe a bug in `mem_read_priv()` I guess. Worth adding a TODO comment at least.

Timmmm (changes requested):
Some more review comments. Please can you fix the formatting? Then I will do some more reviewing.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+// ignore TLB functionality at first reading.
+
+// ****************************************************************
+// Parameters for VM modes sv32, sv39, sv48 and (future) Sv57
+
+// All VM modes use the same page size (4KB, with 12-bit index)
+
+// This two-line idiom constrains the value (2nd line) to be a singleton,
+// which helps in type-checking wherever it is used.
+type PAGESIZE_BITS : Int = 12
```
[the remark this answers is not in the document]
rsnikhil: > The rest of the code uses lowercase for these. Also I think it's proabbly unwise to use the same name for both of these.

* I’m not sure what you mean by “rest of the code uses lowercase for these”.  grep shows no other use of PAGESIZE_BITS, upper or lower case, in any file other than my new files (riscv_vmem.sail, riscv_vmem_tlb.sail) where they are all uppercase.  Are you perhaps referring to code in some other repo or branch (CHERI)?

* I was following Jessica’s suggestion exactly: https://github.com/riscv/sail-riscv/pull/365#discussion_r1416191225 which has both the type and the value in upper case.

But, in any case, I take your point, and I will change the value-use to lowercase.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+       // SATP CSR
+       satp_asid_size_bits = 9,
+       satp_asid_lsb_index = 22,
+
+       // SATP PPN
+       satp_ppn_size_bits  = 22,
+       satp_ppn_lsb_index  = 0,
+
+       // VA
+       va_size_bits	   = 32,
```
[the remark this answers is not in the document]
rsnikhil: I think this was just a tabs artefact.  I’ve removed all tabs.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+// PRIVATE
+val	 undefined_SV_Params : unit -> SV_Params
+function undefined_SV_Params() = sv32_params
+
+// ****************************************************************
+// Fields of virtual addresses
+
+// PRIVATE: Extract full VPN field from VA
+function vpns_of_va(sv_params : SV_Params,
+		    va	      : bits(64)) -> bits(64) = {
```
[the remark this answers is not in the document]
rsnikhil: I’re removed all tabs, for spaces.

model/riscv_vmem.sail
rsnikhil: Ok (shrinking output type from bits(64) to bits[PAGESIZE_BITS] needed a compensating zero_extend() in the two places where it was invoked).

model/riscv_vmem.sail
rsnikhil: Ok; fixed.

model/riscv_vmem.sail
rsnikhil: Ok; fixed.

model/riscv_vmem.sail
rsnikhil: Let’s wait for modules, which I believe will be available soon.

model/riscv_vmem.sail
rsnikhil: Yes, Sv32 phys mem addrs are 34 bits, not 32 bits.  But it is used in 'mem_read_priv()' [riscv_mem.sail, L180], where the address argument is declared as 'xlenbits', so until mem_read_priv() is changed, we have to use 32 bits.

I don't know if it's straightforward to fix 'mem_read_priv()'.  As I understand it, in RV32, phys mem addrs are 32(34) bits when Sv32 is not (is) supported.  Even if Sv32 is supported, what happens when we run with M privilege (are phys addrs 32 bits? Or extended to 34 bits)?

I've mentioned this several times in our meetings (perhaps I should add an "issue" so it doesn't slip thru the cracks), and I'll also add a local TODO comment here in the vmem code.

model/riscv_vmem.sail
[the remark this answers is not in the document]
rsnikhil: I’ve retained uppercase for the type def and changed the value def to lowercase (Haskell-like).  Perhaps we should adopt a uniform convention across the whole repo.

model/riscv_vmem.sail
allenjbaum: Physical addresses at 34 bits if Sv32 is implemented, but Mmode cannot access beyond 32 bits 
**unless** it sets (mstatus.MPRV==1) && (mstatus.MPP !=mmode) && (satp.mode!=0).
If those are set correctly, then it will use VM translation and use all 34 bits. 
Otherwise, the upper 2 bits can't be used in Mmode, and they're set to zero.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+
+// This two-line idiom constrains the value (2nd line) to be a singleton,
+// which helps in type-checking wherever it is used.
+type PAGESIZE_BITS : Int = 12
+let  PAGESIZE_BITS = sizeof(PAGESIZE_BITS)
+
+// PRIVATE
+struct SV_Params = {
+  // SATP CSR				 // Sv32  Sv39	Sv48
+  satp_asid_size_bits : { 9, 16},	 //    9    16	  16
```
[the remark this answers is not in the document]
rsnikhil: I don’t think it’s worth it. They are not (currently) used as parameters to any user-defined functions. They’re used inside functions as parameters to primitives like shifts, extends, etc.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,851 @@
...
+// PRIVATE
+function translate(sv_params : SV_Params,
+		   asid	     : bits(16),
+		   ptb	      : bits(64),
+		   vAddr_arg : bits(64),
+		   ac	      : AccessType(ext_access_type),
+		   priv	     : Privilege,
+		   mxr	      : bool,
+		   do_sum    : bool,
+		   level     : nat,
```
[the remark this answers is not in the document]
rsnikhil: Not using ‘nat’ any more, just ranges.

model/riscv_vmem.sail
rsnikhil: Re. "No need to pass the whole sv_params struct if you only use one field.”: in this instance I think it’s better this way. ‘vpn_size_bits’ should only ever come from this struct, and not any other random expression, and this style documents that provenance.

rsnikhil: > No need to pass the whole `sv_params` struct if you only use one field.
> 
> Also this whole approach does seem to give up on Sail's fancy type system here when it seems like this is exactly the place that would benefit from it.
> 
> Was it not possible to make some of this use type level parameters? E.g.
> 
> ```
> val vpn_j_of_va : forall 'vpn_bits, 'vpn_bits in {10, 9}. (int('vpn_bits), bits(64), ptw_level) -> bits('vpn_bits)
> function vpn_j_of_va(vpn_size_bits, va, level) = {
>   let lsb = PAGESIZE_BITS + level * vpn_size_bits;
>   let msb = lsb + vpn_size_bits - 1;
>   va[msb .. lsb]
> }
> ```

We’re getting the same level of type-checking.  sv_params.vpn_size bits has the type {10,9}, not ’nat'.  I’ve narrowed the ‘level’ argument to have the Level type, which is range(0,3), not ‘nat’.  I’ve also narrowed ‘lsb ‘ to be < xlen.

model/riscv_vmem.sail
```diff
@@ -120,6 +120,9 @@ struct SV_Params = {
   pte_PPN_j_size_bits : {10,  9}         //   10     9     9
 }
 
+// Current level during a page-table walk (0 to SV_Params.levels - 1)
+type Level = range(0,3)    // range(0,4) when add Sv57 (TODO)
```
arichardson, line 124: Level is quite a generic name. Until there is a real way of declaring this file-local in sail, I think we have to use something like PTELevel here.

rsnikhil: Ok, changed name to PTW_Level

model/riscv_pte.sail
Timmmm, line 1: This file still needs to exist - it contains stuff that is overridden by CHERI.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+                                 pte_phys_addr,
+                                 8,                       // atom (8)
+                                 false,                   // aq
+                                 false,                   // rl
+                                 false);                  // res
+
+  match mem_result {
+    MemException(_) => PTW_Failure(PTW_Access(), ext_ptw),
+    MemValue(pte)   => {
+      let pte_lsbs = Mk_PTE_LSBs(pte[7 .. 0]);
```
Timmmm, line 509: Maybe worth a function for this since you do it a couple of times. Also is "LSBs" the best name? I would have gone with "flags".

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+      if pte_is_invalid(pte_lsbs) then
+        PTW_Failure(PTW_Invalid_PTE(), ext_ptw)
+      else {
+        // Non-Leaf PTE
+        let ppns : bits(64) = PPNs_of_PTE(sv_params, pte);
+        let global'         = global | (pte_lsbs.G() == 0b1);
```
Timmmm, line 515: Maybe `next_global` instead of `global'`. Bit weird syntax if you ask me - I mean someone reading this is going to think "what does `'` mean?" whereas `next_global` is obvious. And for the other variables.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+// For (non-standard) extensions: this function gets the extension-available bits
+// of the PTE in extPte, and the accumulated information of the page-table-walk
+// in ext_ptw. It should return the updated ext_ptw in both success and failure cases.
+
+union PTE_Check = {
+  PTE_Check_Success : ext_ptw,
+  PTE_Check_Failure : (ext_ptw, ext_ptw_fail)
+}
+
+// PRIVATE
```
Timmmm, line 327: Not private really since it is overridden by CHERI. Also it needs to go in `riscv_pte.sail`.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+                                 false);                  // res
+
+  match mem_result {
+    MemException(_) => PTW_Failure(PTW_Access(), ext_ptw),
+    MemValue(pte)   => {
+      let pte_lsbs = Mk_PTE_LSBs(pte[7 .. 0]);
+      if pte_is_invalid(pte_lsbs) then
+        PTW_Failure(PTW_Invalid_PTE(), ext_ptw)
+      else {
+        // Non-Leaf PTE
```
Timmmm, line 513: This comment is in the wrong place - should be on the line immediately after `if pte_is_ptr(pte_lsbs) then {`. And I'd move the `// Leaf PTE` comment into the `else {` too.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+          let pte_check = check_PTE_permission(ac, priv, mxr, do_sum, pte_lsbs,
+                                               ext_pte, ext_ptw);
+          match pte_check {
+            PTE_Check_Failure(ext_ptw, ext_ptw_fail) =>
+              PTW_Failure(ext_get_ptw_error(ext_ptw_fail), ext_ptw),
+            PTE_Check_Success(ext_ptw) =>
+              if level > 0 then {
+                // Superpage; construct mask for lower-level PPNs from the PTE
+                let mask_PPN_j : bits(64) = zero_extend(ones(sv_params.pte_PPN_j_size_bits));
+                mask           : bits(64) = zeros();
```
Timmmm, line 540: `var mask` (implicit `var` will be an error eventually).

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+                foreach (j from 0 to (level - 1))
+                  mask = ((mask << sv_params.pte_PPN_j_size_bits) | mask_PPN_j);
+                // Lower-level PPNs must be zero (for aligned superpage)
+                if not((ppns & mask) == zero_extend(0b0)) then
```
Timmmm, line 544: Would this work instead?

```
let mask_bits = level * sv_params.pte_PPN_j_size_bits;
// Clear the lowest `mask_bits` bits.
let ppns_masked = (ppns >> mask_bits) << mask_bits;
if not(ppns == ppns_masked) then ...
```

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+                  // misaligned superpage mapping
+                  PTW_Failure(PTW_Misaligned(), ext_ptw)
+                else {
+                  // Compose final PA in superpage:
+                  // Superpage PPN + lower VPNs + pagesize_bits page-offset
+                  let ppn = ppns | (vpns_of_va (sv_params, va) & mask);
+                  let pa  = ((ppn << pagesize_bits) | zero_extend(offset_of_va(va)));
```
Timmmm, line 551: Extra brackets.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+                  PTW_Success(pa, pte, pte_addr, level, global', ext_ptw)
+                }
+              }
+              else {
+                let pa = ((ppns << pagesize_bits) | zero_extend(offset_of_va (va)));
```
Timmmm, line 556: Extra brackets.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+
+// ----------------
+// Fields of SATP
+
+// ASID is 9b in Sv32, 16b in Sv39/Sv48/Sv57: we use 16b for both
+// PRIVATE
+function satp_to_asid(sv_params : SV_Params, satp_val : xlenbits) -> bits(16) = {
+  // This extend op is a no-op when xlen==64, extends when xlen==32
+  let satp_64b : bits(64) = zero_extend(satp_val);
+  let mask_64b : bits(64) = zero_extend(ones(sv_params.satp_asid_size_bits));
```
Timmmm, line 605: I don't think it makes sense for `satp` field sizes to be in `sv_params` because they only depend on XLEN, not the Sv39,Sv48,etc. mode. In other words this function should be:

```
function satp_to_asid(satp_val : xlenbits) -> bits(16) =
   if sizeof(xlen) == 32 then zero_extend(Mk_Satp32(satp_val)[Asid])
   else if sizeof(xlen) == 64 then Mk_Satp64(satp_val)[Asid]
   else internal_error(__FILE__, __LINE__, "Unsupported xlen" ^ string_of_int(sizeof(xlen)))
```

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+  let asid_64b : bits(64) = (satp_64b >>  sv_params.satp_asid_lsb_index) & mask_64b;
+  asid_64b[15 .. 0]
+}
+
+// Result is 64b to cover both RV32 and RV64 addrs
+// PRIVATE
+function satp_to_PT_base(sv_params : SV_Params, satp_val : xlenbits) -> bits(64) = {
```
Timmmm, line 612: Same for this function.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+  ppn_64b << pagesize_bits
+}
+
+// Compute address translation mode from SATP register
+// PRIVATE
+function translationMode(priv : Privilege) -> SATPMode = {
+  if priv == Machine then
+    Sbare
+  else if sizeof(xlen) == 32 then
+    match satp[31] {
```
Timmmm, line 626: Better to match `Mk_Satp32(satp)[Mode]`.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+    let arch = architecture(get_mstatus_SXL(mstatus));
+    match arch {
+      Some(RV64) => { let mbits : bits(4) = satp[63 .. 60];
+                      match satp64Mode_of_bits(RV64, mbits) {    // see riscv_types.sail
+                        Some(m) => m,
+                        None()  => internal_error(__FILE__, __LINE__,
+                                                  "invalid RV64 translation mode in satp")
+                      }
+                    },
+      Some(RV32) => match satp[31] {
```
Timmmm, line 644: Ditto.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+                           tlb_index : nat,
+                           ent       : TLB_Entry)
+                          -> TR_Result(bits(64), PTW_Error) = {
+  let pte       = ent.pte;
+  let ext_pte   = msbs_of_PTE(sv_params, pte);
+  let pte_lsbs  = Mk_PTE_LSBs(pte[7 .. 0]);
+  let pte_check = check_PTE_permission(ac, priv, mxr, do_sum, pte_lsbs,
+                                       ext_pte,
+                                       ext_ptw);
+  let tr_result : TR_Result(bits(64), PTW_Error)
```
Timmmm, line 688: Can remove this line and just have the last expression be the `match`.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+        match update_PTE_Bits(sv_params, pte, ac, ext_pte) {
+          None()     => TR_Address(ent.pAddr | (vAddr & ent.vAddrMask), ext_ptw),
+          Some(pte') =>
+            // See riscv_platform.sail
+            if not(plat_enable_dirty_update()) then
+              // pte needs dirty/accessed update but that is not enabled
+              TR_Failure(PTW_PTE_Update(), ext_ptw)
+            else {
+              // Writeback the PTE (which has new A/D bits)
+              n_ent : TLB_Entry = ent;
```
Timmmm, line 702: I think this works:

```
let n_ent = {ent with pte=pte'};
```

If not, we should at least use explicit `var n_ent : ...`

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+              n_ent.pte = pte';
+              write_TLB(tlb_index, n_ent);
+              let pte_phys_addr : xlenbits = ent.pteAddr[(sizeof(xlen) - 1) .. 0];
```
Timmmm, line 705: Can delete the `: xlenbits`.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+// PUBLIC: invoked from instr-fetch and load/store/amo
+function translateAddr(vAddr   : xlenbits,
+                       ac      : AccessType(ext_access_type))
+                      -> TR_Result(xlenbits, ExceptionType) = {
+  // Internally the vmem code works with 64-bit values, whether xlen==32 or xlen==64
+  // This 'extend' is a no-op when xlen==64 and extends when xlen==32
+  let vAddr_64b : bits(64) = zero_extend(vAddr);
+  // Effective privilege takes into account mstatus.PRV, mstatus.MPP
+  // See riscv_sys_regs.sail for effectivePrivilege() and cur_privilege
+  let effPriv   : Privilege = effectivePrivilege(ac, mstatus, cur_privilege);
```
Timmmm, line 816: Can delete `: Privilege` and `: SATPMode` and `: ext_ptw`

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+    Sbare => return TR_Address(vAddr, ext_ptw),
+    Sv32  => (true,                                 sv32_params),
+    Sv39  => (is_valid_vAddr(sv39_params, vAddr_64b), sv39_params),
+    Sv48  => (is_valid_vAddr(sv48_params, vAddr_64b), sv48_params),
+    // Sv57 => (is_valid_vAddr(sv57_params, vAddr_64b), sv57_params),    // TODO
+  };
+  if not(valid_va) then
+    TR_Failure(translationException(ac, PTW_Invalid_Addr()), ext_ptw)
+  else {
+    let mxr    : bool     = mstatus.MXR() == 0b1;
```
Timmmm, line 830: Delete all these type annotations.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+                      -> TR_Result(xlenbits, ExceptionType) = {
+  // Internally the vmem code works with 64-bit values, whether xlen==32 or xlen==64
+  // This 'extend' is a no-op when xlen==64 and extends when xlen==32
+  let vAddr_64b : bits(64) = zero_extend(vAddr);
+  // Effective privilege takes into account mstatus.PRV, mstatus.MPP
+  // See riscv_sys_regs.sail for effectivePrivilege() and cur_privilege
+  let effPriv   : Privilege = effectivePrivilege(ac, mstatus, cur_privilege);
+  let mode      : SATPMode = translationMode(effPriv);
+  // PTW extensions (non-standard): initialize the PTW extension state
+  let ext_ptw   : ext_ptw = init_ext_ptw;
```
Timmmm, line 819: Why make this copy? Just use `init_ext_ptw` directly.

Timmmm (changes requested):
Hey, sorry for the review delay - I missed that you replied.

Generally this looks good to me. Review summary:

* One bit definitely looks like a bug (in `update_PTE_Bits`). 
* Lots of unnecessary explicit type annotations. I think it's better to leave those out in almost all cases.
* Lots of unnecessary brackets around expressions. Sometimes it clarifies things, but I think if you have `let foo = (...);` then it just adds noise.
* Probably the biggest thing that needs to change is to move functions back into `riscv_pte.sail`/`riscv_ptw.sail` so that CHERI can override functions without having to completely copy & paste the page table walker.
* Some other misc stuff.

For CHERI I don't think we need to maintain exactly the same function signatures - some small amount of updating work is acceptable there I think. But we should maintain roughly the same code structure so it's at least feasible. And yeah it is kind of annoying that there's no clear indication exactly what the "public overidable" interfaces are, but I guess we'll just have to put up with that for now.

model/riscv_vmem_tlb.sail
```diff
@@ -68,57 +68,108 @@
...
+
+// PUBLIC: invoked in translate() [riscv_vmem.sail]
+function lookup_TLB (asid : bits(16), vaddr : bits(64)) -> option((nat, TLB_Entry)) =
+  match tlb {
+    None()  => None(),
+    Some(e) => if match_TLB_Entry(e, asid, vaddr) then Some((0, e)) else None()
+  }
+
+// PRIVATE
+function add_to_TLB(asid          : bits(16),
```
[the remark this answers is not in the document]
rsnikhil: Ok. Done.

model/riscv_vmem_tlb.sail
```diff
@@ -68,57 +68,108 @@
...
+                    vAddr         : bits(64),
+                    pAddr         : bits(64),
+                    pte           : bits(64),
+                    pteAddr       : bits(64),
+                    level         : nat,
+                    global        : bool,
+                    levelBitSize  : nat,
+                    PAGESIZE_BITS : nat)     -> unit = {
+  let shift      : nat       = PAGESIZE_BITS + (level * levelBitSize);
```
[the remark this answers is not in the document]
rsnikhil: Ok. Removed.

model/riscv_vmem_tlb.sail
```diff
@@ -68,57 +68,108 @@
...
+  assert(shift <= 64);
+  let vAddrMask  : bits(64)  = zero_extend(ones(shift));
+  let vMatchMask : bits(64)  = ~ (vAddrMask);
```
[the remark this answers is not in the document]
rsnikhil: In this case I think it serves as useful documentation, since 'zero_extend (ones (shift))' is polymorphic and so the result width may not be obvious.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+  // Update 'dirty' bit?
+  let update_d : bool = (pte_lsbs.D() == 0b0)
+                        & (match a {
+                             Execute()       => false,
+                             Read()          => false,
+                             Write(_)        => true,
+                             ReadWrite(_, _) => true
+                           });
+  // Update 'accessed'-bit?
+  let update_a : bool = pte_lsbs.A() == 0b0;
```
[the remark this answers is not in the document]
rsnikhil: Ok. Removed.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+
+  if update_d | update_a then {
+    let pte_lsbs1 = update_A (pte_lsbs, 0b1);
```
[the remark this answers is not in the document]
rsnikhil: Ok, fixed.  I was unaware of the new syntax.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+    let pte_lsbs2 = if update_d then update_D (pte_lsbs, 0b1) else pte_lsbs1;
```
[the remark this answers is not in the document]
rsnikhil: I’ve used your formulation which fixes the error and uses the newer ‘with’ syntax.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+  if success then PTE_Check_Success(())
+  else            PTE_Check_Failure((), ())
+}
+
+// Update PTE bits if needed; return new PTE if updated
+// PRIVATE
+function update_PTE_Bits(sv_params : SV_Params,
+                         pte       : bits(64),
+                         a         : AccessType(ext_access_type),
+                         ext       : extPte)
```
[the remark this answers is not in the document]
rsnikhil: Ok. Removed.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+                                          | ((pte_X == 0b1) & mxr)),
+      (Execute(),       Supervisor) => (pte_U == 0b0) & (pte_X == 0b1),
+      (_,               Machine)    => internal_error(__FILE__, __LINE__,
+                                                      "m-mode mem perm check")};
+  if success then PTE_Check_Success(())
+  else            PTE_Check_Failure((), ())
+}
+
+// Update PTE bits if needed; return new PTE if updated
+// PRIVATE
```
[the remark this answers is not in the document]
rsnikhil: See comment below about restoring riscv_pte.sail.

model/riscv_pte.sail
rsnikhil: I've punted on this for now.  It's a little disappointing because I consciously merged that file into riscv_vmem.sail, not knowing that the separation had a reason (facilitating adaptation for CHERI).  Now we have to un-merge.  Also, this may be better done by someone who knows what CHERI requires (which I do not).

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+
+// ****************************************************************
+// Page Table Walk (PTW)
+
+// Note: 'pt_walk()' is recursive => needs separate 'val' decls
+
+// PRIVATE
+val pt_walk : (SV_Params,
+               bits(64),                     // virtual addr
+               AccessType(ext_access_type),  // Read/Write/ReadWWrite/Execute
```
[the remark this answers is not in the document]
rsnikhil: Ok. Fixed.

model/riscv_vmem.sail
rsnikhil: Are you suggestion encapsulating `Mk_PTE_LSBs(pte[7 .. 0])` into a function?  Is that worth it?  Or am I missing the point?

model/riscv_vmem.sail
rsnikhil: In my experience, x, x’, x’’, …. is a common idiom in mathematics and programming to indicate next, next-next, etc.

model/riscv_vmem.sail
rsnikhil: See comment above about CHERI-related factoring.

model/riscv_vmem.sail
rsnikhil: Ok. Fixed.

model/riscv_vmem.sail
rsnikhil: Ok. I was unaware that Sail now recommends (and soon may demand) ‘var’ for updatable variables. But this fix is moot because of the reformulation in the next suggestion, which we’ve adopted, and which has no vars.

model/riscv_vmem.sail
rsnikhil: Ok. Adopted this formulation.

model/riscv_vmem.sail
rsnikhil: Ok. Removed.

model/riscv_vmem.sail
rsnikhil: Ok. Removed.

model/riscv_vmem.sail
rsnikhil: Ok. Adopted this suggestion.

model/riscv_vmem.sail
rsnikhil: Ok. Also adopted the suggestion above.

model/riscv_vmem.sail
rsnikhil: Ok. Fixed.

model/riscv_vmem.sail
rsnikhil: Ok. Fixed (although note that here, in RV64 mode, satp is 64bits, so can’t directly be an arg to Mk_Satp32).

model/riscv_vmem.sail
rsnikhil: Ok. Fixed.  I’ve commented elsewhere that sometimes naming the result is useful for debug-printing the result.

model/riscv_vmem.sail
rsnikhil: Ok. Fixed.  Was a ‘var’ only to do a field update; I was not aware that ‘with’ can be used to do it functionally. The Sail manual documents ‘with’ for vector and. bitfield updates but not for structs (in the main text, not the grammar).

model/riscv_vmem.sail
rsnikhil: Ok. Fixed.

model/riscv_vmem.sail
```diff
@@ -0,0 +1,854 @@
...
+          TR_Address(pAddr, ext_ptw)
+        },
+        Some(pte') =>
+          // See riscv_platform.sail
+          if not(plat_enable_dirty_update()) then
+            // pte needs dirty/accessed update but that is not enabled
+            TR_Failure(PTW_PTE_Update(), ext_ptw)
+          else {
+            // Writeback the PTE (which has new A/D bits)
+            let pte_phys_addr : xlenbits = pteAddr[(sizeof(xlen) - 1) .. 0];
```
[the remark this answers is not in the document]
rsnikhil: Ok. Fixed.

model/riscv_vmem.sail
rsnikhil: Here I think the annotations serve a useful documentation purpose. The RHS functions are non-local and the type of their results may not be obvious.  Personally I  prefer more rather than less type annotations.

model/riscv_vmem.sail
rsnikhil: Ok (partial). I’ve deleted them for the obvious boolean exprs, but not for the others, where I think they serve as useful documentation.

model/riscv_vmem.sail
rsnikhil: Ok. Fixed.

model/riscv_vmem.sail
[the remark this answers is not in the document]
Timmmm: Github won't let me actually resolve this for some reason but it's resolved.

Timmmm:
Just wanted to say I haven't forgotten about this - I'm just working on testing it.

Alasdair:
If you don't want to continue, would it be possible for me to take over this PR? At the very least I would like this on my own fork, even if nothing can be merged here

rsnikhil:
> If you don't want to continue, would it be possible for me to take over this PR? At the very least I would like this on my own fork, even if nothing can be merged here

Hi Alasdair, no I will see it through. This PR (#365) got marked as 'closed' because of my ham-handedness with GitHub/git.  I updated my fork of sail-riscv on GitHub to sync up with the latest commits, then did 'git pull - rebase' to my clone on my laptop, and (a) it closed the PR and (b) threw out all my changes.  Fortunately I had a backup of all my changes.  I have re-incorporated my changes and re-tested, and it looks ok.  I'll create a new PR shortly. Apologies for the confusion.

Alasdair:
Ah, no worries then. I misinterpreted you closing the PR.

That said I remember you said you would be busy at some point soon, so if you do need someone to help drive this forwards please don't hesitate to ask me.

rsnikhil:
> Ah, no worries then. I misinterpreted you closing the PR.
> 
> That said I remember you said you would be busy at some point soon, so if you do need someone to help drive this forwards please don't hesitate to ask me.

I have just now created PR #408 for this (it is running CI tests, which already passed in my fork).

I also separated out code into riscv_vmem_pte.sail and riscv_vmem_ptw.sail per @Timmmm ’s suggestion for convenience for CHERI project which, as I understand it, substitute these files with CHERI-specific files.

It’s been ready for merge for some time now (all PR comments addressed).  @Timmmm said he’d run it through some additional tests that are not part of the CI tests.

Timmmm:
I haven't forgotten! Give me another week or so...

PR #366, opened 2023-12-02
Fix kext warning

Alasdair:
Use fixed-length vectors for AES tables. This fixes a warning about an incomplete match, and should be much more efficient than using linked lists. It also makes the Sail->SystemVerilog translation happy as a nice bonus.

Note that because we have `default Order dec`, and Sail vectors are indexed the same as bitvectors in decreasing order, the indexing in sbox_lookup becomes
```
table[255 - unsigned(x)]
```
rather than
```
table[unsigned(x)]
```
The alternative would be to reverse all the table literals.


Timmmm:
Could you use `vector(256, inc, bits(8))` instead? Not sure that would be better tbh even if it does work.

Timmmm (approved):

billmcspadden-riscv (approved):
nice improvement.

nit - it would be nice to see the tables formed as 16x16 or 32x8.

Alasdair:
> Could you use `vector(256, inc, bits(8))` instead? Not sure that would be better tbh even if it does work.

The Order parameter on vectors no longer does anything (all vectors and bitvectors have the default order). It never worked well and was a big source of internal complexity and headaches.

PR #367, opened 2023-12-03
Remove effect annotations from vector extension

Alasdair:
We removed these a while ago from everything else, but I guess it didn't happen for vector because it was on a branch.

Removes the warning these produce.


billmcspadden-riscv (approved):
lgtm

Issue #368, opened 2023-12-07
Looking for help/suggestion to locate source of a type error


ved-rivos:
Whats a good way to identify the cause for this error?

```
The following expression is unmatched: FVV_VFWREDOSUM

Type error:
No type declaration found for _s8698#
make: *** [Makefile:262: generated_definitions/c/riscv_model_RV64.c] Error 1

```

ved-rivos:
I am developing an extension in my local checkout. I have narrowed it down to one file (a new file I added), which if exclude from make, makes the error go away. I will try to narrow it further by adding stuff into that file little by little.

Alasdair:
If your version of Sail is < 0.17.1, I think an `opam update && opam ugrade sail`, might just fix it.

arichardson:
> If your version of Sail is < 0.17.1, I think an `opam update && opam upgrade sail`, might just fix it.

Can we add a minimum version check? That would fix issues like this?

Alasdair:
That might be a good idea, I made a draft PR to discuss that.

PR #370, opened 2023-12-08
Add a Sail version check to Makefile

Alasdair:
In https://github.com/riscv/sail-riscv/issues/368#issuecomment-1846210592 it was suggested to add a Sail version check to prevent errors where some Makefile invocation fails mysteriously due to a too old version of Sail.

I'm not sure what the best way to do this is: Here I've added a small shell script and a check at the top of the Makefile.

There are some pathological cases where a Sail binary might not know its own version, which would be when it's built from source without either git or opam. I don't think this is likely to ever happen, but I've added an option to skip the version check by setting an environment variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,12 @@
+# Check the current Sail version
+SAIL_VERSION := $(shell etc/version_check.sh > /dev/null; echo $$?)
+ifeq ($(SAIL_VERSION),1)
+ifeq ($(SKIP_SAIL_VERSION_CHECK),)
+  $(info $(shell etc/version_check.sh))
+  $(error 'Sail version check failed. Set SKIP_SAIL_VERSION_CHECK=true to ignore.')
+endif
+endif
+
 # Select architecture: RV32 or RV64.
 ARCH ?= RV64
 
--- a/etc/version_check.sh
+++ b/etc/version_check.sh
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+req_major="0"
+req_minor="17"
+req_patch="1"
+
+function check {
+    local version=$(sail -v | cut -d ' ' -f 2);
+
+    if [ "$?" -ne 0 ]; then
+        echo "Could not find Sail version: 'sail -v' command failed"
+        exit 1
+    fi
+
+    local message="Sail version must be at least ${req_major}.${req_minor}.${req_patch}, got ${version}"
+
+    local major=$(echo $version | cut -d '.' -f 1);
+    local minor=$(echo $version | cut -d '.' -f 2);
+    local patch=$(echo $version | cut -d '.' -f 3);
+
+    # Sail version string may be two digits without a patch version,
+    # in which case we treat it as zero.
+    if [ -z "$patch" ]; then
+        patch="0"
+    fi
+
+    if [ "$major" -lt "$req_major" ]; then
+        echo "${message}"
+        exit 1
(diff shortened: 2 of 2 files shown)


arichardson:
Would it be simpler to use something like the `-has_feature` flag in sail to require a given feature so we don't have to parse arguments? I guess not every release has new features so maybe the version check is better.

Alasdair:
Yes, that's been fairly ad-hoc. If I was more forward-thinking I could have added a feature flag per-release, but I haven't.

I could add something like a `--version-check` flag, but of course that wouldn't serve the goal at first because the older versions we would want to check for wouldn't recognise that flag.

etc/version_check.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
```
arichardson, line 1: This script uses some non-POSIX features so should require bash or remove the non-POSIX features. You could also paste it into shellcheck.net to see those issues.

arichardson:
I think it would be very useful to land this so we have a way of enforcing a minimum version of sail and won't see any more issues filed due to an incorrect version.

Timmmm:
Perhaps to avoid janky Bash scripts Sail could do this check itself, something like `--min-version 0.17` and it exits with a non-zero exit code if it is less than that version. This would be useful for anyone using Sail (not just this repo).

It does mean we need to wait a bit until the minimum version has that flag, but I'd say that's acceptable.

arichardson:
> Perhaps to avoid janky Bash scripts Sail could do this check itself, something like `--min-version 0.17` and it exits with a non-zero exit code if it is less than that version. This would be useful for anyone using Sail (not just this repo).
> 
> It does mean we need to wait a bit until the minimum version has that flag, but I'd say that's acceptable.

Adding support to sail itself sounds like a good plan to me. Alternatively it could pre-define something like a SAIL_VERSION symbol that is checked by prelude.sail and we emit an error if it's insufficient.

allenjbaum:
Or at least a warning?

arichardson:
I like @Timmmm's suggestion, but I think we should really land some kind of version check soon. 
No need to wait for a perfect solution if we can fix the most important issue with this PR.

I think this PR should be good to land with the shellcheck issues fixed.

Timmmm:
I agree we should do this soon, but I also have an entirely rational hatred of Bash, so I've made a PR to add a version check flag to Sail: https://github.com/rems-project/sail/pull/497

Makefile
```diff
@@ -1,3 +1,12 @@
+# Check the current Sail version
+SAIL_VERSION := $(shell etc/version_check.sh > /dev/null; echo $$?)
```
Timmmm, line 2: Now that the new Sail version has been released - and CI for this repo automatically uses the latest version - we should be able to do something like

```
SAIL_VERSION := $(shell $(SAIL) --version --require-version 0.18 > /dev/null; echo $$?)
```

Also I'd suggest calling it `SAIL_VERSION_STATUS` or `SAIL_VERSION_ERROR` or something because it isn't actually the version.

Alternatively...

```
SAIL += --require-version 0.18
```

jordancarlin: It seems like just adding `--require-version 0.18` to `SAIL_FLAGS` is probably the simplest and cleanest solution. And it gives a fairly descriptive error message on its own: `Sail version 0.17.1 is older than requested version 0.18`.

Timmmm: I made a new PR for that here: #534

PR #371, opened 2023-12-11
[DNM before Sail 0.17.2] Catch up to sail mainline

nwf-msr:
These definitions are now provided by sail's libraries, as of https://github.com/rems-project/sail/commit/eb8af69724828181bf0e91f1728399fe8a81e6f0 .

This includes the change in https://github.com/riscv/sail-riscv/pull/349 and, presumably, will also not pass CI until the sail version used there is bumped.


nwf-msr:
Lest anyone get too hasty with the merge button, I'm not sure -- I can check, I have just been a little lazy and focused on trying to get things working on my machine -- that this is compatible with sail 0.17.1 and isn't depending on something between that release (https://github.com/rems-project/sail/commit/8217b279ef55b3c32e792b9683a75f2ce0cfad29) and current `HEAD`.

I do not envy the `sail-riscv` maintainers the task of coordination.

Alasdair:
The CI tests with the last opam published version so that should be fine.

I do plan on fixing up the hex_bits situation in the next Sail release, so I'm not sure if now is the right time to add a `mapping.sail` import. I think it just defines `hex_bits_20` because it was used in one test case, but I recently added a new file that defines all 1-32 in terms of a polymorphic hex_bits mapping.

arichardson (approved):

Timmmm:
I think this is a subset of #318 which is still waiting for review. Possibly also #349 (I guess that's why this failed; I haven't looked).

nwf-msr:
Ah, somehow I'd missed #318!  I'll go comment over there.

PR #373, opened 2023-12-12
Bitfield syntax updates

Alasdair:
Sail for a while now (at least several versions at this point) has supported bitfield syntax that is closer to regular bit vector syntax. This PR updates the model to use this newer syntax. In theory this might be even more efficient (although I have not benchmarked), as the old syntax used special accessor functions that are generated by the `bitfield` declaration, whereas the new syntax avoids that indirection.

To summarise, for a bitfield:
```
bitfield B : bits(32) = {
  F : 7 .. 0
}
```

- Accessing can be `x[F]` rather than `x.F()`.
- Similarly, modifying can be `x[F] = 0x00` rather than `x->F() = 0x00`.
- Updating a field can be done as `[x with F = 0x00]` rather than `update_F(x, 0x00)`.
- Getting all the bits is now `x.bits` rather than `x.bits()`, and the same for setting with `x.bits = 0x0000_0000`.

Alternatively, one can look at the [manual section on bitfields](https://alasdair.github.io/manual.html#_bitfields)

The commit adding the update syntax is separate, as it was slightly more manual than the other changes. In general, code that looked like:
```
let x = update_F1(x, y);
let x = update_F2(x, z);
x
```
can be changed to
```
[x with F1 = y, F2 = z]
```
I avoided making this update change in code I know is going to change soon (like the virtual memory code), or the PMP definitions, and I have no plans to remove the `update_` functions anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -255,9 +255,11 @@ gcovr:
 ocaml_emulator/tracecmp: ocaml_emulator/tracecmp.ml
 	ocamlfind ocamlopt -annot -linkpkg -package unix $^ -o $@
 
+c_preserve_fns=-c_preserve _set_Misa_C
+
 generated_definitions/c/riscv_model_$(ARCH).c: $(SAIL_SRCS) model/main.sail Makefile
 	mkdir -p generated_definitions/c
-	$(SAIL) $(SAIL_FLAGS) -O -Oconstant_fold -memo_z3 -c -c_include riscv_prelude.h -c_include riscv_platform.h -c_no_main $(SAIL_SRCS) model/main.sail -o $(basename $@)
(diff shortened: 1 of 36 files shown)


Alasdair:
It's been around for a while but I don't think I had documented it well enough. It's in my new asciidoc manual now at least as the recommended way.

I might make the `x.F()` syntax and `x->F()` syntax a warning in the next version of Sail. If I ever add namespacing I think I'd want to use `.` as the separator, and having an `<expression> . <id> ()` construct in the grammar will make disambiguating that annoying.

Timmmm (approved):

Timmmm:
@billmcspadden-riscv I think this should be merged ASAP to avoid inflicting painful rebasing on Alasdair.

arichardson:
This is a purely syntactic change and has been open for 1.5 months now. Can we merge it?

PR #376, opened 2023-12-15
Makefile: Make sure OPAMCLI is 2.0 in all subshells

Alasdair:
Follow up to https://github.com/riscv/sail-riscv/pull/364 to make sure OPAMCLI is defined when it is used within the shell commands. Also implement Brian's suggestion that variables defined with opam should use `:=` to avoid calling opam each time the variable is expanded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,14 +128,16 @@ export SAIL_DIR
 EXPLICIT_COQ_SAIL=yes
 else
 # Use sail from opam package
-SAIL_DIR:=$(shell opam config var sail:share)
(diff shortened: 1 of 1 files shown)


Alasdair:
Seems like the warning is not present in the GitHub Actions log, so I think this has actually removed the warning successfully

arichardson (approved):

arichardson (approved):
I guess another option could be to use something like `$(shell $(OPAM) config var)` with OPAM expanding to `OPAMCLI=2.0 opam`? But either seems fine to me.

Timmmm:
Yeah I think @arichardson 's suggestion is mildly better but probably not worth worrying about. LGTM.

PR #377, opened 2023-12-17
Add unratified Zicfiss extension

ved-rivos:
Specification: https://github.com/riscv/riscv-cfi/blob/main/cfi_backward.adoc
Tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/417


Alasdair:
Just letting you know I am away on Holiday until early January, so it might take me a while before I can go through this in detail.

ved-rivos:
Sounds good to me.

PR #379, opened 2023-12-29
Add unratified Ssqosid extension

ved-rivos:
1. Adds `mstateen0/mstateen0h` CSR introduced by Smstateen - only field `PRIV114` defined to control access to `srmcfg` introduced. Other fields need to be added as part of adding related extensions.
2. Adds `srmcfg` CSR introduced by Ssqosid extension. Access to this CSR at S-mode is governed by `PRIV114` bit.
3. Tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/420
4. Specification: https://github.com/riscv/riscv-ssqosid


model/riscv_sys_regs.sail
```diff
@@ -882,6 +888,24 @@ function is_fiom_active() -> bool = {
     User => (menvcfg.FIOM() | senvcfg.FIOM()) == 0b1,
   }
 }
+
+bitfield Mstateen0 : bits(64) = {
+  // controls to new state introduced by the following small extensions of Priv-1.14-era:
+  PRIV114  : 55
+}
```
Alasdair, line 896: Should add spaces between these toplevel definitions

model/riscv_sys_regs.sail
```diff
@@ -882,6 +888,24 @@ function is_fiom_active() -> bool = {
     User => (menvcfg.FIOM() | senvcfg.FIOM()) == 0b1,
   }
 }
+
+bitfield Mstateen0 : bits(64) = {
+  // controls to new state introduced by the following small extensions of Priv-1.14-era:
```
Alasdair, line 893: Could maybe add a link or slightly more explanation of what Priv-1.14-era means here?

Also I'm not sure if this comment makes sense 'following small extensions', but there is no following list of small extensions. Could just remove 'following small'.

model/riscv_sys_regs.sail
```diff
@@ -882,6 +888,24 @@ function is_fiom_active() -> bool = {
...
+  PRIV114  : 55
```
Alasdair, line 895: I would probably do PRIV_1_14, to separate out the 1 and 14 in 1.14

Alasdair (commented):
I left some comments. I think this extension is pretty simple and can be merged once those are addressed.

model/riscv_sys_regs.sail
Alasdair: Or maybe 1p14 as that's how the notation used in the RISC-V config string works

ved-rivos: Using 1p14 or 1_14 may end up not match the name that would be in specification and the [riscv-opcodes](https://github.com/riscv/riscv-opcodes/blob/f51aa42dac51c071c516a96648344a50339fa70d/encoding.h#L171).

model/riscv_sys_regs.sail
ved-rivos: Here is a link to the specification: https://github.com/riscv/riscv-isa-manual/issues/1147. I will update the comment - it should have included Ssqosid state as the state governed by this bit for now but more state may be included by time Priv 1.14 is frozen.

model/riscv_sys_regs.sail
Alasdair: That is an argument for keeping it like it is, but just ambiguously concatenating version numbers like this seems like a bad idea, so I think the inconsistency is probably worth it. I'm not sure we have any other examples of names with version numbers in them. If we want ALLCAPS here maybe 'NAME_V1P14'.

I'm just imagining the inevitable future where we end up with a bunch of versioned identifiers like this, so it might be good to set a convention now.

model/riscv_sys_regs.sail
ved-rivos: done

model/riscv_sys_regs.sail
ved-rivos: I have updated the comments.

model/riscv_sys_regs.sail
ved-rivos: Updated to use PRIV_V1P14.

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,45 @@
...
+// directories except where otherwise noted is subject the BSD
+// two-clause license in the LICENSE file.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+// =======================================================================================
+
+type rcidlen : Int = config extensions.Ssqosid.rcidlen
+type mcidlen : Int = config extensions.Ssqosid.mcidlen
+constraint 1 <= rcidlen <= 12
+constraint 1 <= mcidlen <= 12
```
jordancarlin: @ved-rivos I suspect this is the section that is giving you trouble with the SMT solver. I also don't think it is necessary. You should be fine with just a let binding:
```
let rcidlen : range(1, 12) = config extensions.Ssqosid.rcid_length
let mcidlen : range(1, 12) = config extensions.Ssqosid.mcid_length
```

@Alasdair It would be great to get better error messages for this if possible. The current output with the smt debug flag is the following (which is pretty cryptic even for those of us who are familiar with Sail):
```
Prove (src/lib/type_check.ml/640) (0 <= '_#sys_misaligned_allowed_within_exp & '_#sys_misaligned_allowed_within_exp <= 11), (0 <= '_#sys_misaligned_allowed_within_exp#0 & '_#sys_misaligned_allowed_within_exp#0 <= 11), (0 <= '_#sys_pmp_grain & '_#sys_pmp_grain <= 63), (0 <= '_#sys_pmp_grain#0 & '_#sys_pmp_grain#0 <= 63), (0 <= '_#sys_pmp_usable_count & '_#sys_pmp_usable_count <= 64), (0 <= '_#sys_pmp_usable_count#0 & '_#sys_pmp_usable_count#0 <= 64), '_#sys_pmp_count in {0, 16, 64}, '_#sys_pmp_count#0 in {0, 16, 64}, (1 <= '_#mcidlen & '_#mcidlen <= 12), (1 <= '_#mcidlen#0 & '_#mcidlen#0 <= 12), (1 <= '_#rcidlen & '_#rcidlen <= 12), (1 <= '_#rcidlen#0 & '_#rcidlen#0 <= 12), '_#plat_insns_per_tick > 0, '_#plat_insns_per_tick#0 > 0, (2 <= '_#plat_reservation_set_size_exp & '_#plat_reservation_set_size_exp <= 12), (2 <= '_#plat_reservation_set_size_exp#0 & '_#plat_reservation_set_size_exp#0 <= 12), (0 <= '_#plat_cache_block_size_exp & '_#plat_cache_block_size_exp <= 12), (0 <= '_#plat_cache_block_size_exp#0 & '_#plat_cache_block_size_exp#0 <= 12), (1 <= mcidlen & mcidlen <= 12), (1 <= rcidlen & rcidlen <= 12), ((3 <= elen_exp & elen_exp <= 16) & elen_exp <= vlen_exp), (3 <= vlen_exp & vlen_exp <= 16), xlen in {32, 64} |- ('n + 'm) == 16
SMTLIB2 constraints are:
(push)
(declare-const zbase_E_enabled Bool)
(declare-const zelen_exp Int)
(declare-const zext_d_supported Bool)
(declare-const zvlen_exp Int)
(declare-const zxlen Int)
(declare-const v9 Int)
(declare-const v8 Int)
(declare-const v1 Int)
(declare-const v0 Int)
(declare-const v5 Int)
(declare-const v4 Int)
(declare-const v3 Int)
(declare-const v2 Int)
(declare-const v7 Int)
(declare-const v6 Int)
(declare-const v17 Int)
(declare-const v16 Int)
(declare-const v11 Int)
(declare-const v10 Int)
(declare-const v15 Int)
(declare-const v14 Int)
(declare-const v13 Int)
(declare-const v12 Int)
(declare-const v18 Int)
(declare-const v19 Int)

(assert (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (and (not (= (+ v19 v18) 16)) (and (<= 0 v17) (<= v17 11))) (and (<= 0 v16) (<= v16 11))) (and (<= 0 v15) (<= v15 63))) (and (<= 0 v14) (<= v14 63))) (and (<= 0 v13) (<= v13 64))) (and (<= 0 v12) (<= v12 64))) (or (= v11 0) (= v11 16) (= v11 64))) (or (= v10 0) (= v10 16) (= v10 64))) (and (<= 1 v9) (<= v9 12))) (and (<= 1 v8) (<= v8 12))) (and (<= 1 v7) (<= v7 12))) (and (<= 1 v6) (<= v6 12))) (> v5 0)) (> v4 0)) (and (<= 2 v3) (<= v3 12))) (and (<= 2 v2) (<= v2 12))) (and (<= 0 v1) (<= v1 12))) (and (<= 0 v0) (<= v0 12))) (and (<= 1 zmcidlen) (<= zmcidlen 12))) (and (<= 1 zrcidlen) (<= zrcidlen 12))) (and (and (<= 3 zelen_exp) (<= zelen_exp 16)) (<= zelen_exp zvlen_exp))) (and (<= 3 zvlen_exp) (<= zvlen_exp 16))) (or (= zxlen 32) (= zxlen 64))))
(check-sat)
(pop)

Error:
SMT solver returned unexpected status 1
```

jordancarlin: @ved-rivos Actually, it looks like you have a let binding version of this already in `platform_config.sail` pointing to json keys that don't exist. I'd recommend dropping this portion and updating the config keys you already have in `platform_config.sail`.

Alasdair: True, I think this should work as written though.

Alasdair: It might be the overlap with the same named letbindings that is causing the constraint generation to go wrong though.

jordancarlin: We have several let bindings that overlap with the names of abstract types (like xlen) that work without issue. Those just set the let binding to the size of the abstract type though.

jordancarlin: I just tested it locally, and without these abstract types/constraints, the SMT solver finishes without issues. With the abstract types, it fails even if the let binding is moved after the abstract type or removed completely. So it seems like something else is going on besides the overlapping names.

model/postlude/validate_config.sail
```diff
@@ -283,6 +283,18 @@ function check_misc_extension_dependencies() -> bool = {
...
+    valid = false;
+    print_endline("The Ssqosid extensions is enabled but Zicsr is disabled: supporting Ssqosid requires Zicsr.");
+  };
+  if hartSupports(Ext_Ssqosid) & not(hartSupports(Ext_Zicsr)) then {
+    valid = false;
+    print_endline("The Ssqosid extensions is enabled but Zicsr is disabled: supporting Ssqosid requires Zicsr.");
+  };
+  if hartSupports(Ext_Ssqosid) & not(hartSupports(Ext_S)) then {
+    valid = false;
+    print_endline("The Ssqosid extensions requires supervisor mode (S) but supervisor mode is not enabled.");
```
jordancarlin: Is this true? The spec says
> The Ssqosid extension does not require that S-mode mode be implemented.

model/postlude/validate_config.sail
```diff
@@ -283,6 +283,18 @@ function check_misc_extension_dependencies() -> bool = {
     valid = valid & check_required_sstvala_option("hardware breakpoint exceptions", hardware_breakpoint_writes_xtval);
     valid = valid & check_required_sstvala_option("illegal instruction exceptions", illegal_instruction_writes_xtval);
   };
+  if hartSupports(Ext_Ssqosid) & not(hartSupports(Ext_Zicsr)) then {
+    valid = false;
+    print_endline("The Ssqosid extensions is enabled but Zicsr is disabled: supporting Ssqosid requires Zicsr.");
+  };
+  if hartSupports(Ext_Ssqosid) & not(hartSupports(Ext_Zicsr)) then {
+    valid = false;
+    print_endline("The Ssqosid extensions is enabled but Zicsr is disabled: supporting Ssqosid requires Zicsr.");
```
jordancarlin: Check for Zicsr is repeated twice.

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,43 @@
...
+
+let rcidlen : range(1, 12) = config extensions.Ssqosid.rcid_length
+let mcidlen : range(1, 12) = config extensions.Ssqosid.mcid_length
+
+bitfield Srmcfg : xlenbits = {
+  RCID  : 27 .. 16,
+  MCID  : 11 .. 0
+}
+
+function clause currentlyEnabled(Ext_Ssqosid) = hartSupports(Ext_Ssqosid)
```
jordancarlin: This should also check for `currentlyEnabled(Ext_Zicsr)` and (depending on the resolution of my other comment) `Ext_S`.

model/extensions/Ssqosid/ssqosid.sail
ved-rivos: @jordancarlin @Alasdair thanks for helping resolve the SMT solver issue.

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,43 @@
...
+}
+
+register srmcfg : Srmcfg
+
+// Ssqosid - Quality-of-Service (Qos) Identifiers
+mapping clause csr_name_map = 0x181  <-> "srmcfg"
+
+// TODO: Add dependence on mstateen0 when that is added
+function srmcfg_CSRs_accessible(priv : Privilege) -> bool =
+    ((priv == Machine) | (priv == Supervisor))
```
jordancarlin: I think this is already captured by the `check_CSR_priv` function that all CSR accesses are gated by.

```
function csrPriv(csr : csreg) -> nom_priv_bits = csr[9..8]

// Check that the CSR access is made with sufficient privilege.
function check_CSR_priv(csr : csreg, p : Privilege) -> bool =
  privLevel_to_bits(p) >=_u csrPriv(csr)
```

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,43 @@
...
+
+function clause is_CSR_accessible(0x181, priv, _) =
+    currentlyEnabled(Ext_Ssqosid) & srmcfg_CSRs_accessible(priv)
+
+function clause read_CSR(0x181) = srmcfg.bits[xlen - 1 .. 0]
```
jordancarlin: The bitfield is `xlen` bits, so I think this should work without the bit slicing:

```suggestion
function clause read_CSR(0x181) = srmcfg.bits
```

jordancarlin:
Also make sure to add the extension to the README and Changelog.

model/postlude/validate_config.sail
ved-rivos: You are right. The S-mode need not be implemented.

model/postlude/validate_config.sail
ved-rivos: fixed.

model/extensions/Ssqosid/ssqosid.sail
ved-rivos: Updated. Should only depend on Zicsr.

model/extensions/Ssqosid/ssqosid.sail
ved-rivos: I am leaving it there because it needs to be augmented by mstateen0.SRMCFG being 1. I will be sending Smstateen PR separately as it affects this extension plus other extensions already in the model like Sstc.

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,44 @@
...
+let rcidlen : range(1, 12) = config extensions.Ssqosid.rcid_length
+let mcidlen : range(1, 12) = config extensions.Ssqosid.mcid_length
+
+bitfield Srmcfg : xlenbits = {
+  RCID  : 27 .. 16,
+  MCID  : 11 .. 0
+}
+
+function clause currentlyEnabled(Ext_Ssqosid) = hartSupports(Ext_Ssqosid) &
+                                                hartSupports(Ext_Zicsr)
```
jordancarlin: Minor in the case of Zicsr since it can't be disabled, but for consistency with the rest of the model, this should use `currentlyEnabled`, not `hartSupports`
```suggestion
function clause currentlyEnabled(Ext_Ssqosid) = hartSupports(Ext_Ssqosid) &
                                                currentlyEnabled(Ext_Zicsr)
```

model/extensions/Ssqosid/ssqosid.sail
ved-rivos: fixed

model/extensions/Ssqosid/ssqosid.sail
ved-rivos: fixed

ved-rivos:
@jordancarlin where is Changelog?

jordancarlin:
> @jordancarlin where is Changelog?

`doc/ChangeLog.md`

model/core/extensions.sail
```diff
@@ -686,6 +691,7 @@ let extensions_ordered_for_isa_string = [
   Ext_Sstvala,
   Ext_Sstvecd,
   Ext_Ssu64xl,
+  Ext_Ssqosid,
```
jordancarlin: This list needs to be in canonical order. Ssqosid should be before Sstc.

ved-rivos: updated.

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,44 @@
...
+  ]
+}
+
+register srmcfg : Srmcfg
+
+// Ssqosid - Quality-of-Service (Qos) Identifiers
+mapping clause csr_name_map = 0x181  <-> "srmcfg"
+
+// TODO: Add dependence on mstateen0 when that is added
+function srmcfg_CSRs_accessible(priv : Privilege) -> bool =
```
jordancarlin: There is only one srmcfg CSR, so how about just calling this `srmcfg_accessible`.

ved-rivos: updated.

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,44 @@
...
+  ]
+}
+
+register srmcfg : Srmcfg
+
+// Ssqosid - Quality-of-Service (Qos) Identifiers
+mapping clause csr_name_map = 0x181  <-> "srmcfg"
+
+// TODO: Add dependence on mstateen0 when that is added
+function srmcfg_CSR_accessible(priv : Privilege) -> bool =
```
jordancarlin:
```suggestion
private function srmcfg_CSR_accessible(priv : Privilege) -> bool =
```

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,44 @@
...
+
+bitfield Srmcfg : xlenbits = {
+  RCID  : 27 .. 16,
+  MCID  : 11 .. 0
+}
+
+function clause currentlyEnabled(Ext_Ssqosid) = hartSupports(Ext_Ssqosid) &
+                                                currentlyEnabled(Ext_Zicsr)
+
+function legalize_srmcfg(o: Srmcfg, value : xlenbits) -> Srmcfg = {
```
jordancarlin:
```suggestion
private function legalize_srmcfg(o: Srmcfg, value : xlenbits) -> Srmcfg = {
```

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,44 @@
...
+
+// Ssqosid - Quality-of-Service (Qos) Identifiers
+mapping clause csr_name_map = 0x181  <-> "srmcfg"
+
+// TODO: Add dependence on mstateen0 when that is added
+function srmcfg_CSR_accessible(priv : Privilege) -> bool =
+    ((priv == Machine) | (priv == Supervisor))
+
+function clause is_CSR_accessible(0x181, priv, _) =
+    currentlyEnabled(Ext_Ssqosid) & srmcfg_CSR_accessible(priv)
```
pmundkur: indentation is 2 spaces
```suggestion
  currentlyEnabled(Ext_Ssqosid) & srmcfg_CSR_accessible(priv)
```

pmundkur:
Either this or #910 would need to be updated depending on which gets merged first.

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,44 @@
...
+let rcidlen : range(1, 12) = config extensions.Ssqosid.rcid_length
+let mcidlen : range(1, 12) = config extensions.Ssqosid.mcid_length
+
+bitfield Srmcfg : xlenbits = {
+  RCID  : 27 .. 16,
+  MCID  : 11 .. 0
+}
+
+function clause currentlyEnabled(Ext_Ssqosid) = hartSupports(Ext_Ssqosid) &
+                                                currentlyEnabled(Ext_Zicsr)
```
nadime15: Lets put this on the same line
```suggestion
function clause currentlyEnabled(Ext_Ssqosid) = hartSupports(Ext_Ssqosid) & currentlyEnabled(Ext_Zicsr)
```

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,44 @@
...
+register srmcfg : Srmcfg
+
+// Ssqosid - Quality-of-Service (Qos) Identifiers
+mapping clause csr_name_map = 0x181  <-> "srmcfg"
+
+// TODO: Add dependence on mstateen0 when that is added
+private function srmcfg_CSR_accessible(priv : Privilege) -> bool =
+    ((priv == Machine) | (priv == Supervisor))
+
+function clause is_CSR_accessible(0x181, priv, _) =
```
nadime15:
```suggestion
function clause is_CSR_accessible(0x181, priv, _) = currentlyEnabled(Ext_Ssqosid)
```

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,44 @@
...
+
+register srmcfg : Srmcfg
+
+// Ssqosid - Quality-of-Service (Qos) Identifiers
+mapping clause csr_name_map = 0x181  <-> "srmcfg"
+
+// TODO: Add dependence on mstateen0 when that is added
+private function srmcfg_CSR_accessible(priv : Privilege) -> bool =
+    ((priv == Machine) | (priv == Supervisor))
+
```
nadime15:
```suggestion
```

This is probably not needed, due to check_CSR_priv(csr, p) that already verifies, based on the current privilege mode and the csr address, whether access is allowed so user mode will not be able to access it.

model/postlude/validate_config.sail
```diff
@@ -283,6 +283,10 @@ function check_misc_extension_dependencies() -> bool = {
     valid = valid & check_required_sstvala_option("hardware breakpoint exceptions", hardware_breakpoint_writes_xtval);
     valid = valid & check_required_sstvala_option("illegal instruction exceptions", illegal_instruction_writes_xtval);
   };
+  if hartSupports(Ext_Ssqosid) & not(hartSupports(Ext_Zicsr)) then {
+    valid = false;
+    print_endline("The Ssqosid extensions is enabled but Zicsr is disabled: supporting Ssqosid requires Zicsr.");
+  };
```
nadime15, line 293: Why is this a hard requirement? We have tons of extensions that actually just work with Zicsr and we dont check for that

jordancarlin: The config validation is a newer addition. I think it should be as strict as possible to catch as many config errors as possible. For new extensions, we should check all of their dependencies in the config validation. Eventually we will need to add better validation for all of the older extensions, but that shouldn't prevent us from validating the new ones.

model/extensions/Ssqosid/ssqosid.sail
jordancarlin: See https://github.com/riscv/sail-riscv/pull/379#discussion_r2637302230. Now that there is a standard way of handling Sstateen in #910, we can probably drop this.

jordancarlin: @ved-rivos Are you planning on making these final few updates? If not, one of us can push the last few changes so that we can get this merged.

ved-rivos: @jordancarlin - I do plan to make these updates but I am in the middle of setting my environment back up at my new workplace. So will need a couple of days to do that. If you could push this update then that would be great.

model/extensions/Ssqosid/ssqosid.sail
```diff
@@ -0,0 +1,38 @@
...
+    MCID = v[MCID] & mcid_mask
+  ]
+}
+
+register srmcfg : Srmcfg
+
+// Ssqosid - Quality-of-Service (Qos) Identifiers
+mapping clause csr_name_map = 0x181  <-> "srmcfg"
+
+function clause is_CSR_accessible(0x181, priv, _) = currentlyEnabled(Ext_Ssqosid)
```
jordancarlin:
```suggestion
function clause is_CSR_accessible(0x181, _, _) = currentlyEnabled(Ext_Ssqosid)
```

pmundkur: Now that Smstaten is merged via #910, could this PR add the `mstateen0.SRMCFG` access check as well (in `stateen_allows_CSR_access`)?

jordancarlin: @ved-rivos Seems like this was marked as resolved but the `stateen_allows_CSR_access` check was not added.

model/extensions/Stateen/stateen_regs.sail
```diff
@@ -16,11 +16,12 @@ function is_sstateen_accessible() -> bool = currentlyEnabled(Ext_S) & (currently
 termination_measure is_hstateen_accessible() = 5 // > Ext_H
 termination_measure is_sstateen_accessible() = 2 // > Ext_S
 
-enum stateen_bit = {STATEEN_FCSR, STATEEN_ENVCFG, STATEEN_SE}
+enum stateen_bit = {STATEEN_FCSR, STATEEN_SRMCFG, STATEEN_ENVCFG, STATEEN_SE}
 
 mapping stateen_bit_index : stateen_bit <-> range(0, 63) = {
   STATEEN_SE     <-> 63,
   STATEEN_ENVCFG <-> 62,
+  STATEEN_SRMCFG <-> 55,
```
nadime15, line 24: Thats unrelated to this PR but I just realized that bits 55 and 56 are defined only in mstateen0 and not in hstateen0. Currently, hstateen0 uses the same encoding as mstateen0. I think the correct behavior would be to legalize hstateen0 and set these bits to 1, otherwise, HS could write 0 to hstateen0.SRMCFG, which would prevent VS from accessing it

model/extensions/Stateen/stateen_access_checks.sail
```diff
@@ -48,6 +48,9 @@ function stateen_allows_CSR_access(csr : csreg, priv : Privilege, _access_type :
     // fcsr
     (0x003, priv) => if hartSupports(Ext_Zfinx) then check_stateen_bit(priv, STATEEN_FCSR, 0) else true,
 
+    // srmcfg
+    (0x181, priv) => check_stateen_bit(priv, STATEEN_SRMCFG, 0),
```
nadime15: Sorry, a minor issue, could you move line 52 so that it sits above line 44 and below line 42? This would follow the register encoding.

model/extensions/Stateen/stateen_regs.sail
nadime15: Or to define a new separate bit field for Hstateen0 which is probably cleaner

ved-rivos: This is just a mapper that maps strings to bit positions. The M/H/S state enable CSRs have their own legalization routines and the SRMCFG bit position is reserved in H state enable CSR. The state is enabled when it is set to 1 and is access disabled when it is set to 0. The srmcfg CSR is not defined to be accessible at VS level irrespective of the state enable i.e. even if enabled for HS it is not legal for VS. Further, the SRMCFG bit cannot be set in the hstateen0 CSR.

nadime15: Ah, I see. You are right, I was confused because senvcfg is accessible and writable in VS-mode, while srmcfg is not.

However the problem remains that hstateen0.SRMCFG (bit 55 and 56) is currently writable because hstateen0 is defined as mstateen0. This doesn’t cause a trap (since VS can’t access srmcfg), but it is still incorrect. If someone writes a bunch of 1s into hstateen0 bits 55 or 56 should be read back as 0.

ved-rivos: Using the same routing for legalizing hstateen0 and mstateen0 is a bug. Let me push an update to this PR to fix that.

nadime15: Awesome, thanks!

ved-rivos: hstateenX should be fixed now.

model/extensions/Stateen/stateen_regs.sail
```diff
@@ -67,17 +68,37 @@ function legalize_mstateen0(m : Mstateen0, v : bits(64)) -> Mstateen0 = {
     ENVCFG = if currentlyEnabled(Ext_S) then v[ENVCFG] else 0b0,
     FCSR   = if hartSupports(Ext_Zfinx) & (misa[F] == 0b0) then v[FCSR] else 0b0,
     C      = if config extensions.Stateen.C_readonly_zero then 0b0 else v[C],
+    SRMCFG = if currentlyEnabled(Ext_Ssqosid) then v[SRMCFG] else 0b0,
```
jordancarlin: Minor nit, but move `SRMCFG` up between `ENVCFG` and `FCSR` so that it follows the order of the bitfields.

jordancarlin (approved):
~~Seems like a pretty simple extension~~ (I just hit approve without a message, no idea why that message was filled in).

PR #381, opened 2024-01-02
Writable mtime 

defermelowie:
According to the priv. spec. `mtime` should be readable & writable:

> Platforms provide a real-time counter, exposed as a memory-mapped machine-mode read-write register, `mtime`.

However, the model currently does not allow writes to `mtime`. This pull requests modifies `clint_store` in order to support writes to the `mtime` memory-mapped register.

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -243,6 +243,24 @@ function clint_store(addr, width, data) = {
     mtimecmp = vector_update_subrange(mtimecmp, 63, 32, sail_zero_extend(data, 32)); /* FIXME: Redundant zero_extend currently required by Lem backend */
     clint_dispatch();
     MemValue(true)
+    } else if addr == MTIME_BASE & 'n == 8 then {
+    if   get_config_print_platform()
+    then print_platform("clint<8>[" ^ BitStr(addr) ^ "] <- " ^ BitStr(data) ^ " (mtime)");
+    mtime = zero_extend(data);
+    clint_dispatch();
+    MemValue(true)
+  } else if addr == MTIME_BASE & 'n == 4 then {
+    if   get_config_print_platform()
+    then print_platform("clint<4>[" ^ BitStr(addr) ^ "] <- " ^ BitStr(data) ^ " (mtime)");
+    mtime[31 .. 0] = zero_extend(data);
+    clint_dispatch();
+    MemValue(true)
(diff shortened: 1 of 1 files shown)


model/riscv_platform.sail
```diff
@@ -311,6 +311,24 @@ function clint_store(addr, width, data) = {
     mtimecmp = vector_update_subrange(mtimecmp, 63, 32, sail_zero_extend(data, 32)); /* FIXME: Redundant zero_extend currently required by Lem backend */
     clint_dispatch();
     MemValue(true)
+  } else if addr == MTIME_BASE & 'n == 8 then {
+    if   get_config_print_platform()
+    then print_platform("clint<8>[" ^ BitStr(addr) ^ "] <- " ^ BitStr(data) ^ " (mtime)");
+    mtime = zero_extend(data);
```
Timmmm, line 317: Do you actually need the `zero_extend()`s? In theory Sail could figure out that you don't... but maybe it isn't smart enough?

defermelowie: No, Sail is indeed able to figure this out. I removed them

model/riscv_platform.sail
```diff
@@ -243,6 +243,24 @@ function clint_store(addr, width, data) = {
     mtimecmp = vector_update_subrange(mtimecmp, 63, 32, sail_zero_extend(data, 32)); /* FIXME: Redundant zero_extend currently required by Lem backend */
     clint_dispatch();
     MemValue(true)
+  } else if addr == MTIME_BASE & 'n == 8 then {
+    if   get_config_print_platform()
+    then print_platform("clint<8>[" ^ BitStr(addr) ^ "] <- " ^ BitStr(data) ^ " (mtime)");
+    mtime = data;
```
arichardson, line 249: If this zero_extend is no longer required, can we also remove it from the other cases? Not necessarily as part of this PR, but would be good to clean up the FIXME.

arichardson (approved):

model/riscv_platform.sail
Timmmm: Yeah I don't think `vector_update_subrange` is needed either. Can go in a different MR though.

Timmmm (approved):

Timmmm:
I'll merge this in a few days if nobody objects.

PR #384, opened 2024-01-07
Add Zaamo and Zalrsc

ved-rivos:
Specification: https://github.com/ved-rivos/riscv-zaamo-zalrsc

### DIFF
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -43,8 +43,8 @@ function amo_width_valid(size : word_width) -> bool = {
 /* ****************************************************************** */
 union clause ast = LOADRES : (bool, bool, regidx, word_width, regidx)
 
-mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)                                                if haveAtomics() & amo_width_valid(size)
-  <-> 0b00010 @ bool_bits(aq) @ bool_bits(rl) @ 0b00000 @ rs1 @ 0b0 @ size_enc(size) @ rd @ 0b0101111 if haveAtomics() & amo_width_valid(size)
+mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)                                                if haveZalrsc() & amo_width_valid(size)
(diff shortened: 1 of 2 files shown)


rpsene:
@billmcspadden-riscv this spec was also [recently ratified](https://jira.riscv.org/browse/RVS-1995). Can this be merged?

jrtc27:
What is the official stance on A vs Za*? LLVM doesn’t believe that A implies Zalrsc or Zaamo, yet this PR says it does.

Alasdair:
I think we merged a PR recently that moved the guards to be more in line with other files, so I think this PR needs to be rebased for that.

Other than that I don't see any issues here. Presumably at some point we will have to change the `haveZalrsc` and `haveZaamo` implementations to allow toggling them individually.

ved-rivos:
Ideally PRs like the one the one to refactor the *aext.sail would have been done after applying the related PRs like a) for Zabha, b) Zacas, c) Zaamo, d) Zalrsc that have been queued for a while now. If these had been applied before the refactoring then the refactoring could have more comprehensively addressed these ratified extensions. I will try to get these four reworked/rebased and retested soon.

Timmmm:
Open PRs aren't actually a queue though. This one looks very easy to rebase in any case.

ved-rivos:
> Open PRs aren't actually a queue though. 

Open PRs are not a queue but rewriting upstream for aesthetic reasons when there are queued PRs is not kind.

Alasdair (approved):

Timmmm:
> Open PRs are not a queue but rewriting upstream for aesthetic reasons when there are queued PRs is not kind.

It wasn't entirely aesthetic, and I don't think we should pause refactoring if any open PR conflicts, but in any case it wasn't my intention to be unkind or cause unnecessary rebasing (which is a right pain indeed!). Sorry about that.

I also wasn't aware of this PR tbf - would be nice if Github could highlight conflicting PRs automatically!

Anyway this LGTM and seems to be entirely uncontroversial so I'll merge it tomorrow if nobody has any objections...

Timmmm (approved):

arichardson (approved):

PR #385, opened 2024-01-07
Add B extension

ved-rivos:
Specification: https://github.com/riscv/riscv-b

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -105,6 +105,9 @@ val sys_pmp_grain = {c: "sys_pmp_grain", ocaml: "Platform.pmp_grain", _: "sys_pm
 /* whether misa.v was enabled at boot */
 val sys_enable_vext = {c: "sys_enable_vext", ocaml: "Platform.enable_vext", _: "sys_enable_vext"} : unit -> bool
 
+/* whether misa.b was enabled at boot */
+val sys_enable_bext = {c: "sys_enable_bext", ocaml: "Platform.enable_bext", _: "sys_enable_bext"} : unit -> bool
+
 /* This function allows an extension to veto a write to Misa
    if it would violate an alignment restriction on
    unsetting C. If it returns true the write will have no effect. */
@@ -136,10 +139,11 @@ function haveNExt()    -> bool = misa[N] == 0b1
 /* see below for F and D (and Z*inx counterparts) extension tests */
 
 /* BitManip extension support. */
-function haveZba()  -> bool = true
-function haveZbb()  -> bool = true
+function haveBExt() -> bool = misa[B] == 0b1
+function haveZba()  -> bool = true | haveBExt()
+function haveZbb()  -> bool = true | haveBExt()
+function haveZbs()  -> bool = true | haveBExt()
 function haveZbc()  -> bool = true
-function haveZbs()  -> bool = true
 
 /* Zfa (additional FP) extension */
 function haveZfa()  -> bool = true
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -57,6 +57,11 @@ bool sys_enable_vext(unit u)
   return rv_enable_vext;
 }
 
+bool sys_enable_bext(unit u)
+{
+  return rv_enable_bext;
+}
+
 uint64_t sys_pmp_count(unit u)
 {
   return rv_pmp_count;
--- a/c_emulator/riscv_platform.h
+++ b/c_emulator/riscv_platform.h
@@ -10,6 +10,7 @@ bool sys_enable_zfinx(unit);
 bool sys_enable_writable_misa(unit);
 bool sys_enable_writable_fiom(unit);
 bool sys_enable_vext(unit);
+bool sys_enable_bext(unit);
 
 uint64_t sys_pmp_count(unit);
 uint64_t sys_pmp_grain(unit);
--- a/c_emulator/riscv_platform_impl.c
+++ b/c_emulator/riscv_platform_impl.c
@@ -14,6 +14,7 @@ bool rv_enable_next = false;
 bool rv_enable_writable_misa = true;
 bool rv_enable_fdext = true;
 bool rv_enable_vext = true;
+bool rv_enable_bext = false;
 
 bool rv_enable_dirty_update = false;
 bool rv_enable_misaligned = false;
--- a/c_emulator/riscv_platform_impl.h
+++ b/c_emulator/riscv_platform_impl.h
@@ -18,6 +18,7 @@ extern bool rv_enable_rvc;
 extern bool rv_enable_next;
 extern bool rv_enable_fdext;
 extern bool rv_enable_vext;
+extern bool rv_enable_bext;
 extern bool rv_enable_writable_misa;
 extern bool rv_enable_dirty_update;
 extern bool rv_enable_misaligned;
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -248,6 +248,7 @@ static int process_args(int argc, char **argv)
   while (true) {
     c = getopt_long(argc, argv,
                     "a"
+                    "B"
                     "d"
                     "m"
                     "P"
@@ -281,6 +282,10 @@ static int process_args(int argc, char **argv)
     switch (c) {
     case 'a':
       report_arch();
+      break;
+    case 'B':
+      fprintf(stderr, "enabling B extension.\n");
+      rv_enable_bext = true;
       break;
     case 'd':
       fprintf(stderr, "enabling dirty update.\n");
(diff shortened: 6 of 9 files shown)


model/riscv_insts_zba.sail
```diff
@@ -71,8 +71,8 @@
 /* ****************************************************************** */
 union clause ast = RISCV_SLLIUW : (bits(6), regidx, regidx)
 
-mapping clause encdec = RISCV_SLLIUW(shamt, rs1, rd) if haveZba() & sizeof(xlen) == 64
-  <-> 0b000010 @ shamt @ rs1 @ 0b001 @ rd @ 0b0011011 if haveZba() & sizeof(xlen) == 64
+mapping clause encdec = RISCV_SLLIUW(shamt, rs1, rd) if (haveZba() | haveBExt()) & sizeof(xlen) == 64
```
jrtc27, line 74: B implies Zba, so this is just an ISA parsing (which doesn't really exist) issue? This would be if Zba and B were independent extensions that both implement the same instructions (like M vs Zmmul).

ved-rivos: The proposed B extension include instructions provided by Zba, Zbb, and Zbs extensions.

jrtc27: My reading of the spec is that B *is* Zba_Zbb_Zbs, not that it's a *separate* extension. And I don't see why you'd want it to be anything else?

Like how G is IMAFD; we just say "has F" in the Sail model, since G implies F, rather than putting "has F or G" everywhere.

ved-rivos: Yes include Zba, Zbb, Zbs so B=Zba+Zbb+Zbs. However, if B is not present i.e. misa.B is 0, then one or more of Zba, Zbb, or Zbs may or may not be present.

jrtc27: Yes, I know. The point is, "has B" and "has B or Zba" are entirely equivalent, and it's the job of a centralised ISA parser to turn B into B_Zba_Zbb_Zbs, just like a toolchain does. Not the job of every single instruction to list every single extension that (transitively) provides that instruction. That would get very out of hand very quickly, and isn't helpful.

ved-rivos: So you would prefer if misa.B was set to 1 then haveZba be also set to 1?

jrtc27: Yes, because that's what B actually means

jrtc27: (Though note that haveZba is already hard-coded to true, so there's nothing to do there; in future the Sail model will have a proper ISA string parser, and then it'll have to include all the same rules toolchains have)

ved-rivos: But that not allow expressing behavior of clearing B clearly. For instance, if misa.B and haveZba was both 1 at reset. Then clearing B would leave Zba available. But if misa.B was 1 and haveZba was 0 at reset then clearing B would disable Zba. Forcing state of Zba to misa.B would not allow expressing this.

allenjbaum:
Yes, that's why Zmmul and M can both be set. Clearing misa.M is then just
removing the DIV/REM instructions but not the MUL instructions.
I don't know why anyone would ever do that, but mine is not to reason
why....

model/riscv_insts_zba.sail
ved-rivos: I like the idea that @jrtc27 brought up to defer this to when we have isa string parsing implemented in sail. I will revise this PR to only add the misa.B and if misa.B is 1 set Zba/Zbb/Zbs as supported. For now clearing misa.B will keep Zba/Zbb/Zbs as still available and that option for them to go away can be done when there is ISA string support.

model/riscv_sys_regs.sail
```diff
@@ -187,10 +188,10 @@ function haveNExt()    -> bool = misa.N() == 0b1
 /* see below for F and D (and Z*inx counterparts) extension tests */
 
 /* BitManip extension support. */
-function haveZba()  -> bool = true
-function haveZbb()  -> bool = true
+function haveZba()  -> bool = haveBExt()
```
jrtc27, line 191: These should all remain true

ved-rivos: right. That would be backward compatible change without the isa string parser. Updated.

rpsene:
@billmcspadden-riscv this spec was [recently ratified](https://jira.riscv.org/browse/RVS-2006), can this be merged?

Alasdair:
Does the `haveBExt` function this defines not need to be used somewhere?

ved-rivos:
Please see the conversation thread in this PR.

model/riscv_insts_zba.sail
ved-rivos: That would be incorrect. If `misa.B` is 1 then it implies that `Zba`, `Zbb`, and `Zbs` are present. However, if `misa.B` is 0, it does not imply that `Zba`, `Zbb`, or `Zbs` are not present.

ved-rivos: I have removed the haveBExt() - it can be added back if needed later.

Timmmm: Ah good point, I see what you mean. So would it make sense to do this?

```
misa[B] = bool_to_bits(haveZba() & haveZbb() & haveZbs());
```

That would ensure the bit is set correctly if someone edits `haveZb?()` to return `false`.

ved-rivos: That would still have the issue that having `Zba`, `Zbb`, and `Zbs` does not imply `B`.  It  is valid for `B` to be 0 and  have *all* or some of `Zba`, `Zbb`, and `Zbs`. I am not sure we  need to guard againt people modifying code  in wrong ways - they can do that in many other ways. When we have ISA string parsing then one  would be able to specify "B" in ISA string in which case they do not need to explicitly add `Zba`, `Zbb`, and `Zbs` to the ISA string and `misa.B` becomes 1. Or they can choose to provide any combination of `Zba`, `Zbb`, and `Zbs`  and `misa.B`  stays  0. Or they can choose to provide `B`  and combination of `Zba`, `Zbb`, and `Zbs`  in which case the ones not  specified are forced to 1 because of `B` and `misa.B` is 1.

Timmmm: > I am not sure we need to guard againt people modifying code in wrong ways - they can do that in many other ways.

Fully disagree here. The way this bit works is clearly quite confusing and I don't think it's unlikely that people will change these hard-coded `haveExt()` functions. Anyone using the model for verification will have to do it (we do). And when they do it would be nice for it to prevent B being set incorrectly.

I'm happy for it to be an assertion instead if you like? `assert(misa[B] == 0b0 | have.. & have... & have...)`. Maybe that's better?

Do we need to handle writing this bit for writable misa?

ved-rivos: > I'm happy for it to be an assertion instead if you like? assert(misa[B] == 0b0 | have.. & have... & have...). Maybe that's better?

Yeah an  assertion could be added. I can  see need to for many such we would add to adress people modifying code in  ways that does   not  make sense.

> Do we need to handle writing this bit for writable misa?

When misa is writeable this bit can be written. Clearing  `misa.B`  does not  imply `Zba`, `Zbb`, or `Zbs` go away.

ved-rivos: I made an update that should make it more clearer and also guard against people adding bugs when changing code.

Timmmm: Yeah this looks much better, thanks!

> When misa is writeable this bit can be written.

Even if the extensions are not supported? I would have thought that makes it read-only zero.

To be fair I think all of this code needs some clean up, e.g. it doesn't look like the A bit isn't handled properly at all either, so I think this is good enough to merge now.

Timmmm: Actually, re-reading the code I now understand how the `have...` and `sys...` functions are supposed to work, so I think the fully correct thing is:

```
// Indicates whether the B extension is *supported* (not enabled).
// TODO: Add a CLI parameter.
function sys_enable_b() -> bool = true

...

misa[B] = bool_to_bits(sys_enable_b()); /* Bit-manipulation */

...

function haveBExt() -> bool = misa[B] == 0b1
// These sub-extensions can be enabled even if misa[B] is 0, and currently we do that.
// TODO: Make this configurable.
function haveZba()  -> bool = true | haveBExt()
function haveZbb()  -> bool = true | haveBExt()
function haveZbs()  -> bool = true | haveBExt()
```

I think the naming and organisation of these functions is a bit misleading so I made an issue to fix it: #495

Sorry for all the changes!

ved-rivos: updated.

Timmmm (approved):
LGTM, thanks! @jrtc27 does this look ok to you?

Timmmm:
I'll merge this in a couple of days if nobody objects. @ved-rivos what would you say your highest priority PR is? I think it would be good to get some more merged if possible.

ved-rivos:
@Timmmm - dont really have a priority/preference - maybe this one #393 ?

PR #386, opened 2024-01-08
Updates for Coq output

bacam:
This updates the Coq "extras" file and the Coq output snapshot for the current model.  Note that to build the Coq output from scratch you need a recent development version of Sail that lifts some restrictions that the vector extension runs into.


bacam:
Superseded by #695.

PR #397, opened 2024-01-27
Add missing mstatus.MPP legalization

ved-rivos:
`mstatus.MPP` legal values are User (`00) (if U-mode implemented), Supervisor (`01) (if S-mode implemented) and Machine (`11). Encoding `10 is illegal.

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -222,6 +222,21 @@ function haveZmmul()  -> bool = true
 /* Zicond extension support */
 function haveZicond() -> bool = true
 
+/*
+ * Illegal values legalized to least privileged mode supported.
+ * Note: the only valid combinations of supported modes are M, M+U, M+S+U.
+ */
+function lowest_supported_privLevel() -> Privilege =
+  if haveUsrMode() then User else Machine
+
+function have_privLevel(priv : priv_level) -> bool =
+  match priv {
+    0b00 => haveUsrMode(),
+    0b01 => haveSupMode(),
+    0b10 => false,
+    0b11 => true,
+  }
+
 bitfield Mstatush : bits(32) = {
   MBE  : 5,
   SBE  : 4
@@ -308,6 +323,10 @@ function legalize_mstatus(o : Mstatus, v : xlenbits) -> Mstatus = {
    * explicitly later.
    */
   let m : Mstatus = Mk_Mstatus(zero_extend(v[22 .. 7] @ 0b0 @ v[5 .. 3] @ 0b0 @ v[1 .. 0]));
+
+  /* Legalize MPP */
+  let m = [m with MPP = if have_privLevel(m[MPP]) then m[MPP]
+                        else privLevel_to_bits(lowest_supported_privLevel())];
 
   /* We don't have any extension context yet. */
   let m = [m with XS = extStatus_to_bits(Off)];
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -561,6 +561,7 @@ function init_sys() -> unit = {
   mstatus = set_mstatus_SXL(mstatus, misa[MXL]);
   mstatus = set_mstatus_UXL(mstatus, misa[MXL]);
   mstatus[SD]   = 0b0;
+  mstatus[MPP] = privLevel_to_bits(lowest_supported_privLevel());
 
   /* set to little-endian mode */
   if sizeof(xlen) == 64 then {


model/riscv_sys_regs.sail
```diff
@@ -215,6 +215,14 @@ function haveZmmul()  -> bool = true
 /* Zicond extension support */
 function haveZicond() -> bool = true
 
+function legalize_priv_level (p : priv_level) -> priv_level =
+  match (haveUsrMode(), p) {
+    (false,   _) => privLevel_to_bits(Machine),
+    (true, 0b01) => if haveSupMode() then p else privLevel_to_bits(User),
```
arichardson, line 221: Is there a specific reason this is legalized to User rather than Machine?

If not this could be simplified to always fall back to Machine if the value is not supported.

model/riscv_sys_regs.sail
```diff
@@ -302,6 +310,9 @@ function legalize_mstatus(o : Mstatus, v : xlenbits) -> Mstatus = {
    */
   let m : Mstatus = Mk_Mstatus(zero_extend(v[22 .. 7] @ 0b0 @ v[5 .. 3] @ 0b0 @ v[1 .. 0]));
 
+  /* Legalize MPP */
+  let m = update_MPP(m, legalize_priv_level(m.MPP()));
```
arichardson, line 314:
```suggestion
  let m = [m with MPP = legalize_priv_level(m[MPP])];
```

model/riscv_sys_regs.sail
arichardson: Agreed choosing lower privilege makes sense - I'd just like to see a comment along the lines of "we are free to choose the legalized privileged, but choosing the lowest valid privilege level is the safest choice".

model/riscv_sys_regs.sail
```diff
@@ -215,6 +215,14 @@ function haveZmmul()  -> bool = true
 /* Zicond extension support */
 function haveZicond() -> bool = true
 
+function legalize_priv_level (p : priv_level) -> priv_level =
+  match (haveUsrMode(), p) {
+    (false,   _) => privLevel_to_bits(Machine),
```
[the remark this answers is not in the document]
ved-rivos: The only supported combinations of privilege modes are M, M+U, M+S+U (Table 2; Priv) so configuring the model with M+S would illegal. Such an assertion may want to be placed in the ISA parser or knobs parser to not create a misconfigured model.

model/riscv_sys_regs.sail
```diff
@@ -215,6 +215,14 @@ function haveZmmul()  -> bool = true
...
+    (true, 0b01) => if haveSupMode() then p else privLevel_to_bits(User),
+    (true, 0b10) => privLevel_to_bits(User),
```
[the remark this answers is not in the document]
ved-rivos: For M+U, the likely simplest legalization would be have a single flop for privilege that is updated with the and of the two bits. Besides its a safer default. For M only, MPP would be hardwired to `11. For M+S+U the likely simplest legalization to update the MSB of the priv level to be the AND of the two bits. Besides it leads to a safer legal value.

model/riscv_sys_regs.sail
ved-rivos: This is nicer. Had not seen it used elsewhere. Is this new supported syntax?

arichardson: Yes I was also not aware of it until #373 but apparently it's been supported for quite a long time.

model/riscv_sys_regs.sail
```diff
@@ -215,6 +215,14 @@ function haveZmmul()  -> bool = true
...
+    (true,    _) => p
+  }
```
arichardson, line 224:
```suggestion
  /* 
   * When a mode value is not supported, we legalize it to the lowest
   * supported mode since that is the safest choice.
   * Note: the only valid combinations of supported modes are M, M+U, M+S+U.
   */
  match p {
    0b00 => if haveUsrMode() then p else privLevel_to_bits(Machine),
    0b01 => if haveSupMode() then p else privLevel_to_bits(User),
    0b10 => privLevel_to_bits(if haveUsrMode() then User else Supervisor),
    0b11 => p
  }
```
How about something like this, so we don't need to match a tuple?

ved-rivos: For 0b10 - this code will lead to picking Supervisor on a M only machine... It could be fixed by - `0b10 => privLevel_to_bits(if haveUsrMode() then User else Machine)`

For 0b01 - this code will lead to picking User on a M only machine...it could be fixed by - `0b01 => if haveSupMode() then p else if haveUsrMode() then User else Machine,`

ved-rivos: Yes this would work. Is this preferred? Tuples are avoided because they are slow?

arichardson: I don't think that should matter - I just found the tuple version slightly more difficult to parse in my head. But that could just be a me problem :) 

I'll defer to @jrtc27 or @Alasdair for what the preferred style would be.

model/riscv_sys_regs.sail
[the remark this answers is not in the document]
ved-rivos: added a comment to the function.

nwf-msr:
Something also needs to happen at initialization, I think?  Right now https://github.com/riscv/sail-riscv/blob/d5e89a71e3a84495c1b88a7749c25fd6b9da684b/model/riscv_sys_control.sail#L553 leaves `mstatus[MPP]` as `0b00`, even if that's not a legal value (say, on a M-mode-only core).

ved-rivos:
> Something also needs to happen at initialization, I think? Right now
> 
> https://github.com/riscv/sail-riscv/blob/d5e89a71e3a84495c1b88a7749c25fd6b9da684b/model/riscv_sys_control.sail#L553
> 
> leaves `mstatus[MPP]` as `0b00`, even if that's not a legal value (say, on a M-mode-only core).

Agree. This was missed. I have committed an update to fix this.

arichardson (approved):
LGTM. I slightly prefer using an `if/else` instead of tuple match (since the first element is only needed for one case), but happy either way.
```
  if not(haveUsrMode()) then privLevel_to_bits(Machine)
  else match p {
    0b01 => if haveSupMode() then p else privLevel_to_bits(User),
    0b10 => privLevel_to_bits(if haveSupMode() then Supervisor else Machine),
    _ => p
  }
```

ved-rivos:
> We also came across this, so... alternative implementation, which I think is clearer:

This is nicer. Updated to this.

arichardson (approved):

Timmmm (approved):
LGTM, thanks!

Timmmm:
@billmcspadden-riscv We should merge this fix IMO. It's been a few months and there are no issues with it.

Timmmm:
I'll merge this next week if nobody objects.

PR #398, opened 2024-01-29
Add Zawrs extension

ved-rivos:
Spec: https://github.com/riscv/riscv-zawrs
Test: https://github.com/riscv-non-isa/riscv-arch-test/pull/432

### DIFF
--- a/c_emulator/riscv_platform.h
+++ b/c_emulator/riscv_platform.h
@@ -42,3 +42,6 @@ unit plat_term_write(mach_bits);
 mach_bits plat_htif_tohost(unit);
 
 unit memea(mach_bits, sail_int);
+
+/* Zawrs : Wait-on-Reservation-Set */
+bool reservation_set_valid(unit);
--- a/model/riscv_insts_zawrs.sail
+++ b/model/riscv_insts_zawrs.sail
@@ -1,0 +1,57 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+/* ****************************************************************** */
+/* This file specifies the instructions in the 'Zawrs' extension.  */
+
+val reservation_set_valid = {
+  ocaml:       "Platform.reservation_set_valid",
+  interpreter: "Platform.reservation_set_valid",
+  c:           "reservation_set_valid",
+  lem:         "reservation_set_valid"
+  } : unit -> bool
+
+/* ****************************************************************** */
+union clause ast = WRS : (wrsop)
+
+mapping clause encdec = WRS(WRS_STO) if haveZawrs()
+  <-> 0b000000011101 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011 if haveZawrs()
+
+mapping clause encdec = WRS(WRS_NTO) if haveZawrs()
+  <-> 0b000000001101 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011 if haveZawrs()
+
+/* Hart must resume if a locally enabled interrupt becomes pending, even if delegated */
+function interrupts_pending() -> bool = ((mip.bits() & mie.bits()) != zero_extend(0b0))
+
+function clause execute (WRS(op)) = {
+
+  platform_start_wrs_timeout((op == WRS_STO));
+
+  while reservation_set_valid() & not(interrupts_pending()) & not(platform_is_wrs_timed_out()) do ();
+
+  platform_stop_wrs_timeout();
+
+  let timed_out = reservation_set_valid() & not(interrupts_pending());
+
+  match (op, timed_out) {
+    (_,       false) => RETIRE_SUCCESS,
+    (WRS_STO, true)  => RETIRE_SUCCESS,
+    (WRS_NTO, true)  => match cur_privilege {
+      Machine => { RETIRE_SUCCESS },
+      _       => if mstatus.TW() == 0b1
+                 then { handle_illegal(); RETIRE_FAIL }
+                 else { RETIRE_SUCCESS }
+      }
+   }
+}
+
+mapping wrs_mnemonic : wrsop <-> string = {
+  WRS_STO <-> "wrs.sto",
+  WRS_NTO <-> "wrs.nto",
+}
+mapping clause assembly = WRS(op) <-> wrs_mnemonic(op)
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ SAIL_DEFAULT_INST += riscv_insts_zbkb.sail
 SAIL_DEFAULT_INST += riscv_insts_zbkx.sail
 
 SAIL_DEFAULT_INST += riscv_insts_zicond.sail
+SAIL_DEFAULT_INST += riscv_insts_zawrs.sail
 
 SAIL_DEFAULT_INST += riscv_insts_vext_utils.sail
 SAIL_DEFAULT_INST += riscv_insts_vext_vset.sail
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -176,3 +176,9 @@ unit memea(mach_bits len, sail_int n)
 {
   return UNIT;
 }
+
+/* Zawrs : Wait-on-Reservation-Set */
+bool reservation_set_valid(unit u)
+{
+  return reservation_valid;
+}
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -454,3 +454,15 @@ function platform_wfi() -> unit = {
     mcycle = mtimecmp;
   }
 }
+
+/* Start implementation defined short timeout if short is true
+ * For no timeout, start a implementation defined timeout if privilege
+ * level is less than M.
+ */
+function platform_start_wrs_timeout(short : bool) -> unit = ()
+
+/* Stop any timer started by WRS */
+function platform_stop_wrs_timeout() -> unit =  ()
+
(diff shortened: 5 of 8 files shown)


c_emulator/riscv_platform.h
```diff
@@ -38,3 +38,6 @@ unit plat_term_write(mach_bits);
 mach_bits plat_htif_tohost(unit);
 
 unit memea(mach_bits, sail_int);
+
+// Zawrs : Wait-on-Reservation-Set
```
arichardson, line 42: I think we should be consistent and either always use C++ style // comments or always C-style

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,118 @@
...
+
+mapping clause encdec = WRS(WRS_NTO) if haveZawrs()
+  <-> 0b000000001101 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011 if haveZawrs()
+
+/* Hart must resume if a locally enabled interrupt becomes pending, even if delegated */
+function interrupts_pending() -> bool = ((mip.bits() & mie.bits()) != zero_extend(0b0))
+
+function clause execute (WRS(op)) = {
+
+  platform_start_wrs_timeout(if op == WRS_STO then true else false);
```
Timmmm, line 96: This is equivalent to `op == WRS_STO`.

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,118 @@
...
+
+  while reservation_set_valid() & not(interrupts_pending()) & not(platform_is_wrs_timed_out()) do ();
```
Timmmm, line 98: I'm not sure if the best way to handle this is via waiting in the actual execute function. WFI doesn't work like this (to be fair it doesn't work at all because there's no proper interface for injecting interrupts currently).

Maybe the model should be a state machine and the `step()` function handled the WFI and WRS states explicitly?

I dunno, we probably need to think about this properly. Maybe this is fine for now though since the platform callbacks are just hard-coded to time out immediately.

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,118 @@
...
+
+  platform_stop_wrs_timeout();
+
+  let timed_out : bool = if not(reservation_set_valid()) | interrupts_pending() then false else true;
```
Timmmm, line 102: This can be

```
let timed_out = reservation_set_valid() & not(interrupts_pending());
```

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,118 @@
...
+/* Hart must resume if a locally enabled interrupt becomes pending, even if delegated */
+function interrupts_pending() -> bool = ((mip.bits() & mie.bits()) != zero_extend(0b0))
+
+function clause execute (WRS(op)) = {
+
+  platform_start_wrs_timeout(if op == WRS_STO then true else false);
+
+  while reservation_set_valid() & not(interrupts_pending()) & not(platform_is_wrs_timed_out()) do ();
+
+  platform_stop_wrs_timeout();
```
Timmmm, line 100: I think this is probably not necessary (and doesn't do anything for now anyway).

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,118 @@
...
+
+  let timed_out : bool = if not(reservation_set_valid()) | interrupts_pending() then false else true;
+
+  match (op, timed_out) {
```
Timmmm, line 104: You can put this all in one condition. I think this is a bit clearer.

```
if op == WRS_NTO & timed_out & cur_privilege != Machine & mstatus[TW] == 0b1 then {
  handle_illegal(); RETIRE_FAIL
} else RETIRE_SUCCESS
```

Timmmm (changes requested):

model/riscv_insts_zawrs.sail
ved-rivos: The intent here is two folds:
1. As a precise documentation of the ISA - this shows that the execution stalls while the instruction waits for these three conditions a) a snoop that invalidates the reservation set due to a store from another hart or a device b) an interrupt becomes pending c) the wait times out.
2. Allows the platform test bench to inject these into the mode - a snoop invalidation or the platform specific timer timeout. This would also include advancing cycles and time.

We cannot `step()` as it is presently as we cannot fetch or execute instructions when stalled.

model/riscv_insts_zawrs.sail
ved-rivos: This is necessary to document the ISA behavior that if the stall was terminated due to an interrupt or the reservation set getting invalidated then the timer is stopped.

model/riscv_insts_zawrs.sail
ved-rivos: Its much more explicitly readable documentation of the ISA. This allows one who wants to understand how the ISA is defined to see the conditions clearly spelt out as a truth table  and does not need them to do mental math to evaluate the equation to reconstruct the truth table.

Alasdair: I'm fine with the longer version you have but the `let timed_out : bool =` could be:
```
let timed_out : bool = reservation_set_valid() & not(interrupts_pending())
```
i.e. just rewriting `if not(A) | B  then false else true` to `A & not(B)`.

ved-rivos: updated.

model/riscv_insts_zawrs.sail
ved-rivos: updated.

model/riscv_insts_zawrs.sail
ved-rivos: updated.

model/riscv_insts_zawrs.sail
Timmmm: I mean the `platform_stop_wrs_timeout()`. I guess it doesn't hurt to leave it in if you want.

model/riscv_insts_zawrs.sail
Timmmm: > Its much more explicitly readable documentation of the ISA.

IMO my version is more readable and wasn't meant to be "clever". I feel like the `match` version requires careful decoding to understand the actual conditions... But I'm happy to be overruled if people disagree.

You can omit the `: bool` btw. I don't think it adds anything here.

model/riscv_insts_zawrs.sail
ved-rivos: Yes, I meant the same function. It a) documents when the timer stops b) provides a hook into the platform so the test bench can stop the timer.

model/riscv_insts_zawrs.sail
Timmmm: Yeah I agree it is nice to document the stalls like this. But documentation isn't the only purpose of the model. I think we want to be a bit careful about this because this is the first bit of code in the model that assumes that `step()` can block.

> We cannot step() as it is presently as we cannot fetch or execute instructions when stalled.

My feeling is that `step()` should really mean "step the model state", not "fetch and execute an instruction". At the moment those are exactly the same because none of the execute functions call platform callbacks to wait for an event, including WFI. If `step()` blocks you need threads in lots of situations (e.g. multiple harts, cosimulation), which are quite a pain in C++ and even worse in C.

But it's true that turning `step()` into an explicit state machine would make the code less clear. Hmm.

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,118 @@
...
+/*  PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR                  */
+/*  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,                         */
+/*  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT                     */
+/*  LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF                     */
+/*  USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND                  */
+/*  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,                   */
+/*  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT                   */
+/*  OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF                   */
+/*  SUCH DAMAGE.                                                                         */
+/*=======================================================================================*/
```
Timmmm, line 70: Oh btw we shortened this copyright header - can you copy the new one from another file?

model/riscv_insts_zawrs.sail
ved-rivos: I feel the match construct is more explicit as it matches the syntax of a truth table for the various conditions and each row gives the outcome.

model/riscv_insts_zawrs.sail
ved-rivos: Thats why I wrote "as it is presently". I dont think it would make the code less clear as a state machine. We would need instead of RETIRE_SUCCESS, something like STALL_UNTIL and the code should setup the conditions for "until". The model would then want the execution to resume after the STALL_UNTIL - STALL_UNTIL then becomes sort of like a `swapcontext()`.

model/riscv_insts_zawrs.sail
ved-rivos: bool removed.

model/riscv_insts_zawrs.sail
ved-rivos: done

model/riscv_insts_zawrs.sail
ved-rivos: > there's no proper interface for injecting interrupts currently

#395 provides timer interrupts.

ved-rivos:
closing due to #777

PR #404, opened 2024-02-07
Add support for HPM counters

Timmmm:
Implements Hardware Performance Monitoring counters, but only by making the registers available. Their values only change if explicitly written to and all values are legal.

All 32 HPM counters are always implemented (the spec says this "should" be the case), and a platform callback `sys_writable_hpm_counters` is added to allow making any of them read-only. I haven't hooked `sys_writable_hpm_counters` up to CLI flags because Codasip does not use the CLI flags and it is quite tedious, and hopefully going to change when riscv-config is used.

Finally this adds `menvcfg` and `senvcfg` to the CSR name map which I forgot to do previously.


model/riscv_csr_map.sail
```diff
@@ -113,6 +172,8 @@ mapping clause csr_name_map = 0x143  <-> "stval"
 mapping clause csr_name_map = 0x144  <-> "sip"
 /* supervisor protection and translation */
 mapping clause csr_name_map = 0x180  <-> "satp"
+/* supervisor envcfg */
+mapping clause csr_name_map = 0x10A  <-> "senvcfg"
```
arichardson, line 176: This is also included in #395, maybe makes sense to submit this as a tiny bugfix PR that can hopefully be merged in a reasonable timeframe.

arichardson (commented):
"High Performance Counters" should be "Hardware Performance Counters"?

Also do we need to guard these with a (always-true) `haveZihpm()` function?

model/riscv_csr_map.sail
Timmmm: Good idea, I'll do that.

Timmmm:
> "High Performance Counters" should be "Hardware Performance Counters"?
> 
> Also do we need to guard these with a (always-true) `haveZihpm()` function?

Good call, I'll add that too.

model/riscv_csr_map.sail
Timmmm: In #405 ; I'll remove it from this.

Timmmm: Done. I've also added `haveZihpm()`.

c_emulator/riscv_platform.h
```diff
@@ -8,6 +8,7 @@ bool sys_enable_zfinx(unit);
 bool sys_enable_writable_misa(unit);
 bool sys_enable_writable_fiom(unit);
 bool sys_enable_vext(unit);
+mach_bits sys_writable_hpm_counters(unit u);
```
arichardson, line 11: Unless I'm missing something in the spec I don't think this is needed according to privspec 3.1.10. "Hardware Performance Monitor"
> The counter registers have an arbitrary value after the hart is reset, and can be written with a given value. Any CSR write takes effect after the writing instruction has otherwise completed.

model/prelude.sail
```diff
@@ -178,22 +178,31 @@ val to_bits : forall 'l, 'l >= 0.(atom('l), int) -> bits('l)
 function to_bits (l, n) = get_slice_int(l, n, 0)
 
 infix 4 <_s
+infix 4 >_s
```
arichardson, line 181: This should probably be a separate commit?

model/riscv_insts_zicsr.sail
```diff
@@ -103,6 +103,9 @@ function readCSR csr : csreg -> xlenbits = {
     (0x31A, 32) => menvcfg.bits[63 .. 32],
     (0x320,  _) => zero_extend(mcountinhibit.bits),
 
+    /* Hardware Performance Monitoring event selection */
+    (0b0011001 /* 0x320 */ @ index : bits(5), _) if index >_u 0b00010 => read_mhpmevent(hpmidx_from_bits(index)),
```
arichardson, line 107:
```suggestion
    (0b0011001 /* 0x320 */ @ index : bits(5), _) if unsigned(index) > 2 => read_mhpmevent(hpmidx_from_bits(index)),
```
I guess this would avoid the need for the new overloads? Not sure it's any easier to read though

model/riscv_sys_control.sail
```diff
@@ -143,10 +151,14 @@ function is_CSR_defined (csr : csreg, p : Privilege) -> bool =
     0xC01 => haveUsrMode(),    // time
     0xC02 => haveUsrMode(),    // instret
 
+    0b1100000 /* 0xC00 */ @ index : bits(5) if index >_u 0b00010 => haveZihpm() & haveUsrMode(), // mhpmcounter3..31
```
arichardson, line 154:
```suggestion
    0b1100000 /* 0xC00 */ @ index : bits(5) if index >_u 0b00010 => haveZihpm() & haveUsrMode(), // hpmcounter3..31
```

model/riscv_sys_control.sail
```diff
@@ -143,10 +151,14 @@ function is_CSR_defined (csr : csreg, p : Privilege) -> bool =
...
+
     0xC80 => haveUsrMode() & (sizeof(xlen) == 32),     // cycleh
     0xC81 => haveUsrMode() & (sizeof(xlen) == 32),     // timeh
     0xC82 => haveUsrMode() & (sizeof(xlen) == 32),     // instreth
 
+    0b1100100 /* 0xC80 */ @ index : bits(5) if index >_u 0b00010 => haveZihpm() & haveUsrMode() & (sizeof(xlen) == 32), // mhpmcounterh3..31
```
arichardson, line 160:
```suggestion
    0b1100100 /* 0xC80 */ @ index : bits(5) if index >_u 0b00010 => haveZihpm() & haveUsrMode() & (sizeof(xlen) == 32), // hpmcounterh3..31
```

model/riscv_sys_control.sail
```diff
@@ -162,22 +174,24 @@ function check_CSR_access(csrrw, csrpr, p, isWrite) =
...
-  match(csr, p) {
-    (0xC00, Supervisor) => mcounteren[CY] == 0b1,
-    (0xC01, Supervisor) => mcounteren[TM] == 0b1,
-    (0xC02, Supervisor) => mcounteren[IR] == 0b1,
+// There are several features that are controlled by machine/supervisor enable
+// bits (m/senvcfg, m/scounteren, etc.). This abstracts that logic.
+function mode_bit_enabled(p : Privilege, machine_bit : bit, supervisor_bit : bit) -> bool = match p {
+  Machine => true,
+  Supervisor => machine_bit == bitone,
+  User => machine_bit == bitone & (not(haveSupMode()) | supervisor_bit == bitone),
```
arichardson, line 182: Should there be a reference to the spec here that you need both S and M bits?

> The setting of a bit in mcounteren does not affect whether the corresponding bit in scounteren is writable. However, U-mode may only access a counter if the corresponding bits in scounteren and mcounteren are both set.

Also I was initially confused by the name name of the function. Maybe something like ` function feature_enabled_for_priv(p : Privilege, machine_enable_bit : bit, supervisor_enable_bit : bit) -> ` would be clearer?

model/riscv_sys_control.sail
```diff
@@ -162,22 +174,24 @@ function check_CSR_access(csrrw, csrpr, p, isWrite) =
...
-
-    (_, _) => /* no HPM counters for now */
-              if   0xC03 <=_u csr & csr <=_u 0xC1F
-              then false
-              else true
-  }
+// Return true if the counter is enabled OR the CSR is not a counter.
+function check_Counteren(csr : csreg, p : Privilege) -> bool = {
+  // Check if it is not a counter.
+  if csr <_u 0xC00 | csr >_u 0xC1F
```
arichardson, line 188: Maybe this should be a fatal error instead? Do we even call this for non-counter CSRs?

c_emulator/riscv_platform.h
arichardson: Ah I misunderstood what this was for. I guess in the future this would be required to handle cores that have counters > N hardwired to zero, but since this is not configurable yet (riscv-config in the future?) maybe simpler to just have a hardcoded sail variable?
That variable could still be writable from the C model I guess?

Timmmm: Yeah I did it like this because in our fork we actually do set this from a JSON file. Ideally we'll use riscv-config in future.

I'm not sure about poking variables in the C code from C. Feels like a bad thing to do. Especially in this repo because it still uses a copy/pasted header which can easily deviate from the actual generated code and lead to nightmarish memory corruption bugs.

Mind if I just leave it? Btw if you have a better name suggestion I'm all ears, since the one I chose is apparently confusing!

model/prelude.sail
Timmmm: Yeah can do if you like - I think I just needed one missing operator for this and figured I'd add them all. I'll split it out.

model/riscv_sys_control.sail
Timmmm: Yeah it's called for all CSRs - see `check_CSR()` in `riscv_sys_control.sail`.

model/riscv_sys_control.sail
Timmmm: Changed name to your suggestion.

model/prelude.sail
Timmmm: #406

model/riscv_insts_zicsr.sail
Timmmm: Done

model/riscv_sys_control.sail
Timmmm: Done

model/riscv_sys_control.sail
Timmmm: Done

model/riscv_sys_control.sail
Timmmm: Updated to your suggested name. I didn't add that bit of the spec because I'm going to use this for senvcfg/menvcfg too in a future PR.

I kind of think we should have a general mechanism for linking to the spec, but it doesn't look like they've got around to writing it in a linkable way yet (like [this](https://spec.ferrocene.dev/ownership-and-deconstruction.html) for example). :-/

model/riscv_sys_control.sail
arichardson: Since the same style is used for other checks I guess we keep it as is. But IMO it would be cleaner to push that check into check_CSR, so we don't have all these checks being called for every CSR.

model/riscv_sys_control.sail
arichardson: Sounds like ideally the privspec should publish versioned HTML versions to github pages which would allow linking easily.

model/riscv_sys_regs.sail
```diff
@@ -566,6 +574,40 @@ function retire_instruction() -> unit = {
...
+type hpmidx = range(3, 31)
+
+// Convert the lowest 5 bits of a CSR index to an hpmidx. Asserts if it is 0..2.
+function hpmidx_from_bits(b : bits(5)) -> hpmidx = {
+  let index = unsigned(b);
+  if index >= 3 then index else {
+    assert(false, "unreachable HPM index");
+    undefined
+  }
+}
```
arichardson, line 595:
```suggestion
function hpmidx_from_bits(b : bits(5)) -> hpmidx = {
  assert(unsigned(b) >= 3, "unreachable HPM index");
  unsigned(b)
}
```

Does this type check?

arichardson (approved):
This LGTM but would be good to have someone else double-check

model/riscv_sys_regs.sail
Timmmm: It doesn't, but I updated it to this which does:

```
  let index = unsigned(b);
  assert(index >= 3, "unreachable HPM index");
  index
  ```
  
Not sure what I was thinking :-D

**Comment by Timmmm** [comment]
@billmcspadden-riscv I think it's ok if you double check this since you've also done a lot of HPM stuff.

**Comment by Timmmm** [comment]
@billmcspadden-riscv I've rebased this.

File: model/riscv_sys_regs.sail [hunk 10]
```diff
@@ -152,6 +155,9 @@ function have_privLevel(priv : priv_level) -> bool =
     0b11 => true,
   }
 
+/* Hardware Performance Monitoring counters */
+function haveZihpm()  -> bool = true
```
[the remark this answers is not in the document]
Timmmm: Ok done.

arichardson (approved):

Timmmm:
@billmcspadden-riscv please can you merge this? It's ready to go.

PR #405, opened 2024-02-08
Add m/senvcfg to CSR name map

Timmmm:
These were missed when those CSRs were implemented.


arichardson (approved):
@billmcspadden-riscv are we okay to merge this tiny bugfix PR?

PR #406, opened 2024-02-12
Add missing comparison operators

Timmmm:
Some less-than/greater-than operators were missing. This adds the full set.


billmcspadden-riscv (approved):
LGTM

PR #414, opened 2024-02-25
Fix error in format specifiers

ved-rivos:
Removes two compiler warnings.

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -239,8 +239,8 @@ static int process_args(int argc, char **argv)
 {
   int c;
   uint64_t ram_size = 0;
-  uint64_t pmp_count = 0;
-  uint64_t pmp_grain = 0;
+  int pmp_count = 0;
+  int pmp_grain = 0;
   while (true) {
     c = getopt_long(argc, argv,
                     "a"
@@ -287,8 +287,8 @@ static int process_args(int argc, char **argv)
       rv_enable_misaligned = true;
       break;
     case OPT_PMP_COUNT:
-      pmp_count = atol(optarg);
-      fprintf(stderr, "PMP count: %lld\n", pmp_count);
+      pmp_count = atoi(optarg);
+      fprintf(stderr, "PMP count: %d\n", pmp_count);
       if (pmp_count != 0 && pmp_count != 16 && pmp_count != 64) {
         fprintf(stderr, "invalid PMP count: must be 0, 16 or 64");
         exit(1);
@@ -296,8 +296,8 @@ static int process_args(int argc, char **argv)
       rv_pmp_count = pmp_count;
       break;
     case OPT_PMP_GRAIN:
-      pmp_grain = atol(optarg);
-      fprintf(stderr, "PMP grain: %lld\n", pmp_grain);
+      pmp_grain = atoi(optarg);
+      fprintf(stderr, "PMP grain: %d\n", pmp_grain);
       if (pmp_grain >= 64) {
         fprintf(stderr, "invalid PMP grain: must less than 64");
         exit(1);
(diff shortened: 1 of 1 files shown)


jrtc27:
This is just as wrong; 32-bit architectures, as well as 64-bit macOS, use unsigned long long for uint64_t. Use PRIu64 or similar.

ved-rivos:
You are right that it may not be portable. However, the two variables do not need to be of type uint64_t - their valid range is less than 64. Updated to change their type to int and use %d as format specifier.

c_emulator/riscv_sim.c
```diff
@@ -287,17 +287,17 @@ static int process_args(int argc, char **argv)
...
         exit(1);
       }
       rv_pmp_count = pmp_count;
       break;
     case OPT_PMP_GRAIN:
-      pmp_grain = atol(optarg);
-      fprintf(stderr, "PMP grain: %lld\n", pmp_grain);
+      pmp_grain = atoi(optarg);
+      fprintf(stderr, "PMP grain: %d\n", pmp_grain);
       if (pmp_grain >= 64) {
```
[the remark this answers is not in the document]
ved-rivos: Right. Fixed.

c_emulator/riscv_sim.c
```diff
@@ -287,17 +287,17 @@ static int process_args(int argc, char **argv)
...
       if (pmp_count != 0 && pmp_count != 16 && pmp_count != 64) {
         fprintf(stderr, "invalid PMP count: must be 0, 16 or 64");
         exit(1);
       }
       rv_pmp_count = pmp_count;
       break;
     case OPT_PMP_GRAIN:
-      pmp_grain = atol(optarg);
-      fprintf(stderr, "PMP grain: %lld\n", pmp_grain);
+      pmp_grain = (uint8_t)atoi(optarg);
```
Timmmm, line 299: This seems wrong? You'll silently truncate larger values which is potentially very confusing.

c_emulator/riscv_sim.c
```diff
@@ -287,17 +287,17 @@ static int process_args(int argc, char **argv)
       rv_enable_misaligned = true;
       break;
     case OPT_PMP_COUNT:
-      pmp_count = atol(optarg);
-      fprintf(stderr, "PMP count: %lld\n", pmp_count);
+      pmp_count = (uint8_t)atoi(optarg);
+      fprintf(stderr, "PMP count: %u\n", pmp_count);
```
arichardson, line 291:
```suggestion
      fprintf(stderr, "PMP count: %" PRId64 "\n", pmp_count);
```

No need to change the types, just use the ugly 64-bit format macros instead?

c_emulator/riscv_sim.c
ved-rivos: @Timmmm - would you like to fix the two warnings since you are most familiar with this code? I can close this PR if you provide a fix.

ved-rivos: Please do.

arichardson: > Touché :D
> 
> I would use @arichardson 's suggestion of `PRIu64`. It is very ugly so you could use `%llu` which is less ugly. I think people only recommend `PRIu64` due to theoretical fears about portability (maybe if `unsigned long long` was 128 bits??).
> 
> I can make a new PR if you like?

The problem is that some plaforms use `unsigned long` for uint64_t and some use `unsigned long long` (I believe macOS and glibc systems differ), so whichever one you use you'll get a warning on one platform.

Another option is to add a `(uintmax_t)` cast and the `%j` modifier, but that's just as many characters as the `PRIu64` macro.

Timmmm: Ahh TIL! The things C programmers have to put up with... :-D

I'll make a PR with that.

Issue #374, opened 2023-12-12
Suspicious dead code in vfncvt instructions


bacam:
While experimenting with strengthening Sail's type system to rule out some impossible cases, I noticed that the `SEW = 'm = 8` cases in the `VFNUNARY0` execution function are unreachable.  They are ruled out by the `illegal_fp_variable_width` check, which appears to treat the instructions as if they were widening operations rather than narrowing operations.  Perhaps this check was supposed to be a little different?

aamartin0000:
perhaps it's a placeholder, for eventual inclusion of FP8?

bacam:
I don't think so: there are both integer and float cases, and the float cases are explicit errors.

Alasdair:
It looks suspect to me, but I wasn't able to understand the wording of the vector spec well enough to really understand what it should be doing.

allenjbaum:
vfncvt is explicitly narrowing (that's what the N in vfNcvt is for) from a
double width (64b) source to a singlewidth (32b) result, with at least one
operand being a float
  64b                float-> 32b [un]signed int [truncate]
  64b                float-> 32b float [round to odd]
  64b [un]signed int-> 32b float

I am unaware of any FP8 or even FP16 plans that would change that...yet.
Neither format is part of the existing spec, though I expect that to change
But if there were any, I'd expect FP32->FP16, and FP16->FP8 .

Alasdair:
The current code is implementing `32 -> 16`, and `16 -> 8` conversions, which are being ruled out by the current legality checks. If they are not part of the spec we should remove them I guess?

Alasdair:
Doing the same legality check as the widening operation still seems suspect to me, as the spec states that in one case SEW, LMUL, etc refer to the destination, and in the other they refer to the source. That's the bit I don't fully understand.

PR #415, opened 2024-02-25
Fix compiler warnings in V extension code

ved-rivos:
1. Handle unexpected match values as internal errors - they are checked dynamically. 
2. Remove deprecated set syntax

### DIFF
--- a/model/riscv_insts_vext_red.sail
+++ b/model/riscv_insts_vext_red.sail
@@ -185,7 +185,8 @@ function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_po
         FVV_VFREDOSUM   => fp_add(rm_3b, sum, vs2_val[i]),
         FVV_VFREDUSUM   => fp_add(rm_3b, sum, vs2_val[i]),
         FVV_VFREDMAX    => fp_max(sum, vs2_val[i]),
-        FVV_VFREDMIN    => fp_min(sum, vs2_val[i])
+        FVV_VFREDMIN    => fp_min(sum, vs2_val[i]),
+        _               => internal_error(__FILE__, __LINE__, "Widening op unexpected")
       }
     }
   };
--- a/model/riscv_insts_vext_fp.sail
+++ b/model/riscv_insts_vext_fp.sail
@@ -371,6 +371,7 @@ function clause execute(VFUNARY0(vm, vs2, vfunary0, vd)) = {
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
   if illegal_fp_normal(vd, vm, SEW, rm_3b) then { handle_illegal(); return RETIRE_FAIL };
+  assert(SEW != 8);
 
   let 'n = num_elem;
   let 'm = SEW;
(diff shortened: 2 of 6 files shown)


jrtc27:
Should SEW’s type not instead be constrained to the set of valid values so you get static checking?

ved-rivos:
> Should SEW’s type not instead be constrained to the set of valid values so you get static checking?

They are. But all encoding's are not legal for all ops.

bacam:
It would be better to follow the style of similar functions (e.g., `VFUNARY1`) which `assert` the restriction immediately after the check that should enforce it.

As @jrtc27 said, ideally we would have the type system enforce this statically.  I have experimented a little with this, but some limitations in Sail's type system make it a little too awkward to submit at the moment.  It did manage to detect a problem, though (#374).

ved-rivos:
> It would be better to follow the style of similar functions (e.g., VFUNARY1) which assert the restriction immediately after the check that should enforce it.

That seems like a good idea. I will respin the PR with that fix.

bacam:
That case is quite different.  If you want to avoid using a wildcard, you could split `rfvvfunct6` into separate types for the single-width and widening variants.

ved-rivos:
> If you want to avoid using a wildcard, you could split rfvvfunct6 into separate types for the single-width and widening variants.

Using a wildcard is not an issue. i was however expecting the assert to prune those paths since rfvvfunct6 is a enum. I guess then such pruning using assets only works for sets?

bacam:
The type checker doesn't track enum values, so an assert won't suppress a incomplete pattern match warning.

ved-rivos:
Okay. Using a wildcard is not an issue if the  type checker cannot prune this. This should address the various warnings.

model/riscv_insts_vext_red.sail
```diff
@@ -185,7 +185,8 @@ function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_po
         FVV_VFREDOSUM   => fp_add(rm_3b, sum, vs2_val[i]),
         FVV_VFREDUSUM   => fp_add(rm_3b, sum, vs2_val[i]),
         FVV_VFREDMAX    => fp_max(sum, vs2_val[i]),
-        FVV_VFREDMIN    => fp_min(sum, vs2_val[i])
+        FVV_VFREDMIN    => fp_min(sum, vs2_val[i]),
+        _               => {internal_error(__FILE__, __LINE__, "Widening op unexpected"); sum}
```
Timmmm, line 189: I think you can remove `sum` and just do

```
_ => internal_error(__FILE__, __LINE__, "Widening op unexpected"),
```

`internal_error` is divergent so its return type isn't checked.

Timmmm (changes requested):
LGTM other than the minor comment.

model/riscv_insts_vext_red.sail
ved-rivos: updated.

Timmmm (approved):

arichardson (approved):

PR #420, opened 2024-03-01
Add unratified Smclic, Ssclic, Smclicshv extensions

dansmathers:
Spec: https://github.com/riscv/riscv-fast-interrupt
Tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/436

Note: pulls are not yet available for spike that support CLIC 

Related pulls to run arch-tests against the sail-riscv model:
https://github.com/riscv-software-src/riscv-config/pull/169 
https://github.com/riscv-software-src/riscof/pull/106
https://github.com/riscv-software-src/riscv-isa-sim/pull/1596

To include m-mode CLIC interrupt tests in riscof testlist flow, add Smclic to riscof yaml file, e.g.:
spike/spike_isa.yaml:
  ISA: RV32IMCZicsr_Zifencei_Smclic

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -403,9 +403,12 @@ function legalize_tvec(o : Mtvec, v : xlenbits) -> Mtvec = {
  match (trapVectorMode_of_bits(v[Mode])) {
    TV_Direct => v,
    TV_Vector => v,
+   TV_Smclic => v,
    _         => [v with Mode = o[Mode]]
  }
 }
+
+function haveSmclic()  -> bool = mtvec[Mode] == 0b11
 
 bitfield Mcause : xlenbits = {
   IsInterrupt : xlen - 1,
@@ -421,6 +424,7 @@ function tvec_addr(m : Mtvec, c : Mcause) -> option(xlenbits) = {
     TV_Vector => if   c[IsInterrupt] == 0b1
                  then Some(base + (zero_extend(c[Cause]) << 2))
                  else Some(base),
+    TV_Smclic => Some(base),                 
     TV_Reserved => None()
   }
 }
--- a/model/riscv_types.sail
+++ b/model/riscv_types.sail
@@ -266,13 +266,14 @@ overload to_str = {exceptionType_to_str}
 /* trap modes */
 
 type tv_mode = bits(2)
-enum TrapVectorMode = {TV_Direct, TV_Vector, TV_Reserved}
+enum TrapVectorMode = {TV_Direct, TV_Vector, TV_Smclic, TV_Reserved}
 
 val trapVectorMode_of_bits : tv_mode -> TrapVectorMode
 function trapVectorMode_of_bits (m) =
   match (m) {
     0b00 => TV_Direct,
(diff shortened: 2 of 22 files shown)


model/riscv_sys_regs.sail
```diff
@@ -403,10 +403,13 @@ function legalize_tvec(o : Mtvec, v : xlenbits) -> Mtvec = {
  match (trapVectorMode_of_bits(v[Mode])) {
    TV_Direct => v,
    TV_Vector => v,
+   TV_Smclic => v,
    _         => [v with Mode = o[Mode]]
  }
 }
 
+function haveSmclic()  -> bool = mtvec[Mode] == 0b11
```
jrtc27, line 411: haveFoo means Foo is available for use, not that its additions are currently being used. That is, haveSmclic means "is the model configured to provide the Smclic extension?", not "has software configured mtvec to use the CLIC?".

model/riscv_types.sail
```diff
@@ -266,13 +266,14 @@ overload to_str = {exceptionType_to_str}
...
 type tv_mode = bits(2)
-enum TrapVectorMode = {TV_Direct, TV_Vector, TV_Reserved}
+enum TrapVectorMode = {TV_Direct, TV_Vector, TV_Smclic, TV_Reserved}
 
 val trapVectorMode_of_bits : tv_mode -> TrapVectorMode
 function trapVectorMode_of_bits (m) =
   match (m) {
     0b00 => TV_Direct,
     0b01 => TV_Vector,
+    0b11 => TV_Smclic,
```
jrtc27, line 276: Only if you have the CLIC extension. Plus this is for all xtvec, not just mtvec.

dansmathers: I see this in the Priv spec: 

_The mtvec register must always be implemented, but can contain a read-only value. If mtvec is
writable, the set of values the register may hold can vary by implementation. The value in the
BASE field must always be aligned on a 4-byte boundary, and the MODE setting may impose
additional alignment constraints on the value in the BASE field._

So just as TV_Vector may not be supported by an implementation, TV_Smclic may not be supported by an implementation.  When @billmcspadden-riscv adds WARL config support to the riscv-sail model, TV_Smclic WARL support can be handled the same way as TV_Vector.  So I think it is incorrect to say "only if you have the CLIC extension".  CLIC extension is defined as just another mtvec mode which is labeled Reserved in the current priv spec.

model/riscv_sys_regs.sail
dansmathers: I can change all places that say haveSmclic to SmclicMode.  Is that acceptable?

model/riscv_types.sail
jrtc27: Well there's a bit of a distinction though. Direct and Vectored are both specified as what moved.BASE=0/1 mean, with >=2 reserved. Whereas this CLIC encoding is specified in its own extension on top of the base privileged spec. That is, the philosophical distinction is that vectored is part of the base privileged spec but may not be supported, whereas CLIC support is its own named extension on top.

PR #427, opened 2024-03-06
Implement Zcb extension

Timmmm:
This adds an implementation of the Zcb code size extension.

This fixes all of the comments on the previous PR #322 - and also adds flags to the emulator to enable the extension (it is disabled by default).


model/riscv_sys_regs.sail
```diff
@@ -309,6 +311,9 @@ function haveZfh()     -> bool = (misa[F] == 0b1) & (mstatus[FS] != 0b00)
 /* V extension has to enable both via misa.V as well as mstatus.VS */
 function haveVExt()    -> bool = (misa[V] == 0b1) & (mstatus[VS] != 0b00)
 
+/* Zcb has simple code-size saving instructions. (The Zcb extension depends on the Zca extension.) */
+function haveZcb()     -> bool = sys_enable_zcb()
```
arichardson, line 315: Not opposed to the command line flag, but is there a reason this is not a default-true function like most other extensions? (well it still needs to check for C, but otherwise default true).

Timmmm: Nope, it was a totally arbitrary choice. Happy to change it!

Although it does make the command line flag more awkward (`-disable-zcb`?). Are there any extensions that can be toggled but are on by default?

Timmmm: Internally we use a JSON file for all these settings. Maybe it would be worth upstreaming that because adding these flags is extremely tedious, and using a JSON file makes it slightly less so.

It does require C++ for the JSON parser though (nlohmann's).

model/riscv_insts_zcb.sail
```diff
@@ -0,0 +1,210 @@
...
+
+function clause execute C_NOT(rsdc) = {
+  let r = creg2reg_idx(rsdc);
+  X(r) = ~(X(r));
+  RETIRE_SUCCESS
+}
+
+/* ****************************************************************** */
+
+union clause ast = C_MUL : (cregidx,cregidx)
```
arichardson, line 197:
```suggestion
union clause ast = C_MUL : (cregidx, cregidx)
```

arichardson (approved):
LGTM, although I'm not sure we actually need this command line flag (especially if you're not actually using it).

Timmmm:
> I have not checked the encodings are correct, but the code LGTM. I'm not sure we actually need this command line flag (especially if you're not actually using it).

I mainly added it for other people to use and to reduce the merge conflicts for us. I think it's worth keeping for when/if we eventually actually use `riscv-config`. Though I'm happy to be overruled if it would block this being merged.

model/riscv_insts_zcb.sail
Timmmm: Done.

arichardson:
> > I have not checked the encodings are correct, but the code LGTM. I'm not sure we actually need this command line flag (especially if you're not actually using it).
> 
> I mainly added it for other people to use and to reduce the merge conflicts for us. I think it's worth keeping for when/if we eventually actually use `riscv-config`. Though I'm happy to be overruled if it would block this being merged.

I'm guessing a lot of this command line parsing will go away eventually, so if it makes things easier for you merging as is sounds good.

arichardson (approved):

billmcspadden-riscv (approved):
LGTM

Issue #434, opened 2024-03-18
Use newtype for physical and virtual addresses


Timmmm:
For clarify and to avoid mixing them up, we can use Sail's `newtype` feature to create distinct types for physical and virtual addresses. For example:

```
newtype physaddr = physaddr : xlenbits
newtype virtaddr = virtaddr : xlenbits

function translate(virtaddr(vaddr) : virtaddr) -> physaddr = physaddr(vaddr + 1)

function main() -> unit = {
  let x = physaddr(zeros(32));

  let y = translate(x)
}
```

This will give a compile time error since we accidentally passed a physical address into `translate()`.

See #433

PR #435, opened 2024-03-19
Clean up memory checking functions slightly

Timmmm:
* Move `write_knd_of_flags` into its own function.
* Wrap really long lines
* Move PMP check into `access_check` function. This is needed for two reasons:
  - The CBO extension needs to do explicit access checks.
  - In future it will also check PMAs so the name needs changing.

### DIFF
--- a/model/riscv_mem.sail
+++ b/model/riscv_mem.sail
@@ -50,6 +50,19 @@ function read_kind_of_flags (aq : bool, rl : bool, res : bool) -> option(read_ki
     (false, true, true)   => None()
   }
 
+function write_kind_of_flags (aq : bool, rl : bool, con : bool) -> write_kind =
+  match (aq, rl, con) {
+    (false, false, false) => Write_plain,
+    (false, true,  false) => Write_RISCV_release,
+    (false, false, true)  => Write_RISCV_conditional,
+    (false, true , true)  => Write_RISCV_conditional_release,
+    (true,  true,  false) => Write_RISCV_strong_release,
+    (true,  true , true)  => Write_RISCV_conditional_strong_release,
+    // throw an illegal instruction here?
+    (true,  false, false) => throw(Error_not_implemented("store.aq")),
+    (true,  false, true)  => throw(Error_not_implemented("sc.aq"))
+  }
+
 // only used for actual memory regions, to avoid MMIO effects
 function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext_access_type), paddr : xlenbits, width : int('n), aq : bool, rl: bool, res : bool, meta : bool) -> MemoryOpResult((bits(8 * 'n), mem_meta)) = {
   let result = (match read_kind_of_flags(aq, rl, res) {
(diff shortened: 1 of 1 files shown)


model/riscv_mem.sail
```diff
@@ -66,28 +79,39 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext
...
-    Ext_PhysAddr_OK()     => phys_mem_read(t, paddr, width, aq, rl, res, meta),
-    Ext_PhysAddr_Error(e) => MemException(e)
-  } else match t {
-    Execute()  => MemException(E_Fetch_Access_Fault()),
-    Read(Data) => MemException(E_Load_Access_Fault()),
-    _          => MemException(E_SAMO_Access_Fault())
-  }
+// Check if access is permitted according to PMPs and PMAs.
+val access_check : forall 'n, 0 < 'n <= max_mem_access . (AccessType(ext_access_type), Privilege, xlenbits, int('n)) -> option(ExceptionType)
+function access_check (t, p, paddr, width) = {
```
arichardson, line 84: Maybe name this something like `phys_access_check` to indicate that it's not going through the MMU?

arichardson: I guess https://github.com/riscv/sail-riscv/issues/434 would help here.

model/riscv_mem.sail
```diff
@@ -66,28 +79,39 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext
...
-  else if within_phys_mem(paddr, width)
-  then match ext_check_phys_mem_read(t, paddr, width, aq, rl, res, meta) {
-    Ext_PhysAddr_OK()     => phys_mem_read(t, paddr, width, aq, rl, res, meta),
-    Ext_PhysAddr_Error(e) => MemException(e)
-  } else match t {
-    Execute()  => MemException(E_Fetch_Access_Fault()),
-    Read(Data) => MemException(E_Load_Access_Fault()),
-    _          => MemException(E_SAMO_Access_Fault())
-  }
+// Check if access is permitted according to PMPs and PMAs.
```
arichardson, line 82:
```suggestion
/* Check if access is permitted according to PMPs and PMAs. */
```
Other functions seem to use this comment style - I forget what is preferred.

arichardson (approved):
Minor nit otherwise LGTM.

model/riscv_mem.sail
Timmmm: That makes sense; I'll rename it. And yeah #434 would definitely help!

model/riscv_mem.sail
Timmmm: I was presuming that they're from a distant past when Sail didn't support `//`? I can't see any advantage to using `/*` when `//` is supported - it's more tedious to edit, looks worse IMO and means you can't use `/* */` to comment out blocks of code.

I'm hoping one day Sail's built in formatter will normalise all this stuff!

Alasdair: I think it was just an arbitrary stylistic choice, as I think we have supported both comment styles for as long as sail-riscv has existed.

The block style comments are actually quite a bit easier to auto-format than line comments funnily enough.

model/riscv_mem.sail
Timmmm: Done

model/riscv_mem.sail
arichardson: Marking resuled.

arichardson (approved):

Timmmm:
@billmcspadden-riscv this is ready to merge. No functional changes; just refactoring.

PR #441, opened 2024-03-19
Fix flen=32 compilation

### LINKED ISSUE #351: flen 32 (F extension only) does not compile
You can compile the model with `flen` equal to 32 (for F support only) or 64 (F & D). Currently the `Makefile` hard-codes 64, which means 32 is basically never tested. Currently it does not compile due to a typo (32 should be 31):

```
Type error:
sail-riscv/model/riscv_fdext_regs.sail:102.10-36:
102 |  then if regval [32..16] == 0x_FFFF
    |          ^------------------------^
    | No overloading for (operator ==), tried:
    | * eq_int
    |    No overloading for vector_subrange, tried:
    |    * subrange_bits
    |       Could not resolve quantifiers for subrange_bits
    |       * (0 <= 16 & (16 <= 32 & 32 < 32))
...
```

As far as I can see it looks like it was broken when it was added 22 months ago, which means `flen` 32 has been broken for at least that long.

#348 would help avoid issues like this.

Timmmm:
This has been broken for at least 2 years. This fixes the compilation.

Fixes #351

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -37,7 +37,7 @@ function nan_box_H   val_16b =
 val      nan_unbox_H : flenbits -> bits(16)
 function nan_unbox_H   regval =
   if (sizeof(flen) == 32)
-  then if regval [32..16] == 0x_FFFF
+  then if regval [31..16] == 0x_FFFF
        then regval [15..0]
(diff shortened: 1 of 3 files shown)


arichardson (approved):

billmcspadden-riscv (approved):
LGTM

PR #445, opened 2024-04-05
Q extension

jordancarlin:
Merge branch 'LiuTaowen-Tony/master' (PR #187) into quads

Update to current conventions (remove effects, compatible with Zfh and Zfa extensions that were added since original version written).

Bugs from existing quad implementation still need to be resolved.


model/riscv_types.sail
```diff
@@ -367,26 +367,41 @@ mapping bool_not_bits : bool <-> bits(1) = {
...
+  HALF   => 0b01,
+  WORD   => 0b10,
+  DOUBLE => 0b11
+}
+
+function size_bits_backwards size = match size {
+  0b00 => BYTE,
+  0b01 => HALF,
+  0b10 => WORD,
+  0b11 => DOUBLE
```
jordancarlin, line 383: The version from PR #187 splits up this mapping into a forward and reverse function and adds some extra functions to handle quads. I can't say for sure why this was done, but I assume it is because word_width is currently 2 bits so cannot encode quads.

I'd be inclined to increase it to 3 bits and keep the handling of quads consistent with everything else. Most places size_bits is used it is prepended with a leading 0, so increasing it to 3 bits and giving quads an encoding of 100 should not impact these. For the couple of cases where it is not prepended with the leading 0 we can just pick off the bottom two bits.

I'd love any opinions on one approach vs the other.

martinberger: I think we should should increase to 3 bits for consistency.

LiuTaowen-Tony: Thank you for pointing out! I am sorry that I forgot to leave a comment here. 

Yes, this was because wordwidth was 2 bits, and a few other components refers to word_width. I wasn't too sure about the semantics of other modules relying on word width. And the quad module doesn't rely on this function, as far as what I can recall. Therefore, I just did some minimal change to make my code compile, without changing the semantics.

jordancarlin: @martinberger I attempted to change this to 3 bits and am running into an issue with one particular encoding. All of the others had a 0 bit before so I just removed that. This encoding uses a sign bit, so we need to pull off the bottom two bits. I'm not sure how to do that inside of the encoding statement as I get an error when I try right now. Not sure if there is a different syntax that should be used here.

We could also look at splitting this (and probably the other mappings related to word_width) into two separate mappings—one for integer related instructions and one for floating point—if that would be preferable.

Code in question:
```
mapping clause encdec = LOAD(imm, rs1, rd, is_unsigned, size, false, false) if (word_width_bytes(size) < sizeof(xlen_bytes)) | (not(is_unsigned) & word_width_bytes(size) <= sizeof(xlen_bytes))
<-> imm @ rs1 @ bool_bits(is_unsigned) @ size_bits(size)[1..0] @ rd @ 0b0000011 if (word_width_bytes(size) < sizeof(xlen_bytes)) | (not(is_unsigned) & word_width_bytes(size) <= sizeof(xlen_bytes))
```

Error message:

```
 model/riscv_insts_base.sail:301.58-58:
301 |  <-> imm @ rs1 @ bool_bits(is_unsigned) @ size_bits(size)[1..0] @ rd @ 0b0000011 if (word_width_bytes(size) < sizeof(xlen_bytes)) | (not(is_unsigned) & word_width_bytes(size) <= sizeof(xlen_bytes))
    |                                                          ^
    | current token: [
make: *** [Makefile:261: generated_definitions/c/riscv_model_RV32.c] Error 1
```

martinberger: @jordancarlin OK, I see the problem. I think that is because [1..0] is not a bidirectional mapping. So if you go from right to left how would you get the third bit? You can do something like this
```
mapping strip : bits(2) <-> bits(3) = {
  0b00 <-> 0b000,
  0b01 <-> 0b001,
  0b10 <-> 0b010,
  0b11 <-> 0b011,
}
```
and then replace the `size_bits(size)[1..0]` by `strip(size_bits(size))` I think.

jordancarlin: @martinberger Thanks for the explanation. That builds and passes the test suite. 

That being said, this is making me wonder if having QUAD in size_enc makes sense in the first place. size_enc is currently only used by integer load/store and atomic instructions. The floating point extensions all have their widths hardcoded into the instructions (which seems like something that should be changed). Seeing as QUAD is only supported for floating point (and has the same encoding as unsigned byte for integer loads) it might make more sense to create a separate floating point size mapping down the road and leave this mapping just for integers. I know @Timmmm has been working on some related things if you want to weigh in on this at all.

martinberger: @jordancarlin I agreed that the question should `Quad be encoded needs careful though for the reasons you mention.

jordancarlin: Per RISC-V Golden Model Meeting, Quad will be added to word_width and size_enc will be extended to 3 bits to support it. This 3rd bit can be stripped away when needed (integer loads). This makes sense because quads are being added to amo and byte length floating point formats are expected.

martinberger: After the change, we should do a `grep` through the code where we use `word_width`, in particular where we make a pattern match against something of type `word_width` and add suitable cases either fall-though or matching against `QUAD` so the code is better documented. For example adding a suitable `internal_error(__FILE__, __LINE__, "...")`.

jordancarlin: Done. Thanks for mentioning that.

c_emulator/riscv_softfloat.c
```diff
@@ -15,9 +16,70 @@ static uint_fast8_t uint8_of_rm(mach_bits rm)
...
+  mpz_t *from = v1.bits;
+
+  // I don't know if you mind a small memory leak or not
+  // malloc not being released
+  if (!isInitialized) {
+    mpz_init(tmp);
+    isInitialized = true;
+  }
+
+  mpz_mod_2exp(tmp, *from, 64); // lower 64 bits of v1
```
jordancarlin, line 39: I'm not super familiar with the more obscure Sail functions, but I know Sail implements a bunch of the mpz functions. Is there a way to do this directly with Sail datatypes and functions instead of having to create mpz_t variables? Would love any guidance here. Maybe @Alasdair

model/riscv_insts_aext.sail
```diff
@@ -130,14 +123,15 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
...
+            /* BYTE, HALF, and QUAD would only occur due to invalid decodes, but it doesn't hurt
              * to treat them as valid here; otherwise we'd need to throw an internal_error.
              */
             match width {
               BYTE   => true,
               HALF   => vaddr[0..0] == 0b0,
               WORD   => vaddr[1..0] == 0b00,
-              DOUBLE => vaddr[2..0] == 0b000
+              DOUBLE => vaddr[2..0] == 0b000,
+              QUAD   => vaddr[4..0] == 0b00000
```
Timmmm, line 134: Ditto

model/riscv_insts_base.sail
```diff
@@ -295,10 +295,20 @@ mapping clause assembly = RTYPE(rs2, rs1, rd, op)
 /* ****************************************************************** */
 union clause ast = LOAD : (bits(12), regidx, regidx, bool, word_width, bool, bool)
 
+/* Loads only need the a 2 bit size encoding as quads are not supported.
+The 3rd bit is replaced by a sign bit for loads. */
+
+mapping strip : bits(2) <-> bits(3) = {
```
Timmmm, line 301: I think rather than this, which seems like a bit of a hack, it would make more sense to just add a 2-bit version of `size_enc` (i.e. the original `size_enc`). Like `size_enc_narrow` or something.

Timmmm (changes requested):
Haven't read it all but I skimmed some of it and it mostly looks reasonable. You may want to wait for https://github.com/riscv/sail-riscv/pull/448 which will remove at least some of the changes needed. It's waiting for a new version of the Sail compiler to be released.

model/riscv_insts_base.sail
martinberger: Don't we then get two definitions of `BYTE`, `HALF`, `WORD`  and `DOUBLE`? It would be good to have a single source of truth on width.

c_emulator/riscv_softfloat.c
LiuTaowen-Tony: > Yeah I think it would be much much much simpler just to split it into two 64 bit values in Sail and pass those in as two separate parameters. They will arrive in C as `uint64_t`.

Correct me if I am wrong. I think by the time when I was implementing c to sail interface, one issue is that sail memory model only support at most 64bit memory transaction, therefore, I splitted quad word floating point load and store as two different memory movement.

I am not sure if this would cause any concurrency problem.

model/riscv_insts_base.sail
jordancarlin: Maybe if we did something like signed_size_enc that takes in the word_width and sign bit and returns a 3 bit encoding with the sign bit as the most significant bit and the 2 bit encoding in the lower 2 bits. That might be cleaner than just stripping off a bit and avoids the ambiguity of size_enc and size_enc_narrow.

model/riscv_insts_aext.sail
```diff
@@ -69,14 +61,15 @@ function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
...
+           /* BYTE, HALF, and QUAD would only occur due to invalid decodes, but it doesn't hurt
             * to treat them as valid here; otherwise we'd need to throw an internal_error.
             */
            match width {
              BYTE   => true,
              HALF   => vaddr[0..0] == 0b0,
              WORD   => vaddr[1..0] == 0b00,
-             DOUBLE => vaddr[2..0] == 0b000
+             DOUBLE => vaddr[2..0] == 0b000,
+             QUAD   => vaddr[4..0] == 0b00000
```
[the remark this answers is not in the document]
jordancarlin: Don't think it matters much since QUAD is an invalid decode and will get thrown out later, but I think you're right so I went ahead and changed it.

model/riscv_types.sail
```diff
@@ -368,25 +368,28 @@ mapping bool_not_bits : bool <-> bits(1) = {
...
-  BYTE   <-> 0b00,
-  HALF   <-> 0b01,
-  WORD   <-> 0b10,
-  DOUBLE <-> 0b11
+mapping size_enc : word_width <-> bits(3) = {
+  BYTE   <-> 0b000,
+  HALF   <-> 0b001,
+  WORD   <-> 0b010,
+  DOUBLE <-> 0b011,
+  QUAD   <-> 0b100
```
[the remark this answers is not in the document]
jordancarlin: Thanks for the clarification

model/riscv_sys_control.sail
```diff
@@ -490,11 +490,14 @@ function init_sys() -> unit = {
...
   then internal_error(__FILE__, __LINE__, "F and Zfinx cannot both be enabled!");
 
-  /* We currently support both F and D */
-  misa[F]   = bool_to_bits(sys_enable_fdext());      /* single-precision */
+  /* We currently support F, D and Q */
+  misa[F]   = bool_to_bits(sys_enable_fdext());        /* single-precision */
   misa[D]   = if   sizeof(flen) >= 64
                 then bool_to_bits(sys_enable_fdext())  /* double-precision */
                 else 0b0;
+  misa[Q]   = if   sizeof(flen) >= 128
```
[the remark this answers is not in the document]
jordancarlin: Good catch. I went ahead and changed this for Q and D.

c_emulator/riscv_softfloat.c
jordancarlin: I've changed it so that the functions in riscv_softfloat_interface.sail pass two 64 bit values to the external softfloat implementation. The riscv_softfloat.c file puts them together for the softfloat calls themselves and then splits them apart again to return them to Sail as 64 bit values. Let me know what you think. This eliminates the need for mpz entirely.

c_emulator/riscv_softfloat.c
```diff
@@ -15,9 +16,70 @@ static uint_fast8_t uint8_of_rm(mach_bits rm)
...
+
+#define SOFTFLOAT_POSTLUDE_128 softfloat_postlude_128
+
+static void softfloat_postlude_128(float128_t res)
+{
+  static sail_int _64;
+  static mpz_t lo;
+  static bool isInitialized = false;
+
+  // I don't know if you mind a small memory leak or not
```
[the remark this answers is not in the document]
jordancarlin: See above conversation. MPZ variables have been removed by switching to 64 bit ints.

model/riscv_insts_base.sail
martinberger: In essence, that what the `strip` mapping does implicitly. But I agree, we could make that more explicit.

jordancarlin: @martinberger @Timmmm So would we rather do a 2 bit mapping to a narrower `size_enc` or an additional 3 bit mapping that matches based on both sign and `word_width`?

Timmmm: I mean something like this:

```
mapping size_enc : word_width <-> bits(2) = {
  BYTE   <-> 0b00,
  HALF   <-> 0b01,
  WORD   <-> 0b10,
  DOUBLE <-> 0b11,
}

mapping size_enc_wide : word_width <-> bits(3) = {
  BYTE   <-> 0b000,
  HALF   <-> 0b001,
  WORD   <-> 0b010,
  DOUBLE <-> 0b011,
  QUAD <-> 0b100,
}
```

I originally suggested `size_enc_narrow` / `size_enc` but if you do `size_enc` / `size_enc_wide` then there's no need to change any of the existing uses. Though that may be desirable anyway because it looks like the `0b0 @ size_enc(..)` is really intended to one day be `size_enc_wide(..)`.

> Don't we then get two definitions of BYTE, HALF, WORD and DOUBLE? 

Two definitions of the encodings... kind of. I don't think it really matters in this case. They're not going to change and there are only 4. But I guess if you really wanted to avoid that you could do:

```
mapping size_enc_wide : word_width <-> bits(3) = {
  QUAD <-> 0b100,
  narrow <-> 0b0 @ size_enc(narrow),
}
```

martinberger: That's a good solution, too!

jordancarlin: I went with size_enc_wide referencing size_enc for consistency and clarity about what it is doing.

jordancarlin:
>You may want to wait for #448 which will remove at least some of the changes needed. It's waiting for a new version of the Sail compiler to be released.

It seems like the changes needed to either PR based on each other should be relatively minor, so it probably doesn't matter the order they are merged.

jordancarlin:
@Timmmm Thanks for all your earlier comments. I believe I’ve taken care of all of them at this point. I’d appreciate it if you could give this another look. Aside from more rigorous testing (which is being worked on) I think this PR is ready.

Makefile
```diff
@@ -18,14 +18,15 @@ else
   $(error '$(ARCH)' is not a valid architecture, must be one of: RV32, RV64)
 endif
 
-SAIL_FLEN := riscv_flen_D.sail
+SAIL_FLEN := riscv_flen_Q.sail
```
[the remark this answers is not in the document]
jordancarlin: Would you suggest adding a command line option? I think it would need to be a compile-time option if so. Or do we leave the makefile as is and have people manually change SAIL_FLEN before compiling if they want f128?

This issue will ideally go away once configuration via riscv-config is integrated, but since that might not be for a while, we probably need a solution until then.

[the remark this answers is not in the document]
Timmmm: Yeah I think currently the best option would be to make it a Make variable, so people have to do

```
ARCH=RV64 FLEN=128 make ...
```

I don't think we want to parse `riscv-config` at compile time. The thing that would really help this is https://github.com/rems-project/sail/issues/534 but there's no timeline for that so I think the Make variable is the way to go for now. We should have one for VLENMAX too really.

model/riscv_fdext_regs.sail
```diff
@@ -6,64 +6,100 @@
...
 function canonical_NaN_S() -> bits(32) = 0x_7fc0_0000
 function canonical_NaN_D() -> bits(64) = 0x_7ff8_0000_0000_0000
+function canonical_NaN_Q() -> bits(128) = 0x_7fff_0000_0000_0000_0000_0000_0000_0000
 
 val      nan_box_H : bits(16) -> flenbits
 function nan_box_H   val_16b =
-  if (sizeof(flen) == 32)
-  then 0x_FFFF @ val_16b
-  else 0x_FFFF_FFFF_FFFF @ val_16b
+  match (sizeof(flen)) {
```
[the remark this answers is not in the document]
jordancarlin: I know there is work being done on quad tests for riscv-arch-test which will want to use the sail model to generate signatures. If those are completed before we get all of the floating point updates into the Sail standard library and make the code more generic, then we may want to revisit merging this sooner but otherwise, I agree.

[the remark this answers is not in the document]
martinberger: @jordancarlin Do you know if anyone is planning to make and tape-out a real chip that implements the Q extension?

Makefile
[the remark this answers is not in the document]
jordancarlin: When the Make variable is set to 64, it causes issues with the boxing/unboxing functions. Hopefully switching over to the generic version will solve this.

```
model/riscv_fdext_regs.sail:51.11-27:
51 |    128 if regval [127..16] ==  0x_FFFF_FFFF_FFFF_FFFF_FFFF_FFFF_FFFF
   |           ^--------------^ checking function argument has type bool('p)
   | Could not resolve quantifiers for subrange_bits
   | * (0 <= 16 & (16 <= 127 & 127 < 64))
```

model/riscv_fdext_regs.sail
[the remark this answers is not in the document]
jordancarlin: @martinberger The OpenHW Core-V-Wally processor (https://github.com/openhwgroup/cvw) implements the Q extension in some of its configurations. There is an upcoming tape-out planned for it very soon. I'm double checking which configuration is planned for that, but it sounds like it might include Q.

PR #446, opened 2024-04-09
Add missing check for RV64 on float conversion instructions

Timmmm:
These two instructions are RV64-only, as noted in the comment, but were missing the check. There are other instructions in this file that did have the check so it was just an omission.


Timmmm:
@billmcspadden-riscv this should be uncontroversial.

PR #447, opened 2024-04-09
Derive xlen values from a logarithmic constant

Timmmm:
The motivation for this is that it makes deriving CHERI constants (field widths etc.) which vary between RV32 and RV64 a lot easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ else
   $(error '$(ARCH)' is not a valid architecture, must be one of: RV32, RV64)
 endif
 
+SAIL_XLEN += riscv_xlen.sail
 SAIL_FLEN := riscv_flen_D.sail
 SAIL_VLEN := riscv_vlen.sail
 
--- a/model/riscv_xlen.sail
+++ b/model/riscv_xlen.sail
@@ -6,7 +6,6 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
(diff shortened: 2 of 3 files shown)


arichardson:
Sounds like this needs a new sail version? Would be good if there was a new sail release soon (@Alasdair is this planned soon?) and we can add a version check (#370).

Makefile
```diff
@@ -17,6 +17,7 @@ else ifeq ($(ARCH),RV64)
 else
   $(error '$(ARCH)' is not a valid architecture, must be one of: RV32, RV64)
 endif
+SAIL_XLEN += riscv_xlen.sail
```
arichardson, line 21:
```suggestion

SAIL_XLEN += riscv_xlen.sail
```

Keeping all of these grouped?

arichardson (approved):
Looks reasonable to me and CI seems happy.

Timmmm:
Yeah I think it's because CI doesn't pin the version of the Sail compiler we use, so when the newer version was released which supports these types it magically fixed this PR even though I didn't change anything.

#532 will pin the Sail compiler version.

jordancarlin:
Probably worth merging an updated version of #370 before we merge this to prevent issues for anyone still running Sail 0.17. That way they will at least get a descriptive error message.

arichardson (approved):

jordancarlin (approved):
Should be good to go now that #534 is merged.

PR #454, opened 2024-04-18
Added pmm Support

HAMZA-AFZAL404:
# RISC-V Pointer Masking Extension Implementation

This PR features the implementation of the Pointer Masking Extension for the RISC-V architecture, enhancing the Base, Floating Point, Compressed, Vector and Atomics Instruction Set.

## Changes

1. Implemented the `mseccfg` register with associated changes in <ins>riscv_sys_regs.sail</ins> ,<ins>riscv_insts_zicsr.sail</ins> ,<ins>riscv_sys_control.sail</ins> ,<ins>riscv_csr_map.sail</ins> 

2. Updated the `leglize_envcfg` function and added functions`is_pmm_active` ,`legalize_mseccfg` and `get_pmm` in <ins>riscv_sys_regs.sail</ins>.

3. Introduced functions `transform_VA` and `transform_PA` in <ins>riscv_insts_base.sail</ins>.
       
4. Implemented the function `transform_effective_address`. and Integrated `transform_effective_address` into the execution of LOAD and STORE functions of the targeted instruction files.

## Testing

1. The functionality has been verified with several tests. However, testing is still ongoing. Once completed, the repository link will be shared promptly.

2. The self-checking tests have not been written yet. However, work has commenced on them and they should be available shortly.


Adam-pi3 (approved):

c_emulator/riscv_sim.c
```diff
@@ -136,6 +136,7 @@ static struct option options[] = {
     {"report-arch",                 no_argument,       0, 'a'                     },
     {"test-signature",              required_argument, 0, 'T'                     },
     {"signature-granularity",       required_argument, 0, 'g'                     },
+    {"enable-zpm",                  no_argument,       0, 'Y'                     },
```
Timmmm, line 139: Please can you use a long option (see the `OPT_...` macros)?

model/riscv_insts_aext.sail
```diff
@@ -61,36 +61,38 @@ function process_loadres(rd, addr, value, is_unsigned) =
...
 function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
+  let pmm = is_pmm_active();
   if haveAtomics() then {
     /* Get the address, X(rs1) (no offset).
      * Extensions might perform additional checks on address validity.
      */
     match ext_data_get_addr(rs1, zeros(), Read(Data), width) {
       Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
       Ext_DataAddr_OK(vaddr) => {
+        let  vaddr_ = transform_effective_address(vaddr, pmm);
```
Timmmm, line 72: Could this go inside `translateAddr`? I think the only reason it couldn't is if `transform_effective_address()` would affect alignment but that sounds like it shouldn't be possible? If the address given to `handle_mem_exception()` needs to be different I would say add that as a field of `TR_Failure`.

model/riscv_insts_base.sail
```diff
@@ -321,28 +321,86 @@ function check_misaligned(vaddr : xlenbits, width : word_width) -> bool =
          WORD   => vaddr[0] == bitone | vaddr[1] == bitone,
          DOUBLE => vaddr[0] == bitone | vaddr[1] == bitone | vaddr[2] == bitone
        }
+function transform_VA (effective_address : xlenbits,PMLEN :int) -> xlenbits = {
```
Timmmm, line 324: Add a blank line above here. Also space in `:int`.

model/riscv_insts_base.sail
```diff
@@ -321,28 +321,86 @@ function check_misaligned(vaddr : xlenbits, width : word_width) -> bool =
...
+  let XLEN = sizeof(xlen);
+  let i = XLEN -(PMLEN +1);
+  let effective_bit = effective_address[i..i];
+  let masked_bits = replicate_bits(effective_bit, PMLEN);
+  let unmasked_bits = effective_address[i..0];
+  masked_bits @ unmasked_bits
+}
+
+function transform_PA (effective_address : xlenbits,PMLEN :int) -> xlenbits = {
+  assert(PMLEN >= 0, "PMLEN out of range");
```
Timmmm, line 336: Instead of these asserts you can use `PMLEN : range(0, 63)`. Can it really be 63? Surely there's some reasonable limit? From the code below it looks like it can only be 7 or 16 so the correct type is `pmlen : {7, 16}`.

Also `PMLEN` should be lowercase.

model/riscv_insts_base.sail
```diff
@@ -321,28 +321,86 @@ function check_misaligned(vaddr : xlenbits, width : word_width) -> bool =
...
+  let XLEN = sizeof(xlen);
+  let i = XLEN -(PMLEN + 1);
+  let masked_bits = replicate_bits(0b0, PMLEN);
+  let unmasked_bits = effective_address[i..0];
+  masked_bits @ unmasked_bits
+}
+
+function transform_effective_address(vaddr : xlenbits, pmm : bool) -> xlenbits = {
+ if pmm then {
+    let pmm_bits : bits(2) = get_pmm();
```
Timmmm, line 347: Don't need the `: bits(2)` here.

model/riscv_insts_base.sail
```diff
@@ -321,28 +321,86 @@ function check_misaligned(vaddr : xlenbits, width : word_width) -> bool =
...
+    let mode : SATPMode = translationMode(cur_privilege);
```
Timmmm, line 348: Or `: SATPMode`

model/riscv_insts_base.sail
```diff
@@ -321,28 +321,86 @@ function check_misaligned(vaddr : xlenbits, width : word_width) -> bool =
...
+          (0b11) => transform_VA (vaddr,16),          
+        }
+      },      
+      (Sv48) => {
+        match (pmm_bits){
+          (0b10) => transform_VA (vaddr,7),
+          (0b11) => transform_VA (vaddr,16),          
+        }
+      },
+      (Sv57) => {
```
Timmmm, line 368: Did you really intend all of these branches to be identical? That doesn't seem right.

model/riscv_sys_regs.sail
```diff
@@ -831,6 +875,17 @@ function is_fiom_active() -> bool = {
     User => (menvcfg[FIOM] | senvcfg[FIOM]) == 0b1,
   }
 }
+
+// Return whether or not PMM is currently active, based on the current
+// privilege and the menvcfg,senvcfg and mseccfg settings. 
+function is_pmm_active() -> bool = {
+  match cur_privilege {
+    Machine => (mseccfg.PMM() == 0b11 | mseccfg.PMM() == 0b10),
```
Timmmm, line 883:
```
(match cur_privilege {
  Machine => mseccfg[PMM][1],
  Supervisor => menvcfg[PMM][1],
  User => menvcfg[PMM][1] | senvcfg[PMM][1],
}) == bitone
```

Timmmm (changes requested):
Can you link to the specification? Sorry, I might be being blind but I can't seem to find one that matches this!

Also this needs to be rebased, and the bitfield field access syntax has changed: `mseccfg.PMM()` -> `mseccfg[PMM]`.

c_emulator/riscv_sim.c
```diff
@@ -145,7 +146,7 @@ static struct option options[] = {
     {"trace-output",                required_argument, 0, OPT_TRACE_OUTPUT        },
     {"inst-limit",                  required_argument, 0, 'l'                     },
     {"enable-zfinx",                no_argument,       0, 'x'                     },
-    {"enable-writable-fiom",        no_argument,       0, OPT_ENABLE_WRITABLE_FIOM},
+    {"enable-writable-fiom",        no_argument,       0, OPT_ENABLE_WRITABLE_FIOM},
```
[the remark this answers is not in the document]
HAMZA-AFZAL404: Done

c_emulator/riscv_sim.c
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Basically here i intended to write a code in a way that if we are in Bare mode then then mask the Physical Address  and if we are in the any Translation Scheam mode then Mask the Virtaul Adrees  base on the value of PMM bits.BUT when i wrote a code in this way : `    match mode {
      Sbare => {
        match pmm_bits {
          0b10 => return transform_PA(vaddr, 7),
          0b11 => return transform_PA(vaddr, 16),
          _    => return vaddr // Default case
        }
      },
      Sv39 | Sv48 | Sv57 => {
        match pmm_bits {
          0b10 => return transform_VA(vaddr, 7),
          0b11 => return transform_VA(vaddr, 16),
          _    => return vaddr // Default case
        }
      },`

model/riscv_sys_regs.sail
HAMZA-AFZAL404: Done

model/riscv_insts_aext.sail
HAMZA-AFZAL404: It is possible to do this, but it will complicate the code. But what if the code is misaligned?

Timmmm: Yeah I'm not sure because I haven't seen the specification. Please can you link to it, ideally in the top message?

model/riscv_insts_base.sail
Timmmm: Ah I missed the PA/VA difference. Yes I would do it either like this:

```
if mode == Sbare then {
...
} else {
...
}
```

Or better:

```
match pmm_bits {
  0b10 => if mode == Sbare then transform_PA(vaddr, 7) else transform_VA(vaddr, 7),
  0b11 => if mode == Sbare then transform_PA(vaddr, 16) else transform_VA(vaddr, 16),
  _ => vaddr,
}
```

Note you don't need the `return`.

You could also do this but I don't know if it's clearer. Debatable.

```
match pmm_bits {
  0b1 @ size_or_whatever : bits(1) => {
     let other_size_thing = if size_or_whatever == 0b0 then 7 else 16;
     if mode == Sbare then transform_PA(vaddr, other_size_thing ) else transform_VA(vaddr, other_size_thing )
  },
  _ => vaddr,
}
```

Timmmm:
Please can you link to the specification that this is intended to implement?

martinmaas:
https://github.com/riscv/riscv-j-extension/blob/master/zjpm-spec.pdf

model/riscv_insts_aext.sail
```diff
@@ -202,18 +207,19 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
   match ext_data_get_addr(rs1, zeros(), ReadWrite(Data, Data), width_bytes) {
     Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
     Ext_DataAddr_OK(vaddr) => {
-      if not(is_aligned(vaddr, width))
-      then { handle_mem_exception(vaddr, E_SAMO_Addr_Align()); RETIRE_FAIL }
-      else match translateAddr(vaddr, ReadWrite(Data, Data)) {
-        TR_Failure(e, _) => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
+      let  vaddr_ = transform_effective_address(vaddr, pmm);
```
arichardson, line 210:
```suggestion
      let  vaddr = transform_effective_address(vaddr, pmm);       
```
Should reduce the diff.

model/riscv_insts_base.sail
```diff
@@ -317,7 +317,61 @@ function is_aligned(vaddr : xlenbits, width : word_width) -> bool =
...
   not(plat_enable_misaligned_access()) & not(is_aligned(vaddr, width))
 
+function transform_VA (effective_address : xlenbits,pmlen : {7, 16}) -> xlenbits = {
+  let XLEN = sizeof(xlen);
+  let i = XLEN -(pmlen +1);
+  let effective_bit = effective_address[i..i];
+  let masked_bits = replicate_bits(effective_bit, pmlen);
+  let unmasked_bits = effective_address[i..0];
+  masked_bits @ unmasked_bits
+}
```
Timmmm, line 327: This function can be greatly simplified:

```
function transform_VA(effective_address : xlenbits, pmlen : {7, 16}) -> xlenbits =
  sign_extend(effective_address[sizeof(xlen) - pmlen - 1 .. 0])
```

model/riscv_insts_base.sail
```diff
@@ -317,7 +317,61 @@ function is_aligned(vaddr : xlenbits, width : word_width) -> bool =
...
+
+function transform_PA (effective_address : xlenbits,pmlen : {7, 16}) -> xlenbits = {
+  let XLEN = sizeof(xlen);
+  let i = XLEN -(pmlen +1);
+  let masked_bits = replicate_bits(0b0, pmlen);
+  let unmasked_bits = effective_address[i..0];
+  masked_bits @ unmasked_bits
+}
```
Timmmm, line 335: This function can be similarly simplified:

```
function transform_PA(effective_address : xlenbits, pmlen : {7, 16}) -> xlenbits =
  zero_extend(effective_address[sizeof(xlen) - pmlen - 1 .. 0])
```

model/riscv_insts_base.sail
```diff
@@ -730,4 +787,4 @@ function clause execute SFENCE_VMA(rs1, rs2) = {
 }
 
 mapping clause assembly = SFENCE_VMA(rs1, rs2)
-  <-> "sfence.vma" ^ spc() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+  <-> "sfence.vma" ^ spc() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
```
Timmmm, line 790: Newline

model/riscv_sys_regs.sail
```diff
@@ -826,6 +881,15 @@ bitfield Vtype  : xlenbits = {
 }
 register vtype : Vtype
 
+//Returns the PMM bits value depends on the Privilege mode
+function get_pmm() -> bits(2) = {
```
Timmmm, line 885: Oh wait... you already had that here. Even simpler then!

```
function is_pmm_active() -> bool = get_pmm() != zeros()
```

Probably can just inline that actually.

model/riscv_sys_regs.sail
```diff
@@ -826,6 +881,15 @@ bitfield Vtype  : xlenbits = {
...
+  match cur_privilege {
+    Machine => mseccfg[PMM],
+    Supervisor => menvcfg[PMM],
+    User => if extensionEnabled(Ext_S)  then senvcfg[PMM] else menvcfg[PMM],
```
Timmmm, line 889: Extra space before `then`. You can also omit the outer `{}` on this function.

model/riscv_sys_regs.sail
```diff
@@ -750,13 +752,40 @@ function read_seed_csr() -> xlenbits = {
 /* Writes to the seed CSR are ignored */
 function write_seed_csr () -> option(xlenbits) = None()
 
+/*  mseccfg is a XLEN-bit read/write register that controls */
+/*  security features */
+bitfield MSeccfg : bits(64) = {
+  // Reserved WPRI bits.
+  wpri_2 : 63 .. 34,
+  // PMM bits of Pointer Masking Extension
+  PMM    : 33 .. 32,
```
Timmmm, line 761: Please can you add an enum and mapping for PMM to make the code more readable/robust? I.e.

```
enum PointerMaskingMode {
  Disabled,
  Reserved,
  Enabled7,
  Enabled16,
}

mapping pmm_bits : PointerMaskingMode <-> bits(2) = {
  Disabled <-> 0b00,
  Reserved <-> 0b01,
  Enabled7 <-> 0b10,
  Enabled16 <-> 0b11,
}
```

And then replace all literals with `pmm_bits(Enabled7)` or whatever.

Timmmm (changes requested):
Ok I've read the spec and this seems mostly ok. I've added some more comments.

I think the major thing is that the handling of pointer masking should be inside `translateAddr`. I checked very briefly and I think if you exclude `Execute` accesses then it matches the situations where pointer masking should be applied. (Double check that though  - apart from fetches are there any `translateAddr()` calls that you didn't have to add pointer masking to?

IIRC pointer masking *isn't* applied to the `xtval` address (passed into `handle_mem_exception()`). To deal with that, add an address field to `TR_Failure`. Something like:

```
      if   check_misaligned(vaddr, width)
      then { handle_mem_exception(vaddr, E_Load_Addr_Align()); RETIRE_FAIL }
      else match translateAddr(vaddr, Read(Data)) {
        TR_Failure(xtval_addr, e, _) => { handle_mem_exception(xtval_addr, e); RETIRE_FAIL },
```

jrtc27:
Putting pointer masking inside translateAddr would compose poorly with DDC on CHERI, which should authorise the actual masked virtual address used.

arichardson:
We already have `ext_data_get_addr` which can be used for the CHERI DDC hooks and I presume should also apply in all these cases for pointer masking. I recall the virtual memory managemtn instructions are explicitly opted out in the spec for some reason, but that should be easier to handle than adding yet another hook.

Timmmm:
Ah yeah we don't call `translateAddr()` for `sfence.vma` and friends so I think that should just work too.

You reminded me actually there's a bit of spec I didn't quite follow:

> From an implementation perspective, ignoring bits is deeply connected to the maximum virtual and
physical address space supported by the processor (e.g., Bare, Sv48, Sv57). In particular, applying the
above transformation is cheap if it covers only bits that are not used by any supported address
translation mode (as it is equivalent to switching off validity checks). Masking NVBITS beyond those
bits is more expensive as it requires ignoring them in the TLB tag, and even more expensive if the
masked bits extend into the VBITS portion of the address (as it requires performing the actual sign
extension). 

The Sail model has a TLB so maybe we need to do something, but I didn't quite get what it was trying to say here. It seems to be talking about 3 different cases:

1. it covers only bits that are not used by any supported address translation mode
2. it covers NVBITS beyond those bits
3. the masked bits extend into the VBITS portion of the address

It just doesn't seem to make any sense. There are only two parts of the address - NVBITS and VBITS, so what three cases is this referring to?

Can anyone say if we need to change the TLB code? I'm not sure.

model/riscv_sys_regs.sail
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
```diff
@@ -317,7 +317,61 @@ function is_aligned(vaddr : xlenbits, width : word_width) -> bool =
 function check_misaligned(vaddr : xlenbits, width : word_width) -> bool =
   not(plat_enable_misaligned_access()) & not(is_aligned(vaddr, width))
 
+function transform_VA (effective_address : xlenbits,pmlen : {7, 16}) -> xlenbits = {
```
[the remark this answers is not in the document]
HAMZA-AFZAL404: Done

model/riscv_insts_base.sail
HAMZA-AFZAL404: Done

handwritten_support/riscv_extras.lem
```diff
@@ -251,4 +255,4 @@ val print_bits : string -> bitvector -> unit
 let print_bits msg bs = print_endline (msg ^ (show_bitlist (bits_of bs)))
 
 val print_dbg : string -> unit
-let print_dbg msg = ()
+let print_dbg msg = ()
```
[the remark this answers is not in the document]
HAMZA-AFZAL404: DOne

model/riscv_insts_aext.sail
```diff
@@ -255,4 +261,4 @@ mapping amo_mnemonic : amoop <-> string = {
 }
 
 mapping clause assembly = AMO(op, aq, rl, rs2, rs1, width, rd)
-  <-> amo_mnemonic(op) ^ "." ^ size_mnemonic(width) ^ maybe_aq(aq) ^ maybe_rl(rl) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs2) ^ sep() ^ "(" ^ reg_name(rs1) ^ ")"
+  <-> amo_mnemonic(op) ^ "." ^ size_mnemonic(width) ^ maybe_aq(aq) ^ maybe_rl(rl) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs2) ^ sep() ^ "(" ^ reg_name(rs1) ^ ")"
```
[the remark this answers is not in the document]
HAMZA-AFZAL404: Done

c_emulator/riscv_sim.c
```diff
@@ -151,6 +153,7 @@ static struct option options[] = {
     {"enable-writable-fiom",        no_argument,       0, OPT_ENABLE_WRITABLE_FIOM},
     {"enable-svinval",              no_argument,       0, OPT_ENABLE_SVINVAL      },
     {"enable-zcb",                  no_argument,       0, OPT_ENABLE_ZCB          },
+
```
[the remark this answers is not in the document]
HAMZA-AFZAL404: Done

c_emulator/riscv_sim.c
```diff
@@ -267,7 +270,7 @@ static int process_args(int argc, char **argv)
                     "t:"
                     "T:"
                     "g:"
-                    "h"
+                    "h"
```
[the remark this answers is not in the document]
HAMZA-AFZAL404: Done

UmerShahidengr:
Regarding the comment on @jrtc27 and @Timmmm of implementing Pointer Masking in `translateAddr()`, I would say that pointer masking is different than virtual address translation, this implementation is efficient if not to be bounded inside translateAddr() function, it should be inside a different function (which is defined as `transform_effective_address()` here

arichardson:
> Regarding the comment on @jrtc27 and @Timmmm of implementing Pointer Masking in `translateAddr()`, I would say that pointer masking is different than virtual address translation, this implementation is efficient if not to be bounded inside translateAddr() function, it should be inside a different function (which is defined as `transform_effective_address()` here

IMO this should be part of `ext_data_get_addr` which already exists.

HAMZA-AFZAL404:
I will hopefully complete this in this week> @HAMZA-AFZAL404 can you please complete this one, it is stalled for a long time now, and it requires only a few minutes to complete

nadime15:
@HAMZA-AFZAL404 Any updates?

nadime15:
I am currently rebasing the code on a seperate branch that can be found [here](github.com/nadime15/sail-riscv/tree/ext_zpm)

> > We already have `ext_data_get_addr `which can be used for the CHERI DDC hooks and I presume should also apply in all these cases for pointer masking. I recall the virtual memory managemtn instructions are explicitly opted out in the spec for some reason, but that should be easier to handle than adding yet another hook.

> IMO this should be part of `ext_data_get_addr` which already exists.

I would like to follow the suggestion from @arichardson  and move `transform_effective_address()` into `ext_data_get_addr()`. This would significantly reduce the number of modified files.  

However, the issue is that `ext_data_get_addr()`, which is part of `riscv_addr_checks.sail`, is listed among the first files in `model/CMakeLists.txt`. Because of this, I can't access anything, no CSRs (including `satp`), no virtual memory helper functions like `translationMode()` (I know it's private).

Besides that, I think the whole idea of `ext_data_get_addr()` is to perform checks/modifications by extensions, but in its current structure, this isn't possible because all extensions, including needed CSRs, are defined later.

So, I have two (maybe more?) options:  
1) Follow the approach of this PR and modify nearly every occurrence of explicit memory accesses.  
2) Restructure the source file order so that `ext_data_get_addr()` has access to CSRs and Extensions.  

Any thoughts on that?

nadime15:
That was fast, thanks! :smile: 

Okay, but what does this mean in practice? This extension would greatly benefit from a hook. Would it make more sense to introduce a new function before `translateAddr()` that acts as a hook for potential future extensions to check/modify the address? Or should I be more explicit, like in this PR, by adding `transform_effective_address`() to every occurrence of explicit memory accesses ?

jrtc27:
I imagine we'd want a central function that itself calls into pointer masking-specific functions, ext_data_get_addr, and any others needed by future extensions, with all instructions then using that one function. This means changing every instruction now, but avoids having to do so every time another extension comes along.

nadime15:
Ok makes sense! What do you think about this example for LOAD?
https://github.com/riscv/sail-riscv/blob/4e9b487dfc8c1d9645022ce94bb37976e8ee3562/model/riscv_insts_base.sail#L299-L357

I kept it generic, although I could have kept LOAD as before and simply replaced `process_addr()` with `ext_data_get_addr()` to forward the result. However, we may want to add different custom error types in the future.

Timmmm:
Yeah that looks good to me. I would maybe pick different names but we can bike shed that later.

jordancarlin:
Pointer masking support was completed in #969. Closing this.

Issue #21, opened 2019-09-17
C.ADDI disassembly inaccurate when immediate value is negative


scottj97:
The opcode `0x1541` should decode into `c.addi x10, -16` (according to Spike). Sail is reporting it as `c.addi x10, 48`.

It seems to execute correctly; if `x10` is 0, it gets written with `0xffffffff_fffffff0`. It's only the immediate value shown in the disassembly that looks incorrect.

allenjbaum:
Any progress on this one? It's been 2 years....

allenjbaum:
This is not a problem with the disassembly of C.ADDI; It appears to be a problem for every instruction  with an immediate value..
@jrtc27 We can fix this particular mnemonic pretty easily by defining an signed variant - but it is a fair number of uses in some dozen or so files. Should we do this, or suffer along with a different disassembly? @ben-marshall: the crypto scalar has a bunch of immediate fields. Are the signed or unsigned or neither (in which case they're effectively unsigned)

allenjbaum:
Looking a bit deeper: the following instructions improperly decode signed constants as unsigned:
[F]Loads/Stores,  (LUI,AUIPC), JAL[R], btype ops (B**), Addi[w]/xori/ori/andi/slti[u], c.addi[w/16sp], c.[l[u]/and]i, c.j[al], c.bneq, c.beqz

I am assuming that crypto constants, hints, CSR#s, fences and shift amounts are not sign extended.
There a bunch of compressed immediates that are specifically zero extended

That's 19 lines in 3 files: base, cext, fext (plus defining signed versions of the hex_bits_n functions, or explicitly sign extending in those 19 lines

martinberger:
> That's 19 lines in 3 files: base, cext, fext 

@allenjbaum It's remarkable that this remained undetected for several years.

jrtc27:
We've known about it for years, just not a priority as it only affects the human-readable disassembly. If someone wants to fix it then great, but the string<->int handling is a bit nasty, so be warned...

allenjbaum:
You're right; it is rather...opaque (not good for something that's supposed to be readable and nasty code should be at least comments) and the hex_bits_n function don't seem to have anything to do with hex; it looks strictly binary to decimal string. I'd really prefer if it was hex (I'm not adept at reading Sail, so I could be misreading that... and no comments ....).

KotorinMinami:
[...]

PR #456, opened 2024-04-23
fix disassembly problems

KotorinMinami:
Fix Issue #21 by adding function to deal with the sign number

### DIFF
--- a/model/hex_bits_signed.sail
+++ b/model/hex_bits_signed.sail
@@ -1,0 +1,141 @@
+/*==========================================================================*/
+/*     Sail                                                                 */
+/*                                                                          */
+/*  Sail and the Sail architecture models here, comprising all files and    */
+/*  directories except the ASL-derived Sail code in the aarch64 directory,  */
+/*  are subject to the BSD two-clause licence below.                        */
+/*                                                                          */
+/*  The ASL derived parts of the ARMv8.3 specification in                   */
+/*  aarch64/no_vector and aarch64/full are copyright ARM Ltd.               */
+/*                                                                          */
+/*  Copyright (c) 2013-2021                                                 */
+/*    Kathyrn Gray                                                          */
+/*    Shaked Flur                                                           */
+/*    Stephen Kell                                                          */
+/*    Gabriel Kerneis                                                       */
+/*    Robert Norton-Wright                                                  */
+/*    Christopher Pulte                                                     */
+/*    Peter Sewell                                                          */
+/*    Alasdair Armstrong                                                    */
+/*    Brian Campbell                                                        */
+/*    Thomas Bauereiss                                                      */
+/*    Anthony Fox                                                           */
+/*    Jon French                                                            */
+/*    Dominic Mulligan                                                      */
+/*    Stephen Kell                                                          */
+/*    Mark Wassell                                                          */
+/*    Alastair Reid (Arm Ltd)                                               */
+/*                                                                          */
+/*  All rights reserved.                                                    */
+/*                                                                          */
+/*  This work was partially supported by EPSRC grant EP/K008528/1 <a        */
+/*  href="http://www.cl.cam.ac.uk/users/pes20/rems">REMS: Rigorous          */
+/*  Engineering for Mainstream Systems</a>, an ARM iCASE award, EPSRC IAA   */
+/*  KTF funding, and donations from Arm.  This project has received         */
+/*  funding from the European Research Council (ERC) under the European     */
+/*  Union’s Horizon 2020 research and innovation programme (grant           */
+/*  agreement No 789108, ELVER).                                            */
+/*                                                                          */
+/*  This software was developed by SRI International and the University of  */
+/*  Cambridge Computer Laboratory (Department of Computer Science and       */
+/*  Technology) under DARPA/AFRL contracts FA8650-18-C-7809 ("CIFV")        */
+/*  and FA8750-10-C-0237 ("CTSRD").                                         */
+/*                                                                          */
+/*  Redistribution and use in source and binary forms, with or without      */
+/*  modification, are permitted provided that the following conditions      */
+/*  are met:                                                                */
+/*  1. Redistributions of source code must retain the above copyright       */
+/*     notice, this list of conditions and the following disclaimer.        */
+/*  2. Redistributions in binary form must reproduce the above copyright    */
+/*     notice, this list of conditions and the following disclaimer in      */
+/*     the documentation and/or other materials provided with the           */
+/*     distribution.                                                        */
+/*                                                                          */
+/*  THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS''      */
+/*  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED       */
+/*  TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A         */
+/*  PARTICULAR PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR     */
+/*  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,            */
+/*  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT        */
+/*  LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF        */
+/*  USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND     */
+/*  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,      */
+/*  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT      */
+/*  OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF      */
+/*  SUCH DAMAGE.                                                            */
+/*==========================================================================*/
+
+$ifndef _HEX_BITS_SIGNED
+$define _HEX_BITS_SIGNED
+
+$include <vector.sail>
+$include <string.sail>
+
+val parse_hex_bits_signed : forall 'n, 'n > 0. (int('n), string) -> bits('n)
+val valid_hex_bits_signed : forall 'n, 'n > 0. (int('n), string) -> bool
+
+val hex_bits_signed: forall 'n, 'n > 0. bits('n) <-> (int('n), string)
+
+function hex_bits_signed_forwards(bv) = {
+  if signed(bv) < 0
+  then (length(bv), concat_str("-", hex_str(unsigned(not_vec(bv) + 1))))
+  else (length(bv), hex_str(unsigned(bv)))
+}
+function hex_bits_signed_forwards_matches(bv) = true
+
+function parse_hex_bits_signed(n, str) = {
+  if string_take(str, 1) == "-"
+  then {
+    let str_ = string_drop(str, 1);
+    let bv = parse_hex_bits(n, str_);
+    not_vec(bv) + 1
+  }
+  else parse_hex_bits(n, str)
+}
+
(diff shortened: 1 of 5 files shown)


Timmmm (commented):
I think this is good, and doesn't affect execution so there's no risk.

The only slightly odd thing is that `hex_bits.sail` is only in this repo temporarily until we update the Sail compiler. Then we can `$include <hex_bits.sail>`. However there's no `$include <hex_bits_unsigned.sail>` in the Sail "standard library" so this file will hang around. Maybe we want to add that to the Sail compiler?

In any case I don't think that should block merging this.

Can you squash your commits and write a proper commit message. It should be something like

```
Change immediates to be signed in assembly

These immediates are sign extended and usually interpreted as signed, so it's less confusing to use signed numbers. This also matches SPIKE's disassembly.
```

(Feel free to copy/paste that.)

Timmmm:
No problem at all! This looks ready to merge to me now. @billmcspadden-riscv do you want to do the honours or would I be ok to merge non-risky PRs like this?

@KotorinMinami I think it probably makes sense to add this file to Sail. No need to write any OCaml; you can just add the file next to [the existing `hex_bits.sail`](https://github.com/rems-project/sail/blob/sail2/lib/hex_bits.sail). Maybe extend it to 64 bits to match that file.

Timmmm (approved):

jrtc27:
I'd like to let @Alasdair look over this first given it's delving into the depths of forward and backwards mapping match functions

model/hex_bits_signed.sail
```diff
@@ -0,0 +1,142 @@
...
+$include <string.sail>
+
+val parse_hex_bits_signed : forall 'n, 'n > 0. (int('n), string) -> bits('n)
+val valid_hex_bits_signed : forall 'n, 'n > 0. (int('n), string) -> bool
+
+val hex_bits_signed: forall 'n, 'n > 0. bits('n) <-> (int('n), string)
+
+function hex_bits_signed_forwards(bv) = {
+  if signed(bv) < 0
+  then (length(bv) , concat_str("-" , hex_str(unsigned(not_vec(bv)+1))))
```
jrtc27, line 81: Your , and + whitespace doesn't conform to the style in a bunch of places, and isn't even self-consistent

model/hex_bits_signed.sail
```diff
@@ -0,0 +1,142 @@
...
+    not_vec(bv)+1
+  }
+  else parse_hex_bits(n, str)
+}
+
+function valid_hex_bits_signed(n, str) ={
+  let len = string_length(str);
+  if string_take(str , 1) == "-"
+  then valid_hex_bits(n, string_drop(str , 1))
+  else valid_hex_bits(n, str)
```
jrtc27, line 100: That's not true? An N-bit signed value is only valid if it's an N-1-bit signed value or it's exactly -(1 << N). This would, for example, allow parsing 0xff as a signed 8-bit value.

model/hex_bits_signed.sail
```diff
@@ -0,0 +1,142 @@
...
+  if string_take(str , 1) == "-"
+  then {
+    let str_ = string_drop(str , 1) ;
+    let bv = parse_hex_bits(n, str_);
+    not_vec(bv)+1
+  }
+  else parse_hex_bits(n, str)
+}
+
+function valid_hex_bits_signed(n, str) ={
```
jrtc27, line 96:
```suggestion
function valid_hex_bits_signed(n, str) = {
```

jrtc27 (changes requested):
See comments

model/hex_bits_signed.sail
KotorinMinami: Oh, it's my fault! I apologize for not adhering to the standards.I have made the necessary revisions. Kindly review it once again, please.

model/hex_bits_signed.sail
KotorinMinami: This function is designed to ensure the validity when converting from a string back to an integer. It should correspond to the hex_bits_signed_forwards function. 
When using the hex_bits_signed_forwards function to convert a signed numerical value to a string, it will convert negative numbers to their absolute values and then convert them to strings. Additionally, a negative sign is added at the beginning of the result. This validation is specifically tied to this function. When calling this function, it should be ensured that the input parameter is a string converted from hex_bits_signed_forwards and that the number of bits matches. 
I believe that in other SAIL files, the calls to this function meet these requirements, as other ASTs only invoke it when the value at this location should be interpreted as a signed number, as mentioned in the issue #21 . 
I believe these modifications only make the output of the SAIL simulator more user-friendly and accurate, and they should not cause any other issues.

model/hex_bits_signed.sail
KotorinMinami: Thanks for suggestion

model/hex_bits_signed.sail
jrtc27: Yes, I know what this function is for. I'm saying it's buggy, because it accepts more inputs than it should.

KotorinMinami: I understand then, so I should no matter what, check if the bleedin' string passed in is either a non-negative number or a minus followed by a positive number, ain't it? Just let me have a think on the best way to go about it...

Timmmm: Good spot, though I think the behaviour here is the same as the existing unsigned one:

```
function hex_bits_backwards(n, str) = parse_hex_bits(n, str)
function hex_bits_backwards_matches(n, str) = valid_hex_bits(n, str)
```

```
void parse_hex_bits(lbits *res, const mpz_t n, const_sail_string hex)
{
  if (strncmp(hex, "0x", 2) != 0) {
    goto failure;
  }

  mpz_t value;
  mpz_init(value);
  if (mpz_set_str(value, hex + 2, 16) == 0) {
    res->len = mpz_get_ui(n);
    mpz_set(*res->bits, value);
    mpz_clear(value);
    return;
  }
  mpz_clear(value);

  // On failure, we return a zero bitvector of the correct width
failure:
  res->len = mpz_get_ui(n);
  mpz_set_ui(*res->bits, 0);
}

bool valid_hex_bits(const mpz_t n, const_sail_string hex) {
  if (strncmp(hex, "0x", 2) != 0) {
    return false;
  }

  for (int i = 2; i < strlen(hex); i++) {
    char c = hex[i];
    if (!((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F'))) {
      return false;
    }
  }

  return true;
}
```

If you do `let foo : bits(8) = hex_bits_16("0xFFFF");` it will set `res->len` to 8 but `res->bits` to 0xFFFF. IIRC I *think* that's ok from `lbits` point of view - you're allowed to have "extra bits"; they're just truncated. Don't quote me on that though, it's been a while since I read `sail.c`.

Obviously that isn't ideal though. It probably makes sense for `valid_hex_bits` to actually do something with `n`. You only need to check the total length and the first non-zero digit so it shouldn't be too hard.

Seeing as that code needs changes I wonder if it makes more sense to just add signed support at the same time in `sail.c`.

On the other hand, given that it's an existing bug maybe it shouldn't block this? @Alasdair any thoughts?

KotorinMinami: @jrtc27 Would you mind taking another look at the code, please?

Alasdair: Hmm, I think the Sail library functions are either missing a check on the length of the string, or are missing a call to `normalize_lbits`. In general the `lbits` type should not have bits set outside its width.

When used in a mapping, the `parse_hex_bits` function will always be guarded by `valid_hex_bits` - the intended semantics is when `valid_hex_bits` returns false `parse_hex_bits` will just return a zero bitvector of the correct width on the same input. I think:
```
  if string_take(str, 1) == "-"
  then valid_hex_bits(n, string_drop(str, 1))
  else valid_hex_bits(n, str)
```
should do the correct thing if not for the lack of a length check in valid_hex_bits. I guess the work-around would be to add a check on the length in Sail.

Alasdair: See: https://github.com/rems-project/sail/pull/538

Note that the potential for anything to actually go wrong here is very slim, because we statically check that the string parsing part of the mapping has no semantic effect on the model - which means it never get executed, and therefore the actual behavior of the primitive in the Sail library is mostly irrelevant.

KotorinMinami: @Alasdair  So, if I understand correctly, you're saying that I just need to keep the solution from my [last commit](https://github.com/riscv/sail-riscv/pull/456/commits/998f35a3b8d12f0b19ed3c5adf7e20d62901e6df)? And this buggy problem will be resolved in the new version release of Sail? Does that mean I can revert the commit and the code will move on to the next review? If that's the case, thank you for your help and clarification. Otherwise, could you please provide me with some further advice?

Timmmm: > we statically check that the string parsing part of the mapping has no semantic effect on the model

That's really interesting - can you explain more how/where that works? Presumably this check can be disabled otherwise you can't actually use the reverse mapping?

Alasdair: @KotorinMinami yes, that seems right

@Timmmm Internally we keep track of what side effects expressions have, and anything with a string append pattern gets specially marked. It would be good to expand this check, because model checkers in EDA tools don't typically have good support for SystemVerilog strings, so if you want to do any kind of model checking with the Sail spec you need to ensure that the semantics doesn't rely on logic involving strings (which would be somewhat tasteless in an ISA spec anyway). Right now this does mean there is no practical way to use the 'string -> ast' side of the mapping , other than as part of the documentation generation flow.

KotorinMinami: @jrtc27 @Alasdair  Could you please spare some time to move on to the next review ?

model/hex_bits_signed.sail
```diff
@@ -0,0 +1,141 @@
...
+  if signed(bv) < 0
+  then (length(bv), concat_str("-", hex_str(unsigned(not_vec(bv) + 1))))
+  else (length(bv), hex_str(unsigned(bv)))
+}
+function hex_bits_signed_forwards_matches(bv) = true
+
+function parse_hex_bits_signed(n, str) = {
+  if string_take(str, 1) == "-"
+  then {
+    let str_ = string_drop(str, 1);
```
Timmmm, line 89: Btw you don't actually need a different name here. You can just do `let str = string_drop(str, 1);` and it essentially makes a new variable that shadows the old one. I generally prefer it in situations like this since it prevents you from accidentally using the old one in later code.

model/hex_bits_signed.sail
```diff
@@ -0,0 +1,141 @@
...
+$ifndef _HEX_BITS_SIGNED
+$define _HEX_BITS_SIGNED
+
+$include <vector.sail>
+$include <string.sail>
+
+val parse_hex_bits_signed : forall 'n, 'n > 0. (int('n), string) -> bits('n)
+val valid_hex_bits_signed : forall 'n, 'n > 0. (int('n), string) -> bool
+
+val hex_bits_signed: forall 'n, 'n > 0. bits('n) <-> (int('n), string)
```
Timmmm, line 77: Very minor but `val hex_bits_signed :` (add a space before the `:`).

KotorinMinami: Fixed .
Thanks for pointing out.
Is there any other problem ?

Timmmm:
@jrtc27 any chance you could take another look at this? To refresh your memory, you pointed out that the disassembly direction of the mapping accepted too-large literals, but it turned out that:

1. It was incorrect for the unsigned mapping too.
2. Alasdair has since fixed the issue in the Sail compiler.
3. The Sail compiler doesn't let you use the disassembler anyway.

Thanks!

Alasdair (approved):
I think we can move forwards with this. I have corrected the implementation of `valid_hex_bits` and `parse_hex_bits` so this should now behave as intended.

Even with the previous buggy behaviour this is unlikely to cause any issues with people on because we don't currently generate code for the assembly direction, and even go so far as to check that it is not used by any executable parts of the specification. The only place the assembly side of the mapping would really be visible is in the JSON output for documentation.

jrtc27:
Sure, that's fine. I would like to see this move into the Sail runtime itself though in the long term alongside the unsigned functions, as it's much better equipped to do this efficiently and correctly.

model/hex_bits_signed.sail
billmcspadden-riscv: Resolving this conversation based on jrtc27 response on May 21.

PR #458, opened 2024-04-25
RVWMO support via Sail concurrency interface

Alasdair:
Sail has for a while now had a flexible way of passing additional information to either operational or axiomatic concurrency models by instantiating outcomes (effects) with model-specific types. The set of possible outcomes is defined in the Sail library, and a subset of these can be instantiated by any model.

This PR refactors the base memory access functions to correctly instantiate these outcomes, which enables our concurrency tooling (most relevantly to this PR: https://github.com/rems-project/isla) to work correctly with the RISC-V model. Previously I had been maintaining this on my own fork, but it really should be upstreamed!

Outside of the memory read and write primitives being changed, there are several small changes required for the concurrency model. We need to announce when each instruction has ended, when branches occur, and tell the concurrency model what instruction is being executed after each ifetch. This for dependency analysis (deriving addr, data,  and ctrl). The riscv_analysis.sail file that was previously being used for this is no longer required, so it is removed as part of this PR. There are also a few additional entry points to the model other than `main` used by the symbolic execution tooling to execute a single instruction, with and without any prior initialisation.

This PR was tested on the tests in https://github.com/litmus-tests/litmus-tests-riscv, but I do need to re-run them so we should hold off on merging until that has been done.


Makefile
```diff
@@ -107,15 +107,8 @@ SAIL_ARCH_SRCS += riscv_types_kext.sail    # Shared/common code for the cryptogr
 SAIL_STEP_SRCS = riscv_step_common.sail riscv_step_ext.sail riscv_decode_ext.sail riscv_fetch.sail riscv_step.sail
 RVFI_STEP_SRCS = riscv_step_common.sail riscv_step_rvfi.sail riscv_decode_ext.sail riscv_fetch_rvfi.sail riscv_step.sail
 
-# Control inclusion of 64-bit only riscv_analysis
-ifeq ($(ARCH),RV32)
 SAIL_OTHER_SRCS     = $(SAIL_STEP_SRCS)
 SAIL_OTHER_COQ_SRCS = riscv_termination_common.sail riscv_termination_rv32.sail
```
arichardson, line 111: This now unconditionally uses riscv_termination_rv32.sail, even for RV64. Are these functions still needed? If so the conditional is still required. I've never used the coq build so not sure about that.

Alasdair: That was just an oversight on my part. I actually think they do not need to be separate after Nikhil's refactoring of the virtual memory code.

Alasdair: It might also be the case that `SAIL_OTHER_SRCS` is unnecessary now as there is no `riscv_analysis` file, but I'd rather just make the minimal change to the Makefile.

arichardson (approved):

Makefile
billmcspadden-riscv: Marking as resolved per Alasdair

PR #461, opened 2024-04-29
Don't read or write 8 bytes for 4-byte PTEs

### LINKED ISSUE #459: VM code reads 8 bytes even for 4-byte PTEs
In `pt_walk`:

```
  // Read this-level PTE from mem
  let mem_result = mem_read_priv(Read(Data),              // AccessType
                                 Supervisor,              // Privilege
                                 pte_phys_addr,
                                 8,                       // atom (8)
                                 false,                   // aq
                                 false,                   // rl
                                 false);                  // res

  match mem_result {
    MemException(_) => PTW_Failure(PTW_Access(), ext_ptw),
    MemValue(pte)   => {
      let pte_flags = Mk_PTE_Flags(pte[7 .. 0]);
```

Should be:

```
  // Read this-level PTE from mem
  let mem_result = mem_read_priv(Read(Data),              // AccessType
                                 Supervisor,              // Privilege
                                 pte_phys_addr,
                                 2 ^ sv_params.log_pte_size_bytes,
                                 false,                   // aq
                                 false,                   // rl
                                 false);                  // res

  match mem_result {
    MemException(_) => PTW_Failure(PTW_Access(), ext_ptw),
    MemValue(pte)   => {
      // Extend to 64 bits even on RV32 for simplicity.
      let pte : bits(64) = zero_extend(pte);

      let pte_flags = Mk_PTE_Flags(pte[7 .. 0]);
```

Tbh I think we should probably try to fully use Sail's fancy type system in this new VM code rather than just expanding everything to 64 bits immediately. It's very useful for catching bugs like this.

Timmmm:
For Sv32 Page Table Entries are only 4 bytes, but the old code was unconditionally reading and writing 8 bytes.

Fixes #459

### DIFF
--- a/model/riscv_vmem.sail
+++ b/model/riscv_vmem.sail
@@ -107,7 +107,7 @@ function pt_walk(sv_params,
   let mem_result = mem_read_priv(Read(Data),              // AccessType
                                  Supervisor,              // Privilege
                                  pte_phys_addr,
-                                 8,                       // atom (8)
+                                 2 ^ sv_params.log_pte_size_bytes,
                                  false,                   // aq
                                  false,                   // rl
                                  false);                  // res
@@ -115,6 +115,9 @@ function pt_walk(sv_params,
   match mem_result {
     MemException(_) => PTW_Failure(PTW_Access(), ext_ptw),
     MemValue(pte)   => {
+      // Extend to 64 bits even on RV32 for simplicity.
+      let pte : bits(64) = zero_extend(pte);
+
       let pte_flags = Mk_PTE_Flags(pte[7 .. 0]);
       if pte_is_invalid(pte_flags) then
         PTW_Failure(PTW_Invalid_PTE(), ext_ptw)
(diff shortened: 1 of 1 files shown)


Timmmm:
Actually it's worse than I thought; it also *writes* 8 bytes for Sv32 when updating the accessed/dirty bits. This pretty much means Sv32 is broken at the moment.

jrtc27:
How was the original PR tested if we’re finding this now?

Alasdair:
Are you planning on making a separate PR for the writes, or adjust this one? Otherwise LGTM, and we should get this fix in quickly.

Alasdair:
> How was the original PR tested if we’re finding this now?

We ran it on the tests in this repo, which I thought does include RV32 with virtual memory enabled. We also had a member of the golden model meeting run it on an extra set of virtual memory tests their company had, and we re-ran the basic 'boot FreeBSD' smoke test. It might be the case that there are no tests that actually do anything with the PTE dirty bits though.

jrtc27:
> > How was the original PR tested if we’re finding this now?
> 
> We ran it on the tests in this repo, which I thought does include RV32 with virtual memory enabled. We also had a member of the golden model meeting run it on an extra set of virtual memory tests their company had, and we re-ran the basic 'boot FreeBSD' smoke test. It might be the case that there are no tests that actually do anything with the PTE dirty bits though.

--enable-dirty-update is off by default

Timmmm:
> Are you planning on making a separate PR for the writes, or adjust this one?

I'll update this one.

Timmmm:
Ok I updated it to include the writes too. I am also working on refactoring the virtual memory code to use Sail's type system properly in all of this code, rather than resorting to `bit(64)`.

PR #297, opened 2023-08-24
Add Svinval extension.

kristinbarber:
These changes add the “Svinval” Standard Extension for Fine-Grained
    Address-Translation Cache Invalidation, Version 1.0 to the sail-riscv model.

    This extension defines five new instructions: SINVAL.VMA, SFENCE.W.INVAL,
    SFENCE.INVAL.IR, HINVAL.VVMA, HINVAL.GVMA.

    HINVAL.VVMA & HINVAL.GVMA are omitted since they build on the
    Hypervisor Extension which is yet to be included in the model.

    SFENCE.W.INVAL & SFENCE.INVAL.IR are treated as nops pending integration
    of the coherency model (rmem) with sail.

    The specification says that SINVAL.VMA behaves just as SFENCE.VMA,
    except there are additional ordering constraints with respect to the new
    SFENCE.W.INVAL & SFENCE.INVAL.IR instructions. Since these are nops, we
    can treat SINVAL.VMA as if it were SFENCE.VMA.

PR #463, opened 2024-04-29
Add Svinval extension.

martinberger:
These changes add the "Svinval" Standard Extension for Fine-Grained
Address-Translation Cache Invalidation, Version 1.0 to the sail-riscv model.
    
This extension defines five new instructions: SINVAL.VMA, SFENCE.W.INVAL,
SFENCE.INVAL.IR, HINVAL.VVMA, HINVAL.GVMA.
    
HINVAL.VVMA & HINVAL.GVMA are omitted since they build on the
Hypervisor Extension which is yet to be included in the model.
    
SFENCE.W.INVAL & SFENCE.INVAL.IR are treated as nops pending integration
of the coherency model (rmem) with sail.
    
The specification says that SINVAL.VMA behaves just as SFENCE.VMA,
except there are additional ordering constraints with respect to the new
SFENCE.W.INVAL & SFENCE.INVAL.IR instructions. Since these are nops, we
can treat SINVAL.VMA as if it were SFENCE.VMA.


martinberger:
This is essentially the same PR as https://github.com/riscv/sail-riscv/pull/297 from @kristinbarber but I've added the missing haveSvinval() gates (and configuration support) in the `encdec` functions.

martinberger:
One question I have is: should the new commands for Svinval be in a separate riscv_insts_svinval.sail file?

jrtc27:
> One question I have is: should the new commands for Svinval be in a separate `riscv_insts_svinval.sail` file?

Probably. One extension per file is an easy clear-cut rule (until you ask "which extension?") rather than having to debate it every time.

If you've copied code you should have Co-authored-by: as a trailer in your commit message.

c_emulator/riscv_sim.c
```diff
@@ -53,6 +53,7 @@ const char *RV32ISA = "RV32IMAC";
 #define OPT_ENABLE_WRITABLE_FIOM 1001
 #define OPT_PMP_COUNT 1002
 #define OPT_PMP_GRAIN 1003
+#define OPT_ENABLE_SVINVAL 10017
```
arichardson, line 56: Any reason to pick 17 here? Not that it matters, but it is easier to just keep incrementing by one? I missed that with the ZCB since it looked like 3 -> 4 but was actually 3 -> 14.

arichardson: I don't think they need to be consecutive considering how long it takes to merge even trivial PRs to this repository, but we should keep them in increasing order to see which ones are available.

arichardson (commented):
The original PR had comments explaining that implementing these as NOPs is legal for the model, which I think probably makes sense to keep.

Other than this and the one minor comment on the `#define` this looks good to me.

c_emulator/riscv_sim.c
jrtc27: One could make this an enum with the first value assigned to 1000 so that each subsequent entry is automatically one more than the previous

jrtc27: (Not for this PR, but as future cleanup; or one could not care given it's mostly going to be obsoleted by a proper -march parser)

arichardson: Agreed that there is no need to spend much time on this since it will be removed eventually. I like the enum idea though since it prevents mistakes and mismerges.

martinberger: > Any reason to pick 17 here? Not that it matters, but it is easier to just keep incrementing by one? I missed that with the ZCB since it looked like 3 -> 4 but was actually 3 -> 14.

I pulled 100017 out of thin air, because I could not see any logic in the existing numbers.  I agree that an enum would be better.

martinberger: Changed 10017 to 1004

martinberger:
> The original PR had comments explaining that implementing these as NOPs is legal for the model, which I think probably makes sense to keep.

Added short comments about NOPs

martinberger:
> If you've copied code you should have Co-authored-by: as a trailer in your commit message.

Added Kristin as coauthor.

Alasdair (approved):
Since there isn't much in terms of implementation, I checked the encoding of each instruction against the PDF spec and they look fine.

Once we merge this, I can add the cache invalidate operations that actually interface with our weak memory tooling (once my RVWMO PR is merged), although we don't have a particularly clear semantics for these operations they would at least be visible in our tooling.

Alasdair:
Would have been better to squash merge this so all the fixup commits weren't visible I think.

Issue #399, opened 2024-02-02
Make file error


Ludini1:
When running make, I get the following error:

`File "generated_definitions/lem/RV64/riscv.lem", line 346, character 35 to line 346, character 48
  Type error: unbound variable: parse_hex_bits
make: *** [Makefile:327: generated_definitions/isabelle/RV64/Riscv.thy] Error 1`

I'm using WSL so I already installed the latest development versions of Z3 and Sail but it resulted in the same error anyway.

bacam:
The theorem prover backends have fallen a little behind and don't yet have `parse_hex_bits`.  If you don't need Isabelle output I suggest just making the target you want, probably `c_emulator/riscv_sim_RV64` or `c_emulator/riscv_sim_RV32`.  (It would be nice if these were the default, but unfortunately #329 appears to have stalled.)

shareefj:
Just for anyone else flailing around like I was, the correct make command seems to be:

```
ARCH=32 make c_emulator/riscv_sim_RV32
```

or s/32/64/g if that's your bag.

Incarnation-p-lee:
Just FYI, still have this issue in:
```
[NOTE] Package sail is already installed (current version is 0.17.1).
```

shareefj:
> Just FYI, still have this issue in:
> 
> ```
> [NOTE] Package sail is already installed (current version is 0.17.1).
> ```

If you uninstall the package, then re-install it, that message will go away.  ;-)

Incarnation-p-lee:
Take below step(s), not sure if I missed something or you don't mean the error message.
```
opam remove sail
opam clean
opam install sail -y
  ⬇ retrieved sail.0.17.1  (https://opam.ocaml.org/cache)
  ∗ installed sail.0.17.1 
  Done.
ARCH=RV64 make
```
then still have the error message as below
```
File "generated_definitions/lem/RV64/riscv.lem", line 346, character 35 to line 346, character 48
  Type error: unbound variable: parse_hex_bits
make: *** [Makefile:329: generated_definitions/isabelle/RV64/Riscv.thy] Error 1
```

Alasdair:
I will look into making a PR to fix this later this week. I still think in general the Makefile is trying to build too many targets by default including some targets that are inherently a little brittle.

Incarnation-p-lee:
> I will look into making a PR to fix this later this week. I still think in general the Makefile is trying to build too many targets by default including some targets that are inherently a little brittle.

That's Great!

[...]

PR #464, opened 2024-04-30
Remove theorem prover targets from default Makefile rule

Alasdair:
In general we aren't requiring contributors to implement the correct Lem/Isabelle/HOL4/Coq stubs for new extensions (this would almost certainly be way too high a bar for new contributors anyway) so having these in the default set of build targets just means that typing 'make' is broken until those of us who are invested in maintaining those targets can add updates for those stubs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ RISCV_EXTRAS_LEM = $(addprefix handwritten_support/,$(RISCV_EXTRAS_LEM_FILES))
 
 .PHONY:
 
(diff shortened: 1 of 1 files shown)


Alasdair:
Note, this fixes: https://github.com/riscv/sail-riscv/issues/399 Also, we should merge my PRs that add these stubs: https://github.com/riscv/sail-riscv/pull/444

Anyone who objects to this I will happily forward all the 'why is the sail-riscv build broken' emails I get, and you can respond to them instead 😄

Timmmm (approved):

arichardson (approved):

PR #466, opened 2024-05-09
Refactor MUL instruction

Alasdair:
It was brought to my attention that this instruction had a bit of a case of the 'boolean blindness' code smell, where the mul operation was represented as a triple of booleans. This commit refactors the implemention to use a struct with named fields for high, signed1, and signed2.

The C_MUL instruction in Zcb also needs to be changed appropriately

The mul_op struct was added in riscv_types

While I was there I did some light housekeeping w.r.t a comment about a workaround for Sail < 0.15.1, as this is no longer needed. I could split this into a separate commit, but not sure if it's worthwhile to do so.

### DIFF
--- a/model/riscv_insts_mext.sail
+++ b/model/riscv_insts_mext.sail
@@ -11,27 +11,26 @@
 
 /* ****************************************************************** */
 
-union clause ast = MUL : (regidx, regidx, regidx, bool, bool, bool)
+union clause ast = MUL : (regidx, regidx, regidx, mul_op)
 
-mapping encdec_mul_op : (bool, bool, bool) <-> bits(3) = {
-  (false, true, true)  <-> 0b000,
-  (true, true, true)   <-> 0b001,
-  (true, true, false)  <-> 0b010,
-  (true, false, false) <-> 0b011
+mapping encdec_mul_op : mul_op <-> bits(3) = {
+  struct { high = false, signed_rs1 = true,  signed_rs2 = true  } <-> 0b000,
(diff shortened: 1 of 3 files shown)


model/riscv_insts_mext.sail
```diff
@@ -11,27 +11,26 @@
...
-union clause ast = MUL : (regidx, regidx, regidx, bool, bool, bool)
+union clause ast = MUL : (regidx, regidx, regidx, mul_op)
 
-mapping encdec_mul_op : (bool, bool, bool) <-> bits(3) = {
-  (false, true, true)  <-> 0b000,
-  (true, true, true)   <-> 0b001,
-  (true, true, false)  <-> 0b010,
-  (true, false, false) <-> 0b011
+mapping encdec_mul_op : mul_op <-> bits(3) = {
+  struct { high = false, signed1 = true,  signed2 = true  } <-> 0b000,
```
arichardson, line 17: Maybe call these signed_rs1/signed_rs2?

arichardson (approved):
Looks much nicer!

Timmmm (approved):

model/riscv_insts_mext.sail
billmcspadden-riscv: Can we resolve this conversation?  If so,  then I'll merge.

arichardson: This was not intended as a blocking comment, just a possible clarification. I'll let @Alasdair decide whether to make the change or not.

Alasdair:
I made the suggested change so `signed1` is now `signed_rs1` and likewise for `signed2`, as well as rebased to the latest commit.

PR #468, opened 2024-05-15
Change ext_data_get_addr to use bytes for width

Timmmm:
Instead of `word_width` which can only be up to 8 bytes, just use bytes. This allows larger accesses (the limit is increased to 4096), e.g. for `cbo.zero`.

This is a pre-requisite for #455 but should be uncontroversial so I think we can merge it first.

It will require a corresponding (easy) change to the CHERI repo.


arichardson (approved):

billmcspadden-riscv (approved):
LGTM

PR #470, opened 2024-05-15
Use bool <-> bit mappings from riscv_types.sail

Timmmm:
Use these mappings and change the other functions to be aliases of them.

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -99,17 +99,27 @@ overload zeros = {zeros_implicit}
 val ones : forall 'n, 'n >= 0 . implicit('n) -> bits('n)
 function ones (n) = sail_ones (n)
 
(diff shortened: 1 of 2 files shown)


arichardson (commented):
I feel like the old names made it more obvious which way the conversion goes but happy with either

Timmmm:
> I feel like the old names made it more obvious which way the conversion goes but happy with either

Yeah I agree. Kind of hard to name these mappings.

Timmmm:
I think on reflection I'll add the original functions back as aliases to the mapping, i.e.

```
function bool_to_bits(...) = bool_bits(...)
function bits_to_bool(...) = bool_bits(...)
```

The names are definitely clearer and it will make this change smaller.

arichardson (approved):

PR #471, opened 2024-05-15
Check misalignment of AMOs before address translation

Timmmm:
This is optional according to the spec - you can check afterwards. However

1. it seems extremely unlikely that any real designs will do that for atomics, which (ignoring Zam which the model doesn't support yet), always have to be aligned, and
2. the LR and SC instructions already check before address translation, so this wasn't even consistent.

Ideally in future this would be configurable.

This also includes a minor refactor to reuse the existing alignment checking code.

### DIFF
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -180,7 +180,9 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
   match ext_data_get_addr(rs1, zeros(), ReadWrite(Data, Data), width_bytes) {
     Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
     Ext_DataAddr_OK(vaddr) => {
-      match translateAddr(vaddr, ReadWrite(Data, Data)) {
+      if not(is_aligned(vaddr, width))
+      then { handle_mem_exception(vaddr, E_SAMO_Addr_Align()); RETIRE_FAIL }
+      else match translateAddr(vaddr, ReadWrite(Data, Data)) {
         TR_Failure(e, _) => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
         TR_Address(addr, _) => {
           let eares = mem_write_ea(addr, width_bytes, aq & rl, rl, true);


arichardson (approved):

model/riscv_insts_base.sail
```diff
@@ -313,14 +313,17 @@ function process_load(rd, vaddr, value, is_unsigned) =
     MemException(e)  => { handle_mem_exception(vaddr, e); RETIRE_FAIL }
   }
 
+function is_aligned(vaddr : xlenbits, width : word_width) -> bool =
+  match width {
+    BYTE   => true,
+    HALF   => vaddr[0..0] == zeros(),
```
Mudassir10X, line 319: I think the vaddr[0] should be correct instead of [0..0]

jrtc27: I assume the problem is bit and bits(1) are not the same thing, and this keeps it consistent with the other cases.

Timmmm: Yeah exactly. Ideally I would do this:

```
function is_aligned(vaddr : xlenbits, width : word_width) -> bool =
   vaddr[word_width_bytes(word_width) >.. 0] == zeros()
```

However `>..` isn't supported yet - see [this issue](https://github.com/rems-project/sail/issues/471).

It definitely looks a bit weird to have `[N .. N]`. As well as not being able to do `bitone == 0b1` you also can't do `bitone @ 0b` which leads to weird looking code like this:

```
sign_extend(~(addr[38 .. 38]) @ addr[38 .. 0])
```

Sail does actually support overloading the operators so you can then do this:

```
function prepend_bit forall 'n, 'n >= 1. (b : bit, rest : bits('n)) -> bits('n + 1) =
  (if b == bitone then 0b1 else 0b0) @ rest

function append_bit forall 'n, 'n >= 1. (rest : bits('n), b : bit) -> bits('n + 1) =
  rest @ (if b == bitone then 0b1 else 0b0)

function join_bits(b0 : bit, b1 : bit) -> bits(2) =
  (if b0 == bitone then 0b1 else 0b0) @ (if b1 == bitone then 0b1 else 0b0)

overload append = {prepend_bit, append_bit, join_bits}
```

and then this works:

```
sign_extend(~(addr[38]) @ addr[38 .. 0])
```

which I think from a readability point of view is a good improvement, but I'm not sure about type safety. It's probably fine.

Alasdair: At some point I hope to make `bit` and `bits(1)` the same type. I've almost reached the point where that's possible.

In the past (a long time ago) we just had a polymorphic vector type, and a bitvector would be a vector containing bits `vector('n, dec, bit)`. That turned out to be a mistake, as it made code generation very tricky - we nearly always want to represent bitvectors differently to other vectors. To fix that I split the type into a `vector` and `bitvector` type. However there was also the order parameter that determined the indexing order. This stopped me from being able to just make `bit` into `bits(1)` because actually behind the scenes we would have either `bitvector(inc, 1)` and `bitvector(dec, 1)`. I have subsequently removed the order parameter so now it should be possible...

Timmmm: Ah interesting! I was under the impression that `bits(3)` was still the same as `vector(4, bit)`... but now that I reread the docs you do explicitly say they are different.

Making `bit` and `bits(1)` the same in Sail would be great. I guess that would mean you can eliminate `bitone` and `bitzero` too? In fact... do you need `bit` at all?

Mudassir10X (approved):

Timmmm:
I'll merge this tomorrow if nobody has any objections...

PR #473, opened 2024-05-15
Remove duplicate type declarations for mem_read

Timmmm:
These used to be different because they had different effects annotations, but those have since been removed and now they are the same.

### DIFF
--- a/model/riscv_mem.sail
+++ b/model/riscv_mem.sail
@@ -136,18 +136,10 @@ val rvfi_read : forall 'n, 'n > 0. (xlenbits, int('n), MemoryOpResult((bits(8 * 
 function rvfi_read (addr, width, result) = ()
 $endif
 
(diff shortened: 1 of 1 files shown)


model/riscv_mem.sail
```diff
@@ -137,17 +137,10 @@ function rvfi_read (addr, width, result) = ()
 $endif
 
 /* NOTE: The rreg effect is due to MMIO. */
```
arichardson, line 139: Can also remove this comment. Maybe that was the previous difference that is no longer required?

arichardson (approved):

model/riscv_mem.sail
Timmmm: :+1: done

billmcspadden-riscv (approved):
Hmmm.... I thought this got removed quite a while ago.  Bah.

LGTM

PR #475, opened 2024-05-15
Add some missing explicit var declarations

Timmmm:
Implicit `var` declarations will eventually be an error. This makes some implicit `var` declarations explicit.


arichardson (approved):

Alasdair (approved):

Alasdair:
I need to do a new Sail release soon. Mostly for bug fixes but it will include the `--strict-var` option that can check for this.

PR #477, opened 2024-05-15
Remove unnecessary matches for loads/stores

Timmmm:
Simplify the load/store code by removing the unnecessary matches on the access size. I have not done it for float loads/stores because that requires a separate change to make the float code generic over size.

Note, the first commit is #468 - only review the second.

### DIFF
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -53,46 +53,31 @@ mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)                          
  * call to load_reservation in LR and cancel_reservation in SC.
  */
 
-val process_loadres : forall 'n, 0 < 'n <= xlen_bytes. (regidx, xlenbits, MemoryOpResult(bits(8 * 'n)), bool) -> Retired
-function process_loadres(rd, addr, value, is_unsigned) =
-  match extend_value(is_unsigned, value) {
-    MemValue(result) => { load_reservation(addr); X(rd) = result; RETIRE_SUCCESS },
-    MemException(e)  => { handle_mem_exception(addr, e); RETIRE_FAIL }
-  }
-
 function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
+  let width_bytes = size_bytes(width);
+
+  // This is checked during decoding.
+  assert(width_bytes <= sizeof(xlen_bytes));
+
   /* Get the address, X(rs1) (no offset).
     * Extensions might perform additional checks on address validity.
     */
-  match ext_data_get_addr(rs1, zeros(), Read(Data), size_bytes(width)) {
+  match ext_data_get_addr(rs1, zeros(), Read(Data), width_bytes) {
     Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
     Ext_DataAddr_OK(vaddr) => {
-      let aligned : bool =
-          /* BYTE and HALF would only occur due to invalid decodes, but it doesn't hurt
-          * to treat them as valid here; otherwise we'd need to throw an internal_error.
-          */
-          match width {
-            BYTE   => true,
-            HALF   => vaddr[0..0] == 0b0,
-            WORD   => vaddr[1..0] == 0b00,
-            DOUBLE => vaddr[2..0] == 0b000
-          };
       /* "LR faults like a normal load, even though it's in the AMO major opcode space."
         * - Andrew Waterman, isa-dev, 10 Jul 2018.
         */
-      if not(aligned)
+      if not(is_aligned(vaddr, width))
       then { handle_mem_exception(vaddr, E_Load_Addr_Align()); RETIRE_FAIL }
       else match translateAddr(vaddr, Read(Data)) {
-              TR_Failure(e, _)    => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
-              TR_Address(addr, _) =>
-                match (width, sizeof(xlen)) {
-                  (BYTE, _)    => process_loadres(rd, vaddr, mem_read(Read(Data), addr, 1, aq, aq & rl, true), false),
-                  (HALF, _)    => process_loadres(rd, vaddr, mem_read(Read(Data), addr, 2, aq, aq & rl, true), false),
-                  (WORD, _)    => process_loadres(rd, vaddr, mem_read(Read(Data), addr, 4, aq, aq & rl, true), false),
-                  (DOUBLE, 64) => process_loadres(rd, vaddr, mem_read(Read(Data), addr, 8, aq, aq & rl, true), false),
-                  _            => internal_error(__FILE__, __LINE__, "Unexpected AMO width")
-                }
-            }
+        TR_Failure(e, _)    => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
+        TR_Address(addr, _) =>
+          match mem_read(Read(Data), addr, width_bytes, aq, aq & rl, true) {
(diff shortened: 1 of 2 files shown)


model/riscv_addr_checks.sail
```diff
@@ -54,7 +54,7 @@ type ext_data_addr_error = unit
 
 /* Default data addr is just base register + immediate offset (may be zero).
    Extensions might override and add additional checks. */
-function ext_data_get_addr(base : regidx, offset : xlenbits, acc : AccessType(ext_access_type), width : word_width)
+function ext_data_get_addr(base : regidx, offset : xlenbits, acc : AccessType(ext_access_type), width : range(0, max_mem_access))
```
jrtc27, line 57: 0?

model/riscv_insts_aext.sail
```diff
@@ -53,47 +53,32 @@ mapping clause encdec = LOADRES(aq, rl, rs1, size, rd) if amo_width_valid(size)
...
+
+  // This is checked during decoding.
+  assert(width_bytes <= sizeof(xlen_bytes));
+
   if haveAtomics() then {
     /* Get the address, X(rs1) (no offset).
      * Extensions might perform additional checks on address validity.
      */
-    match ext_data_get_addr(rs1, zeros(), Read(Data), width) {
+    match ext_data_get_addr(rs1, zeros(), Read(Data), size_bytes(width)) {
```
jrtc27, line 66: Use width_bytes?

model/riscv_insts_fext.sail
```diff
@@ -368,45 +336,33 @@ mapping clause encdec = STORE_FP(imm7 @ imm5, rs2, rs1, DOUBLE)
...
-    MemValue(true)  => { RETIRE_SUCCESS },
-    MemValue(false) => { internal_error(__FILE__, __LINE__, "store got false from mem_write_value") },
-    MemException(e) => { handle_mem_exception(vaddr, e); RETIRE_FAIL }
-  }
-
 function clause execute (STORE_FP(imm, rs2, rs1, width)) = {
+  let width_bytes = size_bytes(width);
+
+  // This is checked during decoding.
+  assert(width_bytes <= sizeof(xlen_bytes)); // TODO: Should this actually be flen?
```
jrtc27, line 343: Yes? That's what was being checked in the code you deleted

model/riscv_insts_fext.sail
```diff
@@ -282,59 +282,27 @@ mapping clause encdec = LOAD_FP(imm, rs1, rd, DOUBLE)        if haveDExt()
...
 
-val process_fload32 : (regidx, xlenbits, MemoryOpResult(bits(32)))
-                      -> Retired
-function process_fload32(rd, addr, value) =
-  match value {
-    MemValue(result) => { F(rd) = nan_box(result); RETIRE_SUCCESS },
-    MemException(e)  => { handle_mem_exception(addr, e); RETIRE_FAIL }
-  }
+  // This is checked during decoding.
+  assert(width_bytes <= sizeof(xlen_bytes));
```
jrtc27, line 289: flen

model/riscv_insts_aext.sail
```diff
@@ -53,47 +53,32 @@ mapping clause encdec = LOADRES(aq, rl, rs1, size, rd) if amo_width_valid(size)
...
-                   (HALF, _)    => process_loadres(rd, vaddr, mem_read(Read(Data), addr, 2, aq, aq & rl, true), false),
-                   (WORD, _)    => process_loadres(rd, vaddr, mem_read(Read(Data), addr, 4, aq, aq & rl, true), false),
-                   (DOUBLE, 64) => process_loadres(rd, vaddr, mem_read(Read(Data), addr, 8, aq, aq & rl, true), false),
-                   _            => internal_error(__FILE__, __LINE__, "Unexpected AMO width")
-                 }
-             }
+          TR_Failure(e, _)    => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
+          TR_Address(addr, _) =>
+            match mem_read(Read(Data), addr, 1, aq, aq & rl, true) {
+              MemValue(result) => { load_reservation(addr); X(rd) = zero_extend(result); RETIRE_SUCCESS },
```
jrtc27, line 78: sign_extend

model/riscv_insts_aext.sail
```diff
@@ -269,20 +225,9 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
...
-                    };
-                    let wval : MemoryOpResult(bool) = match (width, sizeof(xlen)) {
-                      (BYTE, _)    => mem_write_value(addr, 1, result[7..0],  aq & rl, rl, true),
-                      (HALF, _)    => mem_write_value(addr, 2, result[15..0], aq & rl, rl, true),
-                      (WORD, _)    => mem_write_value(addr, 4, result[31..0], aq & rl, rl, true),
-                      (DOUBLE, 64) => mem_write_value(addr, 8, result,        aq & rl, rl, true),
-                      _            => internal_error(__FILE__, __LINE__, "Unexpected AMO width")
-                    };
-                    match (wval) {
+                    let rval : xlenbits = sign_extend(loaded[width_bytes * 8 - 1 .. 0]);
```
jrtc27, line 228: Extending loaded on line 211 then re-extending it here is a bit confusing. Maybe just moving this to before line 211 would be enough to make it clearer? Or anything else you have in mind.

model/riscv_addr_checks.sail
Timmmm: Good point! Changed to 1.

model/riscv_insts_aext.sail
Timmmm: :+1: fixed.

model/riscv_insts_aext.sail
Timmmm: Oops, fixed.

model/riscv_insts_fext.sail
Timmmm: Ooops, slight Git mistake - I didn't actually intend to include the fext changes since they don't quite work yet. It only compiles because of the mistake on line 302 -  `2` should be `width_bytes`. With that change it can't compile because `nan_box` isn't generic (needs #448).

I removed these changes.

model/riscv_insts_aext.sail
Timmmm: Ah yeah I agree. This was just because I mechanically translated the code but after actually reading it I agree it was pretty confusing!

I've made an attempt at simplifying it by not extending the values and then later truncating them. I also changed the min/max ones to use `<_u` etc which I think is a bit nicer and plays better with the type system.

Timmmm (commented):
Thanks for the review!

arichardson (approved):
Not familiar with the vector parts but the rest looks like a very nice cleanup.

Timmmm:
This is ready now - I rebased it and fixed a small bug (mixing up `vaddr` and `addr` - see #434 ).

Timmmm:
@billmcspadden-riscv I resolved the threads but I think we should merge #468 first (which is just the first commit from this PR) and then I'll rebase this and we can merge it. Hope that makes sense!

Timmmm:
Ok this is rebased and ready to merge.

rmn30:
IIRC those matches exist because previous Sail versions had trouble with monomorphisation of this code when generating theorem prover definitions. No idea if that is still the case.

Timmmm:
Ah interesting. @Alasdair do you know if that is still true?

bacam:
I don't think any of the output that's currently produced requires monomorphisation (because the Lem/Isabelle output is currently using bit lists).  If I remember correctly, this part of the spec isn't a problem anyway at the moment, but the vector specifications need some tweaks.

PR #479, opened 2024-05-18
Remove unused (f)reg_name_abi functions

jordancarlin:
There seem to be duplicate functions right now for reg and freg names, (f)reg_name_abi and (f)reg_name. These mappings are the same and only (f)reg_name is used, so remove the unused ones.

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -327,47 +327,6 @@ overload F_or_X_H = { rF_or_X_H, wF_or_X_H }
 overload F_or_X_S = { rF_or_X_S, wF_or_X_S }
 overload F_or_X_D = { rF_or_X_D, wF_or_X_D }
 
-/* register names */
-
-val freg_name_abi : regidx <-> string
-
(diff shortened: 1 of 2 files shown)


Timmmm:
I presume the original intent was so that you could assemble e.g. `add a0, x10, 5`, but I went back to the 5 year old commit that added them and they've always been the same so I don't see any reason to keep them.

Also I feel like `to_str()` needs to be fixed anyway. Its overloads include both directions of all the mappings so depending on the order in which you compile files `to_str(0b00000)` can be `zero` or `f0` which is a bit weird. We should probably change the `to_str()` overloads to use `reg_name_backwards`, etc. instead of `reg_name`.

Anyway this looks fine.

Timmmm (approved):

jordancarlin:
Exactly. And the `to_str()` overload for registers is not currently used in the assembly clauses, `reg_name()` is being used directly instead.

Timmmm:
@billmcspadden-riscv I'll merge this if you have no objections?

PR #481, opened 2024-05-20
Sail Crosscheck Testing with ACT Against Spike

Abdulwadoodd:
This PR should complete [this](https://github.com/riscv-admin/dev-partners/issues/38) SOW.

PR adds the RISCOF-based CI infrastructure to test sail-riscv against Spike using ACTs. Both Sail and Spike are configured using ISA yaml configuration (provided under `ci-tests/riscof/`) which are validated by riscv-config using RISCOF

Current infrastructure uses [3.8.10](https://github.com/riscv-non-isa/riscv-arch-test/releases/tag/3.8.10) release of arch-test-suites. The later releases of arch-tests (that add Zfa and Zfh tests) are partially broken for RISCOF - and should be fixed shortly. 

CI runs ACTS in RV32I, RV64I, and RV32E configurations. Details are as follows. After successful completion, test reports for each configuration are uploaded to GitHub Workspace.  

Kindly review, and make suggestions how the CI should be pulled in to this repo. I'll be happy to make the requested changes.

### DIFF
--- a/build_simulators.sh
+++ b/build_simulators.sh
@@ -10,8 +10,8 @@ function test_build () {
     fi
 }
 
-test_build make ARCH=RV32 ocaml_emulator/riscv_ocaml_sim_RV32
-test_build make ARCH=RV64 ocaml_emulator/riscv_ocaml_sim_RV64
+test_build make -j"$(nproc 2> /dev/null || sysctl -n hw.ncpu)" ARCH=RV32 ocaml_emulator/riscv_ocaml_sim_RV32
+test_build make -j"$(nproc 2> /dev/null || sysctl -n hw.ncpu)" ARCH=RV64 ocaml_emulator/riscv_ocaml_sim_RV64
 
-test_build make ARCH=RV32 c_emulator/riscv_sim_RV32
-test_build make ARCH=RV64 c_emulator/riscv_sim_RV64
+test_build make -j"$(nproc 2> /dev/null || sysctl -n hw.ncpu)" ARCH=RV32 c_emulator/riscv_sim_RV32
+test_build make -j"$(nproc 2> /dev/null || sysctl -n hw.ncpu)" ARCH=RV64 c_emulator/riscv_sim_RV64
--- a/ci-tests/riscof/run-tests.sh
+++ b/ci-tests/riscof/run-tests.sh
@@ -1,0 +1,23 @@
+# !bin/bash
+
+set -e
+
+# Check if argument (work directory name) provided
+if [ $# -eq 0 ]; then
+    echo "Usage: $0 <argument>"
+    exit 1
+fi
+
+riscof -v debug run --config=config.ini \
+           --suite=riscv-arch-test/riscv-test-suite/ \
+           --env=riscv-arch-test/riscv-test-suite/env \
+           --no-browser --work-dir "$1"
+
+if grep -rniq "$1"/report.html -e '>0failed<'
+then
+    echo "Test successful!"
+    exit 0
+else
+    echo "Test FAILED!"
+    exit 1
+fi
--- a/ci-tests/riscof/sail_cSim/riscof_sail_cSim.py
+++ b/ci-tests/riscof/sail_cSim/riscof_sail_cSim.py
@@ -1,0 +1,146 @@
+import os
+import re
+import shutil
+import subprocess
+import shlex
+import logging
+import random
+import string
+from string import Template
+
+import riscof.utils as utils
+from riscof.pluginTemplate import pluginTemplate
+import riscof.constants as constants
+from riscv_isac.isac import isac
+
+logger = logging.getLogger()
+
+class sail_cSim(pluginTemplate):
+    __model__ = "sail_c_simulator"
+    __version__ = "0.5.0"
+
+    def __init__(self, *args, **kwargs):
+        sclass = super().__init__(*args, **kwargs)
+
+        config = kwargs.get('config')
+        if config is None:
+            logger.error("Config node for sail_cSim missing.")
+            raise SystemExit(1)
+        self.num_jobs = str(config['jobs'] if 'jobs' in config else 1)
+        self.pluginpath = os.path.abspath(config['pluginpath'])
+        self.sail_exe = { '32' : os.path.join(config['PATH'] if 'PATH' in config else "","riscv_sim_RV32"),
+                '64' : os.path.join(config['PATH'] if 'PATH' in config else "","riscv_sim_RV64")}
+        self.isa_spec = os.path.abspath(config['ispec']) if 'ispec' in config else ''
+        self.platform_spec = os.path.abspath(config['pspec']) if 'ispec' in config else ''
+        self.make = config['make'] if 'make' in config else 'make'
+        logger.debug("SAIL CSim plugin initialised using the following configuration.")
+        for entry in config:
+            logger.debug(entry+' : '+config[entry])
+        return sclass
+
+    def initialise(self, suite, work_dir, archtest_env):
+        self.suite = suite
+        self.work_dir = work_dir
+        self.objdump_cmd = 'riscv64-unknown-elf-objdump -D {0} > {1};'
+        # self.archtest_env = archtest_env.replace("riscv-arch-test", "../my-repo/riscv-arch-test")
+        self.compile_cmd = 'riscv64-unknown-elf-gcc -march={0} \
+         -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles\
+         -T '+self.pluginpath+'/env/link.ld\
+         -I '+self.pluginpath+'/env/\
+         -I ' + archtest_env
+
+        # workaround to avoid clang format error in sail-riscv CI.
+        # *clang suggest style that is syntactically incorrect*
+        # Insert the following macro at run-time.
+        modelTest_path = 'sail_cSim/env/model_test.h'
+        macro = """
+#define RVMODEL_DATA_SECTION .pushsection                                      \\
+  .tohost, "aw", @progbits;                                                    \\
+  .align 8;                                                                    \\
+  .global tohost;                                                              \\
+  tohost:                                                                      \\
+  .dword 0;                                                                    \\
+  .align 8;                                                                    \\
+  .global fromhost;                                                            \\
+  fromhost:                                                                    \\
+  .dword 0;                                                                    \\
+  .popsection;                                                                 \\
+  .align 8;                                                                    \\
+  .global begin_regstate;                                                      \\
+  begin_regstate:                                                              \\
+  .word 128;                                                                   \\
+  .align 8;                                                                    \\
+  .global end_regstate;                                                        \\
+  end_regstate:                                                                \\
+  .word 4;
+"""
+        with open(modelTest_path, 'r') as file:
+            modelTest = file.readlines()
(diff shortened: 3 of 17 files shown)


jrtc27:
Most of this looks like it belongs in a centralised location so code can be reused?

jrtc27:
That or each project should put its riscof config in a place that other projects can use. The Sail model shouldn’t have to have a copy of how to build tests for Spike.

Abdulwadoodd:
> this looks like it it belongs in a centralised location

Not exactly sure what this means but most of the code (python plugins, linkers, model-specific headers or yaml, etc.) here is either produced while setting up RISCOF (configuring Spike as DUT and SAIL as reference model) or added/modified manually as per the requirement by following the [RISCOF documentation](https://riscof.readthedocs.io/en/latest/intro.html).

> Each project should put its riscof config in a place that other projects can use

What do you mean by "project" here?

jrtc27:
> > this looks like it it belongs in a centralised location
> 
> Not exactly sure what this means but most of the code (python plugins, linkers, model-specific headers or yaml, etc.) here is either produced while setting up RISCOF (configuring Spike as DUT and SAIL as reference model) or added/modified manually as per the requirement by following the [RISCOF documentation](https://riscof.readthedocs.io/en/latest/intro.html).

I don't really care about the background, I care about the end result, which does not look like good software engineering to me.

> > Each project should put its riscof config in a place that other projects can use
> 
> What do you mean by "project" here?

The Sail model, the Spike simulator, etc.

ci-tests/riscof/run-tests.sh
```diff
@@ -0,0 +1,23 @@
+# !bin/bash
```
Alasdair, line 1: space in between '#' and '!' here

ci-tests/riscof/spike/riscof_spike.py
```diff
@@ -0,0 +1,179 @@
...
+
+       self.compile_cmd = 'riscv64-unknown-elf-gcc -march={0} \
+         -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -g\
+         -T '+self.pluginpath+'/env/link.ld\
+         -I '+self.pluginpath+'/env/\
+         -I ' + archtest_env + ' {1} -o {2} {3}'
+
+       # workaround to avoid clang format error in sail-riscv CI.
+       # *clang suggest style that is syntactically incorrect*
+       # Insert the following macro at run-time.
```
Alasdair, line 60: why not just use clang-format-ignore, rather than work around it like this?

ci-tests/riscof/spike/riscof_spike.py
```diff
@@ -0,0 +1,179 @@
+import os
```
jrtc27, line 1: My point is, the files in this directory, especially this one, have absolutely nothing to do with the Sail model. They should not be copied to every project/repository/whatever that needs to use Spike, because that quickly becomes an unmaintainable mess where everybody has a differently-outdated version, and you end up with O(n^2) of these files rather than O(n) due to the cross-product between reference and DUT. There should be some way to only have one copy of "how do I build and run tests for platform X", and the most obvious (to me) way to achieve that is to put that code in platform X's repository at a stable published location.

ci-tests/riscof/sail_cSim/riscof_sail_cSim.py
```diff
@@ -0,0 +1,146 @@
...
+        for entry in config:
+            logger.debug(entry+' : '+config[entry])
+        return sclass
+
+    def initialise(self, suite, work_dir, archtest_env):
+        self.suite = suite
+        self.work_dir = work_dir
+        self.objdump_cmd = 'riscv64-unknown-elf-objdump -D {0} > {1};'
+        # self.archtest_env = archtest_env.replace("riscv-arch-test", "../my-repo/riscv-arch-test")
+        self.compile_cmd = 'riscv64-unknown-elf-gcc -march={0} \
```
jrtc27, line 46: I said it years ago for the previous attempt to add something like this to the Sail model and I'll say it again:

1. Do not hard-code GCC/binutils. Clang/LLVM should be just as well supported, and is arguably a better experience for the end user because they don't need to install a special cross-compiler, as every Clang build can compile for every architecture by default.
2. The differences between compiling for platforms X and Y are usually minimal, typically some linker script tweaks and providing model_test.h. That should be reflected in the software implementation by having common library code provided by RISCOF to deal with compilers etc, rather than duplicating all that for every single project. Each project should just provide the project-specific compilation information, as well as how to run the test on it.

ci-tests/riscof/sail_cSim/riscof_sail_cSim.py
```diff
@@ -0,0 +1,146 @@
...
+         -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles\
+         -T '+self.pluginpath+'/env/link.ld\
+         -I '+self.pluginpath+'/env/\
+         -I ' + archtest_env
+
+        # workaround to avoid clang format error in sail-riscv CI.
+        # *clang suggest style that is syntactically incorrect*
```
jrtc27, line 53: I don't know what it was complaining about, but you can suppress clang-format changes when needed. Please elaborate. But this is not the right approach.

Alasdair:
If we want to run ACT cross-checks against spike in our CI, which I think is a good idea, then at least some of the infrastructure for running that needs to be in this repository. While we could pull some of that in from a third-party repository, that isn't necessarily a good idea as it means the CI scripts/config would no-longer be versioned together with the model which creates a bunch of issues.

Overall, I think it would be better if the `ci-tests` directory was moved into the existing tests folder as a subdirectory and renamed (the existing tests are already CI tests!), maybe as `act_cross_check` or similar. I'd also change the formatting of `cSim` to just `csim`, I don't see any reason for the weird capitalisation of the the 'S'.

jrtc27:
> If we want to run ACT cross-checks against spike in our CI, which I think is a good idea, then at least some of the infrastructure for running that needs to be in this repository. While we could pull some of that in from a third-party repository, that isn't necessarily a good idea as it means the CI scripts/config would no-longer be versioned together with the model which creates a bunch of issues.
> 
> Overall, I think it would be better if the `ci-tests` directory was moved into the existing tests folder as a subdirectory and renamed (the existing tests are already CI tests!), maybe as `act_cross_check` or similar. I'd also change the formatting of `cSim` to just `csim`, I don't see any reason for the weird capitalisation of the the 'S'.

The Sail side should be, but the Spike side shouldn't, otherwise *those* files won't be versioned with Spike, creating the same kinds of issues you talk about.

ci-tests/riscof/spike/riscof_spike.py
Alasdair: I agree, but my understanding is the current RISCOF setup essentially expects the user to bring along a certain amount of boilerplate for both the thing you are testing and the thing you are testing against. I think we need to be careful that we review this PR as an attempt to improve our lack of testing by integrating RISCOF rather than look at larger issues in the RISC-V ecosystem that may not be easily actionable by the PR author.

ci-tests/riscof/sail_cSim/riscof_sail_cSim.py
arichardson: 100% agreed. These riscof generated files are awfully non-generic and the whole design of the framework makes no sense to me. Unfortunately there is no alternative so I guess we will have to include these files in the repository until there is a better solution. Maybe in some hidden .github directory?

jrtc27: It's been almost 3 years since I reviewed #97 and made the same comments. I don't see why we should accept poor code here just because people responsible for the RISC-V testing frameworks don't action feedback. All this tells me is that the only way to get it done is to push back hard and force them to, otherwise I have no faith that it will ever get fixed and instead just be declared as done and working despite the poor design.

To quote a popular adage, "poor planning on your part does not necessitate an emergency on mine".

jrtc27 (changes requested):
NAK. Nothing's changed since #97 3 years ago it seems.

ci-tests/riscof/sail_cSim/riscof_sail_cSim.py
Timmmm: We've run into this bug too. `clang-format` will sometimes reformat assembly into invalid code.

In this case it changes the first two lines to

```
#define RVMODEL_DATA_SECTION                                                   \
  .pushsection.tohost, "aw", @progbits;                                        \
```

which I guess is wrong. Anyway I agree, just use `// clang-format off` and `// clang-format on`.

ci-tests/riscof/sail_cSim/riscof_sail_cSim.py
```diff
@@ -0,0 +1,146 @@
...
+        with open(modelTest_path, 'w') as file:
+            file.writelines(modelTest)
+
+    def build(self, isa_yaml, platform_yaml):
+        ispec = utils.load_yaml(isa_yaml)['hart0']
+        self.xlen = ('64' if 64 in ispec['supported_xlen'] else '32')
+        self.isa = ' '  # 'rv' + self.xlen
+        ilp32 = 'ilp32e ' if "E" in ispec["ISA"] else 'ilp32 '
+        self.compile_cmd = self.compile_cmd+' -mabi='+('lp64 ' if 64 in ispec['supported_xlen'] else ilp32)
+        # if "I" in ispec["ISA"]:
```
Timmmm, line 89: Lots of commented out code around here.

ci-tests/riscof/run-tests.sh
```diff
@@ -0,0 +1,23 @@
...
+    echo "Usage: $0 <argument>"
+    exit 1
+fi
+
+riscof -v debug run --config=config.ini \
+           --suite=riscv-arch-test/riscv-test-suite/ \
+           --env=riscv-arch-test/riscv-test-suite/env \
+           --no-browser --work-dir "$1"
+
+if grep -rniq "$1"/report.html -e '>0failed<'
```
Timmmm, line 16: This sort of thing is very fragile in my experience. It's going to say it succeeded if it can't find "failed" which could be for any number of reasons.

Surely riscof has a more robust output method?

build_simulators.sh
```diff
@@ -10,8 +10,8 @@ function test_build () {
     fi
 }
 
-test_build make ARCH=RV32 ocaml_emulator/riscv_ocaml_sim_RV32
-test_build make ARCH=RV64 ocaml_emulator/riscv_ocaml_sim_RV64
+test_build make -j"$(nproc 2> /dev/null || sysctl -n hw.ncpu)" ARCH=RV32 ocaml_emulator/riscv_ocaml_sim_RV32
```
Timmmm, line 13: This needs a comment at least. Why not just `nproc`?

Timmmm (commented):
Quite a proliferation of bash scripts here. I really think we should avoid that - they tend to be extremely fragile and difficult to maintain. IMO we should use Python.

jordancarlin:
@billmcspadden-riscv Seems like this should be closed since we are going to rely on the CI already in the riscv-arch-test repo per last week's meeting

billmcspadden-riscv:
The goal of crosschecking Sail and Spike using the ACTs,  is being done in the 
[ ](https://github.com/riscv-non-isa/riscv-arch-test)
repository.  It will not be supported in the RISC-V Sail repo as it is redundent.

PR #483, opened 2024-05-20
csim: Fix C23 compatability warning

Alasdair:
The plat_get_16_random_bits was missing its unit argument, which produces the following warning when using clang:

```
generated_definitions/c/riscv_model_RV64.c:28041:34: warning: passing arguments to 'plat_get_16_random_bits' without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
  zseed = plat_get_16_random_bits(UNIT);
```

This commit adds the appropriate argument to the function in the C simulator.

### DIFF
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -108,7 +108,7 @@ mach_bits plat_rom_size(unit u)
 }
 
 // Provides entropy for the scalar cryptography extension.
-mach_bits plat_get_16_random_bits()
+mach_bits plat_get_16_random_bits(unit u)
 {
   return rv_16_random_bits();
 }
--- a/c_emulator/riscv_platform.h
+++ b/c_emulator/riscv_platform.h
@@ -26,7 +26,7 @@ mach_bits plat_rom_base(unit);
 mach_bits plat_rom_size(unit);
 
 // Provides entropy for the scalar cryptography extension.
-mach_bits plat_get_16_random_bits();
+mach_bits plat_get_16_random_bits(unit);
 
 mach_bits plat_clint_base(unit);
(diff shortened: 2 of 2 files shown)


c_emulator/riscv_platform.c
```diff
@@ -108,7 +108,7 @@ mach_bits plat_rom_size(unit u)
 }
 
 // Provides entropy for the scalar cryptography extension.
-mach_bits plat_get_16_random_bits()
+mach_bits plat_get_16_random_bits(unit u)
```
Timmmm, line 111: Will this give warnings about unused variables? I think you can just omit the ` u`.

Alasdair: I think you would need `__attribute__((unused)) unit u`, but I don't think we have that warning enabled for that file.

Timmmm: You can just remove the `u` so it's `mach_bits plat_get_16_random_bits(unit)`, or add `(void)u;` in the body, but fair point about not having those warnings enabled.

Timmmm (approved):

c_emulator/riscv_platform.c
Alasdair: That would create another warning, except this time in the other direction because omitting the parameter name is a C2x extension! 😄 
```
c_emulator/riscv_platform.c:111:39: warning: omitting the parameter name in a function definition is a C2x extension [-Wc2x-extensions]
mach_bits plat_get_16_random_bits(unit)
```

Timmmm: Ha TIL - I pretty much always use C++.

billmcspadden-riscv (approved):
LGTM

PR #486, opened 2024-05-23
Sdtrig support

Mudassir10X:
This PR is for the support of sdtrig extension as proof of concept for triggers support in SAIL-RISCV. It is still incomplete and work is being actively done. reviewers and contributors are welcomed. 

This PR is related to this [issue](https://github.com/orgs/riscv-admin/projects/2/views/19?pane=issue&itemId=23160706) in RISCV dev partners.


martinberger:
Thanks for starting to think about debug extensions. Could you say what the plans are for developing `Sdtrig`? The current draft is *very* far from complete.

model/riscv_insts_zicsr.sail
```diff
@@ -72,7 +72,12 @@ function readCSR csr : csreg -> xlenbits = {
     (0xC22, _) => vlenb,
 
     /* trigger/debug */
-    (0x7a0,  _) => ~(tselect),  /* this indicates we don't have any trigger support */
+    (0x7a0,  _) => tselect.bits,
```
martinberger, line 75: This is missing legalisation functions.

model/riscv_sys_regs.sail
```diff
@@ -721,7 +726,266 @@ function legalize_satp32(a : Architecture, o : bits(32), v : bits(32)) -> bits(3
...
+
+function isValidTselect(tselect_val : xlenbits) -> bool = {
+  let N_triggers = 3;// Number of triggers supported
+  (unsigned(tselect_val) >= 0) & (unsigned(tselect_val) < N_triggers);
+}  
+
+function legalize_tselect(o : TSelect, v : xlenbits) -> TSelect = {
+  let v = Mk_TSelect(v);
+  // Update Index field only if Trigger Module is enabled and Written value is legal
+  let o = [o with Index  = if (sys_enable_sdtrig() & isValidTselect(v[Index])) then v[Index] else o[Index]];
```
martinberger, line 985: I don't `legalize_tselect`  covers all the required legalisation.

Mudassir10X:
Yes I am currently working on it thats why its a draft PR. I am trying to complete this ASAP. Will keep you posted about progress.

model/riscv_insts_zicsr.sail
Mudassir10X: It is in the read_CSR function. does this even need legalization?

martinberger: In the RISCV model, we do write legalisation today, but I don't think this is sufficient for the Sdtrig spec.

Mudassir10X: Actually, I am new to Sail and sail-riscv and still learning about many things. Any help is appreciated. Also, this is not complete as I am first completing the registers and will be working on the legalization after they are perfectly integrated and the repo is able to compile. Please, give your suggestions in this regard if you find some bug or something helpful.

martinberger: Sdtrig is a rather complex specification, for example because the legalisation of some registers depends on the values of some other registers. Sdext is much easier

Mudassir10X: Yes I understand, but this is assigned to me as everyone else who was working on it left the team due to other priorities and this is a priority according to [this.](https://github.com/orgs/riscv-admin/projects/2/views/19?pane=issue&itemId=23160706)

Mudassir10X: > In the RISCV model, we do write legalisation today, but I don't think this is sufficient for the Sdtrig spec. But maybe if you only implement the cases where `action=0` and `dmode=0` this may be enough

According to this [comment](https://github.com/riscv-admin/dev-partners/issues/8#issuecomment-2099470840) in the conversation, I think that this should be enough. What do you think?

martinberger: If `action=0` and `dmode=0` then you are not doing any triggering, and write legalisation is probably enough. However none of the hard cases in the verification of the debug extensions will become visible in this way.

Mudassir10X:
Hi @billmcspadden-riscv I didn't change anything involving the P extension (SIMD) then why is it failing on the 
`rv32mi-p-breakpoint.elf: fail`
I dont understand, can you please check this?

model/riscv_sys_regs.sail
Mudassir10X: Since debug spec is not implemented in SAIL and it's still beyond scope, I don't think any other legalization is required.

Mudassir10X:
Hi @billmcspadden-riscv, this PR is failing the CI for breakpoint test in riscv test. The way I see it, when the sdtrig registers were absent, the access would just be ignored as there was no register at tdata1 and go on. but as now the register is present and is writable in debug mode only **(which is still not present in SAIL-RISCV)**, so it gives illegal instruction exception.

jrtc27:
> Hi @billmcspadden-riscv, this PR is failing the CI for breakpoint test in riscv test. The way I see it, when the sdtrig registers were absent, the access would just be ignored as there was no register at tdata1 and go on. but as now the register is present and is writable in debug mode only **(which is still not present in SAIL-RISCV)**, so it gives illegal instruction exception.

Either the test is wrong or your implementation is wrong. Which is it?

martinberger:
> Either the test is wrong or your implementation is wrong. Which is it?

At a guess, the test is now wrong, because the expected outcome is that the expected outcome is an access failure because the test assumes the register is not there. That has changed. So either remove those 4 tests, or change the expected outcome.

allenjbaum:
Is this the problem with the new definition of what happens when you access
an unimplemented CSR (that it basically becomes undefined).]?
From an architectural test point of view, that means that if you can't
discover whether a CSR exists (either programmatically e.g.
 via something like unified discovery, or passing arguments into the test -
which will eventually be the riscv-config)
then you have to skip testing the feature. Changing the test won't work.
If Sail assumes that this CSR is not present, and is now optionally
present, and we have to way to test whether it is present,
then it is untestable.

UmerShahidengr:
There can be 2 ways to solve this problem.
1. If the test is just checking if some CSR is defined or not, and the passing criteria is "access failure", then test is needed to be modified to change the passing criteria. The test should be considered "pass" if CSR is not defined and giving "access failure" or if CSR is defined then legal read/write should be considered as "pass".
2. Instead of checking if a certain CSR is defined or not, a test should check the valid functionality of that CSR and a flag should be added which chooses the test to be run based on the architecture state (or riscv-config string). Test will be selected only if a certain CSR is defined and accessible but if the CSR is undefined then that test should not be selected.

PS: From ACT point of view, option 2 is viable because if a CSR is not defined then you should not test any feature related to that CSR. But this case in more interesting one. In this case, a CSR (tdata1) has been defined but it is not accessible as Debug mode is not available in Sail, so now there is a CSR whose definition and mapping is available in Sail but you cant read/write it in a test because there is no legal way to access it without Debug mode. In this case, option 1 can be chosen as a short term solution.

allenjbaum:
I want to echo what Umer said: there is a difference between an
unimplemented CSR and an inaccessible CSR.
The spec used to say that accessing unimplemented CSR  would cause an
illegal instruction trap - but that has been loosened to say that the
result is Reserved.
That has two implications:
 - tests must be passed a variable that indicates whether a CSR is
implemented or not
   (as a command line option, extracted from the riscv-config YAML file)
 - tests that access a CSR must be conditioned on that variable as part of
the TEST_CASE macro.

It is still just fine to test accessbility (and we should) under conditions
that explore both allowed and disallowed access
( which could be a function of priv mode, or by type of access (read or
write), or a CSR enable/disable bit)

UmerShahidengr:
@Mudassir10X can you please close this one, since new PR has been added.

billmcspadden-riscv:
This PR has been replaced by PR #573

PR #488, opened 2024-05-23
Use multiple threads for LTO

### LINKED ISSUE #487: Sail-riscv uses only 1 cpu thread at a time thus slowing down the compilaton.
The sail-riscv model uses gcc to compile the c_emulator riscv_sim_RV32/64. there is a flag in C_FLAGS in makefile:L188 as -flto due to which a warning appears during building sail-riscv, that is:

`lto-wrapper: warning: using serial compilation of 12 LTRANS jobs`

This is because only one CPU thread is being used at a time and this slows down the building of the model. This can be solved by using 

`-flto=auto`

Mudassir10X:
This PR is for fixing the issue #487 so that all available threads work at the same time and the build is finished a lot faster.

Screenshot from 2024-05-23 15-42-30

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,7 +185,7 @@ ifneq (,$(COVERAGE))
 C_FLAGS += --coverage -O1
 SAIL_FLAGS += -Oconstant_fold
 else
-C_FLAGS += -O3 -flto
+C_FLAGS += -O3 -flto=auto
 endif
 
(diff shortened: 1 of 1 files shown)


Timmmm:
LGTM but can you give it a proper title, like `Use multiple threads for LTO`.

Mudassir10X:
Yeah, sure will change the title.

PR #494, opened 2024-06-11
Implement callbacks for state-changing events

kseniadobrovolskaya:
Add a callback API for register, PC and CSR updates, memory writes and traps. This is used for logging and RVFI instead of embedding that code in the model itself.

### DIFF
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -1,0 +1,136 @@
+#include "riscv_callbacks.h"
+#include "riscv_config.h"
+#include "riscv_platform_impl.h"
+#include <stdlib.h>
+#include <vector>
+#include <inttypes.h>
+
+bool config_enable_rvfi = true; // Only used if RVFI_DII is defined
+
+void print_lbits_hex(lbits val, int length = 0)
+{
+  if (length == 0) {
+    length = val.len;
+  }
+  std::vector<uint8_t> data(length);
+  mpz_export(data.data(), nullptr, -1, 1, 0, 0, *val.bits);
+  for (int i = length - 1; i >= 0; --i) {
+    fprintf(trace_log, "%02" PRIX8, data[i]);
+  }
+  fprintf(trace_log, "\n");
+}
+
+// Implementations of default callbacks for trace printing and RVFI.
+// The model assumes that these functions do not change the state of the model.
+unit mem_write_callback(uint64_t paddr, uint64_t width, lbits value)
+{
+  if (config_print_mem_access) {
+    fprintf(trace_log, "mem[0x%016" PRIX64 "] <- 0x", paddr);
+    print_lbits_hex(value, width);
+  }
+#ifdef RVFI_DII
+  if (config_enable_rvfi) {
+    zrvfi_write(paddr, width, value);
+  }
+#endif
+  return UNIT;
+}
+
+unit mem_read_callback(const char *type, uint64_t paddr, uint64_t width,
+                       lbits value)
+{
+  if (config_print_mem_access) {
+    fprintf(trace_log, "mem[%s,0x%016" PRIX64 "] -> 0x", type, paddr);
+    print_lbits_hex(value, width);
+  }
+#ifdef RVFI_DII
+  if (config_enable_rvfi) {
+    sail_int len;
+    CREATE(sail_int)(&len);
+    CONVERT_OF(sail_int, mach_int)(&len, width);
+    zrvfi_read(paddr, len, value);
+    KILL(sail_int)(&len);
+  }
+#endif
+  return UNIT;
+}
+
+unit mem_exception_callback(uint64_t paddr, uint64_t num_of_exception)
+{
+  (void)num_of_exception;
+#ifdef RVFI_DII
+  if (config_enable_rvfi) {
+    zrvfi_mem_exception(paddr);
+  }
+#else
+  (void)paddr;
+#endif
+  return UNIT;
+}
+
+unit xreg_write_callback(unsigned reg, uint64_t value)
+{
+  if (config_print_reg) {
+    fprintf(trace_log, "x%d <- 0x%016" PRIX64 "\n", reg, value);
+  }
+#ifdef RVFI_DII
+  if (config_enable_rvfi) {
+    zrvfi_wX(reg, value);
+  }
+#endif
+  return UNIT;
+}
+
+unit freg_write_callback(unsigned reg, uint64_t value)
+{
+  // TODO: will only print bits; should we print in floating point format?
+  if (config_print_reg) {
+    fprintf(trace_log, "f%d <- 0x%016" PRIX64 "\n", reg, value);
+  }
+  return UNIT;
+}
+
+unit csr_full_write_callback(const_sail_string csr_name, unsigned reg,
+                             uint64_t value)
+{
+  if (config_print_reg) {
+    fprintf(trace_log, "CSR %s (0x%03X) <- 0x%016" PRIX64 "\n", csr_name, reg,
+            value);
+  }
+  return UNIT;
+}
+
+unit csr_full_read_callback(const_sail_string csr_name, unsigned reg,
+                            uint64_t value)
+{
+  if (config_print_reg) {
+    fprintf(trace_log, "CSR %s (0x%03X) -> 0x%016" PRIX64 "\n", csr_name, reg,
+            value);
+  }
+  return UNIT;
+}
+
+unit vreg_write_callback(unsigned reg, lbits value)
+{
+  if (config_print_reg) {
+    fprintf(trace_log, "v%d <- ", reg);
+    print_lbits_hex(value);
+  }
+  return UNIT;
+}
+
+unit pc_write_callback(uint64_t value)
+{
+  (void)value;
+  return UNIT;
+}
+
+unit trap_callback(unit)
+{
+#ifdef RVFI_DII
+  if (config_enable_rvfi) {
+    zrvfi_trap();
+  }
+#endif
+  return UNIT;
+}
--- a/c_emulator/riscv_callbacks.h
+++ b/c_emulator/riscv_callbacks.h
@@ -1,0 +1,36 @@
+#pragma once
+#include "sail.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+unit mem_write_callback(uint64_t paddr, uint64_t width, lbits value);
+unit mem_read_callback(const char *type, uint64_t paddr, uint64_t width,
+                       lbits value);
+unit mem_exception_callback(uint64_t paddr, uint64_t num_of_exception);
+unit xreg_write_callback(unsigned reg, uint64_t value);
+unit freg_write_callback(unsigned reg, uint64_t value);
+// `full` indicates that the name and index of the CSR are provided.
+// 64 bit CSRs use a long_csr_write_callback Sail function that automatically
+// makes two csr_full_write_callback calls on RV32.
+unit csr_full_write_callback(const_sail_string csr_name, unsigned reg,
+                             uint64_t value);
+unit csr_full_read_callback(const_sail_string csr_name, unsigned reg,
+                            uint64_t value);
+unit vreg_write_callback(unsigned reg, lbits value);
+unit pc_write_callback(uint64_t value);
+unit trap_callback(unit);
+
+#ifdef RVFI_DII
+// TODO: Move these implementations to C.
+unit zrvfi_write(uint64_t paddr, int64_t width, lbits value);
+unit zrvfi_read(uint64_t paddr, sail_int width, lbits value);
+unit zrvfi_mem_exception(uint64_t paddr);
+unit zrvfi_wX(int64_t reg, uint64_t value);
+unit zrvfi_trap();
+#endif
+
+#ifdef __cplusplus
+}
+#endif
--- a/c_emulator/riscv_prelude.cpp
+++ b/c_emulator/riscv_prelude.cpp
@@ -22,20 +22,6 @@ unit print_step(unit)
   return UNIT;
 }
 
-unit print_reg(sail_string s)
-{
-  if (config_print_reg)
-    fprintf(trace_log, "%s\n", s);
-  return UNIT;
-}
-
-unit print_mem_access(sail_string s)
-{
-  if (config_print_mem_access)
-    fprintf(trace_log, "%s\n", s);
-  return UNIT;
-}
-
 unit print_platform(sail_string s)
 {
   if (config_print_platform)
@@ -45,20 +31,10 @@ unit print_platform(sail_string s)
 
 bool get_config_print_instr(unit)
 {
-  return (config_print_instr) ? true : false;
-}
-
-bool get_config_print_reg(unit)
-{
-  return (config_print_reg) ? true : false;
-}
-
-bool get_config_print_mem(unit)
-{
-  return (config_print_mem_access) ? true : false;
+  return config_print_instr;
 }
 
 bool get_config_print_platform(unit)
 {
-  return (config_print_platform) ? true : false;
+  return config_print_platform;
 }
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -155,6 +155,7 @@ foreach (xlen IN ITEMS 32 64)
 
             set(sail_regs_srcs
                 "riscv_csr_begin.sail"
+                "riscv_callbacks.sail"
                 "riscv_reg_type.sail"
                 "riscv_freg_type.sail"
                 "riscv_regs.sail"
@@ -263,6 +264,7 @@ foreach (xlen IN ITEMS 32 64)
                         # Extra #include's.
                         --c-include riscv_prelude.h
                         --c-include riscv_platform.h
+                        --c-include riscv_callbacks.h
                         # Don't dead-code eliminate these functions. These should match the
                         # ones used from riscv_sail.h
                         --c-preserve init_model
@@ -283,6 +285,11 @@ foreach (xlen IN ITEMS 32 64)
                         --c-preserve rvfi_halt_exec_packet
                         --c-preserve print_instr_packet
                         --c-preserve print_rvfi_exec
+                        --c-preserve rvfi_write
+                        --c-preserve rvfi_read
+                        --c-preserve rvfi_mem_exception
+                        --c-preserve rvfi_wX
+                        --c-preserve rvfi_trap
                         # Input files.
                         ${sail_srcs}
                 )
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -55,6 +55,7 @@ function nan_unbox(m, x) = if 'n == 'm then x else (
 newtype fregidx = Fregidx : bits(5)
 newtype fregno = Fregno : range(0, 31)
 function fregidx_to_fregno (Fregidx(b) : fregidx) -> fregno = Fregno(unsigned(b))
+function fregno_to_fregidx (Fregno(b) : fregno) -> fregidx = Fregidx(to_bits(5, b))
 function fregidx_offset(Fregidx(r) : fregidx, o : bits(5)) -> fregidx = Fregidx(r + o)
 function fregidx_bits(Fregidx(r) : fregidx) -> bits(5) = r
 
@@ -68,6 +69,8 @@ function fregidx_to_regidx (Fregidx(b) : fregidx) -> regidx =
   Regidx(trunc(b))
 
 mapping encdec_freg : fregidx <-> bits(5) = { Fregidx(r) <-> r }
+
+val freg_write_callback = pure {c: "freg_write_callback"} : (fregidx, flenbits) -> unit
 
 register f0  : fregtype
 register f1  : fregtype
@@ -106,6 +109,7 @@ function dirty_fd_context() -> unit = {
   assert(hartSupports(Ext_F));
   mstatus[FS] = extStatus_to_bits(Dirty);
   mstatus[SD] = 0b1;
+  long_csr_write_callback("mstatus", "mstatush", mstatus.bits);
 }
 
 function dirty_fd_context_if_present() -> unit = {
@@ -193,12 +197,8 @@ function wF (Fregno(r) : fregno, in_v : flenbits) -> unit = {
     _  => assert(false, "invalid floating point register number")
   };
 
+  freg_write_callback(fregno_to_fregidx(Fregno(r)), in_v);
   dirty_fd_context();
-
-  if   get_config_print_reg()
-  then
-      /* TODO: will only print bits; should we print in floating point format? */
-      print_reg("f" ^ dec_str(r) ^ " <- " ^ FRegStr(v));
 }
 
 function rF_bits(i: fregidx) -> flenbits = rF(fregidx_to_fregno(i))
--- a/model/riscv_insts_vext_mem.sail
+++ b/model/riscv_insts_vext_mem.sail
@@ -167,7 +167,7 @@ function process_vlsegff (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem)
               if i == 0 then return Ext_DataAddr_Check_Failure(e)
               else {
                 vl = to_bits(xlen, i);
-                print_reg("CSR vl <- " ^ BitStr(vl));
+                csr_write_callback("vl", vl);
                 trimmed = true
               }
             },
@@ -176,7 +176,7 @@ function process_vlsegff (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem)
                 if i == 0 then return Memory_Exception(vaddr, E_Load_Addr_Align())
                 else {
                   vl = to_bits(xlen, i);
-                  print_reg("CSR vl <- " ^ BitStr(vl));
+                  csr_write_callback("vl", vl);
                   trimmed = true
                 }
               } else match translateAddr(vaddr, Read(Data)) {
@@ -184,7 +184,7 @@ function process_vlsegff (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem)
                   if i == 0 then return Memory_Exception(vaddr, e)
                   else {
                     vl = to_bits(xlen, i);
-                    print_reg("CSR vl <- " ^ BitStr(vl));
+                    csr_write_callback("vl", vl);
                     trimmed = true
                   }
                 },
@@ -195,7 +195,7 @@ function process_vlsegff (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem)
                       if i == 0 then return Memory_Exception(vaddr, e)
                       else {
                         vl = to_bits(xlen, i);
-                        print_reg("CSR vl <- " ^ BitStr(vl));
+                        csr_write_callback("vl", vl);
                         trimmed = true
                       }
                     }
--- a/model/riscv_insts_vext_vset.sail
+++ b/model/riscv_insts_vext_vset.sail
@@ -48,8 +48,8 @@ function handle_illegal_vtype() = {
    */
   vtype.bits = 0b1 @ zeros(xlen - 1); /* set vtype.vill */
   vl = zeros();
-  print_reg("CSR vtype <- " ^ BitStr(vtype.bits));
-  print_reg("CSR vl <- " ^ BitStr(vl))
+  csr_write_callback("vtype", vtype.bits);
+  csr_write_callback("vl", vl);
 }
 
 function vl_use_ceil() -> bool = config extensions.V.vl_use_ceil
@@ -111,9 +111,9 @@ function clause execute VSETVLI(ma, ta, sew, lmul, rs1, rd) = {
   /* reset vstart to 0 */
   set_vstart(zeros());
 
-  print_reg("CSR vtype <- " ^ BitStr(vtype.bits));
-  print_reg("CSR vl <- " ^ BitStr(vl));
-  print_reg("CSR vstart <- " ^ BitStr(vstart));
+  csr_write_callback("vtype", vtype.bits);
+  csr_write_callback("vl", vl);
+  csr_write_callback("vstart", zero_extend(vstart));
 
   RETIRE_SUCCESS
 }
@@ -163,9 +163,9 @@ function clause execute VSETVL(rs2, rs1, rd) = {
   /* reset vstart to 0 */
   set_vstart(zeros());
 
-  print_reg("CSR vtype <- " ^ BitStr(vtype.bits));
-  print_reg("CSR vl <- " ^ BitStr(vl));
-  print_reg("CSR vstart <- " ^ BitStr(vstart));
+  csr_write_callback("vtype", vtype.bits);
+  csr_write_callback("vl", vl);
+  csr_write_callback("vstart", zero_extend(vstart));
 
   RETIRE_SUCCESS
 }
@@ -200,9 +200,9 @@ function clause execute VSETIVLI(ma, ta, sew, lmul, uimm, rd) = {
   /* reset vstart to 0 */
   set_vstart(zeros());
 
-  print_reg("CSR vtype <- " ^ BitStr(vtype.bits));
-  print_reg("CSR vl <- " ^ BitStr(vl));
-  print_reg("CSR vstart <- " ^ BitStr(vstart));
+  csr_write_callback("vtype", vtype.bits);
+  csr_write_callback("vl", vl);
+  csr_write_callback("vstart", zero_extend(vstart));
 
   RETIRE_SUCCESS
 }
--- a/model/riscv_mem.sail
+++ b/model/riscv_mem.sail
@@ -73,9 +73,7 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext
     (InstructionFetch(), None()) => Err(E_Fetch_Access_Fault()),
     (Read(Data), None()) => Err(E_Load_Access_Fault()),
     (_,          None()) => Err(E_SAMO_Access_Fault()),
-    (_,      Some(v, m)) => { if   get_config_print_mem()
-                              then print_mem("mem[" ^ to_str(t) ^ "," ^ hex_bits_str(bits_of(paddr)) ^ "] -> " ^ BitStr(v));
-                              Ok(v, m) }
+    (_,      Some(v, m)) => Ok(v, m),
   }
 }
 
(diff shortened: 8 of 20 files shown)


model/riscv_insts_vext_vset.sail
```diff
@@ -48,6 +48,11 @@ function handle_illegal_vtype() = {
    */
   vtype.bits = 0b1 @ zeros(sizeof(xlen) - 1); /* set vtype.vill */
   vl = zeros();
+
+  if rv_enable_callbacks then {
+    csr_update_callback("vtype", vtype.bits);
+    csr_update_callback("vl", vl);
+  };
   print_reg("CSR vtype <- " ^ BitStr(vtype.bits));
```
Timmmm, line 56: I guess you can remove these `print_reg()` calls too since the `csr_update_callback()` can do that?

Timmmm (commented):
This looks along the lines I was thinking, though I think we will still need to have RVFI support in the C emulator, so you still need the flags. It will just need to be implemented via the callbacks.

I probably wouldn't use `__attribute__((weak))`. If anyone is modifying the callbacks they'll have their own fork of the emulator anyway and can just delete the default ones. That's what we (Codasip) will do.

I'll get you some code to read `lbits`.

Timmmm:
It's something like this in C++ (adapted from our code, so this exact code isn't tested).

```
std::vector<uint8_t> readLbits(lbits val) {
    std::vector<uint8_t> data(val.len);

    // We should really double check the length of `val.bits` to ensure it
    // isn't actually greated than `val.len`, otherwise we are at risk of
    // a buffer overflow.
    // I think you can maybe use (mpz_sizeinbase(val.bits, 2) + 7) / 8.

    // Export the lbits. Note this doesn't pad with zeros but that's ok
    // because data is prefilled with zeros.
    mpz_export(
        // Destination buffer.
        data.data(),
        // Where to write the output size (we ignore this).
        nullptr,
        // Endianness of words (-1 = little).
        -1,
        // Size of words in bytes.
        1,
        // Endianness *within* words (0 = native). Doesn't matter in this case.
        0,
        // Number of "nail" bits (upper bits to clear in each word).
        0,
        // The GMP value. Note the sign is ignored but it should always be positive.
        val.bits);

    return data;
}
```

In crappy C you will have to deal with `malloc` manually rather than using `vector`.

model/riscv_insts_vext_mem.sail
```diff
@@ -154,6 +154,7 @@ function process_vlsegff (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem)
               if i == 0 then { ext_handle_data_check_error(e); return RETIRE_FAIL }
               else {
                 vl = to_bits(sizeof(xlen), i);
+                if rv_enable_callbacks then csr_update_callback("vl", vl);
```
arichardson, line 157: Having to add these manually is rather error prone and makes the code less readable. Can we do some fancy sail overloading tricks to automatically call these functions for CSR writes (e.g. something like `CSR(vl) = ...` ?

Timmmm: Various CSRs have a `set_foo()` function. We could put it in there. Though that doesn't significantly reduce the risk because people can always forget to use the setter.

I ran into a similar issue recently - trying to memoize/cache a value. Difficult to guarantee that all the code will use setter that updates the cache without some support for `private`.

IMO there are few enough of these that we can probably ignore this issue for the moment, though I agree it's not ideal.

Timmmm:
> mpz_sizeinbase(val.bits, 2) + 7) / 8

Actually reading this back, I wonder if `mpz_sizeinbase(val.bits, 256)` would work (and not be really slow).

model/riscv_insts_vext_mem.sail
bacam: Perhaps it would be better to get Sail to generate the callbacks automatically?

Timmmm: Yeah we discussed this before. Might be nice, but I'm not 100% sure it would do exactly what you want, e.g. you probably want distinct callbacks for writes to `sstatus` and `mstatus`, but under the hood they write to the same `register`. Also there are very few places this callback is needed so the manual burden is not that high, and I suspect Alasdair has enough on his plate already!

model/riscv_insts_vext_vset.sail
kseniadobrovolskaya: Yes, deleted this function.

kseniadobrovolskaya:
I would like to point out that the callback functions are annotated as `pure`. But in reality we have no way to guarantee that users will not actually change the state of the model in these functions. Should we really keep this attribute or should we remove it so that we don't fool anyone?

model/riscv_insts_vext_mem.sail
kseniadobrovolskaya: I added the `ext_notification_write_CSR` (and `ext_notification_read_CSR`) function. It helps to simplify writing to a CSR with a call to a callback.

Timmmm:
> I would like to point out that the callback functions are annotated as `pure`. But in reality we have no way to guarantee that users will not actually change the state of the model in these functions. Should we really keep this attribute or should we remove it so that we don't fool anyone?

I would say if someone edits the code and changes the callback to be impure then it's up to them to also change the Sail code to mark them as impure.

arichardson (commented):
Could you keep the RVFI-DII code around? Alternatively, a good demonstrator of the callbacks would be to implement the RVFI-DII socket interface using those callbacks?

Timmmm:
Yeah to be clear I don't think these callbacks should be user defined. We should have hard-coded callbacks that support RVFI, logging, etc.

kseniadobrovolskaya:
@arichardson, @Timmmm, what do you think of the current version?

model/riscv_sys_control.sail
```diff
@@ -340,8 +340,7 @@ function trap_handler(del_priv : Privilege, intr : bool, c : exc_code, pc : xlen
```
Timmmm, line 340: There's a load of implicit CSR writes here that need to be tracked. We do it like this:

```
function track_trap(p : Privilege, intr : bool, c : exc_code) -> unit = {
  track_writeCSR(csr_name_map("mstatus"));
  match (p) {
    Machine => {
      track_writeCSR(csr_name_map("mcause"));
      track_writeCSR(csr_name_map("mtval"));
      track_writeCSR(csr_name_map("mepc"));
    },
    Supervisor => {
      track_writeCSR(csr_name_map("scause"));
      track_writeCSR(csr_name_map("stval"));
      track_writeCSR(csr_name_map("sepc"));
    },
    User => {
      track_writeCSR(csr_name_map("ucause"));
      track_writeCSR(csr_name_map("utval"));
      track_writeCSR(csr_name_map("uepc"));
    },
  };
}
```

model/riscv_mem.sail
```diff
@@ -243,7 +206,11 @@ function mem_write_value_priv_meta (paddr, width, value, typ, priv, meta, aq, rl
   then MemException(E_SAMO_Addr_Align())
   else {
     let result = checked_mem_write(paddr, width, value, typ, priv, meta, aq, rl, con);
-    rvfi_write(paddr, width, value, meta, result);
+    if rv_enable_callbacks then
+      match result {
+        MemValue(_) => mem_update_callback(paddr, width, value, /* is_exception */ false),
+        MemException(_) => mem_update_callback(paddr, width, value, /* is_exception */ true)
```
Timmmm, line 212: I don't know if it's necessarily better, but we only call out `mem_read_callback()` function on success `MemValue()`, and then we have another callback for traps in `trap_handler()`, where the `print_reg()`s are.

Maybe keep this (because you can easily ignore the exception case) but also add the trap callback.

You might want to add the exception code to the callback to make logging a bit easier:

```
      match result {
        MemValue(_) => mem_write_callback(paddr, width, value),
        MemException(e) => mem_exception_callback(paddr, width, num_of_ExceptionType(e)),
```

Timmmm (commented):
Yeah this is pretty much exactly what I was thinking. I put some minor comments but I it looks like the right direction to me.

model/riscv_fdext_regs.sail
```diff
@@ -157,6 +157,7 @@ function rF r = {
 val wF : forall 'n, 0 <= 'n < 32. (regno('n), flenbits) -> unit
 function wF (r, in_v) = {
   assert(sys_enable_fdext());
+  if rv_enable_callbacks then freg_update_callback(regno_to_regidx(r), in_v);
```
jrtc27, line 160: IMO a better design is to push the enable check down from the caller rather than replicate it everywhere

arichardson (commented):
Overall I think this approach looks good, but I'd push all the rv_enable_callbacks into the callee to simplify the code. Since the variable is not a compile-time constant it shouldn't matter whether it's checked in the C code or sail code. It might be easier to keep this as an internal flag to the C/Ocaml emulators.

c_emulator/riscv_default_callbacks.c
```diff
@@ -0,0 +1,16 @@
...
+                        bool is_exception)
+{
+}
+int mem_read_callback(uint64_t addr, uint64_t width, lbits value,
+                      bool is_exception)
+{
+}
+int xreg_update_callback(unsigned reg, uint64_t value) { }
+int freg_update_callback(unsigned reg, uint64_t value) { }
+int csr_update_callback(const char *reg_name, uint64_t value) { }
```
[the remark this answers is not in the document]
kseniadobrovolskaya: Ok, now in this function argument `unsigned reg`, and in the sail -- `csreg`.

model/riscv_fdext_regs.sail
```diff
@@ -193,11 +193,8 @@ function wF (r : regno, in_v : flenbits) -> unit = {
   };
 
   dirty_fd_context();
-
-  if   get_config_print_reg()
-  then
-      /* TODO: will only print bits; should we print in floating point format? */
-      print_reg("f" ^ dec_str(r) ^ " <- " ^ FRegStr(v));
+  
+  freg_write_callback(regno_to_regidx(r), in_v);
```
arichardson, line 197: The default callback needs to print the register otherwise this removes tracing capabilities from the model.

kseniadobrovolskaya: Added an implementation of the callbacks for trace printing to the `riscv_types.sail`, `riscv_csr_begin.sail`, `riscv_vreg_type.sail` and call them from the default callbacks in the `riscv_default_callbacks.c`.

model/riscv_types.sail
```diff
@@ -410,3 +412,41 @@ function report_invalid_width(f , l, w, k) -> 'a = {
...
+val xreg_write_callback_default : (regidx, xlenbits) -> unit
+function xreg_write_callback_default (reg, value) = {
+  print_reg("x" ^ dec_str(regidx_to_regno(reg)) ^ " <- " ^ BitStr(value))
+}
+
+val freg_write_callback_default : (regidx, flenbits) -> unit
+function freg_write_callback_default (reg, value) = {
+  /* todo: will only print bits; should we print in floating point format? */
+  print_reg("f" ^ dec_str(regidx_to_regno(reg)) ^ " <- " ^ BitStr(value))
+}
```
arichardson, line 450: Now that the OCaml simulator is gone, I wonder if we should just move these default callbacks to C? Doing the sail -> C -> sail roundtrip seems a bit unnecessary?

arichardson (commented):
Not reviewed this in detail, but overall the new approach looks sensible to me.

model/riscv_types.sail
kseniadobrovolskaya: I agree, it's done.

kseniadobrovolskaya:
@arichardson, @Timmmm, what do you think of the current version?

Timmmm:
Hi, sorry for not getting around to this. Just wanted to say it's still on my todo list and I will get to it!

Makefile
```diff
@@ -225,7 +225,7 @@ c_preserve_fns=-c_preserve _set_Misa_C
 
 generated_definitions/c/riscv_model_$(ARCH).c: $(SAIL_SRCS) model/main.sail Makefile
 	mkdir -p generated_definitions/c
-	$(SAIL) $(SAIL_FLAGS) $(c_preserve_fns) -O -Oconstant_fold -memo_z3 -c -c_include riscv_prelude.h -c_include riscv_platform.h -c_no_main $(SAIL_SRCS) model/main.sail -o $(basename $@)
+	$(SAIL) $(SAIL_FLAGS) $(c_preserve_fns) -O -Oconstant_fold -memo_z3 -c -c_include riscv_prelude.h -c_include riscv_platform.h -c_no_main $(SAIL_SRCS) -c_include riscv_default_callbacks.c model/main.sail -o $(basename $@)
```
Timmmm, line 228: I don't think we want to `#include` a `.c` file. This should include a header.

c_emulator/riscv_config.h
```diff
@@ -6,3 +6,4 @@ extern bool config_print_step;
 extern bool config_print_reg;
 extern bool config_print_mem_access;
 extern bool config_print_platform;
+extern bool rv_enable_callbacks;
```
Timmmm, line 9: I'd call this `config_enable_rvfi` instead.

c_emulator/riscv_rvfi_callbacks.c
```diff
@@ -0,0 +1,38 @@
+#include "riscv_config.h"
+
+int zrvfi_write(uint64_t addr, int64_t width, lbits value);
```
Timmmm, line 3: In future these should be in C, not Sail. I think it's probably best to do in a future PR but can you add

```
// TODO: Move these implementations to C.
```

c_emulator/riscv_rvfi_callbacks.c
```diff
@@ -0,0 +1,38 @@
...
+
+int mem_write_callback(uint64_t addr, uint64_t width, lbits value)
+{
+  if (rv_enable_callbacks)
+    zrvfi_write(addr, width, value);
+}
+int mem_read_callback(const char *type, uint64_t addr, uint64_t width,
+                      lbits value)
+{
+  if (rv_enable_callbacks) {
```
Timmmm, line 16: We want a single callback that does *both* the printing, and the RVFI stuff (if enabled). So the contents of this file should be merged into `riscv_default_callbacks.c`. Something like this:

```
int mem_write_callback(uint64_t addr, uint64_t width, lbits value)
{
  if (config_print_mem_access) {
    char *lbits_data = get_lbits_data(value);
    printf("mem[0x%.16lX] <- 0x", addr);
    for (int i = width - 1; i >= 0; --i)
      printf("%02hhX", lbits_data[i]);
    printf("\n");
    free(lbits_data);
  }

#ifdef RVFI_DII
  if (config_enable_rvfi) {
    sail_int len;
    CREATE(sail_int)(&len);
    CONVERT_OF(sail_int, mach_int)(&len, width);
    zrvfi_read(addr, len, value);
    KILL(sail_int)(&len);
  }
#endif
}
```

In future we should remove the `#ifdef` so RVFI_DII support is always enabled. It's unnecessary to have it as a compile time option - it just doubles compilation time. (But don't do it in this PR.)

c_emulator/riscv_sail.h
```diff
@@ -50,6 +50,20 @@ extern bool zhtif_done;
 extern mach_bits zhtif_exit_code;
 extern bool have_exception;
 
+/* Callbacks for state-changing events */
```
Timmmm, line 53: As mentioned above - move these into their own header and `-c_include` that.

c_emulator/riscv_default_callbacks.c
```diff
@@ -0,0 +1,91 @@
...
+int xreg_write_callback(unsigned reg, uint64_t value)
+{
+  if (config_print_reg)
+    printf("x%d <- 0x%.16lX\n", reg, value);
+}
+
+int freg_write_callback(unsigned reg, uint64_t value)
+{
+  /* TODO: will only print bits; should we print in floating point format? */
+  if (config_print_reg)
```
Timmmm, line 53: Minor, but can you always use braces - single-statement blocks a bit of a footgun in C.

I thought we had clang-format set up to enforce that but maybe not. I'll double check that.

Timmmm (changes requested):
Finally got around to this again - sorry for the delay!

I think it's pretty close - the main thing is merging `riscv_rvfi_callbacks.c` into `riscv_default_callbacks.c`.

You could also rename it to just `riscv_callbacks.c`.

jordancarlin (dismissed):

Timmmm:
@kseniadobrovolskaya would it be ok if I did a bit of work on this and pushed to this branch?

kseniadobrovolskaya:
> @kseniadobrovolskaya would it be ok if I did a bit of work on this and pushed to this branch?

Yes, of course.

kseniadobrovolskaya:
> @kseniadobrovolskaya would it be ok if I did a bit of work on this and pushed to this branch?

Hi @Timmmm, I just want to make sure that there is no misunderstanding - do you imply that you going to fix the pending comments from you, or should I continue to work on them?

jordancarlin:
@Timmmm @kseniadobrovolskaya Any updates on this?

Timmmm:
> Any updates on this?

> do you imply that you going to fix the pending comments from you, or should I continue to work on them?

Yes, at least that was my plan, but I haven't had any time to work on it yet! I'll rebase it now at least...

Timmmm:
I squashed, rebased and commented out `mstatus` tracking since it doesn't compile currently.

Also there are linker errors because the C file is no longer compiled in (we don't want to `--c-include` it). I'll take a look tomorrow.

jordancarlin:
@Timmmm Some of what I'm working on is going to need this soon. Any idea of when you might return to it? I can also drive it forward and update it instead if you'd rather.

Timmmm:
Yeah I don't have any time right now so go ahead! I'll post a message if I'm going to start working on it.

jordancarlin:
Rebased and various other cleanup and updates to bring this in line with current model standards. RVFI is currently failing (it was failing on the previous version of this as well) and the `mstatus` callback is still temporarily commented out.

jordancarlin:
Rebased and updated to include callbacks for the trap handler. Also fixed the `mstatus` callbacks. RVFI is still broken and there is more cleanup needed, but getting closer to having this ready for another review.

jordancarlin:
Finally working with RVFI again!

c_emulator/riscv_callbacks.cpp
```diff
@@ -0,0 +1,118 @@
...
+// Implementations of default callbacks for trace printing.
+//
+// The model assumes that these functions do not change the state of the model.
+unit mem_write_callback(uint64_t addr, uint64_t width, lbits value)
+{
+  if (config_print_mem_access) {
+    std::vector<uint8_t> lbits_data = read_lbits(value);
+    printf("mem[0x%.16lX] <- 0x", addr);
+    for (int i = width - 1; i >= 0; --i)
+      printf("%02hhX", lbits_data[i]);
```
arichardson, line 22: How about factoring out the printing loop into a helper function, then we can avoid the temporary local variable for the three prints?

```suggestion
    printf("mem[0x%.16lX] <- 0x", addr);
    print_hex(read_lbits(value));
```

c_emulator/riscv_callbacks.cpp
```diff
@@ -0,0 +1,118 @@
...
+  zrvfi_wX(reg, value);
+#endif
+  return UNIT;
+}
+
+unit freg_write_callback(unsigned reg, uint64_t value)
+{
+  // TODO: will only print bits; should we print in floating point format?
+  if (config_print_reg) {
+    printf("f%d <- 0x%.16lX\n", reg, value);
```
arichardson, line 74: Using l as the modifier won't work for all systems, should probbaly use the PRIX64 macro from inttypes.h

c_emulator/riscv_callbacks.cpp
```diff
@@ -0,0 +1,118 @@
...
+#endif
+  return UNIT;
+}
+
+unit mem_read_callback(const char *type, uint64_t addr, uint64_t width,
+                       lbits value)
+{
+  if (config_print_mem_access) {
+    std::vector<uint8_t> lbits_data = read_lbits(value);
+    printf("mem[%s,0x%.16lX] -> 0x", type, addr);
```
arichardson, line 36: Ah I also noticed we probably want to use `fprintf(trace_log,` here and in (almost?) all the other cases.

c_emulator/riscv_prelude.cpp
```diff
@@ -22,20 +22,6 @@ unit print_step(unit)
...
-  return UNIT;
-}
-
-unit print_mem_access(sail_string s)
-{
-  if (config_print_mem_access)
-    fprintf(trace_log, "%s\n", s);
-  return UNIT;
-}
-
```
arichardson, line 38: We should be able to remove get_config_print_mem and get_config_print_reg from this file.

jordancarlin: Done

c_emulator/riscv_callbacks.cpp
```diff
@@ -0,0 +1,134 @@
...
+  }
+  return UNIT;
+}
+
+unit csr_read_callback(unsigned reg, uint64_t value)
+{
+  if (config_print_reg) {
+    sail_string csr_name;
+    CREATE(sail_string)(&csr_name);
+    zcsr_name_map_forwards(&csr_name, reg);
```
arichardson, line 114: Shouldn't this also be guarded by config_print_csr_name and print an integer name instead?

model/CMakeLists.txt
```diff
@@ -281,6 +282,10 @@ foreach (xlen IN ITEMS 32 64)
                         --c-preserve rvfi_halt_exec_packet
                         --c-preserve print_instr_packet
                         --c-preserve print_rvfi_exec
+                        --c-preserve rvfi_write
+                        --c-preserve rvfi_read
+                        --c-preserve rvfi_mem_exception
+                        --c-preserve rvfi_wX
```
arichardson, line 288:
```suggestion
                        --c-preserve rvfi_wX
                        --c-preserve csr_name_map_forwards
```
Isn't this also needed?

model/riscv_mem.sail
```diff
@@ -151,7 +130,10 @@ function mem_read_priv_meta (typ, priv, paddr, width, aq, rl, res, meta) = {
       (false, true,  true)  => throw(Error_not_implemented("lr.rl")),
       (_, _, _)             => checked_mem_read(typ, priv, paddr, width, aq, rl, res, meta)
     };
-  rvfi_read(paddr, width, result);
+  match result {
+    Ok(value, _) => mem_read_callback(to_str(typ), bits_of(paddr), width, value),
+    Err(e) => mem_exception_callback(bits_of(paddr), num_of_ExceptionType(e)),
+  };
```
arichardson, line 136: Maybe best to move this to checked_mem_read or keep it in phys_mem_read since that's where the actual read happens?

model/riscv_mem.sail
```diff
@@ -242,7 +198,10 @@ function mem_write_value_priv_meta (paddr, width, value, typ, priv, meta, aq, rl
   then Err(E_SAMO_Addr_Align())
   else {
     let result = checked_mem_write(paddr, width, value, typ, priv, meta, aq, rl, con);
-    rvfi_write(paddr, width, value, meta, result);
+    match result {
+      Ok(_) => mem_write_callback(bits_of(paddr), width, value),
+      Err(e) => mem_exception_callback(bits_of(paddr), num_of_ExceptionType(e)),
+    };
```
arichardson, line 204: Same comment here

model/riscv_pc_access.sail
```diff
@@ -28,5 +28,6 @@ function set_next_pc(pc) = {
 
 val tick_pc : unit -> unit
 function tick_pc() = {
+  pc_write_callback(PC);
```
arichardson, line 31: Shouldn't this receive the new pc instead of the previous one?

arichardson (approved):
A few minor comments but overall looking good to me now. LGTM with those addressed, assuming the rvfi code still works.

I wonder if we can clean up the csr_name_map("mcause"), etc. calls in a future commit to use some kind of enum? But not for this PR.

c_emulator/riscv_callbacks.cpp
jordancarlin: Good point. Fixed.

model/CMakeLists.txt
jordancarlin: Not sure why it worked without that, but added it just to be safe.

model/riscv_mem.sail
jordancarlin: Then it wouldn't call the `mem_exception_callback` if it's misaligned and aq or res is true.

arichardson: That's a good point - maybe we should have the success callback happen in phys_mem_read() and only keep the error case here? 

This ensures the callback is called immediately after the effect happens. It also makes sure it is always called even if we decide to add more memory accesses in the future (e.g. some DMA model that goes to phys memory directly and bypasses mem_read_priv_meta)?

jordancarlin: Hmm. Doing it in `phys_mem_read` also brings up the issue of how to deal with mmio reads. Having it at the top level `mem_read_priv_meta` means that mmio reads will be printed as well. Moving it to `phys_mem_read ` skips those. mmio reads seem to have their own custom printing right now, so maybe that's ok (though those should probably be changed to callbacks later as well).

arichardson: hmm maybe the best place would be read_ram/write_ram in prelude_mem.sail?

jordancarlin: So `mem_read_callback` in `read_ram` and `mem_exception_callback` in `mem_read_priv_meta`. And comparable for writes.

arichardson: I like that solution, not sure if there are other opinions (e.g. @Timmmm ?)

jordancarlin: Looked in to this, and moving it to `read_ram` means that we lose the simple access types that are part of the print right now. `read_ram` gets the following types instead of just the `Read`/`Write`/`ReadWrite`/`InstructionFetch` present in the higher level functions.
```
  Read_plain,
  Read_ifetch,
  Read_RISCV_acquire,
  Read_RISCV_strong_acquire,
  Read_RISCV_reserved,
  Read_RISCV_reserved_acquire,
  Read_RISCV_reserved_strong_acquire,
```

jordancarlin: Sounds good. Unless anyone else has a better suggestion, let's keep it here for now and cross the bridge of moving it to lower level memory functions if/when we add support for something like DMA.

model/riscv_sys_control.sail
```diff
@@ -211,8 +211,27 @@ val rvfi_trap : unit -> unit
 function rvfi_trap () = ()
 $endif
```
pmundkur, line 212: Would it make sense to add a `trap_callback()` and move the `rvfi_trap` functionality to it?  This would remove the `ifdef RVFI_DII` from this file as well, leaving it only in `riscv_platform.sail`.

pmundkur (changes requested):
A nice improvement, moving most of RVFI handling out of the main Sail model.

c_emulator/riscv_sim.cpp
```diff
@@ -66,6 +68,7 @@ void set_config_print(char *var, bool val)
     config_print_reg = val;
     config_print_platform = val;
     config_print_rvfi = val;
```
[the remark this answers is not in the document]
jordancarlin: That function is enabling/disabling various forms of logging. `config_print_csr_name` just controls whether the csr logs are printed with the csr name or the csr number. It definitely shouldn't go in the all block, but maybe makes sense as one of the else ifs.

model/riscv_fdext_regs.sail
```diff
@@ -106,6 +109,9 @@ function dirty_fd_context() -> unit = {
   assert(hartSupports(Ext_F));
   mstatus[FS] = extStatus_to_bits(Dirty);
   mstatus[SD] = 0b1;
+  csr_write_callback(csr_name_map("mstatus"), mstatus.bits[xlen - 1 .. 0]);
```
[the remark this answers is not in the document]
jordancarlin: The standard CSR instruction call to the callback is using the bitvector, so there will have to be a back and forth for one of them. It seemed to me like if we had to pick one, then passing the int seemed cleaner than the string. Open to switching it though if you think that would be better.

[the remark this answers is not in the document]
pmundkur: Why not pass `csr_name`, `reg` and `value` to both `csr_{read,write}_callback`s?   That would save one mapping.  And perhaps always print both the CSR name and reg, instead of having to choose one using `config_print_csr_name`?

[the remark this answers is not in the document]
jordancarlin: Seems reasonable. I'll make that change. That also resolves your other comment above.

model/riscv_sys_control.sail
jordancarlin: Done

c_emulator/riscv_sim.cpp
```diff
@@ -66,6 +67,7 @@ void set_config_print(char *var, bool val)
...
+    config_enable_rvfi = val;
```
arichardson, line 70: I don't think we should be enabling rvfi with the all flag.

arichardson (approved):

c_emulator/riscv_sim.cpp
jordancarlin: Coincidentally just changed that. It has nothing to do with printing so shouldn't be part of that function at all. It only has an impact when `RVFI_DII` is defined, so I still have it default to true.

model/riscv_insts_vext_mem.sail
```diff
@@ -167,7 +167,7 @@ function process_vlsegff (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem)
               if i == 0 then return RETIRE_FAIL(Ext_DataAddr_Check_Failure(e))
               else {
                 vl = to_bits(xlen, i);
-                print_reg("CSR vl <- " ^ BitStr(vl));
+                csr_write_callback("vl", csr_name_map("vl"), vl);
```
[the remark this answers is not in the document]
jordancarlin: Good suggestion. Done.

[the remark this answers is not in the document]
jordancarlin: Also added a `long_csr_write_callback` function to handle writing to CSRs that are always 64 bits, like mstatus. That way we don't need separate calls for mstatus and mstatush.

c_emulator/riscv_callbacks.cpp
```diff
@@ -0,0 +1,134 @@
...
+  }
+#endif
+  return UNIT;
+}
+
+unit freg_write_callback(unsigned reg, uint64_t value)
+{
+  // TODO: will only print bits; should we print in floating point format?
+  if (config_print_reg) {
+    fprintf(trace_log, "f%d <- 0x%016" PRIX64 "\n", reg, value);
```
pmundkur, line 88: Doesn't have to be this PR, since the current model doesn't do it, but this print should adapt to whether Zfinx is currently enabled.

model/CMakeLists.txt
```diff
@@ -281,6 +283,12 @@ foreach (xlen IN ITEMS 32 64)
...
+                        --c-preserve rvfi_trap
+                        --c-preserve csr_name_map_forwards
```
pmundkur, line 291: I think this `c-preserve` is no longer needed?

c_emulator/riscv_callbacks.cpp
```diff
@@ -0,0 +1,134 @@
...
+#else
+  (void)addr;
+#endif
+  return UNIT;
+}
+
+unit xreg_write_callback(unsigned reg, uint64_t value)
+{
+  if (config_print_reg) {
+    fprintf(trace_log, "x%d <- 0x%016" PRIX64 "\n", reg, value);
```
pmundkur, line 74: Not in this PR, but we should later add an option to print the ABI names of all the registers, which is almost always more useful.

model/riscv_pc_access.sail
```diff
@@ -28,5 +28,6 @@ function set_next_pc(pc) = {
 
 val tick_pc : unit -> unit
 function tick_pc() = {
+  pc_write_callback(nextPC);
```
pmundkur, line 31: Afaict, this is the only callback made before the actual write effect.  Any reason why?

pmundkur (changes requested):
Other than the `c-preserve` and `pc_write_callback`, looks great!

c_emulator/riscv_callbacks.cpp
jordancarlin: `freg_write_callback` is only called when a write to an f register occurs (it is called within `wF`). If Zfinx is enabled, then `wX` is called instead and that will use `xreg_write_callback`.

model/riscv_pc_access.sail
jordancarlin: Reversed

c_emulator/riscv_callbacks.cpp
jordancarlin: Issue #909 opened to track this.

pmundkur (approved):

Timmmm (commented):
LGTM. Just a couple of minor things and I think we should add some comments to either `riscv_callbacks.h` or `riscv_callbacks.sail` to say exactly what all the callbacks mean, because there are some subtleties. E.g. when exactly is `trap_callback()` called? Are addresses physical or virtual? How is `mstatush` etc. handled?

There are probably some other minor things we can improve but I think this is already a huge improvement so lets get it in!

c_emulator/riscv_callbacks.h
```diff
@@ -0,0 +1,33 @@
...
+extern "C" {
+#endif
+
+unit mem_write_callback(uint64_t addr, uint64_t width, lbits value);
+unit mem_read_callback(const char *type, uint64_t addr, uint64_t width,
+                       lbits value);
+unit mem_exception_callback(uint64_t addr, uint64_t num_of_exception);
+unit xreg_write_callback(unsigned reg, uint64_t value);
+unit freg_write_callback(unsigned reg, uint64_t value);
+unit csr_full_write_callback(const_sail_string csr_name, unsigned reg,
```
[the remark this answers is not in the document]
jordancarlin: Yeah. I tried but couldn't come up with anything better. I'll add a comment to clarify.

Timmmm (approved):
LGTM. We can add the comments later. I imagine the exact API may change slightly when we try to use it in anger anyway.

jordancarlin:
Oops. Fixed. I added some comments to `riscv_callbacks.h`.

Timmmm:
Nice work everyone!

PR #496, opened 2024-06-18
Fix TLB bug

Timmmm:
These lines of code zeroed the top 25 bits of an Sv39 address before translating it. That meant an address like 0xFFFFFFFFFFFFFFFF would be stored in the TLB as 0x0000007fffffffff. This caused a bug with `sfence.vma` when the `rs1` (virtual address) argument was not zero. If you did `sfence.vma 0xFFFFFFFFFFFFFFFF, x0` it should clear the TLB entry but it doesn't because it naively checks `0xFFFFFFFFFFFFF000 == 0x0000007ffffff000`.

Currently there is only one TLB entry so it would need to do an `sfence.vma` for the page that was currently executing to be visible, otherwise the next fetch would clear it anyway.

An alternative fix would be to clear the upper 25 bits of `vMatchMask`, but this is simpler.

### DIFF
--- a/model/riscv_vmem.sail
+++ b/model/riscv_vmem.sail
@@ -384,16 +384,13 @@ function translate_TLB_miss(sv_params : SV_Params,
 function translate(sv_params : SV_Params,
                    asid      : asidbits,
                    ptb       : bits(64),
-                   vAddr_arg : bits(64),
+                   vAddr     : bits(64),
                    ac        : AccessType(ext_access_type),
                    priv      : Privilege,
                    mxr       : bool,
                    do_sum    : bool,
                    ext_ptw   : ext_ptw)
                   -> TR_Result(bits(64), PTW_Error) = {
-  let va_mask : bits(64) = zero_extend(ones(sv_params.va_size_bits));
-  let vAddr              = (vAddr_arg & va_mask);
-
   // On first reading, assume lookup_TLB returns None(), since TLBs
   // are not part of RISC-V archticture spec (see TLB_NOTE above)
   match lookup_TLB(asid, vAddr) {


Alasdair:
We should get this fixed, but I think the question we need to answer is how the upper bits of the `sfence.vma 0xFFFFFFFFFFFFFFFF` instruction should be treated? If the virtual addresses are 39-bits does that mean all bits above 39 should be treated the same? i.e. should something like:
```
sfence.vma 0xABCDFFFFFFFFFFFF
```
also clear that same VA?

jrtc27:
From the priv spec:

> If the value held in rs1 is not a valid virtual address, then the SFENCE.VMA instruction has no effect. No exception is raised in this case.

jrtc27:
Of course, one is free to mask off the high bits too, since it's indistinguishable from an implementation that decided to evict that entry from its TLB of its own accord at that exact moment.

Timmmm:
> If the value held in rs1 is not a valid virtual address, then the SFENCE.VMA instruction has no effect. No exception is raised in this case.

That should be the case because only valid virtual addresses make it into the TLB. Unless... you switch virtual memory modes (e.g. Sv57->Sv39) then do `sfence.vma` on an Sv57 address and then switch back or something?

In any case I think it's ok because this note contradicts the quote above:

> It is always legal to over-fence, e.g., by fencing only based on a subset of the bits in rs1 and/or rs2, and/or by simply treating all SFENCE.VMA instructions as having rs1=x0 and/or rs2=x0. 

Not a very clearly written part of the spec, but I think as long as we invalidate *at least* all the matching entries then it is compliant.

Timmmm:
I think it should really say:

> If the value held in rs1 is not a valid virtual address, then the SFENCE.VMA instruction **is not required to have any effect.** No exception is raised in this case.

Timmmm:
@Alasdair `translate()` is called [here](https://github.com/riscv/sail-riscv/blob/4ab15c0b724d8e6d490a6e5cb8c5ba6d5ec9984f/model/riscv_vmem.sail#L433) and you can see a few lines above it returns TR_Failure if the address is not a valid virtual address, so it must already be correctly sign-extended.

Timmmm:
PR to update the spec wording: https://github.com/riscv/riscv-isa-manual/pull/1510

Alasdair (approved):

PR #497, opened 2024-06-18
Remove redundant CSR privilege checks

### LINKED ISSUE #402: is_CSR_defined has redundant privilege mode checks
`is_CSR_defined` has a ton of `p == Machine` etc. checks that should be replaced by `true`, since the permission check is already handled in a generic way by this:

```
function check_CSR(csr : csreg, p : Privilege, isWrite : bool) -> bool =
    is_CSR_defined(csr, p)
  & check_CSR_access(csrAccess(csr), csrPriv(csr), p, isWrite)    <------------- this checks the privilege mode already
  & check_TVM_SATP(csr, p)
  & check_Counteren(csr, p)
  & check_seed_CSR(csr, p, isWrite)
```

Timmmm:
These mode levels are already checked by `csrPriv()` so there's no need to do it again. Removing these checks from `is_CSR_defined()` also means that function name is now correct.

Fixes #402

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -14,68 +14,68 @@
 function csrAccess(csr : csreg) -> csrRW = csr[11..10]
 function csrPriv(csr : csreg) -> priv_level = csr[9..8]
 
-function is_CSR_defined (csr : csreg, p : Privilege) -> bool =
+function is_CSR_defined (csr : csreg) -> bool =
   match (csr) {
     /* machine mode: informational */
-    0xf11 => p == Machine, // mvendorid
-    0xf12 => p == Machine, // marchdid
-    0xf13 => p == Machine, // mimpid
-    0xf14 => p == Machine, // mhartid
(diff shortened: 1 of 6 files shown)


model/riscv_sys_control.sail
```diff
@@ -17,65 +17,65 @@ function csrPriv(csr : csreg) -> priv_level = csr[9..8]
 function is_CSR_defined (csr : csreg, p : Privilege) -> bool =
```
arichardson, line 17: I think we should also be able to remove the `p` parameter here and remove it from `ext_is_CSR_defined` as well

arichardson (approved):
LGTM but I think we can simplify this even more.

model/riscv_sys_control.sail
Timmmm: Ah good call, I'll do that.

Timmmm: Done

jordancarlin (approved):
LGTM

PR #498, opened 2024-06-18
Increase TLB size to 64

Timmmm:
This has a very significant effect on the emulator performance with virtual memory enabled (e.g. when booting Linux).

Having a more realistic TLB can also help expose bugs, like the recent bug in #496. For example with a larger TLB you can add an assertion in `translate()`. It also allows you to detect missing `sfence.vma`s in software by asserting that the TLB translation is the same as the uncached translation. This is not implemented in this change but we can do it in future (or you can just hack it in when needed). The detection will never be 100% but the bigger the TLB the more missing fences you will detect.

64 was chosen based on benchmarks I did last year. Performance when booting Linux keeps going up until you get to 64.

### DIFF
--- a/model/riscv_vmem.sail
+++ b/model/riscv_vmem.sail
@@ -292,7 +292,7 @@ function translate_TLB_hit(sv_params : SV_Params,
                            mxr       : bool,
                            do_sum    : bool,
                            ext_ptw   : ext_ptw,
-                           tlb_index : nat,
+                           tlb_index : tlb_index_range,
                            ent       : TLB_Entry)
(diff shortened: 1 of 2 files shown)


Timmmm:
Shouldn't be merged before #496 is merged, otherwise it makes the bug that that PR fixes way more likely (which is how I discovered it).

Alasdair (approved):

Timmmm:
@billmcspadden-riscv this should be ok to merge now that the fix in #496 is merged.

PR #501, opened 2024-06-19
Use physical addresses for LR/SC reservations

### LINKED ISSUE #360: LR/SC reservation is based on virtual addresses not physical
Currently LR/SC reservation is done using virtual addresses:

            if match_reservation(vaddr) == false then {

And there is this comment:

```
/* We could set load-reservations on physical or virtual addresses.
 * For now we set them on virtual addresses, since it makes the
 * sequential model of SC a bit simpler, at the cost of an explicit
 * call to load_reservation in LR and cancel_reservation in SC.
 */
```

However I don't think this is correct. The spec says:

> Following this model, in systems with memory translation, an SC is allowed to succeed if
> the earlier LR reserved the same location using an alias with a different virtual address,
> but is also allowed to fail if the virtual address is different.

However just matching on virtual addresses means that in theory you can get a successful SC for a physical address that was never reserved:

1. LR 0x100 (physical address 0x500)
2. Change memory mapping
3. SC 0x100 (physical address 0x900) - this will succeed!

I think the model may be saved by the fact that the reservation is cancelled on all exceptions and traps, so the page fault from changing the memory map would clear the reservation. But it's very implicit and frankly seems like it just works through chance.

IMO it should switch to using physical addresses. This be necessary anyway so that `RsrvNone` memory regions can cause an access fault (though helpfully the spec does not say anything at all about how `RsrvNone` is supposed to work).

Timmmm:
This is how reservation is architecturally defined, how real CPUs are likely to be implemented, and doesn't require spurious cancellations on xRET and traps.

Fixes #360.

### DIFF
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -48,9 +48,14 @@ mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)                          
 
 
 /* We could set load-reservations on physical or virtual addresses.
- * For now we set them on virtual addresses, since it makes the
- * sequential model of SC a bit simpler, at the cost of an explicit
- * call to load_reservation in LR and cancel_reservation in SC.
+ * However the reservation is architecturally defined in terms of physical
+ * addresses, and most chips (especially multi-core) will use physical addresses.
+ * Additionally, matching on physical addresses allows as many SC's to
+ * succeed as the spec allows. This simplifies verification against real chips
+ * since you can force spurious failures from a DUT into the Sail model and
+ * then compare the result (a kind of one-way waiver). Using virtual addresses
+ * requires cancelling the reservation in some additional places, e.g. xRET, and
+ * that will break comparison with a DUT that doesn't require cancellation there.
  */
 
 function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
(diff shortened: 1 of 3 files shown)


defermelowie:
This changes the behaviour of `sc` when address translation fails of an address that has no valid reservation set. In the current model the address translation is omitted and therefore no exception occurs. However after this change, the failing address translation causes an exception. Not sure which behaviour is considered "correct".

At first glance the _new_ behaviour seems to match spike: https://github.com/riscv-software-src/riscv-isa-sim/blob/98d2c29e431f3b14feefbda48c5f70c2f451acf2/riscv/mmu.h#L266

Alasdair:
I suspect the answer is it's implementation defined what happens. Right now the RISC-V axiomatic memory model in the ISA manual does not specify anything regarding virtual memory.

If we do end up with an axiomatic model with virtual memory, we might need something like `load_reservation(vAddr, pAddr)` to capture all the allowed behaviors, assuming implementations have a choice in how they implement LR and SC.

Alasdair (approved):

model/riscv_insts_aext.sail
```diff
@@ -48,9 +48,14 @@ mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)
 
 
 /* We could set load-reservations on physical or virtual addresses.
- * For now we set them on virtual addresses, since it makes the
- * sequential model of SC a bit simpler, at the cost of an explicit
- * call to load_reservation in LR and cancel_reservation in SC.
+ * However the reservation is architecturally defined in terms of physical
```
Alasdair, line 51: Is this actually the case? The LR and SC section of the unpriv spec never uses the word physical. "a set of bytes that subsumes the bytes in the addressed word." seems intentionally fuzzy to allow a range of implementations.

Alasdair: OTOH, when I search the priv spec: "For implementations with both page-based virtual memory and the "A" standard extension, the LR/SC
reservation set must lie completely within a single base physical page (i.e., a naturally aligned 4 KiB
physical-memory region)."

Timmmm: This is based on [this comment](https://github.com/riscv/sail-riscv/issues/360#issuecomment-1890399132); maybe the language is a little too strong for something in an second hand email... Maybe I should just remove that comment? Though the quote you found does indirectly support it.

Timmmm: Ok comment removed.

Timmmm:
@billmcspadden-riscv ok this is ready now.

PR #508, opened 2024-07-04
Update register reads/writes to use range and type aliases instead of forall 'n

### LINKED ISSUE #426: Switch forall to ranges/sets
There are a lot of type declarations using e.g. `forall 'n, 0 < 'n <= 8 .` that could be using the `range(0, 8)`, and some that could be switched to the set notation `{1, 2, 3}`. This would be simpler and easier to read, and we can use type aliases, e.g. instead of

```
type regno ('n : Int), 0 <= 'n < 32 = atom('n)

val rX : forall 'n, 0 <= 'n < 32. regno('n) -> xlenbits
function rX r = {
```

we can do

```
type regno = range(0, 31)

function rX(r : regno) -> xlenbits = {
```

which is much simpler and clearer (especially to people unfamiliar with fancy existential types, which is approximately everyone!).

jordancarlin:
Replace `forall 'n, 0 < 'n <= *` with `range(0, *)` for register access. With this change, many of them can also switch to using the already defined type aliases.

Fixes #426

### DIFF
--- a/model/riscv_types.sail
+++ b/model/riscv_types.sail
@@ -28,10 +28,9 @@ type csreg   = bits(12)   /* CSR addressing */
 
 /* register file indexing */
 
-type regno ('n : Int), 0 <= 'n < 32 = int('n)
+type regno = range(0, 31)
 
-val regidx_to_regno : bits(5) -> {'n, 0 <= 'n < 32. regno('n)}
-function regidx_to_regno b = let 'r = unsigned(b) in r
+function regidx_to_regno (b : regidx) -> regno = unsigned(b)
 
 /* mapping RVC register indices into normal indices */
 val creg2reg_idx : cregidx -> regidx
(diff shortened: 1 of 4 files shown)


model/riscv_types.sail
```diff
@@ -28,10 +28,9 @@ type csreg   = bits(12)   /* CSR addressing */
 
 /* register file indexing */
 
-type regno ('n : Int), 0 <= 'n < 32 = int('n)
+type regno = range(0, 31)
 
-val regidx_to_regno : bits(5) -> {'n, 0 <= 'n < 32. regno('n)}
-function regidx_to_regno b = let 'r = unsigned(b) in r
+function regidx_to_regno (b : regidx) -> regno = let 'r = unsigned(b) in r
```
Timmmm, line 33: I suspect this could just be `= unsigned(b)`

Timmmm (commented):
Great improvement!

model/riscv_types.sail
jordancarlin: Good catch. Updated.

Timmmm (approved):

PR #510, opened 2024-07-04
Add svnapot extention support

Yui5427:


Timmmm (commented):
This looks like it allows any NAPOT size, whereas the spec currently only allows 64kB ones.

I think the bit manipulation can probably be simplified too but I'd have to think about that.

Yui5427:
I check the code in PMP,It really lights me,And I think the additional code would solve the problem.It adds the 64K check.And use the PMP napot bit manipulation to count the napot_mask. And replaced "perfect_one" with zero_extend(0b1).lol :)

model/riscv_vmem.sail
```diff
@@ -164,7 +164,30 @@ function pt_walk(sv_params,
...
+                                (vpns_of_va(sv_params, va) & napot_mask) | 
+                                (vpns_of_va(sv_params, va) & ((zero_extend(0b1) << ptshift) - zero_extend(0b1)));
+
                 let pa = (ppns << pagesize_bits) | zero_extend(offset_of_va(va));
+                let pa_napot = page_base << 12 | zero_extend(offset_of_va(va));
+
+                // For svnapot ext purpose.
+                if (pte & (zero_extend(0b1) << 63)) == (zero_extend(0b1) << 63) then {
+                  if (napot_mask != zero_extend(0xf)) then {
+                    return PTW_Failure(PTW_Invalid_PTE(), ext_ptw)
```
Timmmm, line 185: Did you mean `PTW_Unsupported_Napot` here?

I'd probably call it `PTW_Reserved_Napot` btw.

model/riscv_vmem.sail
```diff
@@ -164,7 +164,30 @@ function pt_walk(sv_params,
...
+                let page_base = (ppn & napot_mask_not) | 
+                                (vpns_of_va(sv_params, va) & napot_mask) | 
+                                (vpns_of_va(sv_params, va) & ((zero_extend(0b1) << ptshift) - zero_extend(0b1)));
+
                 let pa = (ppns << pagesize_bits) | zero_extend(offset_of_va(va));
+                let pa_napot = page_base << 12 | zero_extend(offset_of_va(va));
+
+                // For svnapot ext purpose.
+                if (pte & (zero_extend(0b1) << 63)) == (zero_extend(0b1) << 63) then {
+                  if (napot_mask != zero_extend(0xf)) then {
```
Timmmm, line 184: Easier like this: `napot_mask[3 .. 0] != ones()`

model/riscv_vmem.sail
```diff
@@ -164,7 +164,30 @@ function pt_walk(sv_params,
                 }
               }
               else {
+                let mask_bits = level * sv_params.pte_PPN_j_size_bits;
```
Timmmm, line 167: I haven't gone through the logic properly but are you sure this is right? `level` is guaranteed to be 0 here.

Timmmm:
You also need to update `pte_is_invalid()` to check the `N` bit, and add a `haveSvnapot()` function.

ved-rivos:
Svnapot was queued by an earlier PR https://github.com/riscv/sail-riscv/pull/393
Its unfortunate that the PR backlog is leading to duplicate work.

jrtc27 (changes requested):
Needs to be opt-in/out-able

Timmmm:
> Svnapot was queued by an earlier PR https://github.com/riscv/sail-riscv/pull/393
> Its unfortunate that the PR backlog is leading to duplicate work.

I agree, sorry I forgot about your PR. I think we should close this one in favour of yours which is much more complete.

PR #516, opened 2024-07-22
Fix innocuous issues with extension tagging

ThinkOpenly:
- The `end` statement for `scattered enum extension` misspelled `extension`.
- `end extensionEnabled` was inadvertently dropped, and now been restored.


ThinkOpenly:
These issues snuck into PR #506 .

Alasdair (approved):
I fixed the issue that let these slip past in the first place, so we should get this merged otherwise you'll now see:
```
Type error:
model/riscv_insts_end.sail:30.0-14:
30 |end extensions
   |^------------^
   | extensions is not a scattered definition, so it cannot be ended
```

PR #517, opened 2024-07-28
Use Zc* extensions instead of just the C extension

jordancarlin:
The C extension was subdivided into several smaller extensions (Zca, Zcf, Zcd). The sail model already had the instructions for these separate extensions split into separate files, but they were all enabled by the C extension. This PR switches them to be gated by their respective Zc* extensions.

For now, Zca (which Zcf and Zcd both require) is enabled when C is enabled to keep consistency with the current configuration. Once model configuration is possible we should change this to support any legal configuration (ie RV32FD_Zca without Zcf or Zcd).


model/riscv_insts_zcb.sail
```diff
@@ -7,7 +7,7 @@
 /*=======================================================================================*/
 
 enum clause extension = Ext_Zcb
-function clause extensionEnabled(Ext_Zcb) = sys_enable_zcb()
+function clause extensionEnabled(Ext_Zcb) = sys_enable_zcb() & extensionEnabled(Ext_Zca)
```
jrtc27, line 10: This is a separate behavioural change

jordancarlin: Good point. I can break that out into a separate PR.

arichardson (approved):

billmcspadden-riscv:
@jrtc27 ,  are you satisfied with the changes made per your request?

jrtc27 (approved):
I think this is ok now

billmcspadden-riscv:
@jrtc27 :  Thanks.

Timmmm:
@billmcspadden-riscv for PRs like this please can you squash them and set the commit message to something informative (i.e. in this case it should be the text from the PR "The C extension was subdivided...").

Otherwise the actual Git history - which is the thing that is preserved for all time, and what you see in editors etc. - just says "Use Zc* extentions" which is very unhelpful.

Next to the merge button there is a drop-down where you can select "Squash and Merge":

image

Then when you click it you get to edit the message and you can make sure it contains the relevant information.

For *some* PRs (e.g. the vector one) we probably want to preserve history, but I think for the vast majority it is not relevant and people don't do a good job of writing commit messages anyway.

PR #522, opened 2024-08-13
Zcb requires Zca

jordancarlin:
Per section 29.8 of the Unprivileged ISA manual, extension Zcb requires Zca. This updates the extensionEnabled clause to capture this requirement.

### DIFF
--- a/model/riscv_insts_zcb.sail
+++ b/model/riscv_insts_zcb.sail
@@ -7,7 +7,7 @@
 /*=======================================================================================*/
 
 enum clause extension = Ext_Zcb
-function clause extensionEnabled(Ext_Zcb) = sys_enable_zcb()
+function clause extensionEnabled(Ext_Zcb) = sys_enable_zcb() & extensionEnabled(Ext_Zca)
 
 union clause ast = C_LBU : (bits(2), cregidx, cregidx)
 


jrtc27 (changes requested):
sys_enable_zcb shouldn't be returning true in the first place if Zca isn't enabled

Alasdair:
In general for Sail backends that reason about the spec (Isabelle/Coq/SMT etc) you can only really treat external functions as uninterpreted functions, so there is nothing preventing `sys_enable_zcb` from returning true when `sys_enable_zca` returns false, and e.g. counterexample generation will happily produce a model with that being the case.

jrtc27:
> In general for Sail backends that reason about the spec (Isabelle/Coq/SMT etc) you can only really treat external functions as uninterpreted functions, so there is nothing preventing `sys_enable_zcb` from returning true when `sys_enable_zca` returns false, and e.g. counterexample generation will happily produce a model with that being the case.

F vs Zfinx is a similar case. We don't check `F & not(Zfinx)` everywhere, we just require that they be mutually exclusive (and assert that in some places). If you really wanted you could have a `check_model_config()` that asserts various things like this and can be a precondition for whatever you want to prove.

jordancarlin:
It is also possible for an implementation to support both Zca and Zcb (which is what the sys_enable functions mean) and then disable Zca (for example, by writing a 0 to MISA[C]). In that scenario, sys_enable_zcb would still be returning true as it is supported in the implementation, but the extensionEnabled function should return false.

jrtc27:
Ah that might be a good justification for this specific case. I imagine there’s no precedent yet in the model for a multi-letter extension that depends on a toggleable single-letter one?

jordancarlin:
I don't think so. The most similar would be B vs Zb* and A vs. Zaamo/Zalrsc, but both of those relationships are different from this one.

jordancarlin:
@jrtc27 are there still changes to this that you want based on the above discussion?

jordancarlin:
@jrtc27 Following up on this. Any updates you’d still like to see?

Timmmm:
Seems like the issues have been resolved so I'll merge this in a couple of days. 

I do think we need to at least explicitly state somewhere a policy for exactly how the `sys_enable_` and `extensionEnabled()` functions should work. I think I put a comment for`extensionEnabled()` but there are probably edge cases like dependent extensions that it doesn't cover.

We can probably figure it all out when the new config system is done anyway.

PR #527, opened 2024-08-27
Remove int_power and use built-in 2 ^

Timmmm:
Sail has built-in support for `2 ^`, and this is the only exponent we use so there's no need for the more generic `int_power`. Additionally the type of `int_power` is both way looser than `2 ^` (which is understood at the type level), and actually wrong, e.g. it will let you do `3 ^ -1` and give the result as 3.

Note I had to add spaces because this was required until a very recent Sail version - see https://github.com/rems-project/sail/issues/657

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -39,9 +39,7 @@ function bit_str(b: bit) -> string =
 
 overload BitStr = {bits_str, bit_str}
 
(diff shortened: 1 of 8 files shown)


Alasdair (approved):
LGTM

I think the semantics of int_power should be consistent at least, when the second operand is negative it'll use the absolute value.

Issue #449, opened 2024-04-12
Implementing callbacks in sail-riscv for state-changing events


kseniadobrovolskaia:
Hello! 

I am reaching out to discuss some enhancements we are considering for our RISC-V processors test generator. Currently, we are utilizing SAIL as our golden model. To streamline our workflow, we believe it would greatly benefit our project to receive notifications when the model undergoes changes in its state.

Our proposed solution involves implementing callbacks, which would involve invoking functions via pointers whenever a specified change occurs within the model's state. Specifically, we are seeking to incorporate the following function declarations within the SAIL code:

Type Definitions:
```
Callback handler:   typedef struct callback_handler callback_handler;
Callback Functions:
    mem_update_callback_t
    xreg_update_callback_t
    freg_update_callback_t
    vreg_update_callback_t
    pc_update_callback_t
```
Following the implementation of these declarations, any user would have the ability to furnish their own implementation of the `handler` and associated functions. Consequently, this would enable immediate actions to be executed upon any state changes within the model.

In order to integrate this functionality, we would require calls to these functions (if defined) at the time of the changes within the following function definitions:
```
void zphys_mem_write(struct zMemoryOpResultzIozK *zcbz3590, enum zwrite_kind zwk, uint64_t zpaddr, int64_t zwidth, lbits zdata, unit zmeta)
unit zwX(int64_t zr, uint64_t zin_v)
unit zwF(int64_t zr, uint64_t zin_v)
unit zwrite_single_vreg(sail_int znum_elem, sail_int zSEW, uint64_t zvrid, zz5vecz8z5bvz9 zv)
unit zset_next_pc(uint64_t zpc)
```
Could you please provide your feedback on the feasibility of incorporating this functionality into sail-riscv? We believe it could significantly enhance our testing capabilities, but we would greatly appreciate your insights and guidance on this matter.

Thank you very much for your time and consideration.

Best regards,
Ksenia Dobrovolskaya.

Timmmm:
I agree we need a proper interface for this. Currently the only option is the RVFI code. We have hackily commented all that out and replaced it with our own RVFI-alike.

I'm not really a fan of using callbacks in general but it does nicely avoid some problems with the RVFI approach which I've run into - especially when multiple things can happen in one instruction (e.g. multiple CSR writes). So yeah I think this is a good idea.

The way to do it would be to add external Sail functions like this:

```
[...]

PR #528, opened 2024-08-27
Add an option to delimit instructions in the trace

Alasdair:
See: https://github.com/riscv/sail-riscv/issues/526

An extra newline is inserted after each 'step' in the simulator, if the appropriate trace option is provided. The spaces that were previously in the trace were just an artifact of an extra space inserted when we flushed the trace output (which just happened to coincide with the step). These should actually be guaranteed to delimit each instruction.

Supports both the OCaml and C emulators. For the C emulator this is enabled with `--trace=step`, and for OCaml `-trace-delimit-step`. It works slightly differently in both cases, because of how the two emulators handle the toplevel loop differently - but care was taken to ensure if you did run the sail `loop` function in C, it will behave the same.

The C simulator `--trace` option had it's error message formatted oddly when the passed option was unrecognised, with the expected and provided argument flipped, so that has also been fixed.

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -89,6 +89,7 @@ bool config_print_reg = true;
 bool config_print_mem_access = true;
 bool config_print_platform = true;
 bool config_print_rvfi = false;
+bool config_print_step = false;
 
 void set_config_print(char *var, bool val)
 {
@@ -108,9 +109,11 @@ void set_config_print(char *var, bool val)
     config_print_rvfi = val;
   } else if (strcmp("platform", var) == 0) {
     config_print_platform = val;
+  } else if (strcmp("step", var) == 0) {
+    config_print_step = val;
   } else {
-    fprintf(stderr, "Unknown trace category: '%s' (should be %s)\n",
-            "instr|reg|mem|platform|all", var);
+    fprintf(stderr, "Unknown trace category: '%s' (should be %s)\n", var,
+            "instr|reg|mem|rvfi|platform|step|all");
     exit(1);
   }
 }
@@ -989,6 +992,9 @@ void run_sail(void)
       KILL(sail_int)(&sail_step);
     }
     if (stepped) {
+      if (config_print_step) {
+        fprintf(trace_log, "\n");
+      }
       step_no++;
       insn_cnt++;
       total_insns++;
--- a/c_emulator/riscv_config.h
+++ b/c_emulator/riscv_config.h
@@ -2,6 +2,7 @@
 #include <stdbool.h>
 
 extern bool config_print_instr;
+extern bool config_print_step;
 extern bool config_print_reg;
 extern bool config_print_mem_access;
 extern bool config_print_platform;
--- a/c_emulator/riscv_prelude.c
+++ b/c_emulator/riscv_prelude.c
@@ -15,6 +15,13 @@ unit print_instr(sail_string s)
   return UNIT;
 }
 
+unit print_step(unit u)
+{
+  if (config_print_step)
+    fprintf(trace_log, "\n");
+  return UNIT;
+}
+
 unit print_reg(sail_string s)
 {
   if (config_print_reg)
--- a/c_emulator/riscv_prelude.h
+++ b/c_emulator/riscv_prelude.h
@@ -6,6 +6,7 @@
 unit print_string(sail_string prefix, sail_string msg);
 
 unit print_instr(sail_string s);
+unit print_step(unit u);
 unit print_reg(sail_string s);
 unit print_mem_access(sail_string s);
 unit print_platform(sail_string s);
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -75,6 +75,10 @@ val print_reg      = {ocaml: "Platform.print_reg", interpreter: "print_endline",
 val print_mem      = {ocaml: "Platform.print_mem_access", interpreter: "print_endline", c: "print_mem_access", lem: "print_dbg", _: "print_endline"} : string -> unit
 val print_platform = {ocaml: "Platform.print_platform", interpreter: "print_endline", c: "print_platform", lem: "print_dbg", _: "print_endline"} : string -> unit
 
+val print_step = {ocaml: "Platform.print_step", c: "print_step"} : unit -> unit
+
+function print_step() = ()
+
 val get_config_print_instr = {ocaml: "Platform.get_config_print_instr", c:"get_config_print_instr"} : unit -> bool
 val get_config_print_reg = {ocaml: "Platform.get_config_print_reg", c:"get_config_print_reg"} : unit -> bool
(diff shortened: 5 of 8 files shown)


c_emulator/riscv_sim.c
```diff
@@ -108,9 +109,11 @@ void set_config_print(char *var, bool val)
     config_print_rvfi = val;
   } else if (strcmp("platform", var) == 0) {
     config_print_platform = val;
+  } else if (strcmp("step", var) == 0) {
+    config_print_step = val;
   } else {
-    fprintf(stderr, "Unknown trace category: '%s' (should be %s)\n",
-            "instr|reg|mem|platform|all", var);
+    fprintf(stderr, "Unknown trace category: '%s' (should be %s)\n", var,
+            "instr|reg|mem|rvfi|platform|step|all");
```
billmcspadden-riscv, line 116: Why is 'rvfi' being added here?  Was it never part of this message?

billmcspadden-riscv (commented):
I want to hear back from Umer Shahid before accepting this PR.  Specifically, he
requested 2 newlines  between instructions. 

Is "print_step()" the best function name for this?  Perhaps "print_step_newline()"?

c_emulator/riscv_sim.c
Alasdair: It's a valid option for the --trace flag, it was just omitted so I fixed that

Alasdair:
I think we only need 1 extra newline, because the previous line already ends with a newline. The commit that removed the extra newline also only removed a single one.

Alasdair:
> Is "print_step()" the best function name for this? Perhaps "print_step_newline()"?

I called it `riscv_step` because I think we will probably want to add some extra end-of-instruction output to the trace, and we should discuss what that should be in the meeting.

Parsing based on number of newlines is likely too brittle. We shouldn't make that the official way of detecting an end of instruction. The real issue here is our trace doesn't have a way to show which events are together as part of an instruction. This should solve the riscv-isac issue for now, but this issue is really just Hyrum's law in action:

> _With a sufficient number of users of an API, it does not matter what you promise in the contract: all observable behaviors of your system will be depended on by somebody._

At the end of the day, I don't think we can make very strong guarantees right now about our exact trace format. As e.g. certification needs evolve, we might need to add additional information or change the format. If downstream tools want to parse it with a bunch of regexes that's fine, but they should expect some non-zero amount of breakage and set up their CI appropriately.

UmerShahidengr:
@Alasdair Thanks alot for the quick work. This new flag is generating the log as desired.
@billmcspadden-riscv thanks for your attention too.

arichardson:
Relying on a single extra newline seems rather fragile - if we are adding an option to delimit the traces, can we add another more explicit separator (e.g. a line containing something like `--- end step N`).

I guess this is fine as a quick workaround, but I was under the assumption that the trace output is just debug information used to investigate test executions. Consuming the output should use some form of machine-readable output instead.

Timmmm:
> Consuming the output should use some form of machine-readable output instead.

I very strongly agree with this. IMO it's probably ok to just unilaterally change the output to unblock riscv isac if it really is urgent, but they are clearly Doing It Wrong (tm).

I don't think we should make it configurable to facilitate long term Doing It Wrong. Instead we should finish #449 and add a machine readable output option.

If it were up to me I would regularly change the log output in order to prevent people thinking it's ok to parse it, but I appreciate not everyone cares as much about robustness as me. Maybe we could add a message at the start at least like `This format is not stable or machine readable; do not parse it.` (after we've done #449).

rsnikhil:
>      Consuming the output should use some form of machine-readable output instead.

Does there already exist a defined spec for such a binary trace?

Timmmm:
I don't know of one. There's RVFI-DII, sort of, but it's missing a lot of features.

allenjbaum:
Well, the ISAC parsing was based on how the log file was formatted; it's
not like ISAC forced a specific format.
And I'm unclear what it means to be a machine readable format - it is a
machine readable format now, since it is being programmatically parsed.
I suspect you mean something very specific by that comment - could you
explain what you mean a bit more clearly for me?

Alasdair:
By machine-readable, I mean in a precisely specified and unambiguous format that is intended specifically to be consumed by other programs. For example in Isla we have a S-expression based trace and formal grammar for the trace language. You could also have trace that's e.g. a sequence of JSON objects with a specified JSON schema, or anything along those lines. If a more formally specified trace format changes to accomodate new events, you also can have a changelog/grammar update that precisely documents to downstream tools the ways in which the format has changed, so there's no guesswork involved.

Obviously you can always throw some regular-expressions at a trace that is intended to be human-readable, but we can provide no guarantees that we will not break your regular-expression based parser at some point in the future. In this case, removing the spaces between instructions does not make the trace any harder to understand for a human reader - it's fairly clear where instructions begin and end - so why would we put extra spaces in?

allenjbaum:
I interpret that as simply wanting a documented schema; clearly this can be
consumed by other programs (ISAC for one, but I believe there are others).
It just isn't documented, except by inspection. That's fixable, but clear
there was a change that broke that ad-hoc schema,
and without a documented schema it's hard (impossible) to know that you did
that.
I've been trying to get the log file to be more regular and self describing
(so every line shows what is being fetched, and why
 (e.g. level3 page table fetch, writeback of AD bits to the PTE, etc)

Alasdair:
> It just isn't documented, except by inspection.

It might also be worth noting that in this case the inspection was wrong. The spaces did not delimit instructions, there could be have been extra spaces within a single instruction. This commit also changes where the spaces appear to match the assumption that instructions are delimited by a blank line. Prior to this commit blank lines had no actual defined meaning in the trace.

Timmmm:
"Machine readable" means it is intended to be read by machines and designed in a way to make it easy and reliable for machines to read it.

Neither is true for this log output. We should use something line JSON or TSV or Protobuf or similar.

I've actually had a lot of success with SQLite for similar applications in the past. It's universally supported, gives you a ton of useful features (e.g. indexes, self-documentation, a nice GUI), and is faster and more space efficient than it has any right to be.

However it isn't something you can pipe into another program "live" which maybe rules it out here. I would say JSON lines would be a good first stab but JSON is very space inefficient and these logs can be huge.

rsnikhil:
>    ... "Machine readable" means it is intended to be read by machines
>    ... these logs can be huge.

For these reasons, I think the logs should be in dense binary form, so they're trivially parsed and compact.

Post-processing tools can expand this into human-readable form or any other desired format.

Timmmm:
For the machine readable output definitely. We can support both machine readable and human readable output (i.e. this) via a flag. `--format=text`, `--format=binary`, etc.

rsnikhil:
> For the machine readable output definitely. We can support both machine readable and human readable output (i.e. this) via a flag. `--format=text`, `--format=binary`, etc.

Do we know how much execution-time overhead goes into producing text logs?

Timmmm:
Haven't benchmarked it but I imagine very little.

billmcspadden-riscv (commented):
LGTM

billmcspadden-riscv (approved):

PR #531, opened 2024-08-29
Remove N extension (user mode interrupts)

KotorinMinami:
Since the RISC-V N extension draft was removed in the priv spec 1.2 version, the CSR sedeleg is also not listed in the CSR table in the spec. Therefore, it should be removed. 

By the way, the implementation of the N extension in the master branch of Sail might be redundant or confusing and lack rigor. I believe it may need to be removed or moved to another branch.

### DIFF
--- a/model/riscv_insts_vext_mem.sail
+++ b/model/riscv_insts_vext_mem.sail
@@ -146,7 +146,7 @@ function process_vlsegff (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem)
   var trimmed : bool = false;
   foreach (i from 0 to (num_elem - 1)) {
     if not(trimmed) then {
-      if vm_val[i] then { /* active segments */
+      if mask[i] then { /* active segments */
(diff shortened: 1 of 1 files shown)


Timmmm:
Seems to inconsistently remove `sedeleg` and `sideleg`. And why only remove those aspects of the N extension? Either remove the entire N extension or none of it surely?

arichardson:
> Seems to inconsistently remove `sedeleg` and `sideleg`. And why only remove those aspects of the N extension? Either remove the entire N extension or none of it surely?

Agreed, I think it would make more sense to completely remove the N extension support. I don't believe this was ever used?

KotorinMinami:
@Timmmm Of course I would like to remove the entire N extension. However, some matters have been occupying my time, and due to some unresolved issues in the execution flow of SAIL and in the privileged level interrupt permissions, I haven't been able to complete the work until now. It may be finished at the moment. Could you please spend a convenient time to review my code?

Issue #509, opened 2024-07-04
Remove OCaml emulator


Timmmm:
Can we remove the OCaml emulator? I think this would have significant benefits:

* Less confusing for new users - they no longer need to choose between two very similar but different emulators.
* We no longer have to implement emulator features twice.
* Contributors don't need to learn OCaml.
* Shorter CI time.
* When we use the binary release of the Sail compiler, this will mean we can entirely drop OCaml as a dependency, even in CI.

trdthg:
Discussions from the past: #138

martinberger:
Another problem with the Ocaml emu is 

- Very slow!

arichardson:
I've also found having to add the ocaml callbacks for every new features a bit annoying in the past (especially since I can't write ocaml and just copy-paste). So +1 from me.

jordancarlin:
Is this something that we're planning to proceed with? I can look into it if so.

Timmmm:
Yeah in the weekly meeting we basically agreed to do it and I said I would make a PR, but I've only ~30% completed it.

One thing we didn't discuss which is a unique OCaml backend feature is the thing where it can generate random instructions based on the `encdec` mapping, which I believe is used by TestRig. It doesn't depend on the emulator or OCaml callbacks but we might want to still keep the "generate OCaml code" Makefile targets which it presumably uses (I haven't checked yet).

PR #532, opened 2024-08-29
Install Sail from binary in CI

Timmmm:
This should be a lot faster. It also means the version is pinned properly which wasn't the case before.

I also recommended users to do the same in the README do they don't have to deal with OPAM.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -7,19 +7,19 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - name: Install packages
-      run: sudo apt install -y opam zlib1g-dev pkg-config libgmp-dev z3 device-tree-compiler
+      run: sudo apt install -y --no-install-recommends zlib1g-dev pkg-config libgmp-dev curl
     - name: Check out repository code
       uses: actions/checkout@HEAD
       with:
         submodules: true
     - name: Ensure pre-commit checks pass
-      run: pip install pre-commit && pre-commit run --all-files --show-diff-on-failure --color=always
-    - name: Init opam
-      run: opam init --disable-sandboxing -y
-    - name: Install sail
-      run: opam install -y sail
+      run: python3 -m pip install pre-commit && pre-commit run --all-files --show-diff-on-failure --color=always
+    - name: Install sail from binary
+      run: |
+        sudo mkdir -p /usr/local
+        curl --location https://github.com/rems-project/sail/releases/download/0.18-linux-binary/sail.tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
     - name: Build and test simulators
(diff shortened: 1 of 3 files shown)


.github/workflows/compile.yml
```diff
@@ -7,19 +7,17 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - name: Install packages
-      run: sudo apt install -y opam zlib1g-dev pkg-config libgmp-dev z3 device-tree-compiler
+      run: sudo apt install -y opam zlib1g-dev pkg-config libgmp-dev device-tree-compiler curl
```
jordancarlin, line 10: I think we should be able to get rid of opam now as well.

Timmmm: I think it's still needed to build the OCaml version of the emulator which hasn't been removed yet. Actually I may need to keep the `opam init`, but I'll see what happens.

jordancarlin: We might be able to just install OCaml directly in that case and avoid `opam init`.

Alasdair:
A trick I use in the Sail CI to speed it up is to cache the `~/.opam` directory: https://github.com/rems-project/sail/blob/sail2/.github/workflows/build.yml

Timmmm:
I'm hoping we can get away without OPAM as @jordancarlin suggested.

If this attempt doesn't work I'm writing a Dockerfile so I can try it locally! The edit-build-test cycle for Github CI is ridiculous.

jordancarlin:
> Well, the C side seemed to work, but the OCaml build fails because it can't find `ocamlfind`, and we use `-use-ocamlfind`.
> 
> Maybe I'll just finish #509 first.

That probably makes sense because otherwise it’s just unnecessary work that’ll be quickly obsolete.

.github/workflows/compile.yml
```diff
@@ -7,19 +7,19 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - name: Install packages
-      run: sudo apt install -y opam zlib1g-dev pkg-config libgmp-dev z3 device-tree-compiler
+      run: sudo apt install -y --no-install-recommends zlib1g-dev pkg-config libgmp-dev device-tree-compiler curl
```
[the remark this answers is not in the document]
Timmmm: Done

jordancarlin (approved):
Looks great. Should make it much easier for new users to get going without needing to worry about OPAM. And seems like it speeds up CI by another few minutes too.

Timmmm:
I think this should be ready to go, the only thing I'm not sure about is the Coq stuff. It was last touched 4 years ago by @bacam - Brian do you remember this?

arichardson (approved):

Alasdair (approved):

PR #535, opened 2024-09-04
Remove incorrect privilege assert

PeterRugg:
The condition checks whether it's possible for a mode other than machine mode to take interrupts, but just because it can't, that doesn't mean we can never call this function while in another mode.
In particular, this causes a crash if you run with S mode disabled but U mode enabled, then mret to U mode without NExt.

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -277,19 +277,28 @@ function dispatchInterrupt(priv : Privilege) -> option((InterruptType, Privilege
   /* If we don't have different privilege levels, we don't need to check delegation.
    * Absence of U-mode implies absence of S-mode.
    */
-  if not(extensionEnabled(Ext_U)) | (not(extensionEnabled(Ext_S)) & not(extensionEnabled(Ext_N))) then {
-    let enabled_pending = mip.bits & mie.bits;
-    match findPendingInterrupt(enabled_pending) {
-      Some(i) => let r = (i, Machine) in Some(r),
-      None()  => None()
-    }
-  } else {
+  let multipleModesSupported = extensionEnabled(Ext_U);
+  if not(multipleModesSupported) then {
+    assert(priv == Machine, "invalid current privilege")
+  };
+  /* Even with U-mode, we don't need to check delegation when M-mode is the only
+     one that can take interrupts, i.e. when we don't have S-mode and don't have
+     the N extension
+  */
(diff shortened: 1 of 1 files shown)


Alasdair (approved):

model/riscv_sys_control.sail
```diff
@@ -278,7 +278,6 @@ function dispatchInterrupt(priv : Privilege) -> option((InterruptType, Privilege
    * Absence of U-mode implies absence of S-mode.
```
Timmmm, line 278: I think the comment is a bit wrong and the check for S mode is redundant. Should it not be something like this?

```
// If we don't have S or N then interrupts cannot be delegated and must
// always be handled in machine mode.
if not(extensionEnabled(Ext_S)) & not(extensionEnabled(Ext_N) then {
  let enable_pending...
```

Removing the assertion looks correct to me. I would guess it was correct when written, and the not removed when N was added.

PeterRugg: Yeah, that was my interpretation as well. I'll reword the comment.

PeterRugg: The title of this PR is no longer accurate, but I think that has tidied up the function a lot.

model/riscv_sys_control.sail
```diff
@@ -277,21 +277,29 @@ function dispatchInterrupt(priv : Privilege) -> option((InterruptType, Privilege
...
+  if multipleModesSupported & delegationPossible then {
     match getPendingSet(priv) {
       None()      => None(),
       Some(ip, p) => match findPendingInterrupt(ip) {
                        None()  => None(),
                        Some(i) => let r = (i, p) in Some(r)
                      }
     }
+  } else {
+    let enabled_pending = mip.bits & mie.bits;
```
Timmmm, line 298: Maybe worth adding a comment to say this is a fast path for M-only CPUs, and isn't distinct behaviour from the other branch.

Timmmm: Yeah I don't think it needs to though - later in the function it correctly checks `if not(extensionEnabled(Ext_S)) `.

We probably also want `legalize_mideleg()` to hard-wire `mideleg` to 0 if S mode is not enabled, but I don't think it technically would break things.

I guess we can leave it for now anyway.

Timmmm (approved):

Timmmm:
Merging in a couple of days if nobody objects.

PR #538, opened 2024-09-09
Add Zicbop extension

trdthg:
Add prefetch hint instructions. The Sail model has no caches so this has no effect on execution, however it does make the disassembly output more useful and show what the encoding should be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ For booting operating system images, see the information under the
 - Zicsr extension for CSR instructions, v2.0
 - Zicntr and Zihpm extensions for counters, v2.0
 - Zicond extension for integer conditional operations, v1.0
-- Zicbom and Zicboz extensions for cache-block management (Zicbop not currently supported), v1.0
+- Zicbom, Zicbop and Zicboz extensions for cache-block management, v1.0
 - Zimop extension for May-Be-Operations, v1.0
 - Zihintntl extension for Non-temporal Locality Hints, v1.0
 - Zihintpause extension for Pause Hint, v2.0
--- a/model/riscv.sail_project
+++ b/model/riscv.sail_project
@@ -338,6 +338,19 @@ extensions {
     }
   }
 
+  Zicbop {
+    Zicbop_types {
+      before riscv
+      files riscv_zicbop_types.sail
+    }
+    Zicbop_insts {
+      // PREFETCH.{I,R,W} instructions override ORI hints (rd=0)
+      before I_insts
+      requires riscv, Zicbop_types
+      files riscv_insts_zicbop.sail
+    }
+  }
+
   Zicboz {
     requires riscv
     files riscv_insts_zicboz.sail
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -76,6 +76,10 @@ function clause hartSupports(Ext_H) = false
 enum clause extension = Ext_Zicbom
 mapping clause extensionName = Ext_Zicbom <-> "zicbom"
 function clause hartSupports(Ext_Zicbom) = config extensions.Zicbom.supported
+// Cache-Block Prefetch Instructions
+enum clause extension = Ext_Zicbop
+mapping clause extensionName = Ext_Zicbop <-> "zicbop"
+function clause hartSupports(Ext_Zicbop) = config extensions.Zicbop.supported
 // Cache-Block Zero Instructions
 enum clause extension = Ext_Zicboz
 mapping clause extensionName = Ext_Zicboz <-> "zicboz"
@@ -401,6 +405,7 @@ let extensions_ordered_for_isa_string = [
   // Z extensions, ordered by category and then alphabetically.
   // Zi
   Ext_Zicbom,
+  Ext_Zicbop,
   Ext_Zicboz,
   Ext_Zicntr,
   Ext_Zicond,
--- a/model/riscv_insts_zicbop.sail
+++ b/model/riscv_insts_zicbop.sail
@@ -1,0 +1,35 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+// Cache Block Operations - Prefetch
+
+function clause currentlyEnabled(Ext_Zicbop) = hartSupports(Ext_Zicbop)
+
+/* ****************************************************************** */
+union clause instruction = ZICBOP : (cbop_zicbop, regidx, bits(12))
+
+mapping encdec_cbop_zicbop : cbop_zicbop <-> bits(5) = {
+  PREFETCH_I <-> 0b00000,
+  PREFETCH_R <-> 0b00001,
+  PREFETCH_W <-> 0b00011,
+}
+
+mapping clause encdec = ZICBOP(cbop, rs1, offset11_5 @ 0b00000)
+  <-> offset11_5 @ encdec_cbop_zicbop(cbop) @ encdec_reg(rs1) @ 0b110 @ 0b00000 @ 0b0010011
+  when currentlyEnabled(Ext_Zicbop)
+
+mapping prefetch_mnemonic : cbop_zicbop <-> string = {
+  PREFETCH_I <-> "prefetch.i",
+  PREFETCH_R <-> "prefetch.r",
+  PREFETCH_W <-> "prefetch.w"
+}
+
+mapping clause assembly = ZICBOP(cbop, rs1, offset)
+  <-> prefetch_mnemonic(cbop) ^ spc() ^ hex_bits_12(offset) ^ "(" ^ opt_spc() ^ reg_name(rs1) ^ opt_spc() ^ ")"
+
+function clause execute(ZICBOP(cbop, rs1, offset)) = RETIRE_SUCCESS
--- a/model/riscv_zicbop_types.sail
+++ b/model/riscv_zicbop_types.sail
@@ -1,0 +1,9 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+enum cbop_zicbop = {PREFETCH_I, PREFETCH_R, PREFETCH_W}
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -81,6 +81,9 @@
     "Zicbom": {
       "supported": true
     },
+    "Zicbop": {
+      "supported": true
+    },
     "Zicboz": {
       "supported": true
     },
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -7,6 +7,7 @@
 
 - The following extensions have been added:
 
+  - Zicbop
   - Zihintntl
   - Zihintpause
 


model/riscv_insts_zicbop.sail
```diff
@@ -0,0 +1,78 @@
...
+  match ext_data_get_addr(rs1, offset, ac, cache_block_size) {
+    Ext_DataAddr_Error(e) => { ext_handle_data_check_error(e); RETIRE_FAIL },
+    Ext_DataAddr_OK(vaddr) => {
+
+      // "An implementation may opt to cache a copy of the cache block in a cache
+		  //  accessed by an instruction fetch in order to improve memory access latency,
+		  //  but this behavior is not required."
+      //
+      // This implementation do nothing.
+      match translateAddr(vaddr, ac) {
```
Timmmm, line 58: I don't think you can call `translateAddr()` as-is because it might mark pages as accessed or dirty. You'd need to add a `prefetch` flag or something, but I'm not sure there's any point because ultimately you wouldn't do anything with the result.

model/riscv_insts_zicbop.sail
```diff
@@ -0,0 +1,78 @@
...
+        TR_Failure(e, _) => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
```
Timmmm, line 59: This shouldn't cause any exceptions.

model/riscv_insts_zicbop.sail
```diff
@@ -0,0 +1,78 @@
...
+function clause execute(RISCV_ZICBOP(cbop, rs1, offset)) =
+  match cbop {
+    PREFETCH_I if prefetch_i_enabled(cur_privilege) =>
+      process_prefetch(rs1, offset, Execute(Data)),
+    PREFETCH_R if prefetch_r_enabled(cur_privilege) =>
+      process_prefetch(rs1, offset, Read(Data)),
+    PREFETCH_W if prefetch_w_enabled(cur_privilege) =>
+      process_prefetch(rs1, offset, Write(Data)),
+    _ => {
+      handle_illegal();
```
Timmmm, line 75: This isn't right.

Makefile
```diff
@@ -62,6 +62,7 @@ SAIL_DEFAULT_INST += riscv_insts_vext_red.sail
 SAIL_DEFAULT_INST += riscv_insts_vext_fp_red.sail
 SAIL_DEFAULT_INST += riscv_insts_zicbom.sail
 SAIL_DEFAULT_INST += riscv_insts_zicboz.sail
+SAIL_DEFAULT_INST += riscv_insts_zicbop.sail
```
Timmmm, line 65: This is going to have no effect if you add it here because these instructions overlap `ori` which will already have handled them in `riscv_insts_base.sail` above. This file needs to come before that.

Honestly it feels a bit fragile to just rely on the file order and silently break if you get it wrong though. I don't know if @Alasdair has any good ideas to make it robust.

Timmmm (changes requested):
Seems reasonable to me, especially to get the disassembly.

There are a few issues though which I've commented on.

Makefile
Alasdair: I would like to add a check that file order doesn't matter.

You could take all the decode clauses, and automatically change them so `file1.sail` generates `0b001`, `file2.sail` generates `0b010`, `file3.sail` generates `0b100`, then ask the Sail SMT backend if the sum of each files decode clauses has more than a single bit set. Would likely be a bit heavyweight for a check that would run every single invocation though.

Timmmm: I guess the downside is that you'd need to modify the `ORI` decoder so that it excludes these hints somehow.

I guess you could avoid that by explicitly adding an `encdec_hints` or something and checking that first. Kind of rubbish though.

Alternative add some kind of explicit priority for each clause and have the compiler reorder them. Though that kind of thing usually ends up [like CSS z-index](https://blog.frankmtaylor.com/wp-content/uploads/2019/05/z-index.jpg).

model/riscv_insts_zicbop.sail
trdthg: > You'd need to add a prefetch flag or something

`translateAddr` is called from many places, there are quite a few places to modify. If the final conclusion is this, then I will do it

> but I'm not sure there's any point because ultimately you wouldn't do anything with the result.

I think we should at least check if the address is valid.

I briefly looked at what `translateAddr` does. In addition to updating the TLB, it also performs `is_valid_vAddr` and `check_PTE_permission`, which are also places where errors may occur. don't know if it's a good idea to implement something like `check_addr_translate_and_access` for prefetch

Makefile
trdthg: seems to be a discussion about the positional relationship of scattered clauses. I really can't think of a good solution. At least grammatically, I can only think of adding something like `#[index = 1]` before the clause

model/riscv_insts_zicbop.sail
Timmmm: > I think we should at least check if the address is valid.

Why? Nothing depends on the validity of the address.

I would definitely just do nothing rather than updating `translateAddr`.

Makefile
trdthg: Update, I moved the guard to the ORI decode side, so there is no need to modify the order

model/riscv_insts_zicbop.sail
trdthg: removed

Makefile
jordancarlin: At the meeting this week, we discussed holding off on hints and other extensions that mask existing encodings (like Zicfiss) until the new Sail module system is ready (which should be coming up with the next release). The new module system has support for explicitly saying that certain files (extensions in the way we handle things) need to be included before or after others, which can be used ensure the hints get decoded properly without needing to modify the ORI decode clause.

model/riscv_insts_zicbop.sail
```diff
@@ -0,0 +1,57 @@
...
+  let offset_val = X(rs1) + sign_extend(offset);
+  let cache_block_size_exp = plat_cache_block_size_exp();
+  let cache_block_size = 2 ^ cache_block_size_exp;
+
+  // Offset from offset_val to the beginning of the cache block. This is 0 if offset_val
+  // is aligned to the cache block, or negative if offset_val is misaligned.
+  let offset = (offset_val & ~(zero_extend(ones(cache_block_size_exp)))) - offset_val;
+
+  match ext_data_get_addr(rs1, offset, ac, cache_block_size) {
+    Ext_DataAddr_Error(e) => { ext_handle_data_check_error(e); RETIRE_FAIL },
```
Timmmm, line 47: I don't think this is right either. These are ORI hint instructions. Their observable semantics have to be the same as ORI to x0 (i.e. nothing), plus possibly prefetching (which we don't model). They can't raise exceptions.

Basically it should be:

```
function clause execute(RISCV_ZICBOP(cbop, rs1, offset)) = RETIRE_SUCCESS
```

In future it might be worth keeping all this and just ignoring exceptions... but I think let's handle that later. We may actually want address translation - re-reading the spec it isn't clear if prefetch is allowed to update the PTE `A`ccessed bit.

jordancarlin:
Now that the module system is in place, this would be a good one to revisit. It should follow the same approach as #1208.

README.md
```diff
@@ -77,7 +77,7 @@ For booting operating system images, see the information under the
 - Zicsr extension for CSR instructions, v2.0
 - Zicntr and Zihpm extensions for counters, v2.0
 - Zicond extension for integer conditional operations, v1.0
-- Zicbom and Zicboz extensions for cache-block management (Zicbop not currently supported), v1.0
+- Zicbom, Zicboz, Zicbop extensions for cache-block management, v1.0
```
jordancarlin:
```suggestion
- Zicbom, Zicboz, and Zicbop extensions for cache-block management, v1.0
```

model/riscv_extensions.sail
```diff
@@ -80,6 +80,9 @@ function clause hartSupports(Ext_Zicbom) = config extensions.Zicbom.supported
 enum clause extension = Ext_Zicboz
 mapping clause extensionName = Ext_Zicboz <-> "zicboz"
 function clause hartSupports(Ext_Zicboz) = config extensions.Zicboz.supported
+// Cache-Block Prefetch Instructions
+enum clause extension = Ext_Zicbop
```
jordancarlin: Needs to be added to `extensions_ordered_for_isa_string`.

model/riscv_types.sail
```diff
@@ -286,6 +286,7 @@ enum amoop = {AMOSWAP, AMOADD, AMOXOR, AMOAND, AMOOR, /* AMO ops */
 enum csrop = {CSRRW, CSRRS, CSRRC}                    /* CSR ops */
 
 enum cbop_zicbom = {CBO_CLEAN, CBO_FLUSH, CBO_INVAL}  /* Zicbom ops */
+enum cbop_zicbop = {PREFETCH_I, PREFETCH_R, PREFETCH_W}  /* Zicbop ops */
```
jordancarlin: Let's move this to a separate `riscv_zicbop_types.sail` file (like in #1210).

model/riscv.sail_project
```diff
@@ -296,6 +296,12 @@ extensions {
     files riscv_insts_zicboz.sail
   }
 
+  Zicbop {
+    requires riscv
+    before I // PREFETCHH.{I,R,W} instructions override ORI instructions with rd equal to 0b00000
```
arichardson:
```suggestion
    before I // PREFETCH.{I,R,W} instructions override ORI instructions with rd equal to 0b00000
```

model/riscv_extensions.sail
```diff
@@ -394,6 +397,7 @@ let extensions_ordered_for_isa_string = [
   // Zi
   Ext_Zicbom,
   Ext_Zicboz,
+  Ext_Zicbop,
```
jordancarlin, line 408: Should be before Zicboz for canonical order. For consistency, it would be good to reorder it above and in the config files as well.

model/riscv_zicbop_types.sail
```diff
@@ -0,0 +1,9 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+enum cbop_zicbop = {PREFETCH_I, PREFETCH_R, PREFETCH_W}  /* Zicbop ops */
```
jordancarlin: This comment is probably unnecessary at this point. 
```suggestion
enum cbop_zicbop = {PREFETCH_I, PREFETCH_R, PREFETCH_W}
```

jordancarlin (changes requested):

model/riscv.sail_project
```diff
@@ -296,6 +296,18 @@ extensions {
...
   }
 
+  Zicbop {
+    Zicbop_types {
+      before riscv
+      files riscv_zicbop_types.sail
+    }
+    Zicbop_insts {
+      requires riscv, Zicbop_types
+      before I // PREFETCH.{I,R,W} instructions override ORI instructions with rd equal to 0b00000
```
pmundkur:
```suggestion
      before I_insts // PREFETCH.{I,R,W} instructions override ORI instructions with rd equal to 0b00000
```

model/riscv_extensions.sail
```diff
@@ -399,8 +402,32 @@ let extensions_ordered_for_isa_string = [
...
+  Ext_Zbkx,
+  Ext_Zbs,
+
+  // Zc
+  Ext_Zca,
+  Ext_Zcb,
+  Ext_Zcd,
+  Ext_Zcf,
+  Ext_Zcmop,
+
```
pmundkur: As the top comment says, this list should be ordered by category and then alphabetically _within a category_.  The categories themselves are not ordered alphabetically.   This list was ordered correctly the way it was.  See the "Subset Naming Convention" in the unpriv manual.

pmundkur (changes requested):

model/riscv.sail_project
```diff
@@ -338,6 +338,18 @@ extensions {
     }
   }
 
+  Zicbop {
+    Zicbop_types {
+      before riscv
+      files riscv_zicbop_types.sail
+    }
+    Zicbop_insts {
+      before I_insts // PREFETCH.{I,R,W} instructions override ORI instructions with rd equal to 0b00000
```
Timmmm:
```suggestion
      // PREFETCH.{I,R,W} instructions override ORI hints (rd=x0)
      before I_insts
```

(Nitpicking, but IMO end-of-line comments are usually a bad idea.)

model/riscv_insts_zicbop.sail
```diff
@@ -0,0 +1,35 @@
...
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+// Cache Block Operations - Prefetch
+
+function clause currentlyEnabled(Ext_Zicbop) = hartSupports(Ext_Zicbop)
+
+/* ****************************************************************** */
+union clause instruction = RISCV_ZICBOP : (cbop_zicbop, regidx, bits(12))
```
nadime15: As discussed in the meeting today, lets remove the "RISC-V_" prefix

nadime15 (approved):

jordancarlin (approved):

PR #540, opened 2024-09-09
Fix all Sail compilation warnings

Timmmm:
Add `pure`/`impure` annotations and remove some duplicate function declarations.


Timmmm:
@Alasdair I wasn't sure if printing functions should count as pure or impure from Sail's perspective seeing as they're technically impure but the Sail model can't actually observe the impurity.

Timmmm:
Btw I would recommend using squash and merge, especially when there's only one commit in the PR. Otherwise you get a needlessly convoluted commit history.

PR #543, opened 2024-09-09
Make CSR definitions scattered

### LINKED ISSUE #410: CSRs should use scattered functions
The CSR code is a bit odd. For example `readCSR()` uses a `match`:

```
function readCSR csr : csreg -> xlenbits = {
  let res : xlenbits =
  match (csr, sizeof(xlen)) {
    /* machine mode */
    (0xF11,  _) => zero_extend(mvendorid),
...
```

but *also* a scattered function for extensions.

```
...

    _           => /* check extensions */
                   match ext_read_CSR(csr) {
                     Some(res) => res,
                     None()    => { print_bits("unhandled read to CSR ", csr);
                                    zero_extend(0x0) }
                   }
```

This is silly. It should just work like the scattered functions for instructions - everything uses scattered functions, and there's a `riscv_csrs_begin.sail` and `riscv_csrs_end.sail`, and you add new CSRs just by inserting extra files.

Timmmm:
Resolve the strange split between `read_CSR` (which is not `scattered`) and `ext_read_CSR` (which is). Same for `write_CSR` and `is_CSR_defined`.

I changed the return type of `read_CSR` from `option(xlenbits)` to `xlenbits` since the code must already check `is_CSR_defined` before calling `read_CSR`. The only function that returned `None()` was the `seed` CSR in `write_seed_csr`, which actually meant you would get a weird "Unhandled write to CSR" if you wrote to `seed`.

I renamed & reordered the files slightly to make the scattered mapping work, but I haven't moved any of the actual definitions yet. In future we should actually scatter them.

Fixes #410

### DIFF
--- a/model/riscv_insts_zicsr.sail
+++ b/model/riscv_insts_zicsr.sail
@@ -21,203 +21,180 @@ mapping encdec_csrop : csrop <-> bits(2) = {
 mapping clause encdec = CSR(csr, rs1, rd, is_imm, op)
   <-> csr @ rs1 @ bool_bits(is_imm) @ encdec_csrop(op) @ rd @ 0b1110011
 
-function readCSR csr : csreg -> xlenbits = {
-  let res : xlenbits =
-  match (csr, sizeof(xlen)) {
-    /* machine mode */
-    (0xF11,  _) => zero_extend(mvendorid),
-    (0xF12,  _) => marchid,
-    (0xF13,  _) => mimpid,
-    (0xF14,  _) => mhartid,
-    (0xF15,  _) => mconfigptr,
-    (0x300,  _) => mstatus.bits,
-    (0x301,  _) => misa.bits,
-    (0x302,  _) => medeleg.bits,
-    (0x303,  _) => mideleg.bits,
-    (0x304,  _) => mie.bits,
-    (0x305,  _) => get_mtvec(),
-    (0x306,  _) => zero_extend(mcounteren.bits),
-    (0x30A,  _) => menvcfg.bits[sizeof(xlen) - 1 .. 0],
-    (0x310, 32) => mstatush.bits,
-    (0x31A, 32) => menvcfg.bits[63 .. 32],
-    (0x320,  _) => zero_extend(mcountinhibit.bits),
+// TODO: These read/write_CSR definitions should be moved to the files
+// corresponding to their extensions rather than all be here.
 
-    /* Hardware Performance Monitoring event selection */
-    (0b0011001 /* 0x320 */ @ index : bits(5), _) if unsigned(index) >= 3 => read_mhpmevent(hpmidx_from_bits(index)),
+/* machine mode */
+function clause read_CSR(0xF11) = zero_extend(mvendorid)
+function clause read_CSR(0xF12) = marchid
+function clause read_CSR(0xF13) = mimpid
+function clause read_CSR(0xF14) = mhartid
+function clause read_CSR(0xF15) = mconfigptr
+function clause read_CSR(0x300) = mstatus.bits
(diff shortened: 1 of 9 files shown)


arichardson (commented):
Not reviewed this in detail - but the result is much cleaner IMO. Having two methods for the same thing is weird, and I believe extensions can just override the existing definitions by placing their clause first.

model/riscv_insts_zicsr.sail
```diff
@@ -202,14 +179,19 @@ function clause execute CSR(csr, rs1, rd, is_imm, op) = {
...
-    let csr_val = readCSR(csr); /* could have side-effects, so technically shouldn't perform for CSRW[I] with rd == 0 */
+    let csr_val = read_CSR(csr); /* could have side-effects, so technically shouldn't perform for CSRW[I] with rd == 0 */
     if isWrite then {
       let new_val : xlenbits = match op {
         CSRRW => rs1_val,
         CSRRS => csr_val | rs1_val,
         CSRRC => csr_val & ~(rs1_val)
       };
-      writeCSR(csr, new_val)
+      let final_val = write_CSR(csr, new_val);
```
arichardson, line 189: Since there is only one caller this seems fine but I wonder if it makes sense to have a logging variant.

In the future it might be nice to have logging for all implicit CSR reads and writes, but not for this PR.
I was thinking of maybe something like a scattered enum for all the CSRs so you don't have to hardcode the number and then using something like
`let mtvec = CSR(CSR_MTVEC)` could log the implicit read.

arichardson (approved):
Looked at the whole diff now. LGTM. Removing all the `Some(` is quite nice as well.

model/riscv_insts_zicsr.sail
Timmmm: Yeah we have had to put stuff like `trackCSRWrite(csr_name_map("mstatus"))` in a few places in our code. A scattered enum might make more sense. It would also make the `write_CSR`/`read_CSR` a lot code easier to read.

Something like

```
scattered enum CSR

enum clause CSR = MSTATUS
enum clause CSR = MTVEC
...

scattered mapping csr_name_map : CSR <-> string

mapping class csr_name_map = MSTATUS <-> "mstatus"
mapping class csr_name_map = MTVEC <-> "mtvec"
...

scattered mapping csr_index : CSR <-> bits(12)

mapping clause csr_index = MSTATUS <-> 0x123
mapping clause csr_index = MTVEC <-> 0x456
...

function execute(CSR... = {
  let index : bits(12) = ...
  if not(csr_index_matches(index)) ...
  let csr = csr_index(index);
  write_CSR(csr, ...)
```

I'm not sure I quite follow what you mean by "a logging variant"?

arichardson: I meant I am not sure if we want every read/write of a CSR to be emitted in the logfile since that could end up being really verbose.

arichardson: And by logging variant I meant a version of write_CSR/read_CSR that logs the access as opposed to doing it in the caller.

billmcspadden-riscv:
Minstret and timers are overkill.

model/riscv_insts_zicsr.sail
Timmmm: Ah I see, that makes sense.

Timmmm: (But I think we should do it in a future PR.)

arichardson: agreed, no need to do it here.

Timmmm:
I'll merge this in a few days if nobody objects. I think it's uncontroversial.

Alasdair (approved):
Once this is rebased I think it can be merged

PR #544, opened 2024-09-09
Remove the OCaml emulator

### LINKED ISSUE #509: Remove OCaml emulator

Timmmm:
This has the following benefits:

* Reduces the maintenance burden for making changes to the emulator. They don't have to be done in two places and you don't need to learn OCaml.
* Removes the dependency on OCaml and OPAM, which is a weirdly bug-prone tool.
* Removes confusion for new users about which emulator to use.
* Shorter CI time.

<strike>Note, I left the targets to actually build the OCaml code since I believe it is used by TestRig for its random instruction generation.</strike>

Fixes #509


Alasdair:
The Sail to OCaml translation could also use quite a few changes, as it's the first thing I ever implemented when I started working on Sail. It still uses ocamlbuild for example, whereas the OCaml ecosystem has now mostly move to dune.

Removing the use here would actually free me to make some much needed changes.

Timmmm:
@PeterRugg This MR removes OCaml support from the model, which I think would affect TestRig. Is that a deal breaker? What's the status of TestRig?

PeterRugg:
Yeah, we certainly don't compare against the ocaml model, and can't see any reason why we would. Losing the sail-based test generation is sad, but as @jrtc27 says, we never really got chance to investigate using it properly.

arichardson (commented):
I believe the one thing we are losing here is the ability to generate the DTB for the OS boot. I wonder if this should be added to the C simulator first? Otherwise I am more than happy to see the ocaml simulator being removed since it makes it a lot easier to add new extensions with hooks.

arichardson:
> I believe the one thing we are losing here is the ability to generate the DTB for the OS boot. I wonder if this should be added to the C simulator first? Otherwise I am more than happy to see the ocaml simulator being removed since it makes it a lot easier to add new extensions with hooks.

Looks like it's just a simple string template (`let spike_dts isa_spec mmu_spec cpu_hz insns_per_rtc_tick mems`), so should be easy to move over to C.

Timmmm:
I have actually implemented that for our C++ emulator so I'll port it over.

Timmmm:
> I have actually implemented that for our C++ emulator so I'll port it over.

I actually haven't implemented it - I was thinking of something else. It *is* pretty simple; just generate [a string template](https://github.com/riscv/sail-riscv/blob/c61351e16e803d5ae8c14781d13a4bf214eb51d7/ocaml_emulator/platform_impl.ml#L69) and then [run](https://github.com/riscv/sail-riscv/blob/c61351e16e803d5ae8c14781d13a4bf214eb51d7/ocaml_emulator/platform_impl.ml#L166) `dtc` to compile it.

But doing that in C is a right pain, so I'm voting we defer this and add it again if/when somebody asks for it.

If you want to boot Linux you can still manually make a DTB.

arichardson:
> > I have actually implemented that for our C++ emulator so I'll port it over.
> 
> I actually haven't implemented it - I was thinking of something else. It _is_ pretty simple; just generate [a string template](https://github.com/riscv/sail-riscv/blob/c61351e16e803d5ae8c14781d13a4bf214eb51d7/ocaml_emulator/platform_impl.ml#L69) and then [run](https://github.com/riscv/sail-riscv/blob/c61351e16e803d5ae8c14781d13a4bf214eb51d7/ocaml_emulator/platform_impl.ml#L166) `dtc` to compile it.
> 
> But doing that in C is a right pain, so I'm voting we defer this and add it again if/when somebody asks for it.
> 
> If you want to boot Linux you can still manually make a DTB.

Seems fine to me. It would be nice to generate this so that operating systems can parse it and boot, but can be done as a follow-up when needed.

arichardson (approved):

jordancarlin (approved):
LGTM

README.md
```diff
@@ -36,7 +36,7 @@ What is Sail?
 engineer-friendly language, much like earlier vendor pseudocode, but more precisely defined and with tooling to support a wide range of use-cases.
 <p>
 
-Given a Sail specification, the tool can type-check it, generate documentation snippets (in LaTeX or AsciiDoc), generate executable emulators (in C or OCaml), show specification coverage, generate versions of the ISA for relaxed memory model tools, support automated instruction-sequence test generation, generate  theorem-prover definitions for
+Given a Sail specification, the tool can type-check it, generate documentation snippets (in LaTeX or AsciiDoc), generate an executable emulators, show specification coverage, generate versions of the ISA for relaxed memory model tools, support automated instruction-sequence test generation, generate  theorem-prover definitions for
```
arichardson, line 39:
```suggestion
Given a Sail specification, the tool can type-check it, generate documentation snippets (in LaTeX or AsciiDoc), generate executable emulators, show specification coverage, generate versions of the ISA for relaxed memory model tools, support automated instruction-sequence test generation, generate  theorem-prover definitions for
```
We still have rv32 and rv64 emulators so plural seems fine?

README.md
```diff
@@ -139,13 +139,9 @@ mapping clause assembly = SRET() <-> "sret"
...
 
-The model builds OCaml and C emulators that can execute RISC-V ELF
+The model builds a C emulator that can execute RISC-V ELF
 files, and both emulators provide platform support sufficient to boot
-Linux, FreeBSD and seL4.  The OCaml emulator can generate its own
-platform device-tree description, while the C emulator currently
-requires a consistent description to be manually provided.  The C
-emulator can be linked against the Spike emulator for execution with
-per-instruction tandem-verification.
+Linux, FreeBSD and seL4. The C emulator can be linked against the Spike emulator for execution with per-instruction tandem-verification.
```
arichardson, line 144:
```suggestion
Linux, FreeBSD and seL4. The C emulator can be linked against the Spike emulator for execution with per-instruction tandem-verification.
```
Not for this PR, but is the Spike tandem verification actually still functional? If not it should probably be removed.

arichardson (approved):

Timmmm:
I'll merge this in a few days if nobody objects.

PR #547, opened 2024-09-12
Add enable-svinval option

trdthg:
Looks forgotten. Does it?


jrtc27 (approved):
Going back to the PR this was indeed just missed. The OCaml emulator had the option wired up properly though AFAICT.

PR #549, opened 2024-09-15
Add zvbc extention

rez5427:


jrtc27:
Please read https://github.com/riscv/sail-riscv/blob/master/CONTRIBUTING.md, it doesn't look like you have. Also you've checked in the RV32 C emulator binary...

PR #550, opened 2024-09-15
Add zvbc extension

rez5427:


jrtc27:
Why did you close https://github.com/riscv/sail-riscv/pull/549? Please don't create new PRs when the existing one can be updated just fine...

rez5427:
Sorry...

rez5427:
> Why did you close #549? Please don't create new PRs when the existing one can be updated just fine...

Sorry...

rez5427:
I followed the instructions from the [riscv-vector-tests](https://github.com/chipsalliance/riscv-vector-tests) repository and used the nightly version of the riscv-gnu-toolchain to generate code with the ZVBC extension. I ran six tests from the riscv-vector-tests, all of which successfully passed.
test

nadime15:
Hi @rez5427, your PR looks nearly complete. Would you like to rebase your code? Once that’s done, we can review it again.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,163 @@
...
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+enum clause extension = Ext_Zvbc
+function clause extensionEnabled(Ext_Zvbc) = true
+
+mapping vm_name : bits(1) <-> string = {
```
jordancarlin, line 12: I think you should be able to use the existing `maybe_vmask` function instead of defining this new one. This one also doesn't match the assembly from the isa manual.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,163 @@
...
+  0b0 <-> "0",
+  0b1 <-> "1"
+}
+
+union clause ast = VCLMUL_VV : (bits(1), regidx, regidx, regidx)
```
jordancarlin, line 17: These should use `vregidx` for vector register arguments instead of `regidx`.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,163 @@
...
+
+function clause execute (VCLMUL_VV(vm, vs2, vs1, vd)) = {
+  let SEW                     = get_sew();
+  let LMUL_pow                = get_lmul_pow();
+  let num_elem                = get_num_elem(LMUL_pow, SEW);
+
+  let 'n = num_elem;
+  let 'm = SEW;
+
+  var mask    : vector('n, dec, bool)     = undefined;
```
jordancarlin, line 33: I don't think this line is necessary. It should be sufficient to define mask with the let statement below.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,163 @@
...
+  let vm_val  : bits('n)                  = read_vmask(num_elem, vm, 0b00000);
+  let vd_val  : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
```
jordancarlin, line 35:
```suggestion
  let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
```

Sail no longer supports different ordering for individual vectors so this is unnecessary.

jordancarlin (changes requested):
I haven't compared this with the spec yet, but here are some initial comments on the Sail syntax. The extension also needs to be added to the list in the README.

model/riscv_insts_zvbc.sail
jordancarlin: This one is also why CI is failing.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,163 @@
...
+    };
+};
+  vstart = zeros();
+  RETIRE_SUCCESS
+}
+
+union clause ast = VCLMUL_VX : (bits(1), regidx, regidx, regidx)
+
+mapping clause encdec = VCLMUL_VX (vm, vs2, rs1, vd) if extensionEnabled(Ext_Zvbc)
+  <-> 0b001100 @ vm @ encdec_vreg(vs2) @ encdec_reg(rs1) @ 0b110 @ encdec_vreg(vd) @ 0b1010111 if extensionEnabled(Ext_Zvbc)
```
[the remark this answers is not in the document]
rez5427: thanks for noticing

README.md
```diff
@@ -113,6 +113,7 @@ Supported RISC-V ISA features
 - V extension for vector operations, v1.0
 - Zvbb extension for vector basic bit-manipulation, v1.0
 - Zvkb extension for vector cryptography bit-manipulation, v1.0
+- Zvbc extension for vector carryless multiplication, v1.0
```
jordancarlin, line 116: Move this up after Zvbb before Zvkb (and the same in riscv_extensions.sail)

model/riscv_insts_zvbc.sail
jordancarlin: You should be able to delete this function now that you're no longer using it.

rez5427:
I should install sail 0.19

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,158 @@
...
+
+function clause execute (VCLMUL_VV(vm, vs2, vs1, vd)) = {
+  let SEW                     = get_sew();
+  let LMUL_pow                = get_lmul_pow();
+  let num_elem                = get_num_elem(LMUL_pow, SEW);
+
+  let 'n = num_elem;
+  let 'm = SEW;
+
+  let vm_val  : bits('n)                  = read_vmask(num_elem, vm, 0b00000);
```
jordancarlin, line 32:
```suggestion
  let vm_val  : bits('n)                  = read_vmask(num_elem, vm, zvreg);
```

nadime15 (commented):
Okay, I verified the implementation against the vector test suite, and all tests passed successfully.

Besides the formatting, there is just one thing to change for now.

From the spec:
 > These instructions are only defined for SEW=64.

You could add an assert to check for this.

I just noticed that Crypto Vector extensions have dependencies on Zve32x and Zve64x.
In the case of Zvbc, it's Zve64x, but we can ignore this for now.

rez5427:
@nadime15 maybe like this `assert(SEW == 64, "Only SEW=64 supported for zvbc");`

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,157 @@
...
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zvbc) = true
+
+union clause ast = VCLMUL_VV : (bits(1), vregidx, vregidx, vregidx)
+
+mapping clause encdec = VCLMUL_VV (vm, vs2, vs1, vd)
+  <-> 0b001100 @ vm @ encdec_vreg(vs2) @ encdec_vreg(vs1) @ 0b010 @ encdec_vreg(vd) @ 0b1010111 when extensionEnabled(Ext_Zvbc)
```
nadime15, line 14:
```suggestion
mapping clause encdec = VCLMUL_VV (vm, vs2, vs1, vd)
  <-> 0b001100 @ vm @ encdec_vreg(vs2) @ encdec_vreg(vs1) @ 0b010 @ encdec_vreg(vd) @ 0b1010111
  when extensionEnabled(Ext_Zvbc)
```

See #809

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,157 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zvbc) = true
```
nadime15, line 9: Either we remove the switch to enable Zvbc or we change it to: 
```suggestion
function clause extensionEnabled(Ext_Zvbc) = sys_enable_zvbc()
```
Zvbb is also activated by default and does not have a switch. In my opinion, we could remove the switch for Zvbc (`sys_enable_zvbc()`) and keep it set to true.

jordancarlin:
> @nadime15 maybe like this? `assert(SEW == 64, "Only SEW=64 supported for zvbc");`

It looks like right now this assertion is reachable because there is nothing in the model stopping you from executing this instruction with a different SEW. We want to avoid reachable assertions because they cause the model to stop if they fail.

Since the instruction is only defined for SEW = 64, it would be better to encode this directly in the `encdec` clause so the instruction is not even decoded if SEW is not 64. Something like `when extensionEnabled(Ext_Zvbc) & get_sew() = 64`.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,157 @@
...
+  var result = initial_result;
+
+  foreach (i from 0 to (num_elem - 1)) {
+    if mask[i] == bitone then {
+      foreach (n from 0 to (SEW - 1))
+        if vs2_val[i][n] == bitone then result[i] = result[i] ^ (vs1_val[i] << n);
+    };
+  };
+  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
+  vstart = zeros();
```
jordancarlin, line 44:
```suggestion
  set_vstart(zeros());
```

This ensures V context is dirtied at the end of the instruction. See #816.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,157 @@
...
+  foreach (i from 0 to (num_elem - 1)) {
+    if mask[i] == bitone then {
+      foreach (n from 0 to (SEW - 1))
+        if vs2_val[i][n] == bitone then result[i] = result[i] ^ (rs1_val << n);
+      write_vreg(num_elem, SEW, LMUL_pow, vd, result);
+    };
+  };
+  set_vstart(zeros());
+  RETIRE_SUCCESS
+}
```
nadime15, line 83:
```suggestion
  foreach (i from 0 to (num_elem - 1)) {
    if mask[i] == bitone then {
      foreach (n from 0 to (SEW - 1))
        if vs2_val[i][n] == bitone then result[i] = result[i] ^ (rs1_val << n);
    };
  };
  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
  set_vstart(zeros());
  RETIRE_SUCCESS
}
```

Please do this for each instruction.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,157 @@
...
+  let num_elem                = get_num_elem(LMUL_pow, SEW);
+
+  let 'n = num_elem;
+  let 'm = SEW;
+
+  let vm_val  : bits('n)                  = read_vmask(num_elem, vm, zvreg);
+  let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
+
+  let rs1_val : bits('m)                  = get_scalar(rs1, SEW);
+  let vs2_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
```
nadime15, line 143: Please fix the formatting for each function as well.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,157 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zvbc) = sys_enable_zvbc()
```
nadime15, line 9: Would it actually make sense for all crypto vector extension to check if the Vector Extension is enabled? I would say so.

@pmundkur, @jordancarlin

nadime15 (commented):
Almost done!

rez5427:
@nadime15 maybe this is ok

model/riscv_insts_zvbc.sail
jordancarlin: From the spec:
> The Zvknhb and Zvbc Vector Crypto Extensions --and accordingly the composite extensions Zvkn, Zvknc, Zvkng, and Zvksc -- require a Zve64x base, or application ("V") base Vector Extension.
> 
> All of the other Vector Crypto Extensions can be built on any embedded (Zve*) or application ("V") base Vector Extension.

Right now we only have the `V` extension in the model. We need to go through and fix support for all of the other `Zve*` vector extensions (we have the bulk of the code for all of them already, we just need to fix what is enabled when). #785 does a little bit of this. For now, it seems reasonable to have this require `V`, but we'll need to refactor all of the vector extensionEnabled logic soon, so I don't think it's a big deal.

nadime15 (approved):
Awesome, looks good!

jordancarlin (approved):
LGTM now!

pmundkur (approved):

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,155 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zvbc) = sys_enable_zvbc() & extensionEnabled(Ext_V)
```
nadime15, line 9: Just one last question, do we want to keep this switch here? I think we could remove it.
@pmundkur, @jordancarlin 
```suggestion
function clause extensionEnabled(Ext_Zvbc) = extensionEnabled(Ext_V)
```

jordancarlin: I’d say with #758 almost ready to merge, we should avoid adding new command line switches that will be replaced by configuration flags. Either keep `sys_enable_zvbc()` but hardcode it to true or drop it entirely for now.

rez5427: should I remove all `sys_enable_zvbc` related code?

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,155 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zvbc) = sys_enable_zvbc() | extensionEnabled(Ext_V)
```
jordancarlin, line 9:
```suggestion
function clause extensionEnabled(Ext_Zvbc) = sys_enable_zvbc() & extensionEnabled(Ext_V)
```

c_emulator/riscv_sim.cpp
```diff
@@ -144,6 +145,7 @@ static struct option options[] = {
     {"enable-zicbom",               no_argument,       0, OPT_ENABLE_ZICBOM       },
     {"enable-zicboz",               no_argument,       0, OPT_ENABLE_ZICBOZ       },
     {"enable-zvkb",                 no_argument,       0, OPT_ENABLE_ZVKB         },
+    {"enable-zvbc",                 no_argument,       0, OPT_ENABLE_ZVBC         },
```
jordancarlin, line 148: All of the changes in this file should be removed.

jordancarlin (changes requested):

pmundkur (approved):

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,155 @@
...
+  let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
+  let vs1_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs1);
+  let vs2_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+
+  let (initial_result, mask) = init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val);
+  var result = initial_result;
+
+  foreach (i from 0 to (num_elem - 1)) {
+    if mask[i] == bitone then {
+      foreach (n from 0 to (SEW - 1))
```
Timmmm, line 111: This should be `from 1 to`. Technically doesn't make a difference since 0 will always result in `result[i] = result[i] ^ 0` but they used 1 in the original pseudocode so it probably makes sense here too.

Ditto for line 148.

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,155 @@
...
+mapping clause assembly = VCLMULH_VV (vm, vs2, vs1, vd)
+  <-> "vclmulh.vv" ^ spc() ^ vreg_name(vd) ^ sep() ^ vreg_name(vs2) ^ sep() ^ vreg_name(vs1) ^ maybe_vmask(vm)
+
+function clause execute (VCLMULH_VV(vm, vs2, vs1, vd)) = {
+  let SEW       = get_sew();
+  let LMUL_pow  = get_lmul_pow();
+  let num_elem  = get_num_elem(LMUL_pow, SEW);
+
+  let 'n = num_elem;
+  let 'm = SEW;
```
Timmmm, line 99: Maybe just delete these and the explicit type annotations? Or if we keep the type annotations then don't have two variables for the same thing. You can do

```
let 'n = get_num_elem(...
```

model/riscv_insts_zvbc.sail
```diff
@@ -0,0 +1,155 @@
...
+
+  let rs1_val : bits('m) = get_scalar(rs1, SEW);
+  let vs2_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+
+  let (initial_result, mask) = init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val);
+  var result = initial_result;
+
+  foreach (i from 0 to (num_elem - 1)) {
+    if mask[i] == bitone then {
+      foreach (n from 0 to (SEW - 1))
```
Timmmm, line 75: I feel like it would make sense to extract this into `clmul[h]` functions like in the original psuedocode. Easier to test and then it can also be shared with the scalar code.

We could maybe do that in a future PR though.

Timmmm (commented):
Some minor comments, otherwise LGTM.

nibrunie:
> Okay, I verified the implementation against the vector test suite, and all tests passed successfully.
> 
> Besides the formatting, there is just one thing to change for now.
> 
> From the spec:
> 
> > These instructions are only defined for SEW=64.
> 
> You could add an assert to check for this.
> 
> I just noticed that Crypto Vector extensions have dependencies on Zve32x and Zve64x (which are not implemented yet). In the case of Zvbc, it's Zve64x, but we can ignore this for now.

Shouldn't a comment be added in the code, maybe around
```
function clause extensionEnabled(Ext_Zvbc) = sys_enable_zvbc() & extensionEnabled(Ext_V)
```
to highlight that this will need to be fixed once `Ext_Zve*` are declared to make sure the proper dependency is checked here ?

model/riscv_insts_zvbc.sail
Timmmm: Will do this in a future PR so the scalar code can use it too.

model/riscv_insts_zvbc.sail
Timmmm: Apparently all the other vector code is in this style too so leave it for now and we can update them all at once.

model/riscv_insts_zvbc.sail
Timmmm: Just discussed this in the meeting; good to go after this is done.

Timmmm:
Ah looks like there's a small conflict that needs to be resolved...

PR #563, opened 2024-09-26
Add variable alias for sizeof(xlen)

### LINKED ISSUE #561: Replace sizeof(xlen) with xlen
Currently the model has a load of `sizeof(xlen)` everywhere because `xlen` is a type, not a variable:

```
type xlen       : Int = xlen_bytes * 8
```

You need `sizeof()` to convert the type into a variable. However Sail types and variables are in different namespaces, so you can have a type that is the same name as a variable, like this:

```
type xlen       : Int = xlen_bytes * 8
let xlen = sizeof(xlen)
```

The CHERI model uses this extensively - [see here for example](https://github.com/CTSRD-CHERI/sail-cheri-riscv/blob/c93d5eff4132fd52ddad32020cedae93573bfa29/src/cheri_prelude_128.sail#L64).

I think we should do the same for `xlen` in this model. It makes the code more readable and more like pseudo-code, and I can't see any downsides.

Any thoughts?

Timmmm:
Add variable aliases for `sizeof(xlen)`, `sizeof(xlen_bytes)` and `sizeof(flen)`, and remove unused `flen_bytes`.

This makes the code a bit simpler and more readable.

Fixes #561

### DIFF
--- a/model/main.sail
+++ b/model/main.sail
@@ -15,7 +15,7 @@ $ifdef SYMBOLIC
 
 $include <elf.sail>
 
-function get_entry_point() = to_bits(sizeof(xlen), elf_entry())
+function get_entry_point() = to_bits(xlen, elf_entry())
 
 $else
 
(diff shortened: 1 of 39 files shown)


arichardson (approved):
I always found it odd that we had to use sizeof(*len), this is much nicer.

jordancarlin (approved):
Seems like a nice improvement

Issue #548, opened 2024-09-13
Duplicate vector CSR code


Timmmm:
The code for reading/writing vector CSRs is duplicated [here](https://github.com/riscv/sail-riscv/blob/31f809346e1020d191b46d519ad6435e2428be39/model/riscv_insts_zicsr.sail#L66) and [here](https://github.com/riscv/sail-riscv/blob/31f809346e1020d191b46d519ad6435e2428be39/model/riscv_vext_control.sail#L18). The latter is not used because the former takes priority.

The code is not actually the same, so something is definitely not right.

trdthg:
(One more point, The `ext_read_CSR` functions are also duplicated in the latter file
https://github.com/riscv/sail-riscv/blob/31f809346e1020d191b46d519ad6435e2428be39/model/riscv_vext_control.sail#L18-L24

PR #564, opened 2024-09-28
replace vxrm with vcsr[vxrm]

rez5427:
When I run the vasub test file from the riscv-vector-test suite, I encountered a discrepancy between the outputs of Sail-RISCV and Spike, as shown in the image (highlighted in red). The issue stems from the function get_fixed_rounding_incr, which is incorrectly using the vxrm[1..0] bits instead of vcsr[vxrm]. The vxrm register is not being set by a CSR instruction; it is initialized to zero and remains zero throughout the execution, causing this mismatch.

image


Timmmm:
Presumably this issue also exists for `vxsat`? I think we should also fix the duplicate read_csr and write_csr clauses at the same time.

rez5427:
Is it appropriate to remove the registers vxrm and vxset in this context?

Alasdair:
I don't think vxrm and vxsat should be actual Sail registers no, I think they are just parts of vcsr that should be accessible via the read_csr and write_csr functions.

model/riscv_sys_control.sail
```diff
@@ -557,10 +557,6 @@ function init_sys() -> unit = {
    * See riscv_vlen.sail for details.
    */
   vstart             = zero_extend(0b0);
-  vxsat              = 0b0;
-  vxrm               = 0b00;
-  vcsr[vxrm]       = vxrm;
-  vcsr[vxsat]      = vxsat;
```
Alasdair, line 563: This might want to be `vcsr[vxrm] = 0b00` and `vcsr[vxsat] = 0b0` rather than removing both.

Alasdair (approved):
Note: we should make sure to squash these commits before merging.

model/riscv_insts_zicsr.sail
```diff
@@ -71,8 +71,8 @@ function clause read_CSR(0b1011100 /* 0xB80 */ @ index : bits(5) if sizeof(xlen)
 
 /* vector */
 function clause read_CSR(0x008) = zero_extend(vstart)
-function clause read_CSR(0x009) = zero_extend(vxsat)
-function clause read_CSR(0x00A) = zero_extend(vxrm)
+function clause read_CSR(0x009) = zero_extend(vcsr[vxsat])
```
Timmmm, line 74: I think we should probably fix #548 in this MR though. Move all the vector code from this file to `riscv_vext_control.sail` and then delete the duplicates (there are 3 copies for some of them!).

Note the existing code in that file seems to be correct - it includes this fix and also dirties the status flag which the code here doesn't do.

Timmmm: Let me know if you can't be bothered though - I can have a go.

rez5427: Sorry for the late reply. Is this okay?

Timmmm: Yep LGTM, thanks!

Timmmm (approved):

PR #568, opened 2024-09-29
replace fsplit and fmake with mapping fstruct

KotorinMinami:
As described in the TODO in fext and dext, I attempted to implement a mapping using floating-point number structures. Does it meet the stated requirements?


jordancarlin:
With the work going on upstream in [sail](https://github.com/rems-project/sail) itself to support floats, I'm not sure that this makes sense to update. The new version has float specific structs already that support extracting and creating the different parts, similar to what this PR is doing. See the PRs that have names beginning with "Float: " in the sail repo.

PR #570, opened 2024-10-02
Add Sstc extension

Timmmm:
This adds the stimecmp[h] CSRs. It is enabled by default.

The code is slightly unfortunately structured to handle CSRs that don't fit the standard permission checks. I added a TODO to improve it.


Timmmm:
This is a rebase of #395 by @ved-rivos. The model structure has changed a bit so it required fairly extensive changes. I also added a command line flag.

model/riscv_sys_regs.sail
```diff
@@ -328,11 +331,56 @@ register mip     : Minterrupts /* Pending */
...
+  // Fence of I/O implies Memory
+  FIOM   : 0,
+}
+
+register menvcfg : MEnvcfg
+register senvcfg : SEnvcfg
+
+/* Supervisor timecmp */
+enum clause extension = Ext_Sstc
+function clause extensionEnabled(Ext_Sstc) = sys_enable_sstc() & menvcfg[STCE] == 0b1
```
Timmmm, line 371: Hmm immediately after sending it I realised this isn't quite right because `STCE` only applies to modes less than Machine. I'll fix it.

Alasdair (approved):

Alasdair:
Yes, I think having a `is_csr_accessible` function might be a good idea

Timmmm:
Rebased to handle the CSR rejigging. I had to move some code around to make the dependencies work. I also removed the `platform_wfi()` speed hack since it would no longer be correct with Sstc. I doubt it make much difference to speed in practice.

We've been using this code for a while now with no issues found. It seems like @jscheid-ventana was the original author of the spec - if you could take a look that would be great!

Otherwise I think we should just merge this.

model/riscv_sstc.sail
```diff
@@ -0,0 +1,20 @@
...
+/*=======================================================================================*/
+
+/* Sstc - supervisor time compare */
+mapping clause csr_name_map = 0x14D  <-> "stimecmp"
+mapping clause csr_name_map = 0x15D  <-> "stimecmph"
+
+function clause is_CSR_defined(0x14D) = extensionEnabled(Ext_S) & extensionEnabled(Ext_Sstc)
+function clause is_CSR_defined(0x15D) = extensionEnabled(Ext_S) & extensionEnabled(Ext_Sstc) & xlen == 32
+
+function clause read_CSR(0x14D) = stimecmp[sizeof(xlen) - 1 .. 0]
```
Alasdair, line 16: Now we have [this commit](https://github.com/riscv/sail-riscv/commit/ba5f6d0ee4cb7168499e5ddb739f2685ad3469fd), I think this can just be `xlen - 1`?

model/riscv_sys_control.sail
```diff
@@ -40,6 +40,15 @@ function check_Counteren(csr : csreg, p : Privilege) -> bool = {
   feature_enabled_for_priv(p, mcounteren.bits[index], scounteren.bits[index])
 }
 
+// Return true if the stimecmp[h] CSR is accessible OR the CSR is not stimecmp[h].
+function check_Stimecmp(csr : csreg, p : Privilege) -> bool = {
+  // Check if it is not stimecmp.
+  if csr != 0x14D & csr != 0x15D
+  then return true;
```
Alasdair, line 47: I think this would look better with the then on the preceding line like:
```
  if csr != 0x14D & csr != 0x15D then {
    return true
  };
```
or maybe with the whole if as a single line.

Alasdair:
I left some minor comments, but overall I also think this can just be merged.

model/riscv_sys_control.sail
```diff
@@ -61,8 +70,12 @@ function check_seed_CSR (csr : csreg, p : Privilege, isWrite : bool) -> bool = {
 function check_CSR(csr : csreg, p : Privilege, isWrite : bool) -> bool =
```
arichardson, line 70: Not for this PR, but wouldn't this be cleaner as a scattered function?

arichardson (approved):
LGTM with @Alasdair's comment addressed

model/riscv_sstc.sail
Timmmm: Good spot - updated.

model/riscv_sys_control.sail
Timmmm: I never know how to format these - where's that autoformatter? :-D

I changed it to a single line.

model/riscv_sys_control.sail
Timmmm: I think this one shouldn't be a scattered function because `check_CSR_access()` is always the same, but the other functions should be moved into `is_CSR_defined()`, something like this:

```
function check_CSR(csr : csreg, p : Privilege, isWrite : bool) -> bool =
   check_CSR_access(csrAccess(csr), csrPriv(csr), p, isWrite) & is_CSR_defined(csr, p, isWrite)
```

Or we could potentially leave `isWrite` just for the `seed` CSR. It's the only one that needs it and unfortunately it is actually specified that way:

> Attempts to access the seed CSR using a read-only CSR-access instruction (CSRRS/CSRRC with rs1=x0 or CSRRSI/CSRRCI with uimm=0) raise an illegal instruction exception

I'll make a PR to clean this up after this one.

PR #577, opened 2024-10-05
fix some typo in v ext

rez5427:


Timmmm:
The last one looks obvious but can you edit the PR message to explain the `mask` changes for those not familiar with the vector extension? Ideally with a link to the relevant bit of the spec (I have linkable versions [here](https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#vector)).

rez5427:
When running the test vlseg3e8ff_v-0 from the riscv-vector-tests suite, it initially fails. Upon investigation, I discovered that the issue arises from the handling of the vm_val in the VLSEGTYPE function in the Sail RISC-V model ([riscv_insts_vext_mem.sail](https://github.com/riscv/sail-riscv/blob/master/model/riscv_insts_vext_mem.sail)).

vm_val is always set to 0x0000, which leads to the incorrect behavior during the execution of vector load instructions,
`  let (result, mask) = init_masked_result(num_elem, nf * load_width_bytes * 8, EMUL_pow, vd_seg, vm_val);`
vm_val is only a temperary value.
I think is a typo.
And the [vector load and store](https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#sec-vector-memory)

rez5427:
This one is for pass the vsuxseg4ei8_v-0 test.
image

rez5427:
The address is from 40->44->4c->E0.
I used print_reg to display the relevant value and found that a sign extension issue is causing the value to become negative.

rez5427:
I checked Spike and found that it uses uint8

Timmmm:
We have the same fix for `mask` so I'm pretty sure that one is correct (sorry we've got a load of vector fixes that are not upstreamed yet).

I looked into the signed/unsigned one. Unfortunately the spec doesn't say if should be signed or unsigned - just calls it a "byte offset". However it gives no indication it should be signed and that would seem like a weird default. Also if Spike uses unsigned I think that's good enough. I'll double check with out vector guys.

Should it also be fixed in `process_vlsseg` and `process_vlsseg`? (Look for `signed(get_scalar(`.)

Timmmm:
Ugh these names :-D just search for `signed(`.

Timmmm:
Also if you want to separate the `signed` changes out into a separate PR I think we can merge the `mask` and assembly typo straight away.

Timmmm:
My colleague confirmed `unsigned` is correct. It's [in the spec](https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#_vector_loadstore_addressing_modes) but definitely not obvious!

> If the vector offset elements are narrower than XLEN, they are zero-extended to XLEN before adding to the base effective address.

rez5427:
I'll separate it

rez5427:
> My colleague confirmed `unsigned` is correct. It's [in the spec](https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#_vector_loadstore_addressing_modes) but definitely not obvious!
> 
> > If the vector offset elements are narrower than XLEN, they are zero-extended to XLEN before adding to the base effective address.

New thing to learn today

Timmmm (approved):

PR #581, opened 2024-10-08
Change sign to unsign in v extension

rez5427:


rez5427:
The address is from 40->44->4c->E0.
I used print_reg to display the relevant value and found that a sign extension issue is causing the value to become negative.
image

Timmmm (approved):

PR #585, opened 2024-10-09
Saturation remove set 0

rez5427:
I ran the test for VNClip and noticed that the saturation function includes logic that sets vxsat to zero. This caused the test to fail. After reviewing the Spike implementation, I found that it does not set vxsat to zero. The [spec](https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#_vector_single_width_fractional_multiply_with_rounding_and_saturation) only mentions setting it to one, without specifying any case where it should be reset to zero.

> then saturates the result to fit into SEW bits. If the result causes saturation, the vxsat bit is set.

In vector instructions, if a situation causes saturation midway, vxsat is set to one but is later reset to zero. As a result, the information that vxsat was once set to one is discarded, as shown in the case in the picture.

image

Here is the spike.
image


allenjbaum:
In IEEE FP there are exception bits that can be set by HW, but are only
cleared by SW.
One of them is an underflow exception bit (among others), indicating a
denormalized result.
It is possible for an FP op to yield a normalized result, AND to set the
underflow exception bit.
This can happen if an operation yields an denormalized result, but when
rounded, becomes normalized again.
It sounds like the P extension is doing something similar.

Timmmm:
I confirmed we have the same fix in our fork (sorry... I'll try to make a list of fixes at some point at least).

Issue #566, opened 2024-09-29
Make VLEN and ELEN configurable and remove registers


Timmmm:
VLEN and ELEN are implementation defined configuration parameters for the vector extension. Currently the implementation stores them in registers:

```
register elen : bits(1)

val get_elen_pow : unit -> {5, 6}

function get_elen_pow() = match elen {
    0b0 => 5,
    0b1 => 6
}

register vlen : bits(4)

val get_vlen_pow : unit -> {5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16}

function get_vlen_pow() = match vlen {
    0b0000 => 5,
    0b0001 => 6,
    0b0010 => 7,
    0b0011 => 8,
    0b0100 => 9,
    0b0101 => 10,
    0b0110 => 11,
    0b0111 => 12,
    0b1000 => 13,
    0b1001 => 14,
    0b1010 => 15,
    _      => 16
}
```

And then initialises them in the reset function:

```
  elen               = 0b1; /* ELEN=64 as the common case */
  vlen               = 0b0100; /* VLEN=512 as a default value */
```

This is weird in many ways - they aren't mutable so they shouldn't be registers. They also have no binary encoding. I think that's just because the author didn't realise Sail supports integers or something?

We should change these to functions like this (limits defined [here](https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#_implementation_defined_constant_parameters)):

```
val sys_vector_vlen_exp = pure "sys_vector_vlen_exp" : unit -> range(3, 16)
val sys_vector_elen_exp = pure "sys_vector_elen_exp" : unit -> range(3, 16)

function get_vlen() -> ... = 2 ^ sys_vector_vlen_exp()
function get_elen() -> ... = 2 ^ sys_vector_vlen_exp()
```

The actual values can be hard-coded initially.

See also #565

PR #590, opened 2024-10-12
Make VLEN and ELEN configurable and remove registers

rez5427:
Should sail-riscv use a "--isa=zvl512b_zve64d" like option in spike to implement configurable vlen and elen.
If it should I will do it.

### DIFF
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -86,6 +86,7 @@ uint64_t sys_vector_vlen_exp(unit)
 {
   return rv_vector_vlen_exp;
 }
+
 uint64_t sys_vector_elen_exp(unit)
 {
   return rv_vector_elen_exp;
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -82,6 +82,15 @@ uint64_t sys_pmp_grain(unit u)
   return rv_pmp_grain;
 }
 
+uint64_t sys_vector_vlen_exp(unit)
+{
+  return rv_vector_vlen_exp;
+}
+uint64_t sys_vector_elen_exp(unit)
+{
+  return rv_vector_elen_exp;
+}
+
 bool sys_enable_writable_misa(unit u)
 {
   return rv_enable_writable_misa;
--- a/c_emulator/riscv_platform_impl.c
+++ b/c_emulator/riscv_platform_impl.c
@@ -6,6 +6,9 @@
 uint64_t rv_pmp_count = 0;
 uint64_t rv_pmp_grain = 0;
 
+uint64_t rv_vector_vlen_exp = 0x4;
+uint64_t rv_vector_elen_exp = 0x1;
+
 bool rv_enable_svinval = false;
 bool rv_enable_zcb = false;
 bool rv_enable_zfinx = false;
--- a/c_emulator/riscv_platform_impl.c
+++ b/c_emulator/riscv_platform_impl.c
@@ -6,8 +6,8 @@
 uint64_t rv_pmp_count = 0;
 uint64_t rv_pmp_grain = 0;
 
-uint64_t rv_vector_vlen_exp = 0x4;
-uint64_t rv_vector_elen_exp = 0x1;
+uint64_t rv_vector_vlen_exp = 0x7;
+uint64_t rv_vector_elen_exp = 0x6;
 
 bool rv_enable_svinval = false;
 bool rv_enable_zcb = false;
--- a/c_emulator/riscv_platform_impl.c
+++ b/c_emulator/riscv_platform_impl.c
@@ -6,7 +6,7 @@
 uint64_t rv_pmp_count = 0;
 uint64_t rv_pmp_grain = 0;
 
-uint64_t rv_vector_vlen_exp = 0x7;
+uint64_t rv_vector_vlen_exp = 0x9;
 uint64_t rv_vector_elen_exp = 0x6;
 
 bool rv_enable_svinval = false;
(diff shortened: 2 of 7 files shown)


rez5427:
https://github.com/riscv/sail-riscv/issues/566

model/riscv_vlen.sail
```diff
@@ -8,9 +8,13 @@
 
 register elen : bits(1)
```
Timmmm, line 9: Can remove this now, and `register vlen`.

model/riscv_vlen.sail
```diff
@@ -21,9 +25,13 @@ function get_elen_pow() = match elen {
 
 register vlen : bits(4)
 
+val sys_vector_vlen_exp = pure "sys_vector_vlen_exp" : unit -> bits(4)
```
Timmmm, line 28: There's no reason to use `bits(4)` here. This can simply return `range(3, 16)` (I'm pretty sure the minimum is 3 not 5?).

I think the elen exponent should also be `range(3, 16)`.

model/riscv_vlen.sail
```diff
@@ -21,9 +25,13 @@ function get_elen_pow() = match elen {
...
+
+function get_vlen() -> bits(4) = sys_vector_vlen_exp()
+
 val get_vlen_pow : unit -> {5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16}
 
-function get_vlen_pow() = match vlen {
+function get_vlen_pow() = match get_vlen() {
```
Timmmm, line 34: You'll need to update the code that actually uses `vlen` to call this function (similarly for `elen`). Easy to find once you remove the registers.

Timmmm (changes requested):

rez5427:
@Timmmm Like this?

Timmmm:
Not quite. I was thinking this (feel free to copy/paste):

```
/*  This Sail RISC-V architecture model, comprising all files and                        */
/*  directories except where otherwise noted is subject the BSD                          */
/*  two-clause license in the LICENSE file.                                              */
/*                                                                                       */

val sys_vector_elen_exp = pure "sys_vector_elen_exp" : unit -> range(3, 16)
function get_elen_pow() -> range(3, 16) = sys_vector_elen_exp()

/* Note: ELEN=32 requires a different encoding of the CSR vtype.
 * The current version of vtype implementation corresponds to the ELEN=64 configuration.
 * TODO: the configurarion of ELEN and its corresponding vtype implementations.
 */

/* Note: Currently elen >= vlen is not checked. */

val sys_vector_vlen_exp = pure "sys_vector_vlen_exp" : unit -> range(3, 16)
function get_vlen_pow() -> range(3, 16) = sys_vector_vlen_exp()

type vlenmax : Int = 128
```

Apart from that it looks good.

c_emulator/riscv_platform.c
```diff
@@ -82,6 +82,15 @@ uint64_t sys_pmp_grain(unit u)
   return rv_pmp_grain;
 }
 
+uint64_t sys_vector_vlen_exp(unit)
+{
+  return rv_vector_vlen_exp;
+}
+uint64_t sys_vector_elen_exp(unit)
```
Timmmm, line 89: Slightly ocd (we should get `clang-format`) but can you add a blank line between these functions?

c_emulator/riscv_platform_impl.c
```diff
@@ -6,6 +6,9 @@
 uint64_t rv_pmp_count = 0;
 uint64_t rv_pmp_grain = 0;
 
+uint64_t rv_vector_vlen_exp = 0x4;
+uint64_t rv_vector_elen_exp = 0x1;
```
Timmmm, line 10: Oh also this should be:

```
uint64_t rv_vector_vlen_exp = 7;
uint64_t rv_vector_elen_exp = 6;
```

Timmmm (changes requested):

model/riscv_vlen.sail
```diff
@@ -6,39 +6,20 @@
...
-    0b1000 => 13,
-    0b1001 => 14,
-    0b1010 => 15,
-    _      => 16
-}
-
-type vlenmax : Int = 65536
+type vlenmax : Int = 128
 
 /* Note: At present, the values of elen and vlen need to be manually speficied
```
Timmmm, line 24: This comment needs to be updated - see my earlier suggestion.

model/riscv_vlen.sail
```diff
@@ -6,39 +6,20 @@
...
-    0b0110 => 11,
-    0b0111 => 12,
-    0b1000 => 13,
-    0b1001 => 14,
-    0b1010 => 15,
-    _      => 16
-}
-
-type vlenmax : Int = 65536
+type vlenmax : Int = 128
```
Timmmm, line 22: This shouldn't change, unless I'm missing something?

Timmmm (changes requested):

model/riscv_vlen.sail
Timmmm: Oh sorry I didn't make an earlier suggestion. I think just delete this entire comment.

Timmmm:
LGTM once the comment is removed.

c_emulator/riscv_platform_impl.c
```diff
@@ -6,6 +6,9 @@
 uint64_t rv_pmp_count = 0;
 uint64_t rv_pmp_grain = 0;
 
+uint64_t rv_vector_vlen_exp = 0x7;
```
Timmmm, line 9: Sorry I think I told you the wrong thing before; to match the existing value this should be 9.

Timmmm (approved):

arichardson (approved):

Timmmm:
Thanks!

PR #594, opened 2024-10-15
Fix RV64F compilation, simplify fmv implementation, and make nan boxing functions generic

### LINKED ISSUE #556: Doesn't build with FLEN != XLEN
When FLEN != XLEN there are some compilation issues in the `fmv` D instructions. Current code is:

```
function clause execute (F_UN_TYPE_D(rs1, rd, FMV_X_D)) = {
  assert(sizeof(xlen) >= 64);
  let rs1_val_D            = F(rs1)[63..0];
  let rd_val_X  : xlenbits = sign_extend(rs1_val_D);
  X(rd) = rd_val_X;
  RETIRE_SUCCESS
}
```

On RV64F the `sizeof(xlen) >= 64)` assertion passes, but then `F(rs1)[63..0]` can't compile because `F(rs1)` is only 32 bits. The fix is to add an assertion for `flen` too, since this function is only available when xlen==64 & flen==64.

Also the functions have a crazy level of intermediate variables and explicit type annotation. I'm using these much simpler versions:

```
function clause execute (F_UN_TYPE_S(rs1, rd, FMV_X_W)) = {
  // A narrower n-bit transfer out of the floating-point registers will
  // transfer the lower n bits of the register ignoring the upper FLEN-n bits.
  X(rd) = sign_extend(F(rs1)[31 .. 0]);
  RETIRE_SUCCESS
}

function clause execute (F_UN_TYPE_S(rs1, rd, FMV_W_X)) = {
  // A narrower n-bit transfer, n<FLEN, into the f registers will create a valid
  // NaN-boxed value.
  F(rd) = nan_box(X(rs1)[31..0]);
  RETIRE_SUCCESS
}

function clause execute (F_UN_TYPE_D(rs1, rd, FMV_X_D)) = {
  assert(sizeof(xlen) >= 64 & sizeof(flen) >= 64);
  X(rd) = F(rs1);
  RETIRE_SUCCESS
}

function clause execute (F_UN_TYPE_D(rs1, rd, FMV_D_X)) = {
  assert(sizeof(xlen) >= 64 & sizeof(flen) >= 64);
  F(rd) = X(rs1);
  RETIRE_SUCCESS
}
```

The spec for these for XLEN != FLEN is *really* hard to understand so it is nice to have concise Sail definitions.

Timmmm:
Simplify the implementations with fewer intermediate variables, and fix compilation of RV64F.

I also added relevant quote from the spec because the spec for these instructions is very confusing. This is a prime candidate for getting Sail code into the spec.

Fixes #556

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -15,55 +15,39 @@
 
 /* **************************************************************** */
 /* NaN boxing/unboxing.                                             */
-/* When 16-bit floats (half-precision) are stored in 32/64-bit regs  */
-/* they must be 'NaN boxed'.                                         */
-/* When 16-bit floats (half-precision) are read from 32/64-bit regs  */
-/* they must be 'NaN unboxed'.                                       */
-/* When 32-bit floats (single-precision) are stored in 64-bit regs  */
-/* they must be 'NaN boxed' (upper 32b all ones).                   */
-/* When 32-bit floats (single-precision) are read from 64-bit regs  */
-/* they must be 'NaN unboxed'.                                      */
 
-function canonical_NaN_H() -> bits(16) = 0x_7e00
-function canonical_NaN_S() -> bits(32) = 0x_7fc0_0000
-function canonical_NaN_D() -> bits(64) = 0x_7ff8_0000_0000_0000
+// Canonical NaNs are used when an invalid boxed value is unboxed.
+val canonical_NaN : forall 'n, 'n in {16, 32, 64, 128} . (implicit('n)) -> bits('n)
+function canonical_NaN(n) =
+  match 'n {
+    //     sign  exponent    significand
+    16  => 0b0 @ ones(5)  @ 0b1 @ zeros(9),
+    32  => 0b0 @ ones(8)  @ 0b1 @ zeros(22),
+    64  => 0b0 @ ones(11) @ 0b1 @ zeros(51),
+    128 => 0b0 @ ones(15) @ 0b1 @ zeros(111),
+  }
 
-val      nan_box_H : bits(16) -> flenbits
-function nan_box_H   val_16b =
-  if (flen == 32)
-  then 0x_FFFF @ val_16b
-  else 0x_FFFF_FFFF_FFFF @ val_16b
+// For backwards compatibility with code that existed before the
+// generic version above.
+function canonical_NaN_H() -> bits(16)  = canonical_NaN(16)
+function canonical_NaN_S() -> bits(32)  = canonical_NaN(32)
+function canonical_NaN_D() -> bits(64)  = canonical_NaN(64)
+function canonical_NaN_Q() -> bits(128) = canonical_NaN(128)
 
-val      nan_unbox_H : flenbits -> bits(16)
-function nan_unbox_H   regval =
-  if (flen == 32)
-  then if regval [31..16] == 0x_FFFF
-       then regval [15..0]
-       else canonical_NaN_H()
-  else if regval [63..16] == 0x_FFFF_FFFF_FFFF
-       then regval [15..0]
-       else canonical_NaN_H()
+// n-bit writes to a >n-bit floating point register "box" the value
+// by prepending 1s, which turn it into a qNaN.
+val nan_box : forall 'n 'm, 'n <= 'm . (implicit('m), bits('n)) -> bits('m)
+function nan_box(n, x) = ones('m - 'n) @ x
 
-val nan_box_S : bits(32) -> flenbits
-function nan_box_S val_32b = {
-  assert(sys_enable_fdext());
-  if (flen == 32)
-  then val_32b
-  else 0x_FFFF_FFFF @ val_32b
-}
-
-val nan_unbox_S : flenbits -> bits(32)
-function nan_unbox_S regval = {
-  assert(sys_enable_fdext());
-  if (flen == 32)
-  then regval
-  else if regval [63..32] == 0x_FFFF_FFFF
-       then regval [31..0]
-       else canonical_NaN_S()
-}
-
-overload nan_box = { nan_box_H, nan_box_S }
-overload nan_unbox = { nan_unbox_H, nan_unbox_S }
+// Except for floating-point transfer instructions (FSn and FMV.n.X),
+// n-bit reads of a >n-bit floating point register "unbox" the value stored
+// in the register. If the register does not contain a valid boxed value
+// then a canonical NaN value is returned instead.
+// TODO: Use right-open interval when available. See https://github.com/rems-project/sail/issues/471
+val nan_unbox : forall 'n 'm, 'm in {16, 32, 64, 128} & 'n >= 'm . (implicit('m), bits('n)) -> bits('m)
+function nan_unbox(m, x) = if 'n == 'm then x else (
+  if x['n - 1 .. 'm] == ones() then x['m - 1 .. 0] else canonical_NaN()
+)
 
 /* **************************************************************** */
 /* Floating point register file                                     */
--- a/model/riscv_insts_dext.sail
+++ b/model/riscv_insts_dext.sail
@@ -934,18 +934,14 @@ function clause execute (F_UN_TYPE_D(rs1, rd, FCLASS_D)) = {
 }
 
 function clause execute (F_UN_TYPE_D(rs1, rd, FMV_X_D)) = {
-  assert(xlen >= 64);
-  let rs1_val_D            = F(rs1)[63..0];
-  let rd_val_X  : xlenbits = sign_extend(rs1_val_D);
-  X(rd) = rd_val_X;
+  assert(xlen >= 64 & flen >= 64);
+  X(rd) = sign_extend(F(rs1)[63..0]);
   RETIRE_SUCCESS
 }
 
 function clause execute (F_UN_TYPE_D(rs1, rd, FMV_D_X)) = {
-  assert(xlen >= 64);
-  let rs1_val_X            = X(rs1);
-  let rd_val_D             = rs1_val_X [63..0];
-  F(rd) = rd_val_D;
+  assert(xlen >= 64 & flen >= 64);
+  F(rd) = nan_box(X(rs1)[63..0]);
   RETIRE_SUCCESS
 }
 
--- a/model/riscv_insts_fext.sail
+++ b/model/riscv_insts_fext.sail
@@ -1061,16 +1061,12 @@ function clause execute (F_UN_TYPE_S(rs1, rd, FCLASS_S)) = {
 }
 
 function clause execute (F_UN_TYPE_S(rs1, rd, FMV_X_W)) = {
-  let rs1_val_S            = F(rs1)[31..0];
-  let rd_val_X  : xlenbits = sign_extend(rs1_val_S);
-  X(rd) = rd_val_X;
+  X(rd) = sign_extend(F(rs1)[31..0]);
   RETIRE_SUCCESS
 }
 
 function clause execute (F_UN_TYPE_S(rs1, rd, FMV_W_X)) = {
-  let rs1_val_X            = X(rs1);
-  let rd_val_S             = rs1_val_X [31..0];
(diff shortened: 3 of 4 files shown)


arichardson (approved):
Minor nit: The title of the commit sounds like there is an implementation bug. Maybe `Fix RV64F compilation and simplify fmv implementation`?

Change LGTM.

model/riscv_insts_dext.sail
```diff
@@ -934,18 +934,14 @@ function clause execute (F_UN_TYPE_D(rs1, rd, FCLASS_D)) = {
 }
 
 function clause execute (F_UN_TYPE_D(rs1, rd, FMV_X_D)) = {
-  assert(xlen >= 64);
-  let rs1_val_D            = F(rs1)[63..0];
-  let rd_val_X  : xlenbits = sign_extend(rs1_val_D);
-  X(rd) = rd_val_X;
```
jrtc27, line 940: Please keep the extends/truncates, they generalise the code to support xlen != 64 and flen != 64 (RV128 or RVQ), whereas yours assumes xlen == flen == 64.

jrtc27 (changes requested):

model/riscv_insts_fext.sail
```diff
@@ -1056,16 +1056,16 @@ function clause execute (F_UN_TYPE_S(rs1, rd, FCLASS_S)) = {
 }
 
 function clause execute (F_UN_TYPE_S(rs1, rd, FMV_X_W)) = {
-  let rs1_val_S            = F(rs1)[31..0];
-  let rd_val_X  : xlenbits = sign_extend(rs1_val_S);
-  X(rd) = rd_val_X;
+  // A narrower n-bit transfer out of the floating-point registers will
+  // transfer the lower n bits of the register ignoring the upper FLEN-n bits.
```
jrtc27, line 1060: I don't see the point of these comments? These behave exactly as one would expect, and are entirely consistent with the rest of the ISA.

jrtc27: That is, reading a 32-bit value reads that 32-bit sub register, and writing one either sign-extends, in the case of integer destinations, or nan boxes, in the case of float destinations.

model/riscv_insts_dext.sail
jrtc27: That is, the only code changes needed should be to add ` && flen >= 64` to the asserts.

jordancarlin: Agreed. This will need to be reverted in #445, so no reason to remove it here. Simplifying it to all happen in-line like you did with F seems cleaner though.

Timmmm: Hmm good point! I think it was technically missing the `nan_box`, presumably because the Q version wasn't available here so I added it, and pulled the code from #448 to make the nan boxing functions support all of the float sizes.

model/riscv_insts_fext.sail
Timmmm: Yeah it wasn't meant to contradict any expectations - it's a direct quote from the spec; just explain why the code was doing what it is (because it's not obvious IMO unless you are an expert). I can remove it if you like. Or maybe put it in quotes?

jordancarlin (approved):
Looks good with the latest changes. A nice cleanup + fixes a clear problem with building the model.

model/riscv_insts_fext.sail
jrtc27: I would delete it. The existence of the comment suggest something different to what is expected / done elsewhere, but it is entirely consistent and so to me increases confusion. This code is entirely what someone would expect for an instruction that is “move a 32-bit float to an X register”.

jrtc27: (and similarly for the other direction below)

model/riscv_insts_zfa.sail
```diff
@@ -52,7 +52,7 @@ function clause execute (RISCV_FLI_H(constantidx, rd)) = {
     0b11100 => { 0x7800 },  /* 2^15 */
     0b11101 => { 0x7c00 },  /* 2^16 */
     0b11110 => { 0x7c00 },  /* +inf */
-    0b11111 => { canonical_NaN_H() },
+    0b11111 => { canonical_NaN(16) },
```
jrtc27, line 55: Why have we lost the suffixed versions? It’s nicer being able to consistently name things with _H etc rather than passing in bit widths as numbers.

jrtc27: (And reduces churn if you keep them too)

model/riscv_fdext_regs.sail
```diff
@@ -15,55 +15,30 @@
...
-/* When 16-bit floats (half-precision) are read from 32/64-bit regs  */
-/* they must be 'NaN unboxed'.                                       */
-/* When 32-bit floats (single-precision) are stored in 64-bit regs  */
-/* they must be 'NaN boxed' (upper 32b all ones).                   */
-/* When 32-bit floats (single-precision) are read from 64-bit regs  */
-/* they must be 'NaN unboxed'.                                      */
-
-function canonical_NaN_H() -> bits(16) = 0x_7e00
-function canonical_NaN_S() -> bits(32) = 0x_7fc0_0000
-function canonical_NaN_D() -> bits(64) = 0x_7ff8_0000_0000_0000
```
arichardson, line 29: //                                      sign  exponent    significand
function canonical_NaN_H() -> bits(16) = 0b0 @ ones(5)  @ 0b1 @ zeros(9)
function canonical_NaN_S() -> bits(32) = 0b0 @ ones(8)  @ 0b1 @ zeros(22)
function canonical_NaN_D() -> bits(64) = 0b0 @ ones(11) @ 0b1 @ zeros(51)
function canonical_NaN_Q() -> bits(64) = 0b0 @ ones(15) @ 0b1 @ zeros(111)
```

Could do this instead of the match lower down?

Comment (Timmmm) [reply model/riscv_fdext_regs.sail:29]: Good idea. I did this.

File: model/riscv_insts_zfa.sail (hunk 3 above)
Comment (Timmmm) [reply model/riscv_insts_zfa.sail]: Ok I kept both versions and defined the generic one in terms of the suffixed ones.

File: model/riscv_insts_fext.sail (hunk 2 above)
Comment (Timmmm) [reply model/riscv_insts_fext.sail]: Ok, deleted.

**Comment by arichardson** [comment]
> Btw `canonical_NaN()` can actually be completely generic with some of the new Sail library code:
> 
> ```
> $include <float/common.sail>
> 
> val canonical_NaN : forall 'n, 'n in {16, 32, 64, 128} . (implicit('n)) -> bits('n)
> function canonical_NaN(n) = float_compose(struct {
>     sign = 0b0,
>     exp = ones(),
>     mantissa = 0b1 @ zeros(),
>   })
> ```
> 
> Ok actually that's not quite true. `<float/common.sail>` is missing `float_compose`, but it's very simple (I'll make a PR):
> 
> ```
> val      float_compose : forall 'n, 'n in { 16, 32, 64, 128 }. float_bits('n) -> bits('n)
> function float_compose(op) = op.sign @ op.exp @ op.mantissa
> ```
> 
> And also `0b1 @ zeros()` [doesn't quite work](https://github.com/rems-project/sail/issues/746), so you currently need this full code:
> 
> ```
> $include <float/common.sail>
> 
> val      float_compose : forall 'n, 'n in { 16, 32, 64, 128 }. float_bits('n) -> bits('n)
> function float_compose(op) = op.sign @ op.exp @ op.mantissa
> 
> val leading_one : forall 'n, 'n >= 1. (implicit('n)) -> bits('n)
> function leading_one(n) = 0b1 @ zeros('n - 1)
> 
> val canonical_NaN : forall 'n, 'n in {16, 32, 64, 128} . (implicit('n)) -> bits('n)
> function canonical_NaN(n) = float_compose(struct {
>     sign = 0b0,
>     exp = ones(),
>     mantissa = leading_one(),
>   })
> ```
> 
> Very cool though right? IMO we should do this in future and switch the `_D` suffixed versions to pass integers in.

Yes this looks much nicer!

**Review by arichardson — APPROVED** [review]

**Comment by jordancarlin** [comment]
> Very cool though right? IMO we should do this in future and switch the _D suffixed versions to pass integers in.

Definitely nicer that way. For now you could put the logic for the matching in the `canonical_NaN` function (more similar to what you had initially) and have the _D suffixed versions pass the integer in now. That way there is less changing and it would just be the internals of that one function that get modified later.

**Comment by jrtc27** [comment]
> > Very cool though right? IMO we should do this in future and switch the _D suffixed versions to pass integers in.
> 
> Definitely nicer that way. For now you could put the logic for the matching in the `canonical_NaN` function (more similar to what you had initially) and have the _D suffixed versions pass the integer in now. That way there is less changing and it would just be the internals of that one function that get modified later.

Yeah, having the suffixed versions be wrappers around the integer-taking version was what I was imagining.

**Comment by Timmmm** [comment]
> Yeah, having the suffixed versions be wrappers around the integer-taking version was what I was imagining.

Ok I updated it to this.

File: model/riscv_insts_fext.sail [hunk 5]
```diff
@@ -1061,16 +1061,12 @@ function clause execute (F_UN_TYPE_S(rs1, rd, FCLASS_S)) = {
 }
 
 function clause execute (F_UN_TYPE_S(rs1, rd, FMV_X_W)) = {
-  let rs1_val_S            = F(rs1)[31..0];
-  let rd_val_X  : xlenbits = sign_extend(rs1_val_S);
-  X(rd) = rd_val_X;
+  X(rd) = sign_extend(F(rs1)[31 .. 0]);
```
jrtc27, line 1064:
```suggestion
  X(rd) = sign_extend(F(rs1)[31..0]);
```

model/riscv_fdext_regs.sail
```diff
@@ -15,55 +15,37 @@
...
+function canonical_NaN_S() -> bits(32)  = canonical_NaN(32)
+function canonical_NaN_D() -> bits(64)  = canonical_NaN(64)
+function canonical_NaN_Q() -> bits(128) = canonical_NaN(128)
+
+// When an n-bit float is stored in a larger m-bit register it is "boxed"
+// by prepending 1s, which make it appear as a qNaN.
+val nan_box : forall 'n 'm, 'n <= 'm . (implicit('m), bits('n)) -> bits('m)
+function nan_box(n, x) = ones('m - 'n) @ x
+
+// When an n-bit float is stored ina smaller m-bit register it is "unboxed"
```
jrtc27, line 42:
```suggestion
// When an n-bit float is stored in a smaller m-bit register it is "unboxed"
```

model/riscv_fdext_regs.sail
```diff
@@ -15,55 +15,37 @@
...
+// - only if it is a valid boxed NaN. Otherwise a canonical NaN value is stored.
```
jrtc27, line 43: Not just stored, using as an input to some computation sees the same thing, right? As in this is just a property of any read of a non-NaN-boxed value?

jrtc27: And in particular _transfer_ instructions (FLn/FSn/FMV.n.X/FMV.X.n) _don't_ do this, despite storing in a smaller register.

Timmmm: Ah yeah you're right I didn't really describe it correctly (and I think the spec is not especially readable). How about this:

```
// Except for floating-point transfer instructions (FSn and FMV.n.X),
// n-bit reads of a >n-bit floating point register "unboxes" the value stored
// in the register. If the register does not contain a valid boxed value
// then a canonical NaN value is returned instead.
```

and

```
// n-bit writes to a >n-bit floating point register "boxes" the value
// by prepending 1s, which turn it into a qNaN.
```

jrtc27: (un)boxes should be (un)box as you have plural reads/writes, but yes that looks ok to me otherwise

arichardson (approved):

model/riscv_fdext_regs.sail
Timmmm: Done.

jordancarlin (approved):

PR #596, opened 2024-10-15
Move Sail exception code into its own file

Timmmm:
This is not necessary for the model currently, however internally we wanted to use `internal_error` in a file that was before `riscv_types.sail`. Putting it in its own file makes that possible, and I think it's an organisational improvement even if the model currently does not need it.


arichardson (approved):
Maybe it's simpler to move this into prelude.sail? But either way this LGTM.

PR #597, opened 2024-10-15
Refactor chip reset

Timmmm:
This simplifies, clarifies and fixes the reset functionality.

Until now the model conflates reset and initialisation, and does way more than it should on reset. The RISC-V spec only requires a very small number of things to be reset.

This change:

1. Renames the `init` functions to `reset`, to clarify that they correspond to resetting the chip.
2. Removes the `ext_init` and `ext_rvfi_init` functions. The latter is not used and the former is only used by the old CHERI code.
3. Removes the reset of the X and F registers. These are non-reset.
4. Removes the reset of various CSRs that are non-reset (`mip`, `mie`, `mideleg`, `mtvec`, `mepc`, etc).
5. Adds reset of `mstatus[MIE]` and `mstatus[MPRV]`. As far as I can see they were missing.
6. Add one-time init of `mhartid` etc to 0.

I didn't remove the vector register resets yet. That needs a bigger refactor.

Also note that currently there is no way to actually do a chip reset mid-simulation, but that will be needed eventually.

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -307,22 +307,15 @@ function handle_exception(e: ExceptionType) -> unit = {
 function handle_interrupt(i : InterruptType, del_priv : Privilege) -> unit =
   set_next_pc(trap_handler(del_priv, true, interruptType_to_bits(i), PC, None(), None()))
 
-/* state state initialization */
-
-function init_sys() -> unit = {
-  cur_privilege = Machine;
-
-  mhartid     = zeros();
-  mconfigptr  = zeros();
-
-  misa[MXL] = architecture(if xlen == 32 then RV32 else RV64);
+// Reset misa to enable the maximal set of supported extensions.
+function reset_misa() -> unit = {
   misa[A]   = 0b1;                             /* atomics */
   misa[C]   = bool_to_bits(sys_enable_rvc());  /* RVC */
   misa[B]   = bool_to_bits(sys_enable_bext()); /* Bit-manipulation */
   misa[I]   = 0b1;                             /* base integer ISA */
   misa[M]   = 0b1;                             /* integer multiply/divide */
-  misa[U]   = 0b1;                             /* user-mode */
-  misa[S]   = 0b1;                             /* supervisor-mode */
+  misa[U]   = bool_to_bits(sys_enable_user()); /* user-mode */
+  misa[S]   = bool_to_bits(sys_enable_supervisor()); /* supervisor-mode */
   misa[V]   = bool_to_bits(sys_enable_vext()); /* vector extension */
 
   if   sys_enable_fdext() & sys_enable_zfinx()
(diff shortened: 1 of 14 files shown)


jrtc27:
> Removes the ext_init and ext_rvfi_init functions. The latter is not used and the former is only used by the old CHERI code.

ext_init should remain (possibly renamed). Our sail-cheri-riscv uses it so this would block us updating if we need to keep it going. But, more importantly, we shouldn't delete hooks just because one possible user isn't using them. Any hook that has a real use case is supposed to be there, it's just the only use case so far is sail-cheri-riscv and thus the hooks that exist are the ones that it needed.

jrtc27:
> > Removes the ext_init and ext_rvfi_init functions. The latter is not used and the former is only used by the old CHERI code.
> 
> ext_init should remain (possibly renamed). Our sail-cheri-riscv uses it so this would block us updating if we need to keep it going. But, more importantly, we shouldn't delete hooks just because one possible user isn't using them. Any hook that has a real use case is supposed to be there, it's just the only use case so far is sail-cheri-riscv and thus the hooks that exist are the ones that it needed.

Also, the latter is very much used:

https://github.com/CTSRD-CHERI/sail-cheri-riscv/blob/c93d5eff4132fd52ddad32020cedae93573bfa29/src/cheri_regs.sail#L156-L253

and still used (renamed) in the model for the proposed standard version:

https://github.com/CHERI-Alliance/sail-cheri-riscv/blob/a1ce8f9bebd3397596e683fb90e62e6bbcb8b925/src/cheri_regs.sail#L100-L138

jrtc27 (changes requested):
ext_(rvfi_)init should stay as ext_(rvfi_)reset

jrtc27:
> Removes the reset of the X and F registers. These are non-reset.

They should be initialised for RVFI though.

jrtc27:
> > Removes the reset of the X and F registers. These are non-reset.
> > Removes the reset of various CSRs that are non-reset (mip, mie, mideleg, mtvec, mepc, etc).
> 
> They should be initialised for RVFI though.

model/riscv_sys_control.sail
```diff
@@ -367,40 +360,47 @@ function init_sys() -> unit = {
...
+  // "Upon reset, a hard's privilege mode is set to M."
+  cur_privilege = Machine;
+
+  // "The mstatus fields MIE and MPRV are reset to 0."
+  mstatus[MIE] = 0b0;
+  mstatus[MPRV] = 0b0;
+
+  // "If little-endian memory accesses are supported, the mstatus/mstatush field
+  // MBE is reset to 0."
+  // The model doesn't support little-endian (see legalize_mstatus()).
```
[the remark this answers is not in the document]
Timmmm: Oops, I updated the comment. I think the handling of `mstatush` is pretty wonky at the moment so I made an issue to track that: #639.

Timmmm:
Ah yes you're quite right about the `ext_` functions missing. I renamed it to `ext_reset` [here](https://github.com/CHERI-Alliance/sail-riscv/blob/e9b80042b2f881c045615259dd58c8fcb61b0aa8/model/riscv_step.sail#L130) - I'll add that back in.

> Our sail-cheri-riscv uses it so this would block us updating if we need to keep it going. But, more importantly, we shouldn't delete hooks just because one possible user isn't using them. Any hook that has a real use case is supposed to be there, it's just the only use case so far is sail-cheri-riscv and thus the hooks that exist are the ones that it needed.

Yeah I'm not exactly sure what the story is on these ext_ hooks and what happens to the CHERI repos. It seems like it's making good progress towards ratification and then it would make sense for it to just be directly in the model. We probably still want to keep its code out of the way of people who don't care about CHERI I guess though. Do you know of anyone else using these hooks?

Could we discuss it in the CHERI Tech chat on Thursday?

> They should be initialised for RVFI though.

The generated C code still initialises them to 0 when the model is initialised by `model_init()` (which is also called when RVFI restarts its stimulus). The difference is that if you do a mid-simulation reset then their values will no longer be changed.

Currently there isn't really an interface to do that, but we do it internally via our lock-step verif system. When the real DUT gets reset, we reset the Sail model by calling `sys_reset()`. We also have an implementation-specific `ext_reset()` that does the extra reset that our chip does that isn't required by the standard.

jordancarlin:
@Timmmm I still see the `ext_init` functions being removed and don't see the replacement `ext_reset` being added. I'll wait to review in detail again until that is updated.

Timmmm:
Oops totally forgot about that. I've added `ext_reset`.

arichardson:
Not sure why but ci is unhappy: 

```
Type error:
model/riscv_step.sail:132.2-11:
132 |  ext_reset();
    |  ^-------^
    | Function ext_reset not found
```

Timmmm:
Oops I forgot about RVFI. Should pass now.

arichardson (commented):
Overall this looks good, but as far as I can tell this will break the RVFI support. It should reset registers to all zero so you have a reproducible starting point for divergence testing.

Timmmm:
I'm pretty sure that reset is done automatically by `model_init()` which is called between RVFI sessions. I'll double check though.

arichardson:
> I'm pretty sure that reset is done automatically by `model_init()` which is called between RVFI sessions. I'll double check though.

Ah yes it does call model_init in the rvfi loop. Does this re-run all sail global initializers? If so then please ignore my comment. LGTM.

arichardson (approved):

jordancarlin (approved):
LGTM now. There will definitely be conflicts with #652 though.

jordancarlin:
This would be a good one to get merged before conflicts start building up. @jrtc27 seems like the ext_* hooks have been restored. Any further concerns?

arichardson (approved):

model/riscv_sys_regs.sail
```diff
@@ -288,6 +291,15 @@ function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
   ]
 }
 
+register mstatus : Mstatus = {
+  // Initialise SXL and UXL.
+  let mxl = arch_to_bits(if xlen == 32 then RV32 else RV64);
+  [ Mk_Mstatus(zeros()) with
+    SXL = mxl,
+    UXL = mxl,
```
Timmmm, line 299: This has the same problems identified [here](https://github.com/riscv/sail-riscv/pull/683#issuecomment-2619821148): `SXL`/`UXL` should be zero on RV32 or if the corresponding mode is not implemented.

I'll fix it.

Timmmm: Ok this should be correct now. @jrtc27 @jordancarlin @radimkrcmar

jordancarlin: Since we don't have a modifiable xlen yet, this seems good to me

jordancarlin (approved):

model/riscv_sys_regs.sail
Timmmm: Somehow broke the tests though... I'll investigate.

Timmmm: Ah I figured it out. It's an initialisation order thing - `extensionEnabled(Ext_S/U)` reads `misa[S/U]` but that hasn't been initialised yet.

I changed it to be the other way around so `misa[U/S]` is driven by `extensionEnabled(Ext_S/U)` and those are hardcoded to `true` (pending the new config system).

It's maybe not quite right but we should fix all this stuff when the new config system is available.

Timmmm: Btw the spec isn't super clear but it at least sort of implies that `misa[U/S]` are read-only:

> The "U" and "S" bits will be set if there is support for user and supervisor modes respectively.

jordancarlin: Interesting. That might be worth asking about over on the isa-manual repo.

jordancarlin: Looks like it is supposed to be writable. MISA is supposed to default to the max set of supported CSRs, so if we add `sys_enable_U` and `sys_enable_S` functions (hardcoded to true for now, to be replaced by config system soon) we can use those for the reset value of MISA and then have the `extensionEnabled` clauses be based off of the (writable) current value of MISA.

Timmmm: Makes sense, I've done that.

model/riscv_sys_control.sail
```diff
@@ -307,22 +307,15 @@ function handle_exception(e: ExceptionType) -> unit = {
...
+// Reset misa to enable the maximal set of supported extensions.
+function reset_misa() -> unit = {
   misa[A]   = 0b1;                             /* atomics */
   misa[C]   = bool_to_bits(sys_enable_rvc());  /* RVC */
   misa[B]   = bool_to_bits(sys_enable_bext()); /* Bit-manipulation */
   misa[I]   = 0b1;                             /* base integer ISA */
   misa[M]   = 0b1;                             /* integer multiply/divide */
-  misa[U]   = 0b1;                             /* user-mode */
-  misa[S]   = 0b1;                             /* supervisor-mode */
+  misa[U]   = bool_to_bits(extensionEnabled(Ext_U)); /* user-mode */
```
jordancarlin, line 317: Pretty sure this should be `sys_enable_user()`

Timmmm: Err, oops! Shouldn't code before I've woken up...

Timmmm: Forgot to press Save. :-/ I need some way of blocking `git push` if there are unsaved files :-D

Should be fixed now.

jordancarlin (approved):

Issue #583, opened 2024-10-09
Merge 'with' updates


Timmmm:
There's quite a lot of code like this:

```
function lower_mstatus(m : Mstatus) -> Sstatus = {
  let s = Mk_Sstatus(zero_extend(0b0));
  let s = [s with SD = m[SD]];
  let s = set_sstatus_UXL(s, get_mstatus_UXL(m));
  let s = [s with MXR = m[MXR]];
  let s = [s with SUM = m[SUM]];
  let s = [s with XS = m[XS]];
  let s = [s with FS = m[FS]];
  let s = [s with VS = m[VS]];
  let s = [s with SPP = m[SPP]];
  let s = [s with SPIE = m[SPIE]];
  let s = [s with UPIE = m[UPIE]];
  let s = [s with SIE = m[SIE]];
  let s = [s with UIE = m[UIE]];
  s
}
```

which should be written like this:

```
function lower_mstatus(m : Mstatus) -> Sstatus = {
  let s = Mk_Sstatus(zero_extend(0b0));
  [s with
    SD = m[SD],
...
    SPP = m[SPP],
    SPIE = m[SPIE],
    UPIE = m[UPIE],
    SIE = m[SIE],
    UIE = m[UIE],
  ]
}
```

jordancarlin:
Can be closed with #586 merged

Timmmm:
I think there's a few more that could be merged:

* `legalize_mstatus`
* `lower_mstatus` (mostly done but we missed `SD`)
* `lift_sie`

PR #600, opened 2024-10-19
Merging more with updates

jordancarlin:
Combine additional "with" updates per #583

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -246,35 +246,31 @@ function legalize_mstatus(o : Mstatus, v : xlenbits) -> Mstatus = {
    */
   let m : Mstatus = Mk_Mstatus(zero_extend(v[22 .. 7] @ 0b0 @ v[5 .. 3] @ 0b0 @ v[1 .. 0]));
 
-  /* Legalize MPP */
-  let m = [m with MPP = if have_privLevel(m[MPP]) then m[MPP]
-                        else privLevel_to_bits(lowest_supported_privLevel())];
-
-  /* We don't have any extension context yet. */
-  let m = [m with XS = extStatus_to_bits(Off)];
-
   /* FS is WARL, and making FS writable can support the M-mode emulation of an FPU
    * to support code running in S/U-modes.  Spike does this, and for now, we match it,
    * but only if Zfinx isn't enabled.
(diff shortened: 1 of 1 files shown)


model/riscv_sys_regs.sail
```diff
@@ -271,11 +272,7 @@ function legalize_mstatus(o : Mstatus, v : xlenbits) -> Mstatus = {
   let m = if xlen == 64 then {
```
Timmmm, line 272: You can drop the `m`, i.e. just end with

```
if xlen == 64 then {
...
} else m
```

model/riscv_sys_regs.sail
```diff
@@ -246,22 +246,23 @@ function legalize_mstatus(o : Mstatus, v : xlenbits) -> Mstatus = {
...
    * but only if Zfinx isn't enabled.
    * FIXME: This should be made a platform parameter.
    */
-  let m = if sys_enable_zfinx() then [m with FS = extStatus_to_bits(Off)] else m;
   let dirty = extStatus_of_bits(m[FS]) == Dirty | extStatus_of_bits(m[XS]) == Dirty |
               extStatus_of_bits(m[VS]) == Dirty;
-  let m = [m with SD = bool_to_bits(dirty)];
+
+  /* Legalize MPP */
+  let m = [m with MPP = if have_privLevel(m[MPP]) then m[MPP]
```
Timmmm, line 258: I dunno the exact best way to format these but elsewhere I've been doing it like this:

```
let m = [m with
  MPP = ...
  ...
  MPRV = ...,
];
```

Also note the trailing comma.

Hopefully one day we can use `sail -format` or whatever.

Timmmm (changes requested):

Timmmm (approved):

Timmmm:
LGTM; I'll merge in a couple of days if nobody objects.

arichardson (approved):

model/riscv_sys_regs.sail
jordancarlin: I believe I changed it to use that style already, but if you think further modifications would help let me know.

arichardson: Sorry that was an old pending comment that was only submitted when I added my review. This has been addressed.

PR #601, opened 2024-10-19
Added support of Smepmp

KotorinMinami:
added command line support for Smepmp. command line switch: --enable-smepmp.  Sail function: haveSmepmp

Added support of Smepmp

This pr was written by @HamzaKh01, merged with the command-line switch code provided by Bill, and adjusted by myself

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -340,7 +340,11 @@ private function legalize_mseccfg(o : Seccfg, v : bits(64)) -> Seccfg = {
   ]
 }
 
-register mseccfg : Seccfg = legalize_mseccfg(Mk_Seccfg(zeros()), zeros())
+// TODO: Ideally this would use legalize_mseccfg however it can't due to a Sail
+// compiler bug. In this case that's fine since the result is actually all zeros anyway.
+// See https://github.com/rems-project/sail/issues/1653
+register mseccfg : Seccfg = Mk_Seccfg(zeros())
+// register mseccfg : Seccfg = legalize_mseccfg(Mk_Seccfg(zeros()), zeros())
 
 mapping clause csr_name_map = 0x747  <-> "mseccfg"
 mapping clause csr_name_map = 0x757  <-> "mseccfgh"


Timmmm:
[Specification is here](https://riscv-specs.timhutt.co.uk/spec/20240411/priv-isa-asciidoc.html#smepmp)

model/riscv_pmp_regs.sail
```diff
@@ -43,6 +43,15 @@ bitfield Pmpcfg_ent : bits(8) = {
...
 register pmpaddr_n : vector(64, xlenbits)
 
+bitfield Mseccfg_ent : xlenbits = {
+  RLB  : 2,       /* Rule Locking Bypass */
+  MMWP : 1,       /* Machine Mode Whitelist Policy */
+  MML  : 0        /* Machine Mode Lockdown */
+}
+
+register mseccfg  : Mseccfg_ent
+register mseccfgh : bits(32)
```
Timmmm, line 53: Normally we don't treat this as a separate register, we just make `bitfield Mseccfg : bits(64)` and then read/write the appropriate bits of it.

We have an implementation of `mseccfg` like that internally; I'll make a PR for it.

model/riscv_sys_regs.sail
```diff
@@ -105,6 +105,11 @@ val sys_pmp_grain = pure "sys_pmp_grain" : unit -> range(0, 63)
 /* Which HPM counters are supported (as a bit mask). Bits [2 .. 0] are ignored. */
 val sys_writable_hpm_counters = pure "sys_writable_hpm_counters" : unit -> bits(32)
 
+/* Enable Smepmp */
+val sys_enable_smepmp = pure "sys_enable_smepmp" : unit -> bool
+/* TODO: for extension smmpm and zkr to enable */
+function sys_has_mseccfg() -> bool = sys_enable_smepmp()
```
Timmmm, line 111: You can implement `mseccfg` even if you don't have any of the extensions that use it (and that isn't a weird edge case). So ideally this would be

```
// Callback to C
val sys_have_mseccfg : "sys_have_mseccfg" unit -> bool

function have_mseccfg() -> bool = sys_have_mseccfg() | sys_enable_smepmp() | sys_enable_zkr() ...
```

model/riscv_pmp_control.sail
```diff
@@ -48,14 +48,43 @@ function pmpCheckRWX(ent, acc) = {
   }
 }
 
-// this needs to be called with the effective current privilege.
+/* this needs to be called with the effective current privilege */
```
Timmmm, line 51: `//` style comments are nicer IMO. I don't see why we would change them back?

Timmmm (commented):
Still working on reading all the boolean logic...

model/riscv_pmp_control.sail
KotorinMinami: This is the part that was changed by @HamzaKh01, I didn't check this as I was too focused on the code implementation part, we can change it back if needed, but there are many one line comments like `/* ... */` in Sail, we might need to declare it better in the codestyle after some discussion?

model/riscv_sys_regs.sail
KotorinMinami: You mean we need to add an extra command line parameter to act as an extra switch to control the presence of mseccfg to make sure that even if we don't implement extensions that use this register, we can still implement this register, right ?

Timmmm: Yeah exactly. I don't think we need to actually bother hooking it up to a CLI arguments at the moment - we can wait until Alasdair's new config system exists and then it will be a whole lot less tedious. But we can just have a function that returns `true` as a placeholder so we don't forget.

model/riscv_pmp_control.sail
Timmmm: Yeah there is an experimental auto-formatter in the Sail compiler but I don't know what state it got to and I'd guess it doesn't change this sort of thing.

KotorinMinami:
So does someone review this code and merge it in?

model/riscv_pmp_control.sail
```diff
@@ -123,7 +152,18 @@ function pmpCheck forall 'n, 'n > 0. (addr: xlenbits, width: int('n), acc: Acces
       PMP_Continue => (),
     }
   };
-  if priv == Machine then None() else Some(accessToFault(acc))
+  if priv == Machine
+  then {
+    if sys_has_mseccfg() & mseccfg[MMWP] == 0b1                /* Make Read, Write and execute denied by default, if condition meets for M mode */
```
Timmmm, line 157: Can you move the comments to their own line with `//`. These are too long, and end-of-line comments are kind of awkward.

model/riscv_pmp_control.sail
```diff
@@ -123,7 +152,18 @@ function pmpCheck forall 'n, 'n > 0. (addr: xlenbits, width: int('n), acc: Acces
...
+    then Some(accessToFault(acc))
+    else if sys_has_mseccfg() & mseccfg[MML] == 0b1            /* Make execute denied, if condition meets for M mode */
+    then match acc {
+      Execute(_) => Some(accessToFault(acc)),
+      _          => None()
+    }
+    else None()
+  }
+  else Some(accessToFault(acc))
```
Timmmm, line 166: I think this could be a lot simpler:

```
if priv == Machine & mseccfg[MMWP] != 0b1 & not(mseccfg[MML] == 0b1 & isExecute(acc))
then None() else Some(accesstoFault(acc)
```

(add `isExecute()` as a helper)

Timmmm (changes requested):
I did a bit more reviewing. I'll do more when it's been updated.

model/riscv_pmp_control.sail
```diff
@@ -48,14 +48,43 @@ function pmpCheckRWX(ent, acc) = {
...
+          _       => false
+        },
+      (_, _, _, _) =>
+        if (priv == Machine) == pmpLockBit(ent)
+        then pmpCheckRWX(ent, acc)
+        else false
+    }
+  }
+  else{
+    match priv {
```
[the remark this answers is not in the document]
Timmmm: Oh yeah look it was me... #738 :-D

I've merged that now.

model/riscv_pmp_regs.sail
```diff
@@ -254,3 +276,43 @@ function clause write_CSR(0x3B @ idx : bits(4), value) = { let idx = unsigned(0b
...
+    // Implements stickiness of MML bit, if once set remains set
+    (0b0, 0b0) => [reg with MMWP = legal_v[MMWP], MML = legal_v[MML]],
+    (0b0, 0b1) => [reg with MML = legal_v[MML]],
+    (0b1, 0b0) => [reg with MMWP = legal_v[MMWP]],
+    (0b1, 0b1) => reg
+  };
+  reg
+}
+
+mapping clause csr_name_map = 0x747  <-> "mseccfg"
```
jordancarlin, line 306: All of this mseccfg related code should probably move to sys_regs or another similar file because this CSR is not exclusive to pmp

KotorinMinami: Oh, that can be awkward, because mseccfg's RLB field depends on pmpNcfg's L field, and pmpNcfg's L field depends on the RLB field, and their highly coupled relationship would be difficult to handle if they were to be written separately

Timmmm: Yeah this is an ongoing problem with these "global" CSRs (also `menvcfg`, `mstatus`, etc.). But I agree with @jordancarlin we should move it.

To solve the circular dependency you can forward-declare functions using `val`. We'll hopefully come up with a better way of dealing with this after the new project system is in place.

KotorinMinami: Okay, I've pushed the moved version.

UmerShahidengr:
Has there been any update on this one? Will this be merged soon?

KotorinMinami:
I may rebase it once again. However, I don't know how it's going

Timmmm:
Hey, I've rebased this and tweaked the code a bit, and also added a workaround for the segfault. Hope that's ok. I haven't given the actual logic a full review yet though.

KotorinMinami:
Oh, thank you—initially, the Sail compile bugs did cause me to pause development for a while, but looking at the code now, I think it's ok.

PR #602, opened 2024-10-20
Documentation Updates: reorder README and combine with STATUS.md

### LINKED ISSUE #576: Outdated Status.md file
The [docs/Status.md](https://github.com/riscv/sail-riscv/blob/master/doc/Status.md) file has fallen very far out of date. It is missing many of the extensions now included in the model and only includes a small subset of the supported configuration options. 

It should either be updated or removed. I'd be in favor of removing it since it clearly hasn't been missed, but could also see a case for wanting to keep it and ensure that it stays up to date going forward.

jordancarlin:
Closes #576, the current `STATUS.md` files has fallen very far out of date. This removes it in favor of a list of supported extensions/features in the main README. I also reordered the README to put the "Getting Started" section near the top and did some other general cleanup with it.

Also wondering if it is worth keeping the entire section on what Sail is. We already link to the Sail repo, so it might make more sense to remove this and just include a quick sentence or two overview.

Going through the list of supported extensions also discovered a few missing extensions that we basically already have support for. See #603 and #604.


README.md
```diff
@@ -51,12 +39,112 @@ permission), RISC-V, CHERI-RISC-V, CHERIoT, MIPS, and CHERI-MIPS; all these are
...
+- Zfa extension for additional floating-point instructions, v1.0
+- Zfinx, Zdinx, and Zhinx extensions for floating-point in integer registers, v1.0
+- C extension for compressed instructions, v2.0
+- Zca, Zcf, Zcd, and Zcb extensions for code size reduction, v1.0
+- B (Zba, Zbb, Zbs) and Zbc extensions for bit manipulation, v1.0
+- Zbkb, Zbkc, and Zbkx extensions for bit manipulation for cryptography, v1.0
+- Zkn (Zknd, Zkne, Zknh) and Zks (Zksed, Zksh) extensions for scalar cryptography, v1.0.1
+- Zkr extension for entropy source, v1.0
+- V extension for vector operations, v1.0
+- Machine, Supervisor, and User modes, v1.12?
```
Timmmm, line 138: Yeah good question about the version. It's definitely not the latest everywhere (e.g. for PTE validity checks). Maybe just omit the version for now.

doc/Status.md
```diff
@@ -1,58 +0,0 @@
...
-
-- handling of misaligned data accesses with or without M-mode traps.
-
-- updating of the PTE dirty bit with or without architectural
-  exceptions.
-
-- the contents of the `mtval` register on an illegal instruction
-  exception.
-
-The following ISA features are specified in the
```
Timmmm, line 36: I think it would be worth keeping this list of notable omissions: 

* Hypervisor
* Mutable MXLEN/SXLEN
* Big endian
* RV32E
* PMAs

I don't think we need to mention RV128 or RV64E since I'm pretty sure they don't exist. Or hardware counters because that's kind of expected and fairly minor.

It would be cool to get a table of the RVA22/23 profiles with ticks and crosses for our support, but we don't need to in this PR.

README.md
jordancarlin: Yeah. I know @billmcspadden-riscv has mentioned that he’s working on Priv 1.13, but I’m not sure if we have full Priv 1.12 right now.

doc/Status.md
jordancarlin: Added a section with these omissions. RV64E is ratified, so I included it with RV32E. There are a lot of other extensions/features that are not supported at the moment so it's hard to say where to draw the line, but for now we can just go with this list.

README.md
Timmmm: We definitely don't. This bit is missing from the PTE validity check:

> If pte.v = 0, or if pte.r = 0 and pte.w = 1, **or if any bits or encodings that are reserved for
future standard use are set**

I think just delete the version.

jordancarlin: Seems like it is nebulous what version is currently supported, so removed the version for priv as suggested.

Timmmm (approved):

Timmmm:
LGTM; I'll merge this if nobody objects.

PR #605, opened 2024-10-21
Add parameter name to sys_vector_vlen_exp and sys_vector_elen_exp

jordancarlin:
PR #590 added new functions to `riscv_platform.c` that omit a parameter name. While this works with newer versions of GCC, it causes an error on GCC 10 and earlier. This adds the parameter name like all of the other functions in that file to avoid breaking compatibility with older GCC.

### DIFF
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -82,12 +82,12 @@ uint64_t sys_pmp_grain(unit u)
   return rv_pmp_grain;
 }
 
-uint64_t sys_vector_vlen_exp(unit)
+uint64_t sys_vector_vlen_exp(unit u)
 {
   return rv_vector_vlen_exp;
 }
 
-uint64_t sys_vector_elen_exp(unit)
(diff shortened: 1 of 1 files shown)


Timmmm:
Really? I'm surprised it doesn't work before - it's perfectly legal C. In fact the new code is going to give unused variable warnings (assuming we have warnings turned on; I don't remember).

What error do you get?

jrtc27:
It's legal C for a function declaration but not a function definition (every parameter declaration must have an identifier, except for the special case of `(void)` which must not). At least that's the case in C99, I didn't bother researching whether other versions of the standard have ever allowed that.

Alasdair:
It's now legal in C23 to omit parameter names. It's (always?) been allowed in C++ afaik.

Alasdair (approved):

Timmmm:
Ah fair enough! Nice that the C and C++ folks seem to be fixing some of these discrepancies at least!

PR #606, opened 2024-10-25
remove function sys_enable_next

KotorinMinami:
In the master branch, the N extension has been removed, but at that time, the [commit](https://github.com/riscv/sail-riscv/commit/d79e0bab7b18e0d505383b29db4fa3e76f97ab81) only removed the code in the SAIL section, ignoring the C simulator's support for the N extension (Maybe my fault) , i.e., the sys_enable_next switching function in it was not removed, so pr removed this to avoid inconsistency between the two


Timmmm:
Oops, thanks!

PR #608, opened 2024-10-29
Make calculate new vl configurable

rez5427:
I don't know whether sail-riscv needs this.
The spike uses VLMAX to calculate new vl, and sail-riscv uses (AVL + 1) / 2.

The [spec](https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view?usp=drive_link) says:
> The number of bits implemented in vl depends on the implementation’s maximum vector
> length of the smallest supported type. The smallest vector implementation with VLEN=32
> and supporting SEW=8 would need at least six bits in vl to hold the values 0-32
> (VLEN=32, with LMUL=8 and SEW=8, yields VLMAX=32).

### DIFF
--- a/c_emulator/riscv_platform_impl.c
+++ b/c_emulator/riscv_platform_impl.c
@@ -32,6 +32,8 @@ uint64_t rv_ram_size = UINT64_C(0x4000000);
 uint64_t rv_rom_base = UINT64_C(0x1000);
 uint64_t rv_rom_size = UINT64_C(0x100);
 
+bool rv_vext_vl_use_ceil = false;
+
 // Default 64, which is mandated by RVA22.
 uint64_t rv_cache_block_size_exp = UINT64_C(6);
 
--- a/c_emulator/riscv_platform.c
+++ b/c_emulator/riscv_platform.c
@@ -97,6 +97,11 @@ mach_bits sys_writable_hpm_counters(unit u)
   return rv_writable_hpm_counters;
 }
 
+bool sys_vext_vl_use_ceil(unit u)
+{
+  return rv_vext_vl_use_ceil;
+}
+
 bool plat_enable_dirty_update(unit u)
 {
   return rv_enable_dirty_update;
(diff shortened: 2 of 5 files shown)


Timmmm:
LGTM but I would add some explanation and move the if to the place where it is actually used. Also move the `to_bits` outside the body to make it clearer.

```
val sys_vext_vl_use_ceil = pure "sys_vext_vl_use_ceil" : unit -> bool

function calculate_new_vl(AVL : int, VLMAX : int) -> xlenbits = {
  // See "Constraints on Setting vl" in the vector spec.
  let vl =
    if AVL <= VLMAX then AVL
    else if AVL < 2 * VLMAX then {
      // If VLMAX < AVL < 2 * VLMAX then we can use any value
      // such that ceil(AVL / 2) <= vl <= VLMAX. Here we provide
      // two options: ceil(AVL / 2) or VLMAX.
      if sys_vext_vl_use_ceil() then (AVL + 1) / 2 else VLMAX
    }
    else VLMAX;

  to_bits(xlen, vl)
}
```

(Feel free to copy/paste.)

Timmmm (approved):
LGTM! I'm not a vector expert though so we might want to get the opinion of someone who is (although the spec seems pretty unambiguous on this point). I can ask a colleague if nobody is forthcoming.

avakar:
@Timmmm's colleague here. LGTM.

c_emulator/riscv_platform_impl.c
```diff
@@ -32,6 +32,8 @@ uint64_t rv_ram_size = UINT64_C(0x4000000);
 uint64_t rv_rom_base = UINT64_C(0x1000);
 uint64_t rv_rom_size = UINT64_C(0x100);
 
+bool rv_vext_vl_use_ceil = false;
```
jrtc27, line 35: The commit message doesn't mention that you're changing the default behaviour

rez5427: Sorry..., Is this commit message ok?

Timmmm: I'm not sure which commit message you changed... but don't worry I will squash it and write a descriptive message.

PR #611, opened 2024-11-05
Zero mstatus.UPIE/UIE bits in legalize_mstatus

Timmmm:
This was missed when we removed the N extension.

I'm not really a fan of this code style - in other places we explicitly copy the bits that *are* implemented using their field names, which is also not ideal but definitely better and less error prone. We should change this code to do that at some point. I added a TODO.


arichardson (approved):
There aren't that many fields so could do the assignment explicitly as part of this PR?

Timmmm:
Hmm I started it but it's a decently large refactoring which I don't really have time to do now, sorry. :-/

I think I'll just merge the fix and we can hopefully do it later.

PR #613, opened 2024-11-06
Fix delegated interrupt handling

Timmmm:
When the N extension was removed this code was accidentally broken because it doesn't respect mie for delegated interrupts any more.

The `internal_error`s should be unreachable, at least for the sequential output. I believe formal backends may decide `mideleg` is going to be non-zero even without S, but I think that was an existing issue.

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -244,7 +244,20 @@ function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
     match processPending(mip, mie, mideleg.bits, mIE) {
       Ints_Empty()      => None(),
       Ints_Pending(p)   => Some((p, Machine)),
-      Ints_Delegated(d) => if sIE then Some((d, Supervisor)) else None(),
+      Ints_Delegated(d) => {
+        if not(extensionEnabled(Ext_S)) then {
+          // Can't delegate to user mode. This code is unreachable because
+          // `mideleg.bits` is 0 without supervisor mode.
+          internal_error(__FILE__, __LINE__, "N extension not supported");
+        } else {
+          /* the delegated bits are pending for S-mode */
+          match processPending(Mk_Minterrupts(d), mie, zeros(), sIE) {
+            Ints_Empty()      => None(),
+            Ints_Pending(p)   => Some((p, Supervisor)),
+            Ints_Delegated(d) => internal_error(__FILE__, __LINE__, "N extension not supported"),
+          }
+        }
+      }
     }
   }
 }


model/riscv_sys_control.sail
```diff
@@ -244,7 +244,20 @@ function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
     match processPending(mip, mie, mideleg.bits, mIE) {
       Ints_Empty()      => None(),
       Ints_Pending(p)   => Some((p, Machine)),
-      Ints_Delegated(d) => if sIE then Some((d, Supervisor)) else None(),
+      Ints_Delegated(d) => {
+        if not(extensionEnabled(Ext_S)) then {
+          // Can't delegate to user mode. This code is unreachable because
+          // `mideleg.bits` is 0 without supervisor mode.
```
jrtc27, line 250: Well, the CSR doesn't exist so can't be written to by software and thus the Sail register remains its initial 0 value. The comment written here suggests that the CSR exists and is hardwired to 0 (which was the case in early specs).

Timmmm: The Sail `register mideleg` always exists. That's what I was referring to. Not sure how to make that clearer...

Timmmm:
@KotorinMinami does this look right to you?

KotorinMinami:
@Timmmm 
The [spec](https://riscv-specs.timhutt.co.uk/spec/20240411/priv-isa-asciidoc.html#_machine_trap_delegation_registers_medeleg_and_mideleg:~:text=In%20harts%20with%20S%2Dmode%2C%20the%20medeleg%20and%20mideleg%20registers%20must%20exist%2C%20and%20setting%20a%20bit%20in%20medeleg%20or%20mideleg%20will%20delegate%20the%20corresponding%20trap%2C%20when%20occurring%20in%20S%2Dmode%20or%20U%2Dmode%2C%20to%20the%20S%2Dmode%20trap%20handler.%20In%20harts%20without%20S%2Dmode%2C%20the%20medeleg%20and%20mideleg%20registers%20should%20not%20exist.) mentions:
> In harts with S-mode, the medeleg and mideleg registers must exist, and setting a bit in medeleg or mideleg will delegate the corresponding trap, when occurring in S-mode or U-mode, to the S-mode trap handler. In harts without S-mode, the medeleg and mideleg registers should not exist.

So I think we might need to reflect this in the implementation of the sail's mideleg register, maybe this is the way to solve delegated interrupt handling?

Moreover, based on what is said in spec, we can handle whether medeleg and mideleg exist and the delegated issues by checking misa
> In versions 1.9.1 and earlier , these registers existed but were hardwired to zero in M-mode only, or M/U without N harts. There is no reason to require they return zero in those cases, as the misa register indicates whether they exist.

Timmmm:
> So I think we might need to reflect this in the implementation of the sail's mideleg register

This is handled by `is_CSR_defined()` returning `false` for `m*deleg` when S mode is not implemented.

KotorinMinami:
> This is handled by is_CSR_defined() returning false for m*deleg when S mode is not implemented.

I'm sorry, I didn't notice

But regarding the interrupt handler in the delegate, I think your modification is fine. However, why not change the function `processPending` to such like this, considering that each entry in mie is independent. It can reduce multiple calls to the function and some processing.
```
function processPending(xip : Minterrupts, xie : Minterrupts, xideleg : xlenbits,
                        priv_enabled : bool) -> interrupt_set = {
  /* interrupts that are enabled but not delegated are pending */
  let  effective_pend = xip.bits & xie.bits & ~(xideleg);
  /* the others are delegated */
  let  effective_delg = xip.bits & xie.bits & xideleg;
  /* we have pending interrupts if this privilege is enabled */
  if      priv_enabled & (effective_pend != zeros())
  then    Ints_Pending(effective_pend)
  else if effective_delg != zeros()
  then    Ints_Delegated(effective_delg)
  else    Ints_Empty()
}
```
And maybe we should ensure that S-mode exists before calling the function, which guarantees that the mideleg register exists during reading

Timmmm:
Good point; we've already calculated `effective_pending` so we can do this:

```
function processPending(effective_pending: Minterrupts, xideleg : xlenbits, priv_enabled : bool) -> interrupt_set = {
  /* interrupts that are enabled but not delegated are pending */
  let  effective_non_deleg = effective_pending & ~(xideleg);
  /* the others are delegated */
  let  effective_deleg = effective_pending & xideleg;
  /* we have pending interrupts if this privilege is enabled */
  if      priv_enabled & (effective_non_deleg != zeros())
  then    Ints_Pending(effective_non_deleg)
  else if effective_delg != zeros()
  then    Ints_Delegated(effective_deleg)
  else    Ints_Empty()
}

function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
  let effective_pending = mip.bits & mie.bits;
  if  effective_pending == zeros() then None() /* fast path */
  else {
    /* Higher privileges than the current one are implicitly enabled,
     * while lower privileges are blocked.  An unsupported privilege is
     * considered blocked.
     */
    let mIE = priv != Machine | (priv == Machine & mstatus[MIE] == 0b1);
    let sIE = extensionEnabled(Ext_S) & (priv == User | (priv == Supervisor & mstatus[SIE] == 0b1));
    match processPending(effective_pending, mideleg.bits, mIE) {
      Ints_Empty()      => None(),
      Ints_Pending(p)   => Some((p, Machine)),
      Ints_Delegated(d) => if sIE then Some((d, Supervisor)) else None(),
    }
  }
}
```

Maybe it's clearer just to in-line it since we only call it once then anyway. And I think we can ditch the fast path too. The slow path is going to be fast enough.

```
function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
  assert(extensionEnabled(Ext_S) | mideleg == zeros());

  let pending_m = mip.bits & mie.bits & ~(mideleg);
  let pending_s = mip.bits & mie.bits & mideleg;

  let mIE = (priv == Machine & mstatus[MIE] == 0b1) | priv == User | priv == Supervisor;
  let sIE = (priv == Supervisor & mstatus[SIE] == 0b1)) | priv == User;

  if     mIE & (pending_m != zeros()) then Some((pending_m, Machine))
  else if sIE & (pending_s != zeros()) then Some((pending_s, Supervisor))
  else None()
}
```

That's so simple I can actually understand it. Simple enough that I think it would be suitable for embedding in the ISA manual. Haven't thoroughly checked it yet though.

I think I might merge this first since it is close to the original code which is very well tested. I'll do a follow-up PR to simplify it.

model/riscv_sys_control.sail
billmcspadden-riscv: @jrtc27, @Timmmm : Can we mark this conversation as resolved?

Timmmm: I'm going to merge this; happy to change the comment in future if desired.

PR #615, opened 2024-11-12
use newtype to reflect address type

KotorinMinami:
As mentioned in the #434 , using newtype to refactor the types of virtual memory and physical memory, the xlenbits type in the sail code part of the model which I think that should be virtaddrs or physaddr is refactored to correspond to the virtaddrs/physaddr type. 

Additionally, there seems to be excessive structurally similar redundant code in the memory read and write of the V extension, and I may create a simplified PR later.

### DIFF
--- a/model/prelude_mem_addrtype.sail
+++ b/model/prelude_mem_addrtype.sail
@@ -1,0 +1,18 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+/* Note: virtaddr includes physical addresses when address translation is disabled.
+ * It is used where the address would be virtual if address translation is active.
+ */
+
+newtype physaddr = physaddr : xlenbits
+newtype virtaddr = virtaddr : xlenbits
+
+function physaddr_bits(physaddr(paddr) : physaddr) -> xlenbits = paddr
+
+function virtaddr_bits(virtaddr(vaddr) : virtaddr) -> xlenbits = vaddr
--- a/model/riscv_addr_checks.sail
+++ b/model/riscv_addr_checks.sail
@@ -18,7 +18,7 @@ type ext_fetch_addr_error = unit
  *           any address translation errors are reported for pc, not the returned value.
  */
 function ext_fetch_check_pc(start_pc : xlenbits, pc : xlenbits) -> Ext_FetchAddr_Check(ext_fetch_addr_error) =
-  Ext_FetchAddr_OK(pc)
+  Ext_FetchAddr_OK(virtaddr(pc))
 
 function ext_handle_fetch_check_error(err : ext_fetch_addr_error) -> unit =
   ()
@@ -38,11 +38,11 @@ type ext_control_addr_error = unit
 
 /* the control address is derived from a non-PC register, e.g. in JALR */
 function ext_control_check_addr(pc : xlenbits) -> Ext_ControlAddr_Check(ext_control_addr_error) =
-  Ext_ControlAddr_OK(pc)
+  Ext_ControlAddr_OK(virtaddr(pc))
 
 /* the control address is derived from the PC register, e.g. in JAL */
 function ext_control_check_pc(pc : xlenbits) -> Ext_ControlAddr_Check(ext_control_addr_error) =
-  Ext_ControlAddr_OK(pc)
+  Ext_ControlAddr_OK(virtaddr(pc))
 
 function ext_handle_control_check_error(err : ext_control_addr_error) -> unit =
   ()
@@ -56,7 +56,7 @@ type ext_data_addr_error = unit
    Extensions might override and add additional checks. */
 function ext_data_get_addr(base : regidx, offset : xlenbits, acc : AccessType(ext_access_type), width : range(1, max_mem_access))
          -> Ext_DataAddr_Check(ext_data_addr_error) =
-  let addr = X(base) + offset in
+  let addr = virtaddr(X(base) + offset) in
   Ext_DataAddr_OK(addr)
 
 function ext_handle_data_check_error(err : ext_data_addr_error) -> unit =
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -86,13 +86,13 @@ function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
       /* "LR faults like a normal load, even though it's in the AMO major opcode space."
         * - Andrew Waterman, isa-dev, 10 Jul 2018.
         */
-      if not(is_aligned(vaddr, width))
+      if not(is_aligned(virtaddr_bits(vaddr), width))
       then { handle_mem_exception(vaddr, E_Load_Addr_Align()); RETIRE_FAIL }
       else match translateAddr(vaddr, Read(Data)) {
         TR_Failure(e, _)    => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
         TR_Address(addr, _) =>
           match mem_read(Read(Data), addr, width_bytes, aq, aq & rl, true) {
-            MemValue(result) => { load_reservation(addr); X(rd) = sign_extend(result); RETIRE_SUCCESS },
+            MemValue(result) => { load_reservation(physaddr_bits(addr)); X(rd) = sign_extend(result); RETIRE_SUCCESS },
             MemException(e)  => { handle_mem_exception(vaddr, e); RETIRE_FAIL }
           },
       }
@@ -131,7 +131,7 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
     match ext_data_get_addr(rs1, zeros(), Write(Data), width_bytes) {
       Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
       Ext_DataAddr_OK(vaddr) => {
-        if not(is_aligned(vaddr, width))
+        if not(is_aligned(virtaddr_bits(vaddr), width))
         then { handle_mem_exception(vaddr, E_SAMO_Addr_Align()); RETIRE_FAIL }
         else {
           match translateAddr(vaddr, Write(Data)) {  /* Write and ReadWrite are equivalent here:
@@ -139,7 +139,7 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
             TR_Failure(e, _) => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
             TR_Address(addr, _) => {
               // Check reservation with physical address.
-              if not(match_reservation(addr)) then {
+              if not(match_reservation(physaddr_bits(addr))) then {
                 /* cannot happen in rmem */
                 X(rd) = zero_extend(0b1); cancel_reservation(); RETIRE_SUCCESS
               } else {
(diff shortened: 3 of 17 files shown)


Timmmm:
This looks great! I'll try and review it when I get a chance.

model/prelude_mem_addrtype.sail
```diff
@@ -0,0 +1,14 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+newtype physaddr = physaddr : xlenbits
```
arichardson, line 9: Does not necessarily need fixing for this PR, but the privileged spec defines physical addresses as 34 bits for RV32.

Timmmm: Yeah I think there's a long-standing TODO for that. Hopefully this PR will make fixing it a little easier!

model/riscv_addr_checks.sail
```diff
@@ -18,7 +18,7 @@ type ext_fetch_addr_error = unit
  *           any address translation errors are reported for pc, not the returned value.
  */
 function ext_fetch_check_pc(start_pc : xlenbits, pc : xlenbits) -> Ext_FetchAddr_Check(ext_fetch_addr_error) =
```
Timmmm, line 20: Should the parameters here be changed to virtaddr instead? And in the other three functions below.

model/riscv_insts_aext.sail
```diff
@@ -86,14 +86,14 @@ function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
       /* "LR faults like a normal load, even though it's in the AMO major opcode space."
         * - Andrew Waterman, isa-dev, 10 Jul 2018.
         */
-      if not(is_aligned(vaddr, width))
-      then { handle_mem_exception(vaddr, E_Load_Addr_Align()); RETIRE_FAIL }
+      if not(is_aligned(virtaddr_bits(vaddr), width))
+      then { handle_mem_exception(virtaddr_bits(vaddr), E_Load_Addr_Align()); RETIRE_FAIL }
```
Timmmm, line 90: I think handle_mem_exception always takes a virtual address, so maybe you can change its parameter to virtaddr?

model/riscv_insts_aext.sail
```diff
@@ -131,27 +131,27 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
     match ext_data_get_addr(rs1, zeros(), Write(Data), width_bytes) {
       Ext_DataAddr_Error(e)  => { ext_handle_data_check_error(e); RETIRE_FAIL },
       Ext_DataAddr_OK(vaddr) => {
-        if not(is_aligned(vaddr, width))
-        then { handle_mem_exception(vaddr, E_SAMO_Addr_Align()); RETIRE_FAIL }
+        if not(is_aligned(virtaddr_bits(vaddr), width))
```
Timmmm, line 134: I would probably turn `is_aligned` into an overload that can take bits(), physaddr or virtaddr. Maybe not in this PR...

model/prelude_mem_addrtype.sail
```diff
@@ -0,0 +1,14 @@
...
+newtype virtaddr = virtaddr : xlenbits
```
Timmmm, line 10: Can you add a note:

> Note virtaddr includes physical addresses when address translation is disabled. It is used where the address would be virtual if address translation is active.

It's slightly confusing otherwise and I don't think we're going to come up with a clearer name (`untranslatedaddr`?...)

Timmmm (changes requested):
Had a skim. Looks pretty good. I think there are some places where I would move the types slightly. In particular I think `handle_mem_exception` should take `virtaddr` and that would greatly reduce the size of the PR too. I haven't checked in detail though so I might be wrong. Can you double check, and if so, make that change and then I'll take another look?

model/riscv_addr_checks.sail
KotorinMinami: Hmm, my thought is that although we know the parameters of these four functions are supposed to be virtual addresses, the naming of this function implies that it should be responsible for validating the incoming register values, which is xlenbits,  through type checks and then converting them into virtual addresses, or redirecting to other error types. Although the current implementation does not achieve this, we should keep this as an unimplemented interface for future expansion

model/prelude_mem_addrtype.sail
arichardson: I find this comment confusing and I believe this is not needed? 

We already have a "no-op" translation from virtaddr to physaddr for the bare case (`Sbare => return TR_Address(physaddr(virtaddr_bits(vAddr)), init_ext_ptw),`) so we can just always start out with a virtual address, but in the SBare case it just happens to be 1:1 mapped to a physical one?

KotorinMinami: However, in chip implementations that do not support virtual addresses, there is no concept of a virtual address, so virtaddr should be interpreted as physaddr, rather than through translation. The Sbare implementation just happens to be compatible with this situation while also being compatible with direct read and write of physical memory in M-mode

model/riscv_insts_aext.sail
KotorinMinami: Thank you for pointing that out, so I won't be tormented by multiple calls in vext_mem, I will fix it.

model/prelude_mem_addrtype.sail
Timmmm: Yeah exactly. Also in `Sbare` mode I think it's natural to think of the address as physical addresses rather than virtual addresses that go through an identity translation.

Maybe my comment didn't explain that very well, but I think we need to have some kind of explanation of this subtlety.

model/riscv_addr_checks.sail
Timmmm: I don't think PC is an architecturally visible register in the same way that `x1` is though - it's kind of an invisible thing that only stores addresses (which aren't necessarily xlenbits).

I haven't made a PR for it yet but in our fork we explicitly handle that via an invalid address conversion function that is called in `set_next_pc()`.

But... it's pretty minor. I don't mind if you leave it.

model/prelude_mem_addrtype.sail
arichardson: If you would like to define it that way, I would prefer to add something like a "cpuaddr" that is explicitly converted to virtaddr/physaddr depending on whether a MMU is present/enabled. That is IMO cleaner than saying virtaddr can be physaddr depending on where it is being used. For a m-mode only CPU I'd be fine with cpuaddr being physaddr, but I still view sbare as a 1:1 mapping rather than addresses being physical when switching off the MMU (since the MMU could be turned on again).

Once hypervisor is implemented we should add intermediate physical addresses that are stage2 translated before being turned into physaddrs, this also means directly using physaddr when the stage1 MMU is disabled is not correct.

model/riscv_insts_base.sail
```diff
@@ -70,11 +70,11 @@ function clause execute (RISCV_JAL(imm, rd)) = {
       ext_handle_control_check_error(e);
       RETIRE_FAIL
     },
-    Ext_ControlAddr_OK(target) => {
+    Ext_ControlAddr_OK(virtaddr(target)) => {
       /* Perform standard alignment check */
       if bit_to_bool(target[1]) & not(extensionEnabled(Ext_Zca))
       then {
-        handle_mem_exception(target, E_Fetch_Addr_Align());
+        handle_mem_exception(virtaddr(target), E_Fetch_Addr_Align());
```
Timmmm, line 77: Hmm it's unfortunate that we have to destructure to `xlenbits` to access the bits and then reconstruct the type. @Alasdair since a newtype is always a known type is there any way we can access the inner type without destructuring it? E.g. in Rust you have this idiom:

```
struct virtaddr(xlenbits);
```

And then you can access the `xlenbits` via `my_virt_addr.0`.

Oh actually @KotorinMinami  I guess we could use your `virtaddr_bits` function like:

```
      /* Perform standard alignment check */
      if bit_to_bool(virtaddr_bits(target)[1]) & not(extensionEnabled(Ext_Zca))
      then {
        handle_mem_exception(target, E_Fetch_Addr_Align());
```

It's a bit verbose though. Hmm.

Timmmm (commented):
The changes look good to me. What do people think about the ergonomics of this?

model/prelude_mem_addrtype.sail
```diff
@@ -0,0 +1,18 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+/* Note: virtaddr includes physical addresses when address translation is disabled.
```
arichardson, line 9: As noted below, I don't see how this can work once physaddr is actually `bits(34)` on RV32. But I guess that can be fixed once this is actually changed.

arichardson (approved):
LGTM, my comments can be addressed once physaddr is actually changed.

model/prelude_mem_addrtype.sail
jrtc27: Well with satp.MODE = Bare you can't address more than XLEN bits of PA space, you have to have translation for that

KotorinMinami:
@Timmmm Could you please review my code again sometime?

Timmmm:
Yeah LGTM - if @arichardson is happy with it then I am too.

I think after this change it should actually be fairly simple to properly support 34-bit physical addresses right? We just have to change it to something like this:

```
type physaddrbits : bits(if xlen == 32 then 34 else 64) // Or maybe better to put it in `riscv_xlen32/64.sail`

newtype physaddr = physaddr : physaddrbits
```

and then fix all the inevitable type errors. What happens if you make that change?

Timmmm (approved):

arichardson (approved):

KotorinMinami:
hmm, this change probably needs some minor tweaking, but it just so happens that this pr splits all the physical and virtual address definitions, so the 34-bit physical address issue shouldn't be too difficult, and I'll probably submit another pr based on this commit once I've done the tweaking properly

KotorinMinami:
Well, it may still be a bit difficult, after I changed the physical address to 34-bits, I found lib/concurrency_interface/read_write.sail in the Sail lib:108-113
```
$iftarget isla
val read_mem# = impure "read_mem" : forall ('a: Type) 'n 'addrsize, 'n >= 0 & 'addrsize in {32, 64}.
    ('a, int('addrsize), bits('addrsize), int('n)) -> bits(8 * 'n)
$else
val read_mem# : forall ('a: Type) 'n 'addrsize, 'n >= 0 & 'addrsize in {32, 64}.
    ('a, int('addrsize), bits('addrsize), int('n)) -> bits(8 * 'n)
```
and lib/concurrency_interface/emulator_memory.sail:151-156
```
             if length(pa) == 32 then {
                Ok((read_mem_exclusive#(request, 32, pa, request.size), tag))
            } else {
                Ok((read_mem_exclusive#(request, 64, pa, request.size), tag))
            }
```
Both list above and other code only accept 32-bit and 64-bit bitvectors, which means we'll probably have to add the corresponding patches upstream?

Timmmm:
Just spoke to Alasdair about this. He suggested that we just always use 64-bits for physical addresses under the hood. I think you need to edit `prelude_mem.sail`'s `read_mem` and `write_mem` functions. Change the `xlenbits` in the `Mem_write_request`/`Mem_read_request` types to `bits(64)` and then also `pa = addr,` to `pa = zero_extend(addr),`. Something like that anyway. Don't do it in this PR though.

Alasdair:
I think you just need to change this:
```
instantiation sail_mem_write with
  'pa = xlenbits,
  pa_bits = xlenbits_identity,
```
You can set the physical address type there to your new type synonym, then the pa_bits function should be a function that extracts the actual bits from the physical address type and extends them to 64 bits.

arichardson:
I'd suggest those changes happen in a separate pr though. Can we go ahead an merge this soon to avoid conflicts?

Alasdair:
Yes, those changes can happen in a separate PR

PR #617, opened 2024-11-16
Use newtypes for register indices and numbers

nwf:
In preparation for more seriously tackling #523.  This PR...

1. Changes the definition of `regidx` and `regno` to be `newtypes` and updates all use sites appropriately.
2. Splits `vregidx` from `regidx` (and friends) for Vector (vext) registers, which, AFAIK, are a completely separate set of 32 registers.
3. Splits `fregidx` from `regidx` (and friends) for F/D/Zcd/Zcf/Zfa/Zfh and V (again) extensions.  There's a little muss and fuss around the definition to do with `Zfinx`, but otherwise it's also pretty rote.
4. Stops some punning in Zicsr and splits the `CSR` `ast` ctor into `CSRImm` and `CSRReg`.

This was a slog.  Most of the changes can be described as "type-directed" and "mechanical" but were all done... er, artisanally, so some more careful review is probably merited.  Perhaps best viewed with word diff, not line diff.

The curious are welcome to look at https://github.com/nwf/sail-riscv/tree/202411-eext to see how I think we can properly support E atop this base, and indeed, most of these changes were motivated or directed by that work, but I would like to get this reviewed independently of that shuffling and further refactoring.

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -96,7 +96,7 @@ function dirty_fd_context_if_present() -> unit = {
   if sys_enable_fdext() then dirty_fd_context()
 }
 
-function rF (r : regno) -> flenbits = {
+function rF (Regno(r) : regno) -> flenbits = {
   assert(sys_enable_fdext());
   let v : fregtype =
     match r {
@@ -137,7 +137,7 @@ function rF (r : regno) -> flenbits = {
   fregval_from_freg(v)
 }
 
-function wF (r : regno, in_v : flenbits) -> unit = {
+function wF (Regno(r) : regno, in_v : flenbits) -> unit = {
   assert(sys_enable_fdext());
   let v = fregval_into_freg(in_v);
   match r {
@@ -184,10 +184,10 @@ function wF (r : regno, in_v : flenbits) -> unit = {
       print_reg("f" ^ dec_str(r) ^ " <- " ^ FRegStr(v));
 }
 
-function rF_bits(i: regidx) -> flenbits = rF(unsigned(i))
+function rF_bits(i: regidx) -> flenbits = rF(regidx_to_regno(i))
 
 function wF_bits(i: regidx, data: flenbits) -> unit = {
-  wF(unsigned(i)) = data
+  wF(regidx_to_regno(i)) = data
 }
 
 overload F = {rF_bits, wF_bits, rF, wF}
@@ -265,8 +265,8 @@ function rF_or_X_D(i) = {
   else if xlen >= 64
   then X(i)[63..0]
   else {
-    assert(i[0] == bitzero);
-    if i == zeros() then zeros() else X(i + 1) @ X(i)
+    assert(regidx_bits(i)[0] == bitzero);
+    if i == zreg then zeros() else X(regidx_offset(i, 0b00001)) @ X(i)
   }
 }
 
@@ -297,10 +297,10 @@ function wF_or_X_D(i, data) = {
   else if xlen >= 64
   then X(i) = sign_extend(data)
   else {
-    assert (i[0] == bitzero);
-    if i != zeros() then {
-      X(i)     = data[31..0];
-      X(i + 1) = data[63..32];
+    assert (regidx_bits(i)[0] == bitzero);
+    if i != zreg then {
+      X(i)                         = data[31..0];
+      X(regidx_offset(i, 0b00001)) = data[63..32];
     }
   }
 }
@@ -311,8 +311,7 @@ overload F_or_X_D = { rF_or_X_D, wF_or_X_D }
 
 /* mappings for assembly */
 
-val freg_name : regidx <-> string
-mapping freg_name = {
+mapping freg_name_raw : bits(5) <-> string = {
     0b00000 <-> "ft0",
     0b00001 <-> "ft1",
     0b00010 <-> "ft2",
@@ -346,6 +345,8 @@ mapping freg_name = {
     0b11110 <-> "ft10",
     0b11111 <-> "ft11"
 }
+
+mapping freg_name : regidx <-> string = { Regidx(i) <-> freg_name_raw(i) }
 
 val freg_or_reg_name : regidx <-> string
 mapping freg_or_reg_name = {
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -52,6 +52,23 @@ function nan_unbox(m, x) = if 'n == 'm then x else (
 /* **************************************************************** */
 /* Floating point register file                                     */
 
+newtype fregidx = Fregidx : bits(5)
+newtype fregno = Fregno : range(0, 31)
+function fregidx_to_fregno (Fregidx(b) : fregidx) -> fregno = Fregno(unsigned(b))
+function fregidx_offset(Fregidx(r) : fregidx, o : bits(5)) -> fregidx = Fregidx(r + o)
+function fregidx_bits(Fregidx(r) : fregidx) -> bits(5) = r
+
+function cregidx_to_fregidx (Cregidx(b) : cregidx) -> fregidx = Fregidx(0b01 @ b)
+
+/*
+ * TODO: This is not quite right for having both Zfinx and E, but at least it
+ * marks the places we need to consider it
+ */
+function fregidx_to_regidx (Fregidx(b) : fregidx) -> regidx =
+  Regidx(truncate(b, let Regidx(zreg_bits) = zreg in length(zreg_bits)))
+
+mapping encdec_freg : fregidx <-> bits(5) = { Fregidx(r) <-> r }
+
 register f0  : fregtype
 register f1  : fregtype
 register f2  : fregtype
@@ -96,7 +113,7 @@ function dirty_fd_context_if_present() -> unit = {
   if sys_enable_fdext() then dirty_fd_context()
 }
 
-function rF (Regno(r) : regno) -> flenbits = {
+function rF (Fregno(r) : fregno) -> flenbits = {
   assert(sys_enable_fdext());
   let v : fregtype =
     match r {
@@ -137,7 +154,7 @@ function rF (Regno(r) : regno) -> flenbits = {
   fregval_from_freg(v)
 }
 
-function wF (Regno(r) : regno, in_v : flenbits) -> unit = {
+function wF (Fregno(r) : fregno, in_v : flenbits) -> unit = {
   assert(sys_enable_fdext());
   let v = fregval_into_freg(in_v);
   match r {
@@ -184,50 +201,50 @@ function wF (Regno(r) : regno, in_v : flenbits) -> unit = {
       print_reg("f" ^ dec_str(r) ^ " <- " ^ FRegStr(v));
 }
 
-function rF_bits(i: regidx) -> flenbits = rF(regidx_to_regno(i))
+function rF_bits(i: fregidx) -> flenbits = rF(fregidx_to_fregno(i))
 
-function wF_bits(i: regidx, data: flenbits) -> unit = {
-  wF(regidx_to_regno(i)) = data
+function wF_bits(i: fregidx, data: flenbits) -> unit = {
+  wF(fregidx_to_fregno(i)) = data
 }
 
 overload F = {rF_bits, wF_bits, rF, wF}
 
-val rF_H : regidx -> bits(16)
+val rF_H : fregidx -> bits(16)
 function rF_H(i) = {
   assert(flen >= 16);
   assert(sys_enable_fdext() & not(sys_enable_zfinx()));
   nan_unbox(F(i))
 }
 
-val wF_H : (regidx, bits(16)) -> unit
+val wF_H : (fregidx, bits(16)) -> unit
 function wF_H(i, data) = {
   assert(flen >= 16);
   assert(sys_enable_fdext() & not(sys_enable_zfinx()));
   F(i) = nan_box(data)
 }
 
-val rF_S : regidx -> bits(32)
+val rF_S : fregidx -> bits(32)
 function rF_S(i) = {
   assert(flen >= 32);
   assert(sys_enable_fdext() & not(sys_enable_zfinx()));
   nan_unbox(F(i))
 }
 
-val wF_S : (regidx, bits(32)) -> unit
+val wF_S : (fregidx, bits(32)) -> unit
 function wF_S(i, data) = {
   assert(flen >= 32);
   assert(sys_enable_fdext() & not(sys_enable_zfinx()));
   F(i) = nan_box(data)
 }
 
-val rF_D : regidx -> bits(64)
+val rF_D : fregidx -> bits(64)
 function rF_D(i) = {
   assert(flen >= 64);
   assert(sys_enable_fdext() & not(sys_enable_zfinx()));
   F(i)
 }
 
-val wF_D : (regidx, bits(64)) -> unit
+val wF_D : (fregidx, bits(64)) -> unit
 function wF_D(i, data) = {
   assert(flen >= 64);
   assert(sys_enable_fdext() & not(sys_enable_zfinx()));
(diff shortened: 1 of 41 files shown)


arichardson (commented):
Not reviewed the full diff in detail but overall I think this is a nice improvement!

Timmmm:
Is there any chance this could be split up into multiple PRs to make it easier to review? E.g. can the "initial support for E" be a separate PR?

nwf:
I suppose, but GitHub also allows review commit-by-commit and that's as fine a granularity as I think makes sense to split it?

jrtc27:
With Zfinx I'm not sure how good an idea this is, it seems to get messy quickly. Also I can't actually see your E support anywhere, the tip of the linked branch is https://github.com/nwf/sail-riscv/commit/359fdd3780c18047515ef509afa9da05780415c1 but that doesn't add E, only factors I out.

Timmmm:
Yeah I think I'm happy with the motivation! I think splitting it up has other advantages:

* More commits can get merged into main, tested by CI, be part of `git bisect` etc.
* It's not all or nothing - hopefully we can get some of it merged before all of it is ready, then rebasing etc. is less of a pain if other people are making simultaneous changes.
* Just a bit easier to keep track and less intimidating IMO.

Could you also split out `Simplify Zicsr ast`? Actually I think we probably shouldn't do that. It *is* simpler and totally in the spirit of how it is supposed to work, and in isolation definitely good. Unfortunately there are two future spanners in the works:

1. CLIC, which adds all sorts of crazy CSRs that act differently depending on whether you use `csrrwi` or `csrrw`.
2. CHERI, which also treats them differently (because `csrrwi/c/s` can't write metadata). I think it's way more reasonable in CHERI's case. CLIC is just ... lightly thought through.

nwf:
OK, I've backed this PR back to 000a17c412b6935c5d2fb6ccd68ef9d526e51c0f and have raised #645 and #646 with the two other commits to facilitate discussion.

nwf:
> With Zfinx I'm not sure how good an idea this is, it seems to get messy quickly. Also I can't actually see your E support anywhere, the tip of the linked branch is [nwf@359fdd3](https://github.com/nwf/sail-riscv/commit/359fdd3780c18047515ef509afa9da05780415c1) but that doesn't add E, only factors I out.

Sorry, I missed a `git add`.  #646 has the files added.

As to the complexity around Zfinx and E, well... I don't think the sail is going to be _that_ hairy, and I think the improved clarity in the type system at all use sites outweighs a little complexity at the one definition site.

arichardson (approved):
I think being explicit here is a nice improvement and it would be good to merge this ASAP so we don't accumulate merge conflicts.

jrtc27:
> I think being explicit here is a nice improvement and it would be good to merge this ASAP so we don't accumulate merge conflicts.

I'm not sold given Zfinx; this PR introduces some ugliness there. By introducing new types for everything it ends up having to use the wrong explicit type when previously they were all the same and so there was no wrong type.

nwf:
I'm open to suggestions, but the status quo seems somewhat challenging for E support, especially E machines that reuse the resulting voids in the encoding space, without just forking the whole model.

IMHO Zfinx being a model run-time decision is a deliberate type confusion, and it'd have been better for it to be a model compile-time decision, especially since it cannot (AFAICT) be changed after initialization.  It's not surprising, then, that it results in some violence to the types.  In particular, I think it is acceptable to the specification to have an implementation of E and F and not-Zfinx that has 16 GPRs and 32 FPU registers, which means that `regidx` and `fregidx` really ought to be separable newtypes (perhaps of the same underlying type, of course).

Would you, @jrtc27, accept the first commit, or first two commits, in this PR on its own, without the subsequent refinement of (`vregidx` and) `fregidx` as a types-as-documentation improvement, even if you disagree with the approach of "statically E with a smaller `regidx`" that motivated this series of PRs?

arichardson:
> > I think being explicit here is a nice improvement and it would be good to merge this ASAP so we don't accumulate merge conflicts.
> 
> I'm not sold given Zfinx; this PR introduces some ugliness there. By introducing new types for everything it ends up having to use the wrong explicit type when previously they were all the same and so there was no wrong type.

Can't we solve the zfinx problem by adding an explicit conversion? IMO you're still using a freg index in the instruction, it just happens to be mapped to a for internally?

I also think it's odd this is a runtime option an IMO would be cleaner as compile time.

nwf:
> Can't we solve the zfinx problem by adding an explicit conversion? IMO you're still using a freg index in the instruction, it just happens to be mapped to a for internally?

Yeah, `fregidx`-en are decoded (`encdec_freg`) and passed through the model until they get to the Zfinx-aware register accessors, `[rw]F_or_X_[FDH]()`, which use `fregidx_to_regidx()` if they need something to pass to `X()`.

> I also think it's odd this is a runtime option an IMO would be cleaner as compile time.

A future PR, perhaps.

jrtc27:
> > > I think being explicit here is a nice improvement and it would be good to merge this ASAP so we don't accumulate merge conflicts.
> > 
> > 
> > I'm not sold given Zfinx; this PR introduces some ugliness there. By introducing new types for everything it ends up having to use the wrong explicit type when previously they were all the same and so there was no wrong type.
> 
> Can't we solve the zfinx problem by adding an explicit conversion? IMO you're still using a freg index in the instruction, it just happens to be mapped to a for internally?

No you're not, you're using an X register, there are no F registers.

> I also think it's odd this is a runtime option an IMO would be cleaner as compile time.

The whole intent is to reduce compile-time options and have as much configurable at run time as possible. Having something like Zfinx be compile-time option would completely go against that.

arichardson:
> > > > I think being explicit here is a nice improvement and it would be good to merge this ASAP so we don't accumulate merge conflicts.
> > > 
> > > 
> > > I'm not sold given Zfinx; this PR introduces some ugliness there. By introducing new types for everything it ends up having to use the wrong explicit type when previously they were all the same and so there was no wrong type.
> > 
> > 
> > Can't we solve the zfinx problem by adding an explicit conversion? IMO you're still using a freg index in the instruction, it just happens to be mapped to a for internally?
> 
> No you're not, you're using an X register, there are no F registers.

That's true, but we can still have a type abstraction for "register index of a floating point value"? For RV32 Zdinx we could even assert that it's always an even index when converting it to a GPR index?

> 
> > I also think it's odd this is a runtime option an IMO would be cleaner as compile time.
> 
> The whole intent is to reduce compile-time options and have as much configurable at run time as possible. Having something like Zfinx be compile-time option would completely go against that.


Reading the spec there is no way to dynamically switch this behaviour at runtime since it seems to be gated on the extension being implemented:
> In the standard privileged architecture defined in Volume II, the mstatus field FS is hardwired to 0 if the Zfinx extension is implemented, and FS no longer affects the trapping behavior of floating-point instructions or fcsr accesses.

> The misa bits F, D, and Q are hardwired to 0 when the Zfinx extension is implemented.


But that point I don't really care about, I just think being explicit about what type of register we are indexing is still useful.

nwf:
Well, it looks like we now (again) have tons of conflicting files.  Is it worth dusting this off for merge or would it just languish again?

arichardson:
> Well, it looks like we now (again) have tons of conflicting files. Is it worth dusting this off for merge or would it just languish again?

I think it would be worthwhile, but I can't guarantee it will be merged :(

jordancarlin:
> > Well, it looks like we now (again) have tons of conflicting files. Is it worth dusting this off for merge or would it just languish again?
> 
> I think it would be worthwhile, but I can't guarantee it will be merged :(

We will eventually need to support E, and I haven’t seen any other proposals for how to do that, so it sort of seems like something along these lines is necessary. Even without that aspect, the additional type safety seems like a good thing and is a direction the model is generally moving in (see the recent virtual/physical address new types). So I’d be in favor of it. If you’re concerned about doing the work only for it not to get merged, maybe we can hold off and discuss it at the next weekly meeting to see if we can come to some sort of consensus on if we want to go through with this.

nwf:
To be clear, I am not adamant that _this_ PR merge, and if someone has a better proposal, please, go for it.  But the lack of structured support for E has bitten CHERIoT already, in that our currently in-tree "throw an exception on invalid register access" approach is subtly wrong (because the model does not model instructions with transactions and sail's sequential/monadic behavior is such that state effects survive thrown exceptions, so the CHERIoT model can, for example, write to a CSR before trapping on a GPR access) and more proper fixes are unpleasant, holographic, and eternal divergences from the upstream model, such as changing all arms of the en-/de-coding definition to have additional `0 @`.

pmundkur:
@nwf  We discussed this PR at the golden-model meeting, and the consensus was that the three commits in this PR were worth merging despite the localized increase in complexity for Zfinx.  We can address the other PRs to support E once this is in.   If your patience has been (understandably) exhausted, I can help with getting the patches updated.

nwf:
OK, `N`th time's the charm, I hope!

- I've done the rebase (and renamed the data constructors to be initial-capital as per [Alasdair's comment on #646](https://github.com/riscv/sail-riscv/pull/646#issuecomment-2694822390)).
- These all build and pass the repo's `ctest` tests.

arichardson (approved):

model/riscv_insts_dext.sail
```diff
@@ -251,25 +254,25 @@ function validDoubleRegs(n, regs) = {
...
 
-union clause ast = F_MADD_TYPE_D : (regidx, regidx, regidx, rounding_mode, regidx, f_madd_op_D)
+union clause ast = F_MADD_TYPE_D : (fregidx, fregidx, fregidx, rounding_mode, fregidx, f_madd_op_D)
 
 /* AST <-> Binary encoding ================================ */
 
 mapping clause encdec =
     F_MADD_TYPE_D(rs3, rs2, rs1, rm, rd, FMADD_D)                         if haveDoubleFPU() & validDoubleRegs([rs3, rs2, rs1, rd])
-<-> rs3 @ 0b01 @ rs2 @ rs1 @ encdec_rounding_mode (rm) @ rd @ 0b100_0011  if haveDoubleFPU() & validDoubleRegs([rs3, rs2, rs1, rd])
+<-> encdec_freg(rs3) @ 0b01 @ encdec_freg(rs2) @ encdec_freg(rs1) @ encdec_rounding_mode (rm) @ encdec_freg(rd) @ 0b100_0011 if haveDoubleFPU() & validDoubleRegs([rs3, rs2, rs1, rd])
```
pmundkur, line 263: It would be good to preserve the  alignment of the `if` conditionals on either side of `<->` to help visually ensure the conditions are the same, even though it does add an extra line to the diff.

pmundkur (commented):
Minor alignment nit but otherwise LGTM.

model/riscv_fdext_regs.sail
```diff
@@ -265,8 +265,8 @@ function rF_or_X_D(i) = {
   else if xlen >= 64
   then X(i)[63..0]
   else {
-    assert(i[0] == bitzero);
-    if i == zeros() then zeros() else X(i + 1) @ X(i)
+    assert((let Regidx(i_bits) = i in i_bits)[0] == bitzero);
```
Timmmm: For `physaddr`/`virtaddr` we made a `physaddr_bits()` function.

```
function physaddr_bits(physaddr(paddr) : physaddr) -> physaddrbits = paddr

function virtaddr_bits(virtaddr(vaddr) : virtaddr) -> xlenbits = vaddr
```

Though I'm wondering it if makes sense to have a `bits_of` override function for all three cases?

We could do that in a new PR I guess. I think add a `regidx_bits` function though at least?

model/riscv_fdext_regs.sail
```diff
@@ -265,8 +265,8 @@ function rF_or_X_D(i) = {
...
+    if i == zreg then zeros() else X(regidx_offset(i, 0b00001)) @ X(i)
```
Timmmm: Similarly, we have some operator overloads for `physaddr`:

```
function sub_virtaddr_xlenbits(virtaddr(addr) : virtaddr, offset : xlenbits) -> virtaddr = virtaddr(addr - offset)

overload operator - = { sub_virtaddr_xlenbits }
```

Timmmm (commented):
LGTM - some minor suggestions. I think we should add `to_bits()` later.

model/riscv_fdext_regs.sail
nwf: `regidx_bits` is beyond my knowledge of how to do in `sail`, because it's type, `... regidx -> bits(N)` involves the `N` from the `bits(N)` within the definition of the `regidx` `newtype`.

model/riscv_fdext_regs.sail
nwf: If you'd prefer to eliminate the `regidx_offset` in favor of an overload, can I ask that it be done in a later PR?

model/riscv_fdext_regs.sail
```diff
@@ -347,11 +364,23 @@ mapping freg_name = {
...
-  reg if sys_enable_zfinx() <-> reg_name(reg)  if sys_enable_zfinx()
-}
+mapping freg_name : fregidx <-> string = { fregidx(i) <-> freg_name_raw(i) }
+
+val freg_or_reg_name : fregidx <-> string
+function freg_or_reg_name_forwards_matches(_ : fregidx) -> bool = true
+function freg_or_reg_name_forwards(f : fregidx) -> string =
+  if      sys_enable_fdext() then freg_name(f)
+  else if sys_enable_zfinx() then reg_name(fregidx_to_regidx(f))
+  else ""
```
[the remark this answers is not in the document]
nwf: Those didn't, I think, exist when I first wrote this.  Sure; will figure out how.

model/riscv_fdext_regs.sail
Alasdair: I think it would have the type `regidx -> {'n, 'in in {4, 5}. bits('n)}`, which you could read as `regidx` to some bitvector that has either length 4 or 5. It can't give a specific length because that's hidden inside regidx.

model/riscv_fdext_regs.sail
Alasdair: I agree it makes sense to do that in a follow-up PR

model/riscv_fdext_regs.sail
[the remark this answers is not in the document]
Alasdair: Without checking super carefully, I beleive the following is equivalent to what you have:
```
val freg_or_reg_name : regidx <-> string

mapping freg_or_reg_name = {
  forwards f if sys_enable_fdext() => freg_name(f)
  forwards f if sys_enable_zfinx() => reg_name(fregidx_to_regidx(f))
  forwards f => "",
  backwards s if sys_enable_fdext() & freg_name_backwards_matches(s) => freg_name_backwards(s),
  backwards s if sys_enable_zfinx() & reg_name_backwards_matches(s) =>
    let Regidx(i) = reg_name_backwards(s) in Fregidx(zero_extend(i)),
}
```
Not sure we need the 3rd case with `=> ""`?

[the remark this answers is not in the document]
Alasdair: Thinking a bit more, maybe we can just take the `fdext` behaviour as the default when dealing with floating point registers and only use conditions for the Zfinx case.
```
mapping freg_or_reg_name = {
  forwards f if sys_enable_zfinx() => reg_name(fregidx_to_regidx(f),
  backwards reg_name(Regidx(i)) if sys_enable_zfinx() => Fregidx(zero_extend(i)),
  freg_name(f) <-> f
```
if that works it seems like the cleanest way?

model/riscv_insts_dext.sail
nwf: That was only inconsistently the case beforehand, but I think I have improved it.

model/riscv_fdext_regs.sail
Timmmm: Isn't `Regidx` always 5 bits currently? Or are you thinking of a future definition?

model/riscv_fdext_regs.sail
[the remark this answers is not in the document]
nwf: Looks like that works great; thank you!

model/riscv_fdext_regs.sail
nwf: Hm.  I was thinking of the I-or-E case, I think.  Indeed, for either always-I or always-E, it's trivial and I've introduced `regidx_bits` and used it where appropriate (I think).

Timmmm (approved):
LGTM. Let's get this in before any more nightmare rebases!

Timmmm:
Actually before we merge this I just wanted to double check one thing (which I think I've probably asked like 3 times, sorry):

@Alasdair will we be able to do something like this?

```
newtype regidx = Regidx : if config base_isa.e then bits(4) else bits(5)
```

Timmmm:
Actually I guess it would be

```
newtype regidx = Regidx : bits(if config base_isa.e then 4 else 5)
```

Alasdair:
You would need:
```
type have_e : Bool = config base_isa.e
```
But otherwise yes

pmundkur (approved):
LGTM

Issue #593, opened 2024-10-14
Various issues around clearing the bottom bits of xepc


Timmmm:
Basically we should remove `pc_alignment_mask()`:

```
function pc_alignment_mask() -> xlenbits =
  ~(zero_extend(if misa[C] == 0b1 then 0b00 else 0b10))
```

The logic here is kind of hard to follow but it returns `0b1111..1111` if C is enabled, otherwise `0b1111...11101`. That is weird. I think we got away with it because in all the places it is used the bottom bit is already cleared, but that isn't necessarily the case when we implement CLIC.

It's also weird that it is applied kind of ad-hoc after reading `xepc`, rather than just having a `get_xepc()` function that applies the read legalisation.

There is actually a `get_xpec` type function `get_xret_target()`. In our code I've renamed it to `get_xepc()` (clearer, and for CHERI/CLIC reasons). Then we have

```
function get_xepc(p : Privilege) -> xlenbits =
  match p {
    Machine    => get_mepc(),
    Supervisor => get_sepc(),
    User       => get_uepc(),
  }

...

function get_mepc() -> xlenbits = align_pc(mepc)  // This gets overridden for CHERI.

...

function align_pc(addr : xlenbits) -> xlenbits =
  if misa[C] == 0b1
  then [addr with 0 = bitzero]
  else [addr with 1..0 = zeros()]
```

This code also needs fixing:

```
function legalize_xepc(v : xlenbits) -> xlenbits =
  /* allow writing xepc[1] only if misa.C is enabled or could be enabled
     XXX specification says this legalization should be done on read */
  if   (sys_enable_writable_misa() & sys_enable_rvc()) | misa[C] == 0b1
  then [v with 0 = bitzero]
  else v & sign_extend(0b100)
```

The `XXX` comment is talking about this exact same issue. There are two issues:

1. Weird mix of `[with]` and `&`.
2. `sys_enable_writable_misa()` doesn't matter here.

It should be:

```
function legalize_xepc(v : xlenbits) -> xlenbits = {
  // allow writing xepc[1] only if misa.C is enabled or could be enabled.
  if   sys_enable_rvc()
  then [v with 0 = bitzero]
  else [v with 1..0 = zeros()]
}
```

rmn30:
I agree that the calculation of `pc_alignment_mask` is incorrect but we do still need it.

> 3.1.14. Machine Exception Program Counter (mepc)
[...]

PR #618, opened 2024-11-19
Remove pc alignment mask

rez5427:
#593

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -293,7 +293,7 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       if   get_config_print_platform()
       then print_platform("ret-ing from " ^ to_str(prev_priv) ^ " to " ^ to_str(cur_privilege));
 
-      prepare_xret_target(Machine) & pc_alignment_mask()
+      align_pc(prepare_xret_target(Machine))
     },
     (_, CTL_SRET())  => {
       let prev_priv   = cur_privilege;
@@ -310,7 +310,7 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       then print_platform("ret-ing from " ^ to_str(prev_priv)
                           ^ " to " ^ to_str(cur_privilege));
 
-      prepare_xret_target(Supervisor) & pc_alignment_mask()
+      align_pc(prepare_xret_target(Supervisor))
     },
   }
 }
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -293,7 +293,7 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       if   get_config_print_platform()
       then print_platform("ret-ing from " ^ to_str(prev_priv) ^ " to " ^ to_str(cur_privilege));
 
-      align_pc(prepare_xret_target(Machine))
+      prepare_xret_target(Machine)
     },
     (_, CTL_SRET())  => {
       let prev_priv   = cur_privilege;
@@ -310,7 +310,7 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       then print_platform("ret-ing from " ^ to_str(prev_priv)
                           ^ " to " ^ to_str(cur_privilege));
 
-      align_pc(prepare_xret_target(Supervisor))
+      prepare_xret_target(Supervisor)
     },
   }
 }
--- a/model/riscv_sys_exceptions.sail
+++ b/model/riscv_sys_exceptions.sail
@@ -89,9 +89,9 @@ function clause is_CSR_defined(0x305) = true // mtvec
 function clause is_CSR_defined(0x341) = true // mepc
 
 function clause read_CSR(0x105) = get_stvec()
-function clause read_CSR(0x141) = align_pc(get_xepc(Supervisor))
+function clause read_CSR(0x141) = get_xepc(Supervisor)
 function clause read_CSR(0x305) = get_mtvec()
-function clause read_CSR(0x341) = align_pc(get_xepc(Machine))
+function clause read_CSR(0x341) = get_xepc(Machine)
 
 function clause write_CSR(0x105, value) = { set_stvec(value) }
 function clause write_CSR(0x141, value) = { set_xepc(Supervisor, value) }
--- a/model/riscv_sys_exceptions.sail
+++ b/model/riscv_sys_exceptions.sail
@@ -32,8 +32,8 @@ function prepare_trap_vector(p : Privilege, cause : Mcause) -> xlenbits = {
 
 /* xRET handling involves three functions:
  *
- * get_xepc:     used to read the value of the xret target  (no control flow transfer)
- * set_xepc:     used to write a value of the xret target   (no control flow transfer)
+ * get_xepc:            used to read the value of the xret target  (no control flow transfer)
+ * set_xepc:            used to write a value of the xret target   (no control flow transfer)
  * prepare_xret_target: used to get the value for control transfer to the xret target
  */
 
--- a/model/riscv_sys_exceptions.sail
+++ b/model/riscv_sys_exceptions.sail
@@ -40,8 +40,8 @@ function prepare_trap_vector(p : Privilege, cause : Mcause) -> xlenbits = {
 val get_xepc : Privilege -> xlenbits
 function get_xepc(p) =
   match p {
-    Machine    => mepc,
-    Supervisor => sepc,
+    Machine    => align_pc(mepc),
+    Supervisor => align_pc(sepc),
     User       => internal_error(__FILE__, __LINE__, "Invalid privilege level"),
   }
 
--- a/model/riscv_sys_exceptions.sail
+++ b/model/riscv_sys_exceptions.sail
@@ -32,33 +32,32 @@ function prepare_trap_vector(p : Privilege, cause : Mcause) -> xlenbits = {
 
 /* xRET handling involves three functions:
  *
- * get_xret_target:     used to read the value of the xret target  (no control flow transfer)
- * set_xret_target:     used to write a value of the xret target   (no control flow transfer)
+ * get_xepc:     used to read the value of the xret target  (no control flow transfer)
+ * set_xepc:     used to write a value of the xret target   (no control flow transfer)
  * prepare_xret_target: used to get the value for control transfer to the xret target
  */
 
-val get_xret_target : Privilege -> xlenbits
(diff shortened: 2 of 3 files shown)


rmn30:
There is a bear trap here. From the RISC-V privileged spec:

> mepc is an MXLEN-bit read/write register formatted as shown in Figure 21. The low bit of mepc
(mepc[0]) is always zero. On implementations that support only IALIGN=32, the two low bits
(mepc[1:0]) are always zero.
If an implementation allows IALIGN to be either 16 or 32 (by changing CSR misa, for example), then,
whenever IALIGN=32, bit **mepc[1] is masked on reads so that it appears to be 0. This masking occurs
also for the implicit read by the MRET instruction. Though masked, mepc[1] remains writable when
IALIGN=32**.

(emphasis mine)

So the legalization that occurs on write is not necessarily the same as the one for the implicit read in MRET. Please check whether we're getting this right.

model/riscv_sys_control.sail
```diff
@@ -293,7 +293,7 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       if   get_config_print_platform()
       then print_platform("ret-ing from " ^ to_str(prev_priv) ^ " to " ^ to_str(cur_privilege));
 
-      prepare_xret_target(Machine) & pc_alignment_mask()
+      legalize_xepc(prepare_xret_target(Machine))
```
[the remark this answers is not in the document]
rmn30: As per comment below I think this was correct previously except for the calculation of `pc_alignment_mask` which should be something like `sign_extend(if misa.C then 0b110 else 0b100)`.

model/riscv_sys_control.sail
```diff
@@ -310,7 +310,7 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       then print_platform("ret-ing from " ^ to_str(prev_priv)
                           ^ " to " ^ to_str(cur_privilege));
 
-      prepare_xret_target(Supervisor) & pc_alignment_mask()
+      prepare_xret_target(Supervisor) & sign_extend(if misa[C] then 0b110 else 0b100)
```
rmn30, line 313: Please don't repeat this in two places. You use @Timmmm 's `align_pc` function from #593 .

model/riscv_sys_exceptions.sail
```diff
@@ -89,11 +89,11 @@ function clause is_CSR_defined(0x305) = true // mtvec
 function clause is_CSR_defined(0x341) = true // mepc
 
 function clause read_CSR(0x105) = get_stvec()
-function clause read_CSR(0x141) = get_xret_target(Supervisor) & pc_alignment_mask()
+function clause read_CSR(0x141) = get_xepc(Supervisor)
```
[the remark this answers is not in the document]
jrtc27: I believe that's true, though the slightly less weird sequence that's definitely true is:

```
write mepc <- 0b11
read mepc -> 0b10
write misa.C <- 0
read mepc -> 0b00
write misa.C <- 1
read mepc -> 0b10
```

That alone requires masking on read, something which is explicitly mentioned in the privileged spec.

jrtc27:
I don't know _what_ you've done to cause it, but your commit history is an inscrutable mess. Please clean it up per the second paragraph in https://github.com/riscv/sail-riscv/blob/master/CONTRIBUTING.md (and just because it's the expectation for contributing to *any* project).

model/riscv_sys_control.sail
jrtc27: Is there a reason one should not just reuse legalize_xepc? Perhaps because we want to capture the fact that the LSB isn't ever set in this case, it's only the second bit that needs handling specially?

jrtc27:
This is getting worse not better; you're piling on more commits to an already mess of a history, you're committing some random test log file, you're adding in debug prints and you're removing more code that needs to be there.

rez5427:
> This is getting worse not better; you're piling on more commits to an already mess of a history, you're committing some random test log file, you're adding in debug prints and you're removing more code that needs to be there.

sorry... I'm try to remove the mess commit.... but it didn't work

jrtc27:
> > This is getting worse not better; you're piling on more commits to an already mess of a history, you're committing some random test log file, you're adding in debug prints and you're removing more code that needs to be there.
> 
> sorry... I'm try to remove the mess commit.... but it didn't work Is there a way to remove these commit in a pushed pr. I've learn how to better commit, It won't happen again.. really sorry..

Look up how to use git rebase, e.g. https://about.gitlab.com/blog/2020/11/23/keep-git-history-clean-with-interactive-rebase/ looks like it could be a decent introduction

rez5427:
> > > This is getting worse not better; you're piling on more commits to an already mess of a history, you're committing some random test log file, you're adding in debug prints and you're removing more code that needs to be there.
> > 
> > 
> > sorry... I'm try to remove the mess commit.... but it didn't work Is there a way to remove these commit in a pushed pr. I've learn how to better commit, It won't happen again.. really sorry..
> 
> Look up how to use git rebase, e.g. https://about.gitlab.com/blog/2020/11/23/keep-git-history-clean-with-interactive-rebase/ looks like it could be a decent introduction

Copy that Sir

rmn30:
It's worth pointing out that after a rebase you will have to use `git push -f` to force update the branch for this PR. That goes slightly against what the article Jessica linked says, namely "should NOT be used on commit history that has already been pushed", but it is pretty normal for PRs prior to merging. It can cause slight pain for anyone who has pulled the PR branch and wants to update but it's possible to recover.

rez5427:
> So I got permission to push most of our Sail code in one big go, to support some CHERI stuff (long story)... anyway the point is you can see what we've done [here](https://github.com/CHERI-Alliance/sail-riscv/blob/e9b80042b2f881c045615259dd58c8fcb61b0aa8/model/riscv_sys_regs.sail#L691-L707):
> 
> ```
> /* The xepc legalization zeroes xepc[1:0] when misa.C is hardwired to 0.
>  * When misa.C is writable, it zeroes only xepc[0].
>  */
> function legalize_xepc(v : xlenbits) -> xlenbits = {
>   // allow writing xepc[1] only if misa.C is enabled or could be enabled.
>   if   sys_enable_rvc()
>   then [v with 0 = bitzero]
>   else [v with 1..0 = zeros()]
> }
> 
> // Align value to min supported PC alignment. This is used to
> // legalize xepc reads.
> function align_pc(addr : xlenbits) -> xlenbits = {
>   if misa[C] == 0b1
>   then [addr with 0 = bitzero]
>   else [addr with 1..0 = zeros()]
> }
> ```
> 
> And elsewhere:
> 
> ```
> function get_mepc() -> xlenbits = align_pc(mepc)
> 
> function set_mepc(value : xlenbits) -> unit = {
>   mepc = legalize_xepc(convertInvalidAddr(value));
> }
> 
> function get_sepc() -> xlenbits = align_pc(sepc)
> 
> function set_sepc(value : xlenbits) -> unit = {
>   sepc = legalize_xepc(convertInvalidAddr(value));
> }
> 
> ...
> 
> function clause read_CSR(0x341, _) = get_mepc()
> 
> function clause write_CSR(0x341, value) = { set_mepc(value); get_mepc() }
> 
> ...
> 
> function get_xepc(p : Privilege) -> xlenbits =
>   match p {
>     Machine    => get_mepc(),
>     Supervisor => get_sepc(),
>     User       => internal_error(__FILE__, __LINE__, "N extension not supported"),
>   }
> 
> ...
> 
> function prepare_xret_target(p : Privilege) -> xlenbits =
>   if inSmclicMode() & get_xcause(p)[Xinhv] == 0b1
>   then prepare_clic_xret_vectored_target(p)
>   else get_xepc(p)
> ```
> 
> Ignore my earlier comment about removing `prepare_xret_target`; let's leave it in so when we add CLIC it's easier.
> 
> Also it was maybe overkill having _both_ `get_mepc` and `get_xepc`. I would just have `get_xepc` I think. IIRC it kind ended up that way because of the pain of handling CHERI CSRs.

I'm not sure. Is the new commit ok?

Timmmm:
Yes looks good to me with one change: move the `align_pc()` calls into `get_xepc()`, since that legalisation is always done when reading it.

```
val get_xepc : Privilege -> xlenbits
function get_xepc(p) =
  match p {
    Machine    => align_pc(mepc),
    Supervisor => align_pc(sepc),
    User       => internal_error(__FILE__, __LINE__, "Invalid privilege level"),
  }
```

(Or you can put the `align_pc` around the whole `match` but I think it makes it slightly less readable; up to you.)

model/riscv_sys_exceptions.sail
```diff
@@ -89,11 +89,11 @@ function clause is_CSR_defined(0x305) = true // mtvec
 function clause is_CSR_defined(0x341) = true // mepc
 
 function clause read_CSR(0x105) = get_stvec()
-function clause read_CSR(0x141) = get_xret_target(Supervisor) & pc_alignment_mask()
+function clause read_CSR(0x141) = align_pc(get_xepc(Supervisor))
```
Timmmm, line 92: Oh sorry you also don't need `align_pc` here, or on line 94.

Timmmm (approved):
LGTM

Timmmm:
@rmn30 @jrtc27 does this look ok to you?

rmn30:
Yes, I think it looks fine but the commit history is a bit messy still. Could use squash merge but would need to update PR with proper description first.

Timmmm:
Ok cool, I will merge this in a few days with a nice squashed commit if nobody objects. Thanks for reviewing!

jrtc27:
I'll take a look over the weekend

model/riscv_sys_exceptions.sail
```diff
@@ -32,21 +32,21 @@ function prepare_trap_vector(p : Privilege, cause : Mcause) -> xlenbits = {
 
 /* xRET handling involves three functions:
  *
- * get_xret_target:     used to read the value of the xret target  (no control flow transfer)
- * set_xret_target:     used to write a value of the xret target   (no control flow transfer)
+ * get_xepc:     used to read the value of the xret target  (no control flow transfer)
+ * set_xepc:     used to write a value of the xret target   (no control flow transfer)
  * prepare_xret_target: used to get the value for control transfer to the xret target
```
jrtc27, line 37:
```suggestion
 * get_xepc:            used to read the value of the xret target  (no control flow transfer)
 * set_xepc:            used to write a value of the xret target   (no control flow transfer)
 * prepare_xret_target: used to get the value for control transfer to the xret target
```

jrtc27 (commented):
Minor comment otherwise I think the change is correct

PR #619, opened 2024-11-21
Fix mstatus legalisation of SD bit

Timmmm:
Unfortunately ded340f268e3ca46ac35e0d7ec11bec3c318c9ea subtly broke the mstatus legalisation. It needs to calculate FS/XS bits before it calculates SD as it's the OR of the others.


Timmmm:
@jordancarlin please could you review this?

jordancarlin (approved):
I’m surprised I didn’t catch this while testing #600. This looks like a clean solution to something that is definitely an issue and should be merged asap.

PR #620, opened 2024-11-21
fix: Print MTI value upon hart update for machine timer interrupt visibility

Ali-Faraz-10xe:
Hi!

I have developed the ACTs (Architectural Compliance Tests) and coverage points for [RISC-V compliance verification](https://github.com/riscv-non-isa/riscv-arch-test). As part of the coverage points, I needed support for printing the MTI (Machine Timer Interrupt) value upon a hart update to improve the visibility and tracking of machine interrupts during the test execution.

To achieve this, I made a small change to one of the print functions, ensuring that the MTI value is printed whenever the hart state is updated. This change is essential for accurate compliance verification and enhances the observability of machine interrupts.

Please review and consider adding this change to the original repository for inclusion in the RISC-V architecture test suite.

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -212,7 +212,6 @@ function clint_load(t, addr, width) = {
 function clint_dispatch() -> unit = {
   if   get_config_print_platform()
   then print_platform("clint::tick mtime <- " ^ BitStr(mtime));
-<<<<<<< HEAD
   mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);
   
   if get_config_print_platform()
@@ -224,15 +223,7 @@ function clint_dispatch() -> unit = {
     
   if get_config_print_platform()
   then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.STI <- " ^ BitStr(mip[STI]) ^ ")");
-  }; 
-=======
-  mip[MTI] = 0b0;
-  if mtimecmp <=_u mtime then {
-    mip[MTI] = 0b1;
-    if   get_config_print_platform()
-    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
-  }
->>>>>>> 25748bd9 (remove problematic  else condition)
+  };
 }
 
 /* The rreg effect is due to checking mtime. */
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -211,14 +211,12 @@ function clint_load(t, addr, width) = {
 
 function clint_dispatch() -> unit = {
   mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);
-  if get_config_print_platform()
-  then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
-  /* Sstc - supervisor timer register */
   if extensionEnabled(Ext_Sstc) then {
     mip[STI] = bool_to_bits(stimecmp <=_u mtime); 
-    if get_config_print_platform()
-    then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.STI <- " ^ BitStr(mip[STI]) ^ ")");
   };
+  if get_config_print_platform()
+  then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^
+    (if extensionEnabled(Ext_Sstc) then ", mip.STI <- " ^ BitStr(mip[STI]) else "") ^ ")");
 }
 
 /* The rreg effect is due to checking mtime. */
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -209,7 +209,6 @@ function clint_load(t, addr, width) = {
   }
 }
 
-
 function clint_dispatch() -> unit = {
   mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);
   if get_config_print_platform()
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -209,20 +209,16 @@ function clint_load(t, addr, width) = {
   }
 }
 
+
 function clint_dispatch() -> unit = {
-  if   get_config_print_platform()
-  then print_platform("clint::tick mtime <- " ^ BitStr(mtime));
   mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);
-  
   if get_config_print_platform()
   then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
-  
   /* Sstc - supervisor timer register */
   if extensionEnabled(Ext_Sstc) then {
-  mip[STI] = bool_to_bits(stimecmp <=_u mtime);
-    
-  if get_config_print_platform()
-  then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.STI <- " ^ BitStr(mip[STI]) ^ ")");
+    mip[STI] = bool_to_bits(stimecmp <=_u mtime); 
+    if get_config_print_platform()
+    then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.STI <- " ^ BitStr(mip[STI]) ^ ")");
   };
 }
 
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -215,7 +215,7 @@ function clint_dispatch() -> unit = {
     mip[STI] = bool_to_bits(stimecmp <=_u mtime); 
   };
   if get_config_print_platform()
-  then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^
+  then print_platform("clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^
     (if extensionEnabled(Ext_Sstc) then ", mip.STI <- " ^ BitStr(mip[STI]) else "") ^ ")");
 }
 
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -212,21 +212,18 @@ function clint_load(t, addr, width) = {
 function clint_dispatch() -> unit = {
   if   get_config_print_platform()
   then print_platform("clint::tick mtime <- " ^ BitStr(mtime));
-  mip[MTI] = 0b0;
-  if mtimecmp <=_u mtime then {
-    if   get_config_print_platform()
-    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime));
-    mip[MTI] = 0b1;
-  };
+  mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);
+  
+  if get_config_print_platform()
+  then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
+  
   /* Sstc - supervisor timer register */
   if extensionEnabled(Ext_Sstc) then {
-    mip[STI] = 0b0;
-    if stimecmp <=_u mtime then {
-      if   get_config_print_platform()
-      then print_platform(" supervisor timer pending at mtime " ^ BitStr(mtime));
-      mip[STI] = 0b1;
-    }
-  };
+  mip[STI] = bool_to_bits(stimecmp <=_u mtime);
+    
+  if get_config_print_platform()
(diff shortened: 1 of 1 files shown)


model/riscv_platform.sail
```diff
@@ -209,9 +209,14 @@ function clint_dispatch() -> unit = {
   then print_platform("clint::tick mtime <- " ^ BitStr(mtime));
   mip[MTI] = 0b0;
   if mtimecmp <=_u mtime then {
+    mip[MTI] = 0b1;
```
Timmmm, line 212: Can delete this line.

model/riscv_platform.sail
```diff
@@ -209,9 +209,14 @@ function clint_dispatch() -> unit = {
...
     if   get_config_print_platform()
-    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime));
-    mip[MTI] = 0b1
+    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
+  }
+  else if mtimecmp >_u mtime then {
```
Timmmm, line 216: Can just be `else` no?

model/riscv_platform.sail
```diff
@@ -209,9 +209,14 @@ function clint_dispatch() -> unit = {
...
+    mip[MTI] = 0b0;
+    if   get_config_print_platform()
+    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
```
Timmmm, line 219: Message is wrong

Timmmm (changes requested):
This doesn't seem to change functionality... But there's no way we're going to guarantee that this message output is stable.

Tests shouldn't be parsing log output. However since we haven't got a proper solution yet and this is a pretty small change it's probably ok for now. Just don't expect it to stay like this.

model/riscv_platform.sail
```diff
@@ -209,9 +209,14 @@ function clint_dispatch() -> unit = {
...
   mip[MTI] = 0b0;
   if mtimecmp <=_u mtime then {
+    mip[MTI] = 0b1;
     if   get_config_print_platform()
-    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime));
-    mip[MTI] = 0b1
+    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
+  }
+  else if mtimecmp >_u mtime then {
+    mip[MTI] = 0b0;
```
arichardson, line 217: This is redundant due to the assignment above. Either change this to an `else` and delete the assignment above or drop this assignment.

allenjbaum:
Let me see if I've got this straight. hart state gets updated at every
cycle (even a noop updates the PC.
so, does this message get printed every cycle? The count rate is then a
constant 1/cycle, and track mcycle. That's legal.
But I'm not sure how you can easily write a test where the reference model
(e.g. Sail) updates at a different rate than the DUT
 - which is inevitably true since the rate is implementation dependent.
We could pass a cycle divider number into sail, so it only counts on every
n-th cycle, but that still isn't perfect.

Test don't parse the log output, as such - but tools (e.g. ISAC) rely on it.

Ali-Faraz-10xe:
Thank you all for your detailed reviews and concerns. I've found a simpler solution that addresses the concerns:

1. I tested by removing the else condition entirely, since MTI=0 is already set at the start:

```
function clint_dispatch() -> unit = {
  then print_platform("clint::tick mtime <- " ^ BitStr(mtime));
  mip[MTI] = 0b0;
  if mtimecmp <=_u mtime then {
    mip[MTI] = 0b1;
    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
  }
}
```
This simpler version:

- Removes **redundant** code 

- **Simplifies** the logic 

- **Only prints** when timer is actually pending 

- Still achieves 100% coverage for my test cases

I've verified this gives me the coverage I need while making the code cleaner. Would this be a better approach?

Timmmm:
That appears to be very similar to the existing code no? I think that's fine, but why do you need to change it at all?

Also you can hardcode "mip.MTI <- 1" in the message.

MuhammadHammad001:
> Also you can hardcode "mip.MTI <- 1" in the message.

But, IMO the value of mip.MTI needs to be updated using `mip[MTI] = 0b1;` for architectural state maintenance. (If you are saying to update it and then simply use 1 instead of the variable in the print statement, then, I guess I misinterpreted it:) )


I think the reason for updating the log message for mip.MTI value is to have a similar output format as of the mip.MSI [here](https://github.com/riscv/sail-riscv/blob/07fa23ebd7a42a357f8912a870879e23a3f3695d/model/riscv_platform.sail#L223C1-L228C1)
```
    then print_platform("clint[" ^ BitStr(addr) ^ "] <- " ^ BitStr(data) ^ " (mip.MSI <- " ^ BitStr(data[0]) ^ ")");
    mip[MSI] = [data[0]];
    clint_dispatch();
    MemValue(true)
```
To explain it a bit, whenever we have a software interrupt pending (by writing a value on the base address of MSIP), we see the following message in the log:
```
mem[X,0x8000029C] -> 0xA023
mem[X,0x8000029E] -> 0x01E7
[140] [M]: 0x8000029C (0x01E7A023) sw t5, 0x0(a5)
clint[0x00000000] <- 0x00000001 (mip.MSI <- 0b1)
clint::tick mtime <- 0x0000000000000001
```
But, for the case of timer interrupt pending which we can only be brought by making the mtime>= mtimecmp, we just simply see the following code, while on the back end (by the hart) the value of mip.MTI is updated to 1 which is not visible to the user:
```
mem[X,0x80000298] -> 0x2023
mem[X,0x8000029A] -> 0x01E8
[139] [M]: 0x80000298 (0x01E82023) sw t5, 0x0(a6)
clint<4>[0x0000BFF8] <- 0x0000FFFF (mtime)
clint::tick mtime <- 0x000000000000FFFF
 clint timer pending at mtime 0x000000000000FFFF
```
Therefore, updating the log message will change it to the following and will make the updated mip.MTI visible to the user as in the case of software interrupt:
```
mem[X,0x80000298] -> 0x2023
mem[X,0x8000029A] -> 0x01E8
[139] [M]: 0x80000298 (0x01E82023) sw t5, 0x0(a6)
clint<4>[0x0000BFF8] <- 0x0000FFFF (mtime)
clint::tick mtime <- 0x000000000000FFFF
clint timer pending at mtime 0x000000000000FFFF (mip.MTI <- 0b1)
```

MuhammadHammad001:
Additionally, as soon as we write something on the memory mapped address MSIP, we see the reflected value updated in the mip.MSi in the log something like:
```
mem[X,0x800008FC] -> 0xA023
mem[X,0x800008FE] -> 0x0002
[203] [M]: 0x800008FC (0x0002A023) sw zero, 0x0(t0)
clint[0x00000000] <- 0x00000000 (mip.MSI <- 0b0)
clint::tick mtime <- 0x0000000000000002
``` 
Since, the value of mip.MTI depends on the mtime/mtimecmp update and IMO mip.MTI should also be visible on every mtime update by using the logic something similar to:
```
function clint_dispatch() -> unit = {
  if mtimecmp <=_u mtime then {
    mip[MTI] = 0b1;
    then print_platform(" clint timer pending at mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")");
  }
  else{
  mip[MTI] = 0b0;
  then print_platform("clint::tick mtime <- " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^ ")" );
}
}
```

But, that's just a suggestion.

Timmmm:
So as I understand it you just want to change the logging. How about this?

```
function clint_dispatch() -> unit = {
  mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);

  then print_platform(" clint mtime " ^ bits_str(mtime) ^ ": mip.MTI <- " ^ bits_str(mip[MTI]));
}
```

Much simpler.

model/riscv_platform.sail
```diff
@@ -212,20 +212,17 @@ function clint_load(t, addr, width) = {
...
+  
   /* Sstc - supervisor timer register */
   if extensionEnabled(Ext_Sstc) then {
-    mip[STI] = 0b0;
-    if stimecmp <=_u mtime then {
-      if   get_config_print_platform()
-      then print_platform(" supervisor timer pending at mtime " ^ BitStr(mtime));
-      mip[STI] = 0b1;
-    }
+  mip[STI] = bool_to_bits(stimecmp <=_u mtime);
```
Timmmm, line 222: Needs to be indented.

model/riscv_platform.sail
```diff
@@ -212,20 +212,17 @@ function clint_load(t, addr, width) = {
 function clint_dispatch() -> unit = {
   if   get_config_print_platform()
   then print_platform("clint::tick mtime <- " ^ BitStr(mtime));
```
Timmmm, line 214: This print is redundant now.

Timmmm (changes requested):

model/riscv_platform.sail
```diff
@@ -209,23 +209,16 @@ function clint_load(t, addr, width) = {
   }
 }
 
+
```
arichardson, line 212:
```suggestion
```
unnecessary newline

arichardson (approved):

model/riscv_platform.sail
```diff
@@ -210,22 +210,14 @@ function clint_load(t, addr, width) = {
...
   if extensionEnabled(Ext_Sstc) then {
-    mip[STI] = 0b0;
-    if stimecmp <=_u mtime then {
-      if   get_config_print_platform()
-      then print_platform(" supervisor timer pending at mtime " ^ BitStr(mtime));
-      mip[STI] = 0b1;
-    }
+    mip[STI] = bool_to_bits(stimecmp <=_u mtime); 
+    if get_config_print_platform()
+    then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.STI <- " ^ BitStr(mip[STI]) ^ ")");
```
jrtc27, line 220: Why do these have a leading space? And why is this the same string as above?

Timmmm: It's not exactly the same - it prints STI instead of MTI.

But I think having two prints is needlessly verbose. @Ali-Faraz-10xe can you change it to this?

```
function clint_dispatch() -> unit = {
  mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);
  if extensionEnabled(Ext_Sstc) then {
    mip[STI] = bool_to_bits(stimecmp <=_u mtime); 
  }
  
  if get_config_print_platform()
  then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^
    (if extensionEnabled(Ext_Sstc) then ", mip.STI <- " ^ BitStr(mip[STI]) else "") ^ ")");
}
```

jrtc27: I still don't understand the desire to have the leading space though

Timmmm: Ah yeah I think that made sense with the output before when it was a separate line after the `clint::tick` one, but we can remove it now.

model/riscv_platform.sail
```diff
@@ -210,23 +210,13 @@ function clint_load(t, addr, width) = {
...
-    mip[STI] = 0b0;
-    if stimecmp <=_u mtime then {
-      if   get_config_print_platform()
-      then print_platform(" supervisor timer pending at mtime " ^ BitStr(mtime));
-      mip[STI] = 0b1;
-    }
+    mip[STI] = bool_to_bits(stimecmp <=_u mtime); 
   };
+  if get_config_print_platform()
+  then print_platform(" clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^
```
Timmmm, line 218:
```suggestion
  then print_platform("clint mtime " ^ BitStr(mtime) ^ " (mip.MTI <- " ^ BitStr(mip[MTI]) ^
```

Timmmm (approved):
LGTM - I'll squash and merge in a few days if nobody objects.

model/riscv_platform.sail
```diff
@@ -210,23 +210,13 @@ function clint_load(t, addr, width) = {
...
-  /* Sstc - supervisor timer register */
+  mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);
   if extensionEnabled(Ext_Sstc) then {
-    mip[STI] = 0b0;
-    if stimecmp <=_u mtime then {
-      if   get_config_print_platform()
-      then print_platform(" supervisor timer pending at mtime " ^ BitStr(mtime));
-      mip[STI] = 0b1;
-    }
+    mip[STI] = bool_to_bits(stimecmp <=_u mtime);
```
Timmmm, line 215:
```suggestion
    mip[STI] = bool_to_bits(stimecmp <=_u mtime);
```

Issue #326, opened 2023-10-26
mie, mip bits should be read-only without supervisor mode


Timmmm:
This code:

```
function legalize_mip(o : Minterrupts, v : xlenbits) -> Minterrupts = {
  /* The only writable bits are the S-mode bits, and with the 'N'
   * extension, the U-mode bits. */
  let v = Mk_Minterrupts(v);
  let m = update_SEI(o, v.SEI());
  let m = update_STI(m, v.STI());
  let m = update_SSI(m, v.SSI());
  if haveUsrMode() & haveNExt() then {
    let m = update_UEI(m, v.UEI());
    let m = update_UTI(m, v.UTI());
    let m = update_USI(m, v.USI());
    m
  } else m
}

function legalize_mie(o : Minterrupts, v : xlenbits) -> Minterrupts = {
  let v = Mk_Minterrupts(v);
  let m = update_MEI(o, v.MEI());
  let m = update_MTI(m, v.MTI());
  let m = update_MSI(m, v.MSI());
  let m = update_SEI(m, v.SEI());
  let m = update_STI(m, v.STI());
  let m = update_SSI(m, v.SSI());
  /* The U-mode bits will be modified if we have the 'N' extension. */
  if haveUsrMode() & haveNExt() then {
    let m = update_UEI(m, v.UEI());
    let m = update_UTI(m, v.UTI());
    let m = update_USI(m, v.USI());
    m
  } else m
}
```

should be this (not tested):

```
function legalize_mip(o : Minterrupts, v : xlenbits) -> Minterrupts = {
  /* The only writable bits are the S-mode bits, and with the 'N'
   * extension, the U-mode bits. */
  let v = Mk_Minterrupts(v);
  let m = if haveSupMode() then {
    let m = update_SEI(o, v.SEI());
    let m = update_STI(m, v.STI());
    let m = update_SSI(m, v.SSI());
  } else o;
  if haveUsrMode() & haveNExt() then {
    let m = update_UEI(m, v.UEI());
    let m = update_UTI(m, v.UTI());
    let m = update_USI(m, v.USI());
    m
  } else m
}

function legalize_mie(o : Minterrupts, v : xlenbits) -> Minterrupts = {
  let v = Mk_Minterrupts(v);
  let m = update_MEI(o, v.MEI());
  let m = update_MTI(m, v.MTI());
  let m = update_MSI(m, v.MSI());
  let m = if haveSupMode() then {
    let m = update_SEI(m, v.SEI());
[...]

PR #621, opened 2024-11-22
Fix bits SEIP, STIP, and SSIP of mip and SEIE, STIE, and SSIE of mie read-only problem

KotorinMinami:
As stated in #326, the spec mentions that
>If supervisor mode is not implemented, bits SEIP, STIP, and SSIP of mip and SEIE, STIE, and SSIE of mie are read-only zeros.

Hence the change.


Timmmm:
Lgtm thanks! Very minor nit but can you add trailing commas?

Timmmm:
Yep, exactly. Thanks!

Timmmm (approved):

jordancarlin (approved):
LGTM

Issue #553, opened 2024-09-20
Add >=0 constraints to quot_round_zero and rem_round_zero


Timmmm:
The type declarations for those division operators are missing `>0` and `>=0` constraints.

See https://github.com/riscv/sail-riscv/pull/552/files#r1768223964

Alasdair:
I had a go at doing this. It should be possible, but there are a few places where we will need to add explicit asserts for divisors, even if it looks somewhat obvious:
```
  if EMUL_pow < 0 then nf / (2 ^ (0 - EMUL_pow)) <= 8
```
This is never division by zero, but it requires reasoning about non-linear arithmetic to prove it which the SMT solver can't do.

Timmmm:
Hmm isn't the output of `2 ^` always greater than zero? That does feel obvious...

Alasdair:
Yes, but in practice the SMT solver just gives up whenever it sees `2 ^ n` and `n` isn't bounded. I have a fallback routine that replaces `2 ^` with an arbitrary function that is equivalent up to `2 ^ 64` so I can check any problem that is unsatisfiable for such a function must be unsatisfiable in general, and I could improve that fallback to handle this case but it doesn't work without an explicit assert right now.

Alasdair:
That does mean the other solution is to really understand the vector spec and add tight bounds to every variable.

Timmmm:
I think this was fixed in #1082

PR #622, opened 2024-11-24
Remove vector of bools in vext

### LINKED ISSUE #554: Use bit vectors for read_vmask and read_vmask_carry
There are some vector functions that return bit masks as a vector of bools, rather than a `bits()`. This makes the code unnecessarily complex.

See https://github.com/riscv/sail-riscv/pull/552#discussion_r1768236101

rez5427:
fix #554
emmm...
I will test the code with riscv-vector-tests later...
And to make sure `assert('n > 0);` is right

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -118,6 +118,17 @@ function bits_to_bool(x : bits(1)) -> bool = bool_bits(x)
 val to_bits : forall 'l, 'l >= 0.(int('l), int) -> bits('l)
 function to_bits (l, n) = get_slice_int(l, n, 0)
 
+val flip_bits : forall 'n, 'n >= 0. (int('n), bits('n)) -> bits('n)
+function flip_bits (num : int('n), x : bits('n)) -> bits('n) = {
+  var result : bits('n) = zeros();
+  foreach (i from 0 to (num - 1))
+    if x[i] == bitone then
+      result[i] = bitzero
+    else
+      result[i] = bitone;
+  result
+}
+
 infix 4 <_s
 infix 4 >_s
 infix 4 <=_s
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -118,17 +118,6 @@ function bits_to_bool(x : bits(1)) -> bool = bool_bits(x)
 val to_bits : forall 'l, 'l >= 0.(int('l), int) -> bits('l)
 function to_bits (l, n) = get_slice_int(l, n, 0)
 
-val flip_bits : forall 'n, 'n >= 0. (int('n), bits('n)) -> bits('n)
-function flip_bits (num : int('n), x : bits('n)) -> bits('n) = {
-  var result : bits('n) = zeros();
-  foreach (i from 0 to (num - 1))
-    if x[i] == bitone then
-      result[i] = bitzero
-    else
-      result[i] = bitone;
-  result
-}
-
 infix 4 <_s
 infix 4 >_s
 infix 4 <=_s
--- a/model/riscv_insts_vext_mask.sail
+++ b/model/riscv_insts_vext_mask.sail
@@ -290,8 +290,6 @@ function clause execute(VIOTA_M(vm, vs2, vd)) = {
   let 'n = num_elem;
   let 'm = SEW;
 
-  assert('n > 0);
-
   let vm_val  : bits('n)     = read_vmask(num_elem, vm, 0b00000);
   let vs2_val : bits('n)     = read_vmask(num_elem, 0b0, vs2);
   let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
@@ -330,8 +328,6 @@ function clause execute(VID_V(vm, vd)) = {
 
   let 'n = num_elem;
   let 'm = SEW;
-
-  assert('n > 0);
 
   let vm_val  : bits('n)     = read_vmask(num_elem, vm, 0b00000);
   let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
--- a/model/riscv_insts_vext_mask.sail
+++ b/model/riscv_insts_vext_mask.sail
@@ -50,11 +50,11 @@ function clause execute(MMTYPE(funct6, vs2, vs1, vd)) = {
       result[i] = match funct6 {
         MM_VMAND     => (vs2_val & vs1_val)[i],
         MM_VMNAND    => if (vs2_val & vs1_val)[i] == bitone then bitzero else bitone,
-        MM_VMANDN    => (vs2_val & flip_bits(num_elem, vs1_val))[i],
+        MM_VMANDN    => (vs2_val & (vs1_val ^ ones()))[i],
         MM_VMXOR     => (vs2_val ^ vs1_val)[i],
         MM_VMOR      => (vs2_val | vs1_val)[i],
-        MM_VMNOR     => flip_bits(num_elem, vs2_val | vs1_val)[i],
-        MM_VMORN     => (vs2_val | flip_bits(num_elem, vs1_val))[i],
+        MM_VMNOR     => ((vs2_val | vs1_val) ^ ones())[i],
+        MM_VMORN     => (vs2_val | (vs1_val ^ ones()))[i],
         MM_VMXNOR    => (vs2_val & vs1_val)[i]
       }
     }
@@ -256,7 +256,7 @@ function clause execute(VMSOF_M(vm, vs2, vd)) = {
   var found_elem : bits('n) = zeros();
   foreach (i from 0 to (num_elem - 1)) {
     if mask[i] == bitone then {
-      if (vs2_val & flip_bits('n, found_elem))[i] == bitone then {
+      if (vs2_val & (found_elem ^ ones()))[i] == bitone then {
         result[i] = bitone;
         found_elem = sign_extend(0b1);
       } else {
--- a/model/riscv_insts_vext_mask.sail
+++ b/model/riscv_insts_vext_mask.sail
@@ -50,11 +50,11 @@ function clause execute(MMTYPE(funct6, vs2, vs1, vd)) = {
       result[i] = match funct6 {
         MM_VMAND     => (vs2_val & vs1_val)[i],
         MM_VMNAND    => ~(vs2_val & vs1_val)[i],
-        MM_VMANDN    => (vs2_val & (vs1_val ^ ones()))[i],
+        MM_VMANDN    => (vs2_val & ~(vs1_val))[i],
         MM_VMXOR     => (vs2_val ^ vs1_val)[i],
         MM_VMOR      => (vs2_val | vs1_val)[i],
-        MM_VMNOR     => ((vs2_val | vs1_val) ^ ones())[i],
-        MM_VMORN     => (vs2_val | (vs1_val ^ ones()))[i],
+        MM_VMNOR     => (~(vs2_val | vs1_val))[i],
+        MM_VMORN     => (vs2_val | ~(vs1_val))[i],
         MM_VMXNOR    => (vs2_val & vs1_val)[i]
       }
     }
@@ -177,7 +177,7 @@ function clause execute(VMSBF_M(vm, vs2, vd)) = {
   foreach (i from 0 to (num_elem - 1)) {
     if mask[i] == bitone then {
       if vs2_val[i] == bitone then found_elem = true;
-      result[i] = if found_elem then bitzero else bitone
+      result[i] = bool_to_bit(not(found_elem))
     }
   };
 
@@ -216,7 +216,7 @@ function clause execute(VMSIF_M(vm, vs2, vd)) = {
   var found_elem : bool = false;
   foreach (i from 0 to (num_elem - 1)) {
     if mask[i] == bitone then {
-      result[i] = if found_elem then bitzero else bitone;
+      result[i] = bool_to_bit(not(found_elem));
       if vs2_val[i] == bitone then found_elem = true
     }
   };
--- a/model/riscv_insts_vext_mask.sail
+++ b/model/riscv_insts_vext_mask.sail
@@ -38,24 +38,24 @@ function clause execute(MMTYPE(funct6, vs2, vs1, vd)) = {
   let 'n = num_elem;
   let 'm = SEW;
 
-  let vs1_val : vector('n, bool) = read_vmask(num_elem, 0b0, vs1);
-  let vs2_val : vector('n, bool) = read_vmask(num_elem, 0b0, vs2);
-  let vd_val  : vector('n, bool) = read_vmask(num_elem, 0b0, vd);
+  let vs1_val : bits('n) = read_vmask(num_elem, 0b0, vs1);
+  let vs2_val : bits('n) = read_vmask(num_elem, 0b0, vs2);
+  let vd_val  : bits('n) = read_vmask(num_elem, 0b0, vd);
 
   let (initial_result, mask) = init_masked_result_carry(num_elem, SEW, 0, vd_val);
   var result = initial_result;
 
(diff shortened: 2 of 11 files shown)


rez5427:
The 'n produced in here, and I believe it can't be 0.
/* num_elem means max(VLMAX,VLEN/SEW)) according to Section 5.4 of RVV spec */
val get_num_elem : (int, int) -> nat
function get_num_elem(LMUL_pow, SEW) = {
  /* Ignore lmul < 1 so that the entire vreg is read, allowing all masking to
   * be handled in init_masked_result */
  let num_elem = 2 ^ (LMUL_pow_reg) * VLEN / SEW;
  assert(num_elem > 0);
  num_elem
}

And the result of this commit run with riscv-vector-tests is in [here](https://github.com/rez5427/rvv-tests-result/blob/main/all_test_results.csv)
https://github.com/rez5427/rvv-tests-result/blob/main/all_test_results.csv

model/prelude.sail
```diff
@@ -118,6 +118,17 @@ function bits_to_bool(x : bits(1)) -> bool = bool_bits(x)
 val to_bits : forall 'l, 'l >= 0.(int('l), int) -> bits('l)
 function to_bits (l, n) = get_slice_int(l, n, 0)
 
+val flip_bits : forall 'n, 'n >= 0. (int('n), bits('n)) -> bits('n)
+function flip_bits (num : int('n), x : bits('n)) -> bits('n) = {
```
arichardson, line 122: It seems like you are always flipping all bits? Can't you just xor with ones()?

If this function is needed with a partial flip, you can simplify that to an xor with `ones(n) @ zeros()`

model/riscv_insts_vext_mask.sail
```diff
@@ -38,24 +38,24 @@ function clause execute(MMTYPE(funct6, vs2, vs1, vd)) = {
...
-        MM_VMAND     => vs2_val[i] & vs1_val[i],
-        MM_VMNAND    => not(vs2_val[i] & vs1_val[i]),
-        MM_VMANDN    => vs2_val[i] & not(vs1_val[i]),
-        MM_VMXOR     => vs2_val[i] != vs1_val[i],
-        MM_VMOR      => vs2_val[i] | vs1_val[i],
-        MM_VMNOR     => not(vs2_val[i] | vs1_val[i]),
-        MM_VMORN     => vs2_val[i] | not(vs1_val[i]),
-        MM_VMXNOR    => vs2_val[i] == vs1_val[i]
+        MM_VMAND     => (vs2_val & vs1_val)[i],
+        MM_VMNAND    => if (vs2_val & vs1_val)[i] == bitone then bitzero else bitone,
```
arichardson, line 52: I feel like it would be cleaner to add a `not()` overload that works on `bits('n)`. That also makes the operator precedence more obvious. @Timmmm / @Alasdair what do you think?

Timmmm: I think the `~()` function is already overridden for `bit` so `=> ~((vs2_val & vs1_val)[i])` should work.

arichardson: That's even better 👍

Timmmm: Also since `mask` is now a bit vector we don't need to do this bit by bit. We can do the whole `result = vs2_val & vs1_val & mask` (or whatever) in one go.

I did start working on that a while ago however it is a ton of code and I ran out of steam, so I would suggest we explicitly do *not* fix all that in this PR - we can do it in a future one.

model/riscv_insts_vext_vm.sail
```diff
@@ -689,7 +719,7 @@ function clause execute(VICMPTYPE(funct6, vm, vs2, simm, vd)) = {
         VICMP_VMSGTU   => unsigned(vs2_val[i]) > unsigned(imm_val),
         VICMP_VMSGT    => signed(vs2_val[i]) > signed(imm_val)
       };
-      result[i] = res
+      result[i] = if res then bitone else bitzero
```
arichardson, line 722:
```suggestion
      result[i] = bool_to_bit(res)
```

model/riscv_vext_regs.sail
```diff
@@ -359,51 +359,51 @@ function write_single_element(EEW, index, vrid, value) = {
 }
 
 /* Mask register reading operation with num_elem as max(VLMAX,VLEN/SEW)) */
-val read_vmask : forall 'n, 'n >= 0. (int('n), bits(1), regidx) -> vector('n, bool)
+val read_vmask : forall 'n, 'n > 0. (int('n), bits(1), regidx) -> bits('n)
 function read_vmask(num_elem, vm, vrid) = {
   assert(num_elem <= sizeof(vlenmax));
   let vreg_val : vregtype = V(vrid);
-  var result   : vector('n, bool) = vector_init(true);
+  var result   : bits('n) = sign_extend(0b1);
```
arichardson, line 366:
```suggestion
  var result   : bits('n) = ones();
```

model/riscv_insts_vext_vm.sail
```diff
@@ -626,9 +653,10 @@ function clause execute(VIMSTYPE(funct6, vs2, simm, vd)) = {
   var result = initial_result;
 
   foreach (i from 0 to (num_elem - 1)) {
-    if mask[i] then {
+    if mask[i] == bitone then {
+      let vm_bit : bool = if vm_val[i] == bitone then true else false;
       result[i] = match funct6 {
-        VIMS_VADC     => to_bits(SEW, unsigned(vs2_val[i]) + unsigned(imm_val) + unsigned(bool_to_bits(vm_val[i])))
+        VIMS_VADC     => to_bits(SEW, unsigned(vs2_val[i]) + unsigned(imm_val) + unsigned(bool_to_bits(vm_bit)))
```
arichardson, line 659:
```suggestion
        VIMS_VADC     => to_bits(SEW, unsigned(vs2_val[i]) + unsigned(imm_val) + unsigned(vm_val[i]))
```
Can't we just do this and drop the `let vm_bit`?

model/riscv_insts_vext_vm.sail
```diff
@@ -508,20 +528,23 @@ function clause execute(VIMTYPE(funct6, vs2, simm, vd)) = {
...
   let (initial_result, mask) = init_masked_result_carry(num_elem, SEW, LMUL_pow, vd_val);
   var result = initial_result;
 
   foreach (i from 0 to (num_elem - 1)) {
-    if mask[i] then {
+    if mask[i] == bitone then {
+      let vm_bit : bool = if vm_val[i] == bitone then true else false;
       let res : bool = match funct6 {
-        VIM_VMADC    => unsigned(vs2_val[i]) + unsigned(imm_val) + unsigned(bool_to_bits(vm_val[i])) > 2 ^ SEW - 1
+        VIM_VMADC    => unsigned(vs2_val[i]) + unsigned(imm_val) + unsigned(bool_to_bits(vm_bit)) > 2 ^ SEW - 1
```
arichardson, line 545: same as below

model/riscv_insts_vext_vm.sail
```diff
@@ -462,7 +482,7 @@ function clause execute(VXCMPTYPE(funct6, vm, vs2, rs1, vd)) = {
         VXCMP_VMSGTU   => unsigned(vs2_val[i]) > unsigned(rs1_val),
         VXCMP_VMSGT    => signed(vs2_val[i]) > signed(rs1_val)
       };
-      result[i] = res
+      result[i] = if res then bitone else bitzero
```
arichardson, line 485:
```suggestion
      result[i] = bool_to_bit(res)
```

model/riscv_insts_vext_vm.sail
```diff
@@ -393,10 +410,11 @@ function clause execute(VXMSTYPE(funct6, vs2, rs1, vd)) = {
   var result = initial_result;
 
   foreach (i from 0 to (num_elem - 1)) {
-    if mask[i] then {
+    if mask[i] == bitone then {
+      let vm_bit : bool = if vm_val[i] == bitone then true else false;
       result[i] = match funct6 {
-        VXMS_VADC     => to_bits(SEW, unsigned(vs2_val[i]) + unsigned(rs1_val) + unsigned(bool_to_bits(vm_val[i]))),
-        VXMS_VSBC     => to_bits(SEW, unsigned(vs2_val[i]) - unsigned(rs1_val) - unsigned(bool_to_bits(vm_val[i])))
+        VXMS_VADC     => to_bits(SEW, unsigned(vs2_val[i]) + unsigned(rs1_val) + unsigned(bool_to_bits(vm_bit))),
```
arichardson, line 416: same as below

model/riscv_insts_vext_vm.sail
```diff
@@ -378,13 +393,15 @@ function clause execute(VXMSTYPE(funct6, vs2, rs1, vd)) = {
   let 'n = num_elem;
   let 'm = SEW;
 
+  assert('n > 0);
+
   /* for bypassing normal masking in init_masked_result */
-  var vec_trues : vector('n, bool) = undefined;
+  var vec_trues : bits('n) = undefined;
```
arichardson, line 399: Can't this just be `let vec_trues : bits('n) = ones();`?

model/riscv_insts_vext_vm.sail
rez5427: I change it to `(if vm_val[i] == bitone then 1 else 0)`, and I'm not sure is it ok or not.

Timmmm:
But I would recommend *not* trying to fix the types in this PR! I started going down that route and you end up rewriting half the vector code and never finishing (I have also done the same for the virtual memory code, though that is a less code so I've almost finished). Let's use asserts temporarily and then we can gradually fix the types and remove the asserts.

rez5427:
I change the def below, change 'n >= 0 to 'n > 0. Because it all use get_num_elem to produce it.
```
val process_rfvv_widen: forall 'n 'm 'p, 'n > 0 & 'm in {8, 16, 32, 64}. (rfvvfunct6, bits(1), regidx, regidx, regidx, int('n), int('m), int('p)) -> Retired
function process_rfvv_widen(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_pow) = {
```
```
function clause execute(RFVVTYPE(funct6, vm, vs2, vs1, vd)) = {
  let SEW      = get_sew();
  let num_elem_vs = get_num_elem(LMUL_pow, SEW);

  if funct6 == FVV_VFWREDOSUM | funct6 == FVV_VFWREDUSUM then
    process_rfvv_widen(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_pow)
  else
    process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_pow)
}
```

model/riscv_insts_vext_mask.sail
```diff
@@ -38,24 +38,24 @@ function clause execute(MMTYPE(funct6, vs2, vs1, vd)) = {
...
-        MM_VMNAND    => not(vs2_val[i] & vs1_val[i]),
-        MM_VMANDN    => vs2_val[i] & not(vs1_val[i]),
-        MM_VMXOR     => vs2_val[i] != vs1_val[i],
-        MM_VMOR      => vs2_val[i] | vs1_val[i],
-        MM_VMNOR     => not(vs2_val[i] | vs1_val[i]),
-        MM_VMORN     => vs2_val[i] | not(vs1_val[i]),
-        MM_VMXNOR    => vs2_val[i] == vs1_val[i]
+        MM_VMAND     => (vs2_val & vs1_val)[i],
+        MM_VMNAND    => ~(vs2_val & vs1_val)[i],
+        MM_VMANDN    => (vs2_val & (vs1_val ^ ones()))[i],
```
arichardson, line 53: Use ~ here?

arichardson: This whole block could be simplified by hoisting the full bits operation outside of the foreach. But that cleanup can be done later.
I'd suggest using ~ here instead of cor with ones though.

model/riscv_insts_vext_mask.sail
```diff
@@ -166,18 +166,18 @@ function clause execute(VMSBF_M(vm, vs2, vd)) = {
...
   var result = initial_result;
 
   var found_elem : bool = false;
   foreach (i from 0 to (num_elem - 1)) {
-    if mask[i] then {
-      if vs2_val[i] then found_elem = true;
-      result[i] = if found_elem then false else true
+    if mask[i] == bitone then {
+      if vs2_val[i] == bitone then found_elem = true;
+      result[i] = if found_elem then bitzero else bitone
```
arichardson, line 180: bool_to_bit

arichardson: Well with not()

model/riscv_insts_vext_mask.sail
```diff
@@ -206,18 +206,18 @@ function clause execute(VMSIF_M(vm, vs2, vd)) = {
...
   let (initial_result, mask) = init_masked_result_cmp(num_elem, SEW, 0, vd_val, vm_val);
   var result = initial_result;
 
   var found_elem : bool = false;
   foreach (i from 0 to (num_elem - 1)) {
-    if mask[i] then {
-      result[i] = if found_elem then false else true;
-      if vs2_val[i] then found_elem = true
+    if mask[i] == bitone then {
+      result[i] = if found_elem then bitzero else bitone;
```
arichardson, line 219: bool_to_bit(not(found_elem))

model/riscv_insts_vext_vm.sail
```diff
@@ -35,21 +35,21 @@ function clause execute(VVMTYPE(funct6, vs2, vs1, vd)) = {
...
   let (initial_result, mask) = init_masked_result_carry(num_elem, SEW, LMUL_pow, vd_val);
   var result = initial_result;
 
   foreach (i from 0 to (num_elem - 1)) {
-    if mask[i] then {
+    if mask[i] == bitone then {
       let res : bool = match funct6 {
-        VVM_VMADC    => unsigned(vs2_val[i]) + unsigned(vs1_val[i]) + unsigned(bool_to_bits(vm_val[i])) > 2 ^ SEW - 1,
-        VVM_VMSBC    => unsigned(vs2_val[i]) - unsigned(vs1_val[i]) - unsigned(bool_to_bits(vm_val[i])) < 0
+        VVM_VMADC    => unsigned(vs2_val[i]) + unsigned(vs1_val[i]) + (if vm_val[i] == bitone then 1 else 0) > 2 ^ SEW - 1,
```
arichardson, line 49: Can't this just be unsigned(vm_val[I])?

arichardson (commented):
This generally looks good to me. There are a few more simplifications that I've pointed out (commented on the first instance only but there are often multiple).

Further cleanup can be done in a follow up

model/riscv_insts_vext_vm.sail
rez5427: I remember the last time I tried with unsigned(vm_val[I]), and unsigned take bits not bit, also tried with to_bits,but it also not take type bit. sorry for not mention that...
```
val unsigned = pure {
  ocaml: "uint",
  lem: "uint",
  interpreter: "uint",
  coq: "uint",
  lean: "BitVec.toNat",
  _: "sail_unsigned"
} : forall 'n. bits('n) -> range(0, 2 ^ 'n - 1)
```

arichardson (approved):

rez5427:
About this issue https://github.com/riscv/sail-riscv/issues/553.
Should `assert('n > 0);` be back, and add >= constraints?

Timmmm:
I think we should just merge this first and then try to fix the `get_num_elem` and division constraints in a future PR. There are so many types to fix in the vector code we definitely need to do a little bit at a time as much as possible.

Timmmm (approved):

Timmmm:
I'll give this a week or two before merging in case anyone wants to chime in since it's such a big change (in terms of lines of code anyway) and we have zero vector tests currently.

PR #623, opened 2024-11-26
Make mstatus[VS] dirty when write to vstart

### LINKED ISSUE #569: Vector Status flag mstatus[VS] is not updated correctly
The spec says:

> The VS field encodes the status of the vector extension state, including the vector registers v0–v31 and the CSRs vcsr, vxrm, vxsat, vstart, vl, vtype, and vlenb.

However currently VS is only dirtied when you write to `vcsr`. I think a call to `dirty_v_context_if_present()` is missing for most of the other CSRs.

Also I'm not sure about `dirty_v_context_if_present()` since it seem to be checking whether vector is *supported* which is always the case if we're modifying those CSRs so it's at best redundant, and at worst should it check `misa[V]` instead of `sys_enable_vext()`?

rez5427:
fix #569

### DIFF
--- a/model/riscv_fdext_control.sail
+++ b/model/riscv_fdext_control.sail
@@ -37,8 +37,8 @@ function clause read_CSR (0x001) = zero_extend(fcsr[FFLAGS])
 function clause read_CSR (0x002) = zero_extend(fcsr[FRM])
 function clause read_CSR (0x003) = zero_extend(fcsr.bits)
 
-function clause write_CSR (0x001, value) = { dirty_fd_context(); write_fcsr(fcsr[FRM], value[4..0]); zero_extend(fcsr[FFLAGS]) }
-function clause write_CSR (0x002, value) = { dirty_fd_context(); write_fcsr(value[2..0], fcsr[FFLAGS]); zero_extend(fcsr[FRM]) }
-function clause write_CSR (0x003, value) = { dirty_fd_context(); write_fcsr(value[7..5], value[4..0]); zero_extend(fcsr.bits) }
+function clause write_CSR (0x001, value) = { write_fcsr(fcsr[FRM], value[4..0]); zero_extend(fcsr[FFLAGS]) }
+function clause write_CSR (0x002, value) = { write_fcsr(value[2..0], fcsr[FFLAGS]); zero_extend(fcsr[FRM]) }
+function clause write_CSR (0x003, value) = { write_fcsr(value[7..5], value[4..0]); zero_extend(fcsr.bits) }
 
 /* **************************************************************** */
--- a/model/riscv_fdext_control.sail
+++ b/model/riscv_fdext_control.sail
@@ -37,8 +37,8 @@ function clause read_CSR (0x001) = zero_extend(fcsr[FFLAGS])
 function clause read_CSR (0x002) = zero_extend(fcsr[FRM])
 function clause read_CSR (0x003) = zero_extend(fcsr.bits)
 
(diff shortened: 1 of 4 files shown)


jrtc27:
`misa[V] == 0b1` implies `sys_enable_vext()`, and the commit message is extremely devoid of any detail. Per CONTRIBUTING.md:

> Each commit should be self-contained, provide a meaningful short summary in the subject and, if not clear from the subject alone, provide more detail in the commit description.

But that's also just a common expectation for software engineering in general, it shouldn't need to be spelled out.

Timmmm:
I'm not sure this is correct either. I think we can get rid of `dirty_v_context_if_present()` and just call `dirty_v_context()`. Currently it is only called from `ext_write_vcsr` and that should only happen if V is enabled.

Emphasis on *should* because currently the `is_CSR_defined`s are missing `extensionEnabled(Ext_V)`:

```
function clause is_CSR_defined (0x009) = true
```

I opened #624 to track that.

I believe `dirty_fd_context_if_present()` exists because of Zfinx... but I don't think that's correct either. When you write to `fcsr` it should dirty `mstatus[FS]` even with Zfinx but that currently doesn't happen. I opened #625 to track that.

Also this doesn't completely fix #569 - that issue also says that there are missing calls to `dirty_v_context...`.

rez5427:
Maybe this is good?

Timmmm:
This looks a lot better. There are still some more places where the context needs to be dirtied, e.g. search for `vcsr[vxsat] = 0b1;` (I would extract it into a `set_vcsr_vxsat()` function or something). There's also a gazillion places where `vstart` is written directly. I would make a `set_vstart()` too.

Doesn't have to be in this PR though. In fact maybe better if it isn't...

model/riscv_fdext_control.sail
```diff
@@ -37,8 +37,8 @@ function clause read_CSR (0x001) = zero_extend(fcsr[FFLAGS])
 function clause read_CSR (0x002) = zero_extend(fcsr[FRM])
 function clause read_CSR (0x003) = zero_extend(fcsr.bits)
 
-function clause write_CSR (0x001, value) = { write_fcsr(fcsr[FRM], value[4..0]); zero_extend(fcsr[FFLAGS]) }
-function clause write_CSR (0x002, value) = { write_fcsr(value[2..0], fcsr[FFLAGS]); zero_extend(fcsr[FRM]) }
-function clause write_CSR (0x003, value) = { write_fcsr(value[7..5], value[4..0]); zero_extend(fcsr.bits) }
+function clause write_CSR (0x001, value) = { dirty_fd_context(); write_fcsr(fcsr[FRM], value[4..0]); zero_extend(fcsr[FFLAGS]) }
+function clause write_CSR (0x002, value) = { dirty_fd_context(); write_fcsr(value[2..0], fcsr[FFLAGS]); zero_extend(fcsr[FRM]) }
+function clause write_CSR (0x003, value) = { dirty_fd_context(); write_fcsr(value[7..5], value[4..0]); zero_extend(fcsr.bits) }
```
Timmmm, line 42: This is already called in `write_fcsr`. Let's leave the float stuff for another PR.

Timmmm (approved):
LGTM. I'll squash and fix the commit message when merging, which I will do in a few days if nobody objects.

jordancarlin (approved):

arichardson (approved):

rez5427:
I drop the changes related to this [pr](https://github.com/riscv/sail-riscv/pull/628)

PR #626, opened 2024-11-27
Support static compilation of the emulator

Timmmm:
Optionally wrap the libs with `-Wl,-Bstatic`, `-Wl,-Bdynamic` which will enable static compilation for those libs. We don't really want to use `-static` because it statically links glibc, which is a bad idea, and also increases the binary size. Linking with a sufficiently old version of glibc is a better option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,7 @@ ifneq (,$(STATIC))
         GMP_LIBS = $(shell pkg-config --variable=libdir gmp)/libgmp.a
         C_LIBS_WRAPPED = $(C_LIBS)
     else
-        C_LIBS_WRAPPED = -Wl,--push-state,-Bstatic $(C_LIBS) -Wl,--pop-state
+        C_LIBS_WRAPPED = -Wl,--push-state -Wl,-Bstatic $(C_LIBS) -Wl,--pop-state
     endif
 else
     C_LIBS_WRAPPED = $(C_LIBS)
--- a/Makefile
+++ b/Makefile
@@ -193,7 +193,7 @@ SAIL_FLAGS += -c_coverage $(ALL_BRANCHES) -c_include sail_coverage.h
 C_LIBS += $(SAIL_LIB_DIR)/coverage/libsail_coverage.a -lm -lpthread -ldl
 endif
 
-# Optional link C_LIBS statically. Unlike -static this will not
+# Optionally link C_LIBS statically. Unlike -static this will not
 # link glibc statically which is generally a bad idea.
 ifneq (,$(STATIC))
     UNAME_S := $(shell sh -c 'uname -s 2>/dev/null || echo not')
--- a/Makefile
+++ b/Makefile
@@ -158,6 +158,8 @@ SOFTFLOAT_SPECIALIZE_TYPE = RISCV
 GMP_FLAGS = $(shell pkg-config --cflags gmp)
 # N.B. GMP does not have pkg-config metadata on Ubuntu 18.04 so default to -lgmp
 GMP_LIBS = $(shell pkg-config --libs gmp || echo -lgmp)
+
+# TODO: Remove Zlib when upgrading to Sail 0.19; it is no longer a requirement.
 ZLIB_FLAGS = $(shell pkg-config --cflags zlib)
 ZLIB_LIBS = $(shell pkg-config --libs zlib)
 
@@ -189,6 +191,21 @@ ALL_BRANCHES = generated_definitions/c/all_branches
 C_FLAGS += -DSAILCOV
 SAIL_FLAGS += -c_coverage $(ALL_BRANCHES) -c_include sail_coverage.h
 C_LIBS += $(SAIL_LIB_DIR)/coverage/libsail_coverage.a -lm -lpthread -ldl
+endif
+
+# Optional link C_LIBS statically. Unlike -static this will not
+# link glibc statically which is generally a bad idea.
+ifneq (,$(STATIC))
+    UNAME_S := $(shell sh -c 'uname -s 2>/dev/null || echo not')
+    ifeq ($(UNAME_S),Darwin)
+        # Unfortunately the Mac linker does not support -Bstatic.
+        GMP_LIBS = $(shell pkg-config --variable=libdir gmp)/libgmp.a
+        C_LIBS_WRAPPED = $(C_LIBS)
+    else
+        C_LIBS_WRAPPED = -Wl,--push-state,-Bstatic $(C_LIBS) -Wl,--pop-state
+    endif
+else
+    C_LIBS_WRAPPED = $(C_LIBS)
 endif
 
 RISCV_EXTRAS_LEM_FILES = riscv_extras.lem mem_metadata.lem riscv_extras_fdext.lem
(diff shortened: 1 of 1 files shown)


Timmmm:
Tested with this Dockerfile:

```
FROM rockylinux:8.9.20231119

RUN dnf update -y --enablerepo=devel && \
    dnf install -y --enablerepo=devel \
        pkg-config \
        gmp-static \
        zlib-static \
        curl \
        git \
        make \
        gcc \
        gcc-c++ \
    && dnf clean all && rm -rf /var/cache/yum

# Install the Sail compiler.
ADD https://github.com/rems-project/sail/releases/download/0.18-linux-binary/sail.tar.gz /usr/local/
RUN tar --directory=/usr/local --strip-components=1 --extract --file /usr/local/sail.tar.gz

WORKDIR /work

ADD . .

RUN make c_emulator/riscv_sim_RV64 ARCH=RV64 STATIC=1

# To build and extract the emulator binary.
#
#     docker build -t sail_riscv .
#     id=$(docker create sail_riscv)
#     docker cp $id:/work/c_emulator/riscv_sim_RV64 .
#     docker rm -v $id
```

Stripped binary:

[riscv_sim_RV64.tar.gz](https://github.com/user-attachments/files/17933831/riscv_sim_RV64.tar.gz)

```
❯ ldd riscv_sim_RV64 
        linux-vdso.so.1 (0x00007ffd4e493000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f9fd1d21000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f9fd20f7000)
```

jordancarlin (approved):
LGTM. Even if we don't end up adding a precompiled binary release (which I think we should), it's still a good idea to give users the option to link statically if they want.

Makefile
```diff
@@ -191,6 +191,14 @@ SAIL_FLAGS += -c_coverage $(ALL_BRANCHES) -c_include sail_coverage.h
 C_LIBS += $(SAIL_LIB_DIR)/coverage/libsail_coverage.a -lm -lpthread -ldl
 endif
 
+# Optional link C_LIBS statically. Unlike -static this will not
+# link glibc statically which is generally a bad idea.
+ifneq (,$(STATIC))
+    C_LIBS_WRAPPED = -Wl,-Bstatic $(C_LIBS) -Wl,-Bdynamic
```
jrtc27, line 197: Maybe worth noting this isn't a thing that works on macOS (nor does it have an equivalent beyond providing paths to the .a files or messing with the search paths)? Also `-Wl,--push-state,-Bstatic ... -Wl,--pop-state` is probably better practice, though in reality you're unlikely to be in an environment where it's not -Bdynamic.

Timmmm: Ah very good point! I kind of totally forgot about MacOS existing! I will try and make this work there too. Mercifully `gmp` in Brew provides the static version.

And thanks for telling me about `--push-state`, `--pop-state`; I didn't know about that and it is definitely better.

jordancarlin:
> Tested with this Dockerfile:
> 
> ... 
>
> Stripped binary:
> 
> [riscv_sim_RV64.tar.gz](https://github.com/user-attachments/files/17933831/riscv_sim_RV64.tar.gz)
> 

I tested this binary on several distros and got the following error on all of them:

```
File too small, not big enough even for 32-bit ELF header
```

It's possible I did something wrong with it, but using the same elf file with a version of the model that was compiled on that system worked fine.

Timmmm:
That sounds like you've had a download error. What's the size / MD5 of the executable? It should be 1685008 bytes, a42c0a0f022c44107c6c1de324dff1d5.

Timmmm:
Oh wait, I misunderstood - thought you meant the emulator wouldn't even run. That error is probably because I committed some crimes to remove avoid the zlib dependency and I must have got it wrong...

jordancarlin:
Yes. The second issue. That makes sense. Now that the zlib dependency has been removed from Sail this hopefully won't be an issue going forward.

Timmmm:
Here is one compiled without my broken hack using a recent Sail compiler instead:

[riscv_sim_RV64.tar.gz](https://github.com/user-attachments/files/17976143/riscv_sim_RV64.tar.gz)

Makefile
Timmmm: Updated with `--push/pop-state`, and also Mac support. Make and Mac's linker don't make it easy unfortunately, but it works.

Makefile
```diff
@@ -155,9 +155,28 @@ SOFTFLOAT_FLAGS  = -I $(SOFTFLOAT_INCDIR)
 SOFTFLOAT_LIBS   = $(SOFTFLOAT_LIBDIR)/softfloat.a
 SOFTFLOAT_SPECIALIZE_TYPE = RISCV
 
+UNAME_S := $(shell sh -c 'uname -s 2>/dev/null || echo not')
+
+STATIC_GMP_A :=
+STATIC_B_STATIC :=
```
jordancarlin, line 161: Might be better to use a more descriptive name here that indicates that these are for different OSs. Something like `STATIC_GMP_MACOS` and `STATIC_LINUX`.

Timmmm: I had that originally but it felt clearer to have the logic of how the OS affects what actions are performed in one place, if that makes sense. Maybe `STATIC_LINUX`, `STATIC_MAC` are better though.. anyone else have an opinion?

jordancarlin:
> Here is one compiled without my broken hack using a recent Sail compiler instead:
> 
> [riscv_sim_RV64.tar.gz](https://github.com/user-attachments/files/17976143/riscv_sim_RV64.tar.gz)

New version works great. Able to run elfs through the simulator without installing anything else. Tested the binary on Ubuntu 20.04, 22.04, and 24.04, Debian 11 and 12, Rocky 8 and 9, and OpenSUSE 15.6. All of them are able to run all of the rv64 tests in the `tests/riscv-tests` directory.

Makefile
jordancarlin: What about something like this then if you want to keep the OS logic together? And just leave the existing `GMP_LIBS` definition as it is.

```make
ifneq (,$(STATIC))
    UNAME_S := $(shell sh -c 'uname -s 2>/dev/null || echo not')
    ifeq ($(UNAME_S),Darwin)
        GMP_LIBS := $(shell pkg-config --variable=libdir gmp)/libgmp.a
        C_LIBS_WRAPPED = $(C_LIBS)
    else
        C_LIBS_WRAPPED = -Wl,--push-state,-Bstatic $(C_LIBS) -Wl,--pop-state
    endif
else
    C_LIBS_WRAPPED = $(C_LIBS)
endif
```

Timmmm: Updated to that suggestion, thanks!

Timmmm:
Ok I'll merge this in a few days. It doesn't change the default behaviour so it's very low risk.

I'll try and get the Sail binary release for Mac working before the next compiler release (getting some weird Git errors at the moment) and then I will make a release workflow for `sail-riscv` so we can release some binaries.

Makefile
```diff
@@ -191,6 +193,21 @@ SAIL_FLAGS += -c_coverage $(ALL_BRANCHES) -c_include sail_coverage.h
 C_LIBS += $(SAIL_LIB_DIR)/coverage/libsail_coverage.a -lm -lpthread -ldl
 endif
 
+# Optional link C_LIBS statically. Unlike -static this will not
```
jordancarlin, line 196:
```suggestion
# Optionally link C_LIBS statically. Unlike -static this will not
```

jordancarlin (approved):
LGTM

PR #627, opened 2024-11-27
Move HPM code into hpm sail file and rename it

Timmmm:
* Rename `riscv_hpm_control.sail` to `riscv_zihpm.sail`, so that it's named the same as the extension. I don't think we need separate `control` and `regs` files for small extensions, and it's kind of unclear what code should go in which of those anyway.
* Move the HPM register access code into that file.


Timmmm:
Yeah I think that was the right call! I'll merge this in a few days if nobody objects.

Issue #624, opened 2024-11-26
Vector registers missing extensionEnabled(Ext_V) guards


Timmmm:
All the vector registers are currently always available, irrespective of whether vector support was enabled, or `misa[V]`:

```
function clause is_CSR_defined (0x008) = true
function clause is_CSR_defined (0x009) = true
function clause is_CSR_defined (0x00A) = true
function clause is_CSR_defined (0x00F) = true
function clause is_CSR_defined (0xC20) = true
function clause is_CSR_defined (0xC21) = true
function clause is_CSR_defined (0xC22) = true
```

Should be

```
enum clause extension = Ext_V
function clause extensionEnabled(Ext_V) = (misa[V] == 0b1) & (mstatus[VS] != 0b00)

...

function clause is_CSR_defined (0x008) = extensionEnabled(Ext_V) // vstart
function clause is_CSR_defined (0x009) = extensionEnabled(Ext_V) // vxsat
function clause is_CSR_defined (0x00A) = extensionEnabled(Ext_V) // vxrm
function clause is_CSR_defined (0x00F) = extensionEnabled(Ext_V) // vcsr
function clause is_CSR_defined (0xC20) = extensionEnabled(Ext_V) // vl
function clause is_CSR_defined (0xC21) = extensionEnabled(Ext_V) // vtype
function clause is_CSR_defined (0xC22) = extensionEnabled(Ext_V) // vlenb
```

PR #628, opened 2024-11-29
Add extensionEnabled(Ext_V) guards for vector registers

KotorinMinami:
As mentioned in #624 , the registers involved in the V extension are missing extensionEnabled guards, so this is added.

In this pr, I've changed the location of the Ext_V definition to ensure compilation passes and compliance with the ExtendingGuide.


jordancarlin (approved):
LGTM. Simple fix

PR #630, opened 2024-11-30
Update recommended use of xlen in CODE_STYLE

jordancarlin:
Update CODE_STYLE.md to recommend using `xlen` instead of `sizeof(xlen)` to align with our current usage.

### DIFF
--- a/CODE_STYLE.md
+++ b/CODE_STYLE.md
@@ -62,11 +62,11 @@ Implementation
 
 * All instructions should be built as part of both the RV32 and RV64 models so
   as to provide a path to supporting mutable MXL/SXL/UXL; if necessary,
-  constructs like `assert(sizeof(xlen) == 32)` at the start of the body can be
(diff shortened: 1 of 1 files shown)


CODE_STYLE.md
```diff
@@ -62,11 +62,11 @@ Implementation
 
 * All instructions should be built as part of both the RV32 and RV64 models so
   as to provide a path to supporting mutable MXL/SXL/UXL; if necessary,
-  constructs like `assert(sizeof(xlen) == 32)` at the start of the body can be
+  constructs like `xlen == 32)` at the start of the body can be
```
jrtc27, line 65:
```suggestion
  constructs like `assert(xlen == 32)` at the start of the body can be
```

Not sure what happened here...

jordancarlin: Oops. Not sure how that happened. Fixed.

PR #634, opened 2024-12-06
Move extension scattered enum to single file

### LINKED ISSUE #633: Make `extension` enum not scattered
Currently we have a scattered `extension` enum that we generally add entries to in the files that implement the extension, e.g. `riscv_svinval.h` contains:

```
enum clause extension = Ext_Svinval
function clause extensionEnabled(Ext_Svinval) = sys_enable_svinval()
```

I've found that this causes quite annoying circular dependencies, especially with functions that depend on lots of extensions (e.g. stuff around legalizing `mie`, `mstatus`, `misa`, etc.).

I propose making `extension` not scattered, and just putting it all in a big list in one file (`riscv_extensions.sail`)

```
enum extension = {
  Ext_S,
  Ext_Svinval,
  Ext_...
}
```

This can then be added early on in the compile order. I don't think there are any real downsides to this and it might actually be quite nice to see them all in one place.

The `extensionEnabled()` clauses should still be scattered.

Timmmm:
Having this distributed through the code can lead to slightly awkward dependency order issues. We need to keep it scattered though so e.g. CHERI can add its extensions.

Fixes #633

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -1,0 +1,107 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+scattered enum extension
+
+// Note, these are sorted according to the canonical ordering vaguely described
+// in the `Subset Naming Convention` section of the unprivileged spec.
+
+// Integer Multiplication and Division; not Machine!
+enum clause extension = Ext_M
+// Single-Precision Floating-Point
+enum clause extension = Ext_F
+// Double-Precision Floating-Point
+enum clause extension = Ext_D
+// Compressed Instructions
+enum clause extension = Ext_C
+// Bit Manipulation
+enum clause extension = Ext_B
+// Vector Operations
+enum clause extension = Ext_V
+// Supervisor
+enum clause extension = Ext_S
+// User
+enum clause extension = Ext_U
+
+// Cache-Block Management Instructions
+enum clause extension = Ext_Zicbom
+// Cache-Block Zero Instructions
+enum clause extension = Ext_Zicboz
+// Integer Conditional Operations
+enum clause extension = Ext_Zicond
+// Instruction-Fetch Fence
+enum clause extension = Ext_Zifencei
+// Hardware Performance Counters
+enum clause extension = Ext_Zihpm
+
+// Multiplication and Division: Multiplication only
+enum clause extension = Ext_Zmmul
+// Atomic Memory Operations
+
+enum clause extension = Ext_Zaamo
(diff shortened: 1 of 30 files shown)


jrtc27:
This doesn’t work for sail-cheri-riscv.

Alasdair:
I think the enumeration should probably remain scattered, but it seems to me there are two cases:

- Standard extensions that are declared up-front, and that appear in `misa`. We could declare all those standard single letter extensions in one place

- Other extensions that don't appear in `misa`, e.g. `Xcheri`

Timmmm:
Hmm good point about CHERI. Unfortunately it isn't just misa. There's also things like menvcfg/mseccfg which have circular dependencies.

I guess technically the right thing to do would be to make `legalize_menvcfg` etc. scattered functions, but that just seems like overkill.

How about we keep it scattered but still move them all into one file?

arichardson:
> This doesn’t work for sail-cheri-riscv.

True, but IMO sail-cheri-riscv should change to just be a diff on top of the base model. That would make upstreaming a lot easier as well.

jrtc27:
> > This doesn’t work for sail-cheri-riscv.
> 
> True, but IMO sail-cheri-riscv should change to just be a diff on top of the base model. That would make upstreaming a lot easier as well.

It should, but it won’t do so immediately, and the model has been designed to support the current approach, so we do need to uphold that.

model/riscv_extensions.sail
```diff
@@ -0,0 +1,61 @@
...
+// Double precision float
+enum clause extension = Ext_D
+// Single precision float
+enum clause extension = Ext_F
+// Multiply/Divide; not Machine!
+enum clause extension = Ext_M
+// Supervisor
+enum clause extension = Ext_S
+// Supervisor time compare
+enum clause extension = Ext_Sstc
```
jordancarlin, line 24: Other than `Sstc`, only single-letter extensions have comments. This should probably be consistent. 

Though to be honest, I think it makes more sense to either have comments for all the extensions or none of them. There is nothing special about single-letter ones in this regard.

Timmmm: I was mainly being lazy, but I have now tediously gone through the spec and copied the official extension names into comments for every one (where possible; Zmmul and Zca didn't have explicit names).

Timmmm: I guess this could be a nice `mapping extensionName` at some point but lets do that later.

jordancarlin (approved):
LGTM now that comments are consistently added. Keeping it as a scattered enum should avoid the issues with CHERI while giving us the benefit of simpler file ordering.

I wonder if it might be nicer to order the extensions by canonical order instead of alphabetically so that extensions like Zfinx and Zdinx are next to each other, but it's also fine as it is. If you wanted to use the same ordering that I put in the README that would work.

Timmmm:
Sorted according to the canonical ordering. Which actually doesn't make any sense for `Zhinh` since it's supposed to be related to Hypervisor. Eh whatever.

jordancarlin:
Hmm. Really should've been named `Zfhinx` or something. I'd say this is still an improvement though.

jordancarlin (approved):

PR #635, opened 2024-12-06
Allow writing CBO bits in xenvcfg

Timmmm:
This was accidentally missed from the CBO implementation.

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -349,6 +349,9 @@ function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
   let v = Mk_MEnvcfg(v);
   [o with
     FIOM = if sys_enable_writable_fiom() then v[FIOM] else 0b0,
+    CBZE = if extensionEnabled(Ext_Zicboz) then v[CBZE] else 0b0,
+    CBCFE = if extensionEnabled(Ext_Zicbom) then v[CBCFE] else 0b0,
+    CBIE = if extensionEnabled(Ext_Zicbom) then (if v[CBIE] != 0b10 then v[CBIE] else 0b00) else 0b00,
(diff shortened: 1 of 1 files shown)


jordancarlin:
Shouldn't these be gated on `sys_enable_zicbom` and `sys_enable_zicboz`?

Timmmm:
Ah yes you're right. I'll update it.

Timmmm:
I'll wait for #634 since this is exactly the sort of problem it helps with.

jordancarlin:
I’ll review once this is rebased on top of #634 now that it’s merged.

model/riscv_sys_regs.sail
```diff
@@ -349,6 +349,9 @@ function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
   let v = Mk_MEnvcfg(v);
   [o with
     FIOM = if sys_enable_writable_fiom() then v[FIOM] else 0b0,
+    CBZE = if extensionEnabled(Ext_Zicboz) then v[CBZE] else 0b0,
+    CBCFE = if extensionEnabled(Ext_Zicbom) then v[CBCFE] else 0b0,
+    CBIE = if extensionEnabled(Ext_Zicbom) then (if v[CBIE] != 0b10 then v[CBIE] else 0b00) else 0b00,
```
jordancarlin, line 354: Might be easier to read if this was `if extensionEnabled(Ext_Zicbom) & (v[CBIE] != 0b10) then v[CBIE] else 0b00`, but the current version more closely matches the intention of the spec so either way seems fine.

jordancarlin (approved):
LGTM. A simple fix for something that was missed when this extension was added.

model/riscv_sys_regs.sail
Timmmm: Yeah I think I could go either way on that. I'll just leave it for now I think.

PR #637, opened 2024-12-06
Use better types for virtual memory translation code and support Sv57

Timmmm:
Change the VM code from using `bits(64)` everywhere to using more accurate types. This also enables Sv57 support.


Timmmm:
I wrote this code quite a while ago and kind of got sidetracked by other stuff, but it looks like I finished it so I thought I may as well push a draft and see what people think. It hasn't been tested extensively yet but I can do that.

One slightly awkward thing I found is that you can use really nice strict types everywhere for the actual translation because the lifetime of `sv_width` is the scope of the translation function, so it can only have one value. But you can't really do it for the TLB because the lifetime of `sv_width` for the TLB is infinite so you can end up storing Sv39 addresses in it one cycle and the Sv57 a few cycles later. I just used the widths for Sv57 (the max) and then truncate/zero_extend as appropriate.

Timmmm:
Ok I think this is ready. I will run it through our CI system and report the test results.

model/riscv_vmem_tlb.sail
```diff
@@ -33,112 +70,101 @@ type num_tlb_entries : Int = 64
...
-                                 vMatchMask = vMatchMask,
-                                 vAddr      = vAddr & vMatchMask,
-                                 pAddr      = pAddr & vMatchMask,
-                                 age        = mcycle};
+function add_to_TLB forall 'v, is_sv_mode('v) . (
+  sv_width : int('v),
+  asid     : asidbits,
+  vpn      : bits('v - pagesize_bits),
+  ppn      : bits(if 'v == 32 then 22 else 44),
+  pte      : bits(if 'v == 32 then 32 else 64),
```
arichardson, line 133:
```suggestion
  ppn      : ppn_bits('v),
  pte      : pte_bits('v),
```
Does this work?

model/riscv_vmem_tlb.sail
```diff
@@ -33,112 +70,101 @@ type num_tlb_entries : Int = 64
...
-  None(), None(), None(), None(), None(), None(), None(), None(),
-]
+register tlb : vector(num_tlb_entries, option(TLB_Entry)) = vector_init(None())
 
 // Indexed by the lowest bits of the VPN.
-function tlb_hash(vaddr : bits(64)) -> tlb_index_range =
-  unsigned(vaddr[17 .. 12])
+function tlb_hash forall 'v, is_sv_mode('v) . (
+  sv_mode : int('v),
+  vpn     : bits('v - pagesize_bits),
```
arichardson, line 78: bits('v - pagesize_bits) seems to be used quite a lot, should there be a named alias?

arichardson (commented):
I have not reviewed this super thoroughly and am probably not the best person to review the vmem code but it looks good to me.

model/riscv_vmem_tlb.sail
Timmmm: It does, thanks!

model/riscv_vmem_tlb.sail
Timmmm: Good point, I added `vpn_bits` for that.

arichardson (approved):
LGTM but someone else should also take a look

model/riscv_sys_regs.sail
```diff
@@ -852,24 +852,31 @@ bitfield Satp64 : bits(64) = {
...
 }
 
-function legalize_satp32(a : Architecture, o : bits(32), v : bits(32)) -> bits(32) = {
-  /* all 32-bit satp modes are valid */
-  v
+// TODO: This currently assumes all ASID bits are implemented
+// and does not legalize the asid field if asidlen is not the maximum.
+function legalize_satp(
+  a : Architecture,
+  o : xlenbits,
```
rsnikhil, line 865: Suggest adding comment:  /* previous value of satp */

model/riscv_sys_regs.sail
```diff
@@ -852,24 +852,31 @@ bitfield Satp64 : bits(64) = {
...
+  v : xlenbits,
```
rsnikhil, line 866: Suggest adding comment: /* proposed new value of satp */

model/riscv_vmem_tlb.sail
```diff
@@ -36,100 +73,98 @@ type tlb_index_range = range(0, num_tlb_entries - 1)
...
-function tlb_hash(vaddr : bits(64)) -> tlb_index_range =
-  unsigned(vaddr[17 .. 12])
+function tlb_hash forall 'v, is_sv_mode('v) . (
+  sv_mode : int('v),
+  vpn     : vpn_bits('v),
+) -> tlb_index_range =
+  unsigned(vpn[5 .. 0])
 
-// PUBLIC: invoked in init_vmem() [riscv_vmem.sail]
-function init_TLB() -> unit = {
```
rsnikhil, line 43: Will we not need 'init_TLB()' if we allow a dynamic reset on the Sail model?

model/riscv_step.sail
```diff
@@ -127,7 +127,6 @@ function loop () : unit -> unit = {
 function init_model () -> unit = {
   init_platform (); /* devices */
   init_sys ();      /* processor */
-  init_vmem ();     /* virtual memory */
```
rsnikhil, line 130: Will we not need 'init_vmem()' if we allow a dynamic reset on the Sail model?

rsnikhil (commented):
I have reviewed all 15 changed files; all generally LGTM.

Some very minor suggestions (see individual comments on particular files/lines).

I did not study the translation algorithm in riscv_vmem.sail in detail; it looked similar (structure, flow) to original, just fixing it up for the new tighter vmem types, so I think it's probably fine.

The previous version was less type-safe, using 64 bits everywhere, but could work for dynamic choice between Sv32, Sv39, Sv48 and Sv57.  I think perhaps this new code requires this to be a static choice, to enable type-checking?

model/riscv_sys_regs.sail
Timmmm: Good point - I updated the variable names to be more descriptive rather than adding comments. We should probably do that for all the other `legalize_` functions which also use incomprehensible single letter variables.

model/riscv_vmem_tlb.sail
Timmmm: Good question. We don't for RVFI since the TLB will automatically be reset by Sail. For "real" resets I guess *technically* you don't need to reset the TLB for a functioning chip. Feels like that would be an unnecessarily risky design though. I'll add it back in.

Timmmm:
> I think perhaps this new code requires this to be a static choice, to enable type-checking?

In most languages it would, but in Sail it actually doesn't due to its fancy dependent type system. Don't ask me how it works but it does!

model/riscv_vmem_tlb.sail
Timmmm: Added back.

model/riscv_step.sail
Timmmm: Added back.

model/riscv_sys_regs.sail
Timmmm: Parameter renamed.

model/riscv_vmem.sail
```diff
@@ -20,148 +20,128 @@
 // However, we model a simple TLB so that
```
jordancarlin, line 20: Doesn't let me comment higher up in the file, but on line 12 `Sv57` should be moved from `TODO` to `Supported VM modes`. It should also be added to the list of supported extensions in the README.

Timmmm: Done.

jordancarlin (approved):

Makefile
```diff
@@ -83,14 +83,13 @@ SAIL_SYS_SRCS += riscv_sys_control.sail     # general exception handling
 # SAIL_RV32_VM_SRCS = riscv_vmem_sv32.sail riscv_vmem_rv32.sail
```
arichardson, line 83: Could delete all these commented out lines?

arichardson (approved):

Makefile
Timmmm: Yes, done.

Timmmm:
Rebased, and I also updated `haveSvnapot` and `haveSvpbmt` to `extensionEnabled()`.

model/riscv_extensions.sail
```diff
@@ -107,3 +107,7 @@ enum clause extension = Ext_Zhinx
 enum clause extension = Ext_Sstc
 // Fine-Grained Address-Translation Cache Invalidation
 enum clause extension = Ext_Svinval
+// NAPOT Translation Contiguity
+enum clause extension = Ext_Svnapot
+// Page-Based Memory Types
+enum clause extension = Ext_Svpmbt
```
jordancarlin, line 113:
```suggestion
enum clause extension = Ext_Svpbmt
```

Same problem everywhere this is used

Timmmm: Whoops! Fixed.

Timmmm:
I've also tested this in our CI system with about 180 STING tests (much better than riscv-tests / ACTs) so I think it's good to go.

jordancarlin:
Looks like the CMake file list needs to be updated now.

Timmmm:
CMake updated. I'll give this a week at least before merging since it's a big change.

Timmmm:
Actually... I will merge this tomorrow. I'm doing some other work on virtual memory and don't really want to do it twice!

PR #638, opened 2024-12-06
Increase RAM size

Timmmm:
This matches Spike, and means a default Linux build has enough RAM to boot properly.

I'm not exactly sure why the old value was not sufficient but it didn't work when I tried and this did, and I think consistency with Spike is valuable anyway.


jordancarlin (approved):
Seems like sufficient reasons to change it and this is already an option that can be overridden with a command line flag if the smaller ram size is needed by anyone.

Alasdair (approved):

PR #641, opened 2024-12-13
make physical address 34-bit in sv32

KotorinMinami:
Restructure the physical address to be 34 bits in the RV32 case and 64 bits in the RV64 case. Idea from @Timmmm and @Alasdair in #615 

In the RV32 case, if sv32 is not enabled, use the Spike approach and zero-extend the virtual address to 34 bits for translation.

### DIFF
--- a/model/prelude_mem.sail
+++ b/model/prelude_mem.sail
@@ -58,15 +58,18 @@ struct RISCV_strong_access = {
 
 /* The Sail concurrency interface lets us have a physical address type
    with additional information, provided we can supply a function that
-   converts it into a bitvector.  Since we are just using xlenbits as a
-   physical address, we need the identity function for xlenbits. */
-val xlenbits_identity : xlenbits -> xlenbits
+   converts it into a bitvector. When in RV32 mode, physaddrbits is
+   bits(34), which needs to be zero-extended to bits(64) for
+   sail_mem_write/read interface. This conversion is necessary because
+   the memory interface only supports 32-bit or 64-bit physical
+   addresses regardless of the actual physical address width. */
+val physaddrbits_zero_extend : physaddrbits -> bits(64)
 
-function xlenbits_identity xs = xs
+function physaddrbits_zero_extend xs = zero_extend(xs)
 
 instantiation sail_mem_write with
-  'pa = xlenbits,
-  pa_bits = xlenbits_identity,
+  'pa = physaddrbits,
+  pa_bits = physaddrbits_zero_extend,
   /* We don't have a relaxed-memory translation model for RISC-V, so
      we just use unit as a dummy type. */
   'translation_summary = unit,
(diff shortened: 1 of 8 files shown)


model/prelude_mem.sail
```diff
@@ -60,13 +60,13 @@ struct RISCV_strong_access = {
    with additional information, provided we can supply a function that
    converts it into a bitvector.  Since we are just using xlenbits as a
```
Timmmm, line 61: Need to update this comment.

Timmmm (commented):
LGTM, very nice! Just a minor comment needs updating.

Do you know of any tests that exist for this? (I don't think it should block merging; just if there is a test somewhere that tests this it would be good to try it.)

model/prelude_mem.sail
```diff
@@ -60,13 +60,13 @@ struct RISCV_strong_access = {
...
    physical address, we need the identity function for xlenbits. */
-val xlenbits_identity : xlenbits -> xlenbits
+val physaddrbits_identity : physaddrbits -> bits(64)
 
-function xlenbits_identity xs = xs
+function physaddrbits_identity xs = zero_extend(xs)
```
Timmmm, line 65: Maybe this should be called `_zero_extend` instead of `_identity`?

Alasdair:
Other than renaming `physaddrbits_identity` as Tim suggested this should be good to merge.

model/prelude_mem.sail
KotorinMinami: Sounds great ! Done

model/prelude_mem.sail
KotorinMinami: I have modified it, I don't know if this comment is appropriate

model/prelude_mem.sail
```diff
@@ -58,15 +58,18 @@ struct RISCV_strong_access = {
 
 /* The Sail concurrency interface lets us have a physical address type
    with additional information, provided we can supply a function that
-   converts it into a bitvector.  Since we are just using xlenbits as a
-   physical address, we need the identity function for xlenbits. */
-val xlenbits_identity : xlenbits -> xlenbits
+   converts it into a bitvector. When in RV32 mode, physaddrbits is
+   bits(34), which needs to be zero-extended to bits(64) for
+   sail_mem_write/read interface. This conversion is necessary because
+   the memory interface expects 64-bit addresses
```
Timmmm, line 64: It supports 32 bits too I guess. I'd say "This conversion is necessary because the memory interface only supports 32-bit or 64-bit physical addresses."

model/prelude_mem.sail
```diff
@@ -58,15 +58,18 @@ struct RISCV_strong_access = {
...
 /* The Sail concurrency interface lets us have a physical address type
    with additional information, provided we can supply a function that
-   converts it into a bitvector.  Since we are just using xlenbits as a
-   physical address, we need the identity function for xlenbits. */
-val xlenbits_identity : xlenbits -> xlenbits
+   converts it into a bitvector. When in RV32 mode, physaddrbits is
+   bits(34), which needs to be zero-extended to bits(64) for
+   sail_mem_write/read interface. This conversion is necessary because
+   the memory interface only supports 32-bit or 64-bit physical
+   addressesregardless of the actual physical address width. */
```
Timmmm, line 65:
```suggestion
   addresses regardless of the actual physical address width. */
```

KotorinMinami: sorry, fixed

Timmmm (approved):
LGTM, thanks! Can't see any reason not to merge this. Tests would be good but it didn't work at all before so it can't be any worse.

arichardson (approved):

PR #645, opened 2024-12-17
Simplify Zicsr AST

nwf:
Split out from and sitting atop #617, the last commit here is the salient one.  This transforms a `bool * regidx` under one `ast` data constructor into two `ast` data constructors, allowing us to stop punning between `regidx` and `bits(5)`.  This will be significant in light of #617 for supporting the E extension, where `regidx` will be (isomorphic to) `bits(4)`; see #646.

### DIFF
--- a/model/riscv_insts_zicsr.sail
+++ b/model/riscv_insts_zicsr.sail
@@ -9,7 +9,8 @@
 /* ****************************************************************** */
 /* This file specifies the instructions in the 'Zicsr' extension.     */
 /* ****************************************************************** */
-union clause ast = CSR  : (csreg, regidx, regidx, bool, csrop)
+union clause ast = CSRReg  : (csreg, regidx, regidx, csrop)
+union clause ast = CSRImm  : (csreg, bits(5), regidx, csrop)
 
 mapping encdec_csrop : csrop <-> bits(2) = {
   CSRRW <-> 0b01,
@@ -17,12 +18,13 @@ mapping encdec_csrop : csrop <-> bits(2) = {
   CSRRC <-> 0b11
 }
 
-mapping clause encdec = CSR(csr, rs1, rd, is_imm, op)
-  <-> csr @ encdec_reg(rs1) @ bool_bits(is_imm) @ encdec_csrop(op) @ encdec_reg(rd) @ 0b1110011
+mapping clause encdec = CSRReg(csr, rs1, rd, op)
+  <-> csr @ encdec_reg(rs1) @ 0b0 @ encdec_csrop(op) @ encdec_reg(rd) @ 0b1110011
 
-function clause execute CSR(csr, rs1, rd, is_imm, op) = {
-  let rs1_val : xlenbits =  if is_imm then zero_extend(regidx_bits(rs1)) else X(rs1);
-  let is_CSR_Write = op == CSRRW | rs1 != zreg;
+mapping clause encdec = CSRImm(csr, imm, rd, op)
+  <-> csr @ imm @ 0b1 @ encdec_csrop(op) @ encdec_reg(rd) @ 0b1110011
+
+function doCSR(csr : csreg, rs1_val : xlenbits, rd : regidx, op : csrop, is_CSR_Write: bool) -> Retired = {
   if not(check_CSR(csr, cur_privilege, is_CSR_Write))
   then { handle_illegal(); RETIRE_FAIL }
   else if not(ext_check_CSR(csr, cur_privilege, is_CSR_Write))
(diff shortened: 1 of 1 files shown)


Timmmm:
Ah I understand the motivation now. In order to avoid losing the `is_imm` information could we use a union instead?

```
union RegOrImm = {
  Reg : regidx,
  Imm : bits(5),
}
```

nwf:
I am... reasonably certain we haven't lost any information and that I haven't changed semantics.  Rather than pass in a union and conditionally call `X()`, I just moved the discrimination into `ast` itself and had the appropriate arm unconditionally call `X()`.  Can you say why you think a `union` is beneficial?

model/riscv_insts_zicsr.sail
```diff
@@ -9,23 +9,25 @@
...
-  let regidx(rs1_bits) = rs1;
-  let rs1_val : xlenbits = if is_imm then zero_extend(rs1_bits) else X(rs1);
+mapping clause encdec = CSRImm(csr, imm, rd, op)
+  <-> csr @ imm @ 0b1 @ encdec_csrop(op) @ encdec_reg(rd) @ 0b1110011
+
+function doCSR(csr : csreg, rs1_val : xlenbits, rd : regidx, op : csrop) -> Retired = {
   let isWrite : bool = match op {
     CSRRW  => true,
-    _      => if is_imm then unsigned(rs1_val) != 0 else rs1 != zreg
+    _      => unsigned(rs1_val) != 0
```
arichardson, line 30: I'm not sure this is correct, looking at https://github.com/riscv/riscv-isa-manual/blob/main/src/zicsr.adoc it only says write = false for imm=0 or rs1=0, but if rs1 happens to hold a zero value it still performs the write.

> Note that if rs1 specifies a register other than x0, and that register holds a zero value, the instruction will not action any attendant per-field side effects, but will action any side effects caused by writing to the entire CSR.

arichardson (changes requested):
I like the split into two separate AST clauses here but I think this changes the behaviour to no longer be correct.

model/riscv_insts_zicsr.sail
nwf: Oh, hah, whoops; I was a victim of the kind of type confusion I was hoping to prevent by making things more explicit.

I'll pass `isWrite` to the `doCSR` and have the case arms do the appropriate discrimination.

arichardson (approved):

Timmmm:
> Can you say why you think a union is beneficial?

Hey, sorry I forgot to reply to this. There are two cases I know of where knowing `is_imm` is needed:

1. CHERI CLEN CSRs can only have their metadata written CSRRW:

```
function handle_csr_op(csr, rs1, rd, is_imm, op, isWrite, isRead) = {
  if is_clen_csr(csr) then {
    // The only way to write metadata is with `csrrw`.
    // For extended CSRs, it also has to be in CapPtrMode mode.
    // Note `op == CSRRW` means `csrrw` or` csrrwi` which is why we also check `not(is_imm)`.
    let clen_write = (effective_cheri_mode() == CapPtrMode | csr_type(csr) == ClenNew) & op == CSRRW & not(is_imm);
    let write_val = if is_imm then {null_cap with address = zero_extend(rs1)}
      else if clen_write then C(rs1)
      else {null_cap with address = X(rs1)};
...
```

2. Some CLIC CSRs have weird behaviour that depends on `is_imm`. At least `xnxti`:

```
function access_xnxti(
  csr : csreg,
  rs1_val : xlenbits,
  op : csrop,
  is_imm : bool,
  isWrite : bool,
  priv : Privilege) -> xlenbits = {
...
    let min_pil = if is_imm then xcause[Xpil] else rs1_val[23..16];
```

(I haven't upstreamed that code yet; CLIC is a mess and I'm not sure how much we should want it in this repo... but it would be nice if we can still easily add it on.)

Neither of those bits of code are actually in this repo yet though, so maybe it is unfair to worry about them. I guess we could always add `is_imm` back as a parameter to `doCSR` when it is needed?

model/riscv_insts_zicsr.sail
```diff
@@ -50,13 +51,24 @@ function clause execute CSR(csr, rs1, rd, is_imm, op) = {
...
+function clause execute CSRReg(csr, rs1, rd, op) =
+  doCSR(csr, X(rs1), rd, op, (op == CSRRW) | (rs1 != zreg))
+
+function clause execute CSRImm(csr, imm, rd, op) =
+  doCSR(csr, zero_extend(imm), rd, op, (op == CSRRW) | (imm != zeros_implicit()))
+
+mapping maybe_i : bool <-> string = {
+  true  <-> "i",
+  false <-> ""
+}
```
arichardson, line 63: is this used anywhere? I don't see it in this commit

arichardson (approved):
One minor note otherwise LGTM

pmundkur:
Perhaps modifying the PR is not necessary; github says this can be just merged without conflicts?

arichardson:
> Perhaps modifying the PR is not necessary; github says this can be just merged without conflicts?

I just clicked the update with rebase button, should now only include the final commit.

model/riscv_insts_zicsr.sail
nwf: No, indeed, looks stale.  I'll remove it.

pmundkur (approved):

arichardson (approved):

Issue #409, opened 2024-02-19
Use vectors for registers


Timmmm:
As far as I can see there's no reason to have this:

```
register x1  : regtype
register x2  : regtype
register x3  : regtype
register x4  : regtype
register x5  : regtype
register x6  : regtype
...
```

Instead of

```
register xrf : vec(regtype, 32)
```

This is done for all the X, F and V registers.

Alasdair:
In the concurrency model, `x1` .. `xN` are separate registers for intra-instruction dependencies (so instructions modifying different registers can be re-ordered), whereas `xrf : vector(...)` would be a single register.

jrtc27:
Hm, that seems less than ideal. Should the concurrency model not differentiate between vector-of-registers and register-of-vectors?

Alasdair:
The concurrency model does differentiate between vector-of-registers and register-of-vectors, I think that's the issue here.

The issue with tracking dependencies at a sub-register granularity is we'd lose the ability to statically determine an instruction footprint (at least in an easy way). It also just seems like the easier thing to say that a single sail `register` represents the unit of atomicity for dependency tracking. Anything smaller we'd have to pick some point between that and tracking dependencies at the level of individual bits which would be impractical.

jrtc27:
Oh sorry, I glanced and saw `vector(regtype, ...)` and interpreted that as vector-of-register, but that's not true, it's `register ... : vector(regtype, ...)` so the whole thing is the register. My bad for not fully reading the original issue.

PR #646, opened 2024-12-17
Support for the E base architecture

nwf:
Allow configuring `regidx` to be 4 bits instead of 5, so only 16 `x` registers are supported. Instruction encodings to access `x16` - `x31` are reserved and will cause an illegal instruction exception if `E` is enabled.

### DIFF
--- a/model/riscv_regs.sail
+++ b/model/riscv_regs.sail
@@ -11,6 +11,15 @@
 register PC       : xlenbits
 register nextPC   : xlenbits
 
+/* Mappings for encoding */
+
+mapping encdec_reg : regidx <-> bits(5) = {
+  forwards Regidx(r) => zero_extend(r),
+  backwards r if not(base_E_enabled) | r[4] == bitzero => Regidx(r[regidx_bit_width - 1 .. 0]),
+}
+
+mapping encdec_creg : cregidx <-> bits(3) = { Cregidx(r) <-> r }
+
 /* mappings for assembly */
 
 mapping reg_abi_name_raw : bits(5) <-> string = {
@@ -84,9 +93,11 @@ mapping reg_arch_name_raw : bits(5) <-> string = {
     0b11111 <-> "x31",
 }
 
+// Mapping from 4/5 back to 5 bits here makes the name mapping much easier.
+// If you call `reg_name("t6")` on an E hart you'll get a mapping failure.
 mapping reg_name : regidx <-> string = {
-  Regidx(i) if get_config_use_abi_names()      <-> reg_abi_name_raw(i),
-  Regidx(i) if not(get_config_use_abi_names()) <-> reg_arch_name_raw(i),
+  encdec_reg(i) if get_config_use_abi_names()      <-> reg_abi_name_raw(i),
+  encdec_reg(i) if not(get_config_use_abi_names()) <-> reg_arch_name_raw(i),
 }
 
 // Special mapping for the sp register for use in assembly for sp-relative
@@ -96,7 +107,7 @@ mapping sp_reg_name : unit <-> string = {
   () if not(get_config_use_abi_names()) <-> "x2",
 }
 
-mapping creg_name : cregidx <-> string = { Cregidx(i) <-> reg_name(Regidx(0b01 @ i)) }
+mapping creg_name : cregidx <-> string = { Cregidx(i) <-> reg_name(encdec_reg(0b01 @ i)) }
 
 function xreg_write_callback(reg : regidx, value : xlenbits) -> unit = {
   let name = reg_name(reg);
@@ -236,10 +247,3 @@ function wX_pair_bits(i : regidx, data : bits(xlen * 2)) -> unit =
   }
 
 overload X_pair = {rX_pair_bits, wX_pair_bits}
-
-
-/* Mappings for encoding */
-
-mapping encdec_reg : regidx <-> bits(5) = { Regidx(r) <-> r }
-
-mapping encdec_creg : cregidx <-> bits(3) = { Cregidx(r) <-> r }
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -275,11 +275,14 @@ function reset_misa() -> unit = {
   misa[A]   = bool_to_bits(hartSupports(Ext_A));   /* atomics */
   misa[C]   = bool_to_bits(hartSupports(Ext_C));   /* RVC */
   misa[B]   = bool_to_bits(hartSupports(Ext_B));   /* Bit-manipulation */
-  misa[I]   = 0b1;                                 /* base integer ISA */
   misa[M]   = bool_to_bits(hartSupports(Ext_M));   /* integer multiply/divide */
   misa[U]   = bool_to_bits(hartSupports(Ext_U));   /* user-mode */
   misa[S]   = bool_to_bits(hartSupports(Ext_S));   /* supervisor-mode */
   misa[V]   = bool_to_bits(hartSupports(Ext_V));   /* vector extension */
+
+  // Base integer ISA (disabled if we only support the 16-register E base).
+  misa[E] = bool_to_bits(base_E_enabled);
+  misa[I] = ~(misa[E]);
 
   if   hartSupports(Ext_F) & hartSupports(Ext_Zfinx)
   then internal_error(__FILE__, __LINE__, "F and Zfinx cannot both be enabled!");
--- a/model/riscv_types.sail
+++ b/model/riscv_types.sail
@@ -17,28 +17,37 @@ type instbits = bits(32)
 type pagesize_bits : Int = 12
 let  pagesize_bits = sizeof(pagesize_bits)
 
+// RV32E and RV64E Base Integer Instruction Sets. If this enabled then
+// the model will only support E (16 integer registers). It is allowed
+// to have a design where `misa[I]` is writable to turn E on and off,
+// but this model does not support that.
+type base_E_enabled : Bool = config base.E
+let base_E_enabled = constraint(base_E_enabled)
+
 /* register identifiers */
 
-newtype regidx = Regidx : bits(5)  /* uncompressed register identifiers */
+type regidx_bit_width = if base_E_enabled then 4 else 5
+let regidx_bit_width = sizeof(regidx_bit_width)
+
+newtype regidx = Regidx : bits(regidx_bit_width)  /* uncompressed register identifiers */
 newtype cregidx = Cregidx : bits(3) /* identifiers in RVC instructions */
 type csreg = bits(12)  /* CSR addressing */
 
-function regidx_offset(Regidx(r) : regidx, o : bits(5)) -> regidx = Regidx(r + o)
+function regidx_offset(Regidx(r) : regidx, o : bits(regidx_bit_width)) -> regidx = Regidx(r + o)
 function regidx_offset_range(Regidx(r) : regidx, o : range(0, 31)) -> regidx = Regidx(r + o)
-function regidx_bits(Regidx(b) : regidx) -> bits(5) = b
+function regidx_bits(Regidx(b) : regidx) -> bits(regidx_bit_width) = b
 overload operator + = { regidx_offset, regidx_offset_range }
 /* register file indexing */
 
-newtype regno = Regno : range(0, 31)
+newtype regno = Regno : range(0, 2 ^ regidx_bit_width - 1)
 
 /* mapping RVC register indices into normal indices */
-val creg2reg_idx : cregidx -> regidx
-function creg2reg_idx(Cregidx(i)) = Regidx(0b01 @ i)
+function creg2reg_idx(Cregidx(i) : cregidx) -> regidx = Regidx(zero_extend(0b1 @ i))
 
 /* some architecture and ABI relevant register identifiers */
-let zreg : regidx = Regidx(0b00000) /* x0, zero register  */
-let ra   : regidx = Regidx(0b00001) /* x1, return address */
-let sp   : regidx = Regidx(0b00010) /* x2, stack pointer  */
+let zreg : regidx = Regidx(zero_extend(0b00)) /* x0, zero register  */
+let ra   : regidx = Regidx(zero_extend(0b01)) /* x1, return address */
+let sp   : regidx = Regidx(zero_extend(0b10)) /* x2, stack pointer  */
 
 /* base architecture definitions */
 
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ For booting operating system images, see the information under the
 ### The Sail specification currently captures the following ISA extensions and features:
 
 - RV32I and RV64I base ISAs, v2.1
+- RV32E and RV64E base ISAs, v2.0
 - Zifencei extension for instruction-fetch fence, v2.0
 - Zicsr extension for CSR instructions, v2.0
 - Zicntr and Zihpm extensions for counters, v2.0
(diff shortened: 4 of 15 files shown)


Timmmm:
It would be really nice to have this as an "initialisation time" option rather than yet another compile time option. I think @Alasdair 's new config system will take care of `regidx`. For the other stuff in `riscv_regs_e.sail` can you keep a single `riscv_regs.sail` and use runtime checks and add an `extensionEnabled(Ext_E)`? E.g.

```

function rX (regno(r) : regno) -> xlenbits = {
  let v : regtype =
    match r {
      0 => zero_reg,
      1 => x1,
      2 => x2,
      3 => x3,
      4 => x4,
      5 => x5,
      6 => x6,
      7 => x7,
      8 => x8,
      9 => x9,
      10 => x10,
      11 => x11,
      12 => x12,
      13 => x13,
      14 => x14,
      15 => x15,
      16 if not(extensionEnabled(Ext_E)) => x16,
      17 if not(extensionEnabled(Ext_E)) => x17,
...
      _  => {assert(false, "invalid register number"); zero_reg}
    };
  regval_from_reg(v)
}
```

I think there's no harm having `x16` etc still existing, even if E is enabled.

Timmmm:
That also makes it easier to support runtime enabling/disabling of E, by writing `misa.I`.

Timmmm:
Although... if we do support that then you can't change the `regidx` type. Hmm.

Alasdair:
We could have a split between a compile-time only E build where regidx is 4 bits and a runtime E mode where it just restricts the set of available GPRs. The specification does specifically mention that E can be enabled and disabled by writing misa[I] so we do need to support the second option.

nwf:
Certainly the compile-time E-only option is a more attractive base for CHERIoT, and I could imagine utility for a compile-time I-only option (as in this PR), but maybe that's less useful. I think SWIM gets to do the dynamic E/I case, but I would love to see it done and hopefully this PR is at least not a terrible place to start.

Timmmm:
Is there any point having the compile-time version if we support the runtime version though and make `misa[I]` read only?

I guess it helps with the formal and SV backends?

> I think SWIM gets to do the dynamic E/I case

What's SWIM?

Alasdair:
I think we should get this in when it gets rebased. My only nitpick is I think constructors should probably be capitalised so `newtype regno = Regno : ...`.

jordancarlin:
Now that #617 and #645 are merged, this would be good to rebase and revisit.

nwf:
Here (1c9dcbb2) is a rebased spin that duplicates a lot less code by making the (x) register file a sail `register` of type `vector('n, regtype)` (for fixed `'n`) rather than many separate `register`s of type `regtype`.

I don't _quite_ think that this can all be keyed off of a single `config` type-level value, but this version is surely _closer_ and I'm happy to be wrong about the remaining gap.

jordancarlin:
See #409

Alasdair:
What you can have is a vector of references to registers, like:
```
register X0 : bits(32)
register X1 : bits(32)

let xregs : vector(2, register(bits(32))) = [ref X1; ref X0]
```
Then assign like:
```
*xregs[0] = <some value>
```

nwf:
That's quite nice -- I hadn't known that `sail` had `ref` types at all -- and allows for some commoning up, but it still requires the raft of `register` declarations... but maybe that can live in the common `_regs.sail` along with the accessor functions and the arrays of references can live in the `_regs_[ie].sail`?  Will ponder.

Separately, are there likely to be simulation slowdowns from using the array of refs to access the register state, or can/will the distinction be compiled away given that the array is constant?

Doubly-separately, I had to rewrite `match r { 0 => (), r => ... }` as `if r == 0 then () else ...`, as the latter form seems to do liquid type propagation (if `r` initially had type `range(0,15)`, it has type `range(1,15)` in the `else` branch) but the former does not?

Alasdair:
At least in Sail to C, it shouldn't cause any significant slowdown.

I guess you had to do that if-then-else to handle the zero register case? Would something like:
```
if r == 0 then {
  <zero register case>
} else match r {
  1 => ...
  2 => ...
}
```
work? match won't infer constraints from previous conditions, as it's not the case that everything match does can be simplified down to a boolean constraint the type system would handle.

Timmmm:
A whole feature of Sail I didn't know exists! I actually thought it was free of reference types intentionally. Maybe worth documenting? Although we seem to have managed without it without any problems so far...

Alasdair:
You can only create references to registers, so it's very restricted. You could do the same thing with an enumeration of register names and some helper functions so it isn't adding much semantically just a bit of occasional convenience.

In Isla there are a few functions taking `register('a)` which is useful for passing a register name to an external function, which is used to tweak the dependency analysis for the axiomatic memory model.

model/CMakeLists.txt
```diff
@@ -143,6 +143,7 @@ foreach (xlen IN ITEMS 32 64)
                 "riscv_reg_type.sail"
                 "riscv_freg_type.sail"
                 "riscv_regs.sail"
+                "riscv_regs_i.sail"
```
arichardson, line 146: Could add something like 
`option(RVE "Build for RV32E/RV64E as the base isa instead of RV32I/RV64I" OFF)`
```suggestion
                "riscv_regs_$<IF:$<BOOL:RVE>,e,i>.sail"
```
Untested, not sure if the `$<BOOL:` is needed.

nwf: AFAICT, the `$<BOOL:` is necessary, and the full form is `$<IF:$<BOOL:${RVE}>,e,i>`.  (And people used to give Perl a hard time for being line noise... yeesh.)

nwf:
OK, here's a spin with the registers back to being registers and not a vector thereof (thanks @jordancarlin!), but they're still exposed to the core of the model via an array of references (thanks @Alasdair!) to cut down on duplication.  I've also folded in @arichardson's `cmake` suggestion (thanks!) and have
- built both `I` and `E` versions locally
- tested that I haven't broken `I` (that is, `ctest` passes its 392 test cases)

model/riscv_regs.sail
```diff
@@ -16,75 +16,16 @@ register instbits : xlenbits
...
-      25 => x25,
-      26 => x26,
-      27 => x27,
-      28 => x28,
-      29 => x29,
-      30 => x30,
-      31 => x31,
-      _  => {assert(false, "invalid register number"); zero_reg}
-    };
+  let v : regtype = if r == 0 then zero_reg else __deref(xreg_by_index(r));
```
pmundkur, line 28: @Alasdair Would it be possible to have a less forbidding name for `__deref` in the Sail library?  It seems useful for non Sail-internal uses.

pmundkur (approved):
This is neat!

pmundkur:
@nwf  Should 'DNM' be removed from the title of this PR?

nwf:
It almost surely still needs review and work, but I suppose it's moved from "a quick sketch" to "maybe this is something we might consider merging", so sure.

pmundkur:
Is there any reason not to merge this?  There's the nit about `__deref` but that could be fixed in the next Sail release?
cc @Alasdair @Timmmm @jordancarlin @arichardson

model/riscv_regs.sail
jrtc27: Is there a reason why it's `*ref = v` but `let v = __deref(ref)`? Would `*` for both not be more natural (just as you can do `X(rd) = v` and `let v = X(rs1)` today)?

Alasdair: `let v = *ref` should work, it's just syntax sugar for `__deref`.

jordancarlin:
My main concern is adding another compile time option. We’ve been doing a lot of work to make things run time configurable, and there are plans to make both xlen/flen and rvfi run time options. That would give us a single unified simulator. This seems like a step backwards from that goal.

nwf:
I'm certainly not opposed to someone doing the dynamic I/E design -- probably it means that the `encdec_reg` `mapping` becomes dependent on `misa.[IE]`, and I think that makes instruction decoding at least moderately exciting(?) -- but as I said above:

> I think Someone Who Isn't Me gets to do the dynamic E/I case, but I would love to see it done

My motivation for this PR was better modeling of CHERIoT, and that is (quite likely) always going to be a RV32E-only extension, and so while I understand the desire to support I and E in the same binary in general, it's not something I think I can justify spending time on at the moment.

pmundkur:
I don't see why we could not add a static E in for now, and then fix it to be dynamic at the same time as xlen/flen when that fix is ready?

Timmmm:
I'll see if I can rebase this and make it work with E being a configure-type option like xlen in #870.

nwf:
Please do.  I apologize for my computer's misbehavior this morning... apparently Zoom and Firefox just _do not_ get along at the moment.

The question I was attempting to ask in the meeting was "what will moving to a configure-type option do to the definitions of `regidx_bits_width` and `encdec_reg`?"  That is, will `regidx_bits_width` always be 6 or can it key on the configure option?  Can `encdec_reg` use `if` (or `when`) to key its match arms on the E/I configuration knob?

Timmmm:
Ok I've rebased it and then changed it to use `config base.E`. It compiles but I didn't test it any further.

Also this version leaves all of `x16..x31` in the model even with `config base.E` enabled - they are present but inaccessible. It doesn't affect the C/OCaml backends but I'm not sure about the others (Lean, SV etc)?

Leaving them in does make it easier to support writable `misa[I]` though. I think it's actually pretty easy to add that. All you need is to change this line:

```
  backwards r if not(base_E_enabled) | r[4] == bitzero => Regidx(r[regidx_bit_width - 1 .. 0]),
```

to

```
  backwards r if misa[I] == 0b1 | r[4] == bitzero => Regidx(r[regidx_bit_width - 1 .. 0]),
```

and then update `legalize_misa()` to allow writing it. But I'm pretty doubtful there any use cases for writable `misa[I]`.

Also, that reminds me we both forgot to update `legalize_misa()`. I'll fix it now.

Timmmm:
Oh also just in case you missed it, this requires a recent build of the Sail compiler (so it's going to fail CI).

Timmmm:
> All you need is to change this line:

Oh and the reason I didn't do this, even though it is easy, is because `misa` is defined after that code and I couldn't be bothered to deal with the reordering of dependencies...

model/riscv_types.sail
```diff
@@ -26,29 +26,41 @@ type instbits = bits(32)
 type pagesize_bits : Int = 12
 let  pagesize_bits = sizeof(pagesize_bits)
 
+// RV32E and RV64E Base Integer Instruction Sets. If this enabled then
+// the model will only support E (16 integer registers). It is allowed
+// to have a design where `misa[I]` is writable to turn E on and off,
+// but this model does not support that.
+type base_E_enabled : Bool = config base.E
+let base_E_enabled = constraint(base_E_enabled)
```
jordancarlin, line 34: Seems like it would be more consistent with the rest of the model to put this is the extensions file and use `hartSupports` (similar to what was done for D).

arichardson:
> Leaving them in does make it easier to support writable `misa[I]` though. I think it's actually pretty easy to add that. All you need is to change this line:
> 
> ```
>   backwards r if not(base_E_enabled) | r[4] == bitzero => Regidx(r[regidx_bit_width - 1 .. 0]),
> ```
> 
> to
> 
> ```
>   backwards r if misa[I] == 0b1 | r[4] == bitzero => Regidx(r[regidx_bit_width - 1 .. 0]),
> ```
> 
> and then update `legalize_misa()` to allow writing it. But I'm pretty doubtful there any use cases for writable `misa[I]`.
> 
> Also, that reminds me we both forgot to update `legalize_misa()`. I'll fix it now.

I agree that having it writable would be unlikely to be supported anywhere. I initially thought the spec forced misa.E to be fixed, but it looks like it does actually allow changing it (just not directly, you have to clear I instead).

> The "E" bit is read-only. Unless misa is all read-only zero, the "E" bit always reads as the complement of the "I" bit. If an execution environment supports both RV32E and RV32I, software can select RV32E by clearing the "I" bit.

model/riscv_types.sail
Timmmm: Maybe, but E isn't actually an extension - it's a different base architecture.

Timmmm:
Rebased. One unfortunate side effect is that register index bit-vectors are now `sbits`. It's probably ok though.

Do we have any way of testing this?

nwf:
I'm confused about the need for `sbits`... nothing about `xreg_write_callback` has changed in this PR, so why does `xreg_full_write_callback` need to change?  If this is a separate change to catch up to the `sail` compiler, can it please be done as a separate commit (and probably separate PR)?

Timmmm:
Nothing in `xreg_write_callback` has changed but one of its parameters is `regidx` and we changed that from `bits(5)` to basically `bits(4 or 5)`. The Sail compiler no longer statically knows how many bits are in a `regidx` and the way it deals with that currently is to use `sbits` everywhere (which has an extra `.len` field to say how many bits there are).

nwf:
> Nothing in `xreg_write_callback` has changed but one of its parameters is `regidx` and we changed that from `bits(5)` to basically `bits(4 or 5)`. The Sail compiler no longer statically knows how many bits are in a `regidx` and the way it deals with that currently is to use `sbits` everywhere (which has an extra `.len` field to say how many bits there are).

Ah, I see!  Should we instead do a conversion from `regidx` to `bits(5)` in `xreg_write_callback`?

Timmmm:
Yeah we could do, or `Regidx` which would always be `uint64_t`. I don't think it would be any better though really - it would just do the conversation in the generated Sail code instead.

model/riscv_sys_control.sail
```diff
@@ -275,12 +275,15 @@ function reset_misa() -> unit = {
...
   misa[B]   = bool_to_bits(hartSupports(Ext_B));   /* Bit-manipulation */
-  misa[I]   = 0b1;                                 /* base integer ISA */
   misa[M]   = bool_to_bits(hartSupports(Ext_M));   /* integer multiply/divide */
   misa[U]   = bool_to_bits(hartSupports(Ext_U));   /* user-mode */
   misa[S]   = bool_to_bits(hartSupports(Ext_S));   /* supervisor-mode */
   misa[V]   = bool_to_bits(hartSupports(Ext_V));   /* vector extension */
 
+  // Base integer ISA (disabled if we only support the 16-register E base).
+  misa[E] = bool_to_bits(base_E_enabled);
+  misa[I] = ~(misa[E]);
```
arichardson, line 285:
```suggestion
  misa[I] = bool_to_bits(not(base_E_enabled));
```

Would be consistent with the use below. Or use ~misa[E] in both cases?

nwf:
> I don't think it would be any better though really - it would just do the conversation in the generated Sail code instead.

It's probably worth benchmarking the wall-clock time for the executable model... passing in a `struct` might get optimized a bit by SROA, but passing in a `uint64_t` is _probably_ faster?  The conversion in the Sail-generated code has the advantage of having more type information, so is probably zero instructions?

Timmmm:
Yeah it probably is slightly slower; unfortunately the run time at the moment is dominated by [suboptimal codegen in the PMP code](https://github.com/rems-project/sail/issues/1367) so it probably won't be measurable. I think the relative slowness of `fbits`, `sbits` and `lbits` is something like 1, 2, 100 so mostly improving performance is a matter of eliminating `lbits`.

model/riscv_sys_control.sail
jordancarlin: ~misa[E] is more consistent with the spec which species misa[I] is always the complement of misa[E].

jordancarlin:
This also needs an update to the README.

pmundkur:
@Timmmm @nwf I rebased and added this to the release notes, hope that's okay.

nwf:
`202412` branch deleted!  :tada:  Thank you to everyone for all your help!

PR #649, opened 2024-12-19
Remove obsolete z_set_Misa functions

Timmmm:
This appears to be an obsolete and unused mechanism of disabling extensions by calling into the model to modify `misa`. Now it is done by `sys_enable_` callbacks.


jordancarlin (approved):
LGTM

PR #650, opened 2024-12-19
Remove c2 Makefile target

Timmmm:
I don't believe this backend exists any more.


Alasdair (approved):
Yeah it's not a thing any more. I need to find the time to merge some of it's features into the regular C generation.

Issue #639, opened 2024-12-10
Simplify handling of mstatush


Timmmm:
`mstatush` is not really handled correctly at the moment. The `Mstatus` type is:

```

bitfield Mstatus : xlenbits = {
  SD   : xlen - 1,

  // The MBE and SBE fields are in mstatus in RV64 and absent in RV32.
  // On RV32, they are in mstatush, which doesn't exist in RV64.  For now,
  // they are handled in an ad-hoc way.
  // MBE  : 37
  // SBE  : 36

  // The SXL and UXL fields don't exist on RV32, so they are modelled
  // via explicit getters and setters; see below.
  // SXL  : 35 .. 34,
  // UXL  : 33 .. 32,

  TSR  : 22,
...
  SIE  : 1,
}

register mstatus : Mstatus
```

There *is* an `mstatush` register:

```
bitfield Mstatush : bits(32) = {
  MBE  : 5,
  SBE  : 4
}
register mstatush : Mstatush
```

But it is never modified and we ignore writes to `mstatush` (because the model doesn't support big endian.

This is quite awkward. For other 64-bit registers, e.g. `menvcfg` the `h` version is just a view into the top 32 bits of the 64-bit value (search for `Upper 32 bits of ` in the priv spec and you see a load of these). That is easy to handle because we just have a single `bitfield Menvcfg : bits(64)` and then read/write the relevant parts in the CSR access functions.

However `mstatus` is very annoyingly not *quite* like that because the `SD` bit is always at `xlen - 1`. It is in a different position on RV32 and RV64.

The best solution I can think of is this:

```
bitfield Mstatus : bits(64) = {
  SD_64   : 63, // SD bit for RV64.

  MBE  : 37
  SBE  : 36

  SXL  : 35 .. 34,
  UXL  : 33 .. 32,

  SD_32 : 31, // SD bit for RV32.

  TSR  : 22,
...
  SIE  : 1,
}
```

And then you have to do something like:

```
let sd = mstatus[if xlen == 32 then SD_32 else SD_64];
```

We can put that in a function.

I think that is better than the current situation, and it would remove the need to have getters/setters for MBE/SBE/SXL/UXL.

Timmmm:
For reference, [here are the latest fields for the registers](https://riscv-specs.timhutt.co.uk/spec/riscv-isa-release-89e3947-2024-12-10/riscv-privileged.html#_machine_status_mstatus_and_mstatush_registers).

[...]

PR #652, opened 2024-12-20
refactor mstatus handler function

KotorinMinami:
As mentioned in #639, and there are also some legitimacy issues in legalize_mstatus, so this PR is intended to help make corrections

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -370,16 +370,17 @@ function init_sys() -> unit = {
                 then bool_to_bits(sys_enable_fdext())  /* double-precision */
                 else 0b0;
 
-  mstatus = set_mstatus_SXL(mstatus, misa[MXL]);
-  mstatus = set_mstatus_UXL(mstatus, misa[MXL]);
-  mstatus[SD]   = 0b0;
-  mstatus[MPP] = privLevel_to_bits(lowest_supported_privLevel());
-
-  /* set to little-endian mode */
-  if xlen == 64 then {
-    mstatus = Mk_Mstatus([mstatus.bits with 37 .. 36 = 0b00])
-  };
-  mstatush.bits = zeros();
+  mstatus = [mstatus with
+    SD_64 = 0b0,
+    SD_32 = 0b0,
+    /* SXL and UXL is only valid for 64-bit systems and in RV32 WPRI */
+    SXL = misa[MXL],
+    UXL = misa[MXL],
+    MPP = privLevel_to_bits(lowest_supported_privLevel()),
+    /* set to little-endian mode */
+    MBE = 0b0,
+    SBE = 0b0,
+  ];
 
   mip.bits     = zeros();
   mie.bits     = zeros();
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -167,26 +167,25 @@ function have_privLevel(priv : priv_level) -> bool =
     0b11 => true,
   }
 
-bitfield Mstatush : bits(32) = {
-  MBE  : 5,
-  SBE  : 4
-}
-register mstatush : Mstatush
+bitfield Mstatus : bits(64) = {
+  SD_64: 63,
 
-bitfield Mstatus : xlenbits = {
-  SD   : xlen - 1,
+  //MDT  : 42,
+  //MPELP: 41,
 
-  // The MBE and SBE fields are in mstatus in RV64 and absent in RV32.
-  // On RV32, they are in mstatush, which doesn't exist in RV64.  For now,
-  // they are handled in an ad-hoc way.
-  // MBE  : 37
-  // SBE  : 36
+  //MPV  : 39,
+  //GVA  : 38,
 
-  // The SXL and UXL fields don't exist on RV32, so they are modelled
-  // via explicit getters and setters; see below.
-  // SXL  : 35 .. 34,
-  // UXL  : 33 .. 32,
+  MBE  : 37,
+  SBE  : 36,
 
+  SXL  : 35 .. 34,
+  UXL  : 33 .. 32,
+
+  SD_32: 31,
+
+  //SDT  : 24,
+  //SPELP: 23,
   TSR  : 22,
   TW   : 21,
   TVM  : 20,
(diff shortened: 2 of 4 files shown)


model/riscv_sys_regs.sail
```diff
@@ -220,81 +219,80 @@ function get_mstatus_SXL(m : Mstatus) -> arch_xlen = {
...
-  else {
-    let m = vector_update_subrange(m.bits, 35, 34,  a);
-    Mk_Mstatus(m)
-  }
-}
-
 function get_mstatus_UXL(m : Mstatus) -> arch_xlen = {
   if   xlen == 32
   then arch_to_bits(RV32)
   else m.bits[33 .. 32]
```
Timmmm, line 225:
```suggestion
  else m[UXL]
```

model/riscv_sys_regs.sail
```diff
@@ -220,81 +219,80 @@ function get_mstatus_SXL(m : Mstatus) -> arch_xlen = {
...
-function clause read_CSR(0x300) = mstatus.bits
-function clause read_CSR(0x310 if xlen == 32) = mstatush.bits
+function clause read_CSR(0x300) = mstatus.bits[xlen - 1 .. 0]
+function clause read_CSR(0x310 if xlen == 32) = mstatus.bits[63 .. 32]
 
-function clause write_CSR(0x300, value) = { mstatus = legalize_mstatus(mstatus, value); mstatus.bits }
-function clause write_CSR((0x310, value) if xlen == 32) = { mstatush.bits } // ignore writes for now
+function clause write_CSR((0x300, value) if xlen == 64) = { mstatus = legalize_mstatus(mstatus, value); mstatus.bits }
+function clause write_CSR((0x300, value) if xlen == 32) = { mstatus = legalize_mstatus(mstatus, mstatus.bits[63 .. 32] @ value); mstatus.bits[31 .. 0] }
+function clause write_CSR((0x310, value) if xlen == 32) = { mstatus.bits[63 .. 32] } // ignore writes for now
```
Timmmm, line 295:
```suggestion
function clause write_CSR((0x310, value) if xlen == 32) = { mstatus = legalize_mstatus(mstatus, value @ mstatus.bits[31 .. 0]); mstatus.bits[63 .. 32] }
```

model/riscv_vext_regs.sail
```diff
@@ -79,7 +79,8 @@ mapping vreg_name = {
 function dirty_v_context() -> unit = {
   assert(sys_enable_vext());
   mstatus[VS] = extStatus_to_bits(Dirty);
-  mstatus[SD] = 0b1
+  if xlen == 32 then mstatus[SD_32] = 0b1
```
Timmmm, line 82: Maybe add a `set_mstatus_sd()` function since there's already a `get` version and you use this twice.

model/riscv_sys_regs.sail
```diff
@@ -220,81 +219,80 @@ function get_mstatus_SXL(m : Mstatus) -> arch_xlen = {
   else m.bits[35 .. 34]
```
Timmmm, line 219:
```suggestion
  else m[SXL]
```

I think in future I think we can just use `m[SXL]` directly and initialise it to the RV32 value on RV32: https://github.com/riscv/sail-riscv/pull/597/files#diff-e1a3fe7050c3af4c308a5e39bfb7f0f31094022b9892cee9a611bcb57ccb6726R283

model/riscv_sys_regs.sail
```diff
@@ -673,39 +671,34 @@ function clause read_CSR(0xF15) = mconfigptr
...
-}
-
 function lower_mstatus(m : Mstatus) -> Sstatus = {
   let s = Mk_Sstatus(zeros());
-  let s = set_sstatus_UXL(s, get_mstatus_UXL(m));
+  let sd = get_mstatus_SD(m);
 
   [s with
-    SD = m[SD],
+    SD_64 = if xlen == 64 then sd else 0b0,
```
Timmmm, line 697: I think I would just do `SD_64 = m[SD_64]` and `SD_32 = m[SD_32]`. The rest of the code should ensure that they don't get set incorrectly.

model/riscv_sys_control.sail
```diff
@@ -370,16 +370,17 @@ function init_sys() -> unit = {
...
-  if xlen == 64 then {
-    mstatus = Mk_Mstatus([mstatus.bits with 37 .. 36 = 0b00])
-  };
-  mstatush.bits = zeros();
+  mstatus = [mstatus with
+    SD_64 = 0b0,
+    SD_32 = 0b0,
+    /* SXL and UXL is only valid for 64-bit systems and in RV32 WPRI */
+    SXL = if xlen == 64 then misa[MXL] else mstatus[SXL],
+    UXL = if xlen == 64 then misa[MXL] else mstatus[UXL],
```
Timmmm, line 378: I think just set these to `S/UXL = misa[MXL]`. They are non-reset, and #597 moves the initialisation to a more sensible place so the reset function doesn't have to touch them at all.

model/riscv_sys_regs.sail
```diff
@@ -220,81 +219,80 @@ function get_mstatus_SXL(m : Mstatus) -> arch_xlen = {
...
    * that does not have a matching bitfield entry. All bits above 32 are handled
    * explicitly later.
    */
-  // TODO: This method is error prone. We should instead explicitly copy the
-  // fields that are implemented. See legalize_mie() for an example.
-  let m : Mstatus = Mk_Mstatus(zero_extend(v[22 .. 7] @ 0b0 @ v[5 .. 5] @ 0b0 @ v[3 .. 3] @ 0b0 @ v[1 .. 1] @ 0b0));
+  let m : Mstatus = Mk_Mstatus(v);
 
+  // Set dirty bit to OR of other status bits.
+  let dirty = extStatus_of_bits(m[FS]) == Dirty |
```
Timmmm, line 242: This is subtly incorrect - `XS` is read-only zero but if you write `Dirty` to it you'll still set `SD`. `FS` and `VS` can be read-only zero in some cases too.

I would do it in two stages. First set everything except SD. Then set SD based on the new values.

model/riscv_sys_regs.sail
```diff
@@ -220,81 +219,80 @@ function get_mstatus_SXL(m : Mstatus) -> arch_xlen = {
...
+              extStatus_of_bits(m[XS]) == Dirty |
+              extStatus_of_bits(m[VS]) == Dirty;
+
+  [o with
+    SD_64 = if xlen == 64 then bool_to_bits(dirty) else o[SD_64],
+    SD_32 = if xlen == 32 then bool_to_bits(dirty) else o[SD_32],
+    MDT = m[MDT],
```
Timmmm, line 249: I think we shouldn't add fields and make them writeable if we don't actually support them. I would probably just leave these fields out entirely for now. It's less confusing if we add them when we actually implement their functionality.

model/riscv_sys_regs.sail
```diff
@@ -220,81 +219,80 @@ function get_mstatus_SXL(m : Mstatus) -> arch_xlen = {
...
-  /* We don't support dynamic changes to SXL and UXL. */
-  let m = set_mstatus_SXL(m, get_mstatus_SXL(o));
-  let m = set_mstatus_UXL(m, get_mstatus_UXL(o));
-
-  /* We don't currently support changing MBE and SBE. */
-  if xlen == 64 then {
-    Mk_Mstatus([m.bits with 37 .. 36 = 0b00])
-  } else m
+    MPP = if have_privLevel(m[MPP]) then m[MPP] else privLevel_to_bits(lowest_supported_privLevel()),
+    VS = m[VS],
```
Timmmm, line 276: Hmm I think we should making `FS` and `VS` read-only zero in some cases. That's an existing bug though - I made an issue: #661

model/riscv_sys_regs.sail
```diff
@@ -220,81 +219,80 @@ function get_mstatus_SXL(m : Mstatus) -> arch_xlen = {
...
+function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
   /*
    * Populate all defined fields using the bits of v, stripping anything
    * that does not have a matching bitfield entry. All bits above 32 are handled
    * explicitly later.
    */
-  // TODO: This method is error prone. We should instead explicitly copy the
-  // fields that are implemented. See legalize_mie() for an example.
-  let m : Mstatus = Mk_Mstatus(zero_extend(v[22 .. 7] @ 0b0 @ v[5 .. 5] @ 0b0 @ v[3 .. 3] @ 0b0 @ v[1 .. 1] @ 0b0));
+  let m : Mstatus = Mk_Mstatus(v);
```
Timmmm, line 239:
```suggestion
  let m = Mk_Mstatus(v);
```

Timmmm (changes requested):

KotorinMinami:
@Timmmm Is there anything else that needs improvement in my code?

Timmmm (approved):
LGTM - would be nice to get a second opinion though. Any takers @arichardson @jordancarlin etc?

arichardson (approved):
This looks like an improvement to me. Nit: not sure we need to add all the future fields in this PR but I don't mind keeping them here.

model/riscv_sys_regs.sail
```diff
@@ -217,84 +216,92 @@ function effectivePrivilege(t : AccessType(ext_access_type), m : Mstatus, priv :
...
+  then [m with SD_32 = v]
+  else [m with SD_64 = v]
 }
 
-function legalize_mstatus(o : Mstatus, v : xlenbits) -> Mstatus = {
+function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
   /*
    * Populate all defined fields using the bits of v, stripping anything
    * that does not have a matching bitfield entry. All bits above 32 are handled
    * explicitly later.
```
jordancarlin, line 243: Comment needs to be updated. All bits are now handled directly in this function.

jordancarlin (approved):
Minor nit in a comment but otherwise LGTM

model/riscv_sys_regs.sail
KotorinMinami: Thanks! I have edited this commit. What do you think of the commit now?

jordancarlin: Seems fine now

Issue #182, opened 2022-09-12
switch to disable tlbs


neelgala:
I believe there is some sort of a TLB like structure used in SAIL which prevents PTW to be triggered for each memory/instruction access. It would be beneficial if we can have a switch via the cli to disable the TLBs so that all memory/instruction accesses undergo a PTW by default.

This requirement comes from the SIG-ACT with regards to capturing ISA coverage for tests dealing with translation schemes.

Timmmm:
Can you explain more why this is needed (assuming it still is). If tests want to guarantee a page table walk can't they just do an sfence.vma? Doesn't seem like the ACTs should rely on microarchitectural details about the TLB.

allenjbaum:
Good question.  ACTs are trying to avoid microarchitectural details (e.g the presence of a TLB or it's organization and behavior), not rely on them.  I think it was to ensure that certain implicit accesses were made properly (i.e.the full table walk). Yes, we could modify the tests to add the fence op, but it would be a lot of work. 
If this has already been implemented, I'd rather it continue for now, and I'll file an issue to add the fence.vma before every test access for the VMM tests.  And

Timmmm:
I think I'd be more in favour of just fixing the tests, but I'm not sure I fully understand the issue. I'll tag this for discussion in the meeting.

Timmmm:
I think this is probably obsolete with ACT4. Please comment or open a new issue if there's a good reason to do this.

PR #654, opened 2024-12-30
add disable tlb option

KotorinMinami:
As requested in the #182 , riscv-ctg needs to support the switch to disable TLB


Timmmm:
Any objection to closing this? Doesn't seem like anyone knows a reason why it should be needed.

KotorinMinami:
just do your favor

Alasdair:
Would probably need something like this if we were ever to do something similar to our Arm virtual memory semantics for RISC-V, but I don't think we have any plans to do that in the near future.

PR #658, opened 2024-12-31
Gate base counters on Zicntr extension

jordancarlin:
The cycle, time, and instret counters (and their variants) are defined by the Zicntr extension and should not be accessible without that extension. Tying Zicntr to true for now pending the new config system.

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -32,6 +32,8 @@ enum clause extension = Ext_U
 enum clause extension = Ext_Zicbom
 // Cache-Block Zero Instructions
 enum clause extension = Ext_Zicboz
+// Base Counters and Timers
+enum clause extension = Ext_Zicntr
 // Integer Conditional Operations
 enum clause extension = Ext_Zicond
 // Instruction-Fetch Fence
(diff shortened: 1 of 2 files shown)


arichardson (approved):

Timmmm (approved):

rmn30:
By coincidence I was thinking about these CSRs recently and wondered: do you need to have user mode to have the user mode aliases? I think it's potentially useful to have these even on machine mode only implementations, especially the `time` alias for `mtime`. For example see CHERIoT-Platform/cheriot-sail#95 for a couple of reasons.

Timmmm:
Hmm I think we might have got it wrong anyway actually. The CSR listing table in the spec lists then under "Unprivileged CSRs", not "User CSRs" and they are defined in the "Machine Level ISA" chapter.

Can you make an issue here so we don't forget? We should probably make an issue on the riscv-isa-manual repo to clarify this.

jrtc27:
The spec is already clear:

> Tables 2.2–2.6 list the CSRs that have currently been allocated CSR addresses. The timers, coun-
ters, and floating-point CSRs are standard unprivileged CSRs. The other registers are used by
privileged code, as described in the following chapters. Note that not all registers are required on
all implementations.

jrtc27:
(See https://github.com/riscv/riscv-isa-manual/pull/684)

Timmmm:
Clear by the usual RISC-V "language puzzle" standards maybe! What is "privileged code" or "unprivileged code"? Are those terms defined anywhere?

Timmmm:
That does seem to be what they mean though. I'll make a PR to clarify this in the manual. It's demonstrably *not* clear if two separate implementations interpreted it differently.

PR #662, opened 2025-01-02
Use vector_init for TLB

Timmmm:
This Sail feature was not available when this code was written, but now we can use it to remove the duplication.


jordancarlin (approved):
Simple cleanup

arichardson (approved):

PR #665, opened 2025-01-02
Update LICENCE copyright years for 2025

jordancarlin:


Timmmm:
IMO we should remove the years so we don't have to do this busy work. I've looked into the legalities of it in the past and there's no legal reason to have them. The purpose of the date is to provide evidence of authorship date in case of copyright disputes where the copier claims the work is out of copyright.

However that is not relevant for computer code because:

1. We have Git which is a much better record of when things were written.
2. Sail didn't even exist until 2015 so it's in copyright until at least 2110, and unlikely to be relevant by then!

But we can just update it for now if you like.

Timmmm (approved):

jordancarlin:
I'm fine with removing the years entirely, but we should definitely do one of the two. Leaving it at 2024 doesn't make a whole lot of sense.

PR #667, opened 2025-01-06
Report the correct address for CBO errors

Timmmm:
CBO faults report the address in rs1, not the cache block address.

The spec is not explicit about this yet but I have a pending PR to clarify that: https://github.com/riscv/riscv-isa-manual/pull/1433


jordancarlin (approved):

Timmmm:
Spec update has been merged.

PR #672, opened 2025-01-07
Overload operators for physaddr and virtaddr

### LINKED ISSUE #668: Define operators for physaddr and virtaddr
The new `physaddr` and `virtaddr` types are quite nice, but they do make some code uglier. For example instead of

```
handle_mem_exception(vaddr - offset), e);
```

we have to do

```
handle_mem_exception(virtaddr(virtaddr_bits(vaddr) - offset), e);
```

I think we can overload + and - (and maybe others) so that they work with `physaddr` and `virtaddr` too.

trdthg:
try resolve #668 

Added operator overloading for `physaddr` and `virtaddr`

I think we don't need to implement `==` and `!=`? 
Because `eq_anything` (I just discovered this) has already been introduced in `prelude.sail`.

https://github.com/riscv/sail-riscv/blob/c0ebea2ff14e6a39d53161c3e51dcbabe3e9a0f1/model/prelude.sail#L17

### DIFF
--- a/model/prelude_mem_addrtype.sail
+++ b/model/prelude_mem_addrtype.sail
@@ -19,3 +19,7 @@ newtype virtaddr = virtaddr : xlenbits
 function physaddr_bits(physaddr(paddr) : physaddr) -> physaddrbits = paddr
 
 function virtaddr_bits(virtaddr(vaddr) : virtaddr) -> xlenbits = vaddr
+
+function sub_virtaddr_xlenbits(virtaddr(addr) : virtaddr, offset : xlenbits) -> virtaddr = virtaddr(addr - offset)
+
+overload operator - = { sub_virtaddr_xlenbits }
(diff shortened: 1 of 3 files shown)


trdthg:
- `addr + addr` removed
- `&, |, ^` removed. I haven't found any place where these operators are used, .
- `addr+xlenbits` `addr-xlenbits` added

`physaddr + int` is used in #285

Timmmm (approved):

arichardson (approved):
This looks reasonable, could you add another commit to this PR that shows where they can be used to simplify the code?

Timmmm:
#667 is merged now, so if you rebase you can used them there.

trdthg:
> could you add another commit to this PR

done, it can avoid twice type conversion

Also, I only kept the overload for `sub_virtaddr_xlenbits`, let's do the rest of overloads when they are actually used

jordancarlin (approved):

Timmmm:
I think it was ok to keep the ones you have, but removing them is also fine. LGTM.

PR #676, opened 2025-01-09
Change *_reservation to take physaddr

### LINKED ISSUE #670: Change load_reservation, match_reservation, etc. to take physaddr
Reservation is done on physical addresses, so these should take `physaddr` not `xlenbits`.

KotorinMinami:
Fix #670, edit Sail's C language code so that the load_reservation and match_reservation functions can accept physaddr type parameters


c_emulator/riscv_sail.h
```diff
@@ -68,3 +67,13 @@ struct zMcause {
 extern struct zMcause zmcause, zscause;
 
 extern mach_bits zminstret;
+
+enum kind_zphysaddr { Kind_zphysaddr };
```
arichardson, line 71: Having to duplicate this sail-internal representation does not sound like a good idea to me. IMO the C callback should just use the underlying bits.

arichardson (changes requested):
I think this just makes it more difficult to read for no real gain. If we want to ensure that load_reservation takes a physaddr we should make it a sail function that calls the C function with the raw bits. Not sure this adds much value over the current `physaddrbits` though.

c_emulator/riscv_sail.h
Timmmm: If/when we switch to the `struct Model { #include "model.c" }` trick then we won't have to do this. But I agree, it's too error-prone until it's actually checked by the compiler.

KotorinMinami:
Emmm, @arichardson  in fact, if follow your requirements, it is indeed simpler to convert it directly to physaddrbits. However, another question is why functions like *_reservation are C functions; would it be more consistent to define them as sail functions?

arichardson:
> Emmm, @arichardson in fact, if follow your requirements, it is indeed simpler to convert it directly to physaddrbits. However, another question is why functions like *_reservation are C functions; would it be more consistent to define them as sail functions?

That is a good question - I am not familiar with this part of the model and why they are not written in sail. Maybe to allow swapping them out for something else in the concurrency model?

But I would imagine we could do the same by moving them into a separate .sail file.

KotorinMinami:
> Emmm, @arichardson in fact, if follow your requirements, it is indeed simpler to convert it directly to physaddrbits. However, another question is why functions like *_reservation are C functions; would it be more consistent to define them as sail functions?

So @Timmmm what do you think ?

Timmmm:
> why functions like *_reservation are C functions; would it be more consistent to define them as sail functions?

I presume it is because reservation is a memory thing and that is also handled outside Sail.

I think a practical reason is that if you are simulating a multicore chip then you would have a single memory in C, and you'd also need to handle reservations there. If you had reservation state in Sail then it would be duplicated for each core and that wouldn't work, e.g. multiple cores could reserve different addresses at the same time.

(I didn't write it though so that's just a guess.)

c_emulator/riscv_sail.h
Timmmm: So we do have a header now and don't have to copy/paste the types. So we could do this if you want.

I'm on the fence if it's worth the extra complexity though. Maybe we can discuss it in the meeting.

pmundkur:
> > why functions like *_reservation are C functions; would it be more consistent to define them as sail functions?
> 
> I presume it is because reservation is a memory thing and that is also handled outside Sail.
> 
> I think a practical reason is that if you are simulating a multicore chip then you would have a single memory in C, and you'd also need to handle reservations there. If you had reservation state in Sail then it would be duplicated for each core and that wouldn't work, e.g. multiple cores could reserve different addresses at the same time.
> 
> (I didn't write it though so that's just a guess.)

Yes, that was the motivation to keep it in C.  The more important change that needs to be made here to add the `mhartid` to the callback, so that the memory system can track reservations on a per-hart basis and cancel matching reservations on other harts.

PR #677, opened 2025-01-10
PTEs are invalid if reserved bits are non-zero

Timmmm:
This has only been the case since version 1.12 of the spec.

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -419,7 +419,7 @@ private function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     STCE = if currentlyEnabled(Ext_Sstc) then v[STCE] else 0b0,
     PMM = if currentlyEnabled(Ext_Smnpm) & is_supported_pmm(PM_SMNPM, pmm_mode(v[PMM])) then v[PMM] else o[PMM],
     ADUE = if currentlyEnabled(Ext_Svadu) then v[ADUE] else 0b0,
-    PBMTE = if hartSupports(Ext_Svpbmt) then v[PBMTE] else 0b0,
+    PBMTE = if currentlyEnabled(Ext_Svpbmt) then v[PBMTE] else 0b0,
     // Other extensions are not implemented yet so all other fields are read only zero.
   ]
 }
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -419,7 +419,7 @@ private function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     STCE = if currentlyEnabled(Ext_Sstc) then v[STCE] else 0b0,
     PMM = if currentlyEnabled(Ext_Smnpm) & is_supported_pmm(PM_SMNPM, pmm_mode(v[PMM])) then v[PMM] else o[PMM],
     ADUE = if currentlyEnabled(Ext_Svadu) then v[ADUE] else 0b0,
-    PBMTE = if currentlyEnabled(Ext_Svpbmt) then v[PBMTE] else 0b0,
+    PBMTE = if hartSupports(Ext_Svpbmt) then v[PBMTE] else 0b0,
     // Other extensions are not implemented yet so all other fields are read only zero.
   ]
 }
--- a/model/sys/vmem_pte.sail
+++ b/model/sys/vmem_pte.sail
@@ -74,7 +74,7 @@ private function pte_is_non_leaf(pte_flags : PTE_Flags) -> bool =
 // NAPOT Translation Contiguity
 function clause currentlyEnabled(Ext_Svnapot) = hartSupports(Ext_Svnapot) & currentlyEnabled(Ext_Sv39)
 // Page-Based Memory Types
-function clause currentlyEnabled(Ext_Svpbmt) = menvcfg[PBMTE] == 0b1
+function clause currentlyEnabled(Ext_Svpbmt) = hartSupports(Ext_Svpbmt) & currentlyEnabled(Ext_Sv39)
 // Svrsw60t59b: PTE Reserved for software bits 60-59
 function clause currentlyEnabled(Ext_Svrsw60t59b) = hartSupports(Ext_Svrsw60t59b) & currentlyEnabled(Ext_Sv39)
 // Obviating Memory-management Instructions after Marking PTEs valid
--- a/model/sys/vmem_pte.sail
+++ b/model/sys/vmem_pte.sail
@@ -74,7 +74,7 @@ private function pte_is_non_leaf(pte_flags : PTE_Flags) -> bool =
 // NAPOT Translation Contiguity
 function clause currentlyEnabled(Ext_Svnapot) = hartSupports(Ext_Svnapot) & currentlyEnabled(Ext_Sv39)
 // Page-Based Memory Types
-function clause currentlyEnabled(Ext_Svpbmt) =  hartSupports(Ext_Svpbmt) & currentlyEnabled(Ext_Sv39)
+function clause currentlyEnabled(Ext_Svpbmt) = menvcfg[PBMTE] == 0b1
 // Svrsw60t59b: PTE Reserved for software bits 60-59
 function clause currentlyEnabled(Ext_Svrsw60t59b) = hartSupports(Ext_Svrsw60t59b) & currentlyEnabled(Ext_Sv39)
 // Obviating Memory-management Instructions after Marking PTEs valid
@@ -93,14 +93,19 @@ private function pte_is_invalid(pte_flags : PTE_Flags, pte_ext : PTE_Ext) -> boo
   // TODO: handle V=1 and henvcfg.SSE=0 at VS/VU levels
   | (pte_flags[R] == 0b0 & pte_flags[W] == 0b1 & pte_flags[X] == 0b0 & menvcfg[SSE] == 0b0)
   | (pte_flags[R] == 0b0 & pte_flags[W] == 0b1 & pte_flags[X] == 0b1)
-  // Note, the following requirements depend on the privileged spec version.
-  // Early versions do not require this check. This will need to be behind
-  // a flag when the Sail model allows specifying the spec version.
-  | (pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros()))
-  | (pte_ext[N] != 0b0 & not(currentlyEnabled(Ext_Svnapot)))
-  | (pte_ext[PBMT] != zeros() & (menvcfg[PBMTE] == 0b0 | not(page_based_mem_type_forwards_matches(pte_ext[PBMT]))))
-  | (pte_ext[RSW_60t59b] != zeros() & not(currentlyEnabled(Ext_Svrsw60t59b)))
-  | pte_ext[reserved] != zeros()
+  // Reserved encodings mean the PTE is invalid.
+  | not(page_based_mem_type_forwards_matches(pte_ext[PBMT])) & currentlyEnabled(Ext_Svpbmt)
+  // Early versions of the spec did not require reserved bits to be zero.
+  | pte_reserved_bits_must_be_zero & (
+    // A/D/U and all ext bits are currently reserved for non-leaf PTEs.
+      pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros())
+    // N/PBMT/60t59b are reserved if not implemented.
+    | pte_ext[N]          != zeros() & not(currentlyEnabled(Ext_Svnapot))
+    | pte_ext[PBMT]       != zeros() & not(currentlyEnabled(Ext_Svpbmt))
+    | pte_ext[RSW_60t59b] != zeros() & not(currentlyEnabled(Ext_Svrsw60t59b))
+    // Reserved bits are reserved even for leaf PTEs.
+    | pte_ext[reserved] != zeros()
+  )
 
 // ----------------
 // Check access permissions in PTE
@@ -112,7 +117,7 @@ union pte_check_failure = {
 }
 
 // For (non-standard) extensions: this function gets the extension-available bits
-// of the PTE in extPte, and the accumulated information of the page-table-walk
+// of the PTE in PTE_Ext, and the accumulated information of the page-table-walk
 // in ext_ptw. It should return the updated ext_ptw in both success and failure cases.
 
 union PTE_Check = {
--- a/model/sys/vmem_pte.sail
+++ b/model/sys/vmem_pte.sail
@@ -101,7 +101,7 @@ private function pte_is_invalid(pte_flags : PTE_Flags, pte_ext : PTE_Ext) -> boo
       pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros())
     // N/PBMT/60t59b are reserved if not implemented.
     | pte_ext[N]          != zeros() & not(currentlyEnabled(Ext_Svnapot))
-    | pte_ext[PBMT]       != zeros() & not(currentlyEnabled(Ext_Svpbmt))
+    | pte_ext[PBMT]       != zeros() & (menvcfg[PBMTE] == 0b0 | not(page_based_mem_type_forwards_matches(pte_ext[PBMT])))
     | pte_ext[RSW_60t59b] != zeros() & not(currentlyEnabled(Ext_Svrsw60t59b))
     // Reserved bits are reserved even for leaf PTEs.
     | pte_ext[reserved] != zeros()
--- a/model/core/isa_version.sail
+++ b/model/core/isa_version.sail
@@ -53,7 +53,6 @@ let priv_isa_version : Privileged_ISA_Version = config base.privileged_isa_versi
 
 // In this version, a new requirement was added that reserved bits
 // in PTEs must be zero otherwise the PTE is invalid.
-// TODO: This is not used yet; this is an example that will be implemented in a future commit.
 let pte_reserved_bits_must_be_zero = priv_isa_version >= Privileged_ISA_1_12
 
 
--- a/model/sys/vmem.sail
+++ b/model/sys/vmem.sail
@@ -180,7 +180,8 @@ function pt_walk(
             PTE_Check_Success(ext_ptw) => {
               let ppn = if level > 0 then {
                 // NAPOT PTEs are currently reserved for level > 0.
-                if pte_ext[N] == 0b1 then return Err(PTW_Invalid_PTE(), ext_ptw);
+                if pte_ext[N] == 0b1 & pte_reserved_bits_must_be_zero
+                then return Err(PTW_Invalid_PTE(), ext_ptw);
 
                 // Compose final PA in superpage:
                 // Superpage PPN @ lower VPNs @ page-offset


model/riscv_types.sail
```diff
@@ -16,6 +16,16 @@
...
 
+// Priv/unpriv spec version.
+// TODO: Replace the `true`s with configuration flags.
+
+// Machine and Supervisor Architectures v1.11
+function clause extensionEnabled(Ext_Sx1p11) = true
+// Machine and Supervisor Architectures v1.12
+function clause extensionEnabled(Ext_Sx1p12) = extensionEnabled(Ext_Sx1p11) | true
+// Machine and Supervisor Architectures v1.13
+function clause extensionEnabled(Ext_Sx1p13) = extensionEnabled(Ext_Sx1p12) | true
```
arichardson, line 27: Should this be an & instead of an | ? I.e. 1.13 requires 1.12?

Not that it really matters since it's all hardcoded to 1.

model/riscv_extensions.sail
```diff
@@ -11,6 +11,13 @@ scattered enum extension
 // Note, these are sorted according to the canonical ordering vaguely described
 // in the `Subset Naming Convention` section of the unprivileged spec.
 
+// Machine and Supervisor Architectures v1.11
+enum clause extension = Ext_Sx1p11
```
jrtc27, line 15: There is no such thing as Sx1p11 etc. Ss1p11 and Sm1p11 are the extension names. We shouldn't be using non-existent extension names in the official model.

model/riscv_types.sail
```diff
@@ -16,6 +16,16 @@
 val extensionEnabled : extension -> bool
 scattered function extensionEnabled
 
+// Priv/unpriv spec version.
+// TODO: Replace the `true`s with configuration flags.
+
+// Machine and Supervisor Architectures v1.11
+function clause extensionEnabled(Ext_Sx1p11) = true
+// Machine and Supervisor Architectures v1.12
+function clause extensionEnabled(Ext_Sx1p12) = extensionEnabled(Ext_Sx1p11) | true
```
jrtc27, line 25: This doesn't make sense to me. You can't have multiple versions of an extension implemented. Ss1p12 is implemented if and only if priv 1.12 is the version your implementation chose. You have the Ss and Sm extensions and they have a version corresponding to whatever's implemented.

What you're after is some `extensionVersion` function that, given Ss (NB: no suffix on the extension name in the enum), tells you what version is implemented, so you can do `>= 1.12` where needed.

model/riscv_extensions.sail
Timmmm: Yeah true, I just thought it was probably overkill to have both (I'm not even entirely sure what it would mean to support Ss1p11 and Sm1p12 for example) so this was an attempt to combine them. I agree it's not very obvious though and unfortunate that it's not a real extension name.

Do you think we should have both separately, or use a different name to combine them?

jrtc27: They should be separate, because from an ISA model reading perspective it's important to know whether something's attributed to Ss or Sm. But I think it's perfectly fine to have an invariant that the versions match between them (and expect that's an implicit assumption made by the privileged spec; it hurts my head to think about what mixing would even look like...).

model/riscv_types.sail
Timmmm: Hmm yeah that is definitely the more correct way to do it, but it seemed a lot simpler to pretend they are separate extensions and then ensure that version N implies versions <N.

Open to other options though.

model/riscv_types.sail
Timmmm: Yeah I think you're right. I'll change it.

model/riscv_extensions.sail
Timmmm: Ok that makes sense, I'll change it.

model/riscv_types.sail
jrtc27: Simpler, sure, but definitely wrong, and for the official model that's a real problem.

Timmmm: Is it actually *wrong*? If a chip supports Ss1p12 and some instruction requires Ss1p11 then isn't it fair to say the chip supports Ss1p12 and Ss1p11... Unless the versions aren't backwards compatible I guess. Which tbf is the sometimes the case.

Timmmm: Ah ok I understand. I'll try and come up with something better.

model/riscv_types.sail
jordancarlin: If we are going to go with this approach and not a proper extension version function, then I think the following would make more sense. The current use of | is definitely not right (enabling 1.11 would result in 1.13 being enabled), but I think using & is also cumbersome because enabling 1.13 would require also manually enabling 1.11 and 1.12.

```
// Machine and Supervisor Architectures v1.11
function clause extensionEnabled(Ext_Sx1p11) = extensionEnabled(Ext_Sx1p12) | true
// Machine and Supervisor Architectures v1.12
function clause extensionEnabled(Ext_Sx1p12) = extensionEnabled(Ext_Sx1p13) | true
// Machine and Supervisor Architectures v1.13
function clause extensionEnabled(Ext_Sx1p13) = true
```

That way the user just needs to enable the latest version that they support and other versions would implicitly be enabled as well. This probably covers most cases (since there are very few backwards incompatible changes) and the few backwards incompatible scenarios can use something like `if extensionEnabled(Ext_Sx1p12) & not(extensionEnabled(Ext_Sx1p13))`

Timmmm: Yeah I think Jessica is right - I'm going to rework this to use a proper versioning function.

jordancarlin: Sounds good. That is definitely the more robust approach.

jrtc27: Great. Happy to review that when you have it.

jordancarlin:
For Priv there are defined versions. 1.13, 1.12, and 1.11 seem sufficient for Sail, though technically 1.10 and 1.9 also exist. I'll see if I can find the documentation where I found this.

Timmmm:
I think they switched from 1.XX to date based versions though. Except... not completely, so now there are both, e.g. 1.12 is 20211203. It seems to be a mess. Maybe we should ask Andrew Waterman.

jordancarlin:
My understanding is there there are date based releases of the manuals, but 1.XX releases of versions of the spec. So a particular date based release will have priv 1.12 and a different date based release will have priv 1.13. The intermittent date based releases might add other extensions to the priv manual.

I think we should base it off of extension versions, so 1.12, 1.13, etc. instead of manual releases since those seem somewhat arbitrary.

Timmmm:
Hmm maybe. I'm going to ask Mr Waterman. This seems like something that could be a lot clearer!

Timmmm:
Well the discussion on `riscv-isa-manual` wasn't very illuminating and after discussing it in the meeting today we came to the conclusion that just using the date was the best option. I thought it would also allow us to easily then support non-released versions (since there have apparently only been 2 releases ever), but because Git commit/author dates are a little woolly it's not quite as nice as I'd hoped. I've pushed a version that does that anyway.

The full giving-up-on-their-version-numbers option would be to use `riscv-isa-manual` commit hashes. I'm actually thinking that might make sense. It has the added benefit that people can easily find where the behaviour changed in the spec (since it doesn't have the greatest changelog). We don't need to add *every* commit; just released ones and ones where something changed. Something like the previous version:

```
enum IsaVersion = {
    Version_31bc179b214fc0374075ea9284aa939b8341b843,
    ... more versions *in order* ...
}

// TODO: Replace with 'config ...priv_version' or similar when the config system is available.
function isaVersion() -> IsaVersion = ...

...
overload operator < = {isa_version_lt}
...
```

I think this is the most robust option. The downside is obviously that nobody really likes reading commit hashes. But I think given that there isn't an "official" versioning scheme that really makes any sense maybe this is the best option?

allenjbaum:
We can't use hashes if we ever need to compare them, if version>x.

Alasdair:
Do we need to support non-released versions? Could we just have some sort of `Version_next` which is greater than all released versions to capture ongoing changes?

I didn't bring it up in the meeting, but the ISA string text in the manuals does specify a versioning scheme as something like `ExtXpY` for version `X.Y` of an extension - is that related or is priv+unpriv just versioned differently?

Timmmm:
> We can't use hashes if we ever need to compare them, if version>x.

This is solved (awkwardly) by listing the ones we use in order in the enum.

> Do we need to support non-released versions?

I think we do, given that there have apparently only been two released versions of the ISA manual. Did people really wait for 4 years after `20191213` without implementing any of the new specs? I'd be really surprised if there aren't chips out there targeting versions between 20191213 and 20240411. The chip I'm working on targets 2023-04-27.

> but the ISA string text in the manuals does specify a versioning scheme as something like ExtXpY for version X.Y of an extension - is that related or is priv+unpriv just versioned differently?

The RISC-V profiles mention:

> Ss1p12 Privileged Architecture version 1.12.

But then it goes on to contradict itself:

> Sm1p11, Machine Architecture v1.11
> Sm1p12, Machine Architecture v1.12
> Ss1p11, Supervisor Architecture v1.11
> Ss1p12, Supervisor Architecture v1.12

Also the actual ISA manual doesn't mention the `Sm` or `Ss` extensions anywhere I can see (difficult to search for though so I may have missed it).

arichardson:
> If we can come up with exact commits for v1.11, 1.12, etc. then we could maybe use those, but that seems to have been lost to history.

1.12 seems to be 20211203: https://github.com/riscv/riscv-isa-manual/releases/tag/Priv-v1.12

arichardson:
For 1.10 I can find https://github.com/riscv/riscv-isa-manual/releases/tag/riscv-priv-1.10 which suggests 20170509

Timmmm:
Ah good detective work! What about unpriv? If we can find sensible commits for all of them then I would say we can use `Priv_1_12` etc and just add a comment saying exactly which commit we mean. I guess the Sail model can become the authoritative source on that :-D

jordancarlin:
At this point, it sort of seems like we would be better off basing things off of extension versions instead of trying to capture a manual release version. Everything(?) in the spec is contained within an extension, and there are not technically any rules preventing someone from implementing versions of extensions from different manual releases, so this seems the most consistent approach.

For most unprivileged extensions there is just the one version (1.0), so almost nothing needs to change. Some of the base extensions are on version 2.0/2.1/2.2, but I doubt we need to worry about older versions of those. The only significant one we might want to support is the older version of `I` that included Zicsr and Zifencei.

For privileged extensions, most have their own version numbers (all 1.0 from a quick glance) except machine mode and supervisor mode themselves, which we'd want to support 1.10, 1.11, 1.12, and 1.13 for.

Seems like for an initial attempt at this, we can just support various versions of Machine and Supervisor modes and address older versions of other extensions as they come up if users need them.

Timmmm:
> For most unprivileged extensions there is just the one version (1.0), so almost nothing needs to change.

Yeah the problem is that RVI hasn't been very disciplined about versioning, so it's unclear exactly which document version N of spec Y is, and there have been occasional "clarifications" without changing version numbers. That's the big advantage of using Git commits - you can actually know with certainty what specification you are talking about.

I think it's fine if we use version numbers and then also say in a comment which commit hash we mean though:

```
enum PrivVersion = {
  // 98964261c931d51884f733664b22efe43de93033
  Version_1_12,
  etc.
}
```

model/riscv_extensions.sail
```diff
@@ -6,6 +6,17 @@
...
 
+// Privileged and unprivileged spec versions. The versioning system for specifications
+// is unclear so we just use the commit date. This also allows versioning for specs changes
+// that are between releases. If there was an official release for this day (for example
+// 20240411) then this is the commit of that release. Otherwise it is for the *last*
+// commit on the given day.
+type SpecVersion = range(20000000, 21000000)
+
+// TODO: Replace with 'config ...priv_version' or similar when the config system is available.
+let configPrivVersion : SpecVersion = 20240411
```
arichardson, line 17: Maybe the easiest approach here to allow using relational operators would be to add variables such as `Priv_1_12 = 20211203` and use those in the comparisons?

arichardson: And a comment above with a commit hash and a link to the release on github?

Timmmm:
Ok new attempt. Unfortunately it doesn't compile yet due to https://github.com/rems-project/sail/issues/1353.

model/sys/vmem_pte.sail
```diff
@@ -77,18 +77,22 @@ function clause currentlyEnabled(Ext_Svpbmt) = false
...
     pte_flags[V] == 0b0
   | (pte_flags[W] == 0b1 & pte_flags[R] == 0b0)
-  // Note, the following requirements depend on the privileged spec version.
-  // Early versions do not require this check. This will need to be behind
-  // a flag when the Sail model allows specifying the spec version.
-  | (pte_ext[N] != 0b0 & not(currentlyEnabled(Ext_Svnapot)))
-  | (pte_ext[PBMT] != zeros() & not(currentlyEnabled(Ext_Svpbmt)))
-  | pte_ext[reserved] != zeros()
+  // Early versions of the spec did not require reserved bits to be zero.
+  | pte_reserved_bits_must_be_zero & (
```
radimkrcmar, line 99: Does this refer to the following?
```
* Clarify that if the address-translation algorithm non-speculatively reaches a PTE in which a bit
reserved for future standard use is set, a page-fault exception must be raised.
```
The manual was "clarified", so it seems as if an implementation that doesn't behave like this is not, and never was, an implementation of the RISC-V ISA.

I don't mind treating it as a backward-incompatible change instead.

model/core/spec_version.sail
```diff
@@ -0,0 +1,69 @@
...
+  Hash_98964261c931d51884f733664b22efe43de93033 <-> "20211203",
+  Hash_6ee57d8997021aaa8025e208164763fe12e03c19 <-> "20240411",
+  Hash_Latest <-> "latest",
+}
+
+let priv_spec_version : IsaHash = priv_spec_name(config base.priv_spec_version)
+
+// Released versions of the unpriv spec.
+let UnprivSpec_20191213 = Hash_66bed2470f68520fadc4bf86c6c000d817fe73f2
+let UnprivSpec_20240411 = Hash_6ee57d8997021aaa8025e208164763fe12e03c19
```
radimkrcmar: This could be `unpriv_spec_name("20240411")`.
Do we even need the extra defines when we have the mapping?

radimkrcmar (changes requested):

model/sys/vmem_pte.sail
Timmmm: Yes exactly, but despite them calling it a "clarification" it's pretty clearly a spec change. I think we should treat such changes as actual changes even if they wish they had always been the case.

model/core/spec_version.sail
Timmmm: Maybe not but I think it's maybe best to avoid strings in code that formal back ends will want to deal with. I dunno if Sail does constant propagation on this...

model/sys/vmem_pte.sail
radimkrcmar: When configuring an incompatible version of the manual, shouldn't we issue a warning that the modeled behavior is not RISC-V?

I expect that architectural tests will catch this, and then the warning would serve as an explanation why Sail is failing the tests.

Timmmm: It is RISC-V though - just an older version. Maybe I don't understand what you mean...

radimkrcmar: Yeah, I don't really understand what RISC-V is doing with the versions...
I interpreted the "clarification" as "it's your problem if you didn't understand what we meant", and assumed that RISC-V introduced short term issues to reduce the long term overhead -- the current manual doesn't give an option to behave like it did before, and RISC-V could have introduced a new extension if both behaviors were ever supposed to be valid.

model/sys/vmem_pte.sail
```diff
@@ -77,18 +77,22 @@ function clause currentlyEnabled(Ext_Svpbmt) = false
...
+    // All bits are currently reserved for non-leaf PTEs.
+      pte_is_non_leaf(pte_flags) & pte_ext.bits != zeros()
```
radimkrcmar: To get the review back into more relevant topics: The `pte_flags` bits A, D, and U are also reserved for non-leaf entries, so I think we might just as well do it here too.

model/core/spec_version.sail
```diff
@@ -0,0 +1,69 @@
...
+
+  // Tag: Priv-v1.12 (AKA 20211203)
+  Hash_98964261c931d51884f733664b22efe43de93033,
+
+  // Tag: 20240411
+  Hash_6ee57d8997021aaa8025e208164763fe12e03c19,
+
+  // Abitrary hash that is later than all real hashes.
+  Hash_Latest,
+}
```
arichardson: Since we only care about the enum order, maybe we could use descriptive names by default and the hash for behaviour changes that happened inbetween releases?


```suggestion
enum IsaSpec = {
  // Tag: Ratified-IMAFDQC (AKA 20191213), git hash 66bed2470f68520fadc4bf86c6c000d817fe73f2
  IsaSpec_20191213,

  // Requires PTE reserved bits to be zero.
  IsaSpec_Hash_a432a8997892d0a103798c2e436d36ae66da7d34,

  // Tag: Priv-v1.12 (AKA 20211203), git hash 98964261c931d51884f733664b22efe43de93033
  IsaSpec_20211203, 

  // Tag: 20240411, git hash 6ee57d8997021aaa8025e208164763fe12e03c19
  IsaSpec_20240411,

  // Abitrary version that is later than all real ISA spec versions.
  IsaSpec_Latest,
}
```

arichardson: But also fine with the current format that uses the let aliases :)

pmundkur (changes requested):
The general approach looks good to me.  I'd add checks to the config validation function to ensure that a recognized string is specified for the spec versions to give a better error message.

model/core/spec_version.sail
Timmmm: Your suggestion worked better - I did it in #1756

Timmmm:
I broke the version changes out into #1756.

model/sys/vmem_pte.sail
```diff
@@ -93,14 +93,18 @@ private function pte_is_invalid(pte_flags : PTE_Flags, pte_ext : PTE_Ext) -> boo
...
-  | (pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros()))
-  | (pte_ext[N] != 0b0 & not(currentlyEnabled(Ext_Svnapot)))
-  | (pte_ext[PBMT] != zeros() & (menvcfg[PBMTE] == 0b0 | not(page_based_mem_type_forwards_matches(pte_ext[PBMT]))))
-  | (pte_ext[RSW_60t59b] != zeros() & not(currentlyEnabled(Ext_Svrsw60t59b)))
-  | pte_ext[reserved] != zeros()
+  | (pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1))
+  // Early versions of the spec did not require reserved bits to be zero.
+  | pte_reserved_bits_must_be_zero & (
+    // All ext bits are currently reserved for non-leaf PTEs.
+      pte_is_non_leaf(pte_flags) & pte_ext.bits != zeros()
```
radimkrcmar: The manual always considered the non-leaf A/D/U bits as reserved bits, so I think we should treat them the same way too:

```suggestion
  // Early versions of the spec did not require reserved bits to be zero.
  | pte_reserved_bits_must_be_zero & (
    // All ext bits are currently reserved for non-leaf PTEs.
      pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros()))
```

Timmmm: Ah yeah I accidentally put that outside the `pte_reserved_bits_must_be_zero` check for some reason. I'll move it.

Timmmm: Fixed.

model/core/sys_regs.sail
```diff
@@ -419,7 +419,7 @@ private function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     STCE = if currentlyEnabled(Ext_Sstc) then v[STCE] else 0b0,
     PMM = if currentlyEnabled(Ext_Smnpm) & is_supported_pmm(PM_SMNPM, pmm_mode(v[PMM])) then v[PMM] else o[PMM],
     ADUE = if currentlyEnabled(Ext_Svadu) then v[ADUE] else 0b0,
-    PBMTE = if currentlyEnabled(Ext_Svpbmt) then v[PBMTE] else 0b0,
+    PBMTE = if hartSupports(Ext_Svpbmt) then v[PBMTE] else 0b0,
```
pmundkur: Might want to also mention Svpbmt changes to the PR message.

pmundkur (commented):
LGTM.

Could you also add the changelog entry for the config file change that got missed in #1756?

model/core/sys_regs.sail
radimkrcmar: Ah, and looking at the changes, they seem to introduce a minor bug:
If `misa.S=0`, then PBMTE should be 0, since Svpbmt depends on Sv39, which depends on S, but hartSupports(Ext_Svpbmt) doesn't care about any of that.

model/sys/vmem_pte.sail
```diff
@@ -93,14 +93,19 @@ private function pte_is_invalid(pte_flags : PTE_Flags, pte_ext : PTE_Ext) -> boo
...
-  | pte_ext[reserved] != zeros()
+  // Reserved encodings mean the PTE is invalid.
+  | not(page_based_mem_type_forwards_matches(pte_ext[PBMT])) & currentlyEnabled(Ext_Svpbmt)
+  // Early versions of the spec did not require reserved bits to be zero.
+  | pte_reserved_bits_must_be_zero & (
+    // A/D/U and all ext bits are currently reserved for non-leaf PTEs.
+      pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros())
+    // N/PBMT/60t59b are reserved if not implemented.
+    | pte_ext[N]          != zeros() & not(currentlyEnabled(Ext_Svnapot))
+    | pte_ext[PBMT]       != zeros() & not(currentlyEnabled(Ext_Svpbmt))
```
radimkrcmar: And since I've already started nagging around Svpbmt... Can Sail configure ancient base RISC-V with Svpbmt?
If so !pte_reserved_bits_must_be_zero will still fail on non-zero reserved bits. :)

Another minor issue is that PBMT is only applicable to leaf PTEs -- non-leaves have bits 62-61 reserved, so we shouldn't accept non-zero fields there even if PBMT is enabled.

I think the correct code would be (edited to fix a bug in it):

```suggestion
  | not(page_based_mem_type_forwards_matches(pte_ext[PBMT])) & currentlyEnabled(Ext_Svpbmt) & not(pte_is_non_leaf(pte_flags))
  // Early versions of the spec did not require reserved bits to be zero.
  | pte_reserved_bits_must_be_zero & (
    // A/D/U and all ext bits are currently reserved for non-leaf PTEs.
      pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros())
    // N/PBMT/60t59b are reserved if not implemented.
    | pte_ext[N]          != zeros() & not(currentlyEnabled(Ext_Svnapot))
    | pte_ext[PBMT]       != zeros() & (not(currentlyEnabled(Ext_Svpbmt)) | pte_is_non_leaf(pte_flags))
```

model/core/sys_regs.sail
pmundkur: Dealing with cleared `misa` bits is a thorny issue which we haven't resolved yet: #1313

model/sys/vmem_pte.sail
pmundkur: > Another minor issue is that PBMT is only applicable to leaf PTEs -- non-leaves have bits 62-61 reserved, so we shouldn't accept non-zero fields there even if PBMT is enabled.

Shouldn't that be handled already by:
```
| pte_reserved_bits_must_be_zero & ( 
      pte_is_non_leaf(pte_flags) & ( ... | pte_ext.bits != zeros())
...
```

model/sys/vmem_pte.sail
```diff
@@ -74,7 +74,7 @@ private function pte_is_non_leaf(pte_flags : PTE_Flags) -> bool =
 // NAPOT Translation Contiguity
 function clause currentlyEnabled(Ext_Svnapot) = hartSupports(Ext_Svnapot) & currentlyEnabled(Ext_Sv39)
 // Page-Based Memory Types
-function clause currentlyEnabled(Ext_Svpbmt) =  hartSupports(Ext_Svpbmt) & currentlyEnabled(Ext_Sv39)
+function clause currentlyEnabled(Ext_Svpbmt) = menvcfg[PBMTE] == 0b1
```
arichardson: Shouldn't this change be a separate PR rather than folding it in here?

model/core/sys_regs.sail
Timmmm: Yeah what should actually happen is when `misa.S=0` then `menvcfg[PBMTE]` should be read-only zero. But we haven't properly implemented that (and the model doesn't support dynamic `misa.S` yet).

Though I think you're right that I have introduced a bug where you can set `menvcfg[PBMTE]=1` even if `misa.S=0`. I'll fix that.

model/sys/vmem_pte.sail
Timmmm: > Can Sail configure ancient base RISC-V with Svpbmt?

Yeah... that did occur to me. I have no idea what the spec has to say about that. I assume you're not supposed to do that, but I think explicitly checking for that is probably overkill anyway.

model/sys/vmem_pte.sail
Timmmm: Yeah I can split it out.

model/sys/vmem_pte.sail
radimkrcmar: > > Another minor issue is that PBMT is only applicable to leaf PTEs -- non-leaves have bits 62-61 reserved, so we shouldn't accept non-zero fields there even if PBMT is enabled.
> 
> Shouldn't that be handled already by:
> 
> ```
> | pte_reserved_bits_must_be_zero & ( 
>       pte_is_non_leaf(pte_flags) & ( ... | pte_ext.bits != zeros())
> ...
> ```

Ah, right, thanks.  The second change is useless and the first change only fixes the situation of Svpbmt + old RISC-V.

> > Can Sail configure ancient base RISC-V with Svpbmt?
> 
> Yeah... that did occur to me. I have no idea what the spec has to say about that. I assume you're not supposed to do that, but I think explicitly checking for that is probably overkill anyway.

I think the extensions are supposed to be independent of the base ISA.  I agree that we don't want anyone doing that, but if we know that there is an incompatibility, I'd rather fix it or forbid Svpbmt with old RISC-V.  (Both are one-line changes.)

pmundkur:
@Timmmm if you could rebase this, we can pull this into the next release.

pmundkur:
@Timmmm will you have time to rebase this, or should I do it?

Timmmm:
Yeah might be best if you do it - I'm going to be pretty busy for the next 6 weeks at least.

pmundkur:
Rebased,  I also removed the Svpbmt-related changes, and will soon submit a separate PR for those.

pmundkur (approved):

jordancarlin (commented):
I think the following check in `vmem.sail` on lines 181-183 needs to be updated as well. Right now it would still trap, which I don't think would be required for priv 1.11.

```sail
let ppn = if level > 0 then {
  // NAPOT PTEs are currently reserved for level > 0.
  if pte_ext[N] == 0b1 then return Err(PTW_Invalid_PTE(), ext_ptw);
```

pmundkur:
Good catch!

jordancarlin:
I can't take credit for it. I had AI do a review and it found that. I was impressed because I hadn't even looked at that file. Anyway, glad it is fixed.

jordancarlin (approved):

nadime15 (approved):

PR #680, opened 2025-01-10
Simplify float load/store

Timmmm:
Use Sail's fancy dependent types instead of separate functions for each float load/store size. Functionality is unchanged.


arichardson (approved):
nice simplification

jordancarlin (approved):
This is great!

PR #681, opened 2025-01-10
Improve store-conditional speculative failures and exception priority

Timmmm:
1. Move `speculate_conditional()` to the same place as `match_reservation()`. I am not 100% sure about this change because I'm not sure what rmem is (and it has a comment about allowing it to fail very early), but it doesn't make sense to me that SC could spuriously fail in a place where it shouldn't be allowed in the sequential model.

Codasip uses `speculate_conditional()` to force spurious SC failures into the model (so it matches the DUT), and this is the right place for that use.

2. Do an explicit access check before checking the reservation. Without this you never get an access fault for accesses to memory that doesn't exist or doesn't support LR/SC.

3. Move `cancel_reservation()` to the end of the function. This doesn't change any functionality, I think it's just clearer and makes it obvious that the omission of `cancel_reservation()` in the `MemException(e)` branch was not a mistake.


Timmmm:
@Alasdair if you could shed any light on what rmem is expecting here that would be great!

model/riscv_insts_aext.sail
```diff
@@ -115,52 +115,58 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
...
-                match eares {
-                  MemException(e) => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
-                  MemValue(_) => {
-                    let rs2_val = X(rs2);
-                    match mem_write_value(addr, width_bytes, rs2_val[width_bytes * 8 - 1 .. 0], aq & rl, rl, true) {
-                      MemValue(true)  => { X(rd) = zero_extend(0b0); cancel_reservation(); RETIRE_SUCCESS },
-                      MemValue(false) => { X(rd) = zero_extend(0b1); cancel_reservation(); RETIRE_SUCCESS },
-                      MemException(e) => { handle_mem_exception(vaddr, e); RETIRE_FAIL }
+  /* Get the address, X(rs1) (no offset).
+    * Extensions might perform additional checks on address validity.
```
arichardson, line 119: I know this was only moved around, but the indentation looks odd.

Timmmm: Ah ha that is due to a slight bug in the VSCode "sticky tabs" option I use. It more or less makes stupid space-based indentation act as if glorious tabs had been used instead, but it seems like increasing/decreasing indentation on lines with incomplete "tabs" doesn't quite work correctly (technically my fault because I implemented that feature).

I changed it to line comments which I think are just better anyway.

Timmmm:
The code has changed a lot since I wrote this, and I think all we need to do really is remove `speculate_conditional` and allow `match_reservation` to spuriously return false - implemented in #1060.

PR #683, opened 2025-01-11
mstatus cleanup

radimkrcmar:
This series started by noticing that we use SXL in RV32 mode, which results in very ugly accessors for mstatus.  The code around mstatus can be improved by changing the architecture  (XLEN) detection.

There is (should be) no functional change.


arichardson:
Thanks for this change, I haven't had a detailed look at it yet but it seems some of the changes conflict with https://github.com/riscv/sail-riscv/pull/652/files which will most likely be merged soon. 

@Timmmm could you take a look and see which one is better or if we want both changes?

radimkrcmar:
The pull request also expands the status to bits(64) and explicitly copies the fields, so I can drop 4 patches.  I will rebase when it lands and add 2 patches to clean up its handling of SXL/UXL and SD.

radimkrcmar:
I rebased on top of https://github.com/riscv/sail-riscv/pull/652/ and noticed a bug in it, so this isn't just a cleanup anymore, but introduces a software observable change.

Timmmm:
Ah what's the bug? Can we fix that quickly in another PR?

radimkrcmar:
The patch fixing it is https://github.com/riscv/sail-riscv/pull/683/commits/dc3e3c0852f0945686e29e889dafb6b01fc295e5. In short, mstatush contains 0b01 in SXL and UXL fields instead of required 0b00.

To fix it quickly, we could replace the SXL/UXL bits with zeros in mstatush accessors, but I don't think we have to rush, because software should ignore those WPRI bits.

jrtc27:
They're also supposed to be 0 if the corresponding privilege mode isn't implemented

radimkrcmar:
> They're also supposed to be 0 if the corresponding privilege mode isn't implemented

Good point.  I didn't really want to change any software-visible behavior in this PR, so I'll add the extensionEnabled in another PR.

radimkrcmar:
I noticed that I didn't commit my last change, so the reset patch was only correct if mstatus was pre-initialized with zeros... It makes no difference, because we have to zero WPRI, but it's still not a good practice.
The force push just removes the last patch, because it would conflict with the reset PR (https://github.com/riscv/sail-riscv/pull/597) anyway.

Timmmm:
LGTM but I think this could be split into several basically independent PRs which would make it a lot easier to review (and give us nicer history). E.g.

```
PR 1:
sys_regs: rename cur_Architecture to cur_architecture
types: simplify architecture return type
types: turn architecture into a mapping

PR 2:
insts_base: ignore architecture in SFENCE_VMA

PR 3:
sys_regs: fix RV32 detection in cur_architecture
sys_regs: remove SXL and UXL accessors

PR 4:
sys_regs: compute mstatus and sstatus SD bit on CSR read

PR 5 (these could also be separate PRs but that's probably overkill):
sys_regs: fix SXL and UXL fields in RV32 mstatush
sys_regs: legalize UXL and SXL mstatus fields
```

Sorry I know it's a bit of a pain! Especially because Github doesn't really support the stacked patch development workflow...

I'll hold off on #597 until all these are merged.

radimkrcmar:
> LGTM but I think this could be split into several basically independent PRs which would make it a lot easier to review (and give us nicer history). E.g.

Will do, although history will be the same.  Please don't tell me you want to squash all the patches. :)

> Sorry I know it's a bit of a pain! Especially because Github doesn't really support the stacked patch development workflow...

No worries, this will be quick and easy compared to all the reshuffling I already did.

> Especially because Github doesn't really support the stacked patch development workflow...

I think Github supports it (Git alone does), both rebase and merge-commit PR resolution should get you the desired result if you merge the PRs in order.

> I'll hold off on #597 until all these are merged.

I would actually prefer to have that one merged first.
It will allow me to end up with cleaner code and more other PRs can depend on it.

arichardson:
> > LGTM but I think this could be split into several basically independent PRs which would make it a lot easier to review (and give us nicer history). E.g.
> 
> Will do, although history will be the same. Please don't tell me you want to squash all the patches. :)
> 
I agree that in cases where the history is nice and clean we should use rebase or merge instead of squash

Timmmm:
> Please don't tell me you want to squash all the patches. :)

Nope!

> I think Github supports it (Git alone does), both rebase and merge-commit PR resolution should get you the desired result if you merge the PRs in order.

Yeah I mean it doesn't really have a concept of saying "this PR depends on this other PR, so don't show that one". Apparently some source hosting systems do support that (Gerrit maybe?)

> I would actually prefer to have that one merged first.

Ok! I'll fix the initial SXL/UXL for RV32 in that one at least.

radimkrcmar:
> Yeah I mean it doesn't really have a concept of saying "this PR depends on this other PR, so don't show that one". Apparently some source hosting systems do support that (Gerrit maybe?)

Right, Github just compares the two branches.  We would have to work around that by using a range on commits, e.g. https://github.com/riscv/sail-riscv/pull/683/files/cf121a0..ddd0c7f

model/riscv_vmem.sail
```diff
@@ -179,7 +179,7 @@ function satp_to_ppn(satp_val) =
 function translationMode(priv : Privilege) -> SATPMode = {
   if priv == Machine then Bare
   else {
-    let arch = architecture(get_mstatus_SXL(mstatus));
+    let arch = if xlen == 32 then RV32 else architecture(mstatus[SXL]);
     let mbits : satp_mode = match arch {
       RV64 => {
         // Can't have an effective architecture of RV64 on RV32.
```
radimkrcmar, line 185: I tried to remove the `assert(xlen >= 64)` in the RV64 match, but Sail didn't compile with this hunk, nor with
```
let arch = if xlen >= 64 then architecture(mstatus[SXL]) else RV32;
```
```
Type error:
riscv_vmem.sail:187.24-28:
187 |        Mk_Satp64(satp)[Mode]
    |                        ^--^
    | Identifier Mode is unbound
    | 
    | Caused by riscv_vmem.sail:187.18-22:
    | 187 |        Mk_Satp64(satp)[Mode]
    |     |                  ^--^ checking function argument has type bitvector(64)
    |     | Failed to prove constraint: 32 == 64
```
Am I missing some better way to provide this constraint?

Thanks.

Timmmm: Does it work if you change it to `if xlen < 64 then RV32 else ...`?

Timmmm: Oh wait I can't read... Never mind.

Alasdair: The constraints mostly flow lexically, i.e. from the condition in the `if` down into the branches. The type-checker isn't smart enough to understand that `arch == RV64` implies that branch in a previous `if` was taken, it's a bit too indirect.

You can use an assert that's (clearly) always true at runtime just to assert a fact to the type system. Sometimes it's better to do that than to try to re-write the logic.

Timmmm:
Where did we get to on this? It looks like the remaining changes in this are fairly small and nice so maybe we can rebase it and get it reviewed/merged?

radimkrcmar:
Most of the patches got merged through other pull requests, and the remaining two weren't that important, so I just let it rot... I did a rebase, and dropped the second patch in the process, because we started putting mstatus into a lot of callbacks, which made dynamic computation of the dirty bit awkward.

model/riscv_vmem.sail
```diff
@@ -185,7 +185,7 @@ function satp_to_ppn(satp_val) =
 function translationMode(priv : Privilege) -> SATPMode = {
   if priv == Machine then Bare
   else {
-    let arch = architecture(get_mstatus_SXL(mstatus));
+    let arch = if xlen == 32 then RV32 else architecture(mstatus[SXL]);
```
radimkrcmar: This line will need changes when the hypervisor extension lands, so I deferred any attempts at making it pretty for later. :)

radimkrcmar:
Forgot to update before rebasing. Rebased again.

model/core/sys_regs.sail
```diff
@@ -288,15 +276,19 @@ function clause write_CSR((0x310, value) if xlen == 32) = { mstatus = legalize_m
...
+  else {
+    let a : arch_xlen =
+      match cur_privilege {
+        Machine    => misa[MXL],
+        Supervisor => mstatus[SXL],
+        User       => mstatus[UXL],
+        VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+        VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+      };
+    architecture(a)
```
arichardson: We could inline this call into each of the match clauses to reduce indentation by one level, but I don't mind either way.

arichardson (approved):

jordancarlin (approved):
LGTM

model/core/sys_regs.sail
radimkrcmar: I would prefer to lower the indentation level by one just by placing match on the let line, rather than repeating architecture.
Another possible way to lower indentation is an imperative return rather than else.

I pushed all of that now, and also got rid of the variable, is that more acceptable?

pmundkur (approved):

Timmmm:
> because we started putting mstatus into a lot of callbacks, which made dynamic computation of the dirty bit awkward.

Aw I thought that was a good idea. Can't you just add a `get_mstatus()` function and you always access it through that. There's actually even a `Private` keyword we can use to try to prevent raw access to the underlying register.

Anyway the remaining commit seems like a slight improvement to me so let's merge that, and then feel free to open another PR for the dirty bit if you want (I wouldn't say it's high priority though).

radimkrcmar:
> > because we started putting mstatus into a lot of callbacks, which made dynamic computation of the dirty bit awkward.
> 
> Aw I thought that was a good idea. Can't you just add a `get_mstatus()` function and you always access it through that. 

Yeah, I mostly dropped it to avoid the runtime overhead of the computation.  The dirty bit is currently memoized, and the state is conveniently placed where the CSR reads want to have it.

> There's actually even a `Private` keyword we can use to try to prevent raw access to the underlying register.

I'd have to read the Sail code to understand what it means by `Private`.  If it's something like the common use of private, I was thinking we might do legalization on internal CSR writes as well.

Neither have I explored how Sail interacts with the verification tools, so if we already have a constraint for `mstatus.SD <-> mstatus.VS == Dirty | mstatus.FS == Dirty | mstatus.XS == Dirty`, then we can keep the current performant version without losing safety.  (Would be even nicer if we could express the constraint in Sail's type system directly.)

> (I wouldn't say it's high priority though)

I sadly haven't had too much time for Sail recently, and it doesn't seem like it will change soon.

Timmmm:
> the runtime overhead of the computation

That's going to be completely insignificant. Mstatus statically fits in 64 bits so it will be inlined to just a few simple instructions.

Issue #433, opened 2024-03-15
Use result type instead of `MemoryOpResult`, `TR_Result`, and `PTW_Result`


Timmmm:
Sail has a standard `result` type in it's library now so we should replace these:

```
union PTW_Result('paddr : Type, 'pte : Type) = {
  PTW_Success: ('paddr, 'pte, 'paddr, nat, bool, ext_ptw),
  PTW_Failure: (PTW_Error, ext_ptw)
}

union TR_Result('paddr : Type, 'failure : Type) = {
  TR_Address : ('paddr, ext_ptw),
  TR_Failure : 'failure,
}

union MemoryOpResult ('a : Type) = {
  MemValue     : 'a,
  MemException : ExceptionType
}
```

with e.g:

```
type MemoryOpResult('a : Type) = result('a, ExceptionType)
```

martinberger:
There is a minor issue here in that Sail's type abbreviations (like `type MemoryOpResult('a : Type) = ...`) are *structural* rather than *nominal*. This might be most easy to see with an example. This compiles

```
type Giraffe = unit
type Dolphin = unit

function giraffe_identity(g : Giraffe) -> Giraffe = { g }
function dolphin_identity(d : Dolphin) -> Dolphin = { giraffe_identity(d) }
```

but the typing system is probably not enforcing what the author of the abbreviations `Giraffe` and `Dolphin` had in mind. This is not a problem in most case, especially with complex types, and indeed not in any of the examples above, but it's worth bearing in mind since it is not documented that Sail's type abbreviations are structural.

@Timmmm 
@Alasdair

Alasdair:
Yes, `type` introduces synonyms for existing types rather than creating new ones. You could do:
```
newtype Giraffe = Giraffe : unit
newtype Dolphin = Dolphin : unit
```
à la Haskell

martinberger:
**NICE** ... but how should we know that newtype exists?

Be that as it may, I think we should replace all occurrences of `type` by `newtype` (where that is possible) in the RISCV model to increase type safety. The model currently has a lot of aliases for e.g. `unit` and many `bits(n)` have aliases, e.g.

```
type vaddr32 = bits(32)
type pte32   = bits(32)

type pte64   = bits(64)
type pte48   = bits(64)
```

Alasdair:
[...]

Issue #443, opened 2024-03-31
Is vector extension still in development?


sequencer:
I'm the architect of [T1](https://github.com/chipsalliance/t1), a standard RVV implementation in Chisel. We originally use spike a golden model for co-simulation. Recently we want to migrate to sail for a formal-based verification flow. Thanks to @XinlaiWan, It's delightful to see there is an on going work on the vector extension support for RISC-V. However I saw the vector branch is 5 months ago, and wonder if there is a plan to merge it?
cc @billmcspadden-riscv, @XinlaiWan

Alasdair:
It's been merged

sequencer:
Nice! I’ll try it. Thank you!

PR #686, opened 2025-01-14
Use result type instead of MemoryOpResult

trdthg:
resolve part of #433

Use `result` type instead of `MemoryOpResult`，for `MemoryOpResult`, `MemValue` is `'a`, so I think it's ok to use `type`

But `PTW_Success` has five members, I did not include it in this PR, I think we should create a struct for it?

```ocaml
union PTW_Result('paddr : Type, 'pte : Type) = {
  PTW_Success: ('paddr, 'pte, 'paddr, nat, bool, ext_ptw),
  PTW_Failure: (PTW_Error, ext_ptw)
}
```

### DIFF
--- a/model/riscv_fetch.sail
+++ b/model/riscv_fetch.sail
@@ -30,8 +30,8 @@ function fetch() -> FetchResult =
            * exceptions.
            */
           match mem_read(Execute(), ppclo, 2, false, false, false) {
-            MemException(e) => F_Error(e, PC),
-            MemValue(ilo)   => {
+            Err(e)  => F_Error(e, PC),
+            Ok(ilo) => {
               if   isRVC(ilo)
               then F_RVC(ilo)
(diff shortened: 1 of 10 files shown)


Timmmm:
I think this is a good idea, but I vaguely recall some people didn't like it for some reason I can't quite remember. Can anyone remember?

Also I don't think you meant #443 ?

trdthg:
sorry，it's #433

Because `type` doesn't guarante type safety. 

But in this case, `MemValue` is any type, and its type safety is guaranteed by the incoming type, so I think it's fine

arichardson (approved):
Seems slightly more readable to me. I never really liked the memvalue/memexception verbosity

jordancarlin (approved):

Timmmm (approved):

PR #687, opened 2025-01-14
Fix match AccessType

trdthg:
```diff
-                            Read()          => false,
+                            Read(_)         => false,
```

I don't know why the previous one could compile, it should be wrong, but after replacing `ext_access_type` with newtype, the compiler detected the error correctly.

```
  Read      : 'a,
  Write     : 'a,
  ReadWrite : ('a, 'a),
  Execute   : unit
}
```

### DIFF
--- a/model/riscv_vmem_pte.sail
+++ b/model/riscv_vmem_pte.sail
@@ -117,7 +117,7 @@ function update_PTE_Bits(sv_params : SV_Params,
   let update_d : bool = (pte_flags[D] == 0b0)
                         & (match a {
                              Execute()       => false,
-                             Read()          => false,
+                             Read(_)         => false,
                              Write(_)        => true,
                              ReadWrite(_, _) => true
                            });


Timmmm:
Ah interesting. So the reason this wasn't an issue before is that Sail treats `unit` union variants specially:

```
union Foo = {
  A : unit,
  B : unit,
  C : unit,
  D : int,
}

function test(f : Foo) -> int =
  match f {
    // These are all ok for unit
    A(()) => 0,
    B() => 1,
    C(_) => 2,
    // But int requires _.
    D(_) => 3,
  }
```

So `Read()` would only work if `ext_access_type` is `unit`. Extended versions of the model are allowed to change it, and then `Read()` wouldn't work... except that the only extended version of the model that really exists and uses this is CHERI and that *also* replaces the entire `riscv_vmem_pte.sail` file too.

So yeah, we should change the `Read()` to `Read(_)` anyway, but I don't see any reason to use `newtype`?

trdthg:
done

arichardson (approved):

Timmmm (approved):

jordancarlin (approved):

PR #691, opened 2025-01-18
Make medeleg 64 bits and add medelegh

jordancarlin:
Per section 3.1.8 of the priv spec, `medeleg` is a 64 bit register and `medelegh` accesses the upper 32 bits on RV32.

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -447,7 +447,7 @@ function legalize_mideleg(o : Minterrupts, v : xlenbits) -> Minterrupts = {
 
 /* exception processing state */
 
-bitfield Medeleg : xlenbits = {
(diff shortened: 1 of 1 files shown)


Timmmm (approved):

Timmmm:
Looks correct and uncontroversial. `medelegh` is since priv version 1.12, but seeing as we don't have a versioning system yet (on my todo list) I don't think we should block this.

arichardson (approved):

jordancarlin:
> `medelegh` is since priv version 1.12, but seeing as we don't have a versioning system yet (on my todo list) I don't think we should block this.

Agreed. Once we have a versioning system we will probably need to go back through and gate a bunch of different things on priv versions, but this is a backwards compatible change so shouldn't be an issue until then.

PR #693, opened 2025-01-21
Added support to recognize hints instruction in the log

itsubaidrehman:
SAIL was unable to recognize hint operations as hints in the log. Added the support so that hints should be recognizable as hints in the log file

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -37,9 +37,14 @@ mapping utype_mnemonic : uop <-> string = {
   RISCV_LUI   <-> "lui",
   RISCV_AUIPC <-> "auipc"
 }
+mapping clause assembly = UTYPE(imm, rd, op)
+  if rd == zreg
+  <-> utype_mnemonic(op) ^ ".hint" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_signed_20(imm)
 
 mapping clause assembly = UTYPE(imm, rd, op)
+  if rd != zreg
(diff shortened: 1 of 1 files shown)


allenjbaum:
This is an interesting case. I am unaware of any tools that would recognize
these opcodes, and I'm unclear what the advantage of having the log
distinguish them.
At some point, these hints will have some (microarchitectural)
functionality, at which point there may be a use case, but until then: can
you articulate why this is needed?

Timmmm:
I assume this was intended just to make the log more informational? But yeah I agree with @jrtc27 - we should probably stick to assembly that actually works.

allenjbaum:
I talked to someone about this. The intent is to simplify tools (riscv-isac
specifically) to look at the log file and be able to easily identify a hint
instruction so they can measure coverage of hint ops.
There are probably other alternatives.
I the CSC meeting today we added (I think) a requirement that there be a
standard logging format, and this would affect that decision.

jrtc27:
Sounds like your tool needs an instruction decoder.

Timmmm:
Yeah I think the proper way to solve this is with the planned callback API or machine readable log format. You can extract the instructions opcode and then use a simple decoder to see if is a hint.

If it must be in Sail we could potentially add a second `instruction_metadata` mapping that maps an `ast` to information like whether it is a hint, which extension it is from, etc. Might be handy.

allenjbaum:
That is another option, I agree..

Timmmm:
I don't think we're going to do this so I'll close it. Hope that's ok!

PR #694, opened 2025-01-22
Add Smstateen/Ssstateen extension regs

trdthg:
Add some basic things, including

- cli param
- regs
	- bitfield
	- legalize_xxx
	- csr_name_map
	- is_CSR_defined
	- read_CSR
	- write_CSR 

This PR does not include any actual functionality

I plan to split other parts like FCSR into other PRs

> For convenience, when the stateen CSRs are implemented and misa.F = 0, then if the FCSR bit of a controlling stateen0 CSR is zero, all floating-point instructions cause an illegal instruction trap (or virtual instruction trap, if relevant),


model/riscv_stateen_regs.sail
```diff
@@ -0,0 +1,327 @@
...
+
+type Sstateen1 = xlenbits
+
+function legalize_sstateen1(c : Sstateen1, v : xlenbits) -> Sstateen1 = {
+  c
+}
+
+register sstateen1 : Sstateen1
+mapping clause csr_name_map = 0x10D  <-> "sstateen1"
+function clause is_CSR_defined(0x10D) = extensionEnabled(Ext_Ssstateen)
```
jordancarlin, line 173: From the spec:
> The Smstateen extension specification comprises the mstateen*, sstateen*, and hstateen* CSRs and their functionality. The Ssstateen extension specification comprises only the sstateen* and hstateen* CSRs and their functionality.

This seems to mean that `Sstateen1` (and all of the other S and H CSRs defined here) should accessible if `extensionEnabled(Ext_ Smstateen) | extensionEnabled(Ext_ Ssstateen)`. 

Alternatively, we could potentially define `extensionEnabled(Ext_ Ssstateen) = sys_enable_ssstateen() | extensionEnabled(Ext_ Smstateen)`.

model/riscv_stateen_regs.sail
```diff
@@ -0,0 +1,327 @@
...
+  m
+}
+
+register mstateen3 : Mstateen3
+mapping clause csr_name_map = 0x30F  <-> "mstateen3"
+function clause is_CSR_defined(0x30F) = extensionEnabled(Ext_Smstateen)
+function clause read_CSR(0x30F) = zero_extend(mstateen3)
+function clause write_CSR(0x30F, value) = { mstateen3 = legalize_mstateen3(mstateen3, value); zero_extend(mstateen3) }
+
+type Mstateen0h = xlenbits
```
jordancarlin, line 90: This doesn't seem true? `Mstateen0h` should be the top 32 bits of the `Mstateen` register, just like all of the other h CSRs. 

`Mstateen` should be defined as 64 bits and the write/read should always access that register, either the entire register (for rv64) or the appropriate half of the register (for rv32). This issue is present for all of the mstateen and hstateen registers. See how `menvcfg` and `menvcfgh` are implemented.

jordancarlin (changes requested):
Haven't looked at the functionality of its interaction with the rest of the system yet, but here are a few initial comments.

model/riscv_stateen_regs.sail
trdthg: I did not find the specific dependencies for the `Sm/Ss` extensions in the spec, so I replaced them with

```
function enable_mstatenn() -> bool = extensionEnabled(Ext_Smstateen)
function enable_sstatenn() -> bool = extensionEnabled(Ext_Smstateen) | extensionEnabled(Ext_Ssstateen)
function enable_hstatenn() -> bool = extensionEnabled(Ext_Smstateen) | extensionEnabled(Ext_Ssstateen)
```

What do you think?

jordancarlin: I don’t love introducing yet another level of indirection to enabling/disabling extensions, but it seems like an acceptable solution for this situation.

model/riscv_stateen_regs.sail
```diff
@@ -0,0 +1,308 @@
...
+
+function clause extensionEnabled(Ext_Smstateen) = sys_enable_smstateen()
+function clause extensionEnabled(Ext_Ssstateen) = sys_enable_ssstateen()
+
+function enable_mstatenn() -> bool = extensionEnabled(Ext_Smstateen)
+function enable_sstatenn() -> bool = extensionEnabled(Ext_Smstateen) | extensionEnabled(Ext_Ssstateen)
+function enable_hstatenn() -> bool = extensionEnabled(Ext_Smstateen) | extensionEnabled(Ext_Ssstateen)
+
+bitfield Mstateen0 : bits(64) = {
+    SEO     : 63,
```
nadime15, line 17: Small detail but its actually SE0 (zero) and not SEO (letter).

nadime15 (commented):
Additionally, on chip reset `mstateen*` CSR's are set to 0.

model/riscv_stateen_regs.sail
```diff
@@ -0,0 +1,308 @@
...
+    ENVCFG  : 62,
+    WRPI    : 61,
```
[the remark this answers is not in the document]
jordancarlin: Seems like there is no reason to define this bit at all. We don't define any of the other WPRI bits since they should all be read-only zero.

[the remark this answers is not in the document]
trdthg: removed

model/riscv_stateen_regs.sail
trdthg: thanks! fixed

model/riscv_stateen_regs.sail
trdthg: Add two additional constraints
```
// 'If supervisor mode is implemented'
function enable_sstatenn() -> bool = extensionEnabled(Ext_S) & (extensionEnabled(Ext_Smstateen) | extensionEnabled(Ext_Ssstateen))
// 'And if the hypervisor extension is implemented'
function enable_hstatenn() -> bool = false & (extensionEnabled(Ext_Smstateen) | extensionEnabled(Ext_Ssstateen))
```

trdthg:
You are right, I will check them again with spike and qemu.

nadime15:
Hi @trdthg, any updates on this PR? If you're no longer working on it, I would be happy to take it over.

trdthg:
> I would be happy to take it over.

Just a little rebase, feel free to pick-up

Here is the Spike stateen PR, which may be helpful https://github.com/riscv-software-src/riscv-isa-sim/pull/1035/files

jordancarlin:
Closing now that #910 is merged.

PR #695, opened 2025-01-22
Update support for Coq output

bacam:
It requires an up-to-date development version of Sail because I fixed a couple of bugs while working on it, but only for Coq output, it shouldn't affect any other target.


arichardson (approved):
Seems good to me. It will conflict with #647, so not sure which one should be merged first.

jordancarlin (approved):
LGTM

jordancarlin:
> It will conflict with #647, so not sure which one should be merged first.

Seems like it would be easier to merge this one and then cleanup the Coq cmake target, but I'll leave that up to @Timmmm

bacam:
AIUI #647 will leave the theorem prover targets in the `Makefile`, so it should just be a matter of rebasing the changes to the `Makefile`, regardless of what order we merge in.  I might have a go at getting them to build in cmake later on.

Timmmm:
Yeah I think we should merge this first. Given that it fixes the Coq target I might be able to move it to CMake.

Timmmm (approved):

PR #697, opened 2025-01-24
Remove unused attempt at a platform memory map.

pmundkur:
The memory segments are hard-coded into `within_phys_mem` anyway and its log messages.  A more general map would need a better API to the C emulator.


Timmmm:
LGTM. We have PMA support [here](https://github.com/CHERI-Alliance/sail-riscv/blob/codasip/model/riscv_pma.sail) but I'm going to wait until the new config system before making a PR.

Timmmm (approved):

jordancarlin (approved):

PR #699, opened 2025-01-28
Move csr_name function back to desired location

jordancarlin:
`csr_name` had been moved because of a bug in the Sail compiler. Seems like this is no longer an issue so move it back to where it is supposed to be.


arichardson (approved):
LGTM if CI is happy with it

Timmmm (approved):

PR #700, opened 2025-01-28
Remove unused report_invalid_width function

jordancarlin:
`report_invalid_width` is unused as of #680 and seems to be a remnant from issues with an earlier version of the compiler.

### DIFF
--- a/model/riscv_types.sail
+++ b/model/riscv_types.sail
@@ -376,17 +376,3 @@ struct mul_op = {
   signed_rs1 : bool,
   signed_rs2 : bool
 }
-
-/*!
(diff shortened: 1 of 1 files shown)


arichardson (approved):
Commit message should mention that the last use was removed in https://github.com/riscv/sail-riscv/pull/680, otherwise looks great

jordancarlin:
> Commit message should mention that the last use was removed in #680, otherwise looks great

Commit and PR messages updated

Timmmm (approved):

arichardson (approved):

PR #701, opened 2025-01-28
Replace uses of sail_zero_extend with zero_extend

jordancarlin:
`sail_zero_extend` is aliased to `zero_extend`, and `zero_extend` is already used in most places


jordancarlin:
Strange. This passed locally. I'll look into why CI is failing.

jordancarlin (commented):
Was a git issue on my end. Seems to be working now.

Timmmm (approved):

arichardson (approved):
Using both variants is confusing. LGTM.

PR #703, opened 2025-01-29
Minor architecture cleanup

radimkrcmar:
First three patches of https://github.com/riscv/sail-riscv/pull/683.

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -712,9 +712,8 @@ function clause execute SFENCE_VMA(rs1, rs2) = {
   match cur_privilege {
     User       => { handle_illegal(); RETIRE_FAIL },
     Supervisor => match (architecture(get_mstatus_SXL(mstatus)), mstatus[TVM]) {
-                    (Some(_), 0b1)  => { handle_illegal(); RETIRE_FAIL },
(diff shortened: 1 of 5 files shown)


model/riscv_insts_base.sail
```diff
@@ -712,9 +712,8 @@ function clause execute SFENCE_VMA(rs1, rs2) = {
   match cur_privilege {
     User       => { handle_illegal(); RETIRE_FAIL },
     Supervisor => match (architecture(get_mstatus_SXL(mstatus)), mstatus[TVM]) {
-                    (Some(_), 0b1)  => { handle_illegal(); RETIRE_FAIL },
```
arichardson, line 715: Can we drop the first tuple element in this match?

arichardson (approved):

model/riscv_insts_base.sail
radimkrcmar: Yes, and patch https://github.com/riscv/sail-riscv/pull/683/commits/cf121a07dd7845e7916005b6d5685148871f83e8 does just that, I was just told to split the series this way.

Timmmm (approved):
LGTM, thanks!

jordancarlin (approved):

PR #705, opened 2025-02-04
Fix a corner case in zicsr: avoid a read-effect in CSRRW when rd is 0.

pmundkur:
While here, remove an unused mapping intended for the assembly clause.

### DIFF
--- a/model/riscv_insts_zicsr.sail
+++ b/model/riscv_insts_zicsr.sail
@@ -22,17 +22,16 @@ mapping clause encdec = CSR(csr, rs1, rd, is_imm, op)
 
 function clause execute CSR(csr, rs1, rd, is_imm, op) = {
   let rs1_val : xlenbits = if is_imm then zero_extend(rs1) else X(rs1);
-  let isWrite : bool = match op {
-    CSRRW  => true,
-    _      => if is_imm then unsigned(rs1_val) != 0 else unsigned(rs1) != 0
-  };
-  if not(check_CSR(csr, cur_privilege, isWrite))
+  let is_CSR_Write = op == CSRRW | rs1 != zeros();
+  if not(check_CSR(csr, cur_privilege, is_CSR_Write))
   then { handle_illegal(); RETIRE_FAIL }
-  else if not(ext_check_CSR(csr, cur_privilege, isWrite))
+  else if not(ext_check_CSR(csr, cur_privilege, is_CSR_Write))
   then { ext_check_CSR_fail(); RETIRE_FAIL }
   else {
-    let csr_val = read_CSR(csr); /* could have side-effects, so technically shouldn't perform for CSRW[I] with rd == 0 */
-    if isWrite then {
+    /* CSRRW should not generate read side-effects if rd == 0 */
+    let is_CSR_Read = not(op == CSRRW & rd == zeros());
+    let csr_val : xlenbits = if is_CSR_Read then read_CSR(csr) else zeros();
+    if is_CSR_Write then {
       let new_val : xlenbits = match op {
         CSRRW => rs1_val,
         CSRRS => csr_val | rs1_val,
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):

model/riscv_insts_zicsr.sail
```diff
@@ -31,7 +31,11 @@ function clause execute CSR(csr, rs1, rd, is_imm, op) = {
   else if not(ext_check_CSR(csr, cur_privilege, isWrite))
   then { ext_check_CSR_fail(); RETIRE_FAIL }
   else {
-    let csr_val = read_CSR(csr); /* could have side-effects, so technically shouldn't perform for CSRW[I] with rd == 0 */
+    let csr_val : xlenbits = match (op, rd) {
+      /* CSRRW should not generate read side-effects if rd == 0 */
+      (CSRRW, 0b00000) => zeros(),
+      _                => read_CSR(csr),
+    };
```
Timmmm, line 38: We did it like this, which I think is a bit nicer:

```
  // CSRRW[I] always writes. CSRRS[I] and CSRRC[I] write if the
  // source register (or immediate) is not x0 (or 0).
  let isWrite = op == CSRRW | rs1 != zeros();
  // CSRRS[I] and CSRRC[I] always read. CSRRW[I] reads if the destination
  // register is not x0.
  let isRead = op != CSRRW | rd != zeros();
...
    let csr_val = if isRead then read_CSR(csr) else zeros();
```

(Sorry; on my list of things to upstream.)

pmundkur: That works too.  I would have preferred 'isCSRWrite' and 'isCSRRead' (or 'is_CSR_Write' ...) to be less confusing, since rs1 and rd are also being read and written.  I can edit the commit to the above if you agree.

Timmmm: Yeah that makes sense.

jordancarlin (approved):

model/riscv_insts_zicsr.sail
```diff
@@ -22,17 +22,16 @@ mapping clause encdec = CSR(csr, rs1, rd, is_imm, op)
 
 function clause execute CSR(csr, rs1, rd, is_imm, op) = {
   let rs1_val : xlenbits = if is_imm then zero_extend(rs1) else X(rs1);
-  let isWrite : bool = match op {
-    CSRRW  => true,
-    _      => if is_imm then unsigned(rs1_val) != 0 else unsigned(rs1) != 0
-  };
-  if not(check_CSR(csr, cur_privilege, isWrite))
+  let is_CSR_Write = op == CSRRW | unsigned(rs1) != 0;
```
Timmmm, line 25: May as well use `zeros()` here too I guess.

```suggestion
  let is_CSR_Write = op == CSRRW | rs1 != zeros();
```

model/riscv_insts_zicsr.sail
```diff
@@ -22,17 +22,16 @@ mapping clause encdec = CSR(csr, rs1, rd, is_imm, op)
...
+  if not(check_CSR(csr, cur_privilege, is_CSR_Write))
   then { handle_illegal(); RETIRE_FAIL }
-  else if not(ext_check_CSR(csr, cur_privilege, isWrite))
+  else if not(ext_check_CSR(csr, cur_privilege, is_CSR_Write))
   then { ext_check_CSR_fail(); RETIRE_FAIL }
   else {
-    let csr_val = read_CSR(csr); /* could have side-effects, so technically shouldn't perform for CSRW[I] with rd == 0 */
-    if isWrite then {
+    /* CSRRW should not generate read side-effects if rd == 0 */
+    let is_CSR_Read = ~(op == CSRRW & rd == zeros());
```
arichardson, line 32:
```suggestion
    let is_CSR_Read = not(op == CSRRW & rd == zeros());
```
Since this is a bool?

arichardson (approved):

Timmmm (approved):

arichardson (approved):

PR #706, opened 2025-02-04
Implement Sscofpmf (counter overflow and privilege mode filtering)

Timmmm:
This adds the `mhpmeventNh` and `scountovf` CSRs. It also fixes `mhpmevent` to be 64 bits instead of `xlenbits` (though this mistake is unobservable without Sscofpmf).

Since the model doesn't ever actually increment any of the HPM counters yet, there is no need to actually implement the filtering logic or the overflow interrupt. If/when we add some example HPM sources we will need to do that.


arichardson (approved):
Looks reasonable to me and matches a quick read of the spec.

jordancarlin (commented):
Seems to match the spec and generally looks good to me. `Sscofpmf` needs to be added to the list of supported extensions in the README.

jordancarlin (approved):

PR #708, opened 2025-02-05
Add tabs check to pre-commit.

pmundkur:
Also fix some whitespace issues.  This PR updates the minimum required version of pre-commit to 3.2.0; I do not know if  that is too recent for CI.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
 exclude: '^(prover_snapshots)|(generated_definitions)|(dependencies/[^/]+/)'
-minimum_pre_commit_version: 2.10.0
+minimum_pre_commit_version: 3.2.0
 repos:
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -21,3 +21,8 @@ repos:
     hooks:
     -   id: clang-format
         types_or: [c, c++]  # Don't try running clang-format on .json files
+-   repo: https://github.com/Lucas-C/pre-commit-hooks
+    rev: v1.5.5
+    hooks:
+    -   id: forbid-tabs
```
jordancarlin, line 27: I wonder if we should be using the `remove-tabs` hook instead of `forbid-tabs`. Seems like that would be more similar to our existing hooks that automatically strip trailing white space and fix file endings.

Timmmm: I dunno I think `forbid-tabs` is probably safer since we haven't been super consistent about indentation size, and we probably don't want it to silently get it wrong. Stripping whitespace is unambiguous.

jordancarlin: Works for me

jordancarlin (approved):

PR #710, opened 2025-02-05
Add support for Sail unit tests

Timmmm:
Note: This is a draft because it relies on [this Sail compiler branch](https://github.com/rems-project/sail/pull/928) that I haven't finished yet.

----------------

This adds basic support for Sail unit tests. These have some pros and cons over C tests:

Pros:

* They don't require a RISC-V C compiler.
* They can test internal code (hence why I've called them unit tests).
* They can "cheat", e.g. to change privilege mode you can just magically `cur_privilege = Supervisor`. With C you have to implement a syscall, etc.

Cons:

* They are tied to internal code which will make refactoring more tedious because it means you are likely to need to update the tests. We can minimise that by using "high level" functions like `execute()`, `read_CSR()`, etc.
* They only run on the Sail model, so we can't e.g. verify them against SPIKE.

Given the sorry state of testing in this repo, and the fact that we don't have a way of writing C tests at all yet, I think this is probably a good medium term approach.

Currently the unit test executable reuses a load of code from the emulator, which is not ideal. When we have wrapped the model in a nice C++ library we can use that instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,9 @@ add_subdirectory("c_emulator")
 # Old pre-compiled riscv-tests.
 add_subdirectory("test/riscv-tests")
 
+# Sail unit tests.
+add_subdirectory("test/unit_tests")
+
 # Convenience targets.
 add_custom_target(csim DEPENDS riscv_sim_rv32d riscv_sim_rv64d)
 add_custom_target(check DEPENDS generated_model_rv32d generated_model_rv64d)
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -189,12 +189,18 @@ foreach (xlen IN ITEMS 32 64)
                 )
             endif()
 
+            set(test_srcs
+                "unit_tests/test_mstatus.sail"
+                "unit_tests/test_pte_update_access_fault.sail"
+            )
+
             # Final file list.
             set(sail_srcs
                 ${sail_arch_srcs}
                 ${sail_seq_inst_srcs}
                 ${sail_step_srcs}
                 "main.sail"
+                ${test_srcs}
             )
 
             # Convert to absolute paths, so we can run
--- a/c_emulator/CMakeLists.txt
+++ b/c_emulator/CMakeLists.txt
@@ -1,4 +1,4 @@
-set(EMULATOR_COMMON_SRCS
+add_library(riscv_platform
     riscv_config.h
     riscv_platform.c
     riscv_platform.h
@@ -7,9 +7,17 @@ set(EMULATOR_COMMON_SRCS
     riscv_prelude.c
     riscv_prelude.h
     riscv_sail.h
-    riscv_sim.c
     riscv_softfloat.c
     riscv_softfloat.h
+)
+
+target_include_directories(riscv_platform
+    # So the generated C can find riscv_platform/prelude.h"
+    PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}"
+)
+
+target_link_libraries(riscv_platform
+    PUBLIC sail_runtime
 )
 
 foreach (xlen IN ITEMS 32 64)
@@ -27,7 +35,7 @@ foreach (xlen IN ITEMS 32 64)
 
             add_executable(riscv_sim_${arch}
                 "${CMAKE_BINARY_DIR}/riscv_model_${arch}.c"
-                ${EMULATOR_COMMON_SRCS}
+                "riscv_sim.c"
             )
 
             if (NOT arch IN_LIST DEFAULT_ARCHITECTURES)
@@ -37,12 +45,7 @@ foreach (xlen IN ITEMS 32 64)
             add_dependencies(riscv_sim_${arch} generated_model_${arch})
 
             target_link_libraries(riscv_sim_${arch}
-                PRIVATE softfloat sail_runtime GMP::GMP ZLIB::ZLIB
-            )
-
-            target_include_directories(riscv_sim_${arch}
-                # So the generated C can find riscv_platform/prelude.h"
-                PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}"
+                PRIVATE riscv_platform
             )
 
             if (arch MATCHES "rvfi")
--- a/model/unit_tests/test_mstatus.sail
+++ b/model/unit_tests/test_mstatus.sail
@@ -1,0 +1,21 @@
+// Verify that the reset sxl and uxl are correct.
+$[test]
+function test_mstatus_sxl_uxl_reset_values() -> unit = {
+  let mstatush_val : bits(32) = match xlen {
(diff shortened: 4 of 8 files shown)


jordancarlin:
Feels like these have the potential to be a lower barrier of entry for writing certain kinds of tests,(especially for tricky to reach edge cases of the system), so seems like a good idea to me. Once we get the C/assembly test situation sorted out we can think about what makes sense to run as a C test versus a unit test, but having both options seems like a net positive.

jordancarlin:
@Timmmm Now that Sail 0.19 is released with unit test support this is something that can be finished up if we still want it. Not sure that there are a whole lot of tests that should exist here instead of as first party assembly tests, but seems reasonable to include support for them.

model/CMakeLists.txt
```diff
@@ -200,12 +200,17 @@ foreach (xlen IN ITEMS 32 64)
                 )
             endif()
 
+            set(test_srcs
```
jordancarlin, line 203: Is it worth adding a flag to control whether these tests are built and included in the model? They're not necessary for end users.

test/unit_tests/main_unit_tests.cpp
```diff
@@ -0,0 +1,41 @@
...
+bool config_print_mem_access = false;
+bool config_print_platform = false;
+bool config_enable_rvfi = false;
+FILE *trace_log = NULL;
+
+int main()
+{
+  trace_log = stdout;
+
+  std::filesystem::path path = std::filesystem::temp_directory_path();
```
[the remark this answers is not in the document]
Timmmm: Done.

CMakeLists.txt
```diff
@@ -166,6 +166,9 @@ else()
 endif()
 include(CPack)
 
+# Sail unit tests.
+add_subdirectory("test/unit_tests")
```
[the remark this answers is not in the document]
Timmmm: Done.

jordancarlin (approved):
Might be worth adding a README to the `test/unit_tests` directory explaining that the tests themselves are in the `model/unit_tests` directory or directly inline with the model. Otherwise, LGTM. Great to get more testing in!

Timmmm:
Rebased & README added.

Timmmm:
I'll see if I can sort out the Lean errors.

Timmmm:
Rebased but it's broken now and awkward to fix due to using the Sail generated header. Will probably have to wait until runtime XLEN is merged.

Timmmm:
Rebased & tweaked a little to handle the changes since. I'll give this a couple of days before merging in case anyone wants to take a look.

test/unit_tests/main_unit_tests.cpp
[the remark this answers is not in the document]
jordancarlin: This can be updated now that we’re using a new compiler.

Timmmm:
Ah yeah good call - also since the XLEN stuff was merged I can move the generated C into the library. I renamed it to `riscv_model` since it's not just the support code now.

jordancarlin:
Seems to have broken Lean:
```
error: build failed
✖ [45/46] Building LeanRV64D
trace: .> LEAN_PATH=/home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean /home/runner/.elan/toolchains/leanprover--lean4-nightly---nightly-2025-06-21/bin/lean --tstack=400000 /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/LeanRV64D.lean -R /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D -o /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D.olean -i /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D.ilean -c /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D.c --setup /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D.setup.json --json
error: LeanRV64D.lean:171:32: unknown identifier 'undefined_RVFI_DII_Instruction_Packet'
error: LeanRV64D.lean:172:30: unknown identifier 'undefined_RVFI_DII_Execution_Packet_InstMetaData'
error: LeanRV64D.lean:173:28: unknown identifier 'undefined_RVFI_DII_Execution_Packet_PC'
error: LeanRV64D.lean:174:29: unknown identifier 'undefined_RVFI_DII_Execution_Packet_Ext_Integer'
error: LeanRV64D.lean:176:29: unknown identifier 'undefined_RVFI_DII_Execution_Packet_Ext_MemAccess'
error: LeanRV64D.lean:355:91: unknown identifier 'sail_main'
```

Alasdair:
That's the same issue I've been having with my modules PR, so maybe the issue is unrelated if it's here too

Timmmm:
Ok I had a look into this and the issue is that Lean is treating the very *last* function in the code as special. On `master` the `LeanRV64D.lean` file starts with

```
import LeanRV64D.Main
```

but on this branch it starts with

```
import LeanRV64D.TestMstatus
```

So the problem is I added a file after `main.sail`. I guess your PR does the same thing?

I'll reorder the files in this PR since it doesn't make much difference but that seems like something that should be fixed.

test/unit_tests/main_unit_tests.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+FILE *trace_log = NULL;
+
+int main()
+{
+  trace_log = stdout;
+
+  sail_config_set_string(DEFAULT_JSON);
+  sail_set_abstract_xlen();
+  sail_set_abstract_ext_d_supported();
```
jordancarlin, line 29: Might eventually be worth factoring these abstract type initializations out so they don't get out of sync with the main simulator as we add new abstract types (vlen, etc.). Doesn't need to happen in this PR though.

pmundkur: Same with the `config_*` and `trace_log`.

pmundkur (approved):
It would be great to finally get this in!

test/unit_tests/main_unit_tests.cpp
Timmmm: Yeah I still don't quite understand why they aren't called automatically in `model_init()`.

Alasdair: Maybe they could be? If we ensured calling them is idempotent moving them into `model_init` shouldn't be a breaking change. I may have been overthinking some ordering issue between setting up abstract types and initialising letbindings and registers.

pmundkur: Slightly different topic: is there a reason for `model_init()` not being included in the generated C header?

PR #713, opened 2025-02-06
Simplify some base instruction execute clauses

Timmmm:
Refactor some of the base `execute` clauses to remove intermediate variables and unnecessary `if`s. The motivation for this is to make them read a bit better when included in the ISA manual.

1. Make `shift_right_arith` generic so you don't need a 32/64-bit version.
2. Use `rs1_val` in fewer places - it's actually longer than `X(rs1)` anyway.
3. Directly assign to `X(rd)` where possible.
4. Avoid single letter variables.

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -249,10 +249,10 @@ function clause execute (RTYPE(rs2, rs1, rd, op)) = {
     RISCV_AND  => X(rs1) & X(rs2),
     RISCV_OR   => X(rs1) | X(rs2),
     RISCV_XOR  => X(rs1) ^ X(rs2),
-    RISCV_SLL  => X(rs1) << X(rs2)[log2_xlen_bytes + 2 .. 0],
-    RISCV_SRL  => X(rs1) >> X(rs2)[log2_xlen_bytes + 2 .. 0],
+    RISCV_SLL  => X(rs1) << X(rs2)[log2_xlen - 1 .. 0],
+    RISCV_SRL  => X(rs1) >> X(rs2)[log2_xlen - 1 .. 0],
     RISCV_SUB  => X(rs1) - X(rs2),
-    RISCV_SRA  => shift_bits_right_arith(X(rs1), X(rs2)[log2_xlen_bytes + 2 .. 0]),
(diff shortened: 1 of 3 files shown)


model/riscv_insts_base.sail
```diff
@@ -215,20 +209,12 @@ mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SRLI) <-> 0b000000 @ sham
...
-    RISCV_SRLI => if   xlen == 32
-                  then rs1_val >> shamt[4..0]
-                  else rs1_val >> shamt,
-    RISCV_SRAI => if   xlen == 32
-                  then shift_right_arith32(rs1_val, shamt[4..0])
-                  else shift_right_arith64(rs1_val, shamt)
+  // The decoder guard ensures that shamt[5] = 0 for RV32.
+  X(rd) = match op {
+    RISCV_SLLI => X(rs1) << shamt,
+    RISCV_SRLI => X(rs1) >> shamt,
```
jordancarlin, line 215: Definitely simpler code this way, but I wonder if the other version is better for sake of inclusion in the spec. If I'm not mistaken it is only the execute clause that gets included, so the distinction about shamt[5] might be lost this way.

I think either way is fine though.

jordancarlin (approved):

model/riscv_insts_base.sail
Timmmm: Yeah ... though I think if we include the execute clause in both places (the RV32 and RV64 bits that talk about shifting) it should still be reasonably clear because they have `shamt[4:0]` and `shamt[5:0]` in the encoding diagrams. I think this code is better enough that I'd prefer to keep it anyway.

Timmmm:
I added `log2_xlen` - I think `[log2_xlen - 1 .. 0]` is much clearer than `[log2_xlen_bytes + 2 .. 0]` (but I still really wish we could do `[log2_xlen >.. 0]`).

model/riscv_insts_base.sail
jordancarlin: Fine with me

jordancarlin (approved):

PR #714, opened 2025-02-06
Refactor PTE update code & fix reachable assertion

Timmmm:
Deduplicate the code to write the PTE, and in the process remove the reachable (I think) assertion "invalid physical address in TLB".

It actually ended up being basically the same amount of code, so we may want to e.g. inline `update_pte_bits`. However I haven't done that because this code is going to get a bit more complex with CHERI.


Timmmm:
I'm working on a test to confirm that assertion is reachable.

pmundkur:
Closing this since this was merged as part of #1599.

Timmmm:
I don't think it was fully merged - e.g. update_PTE_bits wasn't changed. I'll reopen it as a draft & rebase at some point.

Timmmm:
The reachable assertion is now covered by #1664. Some of the other refactoring might still be relevant. I need to double check.

PR #717, opened 2025-02-06
Simplify mstatus[SD] bit handling

Timmmm:
I'm not sure why we didn't think of this before, but you can have the SD field positioned relative to xlen, so it's at 63 on RV64 and 31 on RV31. This simplifies the code a lot.

Note #683 has an alternative solution that just calculates SD on read, but I'm not sure we need that after this simplification. In any case we can still do it if desired.

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -88,7 +88,7 @@ register f31 : fregtype
 function dirty_fd_context() -> unit = {
   assert(sys_enable_fdext());
   mstatus[FS] = extStatus_to_bits(Dirty);
-  mstatus = set_mstatus_SD(mstatus, 0b1);
+  mstatus[SD] = 0b1;
 }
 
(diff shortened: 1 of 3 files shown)


Timmmm:
@KotorinMinami fyi

jordancarlin (approved):
Great solution! I don't know how we didn't think of this before.

radimkrcmar:
It was like this before the patch from @KotorinMinami, so I thought it was somehow not desired.

The dynamic SD bit has three advantages 
* dynamic xlen doesn't need to shuffle the bit around
* we don't need to increase mstatus size to 128 when RV128 lands
* internal sail code cannot make a mistake and use the SD bit with incorrect value

I'll have time to rebase this weekend, so we can discuss it there.

Timmmm:
> It was like this before the patch from @KotorinMinami, so I thought it was somehow not desired.

Oops so it was. I guess I was just being dumb!

PR #718, opened 2025-02-06
Improve Zicbom implementation.

pmundkur:
The bitwise check of xenvcfg.CBIE to check whether cbo.inval was enabled was incorrect, since CBIE is a four-valued setting. Specifically, it decoupled the bits in the reserved value 0b10 of CBIE.

This fixes the handling of the reserved value, and also throws an internal error when it is detected to indicate a possible bug in xenvcfg legalization.

The privilege check should extend to the hypervisor case, when two different illegal exceptions could be thrown.  This check as written doesn't match the pseudo-Sail in the CMO spec, but is arguably clearer.

I've left the CBCFE handling alone, since it works for the single bit case.

I'd like to make sure that the handling of no S-mode case is correct (the pseudo-Sail in the spec doesn't seem to handle it explicitly).

### DIFF
--- a/model/riscv_insts_zicbom.sail
+++ b/model/riscv_insts_zicbom.sail
@@ -11,8 +11,6 @@
 function clause extensionEnabled(Ext_Zicbom) = sys_enable_zicbom()
 
 function cbo_clean_flush_enabled(p : Privilege) -> bool = feature_enabled_for_priv(p, menvcfg[CBCFE][0], senvcfg[CBCFE][0])
-function cbo_inval_enabled(p : Privilege) -> bool = feature_enabled_for_priv(p, menvcfg[CBIE][0], senvcfg[CBIE][0])
-function cbo_inval_as_inval(p : Privilege) -> bool = feature_enabled_for_priv(p, menvcfg[CBIE][1], senvcfg[CBIE][1])
 
 /* ****************************************************************** */
 union clause ast = RISCV_ZICBOM : (cbop_zicbom, regidx)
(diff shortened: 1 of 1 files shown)


Timmmm:
It wasn't *incorrect*. You can do anything with reserved values so I chose the (much) simpler implementation. And as you noted the legalisation prevents that value anyway (though I do like being able to detect mistakes in legalisation).

I think splitting the instructions up is probably a good idea (makes embedding in the spec nicer)... but it is definitely more complex. Let's discuss it in the meeting on Monday.

pmundkur:
Sounds good; thanks for the review.  Yeah, _incorrect_ was probably too strong, apologies.  I need to understand how we should handle reserved values; I'm not convinced we can do anything with them.

Timmmm:
Yeah I agree we don't quite have a clear strategy for reserved values.

I think we should probably have a global flag that allows two modes of operation:

* Any reserved behaviour immediately asserts.
* Reserved behaviour does something sensible, like a real chip would.

Let's discuss that too.

model/riscv_insts_zicbom.sail
```diff
@@ -35,6 +33,38 @@ mapping cbop_mnemonic : cbop_zicbom <-> string = {
 mapping clause assembly = RISCV_ZICBOM(cbop, rs1)
   <-> cbop_mnemonic(cbop) ^ spc() ^ "(" ^ opt_spc() ^ reg_name(rs1) ^ opt_spc() ^ ")"
 
+/* CBIE from xenvcfg */
+enum cbie = {CBIE_ILLEGAL, CBIE_EXEC_FLUSH, CBIE_RESERVED, CBIE_EXEC_INVAL}
+
+mapping encdec_cbie : cbie <-> bits(2) = {
+  CBIE_ILLEGAL    <-> 0b00,
+  CBIE_EXEC_FLUSH <-> 0b01,
+  CBIE_RESERVED   <-> 0b10,
```
radimkrcmar, line 42: How about handing the reserved value here instead?
```
- CBIE_RESERVED   <-> 0b10,
+ backwards 0b10 => internal_error(__FILE__, __LINE__, "reserved CBIE"),
```
This would allow us to remove some matches below.

(Making it impossible to set the reserved value in *envcfg would be better, but that is a bigger change.)

pmundkur: I don't think the encdec decoding of 10 to `CBIE_RESERVED` is an `internal_error`, so I'm not sure that's where it should be.

One of the goals of the PR was to get the privilege check in a form that can replace the pseudo-code in the Zicbom spec, and to be explicit about the handling of the reserved value there.  

Afaict, the `legalize_xenvcfg` already make it impossible to set the reserved value.

jrtc27: > Afaict, the legalize_xenvcfg already make it impossible to set the reserved value.

But that's the point. The architecture's WARL rules mean it is architecturally impossible to ever have a reserved value in the register. So if it does end up happing, that is an internal error / assertion failure / whatever you want to call it; it is an impossible state that does not need any architecturally-defined behaviour. Ideally that state wouldn't even be representable, e.g. by having the actual state use an enum and convert it to bits in CSRRW etc.

pmundkur: That is a good point: the encdec should not make the reserved value representable.  I'll push a fix.

model/riscv_insts_zicbom.sail
```diff
@@ -35,6 +33,35 @@ mapping cbop_mnemonic : cbop_zicbom <-> string = {
...
   <-> cbop_mnemonic(cbop) ^ spc() ^ "(" ^ opt_spc() ^ reg_name(rs1) ^ opt_spc() ^ ")"
 
+/* CBIE from xenvcfg */
+enum cbie = {CBIE_ILLEGAL, CBIE_EXEC_FLUSH, CBIE_EXEC_INVAL}
+
+mapping encdec_cbie : cbie <-> bits(2) = {
+  CBIE_ILLEGAL    <-> 0b00,
+  CBIE_EXEC_FLUSH <-> 0b01,
+  CBIE_EXEC_INVAL <-> 0b11,
+  backwards 0b11   => internal_error(__FILE__, __LINE__, "reserved CBIE"),
```
arichardson, line 43:
```suggestion
  backwards 0b10   => internal_error(__FILE__, __LINE__, "reserved CBIE"),
```
Isn't it this one?

pmundkur: Oops, good catch, thanks!

radimkrcmar: I think it would be nice if the type system caught this kind of mistakes...
Can a bijective mapping be expressed in Sail?

Timmmm (approved):
LGTM

PR #721, opened 2025-02-08
c_emulator: fix cmake install with unbuilt targets

radimkrcmar:
When doing a default installation, CMake fails when it tries to install the unbuilt targets:
```
CMake Error at c_emulator/cmake_install.cmake:60 (file):
  file INSTALL cannot find "/build/source/build/c_emulator/riscv_sim_rv32f":
  No such file or directory.
```
Make all targets optional to avoid the failure;  this will install only the targets that are built.

### DIFF
--- a/c_emulator/CMakeLists.txt
+++ b/c_emulator/CMakeLists.txt
@@ -61,7 +61,7 @@ foreach (xlen IN ITEMS 32 64)
             #     -Wno-unused-parameter
             # )
 
-            install(TARGETS riscv_sim_${arch})
+            install(TARGETS riscv_sim_${arch} OPTIONAL)
         endforeach()
     endforeach()
 endforeach()


jordancarlin (approved):
I had the same issue but hadn’t gotten around to fixing it. LGTM

arichardson (approved):

Timmmm:
Should we make it so that the default targets at least are required? (See `DEFAULT_ARCHITECTURES` above.) Seems weird to potentially silently install nothing? Or can that not happen?

jordancarlin:
> Should we make it so that the default targets at least are required? (See `DEFAULT_ARCHITECTURES` above.) Seems weird to potentially silently install nothing? Or can that not happen?

What is someone is only interested in building/installing an rv32f or rvfi simulator for example?

Timmmm:
Yeah fair point... I just wanted to avoid the confusing if you just `make install` it installs nothing and claims success.

But I don't have strong feelings and this won't be a problem when we don't have to build multiple times for different architectures.

Timmmm (approved):

jrtc27:
Normally one would specify at configure time what targets to build and install, not proceed to build some subset and then install whatever happens to have been built

radimkrcmar:
`cmake --install` will install whatever is built.  `make install` and `ninja install` rebuild the DEFAULT_ARCHITECTURES first but also install whatever else is built.

We could do another simple change to prevent installation of the extra built targets and also fail `cmake --install` if the DEFAULT_ARCHITECTURES haven't been built:
```
diff --git a/c_emulator/CMakeLists.txt b/c_emulator/CMakeLists.txt
index 4cc78e02fae9..cee68e65068b 100644
--- a/c_emulator/CMakeLists.txt
+++ b/c_emulator/CMakeLists.txt
@@ -61,7 +61,9 @@ foreach (xlen IN ITEMS 32 64)
             #     -Wno-unused-parameter
             # )
 
-            install(TARGETS riscv_sim_${arch} OPTIONAL)
+            if (arch IN_LIST DEFAULT_ARCHITECTURES)
+                install(TARGETS riscv_sim_${arch})
+            endif()
         endforeach()
     endforeach()
 endforeach()
```
I chose the one in this PR, because it was closer to the behavior that has been there before.

The hunk above would also need to pick a better name for DEFAULT_ARCHITECTURES and maybe we should also forbid building anything that can't be installed.

Timmmm:
Yeah I wouldn't worry about it too much since it's temporary anyway.

PR #722, opened 2025-02-08
insts_base: ignore architecture in SFENCE_VMA

radimkrcmar:
Another patch from the mstatus cleanup,  https://github.com/riscv/sail-riscv/pull/683#issuecomment-2621133329.

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -714,9 +714,9 @@ function clause execute SFENCE_VMA(rs1, rs2) = {
   let asid = if rs2 != zreg then Some(X(rs2)[asidlen - 1 .. 0]) else None();
   match cur_privilege {
     User       => { handle_illegal(); RETIRE_FAIL },
-    Supervisor => match (architecture(get_mstatus_SXL(mstatus)), mstatus[TVM]) {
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):
LGTM

arichardson (approved):

Timmmm (approved):
LGTM. Any idea why it was there in the first place?

jordancarlin:
> Any idea why it was there in the first place?

Looks like a remnant from this version from 7 years ago:
```
function clause execute SFENCE_VMA(rs1, rs2) = {
  /* TODO: handle PMP/TLB synchronization when executed in M-mode. */
  if   cur_privilege == User
  then { handle_illegal(); false }
  else match (architecture(mstatus.SXL()), mstatus.TVM()) {
    (Some(RV64), true)  => { handle_illegal(); false },
    (Some(RV64), false) => {
      let addr : option(vaddr39) = if rs1 == 0 then None() else Some(X(rs1)[38 .. 0]);
      let asid : option(asid64)  = if rs2 == 0 then None() else Some(X(rs2)[15 .. 0]);
      flushTLB(asid, addr);
      true
    },
    (_, _) => internal_error("unimplemented sfence architecture")
  }
}
```

Seemingly it only supported RV64 at the time.

PR #723, opened 2025-02-09
Implement Zimop and Zcmop Extensions

jordancarlin:
Extracted and updated the `Zimop` and `Zcmop` extensions from #377.

One of the tricky parts with this extension is that it is designed to have its instructions overridden by other extensions (like `Zicfiss`). The implementation in #377 had a function dedicated to Zimop and Zcmop in each of their files that listed which specific instructions were overridden. I'd rather avoid the need to modify this file each time a new extension is added that reuses one of its encodings, so I added new scattered `encdec_overrides` and `encdec_compressed_overrides` mappings for this purpose. These mappings are checked first, and only if there is no match does it check the normal `encdec` mappings. This allows each new extension (like `Zicfiss` and `Zicflp`) to be self-contained. It also opens the door to properly supporting hints from extensions like `Zicbop` and `Zihintpause`. While the hints don't do anything architecturally, having the correct assembly in the log would still be beneficial.

This new mapping can be split out into a separate PR, but I'm introducing it here to clarify the motivation behind it.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -79,6 +79,9 @@ foreach (xlen IN ITEMS 32 64)
                 "riscv_insts_zicbom.sail"
                 "riscv_insts_zicboz.sail"
                 "riscv_insts_zvbb.sail"
+                # Zimop and Zcmop should be at the end so they can be overridden by earlier extensions
+                "riscv_insts_zimop.sail"
+                "riscv_insts_zcmop.sail"
             )
 
             if (variant STREQUAL "rmem")
--- a/model/riscv_insts_zcmop.sail
+++ b/model/riscv_insts_zcmop.sail
@@ -1,0 +1,22 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zcmop) = true & extensionEnabled(Ext_Zca) // TODO: add configuration
+
+union clause ast = ZCMOP : (bits(3))
+
+mapping clause encdec_compressed = ZCMOP(mop)
+  <-> 0b01100 @ mop : bits(3) @ 0b100000 @ 0b01
+  when extensionEnabled(Ext_Zcmop)
+
+mapping clause assembly = ZCMOP(mop)
+  <-> "c.mop." ^ dec_bits_4(mop @ 0b1)
+
+function clause execute ZCMOP(mop) = {
+  RETIRE_SUCCESS
+}
--- a/model/riscv_insts_zimop.sail
+++ b/model/riscv_insts_zimop.sail
@@ -1,0 +1,36 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zimop) = true // TODO: add configuration
+
+union clause ast = ZIMOP_MOP_R : (bits(5), regidx, regidx)
+union clause ast = ZIMOP_MOP_RR : (bits(3), regidx, regidx, regidx)
+
+mapping clause encdec = ZIMOP_MOP_R(mop_30 @ mop_27_26 @ mop_21_20, rs1, rd)
+  <-> 0b1 @ mop_30 : bits(1) @ 0b00 @ mop_27_26 : bits(2) @ 0b0111 @ mop_21_20 : bits(2) @ encdec_reg(rs1) @ 0b100 @ encdec_reg(rd) @ 0b1110011
+  when extensionEnabled(Ext_Zimop)
+
+mapping clause encdec = ZIMOP_MOP_RR(mop_30 @ mop_27_26, rs2, rs1, rd)
+  <-> 0b1 @ mop_30 : bits(1) @ 0b00 @ mop_27_26 : bits(2) @ 0b1 @ encdec_reg(rs2) @ encdec_reg(rs1) @ 0b100 @ encdec_reg(rd) @ 0b1110011
+  when extensionEnabled(Ext_Zimop)
+
+mapping clause assembly = ZIMOP_MOP_R(mop, rs1, rd)
+  <->  "mop.r." ^ dec_bits_5(mop) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1)
+
+mapping clause assembly = ZIMOP_MOP_RR(mop, rs2, rs1, rd)
+  <-> "mop.rr." ^ dec_bits_3(mop) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
+
+function clause execute ZIMOP_MOP_R(mop, rs1, rd) = {
+  X(rd) = zeros();
+  RETIRE_SUCCESS
+}
+
+function clause execute ZIMOP_MOP_RR(mop, rs2, rs1, rd) = {
+  X(rd) = zeros();
+  RETIRE_SUCCESS
+}
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ Supported RISC-V ISA features
 - Zicntr and Zihpm extensions for counters, v2.0
 - Zicond extension for integer conditional operations, v1.0
 - Zicbom and Zicboz extensions for cache-block management (Zicbop not currently supported), v1.0
+- Zimop extension for May-Be-Operations, v1.0
 - M extension for integer multiplication and division, v2.0
 - Zmmul extension for integer multiplication only, v1.0
 - A extension for atomic instructions, v2.1
@@ -106,6 +107,7 @@ Supported RISC-V ISA features
 - Zfinx, Zdinx, and Zhinx extensions for floating-point in integer registers, v1.0
 - C extension for compressed instructions, v2.0
 - Zca, Zcf, Zcd, and Zcb extensions for code size reduction, v1.0
+- Zcmop extension for compressed May-Be-Operations, v1.0
 - B (Zba, Zbb, Zbs) and Zbc extensions for bit manipulation, v1.0
 - Zbkb, Zbkc, and Zbkx extensions for bit manipulation for cryptography, v1.0
 - Zkn (Zknd, Zkne, Zknh) and Zks (Zksed, Zksh) extensions for scalar cryptography, v1.0.1
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -17,6 +17,7 @@ $include <vector_dec.sail>
 $include <generic_equality.sail>
 $include <hex_bits.sail>
 $include <hex_bits_signed.sail>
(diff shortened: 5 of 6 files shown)


Timmmm:
Makes sense. It's a bit hard to follow the flow though... What if you have:

* `encdec_hints`: Just the mappings for hints. Doesn't chain to any other mappings.
* `encdec`: Leave this completely unmodified.
* `encdec_with_hints`:

```
mapping encdec_with_hints : bits(32) -> ast = {
    encdec_hints(s) <-> s
    encdec(s) <-> s
}
```

I think that's a lot easier to follow.

jrtc27:
Why do we need all this rather than just having multiple encdec clauses in decreasing precedence order, i.e. compiling Zicfiss before Zimop?

jordancarlin:
> Makes sense. It's a bit hard to follow the flow though... What if you have:
> 
> * `encdec_hints`: Just the mappings for hints. Doesn't chain to any other mappings.
> * `encdec`: Leave this completely unmodified.
> * `encdec_with_hints`:
> 
> ```
> mapping encdec_with_hints : bits(32) -> ast = {
>     encdec_hints(s) <-> s
>     encdec(s) <-> s
> }
> ```
> 
> I think that's a lot easier to follow.

Good suggestion. That is definitely easier to follow. I think I still prefer `encdec_overrides` as the name since things like Zicfiss which reuse the Zimop encoding space are explicitly described in the ISA manual and NOT hints.

jordancarlin:
> Why do we need all this rather than just having multiple encdec clauses in decreasing precedence order, i.e. compiling Zicfiss before Zimop?

See previous related discussion here: https://github.com/riscv/sail-riscv/pull/538#discussion_r1750281734. The general sense is having different results for the model based on the compilation order seems unnecessarily fragile. Especially as we (hopefully) move to the new module system that makes it easier than ever to include or exclude various files from compilation.

jordancarlin:
> > Makes sense. It's a bit hard to follow the flow though... What if you have:
> > 
> > * `encdec_hints`: Just the mappings for hints. Doesn't chain to any other mappings.
> > * `encdec`: Leave this completely unmodified.
> > * `encdec_with_hints`:
> > 
> > ```
> > mapping encdec_with_hints : bits(32) -> ast = {
> >     encdec_hints(s) <-> s
> >     encdec(s) <-> s
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think that's a lot easier to follow.
> 
> Good suggestion. That is definitely easier to follow. I think I still prefer `encdec_overrides` as the name since things like Zicfiss which reuse the Zimop encoding space are explicitly described in the ISA manual and NOT hints.

Hmm. That approach may actually not work after all. Somewhat surprisingly, the error messages come about in the C compilation after sail itself is done.
```
[4/6] Linking C executable c_emulator/riscv_sim_rv32d
FAILED: c_emulator/riscv_sim_rv32d 
: && /opt/rh/gcc-toolset-13/root/usr/bin/cc -O2 -g -DNDEBUG -Wl,--no-undefined c_emulator/CMakeFiles/riscv_sim_rv32d.dir/__/riscv_model_rv32d.c.o c_emulator/CMakeFiles/riscv_sim_rv32d.dir/riscv_platform.c.o c_emulator/CMakeFiles/riscv_sim_rv32d.dir/riscv_platform_impl.c.o c_emulator/CMakeFiles/riscv_sim_rv32d.dir/riscv_prelude.c.o c_emulator/CMakeFiles/riscv_sim_rv32d.dir/riscv_sim.c.o c_emulator/CMakeFiles/riscv_sim_rv32d.dir/riscv_softfloat.c.o -o c_emulator/riscv_sim_rv32d  dependencies/softfloat/libsoftfloat.a  sail_runtime/libsail_runtime.a  /usr/lib64/libgmp.so  /usr/lib64/libz.so && :
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: c_emulator/CMakeFiles/riscv_sim_rv32d.dir/__/riscv_model_rv32d.c.o: in function `zencdec_with_overrides_forwards':
/home/jcarlin/riscv-projects/sail-riscv/build/riscv_model_rv32d.c:197075: undefined reference to `zencdec_overrides_backwards_matches'
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: /home/jcarlin/riscv-projects/sail-riscv/build/riscv_model_rv32d.c:197082: undefined reference to `zencdec_overrides_backwards'
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: c_emulator/CMakeFiles/riscv_sim_rv32d.dir/__/riscv_model_rv32d.c.o: in function `zencdec_with_overrides_forwards_infallible':
/home/jcarlin/riscv-projects/sail-riscv/build/riscv_model_rv32d.c:198005: undefined reference to `zencdec_overrides_backwards_matches'
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: /home/jcarlin/riscv-projects/sail-riscv/build/riscv_model_rv32d.c:198012: undefined reference to `zencdec_overrides_backwards'
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: c_emulator/CMakeFiles/riscv_sim_rv32d.dir/__/riscv_model_rv32d.c.o: in function `zencdec_compressed_with_overrides_forwards':
/home/jcarlin/riscv-projects/sail-riscv/build/riscv_model_rv32d.c:208184: undefined reference to `zencdec_compressed_overrides_backwards_matches'
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: /home/jcarlin/riscv-projects/sail-riscv/build/riscv_model_rv32d.c:208191: undefined reference to `zencdec_compressed_overrides_backwards'
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: c_emulator/CMakeFiles/riscv_sim_rv32d.dir/__/riscv_model_rv32d.c.o: in function `zencdec_compressed_with_overrides_forwards_infallible':
/home/jcarlin/riscv-projects/sail-riscv/build/riscv_model_rv32d.c:209114: undefined reference to `zencdec_compressed_overrides_backwards_matches'
/opt/rh/gcc-toolset-13/root/usr/libexec/gcc/x86_64-redhat-linux/13/ld: /home/jcarlin/riscv-projects/sail-riscv/build/riscv_model_rv32d.c:209121: undefined reference to `zencdec_compressed_overrides_backwards'
collect2: error: ld returned 1 exit status
[5/6] Building C object c_emulator/CMakeFiles/riscv_sim_rv64d.dir/__/riscv_model_rv64d.c.o
```

Seems like it's not able to come up with a backwards function for the mapping. I'll see if I can manually define one that will work, but it may end up being cleaner to have the tiered mappings that are in the current version.

Timmmm:
Hmm could be a Sail compiler bug? Can you make a small reproducible example?

Also we may be able to work around it with `--c-preserve zencdec... all of those missing functions`?

Alasdair:
We shouldn't produce any broken C code for Sail code that type-checks, so certainly a bug. Looks like the tree-shaking that we do to prune unused definitions is removing things it shouldn't in this case, I'll try to create a smaller test case. I recently introduced an optimisation that introduces `_infalliable` versions of mappings for certain optimisations and it might be that they aren't being tracked correctly.

The original idea was to use the order of files to control this. The module system I've been working on makes this more explicit, as you can write explicitly `after X` or `before Y` in a module `Z` to control the ordering of scattered definitions.

jordancarlin:
Yeah. Seems like if the scattered mapping is empty (which the overrides mappings are in this PR are), then Sail optimizes them away but still leaves the call to them in the other mapping that calls the (now pruned) scattered mapping. I can create a smaller test case if that would be helpful.

kourzanov:
Isn't this overriding better solved by injecting overrides into the generated C zdecode/zexecute?

jordancarlin:
Per the TGMM today, we're going to skip the overrides and rely on the upcoming module system that adds support for explicitly declaring module precedence. This is a good balance between the slightly unreliable pure file ordering and the slightly convoluted nested mappings.

model/CMakeLists.txt
```diff
@@ -75,6 +76,7 @@ foreach (xlen IN ITEMS 32 64)
                 "riscv_insts_zbkb.sail"
                 "riscv_insts_zbkx.sail"
                 "riscv_insts_zicond.sail"
+                "riscv_insts_zimop.sail"
                 ${vext_srcs}
                 "riscv_insts_zicbom.sail"
                 "riscv_insts_zicboz.sail"
```
arichardson, line 82: Shouldn't we put these new files right at the end so they have lowest precedence?

model/riscv_insts_zimop.sail
```diff
@@ -0,0 +1,80 @@
...
+  0b010 <-> "mop.rr.2",
+  0b011 <-> "mop.rr.3",
+  0b100 <-> "mop.rr.4",
+  0b101 <-> "mop.rr.5",
+  0b110 <-> "mop.rr.6",
+  0b111 <-> "mop.rr.7",
+}
+
+mapping clause assembly = ZIMOP_MOP_R(mop, rs1, rd)
+  <->  zimop_r_mnemonic(mop) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1)
```
arichardson, line 67:
```suggestion
mapping clause assembly = ZIMOP_MOP_R(mop, rs1, rd)
  <->  "mop.r." ^ ^ hex_bits_5(mop) ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1)
```

Can we do the same thing as the hint instructions here?

arichardson: And same with all the other mappings?

arichardson (commented):
Not looked at the spec but seems reasonable overall

model/riscv_insts_zimop.sail
jordancarlin: I think it was initially done this way for consistency because the `c.mop` instruction mnemonics don't correspond to the actual bit values, but for the Zimop mappings that should work.

jordancarlin: Done

model/riscv_insts_zcmop.sail
```diff
@@ -0,0 +1,32 @@
...
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zcmop) = true & extensionEnabled(Ext_Zca) // TODO: add configuration
+
+union clause ast = ZCMOP : (bits(3))
+
+mapping clause encdec_compressed = ZCMOP(mop) if extensionEnabled(Ext_Zcmop)
+  <-> 0b01100 @ mop : bits(3) @ 0b100000 @ 0b01 if extensionEnabled(Ext_Zcmop)
+
+mapping zcmop_mnemonic : bits(3) <-> string = {
```
arichardson, line 16: Isn't this one just `"c.mop." ^ hex_bits(mop @ 0b1)`? Or well decimal not hex.

jordancarlin: Yes. That would work. Also, not sure how I messed that up and did hex_bits in Zimop. Should be dec_bits for that one as well. Looks like that hasn't been integrated into sail yet: https://github.com/rems-project/sail/pull/845/

jordancarlin: Updated

jordancarlin:
This either needs the `dec_bits.sail` library file to be included locally for now or wait until the next sail release. Unless there is a high degree of urgency to get it in, I'd vote for waiting since there should be a new release soon.

jordancarlin:
This should be good to go now that Sail 0.19 is released.

Timmmm:
I thought we were going to wait for #572 ?

arichardson (approved):

pmundkur (approved):

nadime15 (approved):

PR #724, opened 2025-02-09
Fix building the model when the sail binary is not in $PATH

arichardson:
In my build sail is installed to a custom prefix that I pass to CMake via `CMAKE_PREFIX_PATH` instead of by overriding `$PATH`. Fix the one place in the build where we were using a bare sail instead of `${SAIL_BIN}`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ find_program(SAIL_BIN "sail")
 if (NOT SAIL_BIN)
     message(FATAL_ERROR "Sail not found. See README.md for installation instructions.")
 endif()
+message(STATUS "Found sail: ${SAIL_BIN}")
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ find_program(SAIL_BIN "sail")
 if (NOT SAIL_BIN)
     message(FATAL_ERROR "Sail not found. See README.md for installation instructions.")
 endif()
+message(STATUS "Found sail: ${SAIL_BIN}")
```
arichardson, line 67: Strictly speaking this is unrelated, but I found it helpful considering we print the path to GMP and libz as well.

jordancarlin (approved):

Timmmm (approved):
Whoops!

PR #725, opened 2025-02-10
Add Svadu/Svade extension

trdthg:
Please check: https://riscv-specs.timhutt.co.uk/spec/riscv-isa-release-b5aa38c-2025-01-17/riscv-privileged.html#sec:svadu

- Add `menvcfg.ADUE` 

	> If the Svadu extension is implemented, the menvcfg.ADUE field is writable. 

- Check A/D when Svade is enabled
	
	> Two schemes to manage the A and D bits are defined:
	>	- The Svade extension: when a virtual page is accessed and the A bit is clear, or is written and the D bit is clear, a page-fault exception is raised.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ Supported RISC-V ISA features
 - Svinval extension for fine-grained address-translation cache invalidation, v1.0
 - Sv32, Sv39, Sv48 and Sv57 page-based virtual-memory systems
 - Physical Memory Protection (PMP)
+- Svadu and Svade Extension for Hardware Updating of A/D Bits, v1.0
 
 **For a list of unsupported extensions and features, see the [Extension Roadmap](https://github.com/riscv/sail-riscv/wiki/Extension-Roadmap).**
 
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -112,6 +112,9 @@ enum clause extension = Ext_Zvkb
 enum clause extension = Ext_Sscofpmf
 // Supervisor-mode Timer Interrupts
 enum clause extension = Ext_Sstc
+// Hardware Updating of A/D Bits
+enum clause extension = Ext_Svadu
+enum clause extension = Ext_Svade
 // Fine-Grained Address-Translation Cache Invalidation
 enum clause extension = Ext_Svinval
 // NAPOT Translation Contiguity
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -127,6 +127,17 @@ val sys_enable_sstc = pure "sys_enable_sstc" : unit -> bool
 // Supervisor timecmp
 function clause extensionEnabled(Ext_Sstc) = sys_enable_sstc()
 
+/* whether hardware updating of A/D bits was enabled at boot */
+val plat_enable_dirty_update = pure {interpreter: "Platform.enable_dirty_update",
+                                     c: "plat_enable_dirty_update",
+                                     lem: "plat_enable_dirty_update"} : unit -> bool
+
+/* whether Svadu was enabled at boot */
+val sys_enable_svadu = pure "sys_enable_svadu" : unit -> bool
+
+/* whether Svadu was enabled at boot */
+val sys_enable_svade = pure "sys_enable_svade" : unit -> bool
+
 /* This function allows an extension to veto a write to Misa
    if it would violate an alignment restriction on
    unsetting C. If it returns true the write will have no effect. */
@@ -330,8 +341,10 @@ bitfield MEnvcfg : bits(64) = {
   STCE   : 63,
   // Page Based Memory Types Extension
   PBMTE  : 62,
+  // Svadu extension
+  ADUE   : 61,
   // Reserved WPRI bits.
-  wpri_1 : 61 .. 8,
+  wpri_1 : 60 .. 8,
   // Cache Block Zero instruction Enable
   CBZE   : 7,
   // Cache Block Clean and Flush instruction Enable
@@ -365,6 +378,7 @@ function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     CBCFE = if extensionEnabled(Ext_Zicbom) then v[CBCFE] else 0b0,
     CBIE = if extensionEnabled(Ext_Zicbom) then (if v[CBIE] != 0b10 then v[CBIE] else 0b00) else 0b00,
     STCE = if extensionEnabled(Ext_Sstc) then v[STCE] else 0b0,
+    ADUE = if sys_enable_svadu() then v[ADUE] else 0b0,
     // Other extensions are not implemented yet so all other fields are read only zero.
   ]
 }
@@ -381,7 +395,10 @@ function legalize_senvcfg(o : SEnvcfg, v : xlenbits) -> SEnvcfg = {
 }
 
 // Initialised to legal values in case some bits are hard-coded to 1.
-register menvcfg : MEnvcfg = legalize_menvcfg(Mk_MEnvcfg(zeros()), zeros())
+register menvcfg : MEnvcfg = legalize_menvcfg(
+  Mk_MEnvcfg(zeros()),
+  [Mk_MEnvcfg(zeros()) with ADUE = bool_to_bits(sys_enable_svadu() & plat_enable_dirty_update())].bits
+)
 register senvcfg : SEnvcfg = legalize_senvcfg(Mk_SEnvcfg(zeros()), zeros())
 
 mapping clause csr_name_map = 0x30A  <-> "menvcfg"
@@ -400,6 +417,11 @@ function clause write_CSR((0x30A, value) if xlen == 32) = { menvcfg = legalize_m
 function clause write_CSR((0x30A, value) if xlen == 64) = { menvcfg = legalize_menvcfg(menvcfg, value); menvcfg.bits }
 function clause write_CSR((0x31A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, value @ menvcfg.bits[31 .. 0]); menvcfg.bits[63 .. 32] }
 function clause write_CSR(0x10A, value) = { senvcfg = legalize_senvcfg(senvcfg, zero_extend(value)); senvcfg.bits[xlen - 1 .. 0] }
+
+/* whether hardware updating of A/D bits is enabled */
+function clause extensionEnabled(Ext_Svadu) = sys_enable_svadu()
+/* whether Svade takes effect */
+function clause extensionEnabled(Ext_Svade) = sys_enable_svade() & menvcfg[ADUE] == 0b0
 
 // Return whether or not FIOM is currently active, based on the current
 // privilege and the menvcfg/senvcfg settings. This means that I/O fences
--- a/model/riscv_vmem.sail
+++ b/model/riscv_vmem.sail
@@ -239,10 +239,11 @@ function translate_TLB_hit forall 'v, is_sv_mode('v) . (
       match update_PTE_Bits(pte, ac) {
         None()     => TR_Address(tlb_get_ppn(sv_width, ent, vpn), ext_ptw),
         Some(pte') =>
-          if not(plat_enable_dirty_update()) then
-            // pte needs dirty/accessed update but that is not enabled
-            TR_Failure(PTW_PTE_Update(), ext_ptw)
-          else {
+          if not(extensionEnabled(Ext_Svadu)) & not(plat_enable_dirty_update()) then {
+              TR_Failure(PTW_PTE_Update(), ext_ptw)
+          } else if extensionEnabled(Ext_Svadu) & extensionEnabled(Ext_Svade) then {
+              TR_Failure(PTW_Access(), ext_ptw)
+          } else {
             // Writeback the PTE (which has new A/D bits)
             write_TLB(tlb_index, tlb_set_pte(ent, pte'));
             match write_pte(ent.pteAddr, pte_width, pte') {
@@ -276,7 +277,7 @@ function translate_TLB_miss forall 'v, is_sv_mode('v) . (
   match ptw_result {
     PTW_Failure(f, ext_ptw) => TR_Failure(f, ext_ptw),
     PTW_Success(struct {ppn, pte, pteAddr, level, global}, ext_ptw) => {
-      let ext_pte   = ext_bits_of_PTE(pte);
+      let ext_pte = ext_bits_of_PTE(pte);
       // Without TLBs, this 'match' expression can be replaced simply
       // by: 'TR_Address(ppn, ext_ptw)'    (see TLB NOTE above)
       match update_PTE_Bits(pte, ac) {
@@ -285,11 +286,11 @@ function translate_TLB_miss forall 'v, is_sv_mode('v) . (
           TR_Address(ppn, ext_ptw)
         },
         Some(pte) =>
-          // See riscv_platform.sail
-          if not(plat_enable_dirty_update()) then
-            // pte needs dirty/accessed update but that is not enabled
-            TR_Failure(PTW_PTE_Update(), ext_ptw)
-          else {
+          if not(extensionEnabled(Ext_Svadu)) & not(plat_enable_dirty_update()) then {
+              TR_Failure(PTW_PTE_Update(), ext_ptw)
+          } else if extensionEnabled(Ext_Svadu) & extensionEnabled(Ext_Svade) then {
+              TR_Failure(PTW_Access(), ext_ptw)
+          } else {
             // Writeback the PTE (which has new A/D bits)
             match write_pte(pteAddr, pte_width, pte) {
               Ok(_) => {
--- a/c_emulator/riscv_platform.cpp
+++ b/c_emulator/riscv_platform.cpp
@@ -78,6 +78,16 @@ bool sys_enable_sstc(unit)
   return rv_enable_sstc;
 }
 
+bool sys_enable_svadu(unit)
+{
+  return rv_enable_svadu;
+}
+
+bool sys_enable_svade(unit)
+{
+  return rv_enable_svadu;
+}
+
 uint64_t sys_pmp_count(unit)
 {
   return rv_pmp_count;
--- a/c_emulator/riscv_platform.h
+++ b/c_emulator/riscv_platform.h
@@ -18,6 +18,8 @@ bool sys_enable_zicbom(unit);
 bool sys_enable_zicboz(unit);
 bool sys_enable_zvkb(unit);
 bool sys_enable_sstc(unit);
+bool sys_enable_svadu(unit u);
+bool sys_enable_svade(unit u);
 
 uint64_t sys_pmp_count(unit);
 uint64_t sys_pmp_grain(unit);
--- a/c_emulator/riscv_platform_impl.cpp
+++ b/c_emulator/riscv_platform_impl.cpp
@@ -21,6 +21,7 @@ bool rv_enable_zicbom = false;
 bool rv_enable_zicboz = false;
 bool rv_enable_zvkb = false;
 bool rv_enable_sstc = false;
+bool rv_enable_svadu = false;
 
 bool rv_enable_dirty_update = false;
 bool rv_enable_misaligned = false;
--- a/c_emulator/riscv_platform_impl.h
+++ b/c_emulator/riscv_platform_impl.h
@@ -25,6 +25,7 @@ extern bool rv_enable_zicbom;
 extern bool rv_enable_zicboz;
 extern bool rv_enable_zvkb;
 extern bool rv_enable_sstc;
+extern bool rv_enable_svadu;
 extern bool rv_enable_writable_misa;
 extern bool rv_enable_dirty_update;
 extern bool rv_enable_misaligned;
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -53,6 +53,7 @@ enum {
   OPT_ENABLE_ZICBOZ,
   OPT_ENABLE_ZVKB,
   OPT_ENABLE_SSTC,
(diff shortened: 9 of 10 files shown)


c_emulator/riscv_sim.c
```diff
@@ -431,7 +432,11 @@ static int process_args(int argc, char **argv)
       break;
     case OPT_ENABLE_SSTC:
       fprintf(stderr, "enabling Sstc extension.\n");
-      rv_enable_sstc = true;
+      rv_enable_svadu = true;
```
arichardson, line 435: Sstc should remain unchanged.

model/riscv_sys_regs.sail
```diff
@@ -121,9 +121,14 @@ val sys_enable_zicboz = pure "sys_enable_zicboz" : unit -> bool
 // Is the Sstc stimecmp extension supported.
 val sys_enable_sstc = pure "sys_enable_sstc" : unit -> bool
 
+val sys_enable_svadu = pure "sys_enable_svadu" : unit -> bool
+
 // Supervisor timecmp
 function clause extensionEnabled(Ext_Sstc) = sys_enable_sstc()
 
+// Supervisor timecmp
```
arichardson, line 129: Stale description

model/riscv_sys_regs.sail
```diff
@@ -378,6 +385,7 @@ function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     CBCFE = if extensionEnabled(Ext_Zicbom) then v[CBCFE] else 0b0,
     CBIE = if extensionEnabled(Ext_Zicbom) then (if v[CBIE] != 0b10 then v[CBIE] else 0b00) else 0b00,
     STCE = if extensionEnabled(Ext_Sstc) then v[STCE] else 0b0,
+    ADUE = if extensionEnabled(Ext_Svade) then v[ADUE] else 0b0,
```
arichardson, line 388: Svade mandates exceptions, I believe you want Svadu here

model/riscv_sys_regs.sail
```diff
@@ -121,9 +121,14 @@ val sys_enable_zicboz = pure "sys_enable_zicboz" : unit -> bool
 // Is the Sstc stimecmp extension supported.
 val sys_enable_sstc = pure "sys_enable_sstc" : unit -> bool
 
+val sys_enable_svadu = pure "sys_enable_svadu" : unit -> bool
+
 // Supervisor timecmp
 function clause extensionEnabled(Ext_Sstc) = sys_enable_sstc()
 
+// Supervisor timecmp
+function clause extensionEnabled(Ext_Svade) = sys_enable_svadu()
```
arichardson, line 130: Also this should be Ext_svadu

model/riscv_extensions.sail
```diff
@@ -105,6 +105,8 @@ enum clause extension = Ext_Zhinx
 
 // Supervisor-mode Timer Interrupts
 enum clause extension = Ext_Sstc
+// Hardware Updating of A/D Bits
+enum clause extension = Ext_Svade
```
arichardson, line 109:
```suggestion
enum clause extension = Ext_Svadu
```

model/riscv_vmem.sail
```diff
@@ -240,8 +240,11 @@ function translate_TLB_hit forall 'v, is_sv_mode('v) . (
         None()     => TR_Address(tlb_get_ppn(sv_width, ent, vpn), ext_ptw),
         Some(pte') =>
           if not(plat_enable_dirty_update()) then
-            // pte needs dirty/accessed update but that is not enabled
-            TR_Failure(PTW_PTE_Update(), ext_ptw)
+            if extensionEnabled(Ext_Svade) & check_PTE_AD_invalid(pte, ac) then {
```
arichardson, line 243: Shouldn't this be checking the ADUE bit instead?

model/riscv_vmem.sail
```diff
@@ -240,8 +240,11 @@ function translate_TLB_hit forall 'v, is_sv_mode('v) . (
         None()     => TR_Address(tlb_get_ppn(sv_width, ent, vpn), ext_ptw),
         Some(pte') =>
           if not(plat_enable_dirty_update()) then
```
arichardson, line 242: Isn't plat_enable_dirty_update now the same as sys_enable_svadu?

trdthg: Yes, you are right, I previously confused svadu/svade/plat_enable_dirty_update

So I added `extensionEnabled(Ext_Svade) = menvcfg[ADUE] == 0b0`

> hence Svade is implemented when ADUE=0.
> When hardware updating of A/D bits is disabled, the Svade extension,  ... , instead takes effect

Should we should replace `plat_enable_dirty_update` with `extensionEnabled(Ext_Svadu)` now?

arichardson: If I understand correctly then `plat_enable_dirty_update() == not(extensionEnabled(Ext_Svade)`. Ext_Svadu just makes the bit configurable.

Maybe plat_enable_dirty_update() should be used for the reset value of menvcfg[ADUE]

model/riscv_sys_regs.sail
```diff
@@ -378,6 +385,7 @@ function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     CBCFE = if extensionEnabled(Ext_Zicbom) then v[CBCFE] else 0b0,
     CBIE = if extensionEnabled(Ext_Zicbom) then (if v[CBIE] != 0b10 then v[CBIE] else 0b00) else 0b00,
     STCE = if extensionEnabled(Ext_Sstc) then v[STCE] else 0b0,
+    ADUE = if extensionEnabled(Ext_Svadu) then v[ADUE] else 0b0,
```
jordancarlin, line 388: Based on the spec, it seems to me like this should be dependent on `sys_enable_svadu` and then `extensionEnabled(Ext_Svadu)` should be based on the ADUE bit being 1/0.

model/riscv_sys_regs.sail
```diff
@@ -121,9 +121,14 @@ val sys_enable_zicboz = pure "sys_enable_zicboz" : unit -> bool
...
 val sys_enable_sstc = pure "sys_enable_sstc" : unit -> bool
 
-// Supervisor timecmp
+val sys_enable_svadu = pure "sys_enable_svadu" : unit -> bool
+
+/* whether misa.b was enabled at boot */
 function clause extensionEnabled(Ext_Sstc) = sys_enable_sstc()
 
+/* whether menvcfg.ADUE, (TODO) henvcfg.ADUE was enabled at boot */
+function clause extensionEnabled(Ext_Svadu) = sys_enable_svadu()
```
jordancarlin, line 130: Put this right after `sys_enable_svadu`. And I think the comment is wrong.

model/riscv_sys_regs.sail
```diff
@@ -121,9 +121,14 @@ val sys_enable_zicboz = pure "sys_enable_zicboz" : unit -> bool
 // Is the Sstc stimecmp extension supported.
 val sys_enable_sstc = pure "sys_enable_sstc" : unit -> bool
 
-// Supervisor timecmp
```
jordancarlin, line 124: Why is this description being removed from Sstc?

model/riscv_vmem.sail
jordancarlin: I think there are several legal possibilities:
1. Svade is supported and Svadu is not. This should always raise a page-fault exception if the A/D bits need to be updated.
2. Neither the Svadu nor Svade extensions are implemented. There should always be hardware updates of the A/D bits. This seems to be what `plat_enable_dirty_update()` implies right now. The important thing to note with this option is that `menvcfg[ADUE]` will be 0, but there will be hardware updating of A/D bits.
3. Svadu is implemented, which requires Svade to also be implemented. There are hardware updates when `menvcfg[ADUE]` is 1 and page faults when `menvcfg[ADUE]` is 0.

Right now it seems like `plat_enable_dirty_update` controls if we are using option 1 (when disabled) or option 2 (when enabled). One potential path forward would be to remove `plat_enable_dirty_update` and have both `sys_enable_svadu` and `sys_enable_svade` options. 
- If neither is set, then we go to option 2 (the equivalent of setting `plat_enable_dirty_update` right now). 
- If `sys_enable_svade` is set and  `sys_enable_svadu` is not set, then `extensionEnabled(Ext_Svade)` is hardwired to true and we always throw page faults (the current behavior when `plat_enable_dirty_update` is not set).
- If `sys_enable_svadu` is set then `menvcfg[ADUE]` becomes writable and when it is a 1 we do hardware updates and when it is a zero we follow the implementation for Svade. `sys_enable_svadu` would imply `sys_enable_svade`.

I think this boils down to `extensionEnabled(Ext_Svadu) = menvcfg[ADUE] == 1` and `extensionEnabled(Ext_Svade) = menvcfg[ADUE] == 0 & (sys_enable_svade | sys_enable_svadu)`. We would then update the A/D bits if `extensionEnabled(Ext_Svadu) | not(sys_enable_svade | sys_enable_svadu)` and throw a page fault otherwise (which should correspond to `extensionEnabled(Ext_Svade)`.

Quite the convoluted set of options...

model/riscv_vmem_pte.sail
```diff
@@ -161,3 +161,24 @@ function update_PTE_Bits forall 'pte_size, 'pte_size in {32, 64} . (
     None()
   }
 }
+
+// PRIVATE: Check A/D bits status for Svadu extension
+// "when a virtual page is accessed and the A bit is clear, or is written
+//  and the D bit is clear, a page-fault exception is raised."
+function check_PTE_AD_invalid forall 'pte_size, 'pte_size in {32, 64} . (
```
jordancarlin, line 168: This seems like a misleadingly named function. Nothing about this is invalid from what I can tell. The A/D bits just need to be updated.

jordancarlin (changes requested):

jordancarlin:
This will also conflict with #714, so we need to figure out how we want these two to work together.

model/riscv_vmem_pte.sail
trdthg: I deleted this function; the `update_PTE_Bits` above has already done the same thing

model/riscv_sys_regs.sail
```diff
@@ -121,9 +121,14 @@ val sys_enable_zicboz = pure "sys_enable_zicboz" : unit -> bool
...
+val sys_enable_svadu = pure "sys_enable_svadu" : unit -> bool
+
+/* whether misa.b was enabled at boot */
```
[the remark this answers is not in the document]
trdthg: done

model/riscv_sys_regs.sail
trdthg: done

model/riscv_sys_regs.sail
trdthg: put after menvcfg def

model/riscv_vmem.sail
trdthg: Update
- Added `-enable_svade` option
- Modified various constraints
- Retained `plat_enable_dirty_update` as init value, I think it makes sense, the `init_spike` on the C side (before init_sail) also depends on this condition

model/riscv_sys_regs.sail
trdthg: done

model/riscv_sys_regs.sail
trdthg: done

model/riscv_sys_regs.sail
trdthg: done

model/riscv_extensions.sail
trdthg: done

model/riscv_sys_regs.sail
```diff
@@ -378,6 +389,7 @@ function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     CBCFE = if extensionEnabled(Ext_Zicbom) then v[CBCFE] else 0b0,
     CBIE = if extensionEnabled(Ext_Zicbom) then (if v[CBIE] != 0b10 then v[CBIE] else 0b00) else 0b00,
     STCE = if extensionEnabled(Ext_Sstc) then v[STCE] else 0b0,
+    ADUE = if sys_enable_svadu() then v[ADUE] else 0b0,
```
arichardson, line 392: This means it can only be set to 1 if it is already 1. But Svadu should make in configurable.

model/riscv_vmem.sail
```diff
@@ -239,10 +239,7 @@ function translate_TLB_hit forall 'v, is_sv_mode('v) . (
       match update_PTE_Bits(pte, ac) {
         None()     => TR_Address(tlb_get_ppn(sv_width, ent, vpn), ext_ptw),
         Some(pte') =>
-          if not(plat_enable_dirty_update()) then
-            // pte needs dirty/accessed update but that is not enabled
-            TR_Failure(PTW_PTE_Update(), ext_ptw)
-          else {
+          if extensionEnabled(Ext_Svadu) | not(sys_enable_svade() | sys_enable_svadu()) then {
```
arichardson, line 242:
```suggestion
          if not(plat_enable_dirty_update()) | menvcfg[ADUE] == 0b0 then {
```
Shouldn't this be equivalent to the old logic and also handle the new case of "Svadu present but updates disabled"?

model/riscv_sys_regs.sail
```diff
@@ -124,6 +124,15 @@ val sys_enable_sstc = pure "sys_enable_sstc" : unit -> bool
 // Supervisor timecmp
 function clause extensionEnabled(Ext_Sstc) = sys_enable_sstc()
 
+/* whether Svadu/Svade was enabled at boot */
+val sys_enable_svadu = pure "sys_enable_svadu" : unit -> bool
+val sys_enable_svade = pure "sys_enable_svade" : unit -> bool
```
arichardson, line 129: Do we need this flag? Isn't sys_enable_svade equivalent to `not(plat_enable_dirty_update()) | sys_enable_svadu()`?

model/riscv_sys_regs.sail
jordancarlin: I think it works the way it is right now because of the distinction between `sys_enable_svadu()` and `extensionEnabled(Ext_Svadu)`. The former will be true if the configuration supports it (so the bit is writable) and the latter will be true when the extension is enabled (when this bit is set to 1).

model/riscv_sys_regs.sail
jordancarlin: I'd be in favor of having svade and svadu flags and getting rid of the plat_enable_dirty_update flag. Having neither of the new flags should be equivalent to the old flag (which does mean a change in default behavior, but a configuration option available to get to whatever a particular implementation does).

arichardson: svadu implies svade is available (but not necessarily the default behaviour), so I think having a flag for svadu is sufficient (and this makes ADUE writable). I think it would be more consistent to have plat_enable_dirty_update + enable SVAdu

| enable_updates | enable_svadu  |   |
| -------------  | ------------- |---|
|  0             | 0  | always exceptions, hardwired ADUE=0  |
|  0             | 1  | exceptions by default, but can set to ADUE=1 for updates  |
|  1             | 0  | always updates, even though hardwired ADUE=0  |
|  1             | 1  | updates by default, but can set to ADUE=0 for exceptions |

I think this is easier to reason about than the svade+svadu case:

| enable_svade | enable_svadu  |   |
| -------------  | ------------- |---|
|  0             | 0  | always updates, even though hardwired ADUE=0  |
|  0             | 1  | illegal? Or do we treat this as the "updates by default, but can set to ADUE=0 for exceptions" case?  |
|  1             | 0  | always exceptions, hardwired ADUE=0  |
|  1             | 1  | ADUE bit exists, but is it hardwired to zero? Or is this the exceptions by default case? |

As the spec says:
> The Svadu extension adds support and CSR controls for hardware updating of PTE A/D bits.
>
>If the Svadu extension is implemented, the menvcfg.ADUE field is writable. If the hypervisor extension is additionally implemented, the henvcfg.ADUE field is also writable. See [sec:menvcfg] and [sec:henvcfg] for the definitions of those fields.
>
>Addressing and Memory Protection defines the semantics of hardware updating of A/D bits. When hardware updating of A/D bits is disabled, the Svade extension, which mandates exceptions when A/D bits need be set, instead takes effect. The Svade extension is also defined in Addressing and Memory Protection.

arichardson: Alternatively, it might be possible to just hardwire ADUE=1 for enable_updates+!enable_svadu? Havne't checked if we could do that? Of course it won't match older implementations that had dirty updates, but not ADUE bit, so maybe not a good idea.

jordancarlin: For additional context, here's what the linux device tree bindings say:

>Both Svade and Svadu extensions control the hardware behavior when
            the PTE A/D bits need to be set. The default behavior for the four
            possible combinations of these extensions in the device tree are:
            1) Neither Svade nor Svadu present in DT => It is technically
               unknown whether the platform uses Svade or Svadu. Supervisor
               software should be prepared to handle either hardware updating
               of the PTE A/D bits or page faults when they need updated.
            2) Only Svade present in DT => Supervisor must assume Svade to be
               always enabled.
            3) Only Svadu present in DT => Supervisor must assume Svadu to be
               always enabled.
            4) Both Svade and Svadu present in DT => Supervisor must assume
               Svadu turned-off at boot time. To use Svadu, supervisor must
               explicitly enable it using the SBI FWFT extension.

model/riscv_sys_regs.sail
trdthg: > But Svadu should make in configurable.

@arichardson It is configurable, I'm currently using `plat_enable_dirty_update()` as the initial value menvcfg during initialization
```
register menvcfg : MEnvcfg = legalize_menvcfg(Mk_MEnvcfg(zeros()), 
	[Mk_MEnvcfg(zeros()) with 
		ADUE = if (sys_enable_svadu() & plat_enable_dirty_update()) then 0b1 else 0b0
	].bits)
```

c_emulator/riscv_sim.c
```diff
@@ -537,7 +547,7 @@ void init_spike(const char *f, uint64_t entry, uint64_t ram_size)
   bool mismatch = false;
   const char *isa = is_32bit_model() ? RV32ISA : RV64ISA;
   s = tv_init(isa, ram_size, 1);
-  if (tv_is_dirty_enabled(s) != rv_enable_dirty_update) {
+  if (tv_is_dirty_enabled(s) != (rv_enable_dirty_update && rv_enable_svadu)) {
```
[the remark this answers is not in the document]
trdthg: I'll do it in another PR later

model/riscv_sys_regs.sail
trdthg: behavior when svadu not implemented https://github.com/riscv/riscv-isa-manual/issues/1846

README.md
```diff
@@ -121,6 +121,7 @@ Supported RISC-V ISA features
 - Svinval extension for fine-grained address-translation cache invalidation, v1.0
 - Sv32, Sv39, Sv48 and Sv57 page-based virtual-memory systems
 - Physical Memory Protection (PMP)
+- Svadu Extension for Hardware Updating of A/D Bits, Version 1.0
```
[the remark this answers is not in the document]
trdthg: done

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
trdthg: done

pmundkur:
It looks like we need to be able to enable Svadu and Svade independently.  How about the following:

| enable_Svadu | enable_Svade |  |
| --- | --- | --- |
| 0 | 0  | current behavior configurable with `-plat_enable_dirty_update` |
| 0  | 1   | `ADUE` is read-only 0,  `-plat_enable_dirty_update`  is an incompatible option error |
| 1   | 0  | `ADUE` is writable, `-plat_enable_dirty_update` selects whether `ADUE` is initialized to 1 |
| 1 | 1  |   `-plat_enable_dirty_update` selects whether `ADUE` is initialized to 1, and enable_Svade is effectively ignored since it is subsumed by `ADUE=0` |

Or we could ignore (or error on) `-plat_enable_dirty_update` for `enable_Svadu`/`enable_Svade` and leave `ADUE` unspecified on reset as the manual says.

jordancarlin:
> It looks like we need to be able to enable Svadu and Svade independently. How about the following:
> 
> enable_Svadu	enable_Svade	
> 0	0	current behavior configurable with `-plat_enable_dirty_update`
> 0	1	`ADUE` is read-only 0, `-plat_enable_dirty_update` is an incompatible option error
> 1	0	`ADUE` is writable, `-plat_enable_dirty_update` selects whether `ADUE` is initialized to 1
> 1	1	`-plat_enable_dirty_update` selects whether `ADUE` is initialized to 1, and enable_Svade is effectively ignored since it is subsumed by `ADUE=0`
> Or we could ignore (or error on) `-plat_enable_dirty_update` for `enable_Svadu`/`enable_Svade` and leave `ADUE` unspecified on reset as the manual says.

@trdthg based on Monday's meeting, it sounds like we want to return to a version similar to this with flags for both `svade` and `svadu`. Later on (once configuration is implemented) we can modify things so that `plat_enable_dirty_update` is not an option when `enable_svade` or `enable_svadu` are set.

pmundkur:
@trdthg do you plan on continuing on this in the near future? Else, I could try and push forward on this.

trdthg:
@pmundkur Not planning for now. Feel free to take over.

nadime15:
@pmundkur @trdthg Do you have any updates on this? If not, I can take it over as well.

pmundkur:
I'm working on this along with Svnapot, Svpbmt.

nadime15:
Is there a branch I can take a look at (at least for Svadu)?

pmundkur:
No branch yet, I'll be working on them right after I wrap up Svnapot.  They're all touching vmem_pte and vmem.

jordancarlin:
Closing now that #1599 is merged.

PR #729, opened 2025-02-11
Fix TLB address matching for sfence.vma

Timmmm:
A slight bug in the latest virtual memory rewrite: virtual addresses were zero extended in the TLB instead of sign extended, which meant that `sfence.vma` sometimes wouldn't match the TLB entries it should.

This was found after further internal testing. The failing test passes after these changes.


jordancarlin (approved):
Good catch and really reaffirms the need for improved testing of the model.

PR #730, opened 2025-02-11
Add support for Zcmp extension

nadime15:
Add the following instructions:

* cm.push
* cm.pop
* cm.popret
* cm.popretz
* cm.mvsa01
* cm.mva01s

### DIFF
--- a/model/extensions/Zcmp/zcmp_insts.sail
+++ b/model/extensions/Zcmp/zcmp_insts.sail
@@ -144,7 +144,7 @@ function clause execute (CMPP(cmop, rlist, spimm)) = {
 
   if cmop == CM_POPRETZ then X(a0) = zeros();
   X(sp) = new_sp;
-  if cmop == CM_POPRET | cmop == CM_POPRETZ then set_next_pc(X(ra));
+  if cmop == CM_POPRET | cmop == CM_POPRETZ then return jump_to([X(ra) with 0 = bitzero]);
   RETIRE_SUCCESS
 }
 
--- a/model/extensions/Zcmp/zcmp_insts.sail
+++ b/model/extensions/Zcmp/zcmp_insts.sail
@@ -1,0 +1,214 @@
+// =======================================================================================
+// This Sail RISC-V architecture model, comprising all files and
+// directories except where otherwise noted is subject the BSD
+// two-clause license in the LICENSE file.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+// =======================================================================================
+
+function clause currentlyEnabled(Ext_Zcmp) = hartSupports(Ext_Zcmp) & currentlyEnabled(Ext_Zca) & not(hartSupports(Ext_Zcd))
+
+mapping zcmp_abi_assembly_rlist : bits(4) <-> string = {
+  0x4 <-> "ra",
+  0x5 <-> "ra" ^ sep() ^ "s0",
+  0x6 <-> "ra" ^ sep() ^ "s0-s1",
+  0x7 <-> "ra" ^ sep() ^ "s0-s2",
+  0x8 <-> "ra" ^ sep() ^ "s0-s3",
+  0x9 <-> "ra" ^ sep() ^ "s0-s4",
+  0xa <-> "ra" ^ sep() ^ "s0-s5",
+  0xb <-> "ra" ^ sep() ^ "s0-s6",
+  0xc <-> "ra" ^ sep() ^ "s0-s7",
+  0xd <-> "ra" ^ sep() ^ "s0-s8",
+  0xe <-> "ra" ^ sep() ^ "s0-s9",
+  0xf <-> "ra" ^ sep() ^ "s0-s11",
+}
+
+mapping zcmp_arch_assembly_rlist : bits(4) <-> string = {
+  0x4 <-> "x1",
+  0x5 <-> "x1" ^ sep() ^ "x8",
+  0x6 <-> "x1" ^ sep() ^ "x8-x9",
+  0x7 <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18",
+  0x8 <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x19",
+  0x9 <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x20",
+  0xa <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x21",
+  0xb <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x22",
+  0xc <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x23",
+  0xd <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x24",
+  0xe <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x25",
+  0xf <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x27",
+}
+
+mapping zcmp_assembly_rlist : bits(4) <-> string = {
+  i if get_config_use_abi_names()      <-> zcmp_abi_assembly_rlist(i),
+  i if not(get_config_use_abi_names()) <-> zcmp_arch_assembly_rlist(i),
+}
+
+mapping zcmp_rlist_regmask : bits(4) <-> bits(32) = {
+  0x4 <-> 0b00000000000000000000000000000010,
+  0x5 <-> 0b00000000000000000000000100000010,
+  0x6 <-> 0b00000000000000000000001100000010,
+  0x7 <-> 0b00000000000001000000001100000010,
+  0x8 <-> 0b00000000000011000000001100000010,
+  0x9 <-> 0b00000000000111000000001100000010,
+  0xA <-> 0b00000000001111000000001100000010,
+  0xB <-> 0b00000000011111000000001100000010,
+  0xC <-> 0b00000000111111000000001100000010,
+  0xD <-> 0b00000001111111000000001100000010,
+  0xE <-> 0b00000011111111000000001100000010,
+  0xF <-> 0b00001111111111000000001100000010,
+}
+
+function get_stack_adj(rlist : bits(4), spimm : bits(2)) -> nat = {
+  let num_regs = count_ones(zcmp_rlist_regmask(rlist));
+  let chunks_for_regs = (num_regs * xlen_bytes + 15) / 16;
+  (chunks_for_regs + unsigned(spimm)) * 16
+}
+
+mapping zcmp_assembly_map : (bits(4), bits(2), bool) <-> string = {
+  forwards (rlist, spimm, is_negative) => {
+    let stack_adj = get_stack_adj(rlist, spimm);
+    let sign = if is_negative then "-" else "";
+
+    "{" ^ zcmp_assembly_rlist(rlist) ^ "}" ^ sep() ^ sign ^ dec_str(stack_adj)
+  },
+  backwards str_input => {
+    // Backwards mapping for cm.push, cm.pop, cm.popret, and cm.popretz
+    // is not implemented due to the complexity of parsing these instructions.
+    not_implemented("Backwards mapping for (cm.push/cm.pop/cm.popret/cm.popretz) is not implemented.");
+  },
+}
+
+union clause instruction = CM_PUSH : (bits(4), bits(2))
+
+mapping clause encdec_compressed = CM_PUSH(rlist, spimm)
+  <-> 0b101 @ 0b11000 @ rlist : bits(4) @ spimm : bits(2) @ 0b10
+  when currentlyEnabled(Ext_Zcmp) & rlist >=_u 0b0100 & (not(base_E_enabled) | rlist <=_u 0b0110)
+
+function clause execute (CM_PUSH(rlist, spimm)) = {
+  let addr      = X(sp);
+  let stack_adj = get_stack_adj(rlist, spimm);
+  let new_sp    = addr - stack_adj;
+
+  let reg_mask  = zcmp_rlist_regmask(rlist);
+  var sp_offset = negate_int(xlen_bytes);
+
+  foreach (i from (if base_E_enabled then 15 else 31) downto 1) {
+    if reg_mask[i] == bitone then {
+      let offset = to_bits_truncate(xlen, sp_offset);
+      let rs2    = Regidx(to_bits(regidx_bit_width, i));
+      let data   = X(rs2);
+      match vmem_write(sp, offset, xlen_bytes, data, Store(Data), false, false, false) {
+        Err(e) => ( return e ),
+        Ok(_)  => { sp_offset = sp_offset - xlen_bytes }
+      };
+    }
+  };
+
+  X(sp) = new_sp;
+  RETIRE_SUCCESS
+}
+
+mapping clause assembly = CM_PUSH(rlist, spimm)
+  <-> "cm.push" ^ spc() ^ zcmp_assembly_map(rlist, spimm, true)
+
+union clause instruction = CMPP : (cmop_zcmp, bits(4), bits(2))
+
+mapping encdec_cmop : cmop_zcmp <-> bits(5) = {
+  CM_POP     <-> 0b11010,
+  CM_POPRET  <-> 0b11110,
+  CM_POPRETZ <-> 0b11100,
+}
+
+mapping clause encdec_compressed = CMPP(cmop, rlist, spimm)
+  <-> 0b101 @ encdec_cmop(cmop) @ rlist : bits(4) @ spimm : bits(2) @ 0b10
+  when currentlyEnabled(Ext_Zcmp) & rlist >=_u 0b0100 & (not(base_E_enabled) | rlist <=_u 0b0110)
+
+function clause execute (CMPP(cmop, rlist, spimm)) = {
+  let addr      = X(sp);
+  let stack_adj = get_stack_adj(rlist, spimm);
+  let new_sp    = addr + stack_adj;
+
+  let reg_mask = zcmp_rlist_regmask(rlist);
+  var sp_offset : int = stack_adj - xlen_bytes;
+
+  foreach (i from (if base_E_enabled then 15 else 31) downto 1) {
+    if reg_mask[i] == bitone then {
+      let offset = to_bits_truncate(xlen, sp_offset);
+      let rd     = Regidx(to_bits(regidx_bit_width, i));
+      match vmem_read(sp, offset, xlen_bytes, Load(Data), false, false, false) {
+        Err(e)   => { return e },
+        Ok(data) => { X(rd) = data; sp_offset = sp_offset - xlen_bytes }
+      };
+    }
+  };
+
+  if cmop == CM_POPRETZ then X(a0) = zeros();
+  X(sp) = new_sp;
+  if cmop == CM_POPRET | cmop == CM_POPRETZ then set_next_pc(X(ra));
+  RETIRE_SUCCESS
+}
+
+mapping zcmp_mnemonic : cmop_zcmp <-> string = {
+  CM_POP     <-> "cm.pop",
+  CM_POPRET  <-> "cm.popret",
+  CM_POPRETZ <-> "cm.popretz",
+}
+
+mapping clause assembly = CMPP(cmop, rlist, spimm)
+  <->  zcmp_mnemonic(cmop) ^ spc() ^ zcmp_assembly_map(rlist, spimm, false)
+
+mapping rns_name : bits(3) <-> string = {
+  0b000 <-> "s0",
+  0b001 <-> "s1",
+  0b010 <-> "s2",
+  0b011 <-> "s3",
+  0b100 <-> "s4",
+  0b101 <-> "s5",
+  0b110 <-> "s6",
+  0b111 <-> "s7",
+}
+
+function rns_to_regidx(Cregidx(r) : cregidx) -> regidx =
+  match r {
+    0b000 => Regidx(zero_extend(0b1000)),                          // s0 / x8
+    0b001 => Regidx(zero_extend(0b1001)),                          // s1 / x9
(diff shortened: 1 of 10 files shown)


rez5427:
what

rez5427:
You should at least add a co-author.

nadime15:
Hey @rez5427, thank you for pointing that out! I didn’t mean to overlook your contribution, which is why I linked to your pull request in my PR message. I’ve now added you as a co-author in the commit.

nadime15:
@Alasdair Yeah sure, will do!

In [#8950389](https://github.com/riscv/sail-riscv/pull/730/commits/8950389693f192e190acb1050b1c539878269027) I added the Zcmp extension to the C emulator along with the --enable-zcmp command-line argument.

I retested everything, and it now works with or without virtual memory.

However, the warnings are still there, even after using `sys_enable_zcmp()` and `sys_enable_zcd`.

Maybe I missed something?

jordancarlin:
> However, the warnings are still there, even after using `sys_enable_zcmp()` and `sys_enable_zcd`.
> 
> Maybe I missed something?

Assuming you’re talking about the 4 warnings in your initial message, they won’t be impacted by these changes. They are new warnings introduced by the unreleased version of sail that I assume you’re using and either need changes to sail itself or #689.

Alasdair:
Yes as Jordan said, those warnings are unrelated to this commit. Don't worry about them.

nadime15:
@jordancarlin Oh yes! Good catch, I am going to fix that!

Btw. I forgot to change one line, which is fixed in https://github.com/riscv/sail-riscv/pull/730/commits/3d9cd3edf642e93056796500415388b742a34f66

Yes exactly I use the unreleased version of sail. But now I know thanks.

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,388 @@
...
+  let addr = X(sp);
+  let stack_adj = negate_int(get_stack_adj_base(rlist) + unsigned(spimm) * 16);
+  let new_sp = addr + to_bits(xlen, stack_adj);
+  let mask = zcmp_regmask(rlist);
+
+  var offset : int = width_bytes;
+
+  foreach (i from 31 downto 1) {
+    if mask[i] == bitone then {
+      match ext_data_get_addr(sp, to_bits(xlen, negate_int(offset)), Write(Data), xlen / 8) {
```
jordancarlin, line 204: Can `xlen / 8` just be `xlen_bytes`?

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,388 @@
...
+}
+
+/* ****************************************************************** */
+union clause ast = CM_PUSH : (bits(4), bits(2))
+
+mapping clause encdec_compressed = CM_PUSH(rlist, spimm)   if extensionEnabled(Ext_Zcmp)
+  <-> 0b101 @ 0b11000 @ rlist : bits(4) @ spimm : bits(2) @ 0b10    if extensionEnabled(Ext_Zcmp)
+
+function process_cmpush (rlist : bits(4), spimm : bits(2)) -> Retired = {
+  let width_bytes = xlen_bytes;
```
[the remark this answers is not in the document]
nadime15: Yes, you are right. I replaced it with `xlen_bytes`. Besides that, I refactored the code a bit further and cleaned it up some more.

pmundkur:
It might be better to factor out the core address logic from `STORE` into a separate function, and call that from `STORE` and `process_cmpush`.

If interrupts are to be taken during the loop (which is up to the implementation, according to the spec), it would need to involve the stepper.  This might be easier after something like #746.

nadime15:
Yes, I agree on both points. Having some form of access to the stepper is probably the way to go, just in case someone wants to run a test case with interrupts between load and store sequences.

Timmmm:
I think `execute` might be nested for RVC... but yeah I would avoid it in general.

As for being interrupted in the middle of the instruction, that is true for all of the vector extension as well. Currently there's no way for it to happen.

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,320 @@
...
+    (rlist, spimm54, is_negative)
+  },
+}
+
+function zcmp_regmask(rlist : bits(4)) -> bits(32) = {
+  var mask : bits(32) = zeros();
+
+  if rlist >=_u 0b0100 then {
+    mask[1] = bitone
+  };
```
pmundkur, line 148:
```suggestion
  if   rlist >=_u 0b0100 
  then mask[1] = bitone;
```

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,320 @@
...
+
+  foreach (i from 5 to unsigned(rlist)) {
+    assert(i <= 15);
+    if( (i - 5) < 2 ) then
+      mask[3 + i] = bitone
+    else {
+      mask[11 + i] = bitone
+    };
```
pmundkur, line 156:
```suggestion
    if   i - 5 < 2 
    then mask[ 3 + i] = bitone
    else mask[11 + i] = bitone;
```

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,320 @@
...
+  };
+
+  if(rlist == zero_extend(0xf)) then
+    mask[27] = bitone;
```
pmundkur, line 160:
```suggestion
  if   rlist == zero_extend(0xf)
  then mask[27] = bitone;
```

pmundkur (changes requested):

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,320 @@
...
+
+mapping zcmp_assembly_mapping : (bits(4), bits(2), bool) <-> string = {
+  forwards (rlist, spimm54, is_negative) => {
+    let stack_adj_base = get_stack_adj_base(rlist);
+    let stack_adj = stack_adj_base + unsigned(spimm54) * 16;
+    let sign = if is_negative then "-" else "";
+
+    "{" ^ zcmp_assembly_rlist(rlist) ^ "}" ^ sep() ^ sign ^ dec_str(stack_adj)
+  },
+  backwards str_input => {
```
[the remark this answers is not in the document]
pmundkur: Lets discuss this in the Monday meeting.

jordancarlin:
Would be good to revisit this now that #861 is merged.

nadime15:
Yes, I am working on it at the moment.

nadime15:
I just noticed in my logs, during a trap, I basically reexecuted the instruction that caused the trap plus the one before it.

I think its due to:

```
// Align value to min supported PC alignment. This is used to
// legalize xepc reads.
function align_pc(addr : xlenbits) -> xlenbits = {
  if misa[C] == 0b1
  then [addr with 0 = bitzero]
  else [addr with 1..0 = zeros()]
}
```
My code is aligned to 2-byte boundaries but misa[C] == 0b0.

So if I understand correctly, `misa[C] == 0b1` should also be set to 1 if either `Zcmp` or `Zcmt` is active.

Modifying the hartSupport clause for `Ext_C` would look something like:

```
function clause hartSupports(Ext_C) = 
    hartSupports(Ext_Zca) & 
    ((hartSupports(Ext_Zcf) | not(hartSupports(Ext_F)) | xlen != 32) & 
    (hartSupports(Ext_Zcd) | not(hartSupports(Ext_D)))) |
    (hartSupports(Ext_Zcmt) & not(hartSupports(Ext_Zcd)) |
    (hartSupports(Ext_Zcmp) & not(hartSupported(Ext_Zcd))
```

Do you think, this makes sense @jordancarlin ?

jordancarlin:
@nadime15 Good catch! Surprised this has slipped through testing so far.

> So if I understand correctly, misa[C] == 0b1 should also be set to 1 if either Zcmp or Zcmt is active.
>
> Modifying the hartSupport clause for Ext_C would look something like:
> 
> ```
> function clause hartSupports(Ext_C) = 
>     hartSupports(Ext_Zca) & 
>     ((hartSupports(Ext_Zcf) | not(hartSupports(Ext_F)) | xlen != 32) & 
>     (hartSupports(Ext_Zcd) | not(hartSupports(Ext_D)))) |
>     (hartSupports(Ext_Zcmt) & not(hartSupports(Ext_Zcd)) |
>     (hartSupports(Ext_Zcmp) & not(hartSupported(Ext_Zcd))
> ```

misa[C] has very specific semantics in the ISA manual and is not dependent on Zcmp or Zcmt. We should not change `hartSupports(Ext_C)`.


I think the correct solution is to change `align_pc` to be based on `currentlyEnabled(Ext_Zca)` instead of `misa[C]`. All of the other compressed extensions depend on Zca, so it should be sufficient to look at it for the alignment. This also means that this problem extends beyond just adding Zcmp/Zcmt. Currently an `RV32IF_Zca` hart would run into the same issue with C not being set while supporting compressed instructions.

jordancarlin:
@nadime15 see #953.

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+  0xf <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x27",
+}
+
+function get_stack_adj_base(rlist : bits(4)) -> int = {
+  if xlen == 32 then
+    match rlist[3 .. 2] {
+      0b01 => 16,
+      0b10 => 32,
+      0b11 => if rlist == 0b1111 then 64 else 48,
+      _    => 0
```
Timmmm, line 45: These `=> 0` cases should be unreachable right? Can you change them to `internal_error(__FILE__, __LINE__, ...)`?

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+  },
+}
+
+function zcmp_regmask(rlist : bits(4)) -> bits(32) = {
+  var mask : bits(32) = zeros();
+
+  if rlist >=_u 0b0100 then mask[1] = bitone;
+
+  foreach (i from 5 to unsigned(rlist)) {
+    assert(i <= 15);
```
Timmmm, line 81: This should be unnecessary.

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+    if i - 5 < 2
+    then mask[ 3 + i] = bitone
+    else mask[11 + i] = bitone
+  };
+
+  if rlist == zero_extend(0xf) then mask[27] = bitone;
```
Timmmm, line 87:
```suggestion
  if rlist == 0xf then mask[27] = bitone;
```

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+    "{" ^ zcmp_assembly_rlist(rlist) ^ "}" ^ sep() ^ sign ^ dec_str(stack_adj)
+  },
+  backwards str_input => {
+    /* Backwards mapping for cm.push, cm.pop, cm.popret, and cm.popretz */
+    /* is not implemented due to the complexity of parsing these instructions. */
+    not_implemented("Backwards mapping for (cm.push/cm.pop/cm.popret/cm.popretz) is not implemented.");
+  },
+}
+
+function zcmp_regmask(rlist : bits(4)) -> bits(32) = {
```
Timmmm, line 75: I wonder if this is just better as a mapping? It's actually less code, and way simpler to understand IMO. Plus it correctly fails on invalid rmasks.

```
mapping zcmp_regmask : bits(4) <-> bits(32) = {
  0x4 <-> 0b00000000...0010,
  0x5 <-> 0b00000000...0110,
...
}
```

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+union clause ast = CM_PUSH : (bits(4), bits(2))
+
+mapping clause encdec_compressed = CM_PUSH(rlist, spimm)
+  <-> 0b101 @ 0b11000 @ rlist : bits(4) @ spimm : bits(2) @ 0b10
+  when currentlyEnabled(Ext_Zcmp) & rlist >=_u 0b0100
+
+function clause execute (CM_PUSH(rlist, spimm)) = {
+  let addr      = X(sp);
+  let stack_adj = negate_int(get_stack_adj_base(rlist) + unsigned(spimm) * 16);
+  let new_sp    = addr + to_bits(xlen, stack_adj);
```
Timmmm, line 101: I think this can just be

```
let addr = addr - (get_stack_adj_base(rlist) + unsigned(spimm) * 16);
```

Or with the other suggestions

```
let addr = addr - get_stack_adj(rlist, spimm) * 16;
```

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+
+  let mask      = zcmp_regmask(rlist);
+  var offset    = negate_int(xlen_bytes);
```
Timmmm, line 104:
```
var offset : xlenbits = zeros();

foreach (i from 31 downto 1) {
  if mask[i] == bitone then {
    let rs2 = bits_regidx(to_bits(5, i));
    let data = X(rs2);
    match vmem_write(sp, offset, xlen_bytes, data, Write(Data), false, false, false) {
       Err(e) => ( return e ),
       Ok(_)  => { offset = offset - xlen_bytes }
     };
  }
};
```

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+  if rlist == zero_extend(0xf) then mask[27] = bitone;
+
+  mask
+}
+
+union clause ast = CM_PUSH : (bits(4), bits(2))
+
+mapping clause encdec_compressed = CM_PUSH(rlist, spimm)
+  <-> 0b101 @ 0b11000 @ rlist : bits(4) @ spimm : bits(2) @ 0b10
+  when currentlyEnabled(Ext_Zcmp) & rlist >=_u 0b0100
```
Timmmm, line 96: Rather than use `bits(4)`, I think it might work better like this:

```
type rlist = range(4, 15)

mapping rlist_bits : bits(4) <-> rlist = {
   r if r <_u 0b0100 <-> unsigned(r)
}
```

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+  0x9 <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x20",
+  0xa <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x21",
+  0xb <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x22",
+  0xc <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x23",
+  0xd <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x24",
+  0xe <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x25",
+  0xf <-> "x1" ^ sep() ^ "x8-x9" ^ sep() ^ "x18-x27",
+}
+
+function get_stack_adj_base(rlist : bits(4)) -> int = {
```
Timmmm, line 39: After actually understanding the spec, I think a clearer way would be something like this:

```
// stack_adj in units of 16 bytes.
function get_stack_adj(rlist_mask : bits(32), spimm : bits(2)) -> range(1, 10) = {
  let num_regs = count_ones(rlist_mask);
  assert(num_regs <= 13);
  let chunks_for_regs = (num_regs * xlen_bytes + 15) / 16;
  chunks_for_regs + unsigned(spimm)
}
```

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+    }
+  };
+
+  X(sp) = new_sp;
+  RETIRE_SUCCESS
+}
+
+mapping clause assembly = CM_PUSH(rlist, spimm)
+  <-> "cm.push" ^ spc() ^ zcmp_assembly_mapping(rlist, spimm, true)
+/* ****************************************************************** */
```
Timmmm, line 124: Add a blank line above this I think.

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+union clause ast = CMPP : (cmop_zcmp, bits(4), bits(2))
+
+mapping encdec_cmop : cmop_zcmp <-> bits(5) = {
+  CM_POP     <-> 0b11010,
+  CM_POPRET  <-> 0b11110,
+  CM_POPRETZ <-> 0b11100
```
Timmmm, line 130: Trailing comma.

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,224 @@
...
+
+function clause execute (CM_MVSA01(r1s', r2s')) = {
+  X(rns_to_regidx(r1s')) = X(a0);
+  X(rns_to_regidx(r2s')) = X(a1);
+  RETIRE_SUCCESS
+}
+
+mapping clause assembly = CM_MVSA01(r1s', r2s')
+  <-> "cm.mvsa01" ^ spc() ^ creg_zcmp_name(r1s') ^ sep() ^ creg_zcmp_name(r2s')
+/* ****************************************************************** */
```
Timmmm, line 210: Add a blank line above this.

model/riscv_types.sail
```diff
@@ -29,6 +29,8 @@ type csreg = bits(12)  /* CSR addressing */
 function regidx_offset(Regidx(r) : regidx, o : bits(5)) -> regidx = Regidx(r + o)
 function regidx_bits (Regidx(b) : regidx) -> bits(5) = b
 
+function bits_regidx (b : bits(5)) -> regidx = Regidx(b)
```
Timmmm, line 32: This doesn't seem necessary - just call `Regidx` directly.

Timmmm (changes requested):
Hey, sorry for the late review and quite large suggestions. I can have a go at it if you like.

model/riscv_insts_zcmp.sail
nadime15: Good idea, I like this much better!

model/riscv_insts_zcmp.sail
nadime15: I might miss something, but is there a reason you set `offset` to 0?

If `offset` is initialized to 0, the first write and the following ones will go to the wrong address.

model/riscv_insts_zcmp.sail
nadime15: `count_ones` Is not defined yet. I saw the [Issue](https://github.com/rems-project/sail/issues/564) which you have opened. I might make sense just define another mapping?

Timmmm: I opened a PR with `count_ones()`: #957

Another mapping would also work, though I quite like that this explains *why* the stack_adj values are what they are.

Up to you.

nadime15: Ok cool, thanks! Yeah I agree, I will wait and use it once its merged.

Timmmm: Yeah I guess it can't reason about the rounding-up division. I guess add an assertion. Unless there's a better way @Alasdair ?

model/riscv_insts_zcmp.sail
```diff
@@ -0,0 +1,206 @@
...
+  0x7 <-> "ra" ^ sep() ^ "s0-s2",
+  0x8 <-> "ra" ^ sep() ^ "s0-s3",
+  0x9 <-> "ra" ^ sep() ^ "s0-s4",
+  0xa <-> "ra" ^ sep() ^ "s0-s5",
+  0xb <-> "ra" ^ sep() ^ "s0-s6",
+  0xc <-> "ra" ^ sep() ^ "s0-s7",
+  0xd <-> "ra" ^ sep() ^ "s0-s8",
+  0xe <-> "ra" ^ sep() ^ "s0-s9",
+  0xf <-> "ra" ^ sep() ^ "s0-s11",
+  // Using X names instead of ABI names.
```
pmundkur, line 24: The compiler should really warn about overlapping mappings.  A better approach might be the way `reg_name` works by checking `get_config_use_abi_names()` and then calling the underlying `raw` mappings for X and ABI names.

pmundkur (changes requested):
One fixup needed; I'll give it another review after reading the spec.

model/riscv_insts_zcmp.sail
nadime15: I am not sure this is an improvement, to me, using `rlist >=_u 0b0100` is just cleaner and its immediately clear to the reader as well. Otherwise, you first have to look up and understand the mapping.

If others prefer this change, I am happy to update it.

This needs to be updated anyway once the E Extension is supported

> rlist values 0 to 3 are reserved for a future EABI variant called cm.pop.e

model/riscv_insts_zcmp.sail
pmundkur: How about
```
mapping zcmp_assembly_rlist : bits(4) <-> string = {
  i if get_config_use_abi_names()      <-> zcmp_abi_assembly_rlist(i),
  i if not(get_config_use_abi_names()) <-> zmp_arch_assembly_rlist(i),
}
```
where the `_abi_` and `_arch_` functions do the appropriate mapping.  Each function will be shorter and the `get_config_use_abi_names()` check doesn't have to be repeated on each line.

pmundkur (changes requested):

model/extensions/Zcmp/zcmp_insts.sail
```diff
@@ -0,0 +1,214 @@
...
+      match vmem_read(sp, offset, xlen_bytes, Load(Data), false, false, false) {
+        Err(e)   => { return e },
+        Ok(data) => { X(rd) = data; sp_offset = sp_offset - xlen_bytes }
+      };
+    }
+  };
+
+  if cmop == CM_POPRETZ then X(a0) = zeros();
+  X(sp) = new_sp;
+  if cmop == CM_POPRET | cmop == CM_POPRETZ then set_next_pc(X(ra));
```
Timmmm: Should we use `jump_to` here? And do you need to clear the lowest bit of `X(ra)`?

nadime15: Ahh yeah, good point, thanks!

nazar-pc:
This is now a blocker for https://github.com/riscv/riscv-arch-test/pull/1535

Issue #215, opened 2023-02-10
Easy model build without dependencies that require sudo acces


martinberger:
In the discussion #212 on v0.15 planning, @allenjbaum stated that *"easy model build without dependencies that require sudo access"* is one of the two most important issues facing Sail at the moment.  This thread is meant as a place where we can bullet-point the core requirements for this. With a clear specification it is be easier to conceive of a solution.

abukharmeh:
I think it could be really helpful if we make the CI upload the built model so people who are not intending to develop the the model can just download it and use it

martinberger:
> upload the built model so ...

Isn't that trivial?

abukharmeh:
Yep, just make the build directory job artifact

Timmmm:
I agree, I think uploading the generated C would be a good first move. Then all you need is a C compiler and libgmp which is not that hard to install.

This is how we work around the pain of installing the Sail compiler internally.

Alternatively (or ideally additionally), it would be nice if there was a binary distribution of Sail (including Z3!) that you could just download instead of having to deal with 🤬ing OPAM.

jordancarlin:
Seems like this can be closed now that we have a flow to distribute precompiled binaries (#715).

Timmmm:
Yes I think so, that and the Sail binary release. On Linux at least! Mac should come with the next Sail release and I'll do Windows when someone asks for it...

PR #731, opened 2025-02-12
Add a DOWNLOAD_GMP option to build against a local GMP build

arichardson:
This downloads the latest release of GMP and builds it as part of the overall emulator build. This should make it easier to build sail without having to install system dependencies.

See https://github.com/riscv/sail-riscv/issues/215

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,14 +63,35 @@ if (NOT ZLIB_FOUND)
     endif()
 endif()
 
-find_package(GMP)
-if (NOT GMP_FOUND)
-    if (APPLE)
-        message(FATAL_ERROR "GMP not found. Try 'brew install gmp'.")
-    elseif (UNIX)
-        message(FATAL_ERROR "GMP not found. Try 'sudo apt install libgmp-dev' or 'sudo dnf install gmp-devel'.")
-    else()
-        message(FATAL_ERROR "GMP not found.")
+option(DOWNLOAD_GMP "Download libgmp and build the library locally instead of using a system installation" OFF)
+if (DOWNLOAD_GMP)
+    include(ExternalProject)
+    ExternalProject_Add(
+        gmp
+        URL https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz
+        INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/gmp
+        CONFIGURE_COMMAND "<SOURCE_DIR>/configure" "--prefix=<INSTALL_DIR>" --disable-shared --enable-static
+    )
+    ExternalProject_Get_property(gmp INSTALL_DIR)
+    add_library(GMP::GMP STATIC IMPORTED GLOBAL)
+    # Work around `Imported target "GMP::GMP" includes non-existent path`.
+    # See https://gitlab.kitware.com/cmake/cmake/-/issues/15052
+    file(MAKE_DIRECTORY ${INSTALL_DIR}/include)
+    set_target_properties(GMP::GMP PROPERTIES
+            IMPORTED_LOCATION "${INSTALL_DIR}/lib/libgmp.a"
+            INTERFACE_INCLUDE_DIRECTORIES "${INSTALL_DIR}/include")
+    add_dependencies(GMP::GMP gmp)
+else()
+    find_package(GMP)
+    if (NOT GMP_FOUND)
+        set(download_gmp "setting -DDOWNLOAD_GMP=TRUE to use a local build of GMP instead of a system library")
+        if (APPLE)
+            message(FATAL_ERROR "GMP not found. Try 'brew install gmp' or ${download_gmp}.")
+        elseif (UNIX)
+            message(FATAL_ERROR "GMP not found. Try 'sudo apt install libgmp-dev' or 'sudo dnf install gmp-devel' or ${download_gmp}.")
+        else()
+            message(FATAL_ERROR "GMP not found. Try ${download_gmp}")
+        endif()
     endif()
 endif()
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,9 @@ build directory, then e.g.
 $ make -C build riscv_sim_rv64f_rvfi
 ```
 
+By default `build_simulators.sh` will download and build [libgmp](https://gmplib.org/).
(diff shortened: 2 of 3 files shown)


arichardson:
If we decide to use this I can add one for libz. However that dependency will go away soon and is commonly available so not sure if it's worthwhile

jordancarlin:
> If we decide to use this I can add one for libz. However that dependency will go away soon and is commonly available so not sure if it's worthwhile

I don't think it is worthwhile to add for zlib since (as you mentioned) it will be gone starting with the next Sail release. More worthwhile perhaps would be to have it automatically download the Sail compiler itself. That way there would be next to no dependences other than a compiler and cmake.

CMakeLists.txt
```diff
@@ -63,14 +63,37 @@ if (NOT ZLIB_FOUND)
...
+    add_dependencies(GMP::GMP gmp)
+else()
+    find_package(GMP)
+    if (NOT GMP_FOUND)
+        if (APPLE)
+            message(FATAL_ERROR "GMP not found. Try 'brew install gmp'.")
+        elseif (UNIX)
+            message(FATAL_ERROR "GMP not found. Try 'sudo apt install libgmp-dev' or 'sudo dnf install gmp-devel'.")
+        else()
+            message(FATAL_ERROR "GMP not found.")
```
jordancarlin, line 95: All of these error messages should be updated to mention the option to download gmp automatically instead of just mentioning the system package manager.

jordancarlin (changes requested):

CMakeLists.txt
```diff
@@ -63,14 +63,37 @@ if (NOT ZLIB_FOUND)
...
+        INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/gmp
+        CONFIGURE_COMMAND "<SOURCE_DIR>/configure" "--prefix=<INSTALL_DIR>" --disable-shared --enable-static
+        USES_TERMINAL_CONFIGURE TRUE
+        BUILD_IN_SOURCE FALSE
+        USES_TERMINAL_BUILD TRUE
+        USES_TERMINAL_INSTALL TRUE
+    )
+    ExternalProject_Get_property(gmp INSTALL_DIR)
+    add_library(GMP::GMP STATIC IMPORTED GLOBAL)
+    # Work around `Imported target "GMP::GMP" includes non-existent path`:
```
Timmmm, line 81: How come you get this error?

CMakeLists.txt
```diff
@@ -63,14 +63,37 @@ if (NOT ZLIB_FOUND)
...
+    include(ExternalProject)
+    ExternalProject_Add(
+        gmp
+        URL https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz
+        INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/gmp
+        CONFIGURE_COMMAND "<SOURCE_DIR>/configure" "--prefix=<INSTALL_DIR>" --disable-shared --enable-static
+        USES_TERMINAL_CONFIGURE TRUE
+        BUILD_IN_SOURCE FALSE
+        USES_TERMINAL_BUILD TRUE
+        USES_TERMINAL_INSTALL TRUE
```
Timmmm, line 77: How come these are needed?

CMakeLists.txt
```diff
@@ -63,14 +63,37 @@ if (NOT ZLIB_FOUND)
...
+option(DOWNLOAD_GMP "Download libgmp and build the library locally instead of using a system installation" OFF)
+if (DOWNLOAD_GMP)
+    include(ExternalProject)
+    ExternalProject_Add(
+        gmp
+        URL https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz
+        INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/gmp
+        CONFIGURE_COMMAND "<SOURCE_DIR>/configure" "--prefix=<INSTALL_DIR>" --disable-shared --enable-static
+        USES_TERMINAL_CONFIGURE TRUE
+        BUILD_IN_SOURCE FALSE
```
Timmmm, line 75: Presumably `BUILD_IN_SOURCE FALSE` is the default?

Timmmm (commented):
This is great! We should probably mention it in the README. Maybe use it in `build_emulators.sh`?

CMakeLists.txt
arichardson: I am not sure why this happens but it seems to only check presence of the include dir during CMake time (even if I add a BYPRODUCTS, to indicate that it will be created during the build). We might be able to turn it into a warning instead with: https://cmake.org/cmake/help/latest/policy/CMP0027.html (https://gitlab.kitware.com/cmake/cmake/-/blob/v3.27.4/Source/cmGeneratorTarget.cxx#L3658)

Seems to be a known issue: https://stackoverflow.com/questions/45516209/cmake-how-to-use-interface-include-directories-with-externalproject

CMakeLists.txt
arichardson: These are not needed, but are helpful if something goes wrong during the build/configure. Can remove it if you prefer

CMakeLists.txt
arichardson: Ah yes, that is a leftover from when I set it true before I realized I could use "<SOURCE_DIR>/configure" as the configure command.
```suggestion
```

CMakeLists.txt
Timmmm: Ah ok maybe we should add a comment linking to https://gitlab.kitware.com/cmake/cmake/-/issues/15052

CMakeLists.txt
Timmmm: Ah does it not show the build process without this, even on errors?

arichardson: Let me try removing it

arichardson:
New version should address all comments raised so far.

Timmmm (approved):
LGTM. Do we want to add a CI matrix entry that uses this option so it doesn't get broken in future?

arichardson:
> LGTM. Do we want to add a CI matrix entry that uses this option so it doesn't get broken in future?

Yes that would be good - let me do that as a follow-up?

jordancarlin (approved):
LGTM now. Agreed that we should test this in CI. Maybe just by running the `build_emulators.sh` script.

Timmmm:
Issue to track adding this to CI: #745

PR #732, opened 2025-02-12
Remove support for linking C emulator to Spike

### LINKED ISSUE #727: Remove `ENABLE_SPIKE`
https://github.com/riscv/sail-riscv/pull/725#discussion_r1948687529

The `ENABLE_SPIKE` parts are quite outdated, I'm removing them.

But I don't know what they are used for. Does anyone still use it?

I checked a bit, and the content is roughly as follows:

```
while not_done {
    /* run a Sail step */
    zstep(sail_step);

#ifdef ENABLE_SPIKE
    /* run a Spike step */
    tv_step(s);
#endif
}
```

trdthg:
resolve #727 

- remove all `#ifdef ENABLE_SPIKE`
- remote `init_spike()`, `init_check()`, `compare_states()`
- remove things about linked to Spike in README and Makefile

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -62,7 +62,6 @@ enum {
   OPT_CACHE_BLOCK_SIZE,
 };
 
-static bool do_dump_dts = false;
 static bool do_show_times = false;
 struct tv_spike_t *s = NULL;
 char *term_log = NULL;
@@ -141,7 +140,6 @@ static struct option options[] = {
     {"disable-vector-ext",          no_argument,       0, 'W'                     },
     {"mtval-has-illegal-inst-bits", no_argument,       0, 'i'                     },
     {"device-tree-blob",            required_argument, 0, 'b'                     },
-    {"dump-dts",                    no_argument,       0, 's'                     },
     {"terminal-log",                required_argument, 0, 't'                     },
     {"show-times",                  required_argument, 0, 'p'                     },
     {"report-arch",                 no_argument,       0, 'a'                     },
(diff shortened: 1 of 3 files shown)


jordancarlin (approved):
Generally LGTM. @Timmmm I'm wondering if any of these functions would be useful for lockstep with the model. I know you've worked on that a little bit already.

c_emulator/riscv_sim.c
```diff
@@ -199,20 +189,7 @@ static bool is_32bit_model(void)
...
-  if (dts_len > 0) {
-    unsigned char *dts = (unsigned char *)malloc(dts_len + 1);
-    dts[dts_len] = '\0';
-    tv_get_dts(s, dts, &dts_len);
-    fprintf(stdout, "%s\n", dts);
-  }
-#else
-  fprintf(stdout, "Spike linkage is currently needed to generate DTS.\n");
-#endif
+  fprintf(stdout, "Sail model currently doesn't support generating DTS.\n");
```
jordancarlin, line 192: We should probably open an issue about this so we can decide if we want to build support for this into the model or just drop the feature entirely and remove the flag.

Timmmm: Yeah I think it would be a handy feature. But we should probably just drop it while it isn't supported at all.

Timmmm (approved):
LGTM, thanks!

arichardson (approved):
The title should be "Remove" and not "Deprecate" but otherwise LGTM.

Timmmm:
> Generally LGTM. @Timmmm I'm wondering if any of these functions would be useful for lockstep with the model. I know you've worked on that a little bit already.

Yeah I think once we have that ready we can add this back a lot more easily (and ideally run it in CI).

c_emulator/riscv_sim.c
trdthg: dropped

trdthg:
> The title should be "Remove" and not "Deprecate" but otherwise LGTM.

done

PR #733, opened 2025-02-12
Add support for Zic64b

trdthg:
Seems not needed, if I understand correctly

> **Zic64b** Cache blocks must be 64 bytes in size, naturally aligned in the address space
>
> **Note**: This is a new extension name for this feature. While the general RISC-V specifications are agnostic to cache block size, selecting a common cache block size simplifies the specification and use of the following cache-block extensions within the application processor profile.
Software does not have to query a discovery mechanism and/or provide dynamic dispatch to the appropriate code. We choose 64 bytes at it is effectively an industry standard.
Implementations may use longer cache blocks to reduce tag cost provided they use 64-byte sub-blocks to remain compatible.
Implementations may use shorter cache blocks provided they sequence cache operations across the multiple cache blocks comprising a 64-byte block to remain compatible.

Spike added isa_parser for hints extension, but this is not even a hint? and sail doesn't have this requirement

The only thing can do is update the Wiki or explain it in README (seems not needed too)

So please feel free to close this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ Supported RISC-V ISA features
 - Zicntr and Zihpm extensions for counters, v2.0
 - Zicond extension for integer conditional operations, v1.0
 - Zicbom and Zicboz extensions for cache-block management (Zicbop not currently supported), v1.0
+- Zic64b: Cache block size is 64 bytes, v1.0
 - Zimop extension for May-Be-Operations, v1.0
 - M extension for integer multiplication and division, v2.0
 - Zmmul extension for integer multiplication only, v1.0
--- a/config/default.json
+++ b/config/default.json
@@ -63,6 +63,9 @@
         "U" : {
             "supported" : true
         },
+        "Zic64b" : {
+            "supported" : true
+        },
         "Zicbom" : {
             "supported" : true
         },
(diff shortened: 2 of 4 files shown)


jordancarlin:
I think it would make more sense to have an option to enable Zic64b that sets `plat_cache_block_size_exp` to 6.

Timmmm:
I'm not sure this is needed at all. It's a profile rule rather than an extension. I think probably the best thing to do is when we have the new config format we just provide some sample configs that conform to the various profiles, so you can do `riscv_sim --config configs/rv22_minimal.json` or whatever.

trdthg:
> have an option to enable Zic64b that sets plat_cache_block_size_exp to 6.

Agree

- Adding an `extensionEnabled(Zic64b)` is useless because the code will not handle cache_block of size 64 specially in the short term
- It makes sense to add an option to set `plat_cache_block_size_exp`, which needs to wait for the configuration system

Timmmm:
I still don't think this makes sense. It's not a "real" extension. It's just a profile requirement. We can do this now though:

> when we have the new config format we just provide some sample configs that conform to the various profiles, so you can do `riscv_sim --config configs/rv22_minimal.json` or whatever.

We probably want something like `config/{rva22,rva23,rvb23,rvc23}_{min,full}.json`.

trdthg:
🤔👍

Issue #712, opened 2025-02-06
SHIFTIOP mapping missing ifs


Timmmm:
```
mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SLLI) <-> 0b000000 @ shamt @ rs1 @ 0b001 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SRLI) <-> 0b000000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SRAI) <-> 0b010000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
```

It needs the `if` on the left hand side too.

Btw @Alasdair I looked through the whole RISC-V model and couldn't find a single case where we use `if` but have different conditions on each side. I wonder if the language should be changed somehow so that a) you don't need to duplicate the condition in the common case, b) you can't make the above mistake so easily (not the first time I've seen it). I'm not sure how though.

PR #735, opened 2025-02-12
Add missing ifs with SHIFTIOP mapping

KotorinMinami:
Minor fix mentioned in #712

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -210,9 +210,9 @@ mapping encdec_sop : sop <-> bits(3) = {
   RISCV_SRAI <-> 0b101
 }
 
-mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SLLI) <-> 0b000000 @ shamt @ rs1 @ 0b001 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
-mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SRLI) <-> 0b000000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
-mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SRAI) <-> 0b010000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
+mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SLLI) if xlen == 64 | shamt[5] == bitzero <-> 0b000000 @ shamt @ rs1 @ 0b001 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
+mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SRLI) if xlen == 64 | shamt[5] == bitzero <-> 0b000000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
+mapping clause encdec = SHIFTIOP(shamt, rs1, rd, RISCV_SRAI) if xlen == 64 | shamt[5] == bitzero <-> 0b010000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
 
 function clause execute (SHIFTIOP(shamt, rs1, rd, op)) = {
   let rs1_val = X(rs1);


jordancarlin (approved):

arichardson (approved):

Timmmm (approved):

radimkrcmar:
When looking at the code, I would assume that the three following mappings are equivalent, because if either side fails, then the mapping should fail as there is nothing to map...
```
SHIFTIOP(shamt, rs1, rd, RISCV_SRAI) if xlen == 64 | shamt[5] == bitzero <->
0b010000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
```
```
SHIFTIOP(shamt, rs1, rd, RISCV_SRAI) <->
0b010000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011 if xlen == 64 | shamt[5] == bitzero
```
```
SHIFTIOP(shamt, rs1, rd, RISCV_SRAI) if xlen == 64 | shamt[5] == bitzero <->
0b010000 @ shamt @ rs1 @ 0b101 @ rd @ 0b0010011
```

If I simplify the mapping into `OP(x) if x == 1 <-> x`:
The patch makes it look like only the forward mapping `OP(x) if x == 1 -> x` fails when `x != 1`, but shouldn't the backwards mapping `OP(x) -> x if x == 1` also fail, because there is no valid rhs for `x`?

One possible implication of this patch is that `shamt`/`rs1`/`rd` on lhs do not have to equal `shamt`/`rs1`/`rd` on rhs, in which case we have more serious issues.

What does Sail actually do here? (What tests fail without this patch?)

Thanks.

Timmmm:
> I would assume that the three following mappings are equivalent

Yeah... they aren't though. The `if` applies when mapping *from* that side. So the difference is that before this change if you did this on `xlen == 32`

```
let x = encdec(SHIFTIOP(0b11111, zeros(), zeros(), RISCV_SRAI));
```

Then before it would have mapped to the xlen 64 encoding, whereas now you'll get a match failure.

It doesn't really matter because we never actually do that, but every other `encdec` has the `if` on both sides so it makes sense to do it for consistency.

I slightly feel like the current `encdec` conflates *encoding* with validity checks in some places. For example it won't even decode the floating point instructions if `mstatus.FS=0`, which makes it kind of annoying to use as a general purpose encoded/decoder (e.g. for generating random instructions or writing a disassembler). Our disassembly function is:

```
function disassemble(instruction: bits(32)) -> string = {
    // Save and restore the FS/VS flags. If they are zero then those extensions are disabled.
    let prev_mstatus = mstatus;

    mstatus[FS] = 0b01;
    mstatus[VS] = 0b01;

    let disassembly = assembly(
        if isRVC(instruction[15..0])
        then encdec_compressed(instruction[15..0])
        else encdec(instruction)
    );

    mstatus = prev_mstatus;

    disassembly
}
```

Which is kind of weird! Probably not a huge issue though.

I do feel like the `if` syntax is a bit confusing - it makes the common case (identical `if`s on both sides) awkward, and if you *don't* remember to do that it kind of silently does the wrong thing. I thought I opened a Sail issue about it but apparently not.

radimkrcmar:
> > I would assume that the three following mappings are equivalent
> 
> Yeah... they aren't though. The `if` applies when mapping _from_ that side. So the difference is that before this change if you did this on `xlen == 32`
> 
> ```
> let x = encdec(SHIFTIOP(0b11111, zeros(), zeros(), RISCV_SRAI));
> ```
> 
> Then before it would have mapped to the xlen 64 encoding, whereas now you'll get a match failure.
> 
> It doesn't really matter because we never actually do that, but every other `encdec` has the `if` on both sides so it makes sense to do it for consistency.

Definitely, thanks.   I am surprised about the Sail behavior and the patch sadly doesn't provide any explanation.

> I slightly feel like the current `encdec` conflates _encoding_ with validity checks in some places. For example it won't even decode the floating point instructions if `mstatus.FS=0`, which makes it kind of annoying to use as a general purpose encoded/decoder (e.g. for generating random instructions or writing a disassembler). [...]
> Which is kind of weird! Probably not a huge issue though.

We cannot disassemble extensions that we haven't implemented, so I guess that not disassembling disabled extensions is fine, and welcome if it simplifies the code.
Would be nice if we were consistent about that, though, probably by not disassembling all disabled extensions.
Extensions might even define the same bit representation for different instructions (e.g. Zcd and Zcmp), so it doesn't seem like we have much other choice.
 
> I do feel like the `if` syntax is a bit confusing - it makes the common case (identical `if`s on both sides) awkward, and if you _don't_ remember to do that it kind of silently does the wrong thing. I thought I opened a Sail issue about it but apparently not.

I agree that Sail would be nicer with a mapping type that prevents these mistakes.

Alasdair:
I could add some syntax for a predicate that applies to both sides, maybe something like (`x <-> y where predicate`)?

I thought it would be reasonable to write mappings that satisfy some weaker property like 
```
forwards(x) ⊏ y ⟺ x ⊏ backwards(y)
```
in which case an `if` guard on one side would be reasonable, but it does seem like it is not really being used.

radimkrcmar:
> I could add some syntax for a predicate that applies to both sides, maybe something like (`x <-> y where predicate`)?

I would like to see a guaranteed 1:1 mapping, which would also indirectly fix this issue, because the `if` would have to behave the same when placed on any side.

> I thought it would be reasonable to write mappings that satisfy some weaker property

Makes sense.  Users could have been similarly unhappy if you forced explicit use of `forwards`/`backwards` for this.

Timmmm:
> Extensions might even define the same bit representation for different instructions (e.g. Zcd and Zcmp), so it doesn't seem like we have much other choice.

Yeah this was my conclusion too.

> `x <-> y where predicate`

That sounds like a great idea!

PR #738, opened 2025-02-14
Simplify PMP code

Timmmm:
Inline `pmpCheckPerms` and use `== Machine` instead of `match`. I think this is a lot easier to understand, and shorter.

### DIFF
--- a/model/riscv_pmp_control.sail
+++ b/model/riscv_pmp_control.sail
@@ -39,25 +39,14 @@ function pmpAddrRange(cfg: Pmpcfg_ent, pmpaddr: xlenbits, prev_pmpaddr: xlenbits
 /* permission checks */
 
 val pmpCheckRWX: (Pmpcfg_ent, AccessType(ext_access_type)) -> bool
-function pmpCheckRWX(ent, acc) = {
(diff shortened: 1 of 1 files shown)


model/riscv_pmp_control.sail
```diff
@@ -95,7 +84,7 @@ function pmpMatchEntry(addr: physaddr, width: xlenbits, acc: AccessType(ext_acce
   match pmpMatchAddr(addr, width, rng) {
     PMP_NoMatch      => PMP_Continue,
     PMP_PartialMatch => PMP_Fail,
-    PMP_Match        => if   pmpCheckPerms(ent, acc, priv)
+    PMP_Match        => if pmpCheckRWX(ent, acc) | (priv == Machine & not(pmpLocked(ent)))
```
arichardson, line 87: I guess the other options would be 
1) pass priv to pmpcheckRWX() and add `if priv == Machine & not(pmpLocked(ent)) then true else match ...`?
2) change pmpCheckPerms to an if instead of a match().

But 2) seems unnecessary since there is only one caller of pmpCheckPerms() so we might as inline it. LGTM.

arichardson (approved):

PR #741, opened 2025-02-14
Fix simulator exit code

Timmmm:
The new CMake testing infrastructure assumed (one might say logically!) that `riscv_sim_..` would exit with a non-zero exit code if a test failed. That actually wasn't the case, so in fact we haven't been testing anything.

This corrects that. Fortunately all the tests still pass.

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -842,10 +842,12 @@ void run_sail(void)
 
     if (zhtif_done) {
       /* check exit code */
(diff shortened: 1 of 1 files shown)


Timmmm:
Note, it probably isn't the cleanest way to exit, but I'm planning to rewrite all this anyway.

Alasdair (approved):

jordancarlin (approved):
Quick fix for a significant problem

Issue #412, opened 2024-02-23
Remove `instbits` and extend `Retired` enum


Timmmm:
Currently there's a global register:

```
/* internal state to hold instruction bits for faulting instructions */
register instbits : xlenbits
```

Which is saved after fetch:

```
                F_Base(w) => {
                  instbits = zero_extend(w);
```

And used in `handle_illegal()`:

```
function handle_illegal() -> unit = {
  let info = if plat_mtval_has_illegal_inst_bits ()
             then Some(instbits)
             else None();
```

This use of a global is unnecessary. It would be more elegant to:

1. Remove the `instbits` register.
2. Add `RETIRE_ILLEGAL` to the `Retire` enum.
3. Change all instances of `handle_illegal(); RETIRE_FAIL` to just `RETIRE_ILLEGAL`
4. Add `instbits` as a parameter to `handle_illegal()`.
5. Call `handle_illegal(...)` from `step()` if `retired == RETIRE_ILLEGAL`.

Timmmm:
Good call, that does sound better and I guess would let you treat other exceptions in the same way (I'll have to double check the code though). I don't think `exception_details` needs to include the instructions bits though since they are already available at the point in `step()` where it's handled (and they'd be a bit awkward to get in `execute()` since the instruction has already been decoded).

jrtc27:
> Good call, that does sound better and I guess would let you treat other exceptions in the same way (I'll have to double check the code though). I don't think `exception_details` needs to include the instructions bits though since they are already available at the point in `step()` where it's handled (and they'd be a bit awkward to get in `execute()` since the instruction has already been decoded).

Sorry I should have been clearer: instbits would be supplied to handle_exception to use in the case that the execute clause returned Exc_Illegal_Inst (names entirely made up, any semblance to real names is coincidental). Or, to put it another way: yes, that's what I meant, I just failed to clearly communicate it.

martinberger:
I'm also in favour of extending the `Retired` data type. There is another important case: debug triggers. You want those to be handled uniformly in `step` as well. But if you get a trigger for example of a load address, that trigger is found in the `execute` clause of the relevant load instruction. For this purpose a `RETIRE_TRIGGER(...)` would be neat.

PR #746, opened 2025-02-17
Draft: refactor the instruction stepper to better support waiting instructions like WFI, WRS.{NTO, STO}

pmundkur:
This tries to address the issues raised in #398 and implements some ideas discussed in #412.

The first commit refactors the stepper and instruction retirement, and addresses #412.
The second adds a Wait state to the stepper and exposes it to the C emulator.
The third adds a configuration option for the max number of wait states before the Wait (for WFI) expires.  This should provide a better foundation for #398.

This is an invasive and backwards-incompatible change, and so mainly submitted for discussion.

I am unsure of the effects on RVFI.  Is it still being used?

### DIFF
--- a/Makefile.old
+++ b/Makefile.old
@@ -106,8 +106,9 @@ SAIL_ARCH_SRCS += riscv_extensions.sail riscv_types_common.sail riscv_types_ext.
 SAIL_ARCH_SRCS += riscv_vmem_types.sail $(SAIL_REGS_SRCS) $(SAIL_SYS_SRCS) riscv_platform.sail
 SAIL_ARCH_SRCS += riscv_sstc.sail
 SAIL_ARCH_SRCS += riscv_mem.sail $(SAIL_VM_SRCS)
-SAIL_ARCH_RVFI_SRCS = $(PRELUDE) rvfi_dii.sail riscv_extensions.sail riscv_types_common.sail riscv_types_ext.sail riscv_types.sail riscv_vmem_types.sail $(SAIL_REGS_SRCS) $(SAIL_SYS_SRCS) riscv_platform.sail riscv_mem.sail $(SAIL_VM_SRCS) riscv_types_kext.sail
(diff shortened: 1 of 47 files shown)


jordancarlin:
> I am unsure of the effects on RVFI. Is it still being used?

It is still being used. There are plans to transition it to a runtime option instead of compile time, but we definitely don't want to break compatibility with it.

pmundkur:
What is the recommended way of testing RVFI?   

I tried using CTSRD-CHERI/TestRig with `utils/scripts/runTestRIG.py -a sail -b toooba -r rv64g`.  It was able to run on both master and this PR, but errors were reported against both.

jordancarlin:
> What is the recommended way of testing RVFI?

That's a good question and another thing that should be added to CI. Right now we only test building the RVFI model in CI.

pmundkur:
I confirmed with TestRig that there is no RVFI regression in this PR vs master.  The command used was:
`utils/scripts/runTestRIG.py -a manual -b manual --implementation-A-port 8080 --implementation-B-port 8081`
where the `riscv_rvfi_RV64` from the PR  and the master  were pre-launched with `-r <port>`

pmundkur:
@Alasdair   I tried out your suggestion to reduce the size of this PR by defining in `riscv_inst_retire.sail`:
```
let RETIRE_SUCCESS : Retired(Retire_Failure) = RETIRE_SUCCESS()
```
But this results in:
```
Warning: Identifier RETIRE_SUCCESS found in pattern is also a union constructor at riscv_inst_retire.sail:33.4-18:
33 |let RETIRE_SUCCESS : Retired(Retire_Failure) = RETIRE_SUCCESS()
   |    ^------------^
   | 
Suggestion: Maybe you meant to match against RETIRE_SUCCESS() instead?

Type error:
riscv_inst_retire.sail:33.4-18:
33 |let RETIRE_SUCCESS : Retired(Retire_Failure) = RETIRE_SUCCESS()
   |    ^------------^
   | Local variable RETIRE_SUCCESS is already bound as a function name

```
i.e. it results in both a warning and a type error.

Alasdair:
You would need to give the constructor a different name.

PR #747, opened 2025-02-17
Fix Makefile.old after f37f9290.

pmundkur:


jordancarlin (approved):
Seems fine. We need to think about when to stop pushing new features into the old Makefile though.

Timmmm:
Yeah there's a few old formal outputs that need to be migrated but I'd be reeeeally surprised if they actually work in the Makefile anyway given they aren't tested in CI and are pretty niche.

I would be in favour of creating issues to restore Isabelle, Hol4, etc. then pasting the old code into those issues and deleting this file.

Not urgent though.

pmundkur:
I ran into this when trying TestRig to test RVFI on my PR, since TestRig is using the makefile build.

arichardson:
> I ran into this when trying TestRig to test RVFI on my PR, since TestRig is using the makefile build.

You can also use the runTestRIG.py script to point it at a manually built+launched simulator using `--implementation-A-port` `--implementation-B-port`

PR #748, opened 2025-02-18
 store `total_insns` and `insn_limit` as `uint64_t`, not `int`

cfbolz:
At a few hundred thousand instructions per second, 32-bit `int total_insns` will overflow in a few hours. If the limit  is too high, it will be parsed incorrectly without error and execution stops at a somewhat arbitrary point (or not at all).

This PR therefore changes the types to `long`. 

I've also switched to using `strtol` for parsing the `insns_limit` command line argument in order to get error checking (and to make it possible to use hex notation).

### DIFF
--- a/c_emulator/riscv_sim.c
+++ b/c_emulator/riscv_sim.c
@@ -112,8 +112,8 @@ void set_config_print(char *var, bool val)
 }
 
 struct timeval init_start, init_end, run_end;
-int total_insns = 0;
-int insn_limit = 0;
+uint64_t total_insns = 0;
+uint64_t insn_limit = 0;
 #ifdef SAILCOV
 char *sailcov_file = NULL;
 #endif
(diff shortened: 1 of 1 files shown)


c_emulator/riscv_sim.c
```diff
@@ -112,8 +112,8 @@ void set_config_print(char *var, bool val)
 }
 
 struct timeval init_start, init_end, run_end;
-int total_insns = 0;
-int insn_limit = 0;
+long total_insns = 0;
+long insn_limit = 0;
```
radimkrcmar, line 116: `long` doesn't have to be any longer than 32-bits (and it's not on Windows) and the counter shouldn't be signed.
How about `uintmax_t`/`uint64_t`/defining own better type?

Timmmm: Yeah there's also no reason for it to be signed. Either `unsigned long long` of (my preference) `uint64_t`.

cfbolz: makes sense! I switched to `uint64_t`, thanks for the feedback.

Timmmm (approved):
LGTM, thanks!

c_emulator/riscv_sim.c
```diff
@@ -378,7 +378,12 @@ static int process_args(int argc, char **argv)
       set_config_print(optarg, true);
       break;
     case 'l':
-      insn_limit = atoi(optarg);
+      char *temp;
+      insn_limit = (uint64_t)strtoull(optarg, &temp, 0);
+      if (*temp != '\0') {
+        fprintf(stderr, "invalid instruction limit %s\n", optarg);
+        exit(1);
+      }
```
jrtc27, line 386: To be robust you want something like:
```c
    case 'l': {
      char *p;
      unsigned long long val;
      errno = 0;
      val = strtoull(optarg, &p, 0);
      if (*p != '\0' || val > UINT64_MAX || (val == ULLONG_MAX && errno == ERANGE)) {
        fprintf(stderr, "invalid instruction limit %s\n", optarg);
        exit(1);
      }
      break;
    }
```
Note the curly braces too for scoping, case statements are just glorified labels.

cfbolz: thanks, done

Timmmm:
I will merge this in a few days if there are no objections.

jordancarlin (approved):

PR #749, opened 2025-02-18
Handwritten support for the Lean backend

ineol:
This is still work in progress

### DIFF
--- a/handwritten_support/RiscvExtras.lean
+++ b/handwritten_support/RiscvExtras.lean
@@ -1,0 +1,56 @@
+
+import THE_MODULE_NAME.Sail.Sail
+
+open Sail
+
+def print_endline (_ : String) : Unit := ()
+def prerr_endline (_: String) : Unit := ()
+def prerr_string (_: String) : Unit := ()
+def putchar {T} (_: T ) : Unit := ()
+def string_of_int (z : Int) := s!"{z}"
+
+axiom sys_enable_writable_misa : Unit -> Bool
+axiom sys_enable_rvc : Unit -> Bool
+axiom sys_enable_fdext : Unit -> Bool
+axiom sys_enable_svinval : Unit -> Bool
+axiom sys_enable_zcb : Unit -> Bool
+axiom sys_enable_zfinx : Unit -> Bool
+axiom sys_enable_writable_fiom : Unit -> Bool
+axiom sys_enable_vext : Unit -> Bool
+axiom sys_enable_bext : Unit -> Bool
+axiom sys_enable_zicbom : Unit -> Bool
+axiom sys_enable_zicboz : Unit -> Bool
+axiom sys_enable_sstc : Unit -> Bool
+axiom sys_writable_hpm_counters : Unit -> BitVec 32
+
+axiom sys_vext_vl_use_ceil : Unit -> Bool
+axiom sys_vector_elen_exp : Unit -> Nat
+axiom sys_vector_vlen_exp : Unit -> Nat
+
+axiom sys_pmp_count : Unit -> Nat
+axiom sys_pmp_count_ok : 0 ≤ sys_pmp_count () ∧ sys_pmp_count () ≤ 64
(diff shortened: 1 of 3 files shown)


handwritten_support/RiscvExtras.lean
```diff
@@ -0,0 +1,136 @@
+
+import THE_MODULE_NAME.Sail.Sail
```
tobiasgrosser, line 2: This breaks the build for me. How is `The_MODULE_NAME` supposed to be updated?

handwritten_support/RiscvExtras.lean
```diff
@@ -0,0 +1,136 @@
...
+abbrev bits_H := BitVec 16  -- Half-precision float value
+abbrev bits_S := BitVec 32  -- Single-precision float value
+abbrev bits_D := BitVec 64  -- Double-precision float value
+
+abbrev bits_W := BitVec 32  -- Signed integer
+abbrev bits_WU := BitVec 32  -- Unsigned integer
+
+abbrev bits_L := BitVec 64  -- Signed integer
+abbrev bits_LU := BitVec 64  -- Unsigned integer
+
```
tobiasgrosser, line 70:
```suggestion
```

These lines are already discussed in the lean output that we generate. I now get duplicate definition warnings. How do we handle this?

handwritten_support/RiscvExtras.lean
ineol: Are you on MacOS or Linux? the lean translator uses `sed` to replace it with the actual name of the module. Until [the PR fixing sed on linux](https://github.com/rems-project/sail/pull/1041) is merged it's not going to work :/

handwritten_support/RiscvExtras.lean
ineol: @javra do you know what is happening?

handwritten_support/RiscvExtras.lean
tobiasgrosser: My CI runs on Linux. I meanwhile fixed it with my own sed call in the CI.

handwritten_support/RiscvExtras.lean
ineol: Would it work to put all the type definitions in a file `Types.lean`, have `RiscvExtras.lean` and `Definitions.lean` import `Types.lean`? or are there more cycles?

handwritten_support/RiscvExtras.lean
tobiasgrosser: Indeed, this works now and this comment is resolved.

handwritten_support/RiscvExtras.lean
tobiasgrosser: I think this could work.

javra: The alternative would be to just inline them here

javra: (Did that, I doubt they change a lot)

tobiasgrosser (approved):
This now runs https://github.com/opencompl/sail-riscv-lean/ and works flawlessly. Happy to see this merged.

tobiasgrosser:
@Timmmm or someone else, can you trigger the CI and potentially get this merged?

Alasdair:
I just ran the CI, but it's only running some tests for the C emulator so it's unlikely this will break anything

Timmmm:
Btw @Alasdair you might consider how indentation is done in the generated output. It's 11 MB currently, and because of the crazy indentation it's actually 81% leading whitespace! I don't know how you'd format it nicely, but a very easy improvement would be to use glorious tabs instead of spaces. That reduces the output size to 6 MB.

This affects the C output too but IIRC not as bad.

tobiasgrosser:
We have a nicer output formatting that is currently buggy but should hopefully address this case.

tobiasgrosser:
Right. The real solution is to get bv_match enabled and do multi-dimensional matching. That should reduce indentation and LoC, AFAIU.

PR #750, opened 2025-02-19
Update mstatus bits guarded by privilege modes

jordancarlin:
A recent PR against the isa manual (https://github.com/riscv/riscv-isa-manual/pull/1864) clarified that mstatus SIE and SPIE should be read-only zero when S mode is not supported. While updating, I noticed several other bits that should also be guarded by this constraint according to the spec.

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -251,11 +251,11 @@ function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
     // UXL = if xlen == 64 then v[UXL] else o[UXL],
     // SDT = v[SDT],
     // SPELP = v[SPELP],
-    TSR = v[TSR],
-    TW = v[TW],
-    TVM = v[TVM],
-    MXR = v[MXR],
-    SUM = v[SUM],
+    TSR = if extensionEnabled(Ext_S) then v[TSR] else 0b0,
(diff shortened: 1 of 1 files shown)


trdthg:
I have always had a small question, what version of spec should we develope with, the release version or the in develop version?
Will sail have a config to change different spec versions in the future?
If yes, then both behaviors will be supported when we have that config?

jordancarlin:
> I have always had a small question, what version of spec should we develope with, the release version or the in develop version? Will sail have a config to change different spec versions in the future? If yes, then both behaviors will be supported when we have that config?

The plan is definitely to support multiple versions of the spec (primarily the privileged specs) once config is up and running.

All of the mstatus bits other than SIE and SPIE that I changed have had the constraint that they were read-only zero if supervisor mode was not implemented going back to at least priv 1.10, so versions shouldn't affect those parts. 

SIE and SPIE are a trickier case because it is being presented as just a clarification, not an actual change. If there were implementations that didn't make these bits read only zero though, then it is most definitely a breaking change. Maybe for now we update it accordingly and then add some configuration option later if needed.

pmundkur (approved):
Good catch!

arichardson (approved):

pmundkur:
Merging this since it's been two weeks since the `will be merged` label.

PR #751, opened 2025-02-19
Big endianness Support added

MuhammadHammad001:
Hi @jordancarlin and @Timmmm 
This PR adds the support for the big endianness in Sail RISC-V.

The test for the verification of this support is also added in this PR.

### DIFF
--- a/model/riscv_mem.sail
+++ b/model/riscv_mem.sail
@@ -228,7 +228,7 @@ function mem_write_value_priv_meta (paddr, width, value, typ, priv, meta, aq, rl
   else {
     let result = checked_mem_write(paddr, width, value, typ, priv, meta, aq, rl, con);
     match result {
-      Ok(_) => mem_write_callback(bits_of(paddr), width, value),
+      Ok(_) => mem_write_callback(to_str(typ), bits_of(paddr), width, value),
       Err(e) => mem_exception_callback(bits_of(paddr), num_of_ExceptionType(e)),
     };
     result
--- a/.github/workflows/compile-lean.yml
+++ b/.github/workflows/compile-lean.yml
@@ -1,6 +1,6 @@
 name: Build and test Lean backend
 
-on: [workflow_dispatch]
+on: [pull_request, workflow_dispatch]
 
 env:
   OPAMVERBOSE: 1
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,17 @@
 cmake_minimum_required(VERSION 3.20)
 cmake_policy(VERSION 3.20...3.31)
 
+# CMake 3.24 and newer don't use the mtimes from archives when extracting
+# GMP in ExternalProject_Add(). This reverts that behaviour so that the
+# mtimes in the archive are used. This is necessary because otherwise
+# Make tries to rebuild the documentation that is already built, and
+# this requires `makeinfo` which we don't want to require.
+# TODO: Remove this when we bump the minimum CMake version to 3.24 and
+# instead add DOWNLOAD_EXTRACT_TIMESTAMP TRUE to ExternalProject_Add().
+if(POLICY CMP0135)
+    cmake_policy(SET CMP0135 OLD)
+endif()
+
 project(sail_riscv)
 
 set(CMAKE_CXX_STANDARD 17)
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -1,6 +1,7 @@
 #include "riscv_callbacks.h"
 #include "riscv_config.h"
 #include "riscv_platform_impl.h"
+#include "riscv_sail.h"
 #include <stdlib.h>
 #include <vector>
 #include <inttypes.h>
@@ -20,10 +21,12 @@ void print_lbits_hex(lbits val, int length = 0)
 
 // Implementations of default callbacks for trace printing and RVFI.
 // The model assumes that these functions do not change the state of the model.
-unit mem_write_callback(uint64_t paddr, uint64_t width, lbits value)
+unit mem_write_callback(const char *type, uint64_t paddr, uint64_t width,
+                        lbits value)
 {
   if (config_print_mem_access) {
-    fprintf(trace_log, "mem[0x%016" PRIX64 "] <- 0x", paddr);
+    fprintf(trace_log, "mem[%s,0x%0*" PRIX64 "] <- 0x", type,
+            static_cast<int>((zphysaddrbits_len + 3) / 4), paddr);
     print_lbits_hex(value, width);
   }
   if (config_enable_rvfi) {
@@ -36,7 +39,8 @@ unit mem_read_callback(const char *type, uint64_t paddr, uint64_t width,
                        lbits value)
 {
   if (config_print_mem_access) {
-    fprintf(trace_log, "mem[%s,0x%016" PRIX64 "] -> 0x", type, paddr);
+    fprintf(trace_log, "mem[%s,0x%0*" PRIX64 "] -> 0x", type,
+            static_cast<int>((zphysaddrbits_len + 3) / 4), paddr);
     print_lbits_hex(value, width);
   }
   if (config_enable_rvfi) {
@@ -102,7 +106,10 @@ unit vreg_write_callback(unsigned reg, lbits value)
 {
   if (config_print_reg) {
     fprintf(trace_log, "v%d <- ", reg);
-    print_lbits_hex(value);
+    // TODO: the width of `value` is currently `vlenmax` bits which can be much
+    // greater than VLEN. In future we will remove `vlenmax`, then we can remove
+    // the `zVLEN / 8` argument here.
+    print_lbits_hex(value, zVLEN / 8);
   }
   return UNIT;
 }
@@ -116,7 +123,7 @@ unit pc_write_callback(uint64_t value)
 unit trap_callback(unit)
 {
   if (config_enable_rvfi) {
-    zrvfi_trap();
+    zrvfi_trap(UNIT);
   }
   return UNIT;
 }
--- a/c_emulator/riscv_callbacks.h
+++ b/c_emulator/riscv_callbacks.h
@@ -5,7 +5,8 @@
 extern "C" {
 #endif
 
-unit mem_write_callback(uint64_t paddr, uint64_t width, lbits value);
+unit mem_write_callback(const char *type, uint64_t paddr, uint64_t width,
+                        lbits value);
 unit mem_read_callback(const char *type, uint64_t paddr, uint64_t width,
                        lbits value);
 unit mem_exception_callback(uint64_t paddr, uint64_t num_of_exception);
@@ -27,7 +28,7 @@ unit zrvfi_write(uint64_t paddr, int64_t width, lbits value);
 unit zrvfi_read(uint64_t paddr, sail_int width, lbits value);
 unit zrvfi_mem_exception(uint64_t paddr);
 unit zrvfi_wX(int64_t reg, uint64_t value);
-unit zrvfi_trap();
+unit zrvfi_trap(unit);
 
 #ifdef __cplusplus
 }
--- a/c_emulator/riscv_sail.h
+++ b/c_emulator/riscv_sail.h
@@ -41,6 +41,8 @@ unit zprint_rvfi_exec(unit);
 unit zprint_instr_packet(uint64_t);
 
 extern mach_bits zxlen;
+extern mach_bits zphysaddrbits_len;
+extern mach_bits zVLEN;
 extern bool zhtif_done;
 extern mach_bits zhtif_exit_code;
 extern bool have_exception;
--- a/config/default.json
+++ b/config/default.json
@@ -218,7 +218,8 @@
       "supported": true
     },
     "Svbare": {
-      "supported": true
+      "supported": true,
+      "sfence_vma_illegal_if_svbare_only": true
     },
     "Sv32": {
       "supported": true
--- a/model/riscv_callbacks.sail
+++ b/model/riscv_callbacks.sail
@@ -7,7 +7,7 @@
 /*=======================================================================================*/
 
 // Callbacks for state-changing events
-val mem_write_callback = pure {c: "mem_write_callback"} : forall 'n, 0 < 'n <= max_mem_access . (/* addr */ physaddrbits, /* width */ int('n), /* value */ bits(8 * 'n)) -> unit
+val mem_write_callback = pure {c: "mem_write_callback"} : forall 'n, 0 < 'n <= max_mem_access . (/* access type */ string, /* addr */ physaddrbits, /* width */ int('n), /* value */ bits(8 * 'n)) -> unit
 function mem_write_callback(_) = ()
 
 val mem_read_callback = pure {c: "mem_read_callback"} : forall 'n, 0 < 'n <= max_mem_access . (/* access type */ string, /* addr */ physaddrbits, /* width */ int('n), /* value */ bits(8 * 'n)) -> unit
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -625,6 +625,7 @@ union clause ast = SFENCE_VMA : (regidx, regidx)
 
 mapping clause encdec = SFENCE_VMA(rs1, rs2)
   <-> 0b0001001 @ encdec_reg(rs2) @ encdec_reg(rs1) @ 0b000 @ 0b00000 @ 0b1110011
+  when virtual_memory_supported() | not(config extensions.Svbare.sfence_vma_illegal_if_svbare_only : bool)
 
 function clause execute SFENCE_VMA(rs1, rs2) = {
   let addr = if rs1 != zreg then Some(X(rs1)) else None();


model/riscv_sys_regs.sail
```diff
@@ -269,6 +270,7 @@ function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
     SPP = if extensionEnabled(Ext_S) then v[SPP] else 0b0,
     VS = v[VS],
     MPIE = v[MPIE],
+    UBE = v[UBE],
```
jordancarlin, line 273: This should be read-only zero if U is not supported.

model/riscv_mem.sail
```diff
@@ -63,6 +63,15 @@ function write_kind_of_flags (aq : bool, rl : bool, con : bool) -> write_kind =
     (true,  false, true)  => throw(Error_not_implemented("sc.aq"))
   }
 
+val check_endianess: unit -> bool
+function check_endianess() = {
+    match cur_privilege {
```
jordancarlin, line 68: Endianness is influenced by `mstatus.MPRV`, so I think this needs be based on something like `effectivePrivilege`.

jordancarlin (changes requested):
Thanks for working on this! I left some initial comments below.

One other thing that I don't see here is the necessary modification to the reset function. Reset should set endianness back to little-endian if supported.

model/prelude.sail
```diff
@@ -236,3 +236,6 @@ type max_mem_access : Int = 4096
 
 // Type used for memory access widths. Zero byte accesses are not allowed.
 type mem_access_width = range(1, max_mem_access)
+
+// Function to reverse endianness.
+val reverse_endianness = pure {c: " reverse_endianness"} : forall 'n . (bits('n * 8)) -> bits('n * 8)
```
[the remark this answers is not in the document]
MuhammadHammad001: Done

nadime15:
I just noticed that `sstatus` also has a `UBE` bit. It this should be added to `bitfield Sstatus` including `lower_mstatus()` and `lift_sstatus()`

model/riscv_sys_control.sail
```diff
@@ -339,6 +339,9 @@ function reset_sys() -> unit = {
   mstatus[MIE] = 0b0;
   mstatus[MPRV] = 0b0;
 
+  // If little-endian memory accesses are supported, the mstatus/mstatush field MBE is reset to 0.
+  mstatus[MBE] = 0b0;
+
```
jordancarlin, line 344: This text is duplicated again right below this.

model/riscv_mem.sail
```diff
@@ -63,6 +63,15 @@ function write_kind_of_flags (aq : bool, rl : bool, con : bool) -> write_kind =
     (true,  false, true)  => throw(Error_not_implemented("sc.aq"))
   }
 
+val check_endianess: (AccessType(ext_access_type)) -> bool
+function check_endianess(typ) = {
```
pmundkur, line 67: The code would be easier to understand if this function was called `use_bigendian` or something similarly explicit.

MuhammadHammad001: Done

pmundkur:
The readme can also be updated to mention this support.

model/riscv_mem.sail
```diff
@@ -73,9 +81,15 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext
     (Execute(),  None()) => Err(E_Fetch_Access_Fault()),
     (Read(Data), None()) => Err(E_Load_Access_Fault()),
     (_,          None()) => Err(E_SAMO_Access_Fault()),
-    (_,      Some(v, m)) => { if   get_config_print_mem()
-                              then print_mem("mem[" ^ to_str(t) ^ "," ^ BitStr(physaddr_bits(paddr)) ^ "] -> " ^ BitStr(v));
-                              Ok(v, m) }
+    (_,      Some(v, m)) => {
+                            let temp_var: bits('n * 8) = match (t, is_big_endian(t)) {
+                                (Execute(), _) => (v),
```
jordancarlin, line 86: Why not move the check for an execute access into `is_big_endian`? It already needs to get AccessType to determine effective privilege, and if it is an execute type it will always be little endian.

model/riscv_mem.sail
```diff
@@ -73,9 +81,15 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext
     (Execute(),  None()) => Err(E_Fetch_Access_Fault()),
     (Read(Data), None()) => Err(E_Load_Access_Fault()),
     (_,          None()) => Err(E_SAMO_Access_Fault()),
-    (_,      Some(v, m)) => { if   get_config_print_mem()
-                              then print_mem("mem[" ^ to_str(t) ^ "," ^ BitStr(physaddr_bits(paddr)) ^ "] -> " ^ BitStr(v));
-                              Ok(v, m) }
+    (_,      Some(v, m)) => {
+                            let temp_var: bits('n * 8) = match (t, is_big_endian(t)) {
```
jordancarlin, line 85: Let's use a more specific/significant name than `temp_var`. Especially as we move to include Sail code in the ISA manual, we want to make sure the meaning of everything is clear.

model/riscv_mem.sail
```diff
@@ -63,6 +63,14 @@ function write_kind_of_flags (aq : bool, rl : bool, con : bool) -> write_kind =
...
   }
 
+function is_big_endian(typ : AccessType(ext_access_type)) -> bool = {
+    match effectivePrivilege(typ, mstatus, cur_privilege) {
+      Machine     => bits_to_bool(mstatus[MBE]),
+      Supervisor  => bits_to_bool(mstatus[SBE]),
+      User        => bits_to_bool(mstatus[UBE]),
+    }
+}
+
```
pmundkur, line 73: Ideally, helper functions like these should not access global state like `mstatus` and `cur_privilege`, and instead receive them as arguments.  This makes them easier to unit test, once we have that infrastructure.

model/prelude.sail
Timmmm, line 241:
```suggestion
val reverse_endianness = pure {c: "reverse_endianness"} : forall 'n . (bits('n * 8)) -> bits('n * 8)
```

model/riscv_mem.sail
Timmmm: Yeah, it's definitely good practice in normal programming, but it seems like Sail really encourages global variables and I think it make it more pseudo-cody.

My plan for dealing with this in unit tests was to re-initialise the model before running each unit test, so although `cur_privilege` is still global you don't have to worry about other tests modifying it, or restoring its state after your test. See [this code](https://github.com/rems-project/sail/pull/928/files#diff-86bc5538625e508bdc56ba5fe9b26858ac5211517533e23f4444255e45aa4136R2244) which I really need to finish!

model/riscv_mem.sail
Timmmm: Yeah we can just reuse `v`. I think

```
(_, Some(v, m)) => {
  let v = endianness_conversion(v, t);
  if  ...
}

function endianness_conversion(data : bits(...), ac : AccessType) -> bits(..) = 
  if is_big_endian(t) & not(isExecute(ac)) then reverse_endian(data) else data
```

Smpmp would also benefit from `isExecute()` so I think it makes sense to add it.

model/riscv_mem.sail
```diff
@@ -196,9 +210,13 @@ $endif
 
 // only used for actual memory regions, to avoid MMIO effects
 function phys_mem_write forall 'n, 0 < 'n <= max_mem_access . (wk : write_kind, paddr : physaddr, width : int('n), data : bits(8 * 'n), meta : mem_meta) -> MemoryOpResult(bool) = {
-  let result = write_ram(wk, paddr, width, data, meta);
+  let temp_var: bits('n * 8) = match (is_big_endian(Write())) {
```
Timmmm, line 213: There's a more normal syntax for `match <bool> {`!

Timmmm (changes requested):
Looking pretty close to me. I think we will need some kind of testing for this... It feels like the sort of thing that is easy to screw up without knowing. How do you compile a big endian ELF though?

Timmmm:
Ah never mind, I saw your test.

jordancarlin:
@MuhammadHammad001 Seems like this is pretty close. Would be good to get merged before conflicts start to build. Any chance you can finish it off soon?

Timmmm:
> The test for the verification of this support is available [here](https://github.com/MuhammadHammad001/big-endianness-test).

I just merged support for first-party tests so we could add this as part of this PR now!

MuhammadHammad001:
> @MuhammadHammad001 Seems like this is pretty close. Would be good to get merged before conflicts start to build. Any chance you can finish it off soon?

Hi Jordan and Tim, sorry for the late response. I will commit the requested changes today.

model/riscv_sys_regs.sail
```diff
@@ -244,8 +245,8 @@ function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
     // MPV = v[MPV],
     // GVA = v[GVA],
     /* We don't currently support changing MBE and SBE. */
```
jordancarlin, line 247: This comment is no longer true

jordancarlin (commented):
Looks very close at this point! One more small comment.

MuhammadHammad001:
> > The test for the verification of this support is available [here](https://github.com/MuhammadHammad001/big-endianness-test).
> 
> I just merged support for first-party tests so we could add this as part of this PR now!

Okayy, I will add this in the next commit.

model/riscv_mem.sail
```diff
@@ -73,9 +84,11 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext
     (Execute(),  None()) => Err(E_Fetch_Access_Fault()),
     (Read(Data), None()) => Err(E_Load_Access_Fault()),
     (_,          None()) => Err(E_SAMO_Access_Fault()),
-    (_,      Some(v, m)) => { if   get_config_print_mem()
-                              then print_mem("mem[" ^ to_str(t) ^ "," ^ BitStr(physaddr_bits(paddr)) ^ "] -> " ^ BitStr(v));
-                              Ok(v, m) }
+    (_,      Some(v, m)) => {
+                        let v = endianness_conversion(v, t);
```
jordancarlin, line 88: I wonder if it makes more sense to name this something like `maybe_endianness_conversion` to make it clear that it doesn't always reverse the endianness. Could be convinced either way though.

jordancarlin (approved):
LGTM once the test is added. Down the road we should add a configuration option to enable or disable big endianness support. Feel free to add a TODO comment in the code where the value of MBE is set to remind us of this.

model/riscv_mem.sail
jordancarlin: @MuhammadHammad001 Can you make this small tweak and add the test so we can get this merged?

MuhammadHammad001: Hi Jordan, I was working on this but there is one problem. When I fetched the latest changes, I am unable to build the Sail. The build process fails at 95%. Am I doing something wrong?

image

Additionally, when trying to build the first party tests, I am getting the following error:
image

Can you please help me with this so this work gets completed?

nadime15: I'm not exactly sure why it's failing, but I checked out your branch and was able to compile it successfully.

MuhammadHammad001: Just resolved it. It was because of the older version of sail. :)
But, the error related to first party tests is still not resolved.
image

model/riscv_mem.sail
```diff
@@ -63,6 +63,17 @@ function write_kind_of_flags (aq : bool, rl : bool, con : bool) -> write_kind =
...
+    match effectivePrivilege(typ, mstatus, cur_privilege) {
+      Machine     => bits_to_bool(mstatus[MBE]),
+      Supervisor  => bits_to_bool(mstatus[SBE]),
+      User        => bits_to_bool(mstatus[UBE]),
+    }
+}
+
+function endianness_conversion forall 'n, 0 < 'n <= max_mem_access . (data : bits(8 * 'n), ac : AccessType(ext_access_type)) -> bits(8 * 'n) =
+  if is_big_endian(ac) & not(ac == Execute()) then reverse_endianness(data) else data
+
```
nadime15, line 76: #757 Loads the address of the jump table, and the memory address it points to implies _execute_ permission access while following the current data endianness.

Is there a straightforward way to make an exception for this specific case?

@jordancarlin Any thoughts on that?

A simple but probably not elegant solution would be adding a new AccessType like `ExecuteJumpTable()`

model/riscv_mem.sail
jordancarlin: Hmm. Strange issue. Glad it’s working now.

pmundkur (changes requested):
`mmio_read()` and `mmio_write()` also need to be modified for endianness control.

pmundkur:
> > `mmio_read()` and `mmio_write()` also need to be modified for endianness control.
> 
> Good point. What about just modifying `checked_mem_write` and `checked_mem_read` so it doesn’t need to be implemented in two different places?

That should work, sure.

model/riscv_mem.sail
jordancarlin: Hmm. The only things I can think of are manually putting `if is_big_endian(ac) then reverse_endianness(data) else data` in the relevant instruction or adding a new AccessType (maybe something like `EndianExecute()`). Either way, this seems like a problem that will need to be resolved in #757 and I don't think it impacts the implementation here.

nadime15: Hmm yeah I see. I agree this PR should not consider this.

Timmmm: I think this is a good point actually. Upon closer reading of the spec it's really *instructions* that need their endianness flipped, not fetches. So if the endianness handling goes in `fetch()` itself then this wouldn't be an issue.

nadime15: Yes, that's true. The handling of this case depends on the specific implementation. Slightly off-topic, but for Zcmt, I avoided using `fetch()` and instead loaded the jump table address through a standard data load operation. On a 64-bit system, `fetch()` retrieves only 32 bits, while the jump table address is 64-bit. Currently, there is no mechanism to fetch the jump table address via the PC and update the PC in a single `fetch()` operation.

In fact, in my initial approach, I extended the fetch() function to support 64-bit loads. However, I found it more straightforward to use the standard load pattern with Execute permission.

Depending on the underlying architecture, particularly in a von Neumann system where instructions and data share the same memory space, the distinction between instruction loads and data loads may not even be necessary, unlike the specification implies (in some way?) the use of a dedicated instruction memory (for cm.jt and cm.jalt) which in turn implies use of `fetch()`.

If I understand correctly, `fetch()` uses the program counter (PC) to load the next instruction. However, as far as I can tell, the specification does not mention using the PC to load the jump table address.

Timmmm: I think there is a distinction between fetches and loads/stores in the architecture to allow non-coherent instruction caches (and probably other reasons). It doesn't make a difference for the Sail model of course because it has no instruction cache. It maybe should but that's another story.

Anyway I think for the Sail model the best thing is to *unconditionally* apply the endianness conversion for memory accesses, and then simply reverse the endianness again on loaded instructions (i.e. in `fetch()`).

I can make a PR or push to this one to show you what I mean if you like.

nadime15: > I think there is a distinction between fetches and loads/stores in the architecture to allow non-coherent instruction caches (and probably other reasons)

Ahh true, thats a good point, thanks!

Yes, I think it would make sense to push to this one

pmundkur: Unconditionally applying endianness conversion for memory accesses including fetches (even if reversed later) would complicate eventually including Sail code in the privileged manual, which explicitly states "Instruction fetches are always little-endian".  It might be best to apply the exception in Zcmt.

jordancarlin: I agree with @pmundkur on this one. While it might be slightly more verbose to explicitly add endianness conversion to Zcmt, I think the priority should be on readability and easy understanding of the model. Having an endianness conversion in the fetch function would likely confuse many readers.

Timmmm: Although... it would be weird to check `mstatus` etc. in `fetch()`. Maybe a new access type is the most "correct" way.

Alasdair: Instruction fetch should probably have it's own access type. In our semantics for relaxed memory behavior of self-modifying code on Armv8 you need one because ifetch accesses are much weaker than other kinds of memory accesses.

pmundkur: Isn't `Execute()` already the access type for instruction fetch?

Timmmm: The issue is that jump vector table reads *aren't* instruction fetches but they *do* require execute permission.

So we could equivalently add a `VectorTableFetch` access type. CLIC has a similar thing and... wow I just checked and it's actually consistent! It also requires execute permission but isn't an instruction fetch.

jordancarlin: Seems like the easiest solution then is to add a new access type for jump vector tables and CLIC. I don’t think anything in this PR would need to change for that since it currently only skips the endian switch for the Execute access type.

Timmmm: Yeah ok let's do that. We should maybe rename `Execute` to `InstructionFetch` too but it can wait.

jordancarlin: Issue #808 opened to keep track of this.

pmundkur:
@MuhammadHammad001  It would be good to merge this soon when the `mmio` issue is fixed and the test is added.

jordancarlin:
@MuhammadHammad001 is this something you'll be able to finish up soon? If not, I might finish it up myself in about a week if that's ok with you.

MuhammadHammad001:
Hi @jordancarlin, I will surely do it by tomorrow.

MuhammadHammad001:
@jordancarlin @Timmmm @pmundkur 
Hi, I just added the test as well and also the support for big endianness.

Can you please review the code at your ease :)))

model/riscv_mem.sail
```diff
@@ -73,9 +81,11 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext
     (Execute(),  None()) => Err(E_Fetch_Access_Fault()),
     (Read(Data), None()) => Err(E_Load_Access_Fault()),
     (_,          None()) => Err(E_SAMO_Access_Fault()),
-    (_,      Some(v, m)) => { if   get_config_print_mem()
-                              then print_mem("mem[" ^ to_str(t) ^ "," ^ BitStr(physaddr_bits(paddr)) ^ "] -> " ^ BitStr(v));
-                              Ok(v, m) }
+    (_,      Some(v, m)) => {
+                        let v = endianness_conversion(v, to_be);
```
Timmmm, line 85: Can you format it like this:

```
    (_,      Some(v, m)) => {
      let v = endianness_conversion(v, to_be);
      ..
      Ok(v, m)
    },
```

model/riscv_mem.sail
```diff
@@ -218,12 +229,12 @@ function checked_mem_write forall 'n, 0 < 'n <= max_mem_access . (
     Some(e) => Err(e),
     None() => {
```
Timmmm, line 230: Wouldn't it be better to handle the endianness conversion here? Then you don't need to handle it separately for `mmio_write` and `phys_mem_write`.

model/riscv_platform.sail
```diff
@@ -24,6 +24,9 @@ val elf_entry = pure {
   c: "elf_entry"
 } : unit -> int
 
+function endianness_conversion forall 'n, 0 < 'n <= max_mem_access . (data : bits(8 * 'n), to_be: bool) -> bits(8 * 'n) =
```
Timmmm, line 27: `to_be` is a bit too short ("to be what?"). I think `big_endian_enabled : bool` would be ok and much clearer.

Even better would be

```
enum Endianness = { BigEndian, LittleEndian }

function endianness_conversion forall 'n, 0 < 'n <= max_mem_access . (data : bits(8 * 'n), current_endianness : Endianness) -> bits(8 * 'n) =
  match current_endianness {
    LittleEndian => data,
    BigEndian => reverse_endianness(data),
  }
```

but... only if you are feeling keen.

Timmmm (changes requested):
Some minor things but it looks pretty close.

Haven't read the test but it sounds like it found some issues so good work there!

model/riscv_mem.sail
```diff
@@ -102,10 +112,10 @@ function checked_mem_read forall 'n, 0 < 'n <= max_mem_access . (
...
       if   within_mmio_readable(paddr, width)
-      then MemoryOpResult_add_meta(mmio_read(t, paddr, width), default_meta)
+      then MemoryOpResult_add_meta(mmio_read(t, paddr, width, is_big_endian(t) & not(t == Execute())), default_meta)
```
pmundkur, line 115: Wouldn't it have been easier to just handle endianness in `checked_mem_read/write` as  suggested here:
https://github.com/riscv/sail-riscv/pull/751#issuecomment-2737698003

That way I think you wouldn't have needed to touch `riscv_platform.sail`

model/prelude.sail
```diff
@@ -238,3 +238,6 @@ type max_mem_access : Int = 4096
 
 // Type used for memory access widths. Zero byte accesses are not allowed.
 type mem_access_width = range(1, max_mem_access)
+
+// Function to reverse endianness.
+val reverse_endianness = pure {c: "reverse_endianness"} : forall 'n . bits(8 * 'n) -> bits(8 * 'n)
```
pmundkur, line 243: Can't we just have 
```
$include <reverse_endianness.sail>
```
at the top of the file instead of this?

pmundkur (changes requested):
Thanks for adding the test; haven't checked it yet.

model/prelude.sail
jordancarlin: See https://github.com/riscv/sail-riscv/pull/751#discussion_r1961301663.

Timmmm: We should probably update the type in reverse_endianness.sail and switch to that eventually...

model/riscv_platform.sail
MuhammadHammad001: I actually meant to_be = to_*b*ig_*e*ndian 
But, I will change it to your suggested variable.

model/prelude.sail
pmundkur: @MuhammadHammad001 Could you add a comment about the type signature compatibility issue above this declaration, to avoid this question being raised repeatedly?

model/riscv_mem.sail
MuhammadHammad001: For write, yes, I have updated it. But, for read, I don't know how we can handle it here as we print the data in the phys_mem_read functions, so we can get the data from those functions and reverse them here but in the log we will have a different value....

Or am I mis-understanding this?

model/riscv_mem.sail
```diff
@@ -63,8 +63,16 @@ function write_kind_of_flags (aq : bool, rl : bool, con : bool) -> write_kind =
     (true,  false, true)  => throw(Error_not_implemented("sc.aq"))
   }
 
+function is_big_endian(typ : AccessType(ext_access_type)) -> bool = {
```
jordancarlin, line 66: Why not have this be something like `get_current_endianness` that returns a type from the Endianness enum and push the check for an `Execute` access into it? So it becomes something like:

```
function get_current_endianness(typ : AccessType(ext_access_type)) -> Endianness = {
     match (typ, effectivePrivilege(typ, mstatus, cur_privilege)) {
       (Execute(), _)   => LittleEndian
       (_, Machine)     => if bits_to_bool(mstatus[MBE]) then BigEndian else LittleEndian,
       (_, Supervisor)  => if bits_to_bool(mstatus[SBE]) then BigEndian else LittleEndian,
       (_, User)        => if bits_to_bool(mstatus[UBE]) then BigEndian else LittleEndian,
     }
 }
```

Then you can replace all of the uses of `let current_endianness: Endianness = if is_big_endian(typ) & typ != Execute() then BigEndian else LittleEndian` with just `get_current_endianness(typ)`.

MuhammadHammad001: Good idea! I will implement this

model/riscv_mem.sail
Timmmm: I think it's better just to always print them little endian in the log. The code will be a lot simpler, it's debatable which way it *should* be printed, and I think virtually nobody will be using big endian anyway so it probably doesn't matter too much.

MuhammadHammad001:
Hi @Timmmm Any update on merging this PR? Is there anybody working on the PR #467 ?

jordancarlin:
> Any update on merging this PR? Is there anybody working on the PR #467 ?

@Alasdair and @pmundkur are working on #467 and it should hopefully be updated soon.

jordancarlin:
@MuhammadHammad001 Thanks for your patience with this one. #861 (which replaced #467) is merged, so it should be possible to update this PR now to move the endianness conversion into the new functions that it introduced.

model/prelude.sail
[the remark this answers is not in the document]
jordancarlin: @Alasdair I know you're planning on a new version of Sail soon. Any chance the `reverse_endianness` library could be updated so that it can be used here?

MuhammadHammad001:
Hi, @jordancarlin @pmundkur @Timmmm , can you please review this and let me know if I need to change anything?

Also, I have removed the testcase for the `mtime` since the value is changed before the check causing it to fail. But, `mtimecmp` and `htif` testcases are present to make sure the endinanness is supported for the MMIOs.

jordancarlin (commented):
Maybe I'm missing something, but I though the whole point of waiting for #861 was to move the endianness handling into `vmem_read`/`vmem_write` so that it didn't need to be duplicated everywhere. Right now it looks like it's still being handled in `mem_write_value_priv_meta` and the individual mmio functions.

MuhammadHammad001:
> Maybe I'm missing something, but I though the whole point of waiting for https://github.com/riscv/sail-riscv/pull/861 was to move the endianness handling into vmem_read/vmem_write so that it didn't need to be duplicated everywhere. Right now it looks like it's still being handled in mem_write_value_priv_meta and the individual mmio functions.

I may be mis-understanding but as I mentioned earlier, I can surely do this for `vmem_write` but for `vmem_read`, I am unable to understand, how this will be possible because in case of a read, we print the data where we read that data. For instance, in case of a mmio read, the following part of the `clint_load` code:

```
val clint_load : forall 'n, 'n > 0. (AccessType(ext_access_type), physaddr, int('n), Endianness) -> MemoryOpResult(bits(8 * 'n))
function clint_load(t, Physaddr(addr), width, current_endianness) = {
  //these two registers are 64 bits, so simple conversion is not accurate.
  //consider 0x78563412 stored at mtime/cmp and when reading in big endian format,
  //we change it to 0x1234567800000000. So position for mtime and mtime_hi are reversed. (not for ld)
  var converted_mtime    = endianness_conversion(mtime, current_endianness);
  var converted_mtimecmp = endianness_conversion(mtimecmp, current_endianness);
  if ((current_endianness == BigEndian) & 'n == 4) then {
                converted_mtime    = converted_mtime[31..0] @ converted_mtime[63..32];
                converted_mtimecmp = converted_mtimecmp[31..0] @ converted_mtimecmp[63..32];
                };
  let addr = addr - plat_clint_base ();
  /* FIXME: For now, only allow exact aligned access. */
  if addr == MSIP_BASE & ('n == 8 | 'n == 4)
  then {
    then print_platform("clint[" ^ BitStr(addr) ^ "] -> " ^ BitStr(mip[MSI]));
    Ok(zero_extend(sizeof(8 * 'n), mip[MSI]))
  }
  else if addr == MTIMECMP_BASE & ('n == 4)
  then {
    then print_platform("clint<4>[" ^ BitStr(addr) ^ "] -> " ^ BitStr(converted_mtimecmp[31..0]));
    /* FIXME: Redundant zero_extend currently required by Lem backend */
    Ok(zero_extend(32, converted_mtimecmp[31..0]))

```
As you can see, the 

```
then print_platform("clint<4>[" ^ BitStr(addr) ^ "] -> " ^ BitStr(converted_mtimecmp[31..0]));
```
is printing the data here instead in the `vmem_read` function. Now, even if the `vmem_read` function gets the data reversed, in the log file, we will see the actual (non-reversed) data.

Note: I am referring reversed == Big endian


Can you please help me in clearing the doubt?

MuhammadHammad001:
But, I am converting the data to be written inside the  `vmem_write` function, so even if it's called seperately for amo instructions, this will cover that as well. Am I right?

pmundkur:
> But, I am converting the data to be written inside the `vmem_write` function, so even if it's called seperately for amo instructions, this will cover that as well. Am I right?

AMO instructions are not using `vmem_{read,write}` functions at all.  So any changes in those functions will not be seen by AMOs.

MuhammadHammad001:
Sorry @pmundkur, I may be mis-understanding, I am currently looking at the file `riscv_instr_aext.sail` and here, it is using the function `vmem_write`
image


Can you please give a pointer to exact function you want me to change?

pmundkur:
> Can you please give a pointer to exact function you want me to change?

There is a `mem_read` and `mem_write_value` in `execute(AMO(...))` in that file that will need handling.

PR #752, opened 2025-02-20
Add support for Zvkned.

KotorinMinami:
Implements the Zvkned (NIST Suite: Vector AES Block Cipher) extension, as of version [Release 1.0.0](https://github.com/riscv/riscv-crypto/releases/tag/v1.0.0).

- vaesef.[vv,vs]
- vaesem.[vv,vs]
- vaesdf.[vv,vs]
- vaesdm.[vv,vs]
- vaeskf1.vi
- vaeskf2.vi
- vaesz.vs

All instructions were tested with VLEN=128 and results were compared with QEMU results of each instruction.

Note
This PR was originally developed by @charmitro , and now the Vector AES Block Cipher has been released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ For booting operating system images, see the information under the
 - Zvbb extension for vector basic bit-manipulation, v1.0
 - Zvbc extension for vector carryless multiplication, v1.0
 - Zvkb extension for vector cryptography bit-manipulation, v1.0
+- Zvkned extension for vector cryptography NIST Suite: Vector AES Block Cipher, v1.0
 - Zvknha and Zvknhb extensions for vector cryptography NIST Suite: Vector SHA-2 Secure Hash, v1.0
 - Zvksh extension for vector cryptography ShangMi Suite: SM3 Secure Hash, v1.0
 - Machine, Supervisor, and User modes
--- a/config/default.json
+++ b/config/default.json
@@ -190,11 +190,11 @@
     "Zvbc": {
       "supported": true
     },
-    "Zvknha": {
+    "Zvkned": {
       "supported": true
     },
-    "Zvkned" : {
-      "supported" : true
+    "Zvknha": {
+      "supported": true
     },
     "Zvknhb": {
       "supported": true
--- a/config/default.json
+++ b/config/default.json
@@ -193,6 +193,9 @@
     "Zvknha": {
       "supported": true
     },
+    "Zvkned" : {
+      "supported" : true
+    },
     "Zvknhb": {
       "supported": true
     },
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -185,6 +185,9 @@ function clause hartSupports(Ext_Zvkb) = config extensions.Zvkb.supported
 // Vector Carryless Multiplication
 enum clause extension = Ext_Zvbc
 function clause hartSupports(Ext_Zvbc) = config extensions.Zvbc.supported
+// NIST Suite: Vector AES Block Cipher
+enum clause extension = Ext_Zvkned
+function clause hartSupports(Ext_Zvkned) = config extensions.Zvkned.supported
 // NIST Suite: Vector SHA-2 Secure Hash
 enum clause extension = Ext_Zvknha
 function clause hartSupports(Ext_Zvknha) = config extensions.Zvknha.supported
@@ -194,7 +197,6 @@ function clause hartSupports(Ext_Zvknhb) = config extensions.Zvknhb.supported
 // ShangMi Suite: SM3 Secure Hash
 enum clause extension = Ext_Zvksh
 function clause hartSupports(Ext_Zvksh) = config extensions.Zvksh.supported
-
 
 // Count Overflow and Mode-Based Filtering
 enum clause extension = Ext_Sscofpmf
--- a/model/riscv_insts_zvkned.sail
+++ b/model/riscv_insts_zvkned.sail
@@ -1,0 +1,379 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+/*
+ * Vector Cryptography Extension - NIST Suite: Vector AES Block Cipher
+ * ----------------------------------------------------------------------
+ */
+
+function clause currentlyEnabled(Ext_Zvkned) = true
+
+val get_velem : forall 'n 'm 'p, 'n > 0 & 'm > 0 & 'p >= 0 & 4 * 'p + 3 < 'n . (vector('n, bits('m)), int('p)) -> bits(4 * 'm)
+function get_velem(v, i) = v[4 * i + 3] @ v[4 * i + 2] @ v[4 * i + 1] @ v[4 * i]
+
+val set_velem : forall 'n 'm 'p, 'n > 0 & 'm > 0 & 'p >= 0 & 4 * 'p + 3 < 'n . (vector('n, bits('m)), bits(4 * 'm), int('p)) -> vector('n, bits('m))
+function set_velem(s, v, i) = {
+  let m = length(v) / 4;
+  [s with
+    (4 * i) = v[m - 1 .. 0],
+    (4 * i + 1) = v[2 * m - 1 .. m],
+    (4 * i + 2) = v[3 * m - 1 .. 2 * m],
+    (4 * i + 3) = v[4 * m - 1 .. 3 * m],
+  ]
+}
+
+function aes_rotword(x : bits(32)) -> bits(32) = x[7.. 0] @ x[31..24] @ x[23..16] @ x[15.. 8]
+
+function zvk_check_elements(VLEN : int, num_elem : int, LMUL : int, SEW : int) -> bool = {
+  ((unsigned(vl)%num_elem) != 0) | ((unsigned(vstart)%num_elem) != 0) | (LMUL*VLEN) < (num_elem*SEW)
+}
+
+function zvk_check_encdec(EGW : int, EGS : int) -> bool = {
+  let LMUL_pow = get_lmul_pow();
+
+  (unsigned(vl) % EGS == 0) & (unsigned(vstart) % EGS == 0) & (2 ^ LMUL_pow * VLEN >= 128)
+}
+
+val vv_or_vs_index : forall 'n , 0 <= 'n . (suffix_kind , int('n)) -> range(0, 'n)
+function vv_or_vs_index(suffix, i) = {
+  match suffix {
+    Vector_Vector => i,
+    Vector_Scalar => 0,
+  }
+}
+
+mapping vv_or_vs : suffix_kind <-> bits(6) = {
+  Vector_Vector <-> 0b101000,
+  Vector_Scalar <-> 0b101001,
+}
+
+mapping vv_or_vs_mnemonic : suffix_kind <-> string = {
+  Vector_Vector <-> "vv",
+  Vector_Scalar <-> "vs",
+}
+
+/* VAESEF.[VV, VS] */
+
+union clause ast = RISCV_VAESEF : (vregidx, vregidx, suffix_kind)
+
+mapping clause encdec = RISCV_VAESEF(vs2, vd, suffix)
+  <-> vv_or_vs(suffix) @ 0b1 @ encdec_vreg(vs2) @ 0b00011 @ 0b010 @ encdec_vreg(vd) @ 0b1110111
+  when currentlyEnabled(Ext_Zvkned) & get_sew() == 32 & zvk_check_encdec(128, 4)
+
+mapping clause assembly = RISCV_VAESEF(vs2, vd, suffix)
+  <-> "vaesef." ^ vv_or_vs_mnemonic(suffix) ^ spc() ^ vreg_name(vd)
+                                            ^ spc() ^ vreg_name(vs2)
+
+function clause execute (RISCV_VAESEF(vs2, vd, suffix)) = {
+  let 'SEW = get_sew();
+  let LMUL_pow = get_lmul_pow();
+  let 'num_elem = get_num_elem(LMUL_pow, SEW);
+
+  assert(SEW == 32);
+  let vs2_val = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+  let vd_val = read_vreg(num_elem, SEW, LMUL_pow, vd);
+  var result : vector('num_elem, bits('SEW)) = vector_init(zeros());
+
+  let eg_len = (unsigned(vl) / 4);
+  let eg_start = (unsigned(vstart) / 4);
+
+  foreach (i from eg_start to (eg_len - 1)) {
+    assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < num_elem);
+
+    let vd_state = get_velem(vd_val, i);
+    let vs2_key  = get_velem(vs2_val, vv_or_vs_index(suffix, i));
+    let sb = aes_subbytes_fwd(vd_state);
+    let sr = aes_shift_rows_fwd(sb);
+    let ark = sr ^ vs2_key;
+
+    result = set_velem(result, ark, i);
+  };
+
+  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
+  set_vstart(zeros());
+  RETIRE_SUCCESS
+}
+
+/* VAESEM.[VV, VS] */
+
+union clause ast = RISCV_VAESEM : (vregidx, vregidx, suffix_kind)
+
+mapping clause encdec = RISCV_VAESEM(vs2, vd, suffix)
+  <-> vv_or_vs(suffix) @ 0b1 @ encdec_vreg(vs2) @ 0b00010 @ 0b010 @ encdec_vreg(vd) @ 0b1110111
+  when currentlyEnabled(Ext_Zvkned) & get_sew() == 32 & zvk_check_encdec(128, 4)
+
+mapping clause assembly = RISCV_VAESEM(vs2, vd, suffix)
+  <-> "vaesem." ^ vv_or_vs_mnemonic(suffix) ^ spc() ^ vreg_name(vd)
+                                            ^ sep() ^ vreg_name(vs2)
+
+function clause execute (RISCV_VAESEM(vs2, vd, suffix)) = {
+  let 'SEW = get_sew();
+  let LMUL_pow = get_lmul_pow();
+  let 'num_elem = get_num_elem(LMUL_pow, SEW);
+
+  assert(SEW == 32);
+  let vs2_val = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+  let vd_val = read_vreg(num_elem, SEW, LMUL_pow, vd);
+  var result : vector('num_elem, bits('SEW)) = vector_init(zeros());
+
+  let eg_len = (unsigned(vl) / 4);
+  let eg_start = (unsigned(vstart) / 4);
+
+  foreach (i from eg_start to (eg_len - 1)) {
+    assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < num_elem);
+
+    let vd_state = get_velem(vd_val, i);
+    let vs2_key  = get_velem(vs2_val, vv_or_vs_index(suffix, i));
+    let sb : bits(128) = aes_subbytes_fwd(vd_state);
+    let sr : bits(128) = aes_shift_rows_fwd(sb);
+    let mix : bits(128) = aes_mixcolumns_fwd(sr);
+    let ark : bits(128) = mix ^ vs2_key;
+
+    result = set_velem(result, ark, i);
+  };
+
+  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
+  set_vstart(zeros());
+  RETIRE_SUCCESS
+}
+
+/* VAESDF.VV */
+
+union clause ast = RISCV_VAESDF : (vregidx, vregidx, suffix_kind)
+
+mapping clause encdec = RISCV_VAESDF(vs2, vd, suffix)
+  <-> vv_or_vs(suffix) @ 0b1 @ encdec_vreg(vs2) @ 0b00001 @ 0b010 @ encdec_vreg(vd) @ 0b1110111
+  when currentlyEnabled(Ext_Zvkned) & get_sew() == 32 & zvk_check_encdec(128, 4)
+
+mapping clause assembly = RISCV_VAESDF(vs2, vd, suffix)
+  <-> "vaesdf." ^ vv_or_vs_mnemonic(suffix) ^ sep() ^ vreg_name(vd)
+                                            ^ sep() ^ vreg_name(vs2)
+
+function clause execute (RISCV_VAESDF(vs2, vd, suffix)) = {
+  let 'SEW = get_sew();
+  let LMUL_pow = get_lmul_pow();
+  let 'num_elem = get_num_elem(LMUL_pow, SEW);
+
+  assert(SEW == 32);
+  let vs2_val = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+  let vd_val = read_vreg(num_elem, SEW, LMUL_pow, vd);
+  var result : vector('num_elem, bits('SEW)) = vector_init(zeros());
+
+  let eg_len = (unsigned(vl) / 4);
+  let eg_start = (unsigned(vstart) / 4);
+
+  foreach (i from eg_start to (eg_len - 1)) {
+    assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < num_elem);
+
+    let vd_state = get_velem(vd_val, i);
+    let vs2_key  = get_velem(vs2_val, vv_or_vs_index(suffix, i));
+    let sr : bits(128) = aes_shift_rows_inv(vd_state);
+    let sb : bits(128) = aes_subbytes_inv(sr);
+    let ark : bits(128) = sb ^ vs2_key;
+
+    result = set_velem(result, ark, i);
+  };
+
+  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
+  set_vstart(zeros());
+  RETIRE_SUCCESS
+}
+
+/* VAESDM.VV */
+
+union clause ast = RISCV_VAESDM : (vregidx, vregidx, suffix_kind)
+
+mapping clause encdec = RISCV_VAESDM(vs2, vd, suffix)
+  <-> vv_or_vs(suffix) @ 0b1 @ encdec_vreg(vs2) @ 0b00000 @ 0b010 @ encdec_vreg(vd) @ 0b1110111
+  when currentlyEnabled(Ext_Zvkned) & get_sew() == 32 & zvk_check_encdec(128, 4)
+
+mapping clause assembly = RISCV_VAESDM(vs2, vd, suffix)
+  <-> "vaesdm." ^ vv_or_vs_mnemonic(suffix) ^ spc() ^ vreg_name(vd)
+                                            ^ sep() ^ vreg_name(vs2)
+
+function clause execute (RISCV_VAESDM(vs2, vd, suffix)) = {
+  let 'SEW = get_sew();
+  let LMUL_pow = get_lmul_pow();
+  let 'num_elem = get_num_elem(LMUL_pow, SEW);
+
+  assert(SEW == 32);
+  let vs2_val = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+  let vd_val = read_vreg(num_elem, SEW, LMUL_pow, vd);
+  var result : vector('num_elem, bits('SEW)) = vector_init(zeros());
+
+  let eg_len = (unsigned(vl) / 4);
+  let eg_start = (unsigned(vstart) / 4);
+
+  foreach (i from eg_start to (eg_len - 1)) {
+    assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < num_elem);
+
+    let vd_state = get_velem(vd_val, i);
+    let vs2_key  = get_velem(vs2_val, vv_or_vs_index(suffix, i));
+    let sr : bits(128) = aes_shift_rows_inv(vd_state);
+    let sb : bits(128) = aes_subbytes_inv(sr);
+    let ark : bits(128) = sb ^ vs2_key;
+    let mix : bits(128) = aes_mixcolumns_inv(ark);
+
+    result = set_velem(result, mix, i);
+  };
+
+  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
+  set_vstart(zeros());
+  RETIRE_SUCCESS
+}
+
+/* VAESKF1.VI */
+
+union clause ast = RISCV_VAESKF1_VI : (vregidx, bits(5), vregidx)
+
+mapping clause encdec = RISCV_VAESKF1_VI(vs2, rnd, vd)
+  <-> 0b1000101 @ encdec_vreg(vs2) @ rnd @ 0b010 @ encdec_vreg(vd) @ 0b1110111
+  when currentlyEnabled(Ext_Zvkned) & get_sew() == 32 & zvk_check_encdec(128, 4)
+
+mapping clause assembly = RISCV_VAESKF1_VI(vs2, rnd, vd)
+  <-> "vaeskf1.vi" ^ sep() ^ vreg_name(vd)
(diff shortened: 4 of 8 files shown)


model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+ <-> vaesef_mnemonic(vv_or_vs(suffix)) ^ spc() ^ vreg_name(vd)
+                                       ^ spc() ^ vreg_name(vs2)
+
+function clause execute (RISCV_VAESEF(vs2, vd, suffix)) = {
+  let SEW      = get_sew();
+  let LMUL_pow = get_lmul_pow();
+  let LMUL     = if LMUL_pow < 0 then 0 else LMUL_pow;
+  let num_elem = get_num_elem(LMUL_pow, SEW);
+
+  if (zvk_check_elements(VLEN, num_elem, LMUL, SEW) == false)
```
Timmmm, line 57:
```suggestion
  if not(zvk_check_elements(VLEN, num_elem, LMUL, SEW))
```

(And in the other places.)

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+}
+
+mapping vv_or_vs : string <-> bits(7) = {
+  "vv" <-> 0b1010001,
+  "vs" <-> 0b1010011,
+}
+
+val aes_rotword : bits(32) -> bits(32)
+function aes_rotword(x) = {
+  let a0 : bits (8) = x[ 7.. 0];
```
Timmmm, line 28: I think I would just simplify this to

```
function aes_rotword(x) = x[7 .. 0] @ x[15 .. 8] @ x[23 .. 16] @ x[31 .. 24]
```

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+  let num_elem = get_num_elem(LMUL_pow, SEW);
+
+  if (zvk_check_elements(VLEN, num_elem, LMUL, SEW) == false)
+  then {
+    handle_illegal();
+    RETIRE_FAIL
+  } else {
+    let 'n = num_elem;
+    let 'm = SEW;
+    assert('m == 32);
```
Timmmm, line 64: Is this necessarily true? Definitely worth a comment explaining why at least.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+
+    let vs2_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+    let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
+
+    let eg_len = (unsigned(vl) / 'n);
+    let eg_start = (unsigned(vstart) / 'n);
+    let keyelem = if suffix == "vv" then 1 else 0;
```
Timmmm, line 71: I think we should avoid having behaviour encoded in strings. `suffix` should be an enum, not strings.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+
+    foreach (i from eg_start to (eg_len - 1)) {
+      assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < 'n);
+
+      let vd_state = get_velem(vd_val, i);
+      let vs2_key = get_velem(vs2_val, keyelem * i);
```
Timmmm, line 77: So this would be something like:

```
let vs2_key = get_velem(vs2_val, match suffix { something_VV => i, something_VS => 0 });
```

(I'll leave you to come up with suitable names for whatever VV and VS mean!)

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+      let sb       : bitvector(128) = aes_subbytes_fwd(vd_state);
+      let sr       : bitvector(128) = aes_shift_rows_fwd(sb);
+      let ark      : bitvector(128) = sr ^ vs2_key;
```
Timmmm, line 80: `bits`? Maybe before my time; is `bitvector` an old version of `bits`?

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+
+    foreach (i from eg_start to (eg_len - 1)) {
+      assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < 'n);
+
+      let vd_state = get_velem(vd_val, i);
+      let vs2_key  = get_velem(vs2_val, keyelem * i);
+      let sb       : bitvector(128, dec) = aes_subbytes_fwd(vd_state);
+      let sr       : bitvector(128, dec) = aes_shift_rows_fwd(sb);
+      let mix      : bitvector(128, dec) = aes_mixcolumns_fwd(sr);
+      let ark      : bitvector(128, dec) = mix ^ vs2_key;
```
Timmmm, line 136: `bits(128)`. You could just delete all these, though it may be appropriate for readability.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+  then {
+    handle_illegal();
+    RETIRE_FAIL
+  } else {
+    let 'n = num_elem;
+    let 'm = SEW;
+    assert('m == 32);
+
+    let vs2_val : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+    let vd_val  : vector('n, dec, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
```
Timmmm, line 122: Remove `, dec` - Sail doesn't support per-vector orders now.

You could just delete the type annotations (I would, but up to you).

Timmmm (commented):
A few random low-level Sail comments (I know nothing about the actual spec).

Also how was it tested against QEMU exactly? It would be nice to integrate tests like that (in future; we don't have the infrastructure for it yet).

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+    RETIRE_SUCCESS
+  }
+}
+
+/* VAESDM.VV */
+
+mapping vaesdm_mnemonic : bits(7) <-> string = {
+  0b1010001 <-> "vaesem.vv",
+  0b1010011 <-> "vaesem.vs",
+}
```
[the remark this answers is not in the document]
Timmmm: Also the rest of the model maps `ast` to assembly; not the bits directly.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+ * ----------------------------------------------------------------------
+ */
+
+function clause extensionEnabled(Ext_Zvkned) = true
+
+val get_velem : forall 'n 'm 'p, 'n > 0 & 'm > 0 & 'p >= 0 & 4 * 'p + 3 < 'n . (vector('n, bits('m)), int('p)) -> bits(4 * 'm)
+function get_velem(v, i) = v[4 * i + 3] @ v[4 * i + 2] @ v[4 * i + 1] @ v[4 * i]
+
+val      zvk_check_elements : (int, int, int, int) -> bool
+function zvk_check_elements(VLEN, num_elem, LMUL, SEW) = {
```
jordancarlin, line 17: With simple function declarations like this, it’s usually easier to eliminate the val line and directly integrate the types into the function line.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
+/*
```
jordancarlin, line 1: Missing copyright header

model/riscv_extensions.sail
```diff
@@ -23,6 +23,7 @@ enum clause extension = Ext_C
 enum clause extension = Ext_B
 // Vector Operations
 enum clause extension = Ext_V
+enum clause extension = Ext_Zvkned
```
jordancarlin, line 26: Missing comment with long form name of the extension. This should also be moved down to be with the rest of the Z* extensions in a new Zv* section.

jordancarlin (commented):
Still need to take a look at the spec, but here are some initial comments.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,428 @@
...
+
+    let eg_len = (unsigned(vl) / 'n);
+    let eg_start = (unsigned(vstart) / 'n);
+
+    foreach (i from eg_start to (eg_len - 1)) {
+      assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < 'n);
+
+      let current_round_key = get_velem(vs2_val, i);
+
+      let round_key_b = get_velem(vs2_val, i);
```
charmitro, line 361: Specification states:
```
  let CurrentRoundKey[3:0] : bits(128) = get_velem(vs2, EGW=128, i);
  let RoundKeyB[3:0] : bits(32) = get_velem(vd, EGW=128, i); // Previous round key
```

model/riscv_insts_zvkned.sail
KotorinMinami: the spec declared that it should be 32 bits, and without that, the Sail compiler will chuck an error, because it won't know how many bits `'m` has.

jordancarlin: I haven't looked at the spec closely enough yet, but if it is reserved it seems like it would be best to catch it at the encdec stage if possible and just not decode it as this instruction. Then it would fall through to the existing illegal instruction catch-all.

KotorinMinami: Yes, but then bits('m) would need to be bits(32) for Sail to prove it.

Timmmm: You could do something like this:

```
  let 'SEW      = get_sew();
  let LMUL     = if LMUL_pow < 0 then 0 else LMUL_pow;
  let 'num_elem = get_num_elem(LMUL_pow, SEW);

  if 'SEW != 32 | not(zvk_check_elements(VLEN, num_elem, LMUL, SEW)) then {
    handle_illegal();
    RETIRE_FAIL
  } else {
    let vs2_val = read_vreg(num_elem, SEW, LMUL_pow, vs2);
    let vd_val  = read_vreg(num_elem, SEW, LMUL_pow, vd);

    let eg_len = (unsigned(vl) / 'num_elem);
   ...
```

I *think* that will compile. Not 100% sure though.

KotorinMinami: That works, thank you!

KotorinMinami:
I think the issues raised by the review have been resolved so far. @Timmmm @jordancarlin  Is there anything else wrong with my code?

jordancarlin:
@KotorinMinami Sorry about the delay in reviewing this. I think it needs some modifications because of the recent type refactoring for registers. Once it is rebased I'll give it another look. Would be good to get this one in soon for RVA23.

KotorinMinami:
I think we can move forward with this. @jordancarlin

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,390 @@
...
+                  ^ sep() ^ vreg_name(vs2)
+                  ^ sep() ^ hex_bits_5(rnd)
+
+function clause execute (RISCV_VAESKF1_VI(vs2, rnd, vd)) = {
+  let 'SEW = get_sew();
+  let LMUL_pow = get_lmul_pow();
+  let LMUL = if LMUL_pow < 0 then 0 else LMUL_pow;
+  let 'num_elem = get_num_elem(LMUL_pow, SEW);
+
+  if 'SEW != 32 | (not(zvk_check_elements(VLEN, num_elem, LMUL, SEW)))
```
nadime15, line 247: Do you still need that? Its part of the `encdec` clause already (`get_sew() == 32`)

jordancarlin: What about even moving the `zvk_check_elements` check into the encdec clause?

KotorinMinami: Need. The Sail compiler cannot obtain specific SEW information here (even if proven in encdec). Without specifying SEW, it will fail type checking.

jordancarlin: If it's just to make the Sail compiler happy then how about using an assert?

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,399 @@
...
+    foreach (i from eg_start to (eg_len - 1)) {
+      assert(0 <= ((i * 4) + 3) & ((i * 4) + 3) < num_elem);
+
+      let vd_state = get_velem(vd_val, i);
+      let vs2_key  = get_velem(vs2_val, vv_or_vs_index(suffix, i));
+      let sb = aes_subbytes_fwd(vd_state);
+      let sr = aes_shift_rows_fwd(sb);
+      let ark = sr ^ vs2_key;
+
+      write_single_element(128, i, vd, ark);
```
nadime15, line 89: Writing an element of size 128 is not possible. This will cause a runtime assertion failure.

`write_single_element() `invokes `read_single_vreg()`, which contains the assertion:
`assert(8 <= SEW && SEW <= 64);`

KotorinMinami: So, why does `write_single_element` even allow 128-bit EEW? If it's going to trigger an assertion, the function definition shouldn't permit a 128-bit EEW (Element Effective Width) in the first place. If that's the case, then we'll just have to perform two 64-bit reads.
>val write_single_element : forall 'm 'x, 8 <= 'm <= 128. (int('m), int('x), vregidx, bits('m)) -> unit

pmundkur: Good point.  It looks like some of the type signatures on these helpers are insufficiently precise.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,383 @@
...
+}
+
+function zvk_check_encdec() -> bool = {
+  let SEW = get_sew();
+  let LMUL_pow = get_lmul_pow();
+  let LMUL = if LMUL_pow < 0 then 0 else LMUL_pow;
+  let num_elem = get_num_elem(LMUL_pow, SEW);
+
+  zvk_check_elements(VLEN, num_elem, LMUL, SEW)
+}
```
nadime15, line 42: I really the idea of `zvk_check_encdec`! We can use this for the other Zvk* extensions too!

Personally, I prefer something like:

```suggestion

function zvk_check_encdec() -> bool = {
  let SEW = get_sew();
  let LMUL_pow = get_lmul_pow();

  (unsigned(vl) % 4 == 0) & (unsigned(vstart) % 4 == 0) | (2 ^ get_lmul_pow() * VLEN) >= 128)
}
```
The spec sets `EGW=128` and `EGS=4`

The spec also says that `vl` and `vstart` must be multiples of 4 (=EGS), not `num_elem`. This means we could have `vl` = 8 and `num_elem` = 8, which is valid, but the current check (`vl % num_elem != 0`) would incorrectly be false when it should be true.

Additionally, the code currently takes the exponent of the current `LMUL` (`get_lmul_pow()`) but ignores negative exponents (!). Right now, if you check `get_lmul_pow()`, the condition `(LMUL*VLEN) < (num_elem*SEW) `effectively does:

`LMUL_pow * VLEN < (2 ^ LMUL_pow * VLEN / SEW) * SEW`

which simplifies to  `LMUL_pow * VLEN < (2 ^ LMUL_pow) ` Which is always true (for now). 

What we really want is

`2 * get_lmul_pow() * VLEN >= 128` (I assume that sail can deal with negative exponents)

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,383 @@
...
+  let LMUL_pow = get_lmul_pow();
+  let 'num_elem = get_num_elem(LMUL_pow, SEW);
+
+  assert(SEW == 32);
+  let vs2_val = read_vreg(num_elem, SEW, LMUL_pow, vs2);
+  let vd_val = read_vreg(num_elem, SEW, LMUL_pow, vd);
+  var result : vector('num_elem, bits('SEW)) = vector_init(zeros());
+
+  let eg_len = (unsigned(vl) / num_elem);
+  let eg_start = (unsigned(vstart) / num_elem);
```
nadime15, line 85: Same for here with EGS = 4:
```suggestion
  let eg_len = (unsigned(vl) / 4);
  let eg_start = (unsigned(vstart) / 4;
```
I ran into cases where `vl` = 4 and `vstart` = 0, but `num_elem` = 8 (which is valid). However, with the old code, we would incorrectly skip the following loop because `eg_len` would be 0, even though it should be 1.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,383 @@
...
+val set_velem : forall 'n 'm 'p, 'n > 0 & 'm > 0 & 'p >= 0 & 4 * 'p + 3 < 'n . (vector('n, bits('m)), bits(4 * 'm), int('p)) -> vector('n, bits('m))
+function set_velem(s, v, i) = {
+  let m = length(v) / 4;
+  [s with
+    (4 * i) = v[m - 1 .. 0],
+    (4 * i + 1) = v[2 * m - 1 .. m],
+    (4 * i + 2) = v[3 * m - 1 .. 2 * m],
+    (4 * i + 3) = v[4 * m - 1 .. 3 * m],
+  ]
+}
```
nadime15, line 27: For #819, I added my own version of the `set_vlem` function, but I actually prefer yours! One suggestion: instead of returning s, how about we write directly to vd within the function and then call write_vreg? Since it's a setter method, it might make more sense to do it that way and we would follow the spec.

KotorinMinami: However, there's a problem: this approach results in multiple calls to `write_vreg`, subsequently leading to redundant calls to `write_single_vreg`, making it less efficient.

nadime15: Yes you are right its not really efficient but Sail's primary focus is readability.

Currently, Sail does not provide a function to write `n` continuous elements starting from a specific position in a given vreg. The available functions for writing to a vreg are:

* `write_vreg()` 
* `write_single_vreg()`
* `write_single_element()`

The implementation I have is:

```
val set_velem : forall 'p 'n 'm, 8 <= 'n <= 64 & 'm > 0 & 'p >= 0. (vregidx, int('n), bits('m), int('p)) -> unit
function set_velem(vd, SEW, input, i) = {
  foreach(j from 0 to 3)
    write_single_element(SEW, 4 * i + j, vd, slice(input, j * SEW, SEW));
}
```

This function enables direct writing (setting) of multiple continuous elements to a vreg without the need to return the value vector and to store it at some pointer later via `write_vreg() `. Additionally, it could be further generalized to allow writing `n` continuous elements instead of just four and it follows more the spec.

It might be helpful to get feedback from others as well. @pmundkur @jordancarlin

KotorinMinami: Maybe we can add function like `write_element_group` ?

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,381 @@
...
+function zvk_check_elements(VLEN : int, num_elem : int, LMUL : int, SEW : int) -> bool = {
+  ((unsigned(vl)%num_elem) != 0) | ((unsigned(vstart)%num_elem) != 0) | (LMUL*VLEN) < (num_elem*SEW)
+}
+
+function zvk_check_encdec(EGW : int, EGS : int) -> bool = {
+  let LMUL_pow = get_lmul_pow();
+  let LMUL = if LMUL_pow < 0 then 0 else LMUL_pow;
+
+  (unsigned(vl) % EGS == 0) & (unsigned(vstart) % EGS == 0) & (2 ^ LMUL * VLEN >= 128)
+}
```
nadime15, line 40: We more or less agreed on the same function but you use

`2 ^ LMUL * VLEN >= 128`

instead of 

`2 ^ LMUL_pow * VLEN >= 128`

This essentially suggests that LMUL = 0.5 (mf2), 0.25 (mf4), 0.125 (mf8) cannot be used with this extension, which is incorrect (I still assume that Sail can handle negative exponents).

nadime15:
If you started reviewing the code, it’s ready for another look. @KotorinMinami and I worked together, and the code now structurally matches the other Vector Cryptography Extensions.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,342 @@
...
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause currentlyEnabled(Ext_Zvkned) = hartSupports(Ext_Zvkned) & currentlyEnabled(Ext_V)
+
+union clause ast = ZVKVAESDFTYPE : (zvkfunct6, vregidx, vregidx)
```
pmundkur, line 11:
```suggestion
union clause ast = VAESDF : (zvkfunct6, vregidx, vregidx)
```
Elsewhere in the PR there are ast nodes like `VAESKF1_VI`, `VAESZ_VS` and similar, which are easier to parse; this will make it more consistent.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,342 @@
...
+    let ark = sb ^ rkey;
+
+    write_velem_quad(vd, SEW, ark, i);
+  };
+
+  set_vstart(zeros());
+  RETIRE_SUCCESS
+}
+
+mapping vaesdf_mnemonic : zvkfunct6 <-> string = {
```
pmundkur, line 52: This is a partial mapping, since there are more elements in `zvkfunct6` than just these two.  `mapping`s should be complete, otherwise use scattered `mapping clause`s.

In this case, since there are only two valid `funct6` values for `VAESDF` and similar nodes , it might be best to split the big `zvkfunct6` enumeration into `{_VV, _VS}` pairs for each ast node.  This will also help make the above `if .. else` safer.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,342 @@
...
+  ZVK_VAESDFVV <-> "vaesdf.vv",
+  ZVK_VAESDFVS <-> "vaesdf.vs",
+}
+
+mapping clause assembly = ZVKVAESDFTYPE(funct6, vs2, vd)
+  <-> vaesdf_mnemonic(funct6) ^ sep() ^ vreg_name(vd) ^ sep() ^ vreg_name(vs2)
+
+union clause ast = ZVKVAESDMTYPE : (zvkfunct6, vregidx, vregidx)
```
pmundkur, line 60:
```suggestion
union clause ast = VAESDM : (zvkfunct6, vregidx, vregidx)
```

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,342 @@
...
+    let mix   = aes_mixcolumns_inv(ark);
+
+    write_velem_quad(vd, SEW, mix, i);
+  };
+
+  set_vstart(zeros());
+  RETIRE_SUCCESS
+}
+
+mapping vaesdm_mnemonic : zvkfunct6 <-> string = {
```
pmundkur, line 102: See above; make this into one complete mapping for `zvkfunct6`.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,342 @@
...
+  ZVK_VAESDMVV <-> "vaesdm.vv",
+  ZVK_VAESDMVS <-> "vaesdm.vs",
+}
+
+mapping clause assembly = ZVKVAESDMTYPE(funct6, vs2, vd)
+  <-> vaesdm_mnemonic(funct6) ^ spc() ^ vreg_name(vd) ^ sep() ^ vreg_name(vs2)
+
+union clause ast = ZVKVAESEFTYPE : (zvkfunct6, vregidx, vregidx)
```
pmundkur, line 110:
```suggestion
union clause ast = VAESEF : (zvkfunct6, vregidx, vregidx)
```

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,342 @@
...
+  let vd_val  = read_vreg(num_elem, SEW, LMUL_pow, vd);
+
+  let eg_len   = (unsigned(vl) / 4);
+  let eg_start = (unsigned(vstart) / 4);
+
+  foreach (i from eg_start to (eg_len - 1)) {
+    assert(i * 4 + 3 < num_elem);
+
+    let state = get_velem_quad(vd_val, i);
+    let rkey  = if funct6 == ZVK_VAESDFVV
```
pmundkur, line 38: This `if ... else` is safer if there are only possible values for `funct6`. See below.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,342 @@
...
+
+mapping vaesef_mnemonic : zvkfunct6 <-> string = {
+  ZVK_VAESEFVV <-> "vaesef.vv",
+  ZVK_VAESEFVS <-> "vaesef.vs",
+}
+
+mapping clause assembly = ZVKVAESEFTYPE(funct6, vs2, vd)
+  <-> vaesef_mnemonic(funct6) ^ spc() ^ vreg_name(vd) ^ spc() ^ vreg_name(vs2)
+
+union clause ast = ZVKVAESEMTYPE : (zvkfunct6, vregidx, vregidx)
```
pmundkur, line 159:
```suggestion
union clause ast = VAESEM : (zvkfunct6, vregidx, vregidx)
```

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,342 @@
...
+    let ark   = sr ^ rkey;
+
+    write_velem_quad(vd, SEW, ark, i);
+  };
+
+  set_vstart(zeros());
+  RETIRE_SUCCESS
+}
+
+mapping vaesef_mnemonic : zvkfunct6 <-> string = {
```
pmundkur, line 151: Splitting up `zvkfunct6` will help make this mapping complete; right now, it isn't.

pmundkur (changes requested):
Looks pretty clean!  Some minor fixes suggested.

model/riscv_insts_zvkned.sail
nadime15: If I understand correctly, do you mean something like

```
enum zvk_vaesdf_funct6 = {ZVK_VAESDF_VV, ZVK_VAESDF_VS}
enum zvk_vaesdm_funct6 = {ZVK_VAESDM_VV, ZVK_VAESDM_VS}
enum zvk_vaesef_funct6 = {ZVK_VAESEF_VV, ZVK_VAESEF_VS}
enum zvk_vaesem_funct6 = {ZVK_VAESEM_VV, ZVK_VAESEM_VS}
```
?

pmundkur: Yes, exactly.

model/riscv_insts_zvkned.sail
pmundkur: See rems-project/sail#1252

nadime15: Yes I noticed that too while testing it with the OCaml backend and it did not work, I was surprised that it worked with the C backend. Thats why I initially assumed its supported somehow.

I propose that we go ahead and merge the function as it is for now. We can then either update it after this is merged and the remaining vector crypto extensions are in, or modify it immediately. 

Personally, I think it makes more sense to keep the current version and follow up with a separate PR that explicitly changes the function.

pmundkur: I'm okay with both too, though we should file an issue with the details to track this.  Is this a hole in the vector test suite that doesn't trigger this?

nadime15: No, because the vector test suite only generates test cases for valid configurations.

KotorinMinami:
I've already made the corresponding changes. Since the `if` section already fully covers the `enum`, I wanted to use `match` to handle it. However, I'm not sure why, without specifying the type, sail can't infer the return type of the `match` expression. If you think using `if` is still more clear, I can switch it back.

model/riscv_insts_zvkned.sail
```diff
@@ -0,0 +1,346 @@
...
+
+union clause ast = VAESEM : (zvk_vaesem_funct6, vregidx, vregidx)
+
+mapping clause encdec = VAESEM(ZVK_VAESEM_VV, vs2, vd)
+  <-> 0b1010001 @ encdec_vreg(vs2) @ 0b00010 @ 0b010 @ encdec_vreg(vd) @ 0b1110111
+  when currentlyEnabled(Ext_Zvkned) & get_sew() == 32 & zvk_check_encdec(128, 4)
+
+mapping clause encdec = VAESEM(ZVK_VAESEM_VS, vs2, vd)
+  <-> 0b1010011 @ encdec_vreg(vs2) @ 0b00010 @ 0b010 @ encdec_vreg(vd) @ 0b1110111
+  when currentlyEnabled(Ext_Zvkned) & get_sew() == 32 & zvk_check_encdec(128, 4) & zvk_valid_reg_overlap(vs2, vd, get_lmul_pow())
```
pmundkur, line 170: Now that you have these enum pairs, you could leverage them to get a single `encdec` clause:
```
mapping encdec_vaesem : bits(7) <-> zvk_vaesem_funct6 = {
  0b1010001 <-> ZVK_VAESEM_VV,
  0b1010011 <-> ZVK_VAESEM_VS,
}

mapping clause encdec = VAESEM(op, vs2, vd)
  <-> encdec_vaesem(op) @ encdec_vreg(vs2) @ 0b00010 @ 0b010 @ encdec_vreg(vd) @ 0b1110111
  when ...

```
for each such pair

pmundkur (changes requested):

pmundkur:
> I've already made the corresponding changes. Since the `if` section already fully covers the `enum`, I wanted to use `match` to handle it. However, I'm not sure why, without specifying the type, sail can't infer the return type of the `match` expression. If you think using `if` is still more clear, I can switch it back.

The match is clearer, no needed to change.

model/riscv_insts_zvkned.sail
nadime15: If you mean merging, for example, `ZVK_VAESEM_VV` and `ZVK_VAESEM_VS`, that "won’t work" because the *_VS instructions include additional checks. That's why there exist a _VV and a _VS encdec clause

README.md
```diff
@@ -112,6 +112,7 @@ For booting operating system images, see the information under the
 - Zvbc extension for vector carryless multiplication, v1.0
 - Zvkb extension for vector cryptography bit-manipulation, v1.0
 - Zvknha and Zvknhb extensions for vector cryptography NIST Suite: Vector SHA-2 Secure Hash, v1.0
+- Zvkned extension for vector cryptography NIST Suite: Vector AES Block Cipher, v1.0
```
nadime15, line 115: Is there a reason why you put Zvkned after Zvknha ? In the spec Zvknh[ab] comes after Zvkned

config/default.json
```diff
@@ -177,6 +177,9 @@
     "Zvknha": {
       "supported": true
     },
+    "Zvkned": {
+      "supported": true
+    },
```
nadime15, line 182: Same for here, why is Zvkned in between Zvknha and Zvknhb?

model/riscv_zvk_utils.sail
```diff
@@ -66,3 +66,15 @@ function zvk_ch(x, y, z) = (x & y) ^ (~(x) & z)
...
+ * Utility functions for Zvkned
+ * ----------------------------------------------------------------------
+ */
+
+enum zvk_vaesdf_funct6 = {ZVK_VAESDF_VV, ZVK_VAESDF_VS}
+enum zvk_vaesdm_funct6 = {ZVK_VAESDM_VV, ZVK_VAESDM_VS}
+enum zvk_vaesef_funct6 = {ZVK_VAESEF_VV, ZVK_VAESEF_VS}
+enum zvk_vaesem_funct6 = {ZVK_VAESEM_VV, ZVK_VAESEM_VS}
+
+function aes_rotword(x : bits(32)) -> bits(32) = x[7.. 0] @ x[31..24] @ x[23..16] @ x[15.. 8]
```
nadime15, line 80: In my opinion this function should be part of `riscv_types_kext.sail`

`function aes_rotword(x : bits(32)) -> bits(32) = x[7.. 0] @ x[31..24] @ x[23..16] @ x[15.. 8]`

model/riscv_insts_zvkned.sail
pmundkur: Ah, I forgot about that, thanks; ignore the suggestion.

nadime15 (approved):
Looks good!

pmundkur (approved):
Nice!

jordancarlin:
Would be good to get this extension finished up before things fall too far out of date. @KotorinMinami can you rebase it and then I'm happy to review?

KotorinMinami:
@jordancarlin Sure！Please ！

jordancarlin (changes requested):
Looks pretty close. Just a few small comments.

jordancarlin (approved):

pmundkur (approved):

PR #754, opened 2025-02-24
Lean: improve handwritten support

ineol:


ineol:
I added a description to the commit. Is that enough or should I also add it to the PR?

Timmmm:
Yep that's fine.

PR #755, opened 2025-02-25
Refactor instruction retire and stepper

pmundkur:
This is a slightly cleaner implementation of the draft proposed in #746.
The change now affects 33 files instead of 47 files, hopefully making it easier to review.

I added @Timmmm, @jrtc27 and @Alasdair as co-authors to the main commit as it's based on their discussions and suggestions in #398, #412 and #746.

If merged, I plan to port #398 to this design.

### DIFF
--- a/model/riscv_inst_retire.sail
+++ b/model/riscv_inst_retire.sail
@@ -1,0 +1,30 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+/* Reasons an instruction retire might fail or be incomplete. */
+
+union Retire_Failure = {
+  // standard reasons
+  Illegal_Instruction            : unit,
+  Wait_For_Interrupt             : unit,
+  Trap                           : (Privilege, ctl_result, xlenbits),
+  Memory_Exception               : (virtaddr, ExceptionType),
+
+  // reasons from external extensions
+  Ext_CSR_Check_Failure          : unit,
+  Ext_ControlAddr_Check_Failure  : ext_control_addr_error,
+  Ext_DataAddr_Check_Failure     : ext_data_addr_error,
+  Ext_XRET_Priv_Failure          : unit,
+}
+
+// To ease the introduction of `Retire_Failure`, this global
+// definition is temporarily used instead of the previous
+// RETIRE_SUCCESS enum value so that we don't have to immediately
+// update all the places RETIRE_SUCCESS was used.
+
+let RETIRE_SUCCESS : Retired(Retire_Failure) = RETIRE_OK()
--- a/model/riscv_inst_retire.sail
+++ b/model/riscv_inst_retire.sail
@@ -27,4 +27,4 @@ union Retire_Failure = {
 // RETIRE_SUCCESS enum value so that we don't have to immediately
 // update all the places RETIRE_SUCCESS was used.
 
-let RETIRE_SUCCESS : Retired(Retire_Failure) = RETIRE_OK()
+let RETIRE_SUCCESS : ExecutionResult(Retire_Failure) = RETIRE_OK()
--- a/model/riscv_insts_vext_arith.sail
+++ b/model/riscv_insts_vext_arith.sail
@@ -51,7 +51,7 @@ function clause execute(VVTYPE(funct6, vm, vs2, vs1, vd)) = {
   let VLEN_pow = get_vlen_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_normal(vd, vm) then { handle_illegal(); return RETIRE_FAIL };
+  if illegal_normal(vd, vm) then return RETIRE_FAIL(Illegal_Instruction());
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -61,7 +61,10 @@ function clause execute(VVTYPE(funct6, vm, vs2, vs1, vd)) = {
   let vs2_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
   let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
 
-  let (initial_result, mask) = init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val);
+  let (initial_result, mask) : (vector('n, bits('m)), bits('n)) = match init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val) {
+    Ok(v)   => v,
+    Err(()) => return RETIRE_FAIL(Illegal_Instruction())
+  };
   var result = initial_result;
 
   foreach (i from 0 to (num_elem - 1)) {
@@ -114,14 +117,14 @@ function clause execute(VVTYPE(funct6, vm, vs2, vs1, vd)) = {
         VV_VMAXU         => to_bits(SEW, max(unsigned(vs2_val[i]), unsigned(vs1_val[i]))),
         VV_VMAX          => to_bits(SEW, max(signed(vs2_val[i]), signed(vs1_val[i]))),
         VV_VRGATHER      => {
-                              if (vs1 == vd | vs2 == vd) then { handle_illegal(); return RETIRE_FAIL };
+                              if (vs1 == vd | vs2 == vd) then return RETIRE_FAIL(Illegal_Instruction());
                               let idx = unsigned(vs1_val[i]);
                               let VLMAX = 2 ^ (LMUL_pow + VLEN_pow - SEW_pow);
                               assert(VLMAX <= 'n);
                               if idx < VLMAX then vs2_val[idx] else zeros()
                             },
         VV_VRGATHEREI16  => {
-                              if (vs1 == vd | vs2 == vd) then { handle_illegal(); return RETIRE_FAIL };
+                              if (vs1 == vd | vs2 == vd) then return RETIRE_FAIL(Illegal_Instruction());
                               /* vrgatherei16.vv uses SEW/LMUL for the data in vs2 but EEW=16 and EMUL = (16/SEW)*LMUL for the indices in vs1 */
                               let vs1_new : vector('n, bits(16)) = read_vreg(num_elem, 16, 4 + LMUL_pow - SEW_pow, vs1);
                               let idx = unsigned(vs1_new[i]);
@@ -187,7 +190,7 @@ function clause execute(NVSTYPE(funct6, vm, vs2, vs1, vd)) = {
 
   if  illegal_variable_width(vd, vm, SEW_widen, LMUL_pow_widen) |
       not(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow))
-  then { handle_illegal(); return RETIRE_FAIL };
+  then return RETIRE_FAIL(Illegal_Instruction());
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -198,7 +201,10 @@ function clause execute(NVSTYPE(funct6, vm, vs2, vs1, vd)) = {
   let vs1_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs1);
   let vs2_val : vector('n, bits('o)) = read_vreg(num_elem, SEW_widen, LMUL_pow_widen, vs2);
 
-  let (initial_result, mask) = init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val);
+  let (initial_result, mask) : (vector('n, bits('m)), bits('n)) = match init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val) {
+    Ok(v)   => v,
+    Err(()) => return RETIRE_FAIL(Illegal_Instruction())
+  };
   var result = initial_result;
 
   assert(SEW_widen <= 64);
@@ -254,7 +260,7 @@ function clause execute(NVTYPE(funct6, vm, vs2, vs1, vd)) = {
 
   if  illegal_variable_width(vd, vm, SEW_widen, LMUL_pow_widen) |
       not(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow))
-  then { handle_illegal(); return RETIRE_FAIL };
+  then return RETIRE_FAIL(Illegal_Instruction());
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -265,7 +271,10 @@ function clause execute(NVTYPE(funct6, vm, vs2, vs1, vd)) = {
   let vs1_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs1);
   let vs2_val : vector('n, bits('o)) = read_vreg(num_elem, SEW_widen, LMUL_pow_widen, vs2);
 
-  let (initial_result, mask) = init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val);
+  let (initial_result, mask) : (vector('n, bits('m)), bits('n)) = match init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val) {
+    Ok(v)   => v,
+    Err(()) => return RETIRE_FAIL(Illegal_Instruction())
+  };
   var result = initial_result;
 
   assert(SEW_widen <= 64);
@@ -308,14 +317,17 @@ mapping clause encdec = MASKTYPEV (vs2, vs1,  vd)
   when currentlyEnabled(Ext_V)
 
 function clause execute(MASKTYPEV(vs2, vs1, vd)) = {
-  let start_element = get_start_element();
+  let start_element : nat = match get_start_element() {
+    Ok(v)   => v,
+    Err(()) => return RETIRE_FAIL(Illegal_Instruction())
+  };
   let end_element   = get_end_element();
   let SEW           = get_sew();
   let LMUL_pow      = get_lmul_pow();
   let num_elem      = get_num_elem(LMUL_pow, SEW); /* max(VLMAX,VLEN/SEW)) */
   let real_num_elem = if LMUL_pow >= 0 then num_elem else num_elem / (0 - LMUL_pow); /* VLMAX */
 
-  if illegal_vd_masked(vd) then { handle_illegal(); return RETIRE_FAIL };
+  if illegal_vd_masked(vd) then return RETIRE_FAIL(Illegal_Instruction());
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -361,7 +373,7 @@ function clause execute(MOVETYPEV(vs1, vd)) = {
   let LMUL_pow = get_lmul_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_vd_unmasked() then { handle_illegal(); return RETIRE_FAIL };
+  if illegal_vd_unmasked() then return RETIRE_FAIL(Illegal_Instruction());
 
   let 'n = num_elem;
   let 'm = SEW;
@@ -370,7 +382,10 @@ function clause execute(MOVETYPEV(vs1, vd)) = {
   let vs1_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs1);
   let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
 
-  let (initial_result, mask) = init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val);
+  let (initial_result, mask) : (vector('n, bits('m)), bits('n)) = match init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val) {
+    Ok(v)   => v,
+    Err(()) => return RETIRE_FAIL(Illegal_Instruction())
+  };
   var result = initial_result;
 
   foreach (i from 0 to (num_elem - 1)) {
@@ -420,7 +435,7 @@ function clause execute(VXTYPE(funct6, vm, vs2, rs1, vd)) = {
   let LMUL_pow = get_lmul_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_normal(vd, vm) then { handle_illegal(); return RETIRE_FAIL };
+  if illegal_normal(vd, vm) then return RETIRE_FAIL(Illegal_Instruction());
 
   let 'n = num_elem;
   let 'm = SEW;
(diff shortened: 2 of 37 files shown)


pmundkur:
I've rebased this.  There were some additional changes due to #760 being merged since.  @Timmmm  there was an interaction with your #760; something to look at during your review.

I found #783 while rebasing, but leaving a fix for that to a later PR.

nwf (commented):
I got curious, so did some reading and would like my conclusions checked.

If this lands (and I think it should; it looks great!), the CHERI extension model will "just" have to remove `RETIRE_FAIL` after [`handle_cheri_cap_exception`](https://github.com/CTSRD-CHERI/sail-cheri-riscv/blob/74a9da1f78bdf50d263ffc98f2a587cf969b3e08/src/cheri_addr_checks.sail#L65-L76), as in
```
handle_cheri_cap_exception(CapEx_TagViolation, auth_idx); RETIRE_FAIL
```
and use a slightly tweaked function
```
val handle_cheri_cap_exception : (CapEx, capreg_idx) -> Retired(Retire_Failure)
```
that always returns something of the `Retired(Trap(...))` union arm?  That seems easy enough!

pmundkur:
@nwf Yes; the changes needed should be individually small but pervasive, affecting each `RETIRE_FAIL`.

model/riscv_sys_control.sail
```diff
@@ -145,6 +147,18 @@ function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
...
 
+/* Check if a locally enabled interrupt is pending.
+ *
+ * This does not examine the global enable bits (MIE and SIE in
+ * mstatus) and their delegation in mideleg. It does honor the
+ * local interrupt enables in mie.
+ *
+ * This is used when the hart is in the Waiting state.
+ */
+function wakeForInterrupt() -> bool = {
```
Timmmm, line 158: I'd probably call this `shouldWakeForInterrupt()`.

model/riscv_step.sail
```diff
@@ -8,8 +8,117 @@
 
 /* The emulator fetch-execute-interrupt dispatch loop. */
 
-/* returns whether to increment the step count in the trace */
-function step(step_no : int) -> bool = {
+register step_state : StepState
+
+function init_stepper() -> unit = {
+  step_state = STEP_ACTIVE;
```
Timmmm, line 14: I think you could just initialise this like

```
register step_state : StepState = STEP_ACTIVE
```

Timmmm (commented):
This is great. I have some initial comments. I haven't read `riscv_step.sail` yet.

model/riscv_types.sail
```diff
@@ -110,9 +110,15 @@ function privLevel_to_str (p) =
...
 overload to_str = {privLevel_to_str}
 
-/* enum denoting whether an executed instruction retires */
+/* Type denoting whether an executed instruction retires
+ * that is generic over the reasons why a retire might
+ * fail or be incomplete.
+ */
 
-enum Retired = {RETIRE_SUCCESS, RETIRE_FAIL}
+union Retired ('a : Type) = {
```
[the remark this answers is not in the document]
pmundkur: I think in the future we might need to make `Retire_Failure` scattered to enable extensions like `Sdext` and `Sdtrig`.  It might help dependency ordering issues if this definition was generic and concretized later, so I'd prefer to leave it generic for now.

model/riscv_step_common.sail
```diff
@@ -6,6 +6,24 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
+// The state of the emulator.
+//
+// The RISC-V hart could either be active (fetching and executing
+// instructions) or waiting (e.g. for an interrupt).  The transition
```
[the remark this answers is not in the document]
pmundkur: Yeah, we could be more explicit but we'd have to keep the comment up-to-date if this list changes so I'm not sure if it help much.  But I'm okay either way.

model/riscv_step_common.sail
```diff
@@ -6,6 +6,24 @@
...
+// from `active` to `wait` occurs on instructions like WFI.  The
+// transition from `wait` to `active` can occur if (i) the hart
+// detects an interrupt, or (ii) if the external non-Sail emulator
+// harness indicates that the waiting instruction should be retired
+// (as a nop).
```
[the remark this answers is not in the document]
pmundkur: Sounds good.

model/riscv_inst_retire.sail
```diff
@@ -0,0 +1,28 @@
...
+  Memory_Exception               : (virtaddr, ExceptionType),
+
+  // reasons from external extensions
+  Ext_CSR_Check_Failure          : unit,
+  Ext_ControlAddr_Check_Failure  : ext_control_addr_error,
+  Ext_DataAddr_Check_Failure     : ext_data_addr_error,
+  Ext_XRET_Priv_Failure          : unit,
+}
+
+// For backwards compatibility, this global definition is used
```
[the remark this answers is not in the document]
pmundkur: That's better, thanks!

model/riscv_insts_vext_mem.sail
```diff
@@ -67,31 +67,36 @@ union clause ast = VLSEGTYPE : (bits(3), bits(1), regidx, vlewidth, vregidx)
...
-val process_vlseg : forall 'f 'b 'n 'p, (0 < 'f & 'f <= 8) & ('b in {1, 2, 4, 8}) & ('n > 0). (int('f), bits(1), vregidx, int('b), regidx, int('p), int('n)) -> Retired
+val process_vlseg : forall 'f 'b 'n 'p, (0 < 'f & 'f <= 8) & ('b in {1, 2, 4, 8}) & ('n > 0). (int('f), bits(1), vregidx, int('b), regidx, int('p), int('n)) -> Retired(Retire_Failure)
 function process_vlseg (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem) = {
   let EMUL_reg : int = if EMUL_pow <= 0 then 1 else 2 ^ EMUL_pow;
   let width_type : word_width = size_bytes(load_width_bytes);
   let vm_val : bits('n) = read_vmask(num_elem, vm, zvreg);
   let vd_seg : vector('n, bits('f * 'b * 8)) = read_vreg_seg(num_elem, load_width_bytes * 8, EMUL_pow, nf, vd);
 
-  let (result, mask) = init_masked_result(num_elem, nf * load_width_bytes * 8, EMUL_pow, vd_seg, vm_val);
+  let 'n = num_elem;
```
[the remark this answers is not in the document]
pmundkur: Good catch, thanks!

model/riscv_sys_control.sail
pmundkur: Thats a better name, sure.

model/riscv_step.sail
pmundkur: I could, but `init_stepper` is also needed for RVFI, since it initializes the model multiple times (C:`reinit_sail()` -> Sail:`init_model()` -> Sail:`init_stepper()`).  So I just left it in `init_stepper()`.

Timmmm: RVFI calls the generated C code that reruns the global initialisers (`model_init()` IIRC), so it's safe to do it using that.

model/riscv_step_common.sail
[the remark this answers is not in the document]
Timmmm: Yeah I'm not aware of any others that exist so I think we may as well list them. Easy to update if more arrive.

model/riscv_types.sail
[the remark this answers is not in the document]
Timmmm: Ok seems reasonable.

model/riscv_step.sail
```diff
@@ -8,8 +8,113 @@
...
+    // There are no pending interrupts; transition out of the Wait
+    // as instructed.
+    if   get_config_print_instr()
+    then print_instr("forced exit from WAIT state at PC " ^ BitStr(PC));
+
+    step_state = STEP_ACTIVE;
+    // "When TW=1, then if WFI is executed in any
+    // less-privileged mode, and it does not complete within an
+    // implementation-specific, bounded time limit, the WFI
+    // instruction causes an illegal-instruction exception."
```
Timmmm, line 58: I don't think this is quite right because `exit_wait` is true when it *does* complete. I think when they say "complete" they mean it retires.

Timmmm: Maybe we need 3 options for that parameter then:

* Keep waiting
* Retire successfully (spuriously)
* Timeout

Does `wrs` have similar options?

pmundkur:
Pushed a version that omits the `max-wait-steps` configuration, but keeps the `wait_is_nop` option, currently hardcoded to `false`.  This can be configured later using the configuration system.

pmundkur:
@Timmmm I think this version has the changes you'd requested on the call.

pmundkur:
@jordancarlin would you have time to review this?  This is holding up Zawrs (#777).

jordancarlin:
> @jordancarlin would you have time to review this? This is holding up Zawrs (#777).

Yes. I'll get to it by this weekend at the latest. Sorry about the delay.

model/riscv_step.sail
```diff
@@ -8,8 +8,116 @@
...
 
-/* returns whether to increment the step count in the trace */
-function step(step_no : int) -> bool = {
+register step_state : StepState
+
+// Checks if the retire returned by the execution of an instruction
+// counts as a step, where a step occurs when an instruction retired
+// successfully or it raised an exception.
+function does_executed_retire_step(r : Retired(Retire_Failure)) -> bool =
+  match r {
```
jordancarlin, line 17: Might be simpler to only explicitly list retires that don't step and have the default case be true. Most retires step, so this draws more attention to those that are different (and has the additional benefit of less code).

model/riscv_step.sail
```diff
@@ -24,80 +132,111 @@ function step(step_no : int) -> bool = {
...
+    },
+    Step_Execute(RETIRE_FAIL(Illegal_Instruction())) => {
+      minstret_increment = false;
+      match instbits {
+        None()         => internal_error(__FILE__, __LINE__, "no instruction bits for illegal instruction"),
+        Some(instbits) => handle_illegal(instbits)
+      }
+    },
+    Step_Execute(RETIRE_FAIL(Wait_For_Interrupt())) =>
+      if wait_is_nop() then {
```
jordancarlin, line 180: I wonder if this option should be called `wfi_is_nop()` or something along those lines. We probably want separate options for `wfi` to be a `nop` and for `wars` instructions to be a `nop`.

jordancarlin (commented):
@pmundkur Couple of small comments, but this is looking very close! Will be great to get this refactor merged.

model/riscv_insts_vext_arith.sail
```diff
@@ -61,7 +61,10 @@ function clause execute(VVTYPE(funct6, vm, vs2, vs1, vd)) = {
   let vs2_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
   let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
 
-  let (initial_result, mask) = init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val);
+  let (initial_result, mask) : (vector('n, bits('m)), bits('n)) = match init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val) {
```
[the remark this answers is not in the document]
pmundkur: In this case, the compiler needs the type annotations; there is a "Cannot infer type of" error without them.

model/riscv_step.sail
pmundkur: I generally prefer to list all the variants instead of using a default `_` match so that when a new variant is added,  the compiler gives a reminder to handle the new case instead of having it silently match the default.

model/riscv_step.sail
pmundkur: I wondered too, and am okay with having separate options.

model/riscv_insts_vext_arith.sail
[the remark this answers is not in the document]
jordancarlin: Unfortunate but makes sense with the added type complexity here.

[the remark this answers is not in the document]
jordancarlin: Unfortunate but makes sense with the added type complexity here.

model/riscv_step.sail
jordancarlin: I think it’s reasonable for an implementation to do different things for the two instructions, and this avoids needing to rename the option later.

model/riscv_step.sail
jordancarlin: I think that new variants will almost always retire, but I’m fine either way with this one.

model/riscv_step.sail
pmundkur: Pushed with rename to `wfi_is_nop`

jordancarlin (approved):

pmundkur:
@Timmmm I've tried to make the comments clearer to address your feedback during the meeting.

pmundkur:
I'd like to merge this by the end of the week.  That simplify working on #835 since there are several collisions in the vector files.  Also, this has been holding up adding Zawrs (#777) which has been ready for a month.

model/riscv_step.sail
```diff
@@ -24,80 +136,111 @@ function step(step_no : int) -> bool = {
...
+      set_next_pc(exception_handler(priv, ctl, pc))
+    },
+    Step_Execute(RETIRE_FAIL(Memory_Exception(vaddr, exc))) => {
+      minstret_increment = false;
+      handle_mem_exception(vaddr, exc)
+    },
+    Step_Execute(RETIRE_FAIL(Illegal_Instruction())) => {
+      minstret_increment = false;
+      match instbits {
+        None()         => internal_error(__FILE__, __LINE__, "no instruction bits for illegal instruction"),
```
Timmmm, line 179: This suggests the types are not quite right. Shouldn't `instbits` go in `Step_Execute`?

model/riscv_step.sail
```diff
@@ -24,80 +136,111 @@ function step(step_no : int) -> bool = {
...
-        }
-      }
+  let ((step_val, did_step), instbits) : ((Step, bool), option(xlenbits)) =
+    match step_state {
+      STEP_WAIT   => (step_wait_state(step_no, exit_wait), None()),
+      STEP_ACTIVE => step_active_state(step_no, exit_wait)
     };
 
-  tick_pc();
+  var stepped : bool = did_step;
```
Timmmm, line 145: I believe you can have these named the same.

```suggestion
  var stepped : bool = stepped;
```

Timmmm (changes requested):
I think there are still way too many things called "step". It's very hard to follow.

How about

* `register step_state : StepState` -> `register hart_state : HartState` (`HART_ACTIVE`, `HART_WAITING`)
* `step_wait_state`, `step_active_state` -> `run_hart_waiting`, `run_hart_active`
* `union Step` -> 

```
union RunResult = {
  RunResult_Interrupt_Pending ...
  RunResult_Ext_Fetch_Failure ...
  RunResult_Execution ...
  RunResult_Waiting ...
}
```

I'm not super happy with "run", but it's better than using "step" again!

Also we should rename `Retired` to something  like `ExecutionResult` at some point because it doesn't mean that the instruction has retired! `Retire(DidntRetire)` is nonsense.

model/riscv_insts_vext_utils.sail
```diff
@@ -183,8 +183,9 @@ function get_start_element() = {
     It is recommended that implementations trap if vstart is out of bounds.
     It is not required to trap, as a possible future use of upper vstart bits
     is to store imprecise trap information. */
-  if start_element > (2 ^ (3 + VLEN_pow - SEW_pow) - 1) then handle_illegal();
-  start_element
+  if start_element > (2 ^ (3 + VLEN_pow - SEW_pow) - 1)
```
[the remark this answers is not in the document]
pmundkur: Hmm, it looks like the vector load/store instructions do set `vstart`, so we should use the right bounds here.  I'll leave a TODO, but perhaps leave a fix to a different PR?

model/riscv_step.sail
```diff
@@ -8,8 +8,120 @@
...
+
+function wfi_is_nop() -> bool = config platform.wfi_is_nop
+
+// The `step` function is the main internal driver of the Sail model.
+// It performs the fetch-decode-execute for an instruction.  It is
+// also the primary interface to the non-Sail execution harness.  The
+// harness calls this function with the current step number (which
+// numbers the active or wait step), and whether it should exit a wait
+// state.  The `step` function returns whether the Sail emulator
+// executed a step, and the stepper state at the end of the step.
```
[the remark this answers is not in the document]
pmundkur: I'll use `try_step` for now.  We could revisit this in #777 since it touches this same code.

Science says both options are right :-)  (paywalled, so I haven't read the article)
https://www.theatlantic.com/science/archive/2018/05/two-spaces-after-a-period/559304/

model/riscv_step.sail
```diff
@@ -24,80 +136,111 @@ function step(step_no : int) -> bool = {
...
+      ext_handle_fetch_check_error(e)
+    },
+    Step_Waiting() => {
+      assert(~(wfi_is_nop()));
+      assert(step_state == STEP_WAIT, "cannot be Waiting in a non-Wait state");
+    },
+    Step_Execute(RETIRE_OK()) => {
+      assert(step_state == STEP_ACTIVE);
+      // default handling of minstret
+      ()
```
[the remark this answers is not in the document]
pmundkur: It looks like I didn't need it, thanks!

model/riscv_step.sail
pmundkur: The compiler does not like that:
`Cannot modify immutable let-bound constant or enumeration constructor stepped`

model/riscv_step.sail
pmundkur: It doesn't quite work; the `internal_error` just moves elsewhere.  The only way to avoid this is to give the `instbits` to `execute` so it can package it into `Illegal_Instruction`.  Or have it available in a global register, which is just going back to #412.  We can't package the instruction ast into `Illegal_Instruction` since it doesn't work for RVC instructions whose execute just calls the execute of the uncompressed ast.

model/riscv_step.sail
```diff
@@ -8,8 +8,120 @@
...
-/* returns whether to increment the step count in the trace */
-function step(step_no : int) -> bool = {
+register step_state : StepState
+
+// Returns whether the retire result returned by the execution of an
+// instruction counts as a step, where a step is said to occur when an
+// instruction retires successfully or it raises an exception.
+// Crucially, if an instruction returns a Wait retire result, it does
+// not step.
+function does_executed_retire_step(r : Retired(Retire_Failure)) -> bool =
```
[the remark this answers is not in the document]
pmundkur: Ok

model/riscv_insts_vext_utils.sail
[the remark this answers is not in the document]
Timmmm: Yeah I don't really know enough about vector to say. [Here's what we're using anyway.](https://github.com/CHERI-Alliance/sail-riscv/blob/c19285f7894b6262d92aea604bcd866c359532f8/model/riscv_insts_vext_utils.sail#L200-L216).

pmundkur:
> I think there are still way too many things called "step". It's very hard to follow.
> 
> How about
> 
>     * `register step_state : StepState` -> `register hart_state : HartState` (`HART_ACTIVE`, `HART_WAITING`)
> 
>     * `step_wait_state`, `step_active_state` -> `run_hart_waiting`, `run_hart_active`

Done.

> 
>     * `union Step` ->
> 
> 
> ```
> union RunResult = {
>   RunResult_Interrupt_Pending ...
>   RunResult_Ext_Fetch_Failure ...
>   RunResult_Execution ...
>   RunResult_Waiting ...
> }
> ```
> 
> I'm not super happy with "run", but it's better than using "step" again!

I split the difference.  Since I kept the `step` function (renamed to `try_step`), I kept the `Step` union. 

> 
> Also we should rename `Retired` to something like `ExecutionResult` at some point because it doesn't mean that the instruction has retired! `Retire(DidntRetire)` is nonsense.

Agreed; I did the rename.

model/riscv_step.sail
pmundkur: I was able to address this by storing the `instbits` in the `hart_state`.

model/riscv_inst_retire.sail
```diff
@@ -0,0 +1,30 @@
...
+  Ext_DataAddr_Check_Failure     : ext_data_addr_error,
+  Ext_XRET_Priv_Failure          : unit,
+}
+
+// To ease the introduction of `Retire_Failure`, this global
+// definition is temporarily used instead of the previous
+// RETIRE_SUCCESS enum value so that we don't have to immediately
+// update all the places RETIRE_SUCCESS was used.
+
+let RETIRE_SUCCESS : ExecutionResult(Retire_Failure) = RETIRE_OK()
```
nwf, line 30: Should/can this be polymorphic and down (up?) in `riscv_types.sail`, where `ExecutionResult` and `RETIRE_OK` are defined?  Is the monomorphism of the not-used `union` arm convenient for type inference at use sites?

model/riscv_sys_control.sail
```diff
@@ -145,6 +147,18 @@ function getPendingSet(priv : Privilege) -> option((xlenbits, Privilege)) = {
...
 }
 
+/* Check if a locally enabled interrupt is pending.
+ *
+ * This does not examine the global enable bits (MIE and SIE in
+ * mstatus) and their delegation in mideleg. It does honor the
+ * local interrupt enables in mie.
+ *
+ * This is used when the hart is in the Waiting state.
+ */
```
nwf, line 157: The documentation might benefit from pointing out that this is specifically used by the WFI instruction (and any of its kindred, perhaps) and citing and quoting [the relevant prose from the privileged spec](https://five-embeddev.com/riscv-priv-isa-manual/Priv-v1.12/machine.html#wfi):
> The operation of WFI must be unaffected by the global interrupt bits in  mstatus (MIE and SIE) and the delegation register mideleg (i.e., the hart must resume if a locally enabled interrupt becomes pending, even if it has been delegated to a less-privileged mode), but should honor the individual interrupt enables (e.g, MTIE) (i.e., implementations should avoid resuming the hart if the interrupt is pending but not individually enabled).

model/riscv_inst_retire.sail
pmundkur: I should have added
```
let RETIRE_SUCCESS = RETIRE_OK()
```
gives a type error.

model/riscv_sys_control.sail
pmundkur: Ok.

model/riscv_inst_retire.sail
nwf: Can top-level `let`s take a type annotation?
```
let RETIRE_SUCCESS : forall 'a . ExecutionResult('a) = RETIRE_OK()
```

pmundkur: No, there is a syntax error at `forall`.

jordancarlin (approved):
LGTM. I think we can continue to do some cleanup in follow up PRs if needed, but this seems ready to merge and unblocks several other things.

pmundkur:
Marking this with `will-be-merged`.  I'll merge this tomorrow.

Timmmm:
I normally use the "Create a merge commit" option if I'm merging a PR and not squashing multiple commits. If you use "Rebase and merge" and there's more than one commit in the PR then you end up with commits on `master` that haven't gone through CI, which can break `git bisect`.

pmundkur:
Understood.  In this case, this was intentional: the rebase had conflicts, and I was able to test each commit in the sequence.  Sorry, I'll mention this next time when merging to avoid surprise.

PR #757, opened 2025-02-26
Add support for Zcmt extension

### LINKED ISSUE #1740: Add Zcmt extension
While running ACT's on cv32e40x issues came up - https://github.com/riscv/riscv-arch-test/issues/1623. This core implements Zcmt extensions. Running SsstrictSm test on this core fails since SsstrictSm checks that every unimplemented/reserved instruction and CSR access must raise an illegal-instruction trap.

There is all ready open PR #757 that would fix this issue, and since cv32e40x configuration is part of ACT this would help.

nadime15:
Add the following instructions:
* cm.jt
* cm.jalt

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ For booting operating system images, see the information under the
 - Zfbfmin extension for scalar bf16 converts, v1.0
 - Zfinx, Zdinx, Zhinx, and Zhinxmin extensions for floating-point in integer registers, v1.0
 - C extension for compressed instructions, v2.0
-- Zca, Zcf, Zcd, and Zcb extensions for code size reduction, v1.0
+- Zca, Zcf, Zcd, Zcb and Zcmt extensions for code size reduction, v1.0
 - Zcmop extension for compressed May-Be-Operations, v1.0
 - B (Zba, Zbb, Zbs) and Zbc extensions for bit manipulation, v1.0
 - Zbkb, Zbkc, and Zbkx extensions for bit manipulation for cryptography, v1.0
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -398,6 +398,9 @@
     "Zcmop": {
       "supported": true
     },
+    "Zcmt": {
+      "supported": false
+    },
     "Zba": {
       "supported": false
     },
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -21,6 +21,7 @@
   - Ziccamoc
   - Ziccrse
   - Zicfiss
+  - Zcmt
   - Ssccptr
   - Sscounterenw
   - Svade
--- a/model/core/extensions.sail
+++ b/model/core/extensions.sail
@@ -210,6 +210,10 @@ function clause hartSupports(Ext_Zcf) = config extensions.Zcf.supported : bool &
 enum clause extension = Ext_Zcmop
 mapping clause extensionName = Ext_Zcmop <-> "zcmop"
 function clause hartSupports(Ext_Zcmop) = config extensions.Zcmop.supported
+// Code Size Reduction: compressed table jump instructions
+enum clause extension = Ext_Zcmt
+mapping clause extensionName = Ext_Zcmt <-> "zcmt"
+function clause hartSupports(Ext_Zcmt) = config extensions.Zcmt.supported
 // Compressed Instructions
 // C is supported if all of the following are true:
 //   - Zca is supported
@@ -596,6 +600,7 @@ function currentlyEnabled_measure(ext : extension) -> int =
     Ext_Zve32f => 2, // > Zve32x
     Ext_Zve64x => 2, // > Zve32x
     Ext_Zcmop => 2, // > Zca
+    Ext_Zcmt => 2, // > Zca
     Ext_Zicfiss => 2, // > Zaamo
 
     Ext_Ssccptr => 3, // > Sv39
--- a/model/extensions/FD/zcd_insts.sail
+++ b/model/extensions/FD/zcd_insts.sail
@@ -6,7 +6,8 @@
 // SPDX-License-Identifier: BSD-2-Clause
 // =======================================================================================
 
-function clause currentlyEnabled(Ext_Zcd) = hartSupports(Ext_Zcd) & currentlyEnabled(Ext_D) & currentlyEnabled(Ext_Zca) & (currentlyEnabled(Ext_C) | not(hartSupports(Ext_C)))
+function clause currentlyEnabled(Ext_Zcd) =
+  hartSupports(Ext_Zcd) & currentlyEnabled(Ext_D) & currentlyEnabled(Ext_Zca) & not(hartSupports(Ext_Zcmt)) & (currentlyEnabled(Ext_C) | not(hartSupports(Ext_C)))
 
 union clause instruction = C_FLDSP : (bits(6), fregidx)
 
--- a/model/extensions/Stateen/stateen_access_checks.sail
+++ b/model/extensions/Stateen/stateen_access_checks.sail
@@ -52,6 +52,9 @@ function stateen_allows_CSR_access(csr : csreg, priv : Privilege, _access_type :
     // srmcfg
     (0x181, priv) => check_stateen_bit(priv, STATEEN_SRMCFG, 0),
 
+    // jvt
+    (0x017, priv) => if hartSupports(Ext_Zcmt) then check_stateen_bit(priv, STATEEN_JVT, 0) else true,
+
     // fflags
     (0x001, priv) => if hartSupports(Ext_Zfinx) then check_stateen_bit(priv, STATEEN_FCSR, 0) else true,
     // frm
--- a/model/extensions/Stateen/stateen_regs.sail
+++ b/model/extensions/Stateen/stateen_regs.sail
@@ -16,12 +16,13 @@ function is_sstateen_accessible() -> bool = currentlyEnabled(Ext_S) & (currently
 termination_measure is_hstateen_accessible() = 5 // > Ext_H
 termination_measure is_sstateen_accessible() = 2 // > Ext_S
 
-enum stateen_bit = {STATEEN_FCSR, STATEEN_SRMCFG, STATEEN_ENVCFG, STATEEN_SE}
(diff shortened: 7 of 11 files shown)


model/riscv_sys_regs.sail
```diff
@@ -902,6 +916,9 @@ function clause write_CSR(0x7a0, value) = { tselect = value; tselect }
 
 /*
  * Entropy Source - Platform access to random bits.
+ * WARNING: This function currently lacks a proper side-effect annotation.
+ *          If you are using theorem prover tool flows, you
+ *          may need to modify or stub out this function for now.
```
Timmmm, line 921: Was this intended to be added here? I vaguely recall removing it so maybe it is a rebase mistake?

model/riscv_sys_regs.sail
```diff
@@ -508,6 +512,16 @@ function clause write_CSR((0x302, value) if xlen == 32) = { medeleg = legalize_m
...
 function clause write_CSR(0x303, value) = { mideleg = legalize_mideleg(mideleg, value); mideleg.bits }
 
+bitfield Jvt : xlenbits = {
+  base : xlen - 1 .. 6,
+  mode : 5 .. 0
+}
+register jvt : Jvt
+mapping clause csr_name_map = 0x017  <-> "jvt"
+function clause is_CSR_defined (0x017) = sys_enable_zcmt()
+function clause write_CSR(0x017, value) = { jvt.bits = (value >> 6 ) << 6; jvt.bits }
```
Timmmm, line 522:
```suggestion
function clause write_CSR(0x017, value) = { jvt.bits = [value with 5 .. 0 = zeros()]; jvt.bits }
```

But it's probably better to make a `legalize_jvt()` function similar to the other `legalize` functions and then you can do something like:

```
function legalize_jvt(orig : Jvt, write : xlenbits) -> Jvt = {
  let write = Mk_Jvt(write);
  [orig with
    base = write.base,
  ]
}
```

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+type target_address = bits(xlen)
+
+union FJT_Result = {
+  FJT_Success : target_address,
+  FJT_Failure : unit
+}
+
+function fetch_jump_table(table_address : bits(xlen)) -> FJT_Result = {
+    /* Fetching jump table address needs execute permission */
+    match ext_data_get_addr_from_bits(table_address, Execute(), xlen_bytes) {
```
Timmmm, line 20: Indentation

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause extensionEnabled(Ext_Zcmt) = extensionEnabled(Ext_Zca) & sys_enable_zcmt() & not(sys_enable_zcd()) & (xlen == 32 | xlen == 64)
+
+type target_address = bits(xlen)
+
+union FJT_Result = {
```
Timmmm, line 13: Probably just use `result(target_address, unit)`

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+      32 => base + index << 2,
+      64 => base + index << 3,
+       _ => internal_error(__FILE__, __LINE__, "Unsupported xlen")
+    };
+    match fetch_jump_table(table_address) {
+        FJT_Failure(_) => {
+          return RETIRE_FAIL
+        },
+        FJT_Success(target_address) => {
+          X(0b00001) = get_next_pc();
```
Timmmm, line 56:
```suggestion
          X(1) = get_next_pc();
```

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+          set_next_pc(target_address & ~(zero_extend(0x1)));
```
Timmmm, line 57:
```suggestion
          set_next_pc([target_address with 0 = 0b0]);
```

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+union clause ast = CM_JALT : (bits(8))
+
+mapping clause encdec_compressed = CM_JALT(index) if extensionEnabled(Ext_Zcmt) & 32 <= unsigned(index)
+  <-> 0b101 @ 0b000 @ index : bits(8) @ 0b10 if extensionEnabled(Ext_Zcmt) & 32 <= unsigned(index)
+
+function clause execute (CM_JALT(index)) = {
+  let base : bits(xlen) = jvt.bits & ~(zero_extend(0x3F));
+  let index : bits(xlen) = zero_extend(index);
+  if jvt[mode] == 0b000000 then {
+    let table_address : bits(xlen) = match xlen {
```
Timmmm, line 46:
```
let table_address = base + index << log2_xlen_bytes;
```

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+}
+
+union clause ast = CM_JALT : (bits(8))
+
+mapping clause encdec_compressed = CM_JALT(index) if extensionEnabled(Ext_Zcmt) & 32 <= unsigned(index)
+  <-> 0b101 @ 0b000 @ index : bits(8) @ 0b10 if extensionEnabled(Ext_Zcmt) & 32 <= unsigned(index)
+
+function clause execute (CM_JALT(index)) = {
+  let base : bits(xlen) = jvt.bits & ~(zero_extend(0x3F));
+  let index : bits(xlen) = zero_extend(index);
```
Timmmm, line 44: I believe you can delete this line.

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+  };
+}
+
+union clause ast = CM_JALT : (bits(8))
+
+mapping clause encdec_compressed = CM_JALT(index) if extensionEnabled(Ext_Zcmt) & 32 <= unsigned(index)
+  <-> 0b101 @ 0b000 @ index : bits(8) @ 0b10 if extensionEnabled(Ext_Zcmt) & 32 <= unsigned(index)
+
+function clause execute (CM_JALT(index)) = {
+  let base : bits(xlen) = jvt.bits & ~(zero_extend(0x3F));
```
Timmmm, line 43:
```suggestion
  let base : bits(xlen) = jvt.base @ 0b000000;
```

(or `@ zeros(6)`)

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+mapping clause assembly = CM_JALT(index) if (xlen == 32 | xlen == 64) <->
+  "cm.jalt" ^ spc() ^ hex_bits_8(index) if (xlen == 32 | xlen == 64)
+
+/* ****************************************************************** */
+union clause ast = CM_JT : (bits(8))
+
+mapping clause encdec_compressed = CM_JT(index) if extensionEnabled(Ext_Zcmt) & unsigned(index) < 32
+  <-> 0b101 @ 0b000 @ index : bits(8) @ 0b10 if extensionEnabled(Ext_Zcmt) & unsigned(index) < 32
+
+function clause execute (CM_JT(index)) = {
```
Timmmm, line 73: Ditto on above comments.

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,95 @@
...
+        FJT_Failure(_) => {
+          return RETIRE_FAIL
+        },
+        FJT_Success(target_address) => {
+          X(0b00001) = get_next_pc();
+          set_next_pc(target_address & ~(zero_extend(0x1)));
+        }
+    };
+  };
+  RETIRE_SUCCESS
```
Timmmm, line 61: Since mode != 0 is reserved we probably want to `RETIRE_ILLEGAL` here. Also use `else RETIRE_ILLEGAL` so you can avoid `return` above.

Timmmm (changes requested):

model/riscv_insts_zcmt.sail
nadime15: This wont compile, I get:

`Failed to prove constraint: 8 == 32`

But I think it makes sense, imagine shifting bits to the left. Since the bit width of `index` remains the same (without extending), the two (or three) most significant bits would be lost.

model/riscv_insts_zcmt.sail
nadime15: Same here, this wont compile and I get:
```
No possible overloading for vector_update
```

nadime15:
Hi @Timmmm, thanks for the suggestions! I've incorporated them and left some comments about a few small issues I ran into.

model/riscv_insts_zcmt.sail
Timmmm: Ah yeah I totally misread that. Actually it would happily lose the 2/3 most significant bits; it's complaining that you can't add a 32-bit bitvector (`base`) to an 8-bit one (`index << log2_xlen_bytes`).

I wonder if it's worth adding brackets to make it clearer: `base + (index << log2_xlen_bytes)`?

model/riscv_insts_zcmt.sail
```diff
@@ -0,0 +1,91 @@
...
+function clause extensionEnabled(Ext_Zcmt) = extensionEnabled(Ext_Zca) & sys_enable_zcmt() & not(sys_enable_zcd()) & (xlen == 32 | xlen == 64)
+
+type target_address = bits(xlen)
+
+union FJT_Result = {
+  FJT_Success : target_address,
+  FJT_Failure : unit
+}
+
+function fetch_jump_table(table_address : bits(xlen)) -> FJT_Result = {
```
Timmmm, line 18: We use `xlenbits` instead of `bits(xlen)` elsewhere.

model/riscv_insts_zcmt.sail
Timmmm: Sorry, it should be `set_next_pc([target_address with 0 = bitzero]);`

model/riscv_insts_zcmt.sail
nadime15: Yeah good point, I will do that

Issue #614, opened 2024-11-11
Improve `to_bits` type checking


Timmmm:
Currently `to_bits` just silently accepts larger integers than will fit in the bitvector and slices off the top. That's a bit sad from a type checking point of view. Sail can perfectly happily check this at compile time:

```
val to_bits : forall 'l 'x, 'l >= 0 & 0 <= 'x < 2 ^ 'l . (int('l), int('x)) -> bits('l)
function to_bits (l, n) = get_slice_int(l, n, 0)
```

I think we should change the definition to that. There are some places where the slicing is intentional (e.g. the bit rotate functions), where we can probably introduce an explicit function:

```
val to_bits_truncate : forall 'l, 'l >= 0 . (int('l), int) -> bits('l)
function to_bits_truncate (l, n) = get_slice_int(l, n, 0)
```

And there are a lot of places where it probably isn't intentional, where we could add

```
val to_bits_unsafe : forall 'l, 'l >= 0 . (int('l), int) -> bits('l)
function to_bits_unsafe (l, n) = get_slice_int(l, n, 0)
```

with TODOs to add the appropriate assertions & types. A lot of them are in the vector extension which has very loose typing so it will take a while to fix them there.

jordancarlin:
Seems like a great type safety improvement that makes a lot of sense.On Nov 11, 2024, at 8:54 AM, Tim Hutt ***@***.***> wrote:﻿
Currently to_bits just silently accepts larger integers than will fit in the bitvector and slices off the top. That's a bit sad from a type checking point of view. Sail can perfectly happily check this at compile time:
val to_bits : forall 'l 'x, 'l >= 0 & 0 <= 'x < 2 ^ 'l . (int('l), int('x)) -> bits('l)
function to_bits (l, n) = get_slice_int(l, n, 0)

I think we should change the definition to that. There are some places where the slicing is intentional (e.g. the bit rotate functions), where we can probably introduce an explicit function:
val to_bits_truncate : forall 'l, 'l >= 0 . (int('l), int) -> bits('l)
function to_bits_truncate (l, n) = get_slice_int(l, n, 0)

And there are a lot of places where it probably isn't intentional, where we could add
val to_bits_unsafe : forall 'l, 'l >= 0 . (int('l), int) -> bits('l)
function to_bits_unsafe (l, n) = get_slice_int(l, n, 0)

[...]

PR #758, opened 2025-02-26
Replace ad-hoc model configuration with Sail-native configuration

Alasdair:
This is the first in a series of PRs (and probably patches to this PR) aiming to replace the ad-hoc C stubs used for configuration with a Sail-native configuration system using a separate configuration file. This is a draft PR as it needs a new release of Sail to work.

This PR is split into two commits:

1. The first simply updates the CMake build system to include the Sail JSON handling code, and adds a simulator option to pass a configuration file but does not change the model code whatsoever. The file config/default.json is added to store all the default configuration options (but is empty in this commit). I took the `-c` short option away from the sail coverage option, as it isn't even built in by default.
2. The second replaces the existing options in the simulator. The only behaviour change is the handling of `rom_base` and `rom_size` is slightly different, in that it'll just print if the configuration values don't match what the C code that creates the reset vector and writes the DTB expect rather than setting the values.

Note that this PR isn't aiming to add all the configuration we want for MC100 or beyond, it's just a direct port of existing functionality.

With this adding a new configurable option to the model is easy, add `config x.y.z : T` anywhere in the model, and then add `x.y.z` into the JSON file, and it'll just work.

TODOs:

1. Update the other Sail targets. This configuration system works across all Sail backends, but the CMake commands for them need to be updated to reflect this
2. Sail can generate a JSON schema that will validate any provided config fits into the model, that hasn't been set up yet. The default config should be JSON schema validated in CI as well.
3. We probably want some more friendly way to write the configuration rather than plain JSON, that could be specific to this repository, or by writing a script that imports values from riscv-config or UDB, or both. I haven't decided the best way to go yet.

### DIFF
--- a/c_emulator/CMakeLists.txt
+++ b/c_emulator/CMakeLists.txt
@@ -50,7 +50,7 @@ foreach (xlen IN ITEMS 32 64)
             add_dependencies(riscv_sim_${arch} generated_model_${arch})
 
             target_link_libraries(riscv_sim_${arch}
-                PRIVATE softfloat sail_runtime GMP::GMP
+                PRIVATE softfloat sail_runtime default_config GMP::GMP
             )
 
             target_include_directories(riscv_sim_${arch}
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -10,9 +10,15 @@
 #include <sys/time.h>
 #include <fcntl.h>
 #include <optional>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <vector>
 
 #include "elf.h"
 #include "sail.h"
+#include "sail_config.h"
 #include "rts.h"
 #ifdef SAILCOV
 #include "sail_coverage.h"
@@ -21,6 +27,7 @@
 #include "riscv_platform_impl.h"
 #include "riscv_sail.h"
 #include "rvfi_dii.h"
+#include "default_config.h"
 
 const char *RV64ISA = "RV64IMAC";
 const char *RV32ISA = "RV32IMAC";
@@ -44,19 +51,12 @@ const char *RV32ISA = "RV32IMAC";
 
 enum {
   OPT_TRACE_OUTPUT = 1000,
-  OPT_ENABLE_WRITABLE_FIOM,
-  OPT_PMP_COUNT,
-  OPT_PMP_GRAIN,
-  OPT_ENABLE_SVINVAL,
-  OPT_ENABLE_ZCB,
-  OPT_ENABLE_ZICBOM,
-  OPT_ENABLE_ZICBOZ,
-  OPT_ENABLE_ZVKB,
-  OPT_ENABLE_SSTC,
-  OPT_CACHE_BLOCK_SIZE,
+  OPT_PRINT_CONFIG,
+  OPT_SAILCOV,
 };
 
 static bool do_show_times = false;
+bool do_report_arch = false;
 char *term_log = NULL;
 static const char *trace_log_path = NULL;
 FILE *trace_log = NULL;
@@ -112,43 +112,26 @@ char *sailcov_file = NULL;
 #endif
 
 static struct option options[] = {
-    {"enable-dirty-update",         no_argument,       0, 'd'                     },
-    {"enable-misaligned",           no_argument,       0, 'm'                     },
-    {"pmp-count",                   required_argument, 0, OPT_PMP_COUNT           },
-    {"pmp-grain",                   required_argument, 0, OPT_PMP_GRAIN           },
-    {"ram-size",                    required_argument, 0, 'z'                     },
-    {"disable-compressed",          no_argument,       0, 'C'                     },
-    {"disable-writable-misa",       no_argument,       0, 'I'                     },
-    {"disable-fdext",               no_argument,       0, 'F'                     },
-    {"disable-vector-ext",          no_argument,       0, 'W'                     },
-    {"mtval-has-illegal-inst-bits", no_argument,       0, 'i'                     },
-    {"device-tree-blob",            required_argument, 0, 'b'                     },
-    {"terminal-log",                required_argument, 0, 't'                     },
-    {"show-times",                  required_argument, 0, 'p'                     },
-    {"report-arch",                 no_argument,       0, 'a'                     },
-    {"test-signature",              required_argument, 0, 'T'                     },
-    {"signature-granularity",       required_argument, 0, 'g'                     },
+    {"device-tree-blob",      required_argument, 0, 'b'             },
+    {"terminal-log",          required_argument, 0, 't'             },
+    {"show-times",            required_argument, 0, 'p'             },
+    {"report-arch",           no_argument,       0, 'a'             },
+    {"test-signature",        required_argument, 0, 'T'             },
+    {"signature-granularity", required_argument, 0, 'g'             },
 #ifdef RVFI_DII
-    {"rvfi-dii",                    required_argument, 0, 'r'                     },
+    {"rvfi-dii",              required_argument, 0, 'r'             },
 #endif
-    {"help",                        no_argument,       0, 'h'                     },
-    {"trace",                       optional_argument, 0, 'v'                     },
-    {"no-trace",                    optional_argument, 0, 'V'                     },
-    {"trace-output",                required_argument, 0, OPT_TRACE_OUTPUT        },
-    {"inst-limit",                  required_argument, 0, 'l'                     },
-    {"enable-zfinx",                no_argument,       0, 'x'                     },
-    {"enable-bitmanip",             no_argument,       0, 'B'                     },
-    {"enable-writable-fiom",        no_argument,       0, OPT_ENABLE_WRITABLE_FIOM},
-    {"enable-svinval",              no_argument,       0, OPT_ENABLE_SVINVAL      },
-    {"enable-zcb",                  no_argument,       0, OPT_ENABLE_ZCB          },
-    {"enable-zicbom",               no_argument,       0, OPT_ENABLE_ZICBOM       },
-    {"enable-zicboz",               no_argument,       0, OPT_ENABLE_ZICBOZ       },
-    {"enable-zvkb",                 no_argument,       0, OPT_ENABLE_ZVKB         },
-    {"cache-block-size",            required_argument, 0, OPT_CACHE_BLOCK_SIZE    },
+    {"help",                  no_argument,       0, 1               },
+    {"config",                required_argument, 0, 'c'             },
+    {"print-default-config",  no_argument,       0, OPT_PRINT_CONFIG},
+    {"trace",                 optional_argument, 0, 'v'             },
+    {"no-trace",              optional_argument, 0, 'V'             },
+    {"trace-output",          required_argument, 0, OPT_TRACE_OUTPUT},
+    {"inst-limit",            required_argument, 0, 'l'             },
 #ifdef SAILCOV
-    {"sailcov-file",                required_argument, 0, 'c'                     },
+    {"sailcov-file",          required_argument, 0, OPT_SAILCOV     },
 #endif
-    {0,                             0,                 0, 0                       }
+    {0,                       0,                 0, 0               }
 };
 
 static void print_usage(const char *argv0, int ec)
@@ -209,17 +192,6 @@ static void read_dtb(const char *path)
   fprintf(stdout, "Read %zd bytes of DTB from %s.\n", dtb_len, path);
 }
 
-// Return log2(x), or -1 if x is not a power of 2.
-static int ilog2(uint64_t x)
-{
-  for (unsigned i = 0; i < sizeof(x) * 8; ++i) {
-    if (x == (UINT64_C(1) << i)) {
-      return i;
-    }
-  }
-  return -1;
-}
-
 /**
  * Parses the command line arguments and returns the argv index for the first
  * ELF file that should be loaded. As getopt transforms the argv array, all
@@ -230,20 +202,12 @@ static int ilog2(uint64_t x)
 static int process_args(int argc, char **argv)
 {
   int c;
-  uint64_t ram_size = 0;
-  uint64_t pmp_count = 0;
-  uint64_t pmp_grain = 0;
-  uint64_t block_size_exp = 0;
+  bool have_config = false;
   while (true) {
     c = getopt_long(argc, argv,
                     "a"
-                    "B"
                     "d"
                     "m"
-                    "C"
-                    "I"
-                    "F"
-                    "W"
                     "i"
                     "p"
                     "z:"
@@ -252,91 +216,23 @@ static int process_args(int argc, char **argv)
                     "T:"
                     "g:"
                     "h"
+                    "c:"
 #ifdef RVFI_DII
                     "r:"
 #endif
-#ifdef SAILCOV
-                    "c:"
-#endif
                     "V::"
                     "v::"
-                    "l:"
-                    "x",
+                    "l:",
                     options, NULL);
     if (c == -1)
       break;
     switch (c) {
     case 'a':
-      report_arch();
-      break;
-    case 'B':
-      fprintf(stderr, "enabling B extension.\n");
-      rv_enable_bext = true;
-      break;
-    case 'd':
-      fprintf(stderr, "enabling dirty update.\n");
-      rv_enable_dirty_update = true;
-      break;
-    case 'm':
-      fprintf(stderr, "enabling misaligned access.\n");
-      rv_enable_misaligned = true;
-      break;
-    case OPT_PMP_COUNT:
-      pmp_count = atol(optarg);
-      fprintf(stderr, "PMP count: %" PRIu64 "\n", pmp_count);
-      if (pmp_count != 0 && pmp_count != 16 && pmp_count != 64) {
-        fprintf(stderr, "invalid PMP count: must be 0, 16 or 64");
-        exit(1);
-      }
-      rv_pmp_count = pmp_count;
-      break;
-    case OPT_PMP_GRAIN:
-      pmp_grain = atol(optarg);
-      fprintf(stderr, "PMP grain: %" PRIu64 "\n", pmp_grain);
-      if (pmp_grain >= 64) {
-        fprintf(stderr, "invalid PMP grain: must less than 64");
-        exit(1);
-      }
-      rv_pmp_grain = pmp_grain;
-      break;
-    case 'C':
-      fprintf(stderr, "disabling RVC compressed instructions.\n");
-      rv_enable_rvc = false;
-      break;
-    case 'I':
-      fprintf(stderr, "disabling writable misa CSR.\n");
-      rv_enable_writable_misa = false;
-      break;
-    case 'F':
-      fprintf(stderr, "disabling floating point (F and D extensions).\n");
-      rv_enable_fdext = false;
-      break;
-    case 'W':
-      fprintf(stderr, "disabling RVV vector instructions.\n");
-      rv_enable_vext = false;
-      break;
-    case 'i':
-      fprintf(stderr, "enabling storing illegal instruction bits in mtval.\n");
-      rv_mtval_has_illegal_inst_bits = true;
-      break;
-    case OPT_ENABLE_WRITABLE_FIOM:
-      fprintf(stderr,
-              "enabling FIOM (Fence of I/O implies Memory) bit in menvcfg.\n");
-      rv_enable_writable_fiom = true;
+      do_report_arch = true;
       break;
     case 'p':
       fprintf(stderr, "will show execution times on completion.\n");
       do_show_times = true;
-      break;
-    case 'z':
-      ram_size = atol(optarg);
-      if (ram_size) {
-        fprintf(stderr, "setting ram-size to %" PRIu64 " MB\n", ram_size);
-        rv_ram_size = ram_size << 20;
-      } else {
-        fprintf(stderr, "invalid ram-size '%s' provided.\n", optarg);
-        exit(1);
-      }
       break;
     case 'b':
       dtb_file = strdup(optarg);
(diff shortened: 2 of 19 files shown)


Timmmm (commented):
This is fantastic.

c_emulator/riscv_sim.c
```diff
@@ -431,6 +304,12 @@ static int process_args(int argc, char **argv)
       break;
     }
   }
+
+  if (!have_config) {
+    fprintf(stderr, "No configuration file provided.\n");
```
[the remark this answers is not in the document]
Alasdair: Does CMake have a built in way to do that?

c_emulator/riscv_sim.c
```diff
@@ -526,24 +425,43 @@ void init_sail_reset_vector(uint64_t entry)
...
-  rv_rom_size = rom_end - rv_rom_base;
+  /* calculate rom size */
+  uint64_t rom_size = rom_end - rom_base;
+
+  /* check calculated rom values match configuration */
+  sail_config_key rom_base_key = {"platform", "rom", "base"};
+  sail_config_key rom_size_key = {"platform", "rom", "size"};
+  if (rom_base != get_config_uint64(3, rom_base_key)) {
+    fprintf(stderr,
+            "Configuration value platform.rom.base does not match %" PRIu64
```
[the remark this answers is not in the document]
Alasdair: The way it used to work is it would set `rv_rom_base`, overriding the default value in the configuration (such as it existed previously). Now it's just checking the configuration setting is correct, not sure if it's the best thing to do.

model/riscv_insts_vext_vset.sail
```diff
@@ -53,7 +53,7 @@ function handle_illegal_vtype() = {
   print_reg("CSR vl <- " ^ BitStr(vl))
 }
 
-val sys_vext_vl_use_ceil = pure "sys_vext_vl_use_ceil" : unit -> bool
+function vl_use_ceil() -> bool = config extensions.v.vl_use_ceil
```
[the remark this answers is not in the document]
Alasdair: Yes, you could do that.

c_emulator/riscv_sim.c
```diff
@@ -496,6 +375,24 @@ uint64_t load_sail(char *f, bool main_file)
...
 }
 
+uint64_t get_config_uint64(size_t key_elems, sail_config_key key)
+{
+  sail_config_json json = sail_config_get(key_elems, key);
+  sail_int big_n;
+  uint64_t n;
+
+  if (!sail_config_is_int(json)) {
+    fprintf(stderr, "Configuration option");
```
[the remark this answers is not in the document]
Alasdair: Hmm not sure why I just stopped writing this error message.

I do also need to improve some of the error messages on the Sail side too, like when a configuration key is mispelled in the source.

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
arichardson: Could probably do this using objcopy or .incbin although that isn't too portable. I imagine a custom_command() that runs something like:
```
(echo 'extern "C" const char default_data[] = R"(' && ${CMAKE} -E cat ${CMAKE_SOURCE_DIR}/default.json && echo ')";')> file.cc
```

c_emulator/riscv_sim.c
```diff
@@ -526,24 +425,43 @@ void init_sail_reset_vector(uint64_t entry)
...
 
-  /* set rom size */
-  rv_rom_size = rom_end - rv_rom_base;
+  /* calculate rom size */
+  uint64_t rom_size = rom_end - rom_base;
+
+  /* check calculated rom values match configuration */
+  sail_config_key rom_base_key = {"platform", "rom", "base"};
+  sail_config_key rom_size_key = {"platform", "rom", "size"};
+  if (rom_base != get_config_uint64(3, rom_base_key)) {
```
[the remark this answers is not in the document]
arichardson: Why don't we just compile the model as C++?

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
Timmmm: Not in an especially nice way but something like this might work I guess? I think @arichardson has the misfortune of being more of a CMake expert than me.

```
file(READ "default.json" DEFAULT_JSON)
configure_file("default_json.h.in" "default_json.h" ESCAPE_QUOTES @ONLY)
# Ensure this runs again if `default.json` is changed.
set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "default_json.h.in")
```

```
// default_json.h.in
#pragma once
const char* DEFAULT_JSON = "@DEFAULT_JSON@";
```

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
Timmmm: Yeah I would support that. Then we can gradually move to nicer C++ code.

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
Timmmm: Oh I see. The whole ROM/RAM thing is wrong anyway so maybe we should just not worry about it too much for now. Once this is merged we can easily add a proper PMA memory map.

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
jordancarlin: Seems reasonable. Would also probably make integrating lockstep functionality easier, as we’ve discussed a few times.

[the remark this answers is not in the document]
Alasdair: I don't think there's a good way to do this in plain C. I originally thought this API would only be used by the code generator, and it's quite happy to just insert the extra number argument wherever needed. It's quite restrictive because it's targetting the intersection of C and C++ and array literals are one place where the languages differ (I had to modify my internal IR to add the concept of an initialiser value that can't be propagated thanks to C++'s rules 😞 ). I think if we want to switch the simulator language to C++ that should be another PR.

I could add a convenience function for human use that would look like `get_config("platform.rom.base")`.

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
Timmmm: That's what the `CMAKE_CONFIGURE_DEPENDS` does (in theory).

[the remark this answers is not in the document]
arichardson: Did not know about that! Sounds like the best solution then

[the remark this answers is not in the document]
Timmmm: Though if we change the code to be C++ I would be in favour of just having it hard-coded as a string literal:

```
#pragma once
constexpr DEFAULT_JSON = r#"
...
"#;
```

Much simpler and still very easy to copy/paste if people want to use it as a base.

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
arichardson: https://github.com/riscv/sail-riscv/pull/762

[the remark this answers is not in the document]
Alasdair: Is now `get_config_unit64("platform.rom.base")`, so marking this as resolved.

c_emulator/riscv_sim.c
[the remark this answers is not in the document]
arichardson: I guess the advantage of having both the file and the builtin default is that you can pass the file with --config=. But if we add something like `--print-default-config` you could run `$(SIM) --config <($(SIM) --print-default-config) ...`.

I guess the main advantage of having it as a real file in the repository is that we can get syntax highlighting+json formatting in IDEs.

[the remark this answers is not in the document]
arichardson: So I think the suggestion in https://github.com/riscv/sail-riscv/pull/758/files#r1972488749 is probably best?

model/riscv_sys_regs.sail
```diff
@@ -83,43 +83,50 @@ register misa : Misa =
...
+function sys_pmp_count() -> {0, 16, 64} = config memory.pmp.count
+
 /* G parameter that specifies the PMP grain size. The grain size is 2^(G+2), e.g.
    G=0 -> 4 bytes, G=10 -> 4096 bytes. */
-val sys_pmp_grain = pure "sys_pmp_grain" : unit -> range(0, 63)
+function sys_pmp_grain() -> range(0, 63) = config memory.pmp.grain
 
 /* Which HPM counters are supported (as a bit mask). Bits [2 .. 0] are ignored. */
-val sys_writable_hpm_counters = pure "sys_writable_hpm_counters" : unit -> bits(32)
+function sys_writable_hpm_counters() -> bits(32) = config base.writeable_hpm_counters
```
bacam, line 115: In case you need an example of why checking against a schema or default config is useful, this should read `base.writable_hpm_counters`.

Alasdair: Should be fixe now

Alasdair:
Updated so it uses the default configuration by default as suggested when one is not provided. A slight oversight is I didn't add a Sail API function to set the configuration to a string, so it has to write the default configuration to a temporary file before reading it... I'll fix that for the next Sail version but for now I think that's the best I can do.

Alasdair:
For the first party tests, I set it up so you can create a `<testname>.filter` file, and the configuration provided to the test will be the output of `jq -f <testname>.filter config/default.json`. We could probably do something more in future but that seems like enough for this PR.

jordancarlin:
This is looking great! Seems like a lot of the handwritten support files for theorem provers probably need to be updated after the changes to all of the sys_* functions?

Timmmm (approved):
LGTM. I feel like maybe we should defer `jq`/`.filter` to a future PR but I'm easy.

The `--pmp-count` was for #773 which I haven't got around to merging yet.

model/riscv_platform.sail
```diff
@@ -27,36 +27,30 @@ val elf_entry = pure {
...
 // because this model performs `cbo.zero` with a single write, and the behaviour
 // with cache blocks larger than a page is not clearly defined.
-val plat_cache_block_size_exp = pure {c: "plat_cache_block_size_exp", interpreter: "Platform.cache_block_size_exp", lem: "plat_cache_block_size_exp"} : unit -> range(0, 12)
+function plat_cache_block_size_exp() -> range(0, 12) = config platform.cache_block_size_exp
 
 /* Main memory */
-val plat_ram_base = pure {c: "plat_ram_base", interpreter: "Platform.dram_base", lem: "plat_ram_base"} : unit -> physaddrbits
-val plat_ram_size = pure {c: "plat_ram_size", interpreter: "Platform.dram_size", lem: "plat_ram_size"} : unit -> physaddrbits
+function plat_ram_base() -> physaddrbits = to_bits(physaddrbits_len, config platform.ram.base : int)
+function plat_ram_size() -> physaddrbits = to_bits(physaddrbits_len, config platform.ram.size : int)
```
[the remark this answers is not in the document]
Alasdair: I think that might be a bit tricky because of how `physaddrbits_len` changes.

Alasdair:
I felt it was probably important to have some way to have first_party tests control their config, and it seemed like the simplest way to get that in this PR.

c_emulator/riscv_sim.cpp
```diff
@@ -502,6 +379,46 @@ uint64_t load_sail(char *f, bool main_file)
...
+    char *c_part = (char *)sail_malloc(part.size() + 1);
+    strncpy(c_part, part.c_str(), part.size() + 1);
+    parts.push_back(c_part);
+  }
+
+  sail_config_json json = sail_config_get(parts.size(), parts.data());
+
+  for (const char *c_part : parts) {
+    sail_free((void *)c_part);
+  }
```
arichardson, line 398: This function seems unnecessarily complex. How about the following and the callers can just do `get_config_uint64({"platform", "rom", "size"})`

```suggestion
uint64_t get_config_uint64(std::vector<const char *> keypath)
{
  sail_config_json json = sail_config_get(keypath.size(), keypath.data());
```

c_emulator/riscv_sim.cpp
```diff
@@ -439,6 +301,21 @@ static int process_args(int argc, char **argv)
...
+    std::filesystem::path path = std::filesystem::temp_directory_path();
+    pid_t pid = getpid();
+    std::ostringstream filename;
+    filename << "default_config" << pid << ".json";
+    path = path / filename.str();
+    std::ofstream tmp(path);
+    tmp << DEFAULT_JSON;
+    tmp.close();
+
+    sail_config_set_file(path.c_str());
```
arichardson, line 315: Would be a bit nicer if we could just pass it a string directly but I guess the tempfile is fine.

Could you also add a flag like "--print-default-config" that outputs the default .json file?

sail_runtime/CMakeLists.txt
```diff
@@ -15,8 +15,15 @@ add_library(sail_runtime
...
     "${sail_dir}/lib/sail_failure.h"
     "${sail_dir}/lib/sail_coverage.h"
+    "${sail_dir}/lib/cJSON.c"
+    "${sail_dir}/lib/cJSON.h"
+    "${sail_dir}/lib/sail_config.c"
+    "${sail_dir}/lib/sail_config.h"
 )
 
+add_compile_options(-fsanitize=undefined)
+add_link_options(-fsanitize=undefined)
```
arichardson, line 25: This should not be the default

arichardson (changes requested):
UBSan needs to be removed, otherwise looking good.

c_emulator/riscv_sim.cpp
Alasdair: Added now

c_emulator/riscv_sim.cpp
Alasdair: Done

Alasdair:
Accidental UBSan should be removed now, and the --print-default-config option has been added, along with the simplification to the config access function.

config/default.json
```diff
@@ -0,0 +1,78 @@
...
+            "size" : 4096
+        },
+        "clint" : {
+            "base" : 33554432,
+            "size" : 786432
+        },
+        "instructions_per_tick" : 100
+    },
+    "extensions" : {
+        "c" : {
```
jordancarlin, line 41: Unless I’m mistaken, JSON is case sensitive. Assuming that’s the case, do we want these single letter extension names to be capitalized?

radimkrcmar: The ISA strings are case insensitive.  I think it would make sense to follow the "Chapter 39. ISA Extension Naming Conventions", which capitalizes single letter extensions.

jordancarlin: Agreed. Especially because these aren’t being specified in an isa string. Each extension is being listed separately.

c_emulator/riscv_sim.cpp
```diff
@@ -502,6 +384,38 @@ uint64_t load_sail(char *f, bool main_file)
...
+uint64_t get_config_uint64(std::vector<const char *> keypath)
+{
+  sail_config_json json = sail_config_get(keypath.size(), keypath.data());
+
+  if (!json) {
+    std::cerr << "Failed to find configuration option '";
+    for (auto part : keypath) {
+      std::cerr << "." << part;
+    }
+    std::cerr << "'.";
```
arichardson, line 396:
```suggestion
    std::cerr << "'.\n";
```

c_emulator/riscv_sim.cpp
```diff
@@ -502,6 +384,38 @@ uint64_t load_sail(char *f, bool main_file)
...
+
+  sail_int big_n;
+  uint64_t n;
+
+  if (!sail_config_is_int(json)) {
+    std::cerr << "Configuration option '";
+    for (auto part : keypath) {
+      std::cerr << "." << part;
+    }
+    std::cerr << "' could not be parsed as an integer.";
```
arichardson, line 408:
```suggestion
    std::cerr << "' could not be parsed as an integer.\n";
```

c_emulator/riscv_sim.cpp
```diff
@@ -532,23 +447,40 @@ void init_sail_reset_vector(uint64_t entry)
...
   /* boot at reset vector */
-  zPC = rv_rom_base;
+  zPC = rom_base;
 }
 
 void init_sail(uint64_t elf_entry)
 {
   zinit_model(UNIT);
   if (rvfi) {
+    /*
```
arichardson, line 475: Looks like this breaks RVFI, do we need a separate config file? Or can we override the sail variables here?

arichardson (approved):
Would be good to make RVFI work again but otherwise LGTM.

c_emulator/riscv_sim.cpp
Alasdair: You would need a separate configuration file with these values.

config/default.json
Timmmm: The canonical ISA string - as vaguely defined by the ISA manual and sort of implemented by riscv-opcodes - has every extension start with a capital letter.

For some reason GCC rejects that though and requires the entire string to be lowercased.

pmundkur:
Some instructions about configuration need to be added to the README (or in a separate file pointed to by the README).

config/default.json
Alasdair: They are capitalised now

c_emulator/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED On)
```
Timmmm, line 2: Can remove these now.

model/riscv_platform.sail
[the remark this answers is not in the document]
Timmmm: Ah right... That makes sense. We should probably have a safe version of `to_bits()` that doesn't just truncate. Happy for it to be in a future PR though.

c_emulator/CMakeLists.txt
Alasdair: Yep, rebased on top of that commit

pmundkur:
The `--report-arch` option is broken.  To fix it, the default config needs to be loaded and `model_init()` called before `zxlen_val` gets defined.

config/default.json
```diff
@@ -0,0 +1,78 @@
...
+            "base" : 2147483648,
+            "size" : 2147483648
+        },
+        "rom" : {
+            "base" : 4096,
+            "size" : 4096
+        },
+        "clint" : {
+            "base" : 33554432,
+            "size" : 786432
```
arichardson, line 36: Does the JSON parser support hex literals? IMO this would be easier to read.

jrtc27: It's using cJSON, which looks like it does (uses strtod), but then Sail itself re-converts the string via mpz_set_str and hard-codes a base of 10: https://github.com/rems-project/sail/blob/ca6092efe10b5c915185cfc82db9cf560901408d/lib/sail_config.c#L260

Also, JSON doesn't officially have hex literals anyway, so that would be relying on an extension.

But then using JSON numbers for these things might be a bad idea anyway, although JSON itself places no restrictions on the ranges of numbers, implementations often behave like JavaScript and treat everything as a double, so can't represent all possible 64-bit integers. cJSON does, but Sail is extracting the string and re-parsing it, so I guess it's not subject to that limitation, though it's a somewhat gross way for it to work.

Alasdair: If it's a Sail bitvector you can encode it as a string, but Sail integers are always JSON integers. The problem with using numbers encoded as strings is you can't really say much (really anything) useful about them in a Schema.

The JSON should be standard JSON without any extensions so Schema validators can read it (Arbitrary precision support is mandatory in JSON schema implementations, so issues with silently truncating to doubles won't happen there)

bacam:
> Seems like a lot of the handwritten support files for theorem provers probably need to be updated after the changes to all of the sys_* functions?

Hopefully it will just be deleting the current ones afterwards and not worry about them anymore, because they'll be provided by the configuration instead.  (In future it might be nice to have the generated theorem prover code be parametrised over some or all of the configuration, but we're not worrying too much about that at the moment.)

jordancarlin:
The README “Configuring platform options” section needs to be updated with these changes. It still mentions several command line flags.

arichardson:
It would be good to fix the RVFI build but considering how much this unblocks other work I think that can be done later. Users of RVFI can always use the commit prior to this being merged.

That said it shouldn't be too difficult to add another .json file with the different hardcoded start addresses. Since everything else should be mostly the same maybe it makes sense for that .json file to be a CMake template that just fills in the two different versions.

Or even better: Maybe there is a way to parse the json, modify it in-memory and then print the modified json to the temporary file for RVFI? Not sure if that can easily be done with cjson.

arichardson:
> It would be good to fix the RVFI build but considering how much this unblocks other work I think that can be done later. Users of RVFI can always use the commit prior to this being merged.
> 
> That said it shouldn't be too difficult to add another .json file with the different hardcoded start addresses. Since everything else should be mostly the same maybe it makes sense for that .json file to be a CMake template that just fills in the two different versions.
> 
> Or even better: Maybe there is a way to parse the json, modify it in-memory and then print the modified json to the temporary file for RVFI? Not sure if that can easily be done with cjson.

I guess the other option would be to do some ugly string replacement.

pmundkur:
If we need an external script to process a config (e.g. to automatically enable extension dependencies for an extension), we could make that script also output an RVFI config.  That way, we're not restricted to cjson.

pmundkur (approved):

pmundkur:
It would be good to update the README, but we could do it in a subsequent PR.

jordancarlin:
> It would be good to update the README, but we could do it in a subsequent PR.

Agreed. Shouldn’t block getting this merged since so much other work depends on it.

jordancarlin (approved):

model/riscv_platform.sail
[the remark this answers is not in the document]
Timmmm: Captured in #614

c_emulator/riscv_sim.cpp
Timmmm: I think if RVFI needs specific values it's probably not unreasonable for it to need a special config file. I think it's fine to leave this for now.

PR #762, opened 2025-02-28
Build most of the C simulator as C++

arichardson:
This will make it easier to add more features in the future.

While touching this code also fix various warnings.

History is intended to be clean, but happy to change this.


CMakeLists.txt
```diff
@@ -71,6 +76,11 @@ if (DOWNLOAD_GMP)
         URL https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz
         INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/gmp
         CONFIGURE_COMMAND "<SOURCE_DIR>/configure" "--prefix=<INSTALL_DIR>" --disable-shared --enable-static
+        # The libgmp.a output is created by the install step but since GMP::GMP
+        # depends on the entire ExternalProject build, we can pretend it is
+        # created by the build.
+        # TODO: Once we depend on CMake 3.26, use INSTALL_BYPRODUCTS
+        BUILD_BYPRODUCTS "<INSTALL_DIR>/lib/libgmp.a"
```
arichardson, line 83: This change is unrelated and I can pull it into a separate commit. Just noticed it while building with many jobs.

Timmmm: Does something break if you don't have this then?

Timmmm (approved):
This is great.

CMakeLists.txt
arichardson: I got a build failure with DOWNLOAD_GMP and `ninja -j 64`. Should I submit it as a separate PR?

Timmmm: Nah I was just trying to understand it. I still don't understand why it fails without this though. The docs just say:

> Files that will be generated by the build command but may or may not have their modification time updated by subsequent builds.

arichardson:
Hmm we could use -nostdlib++ for now since we don't use any library features yet? Or alternatively -static-libstdc++. Let me make this change

arichardson:
Updated to use -static-libstdc++

arichardson:
Do we also need static unwind lib? -static-libgcc?

arichardson:
> Do we also need static unwind lib? -static-libgcc?

Added now so it's back to 
```
ldd cmake-build-release/c_emulator/riscv_sim_rv64d
        linux-vdso.so.1 (0x00007f834fe91000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f834f80a000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f834fe93000)
```

Timmmm:
Nice. Is the binary still a reasonable size?

arichardson:
> Nice. Is the binary still a reasonable size?

```
❯ size cmake-build-release/c_emulator/riscv_sim_rv64d
   text    data     bss     dec     hex filename
3103448   32328   77720 3213496  3108b8 cmake-build-release/c_emulator/riscv_sim_rv64d
❯ size cmake-build-release/c_emulator/riscv_sim_rv64d
   text    data     bss     dec     hex filename
3103084   32328   77752 3213164  31076c cmake-build-release/c_emulator/riscv_sim_rv64d
```

Adding a basic use of iostream only seems to add 300 bytes of text (maybe it's converted to fprintf). We will only pay for code size for the C++ features that we use in the future.

CMakeLists.txt
```diff
@@ -46,6 +51,16 @@ if (STATIC)
         list(INSERT CMAKE_FIND_LIBRARY_SUFFIXES 0 .lib .a)
     else()
         set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
+        check_linker_flag(CXX "-static-libstdc++" HAVE_STATIC_LIBSTDCXX)
+        if (NOT HAVE_STATIC_LIBSTDCXX)
+            message(FATAL_ERROR "Building with -DSTATIC+TRUE requires static C++ std library installation")
```
Timmmm, line 56:
```suggestion
            message(FATAL_ERROR "Building with -DSTATIC=TRUE requires static C++ std library installation")
```

Presumably?

CMakeLists.txt
```diff
@@ -46,6 +51,16 @@ if (STATIC)
...
+        endif()
+        add_link_options(-static-libstdc++)
+        check_linker_flag(CXX "-static-libgcc" HAVE_STATIC_LIBGCC)
+        if (NOT HAVE_STATIC_LIBGCC)
+            message(FATAL_ERROR "Building with -DSTATIC+TRUE requires static libgcc installation")
```
Timmmm, line 61: Ditto? Or is this some sorted syntax I've never seen?

CMakeLists.txt
arichardson: Ah yes, fixed now.

Timmmm (approved):

Timmmm:
LGTM, there was support for this in the meeting so I'll add the 'will be merged' tag but I'll give it a week or so since it's a significant change.

jordancarlin (approved):
LGTM

arichardson:
rebased to fix conflict.

Alasdair:
After this change I get `fatal error: bracket nesting level exceeded maximum of 256`

Alasdair:
Apple clang

jordancarlin:
Looks like gcc supports up to 1024 with no option to change it but clang defaults to 256 and has a flag to increase it. We should probably add a check to the CMake file that adds this option if it detects that it’s running with clang.

Might be worth compiling the model with both GCC and clang in CI to catch these sorts of things.

Timmmm:
Oh I'm getting it too now with Clang (but not GCC). I think it's actually because of #617, but yeah let's add a flag.

Also, it occurs to me that we didn't actually check how compile time and performance were affected by #617 - it could potentially be significant I guess.

Timmmm:
#780  - I'm just checking the compile time too...

Alasdair:
I can do something like:
```
      let cdefs = List.map (Jib_optimize.flatten_cdef ~max_depth:100) cdefs in
```
to flatten any deeply nested C definitions in Sail during code generation.

Alasdair:
It might also be the `newtype regidx` change that triggers it in addition to this, as Sail generates lots of extra code that just wraps and unwraps the type. I'll add the optimisation to make the `newtype` wrappers transparent.

PR #764, opened 2025-02-28
Add first party C tests

Timmmm:
Add first party C tests that can be used for tests specific to this model. This uses Clang as the cross-compiler.

This includes a copy of `nanoprinf` for printing to the UART.


jordancarlin:
I definitely prefer this to the Zig implementation, even if the CMake file is a bit messier. Ideally it would work with both gcc and clang, especially because a lot of other riscv projects point users very strongly towards gcc (like the riscv-arch-test documentation).

test/first_party/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+        set(CROSS_COMPILER_COMMAND_RV64 ${CLANG_BIN} --target=riscv64)
+    else()
+        message(WARNING "Your Clang compiler does not support RISC-V (see '${CLANG_BIN} -print-targets'). Please download one from https://github.com/llvm/llvm-project/releases")
+    endif()
+elseif (GCC_BIN_RV64)
+    set(CROSS_COMPILER_COMMAND_RV64 ${GCC_BIN_RV64})
+    if (GCC_BIN_RV32)
+        set(CROSS_COMPILER_COMMAND_RV32 ${GCC_BIN_RV32})
+    else()
+        # Hope it has multilib support. I'm not sure how to test that.
```
jordancarlin, line 51: The `-v` flag seems to report if it was compiled with multilib, so we could try to parse that. I'm also fine just letting it error like we have right now.

```
$ riscv64-unknown-elf-gcc -v
Using built-in specs.
COLLECT_GCC=riscv64-unknown-elf-gcc
COLLECT_LTO_WRAPPER=/opt/riscv/libexec/gcc/riscv64-unknown-elf/14.2.0/lto-wrapper
Target: riscv64-unknown-elf
Configured with: /opt/riscv/riscv-gnu-toolchain/gcc/configure --target=riscv64-unknown-elf --prefix=/opt/riscv --disable-shared --disable-threads --enable-languages=c,c++ --with-pkgversion= --with-system-zlib --enable-tls --with-newlib --with-sysroot=/opt/riscv/riscv64-unknown-elf --with-native-system-header-dir=/include --disable-libmudflap --disable-libssp --disable-libquadmath --disable-libgomp --disable-nls --disable-tm-clone-registry --src=.././gcc --enable-multilib --with-multilib-generator='rv32e-ilp32e--;rv32i-ilp32--;rv32im-ilp32--;rv32iac-ilp32--;rv32imac-ilp32--;rv32imafc-ilp32f--;rv32imafdc-ilp32d--;rv64i-lp64--;rv64ic-lp64--;rv64iac-lp64--;rv64imac-lp64--;rv64imafdc-lp64d--;rv64im-lp64--;' --with-abi=lp64d --with-arch=rv64gc --with-tune=rocket --with-isa-spec=20191213 'CFLAGS_FOR_TARGET=-Os    -mcmodel=medlow' 'CXXFLAGS_FOR_TARGET=-Os    -mcmodel=medlow'
Thread model: single
Supported LTO compression algorithms: zlib zstd
gcc version 14.2.0 () 
```

test/first_party/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+    )
+
+    # Check if `riscv32` is present in the output
+    if (clang_targets MATCHES "riscv32")
+        set(CROSS_COMPILER_COMMAND_RV32 ${CLANG_BIN} --target=riscv32)
+        set(CROSS_COMPILER_COMMAND_RV64 ${CLANG_BIN} --target=riscv64)
+    else()
+        message(WARNING "Your Clang compiler does not support RISC-V (see '${CLANG_BIN} -print-targets'). Please download one from https://github.com/llvm/llvm-project/releases")
+    endif()
+elseif (GCC_BIN_RV64)
```
jordancarlin, line 46: Seems like right now this will fail if a system has clang that does not support RISC-V even if they have a RISC-V gcc compiler installed. This seems like a somewhat likely scenario (it's what we have set up on our RHEL servers).

test/first_party/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+            COMMAND ${CROSS_COMPILER_COMMAND_RV${xlen}}
+                # The ISA string to compile for.
+                -march=rv${xlen}gc
+                # Calling convention to use. Valid values are 'ilp32' or 'lp64' for 32/64-bit,
+                # optionally followed by 'f' or 'd' for hard-float. All combinations are valid.
+                -mabi=${mabi}d
+                # Indicate we are building in a standalone environment. Implies -fno-builtin
+                # so the compiler won't assume that memcpy etc. are available.
+                -ffreestanding
+                # Don't try to link with libc or libc.
```
jordancarlin, line 97: libc or ???

test/first_party/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+    "src/common/crt0.S"
+    "src/common/encoding.h"
+    "src/common/link.ld"
+    "src/common/nanoprintf.c"
+    "src/common/nanoprintf.h"
+    "src/common/runtime.c"
+    "src/common/runtime.h"
+)
+
+set(tests
```
jordancarlin, line 68: Would probably be easier to have it find all c and assembly files in the test directory instead of needing to manually list them here. That way this file won't need modifications as tests are added. It should be as simple as adding the assembly or c file to the appropriate directory.

test/first_party/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
...
+        endif()
+
+        set(elf "${arch}_${test_source}.elf")
+
+        add_custom_command(
+            OUTPUT ${elf}
+            DEPENDS ${common_deps} "src/${test_source}"
+            COMMAND ${CROSS_COMPILER_COMMAND_RV${xlen}}
+                # The ISA string to compile for.
+                -march=rv${xlen}gc
```
jordancarlin, line 90: As our tests evolve we'll need a more comprehensive march string. We use `rv64gcv_zcb_zfa_zba_zbb_zbc_zbs_zfh_zicboz_zicbop_zicbom_zicond_zbkb_zbkx_zknd_zkne_zknh` for our tests. We really want to enable every extension so that any instruction can be used in a test file for the model.

jordancarlin (commented):
Thanks for adding GCC. A few more comments.

test/first_party/CMakeLists.txt
trdthg: But some extensions are incompatible, such as Zcf and Zclsd, each test should be able to specify its own match string

One example is gcc's test framework, you can specify it in the comments (for reference)

[gcc/testsuite/gcc.target/riscv/cmo-zicbom-1.c](https://github.com/gcc-mirror/gcc/blob/4856292f7a680ec478e7607f1b71781996d7d542/gcc/testsuite/gcc.target/riscv/cmo-zicbom-1.c#L2)
```
/* { dg-do compile } */
/* { dg-options "-march=rv64gc_zicbom -mabi=lp64" } */
...
```

jordancarlin: That is a very good point. Not sure that we need that level of complexity yet in the first version of this PR (would rather get some testing in instead of waiting for it to be perfect and continue with almost no tests), but definitely something we will need to address eventually.

trdthg: completely agree, would be great to see this merged soon.

test/first_party/CMakeLists.txt
Timmmm: Yeah unfortunately we'd need to know *which* ISAs multilib was enabled for. Forgive me if I don't want to parse this!

```
--with-multilib-generator='rv32e-ilp32e--;rv32i-ilp32--;rv32im-ilp32--;rv32iac-ilp32--;rv32imac-ilp32--;rv32imafc-ilp32f--;rv32imafdc-ilp32d--;rv64i-lp64--;rv64ic-lp64--;rv64iac-lp64--;rv64imac-lp64--;rv64imafdc-lp64d--;rv64im-lp64--;'
```

My version has

```
--with-multilib-generator='rv64imafdc-lp64d--v*zicsr*zifencei;rv32imafc-ilp32f--zicsr*zifencei;rv32iafc_zmmul-ilp32f--zicsr*zifencei;rv32e-ilp32e--m*a*c*zicsr*zifencei'
```

test/first_party/CMakeLists.txt
Timmmm: Ooops good point! I'll fix that.

test/first_party/CMakeLists.txt
Timmmm: Think it was meant to be `libm`. I'll fix it.

test/first_party/CMakeLists.txt
Timmmm: [CMake recommends against that with some good reasons](https://cmake.org/cmake/help/latest/command/file.html#:~:text=We%20do%20not%20recommend%20using,to%20ask%20CMake%20to%20regenerate.), and I also think it's better to be explicit in general.

test/first_party/CMakeLists.txt
Timmmm: Yeah my plan was to embed a list of `march` and model configs inside the source files and then we would run each test for all of those. Probably worth waiting until after the new config system is done though.

test/first_party/CMakeLists.txt
Timmmm: Ok this should work now.

test/first_party/CMakeLists.txt
Timmmm: Done

test/first_party/CMakeLists.txt
jordancarlin: Fine with me.

test/first_party/CMakeLists.txt
jordancarlin: Seems reasonable. We should probably open an issue to track this after this PR is merged.

test/first_party/CMakeLists.txt
jordancarlin: I believe any compiler that has the `--enable-multilib` flag has 32 bit support irrespective of which specific extra configs are enabled. I’m fine leaving this as it is though, so no need to pursue it.

test/first_party/CMakeLists.txt
```diff
@@ -0,0 +1,141 @@
...
+if (GCC_BIN_RV32)
+    if (NOT CROSS_COMPILER_COMMAND_RV32)
+        set(CROSS_COMPILER_COMMAND_RV32 ${GCC_BIN_RV32})
+    endif()
+endif()
+
+if (GCC_BIN_RV64)
+    if (NOT CROSS_COMPILER_COMMAND_RV32)
+        # It might support multilib.
+        set(CROSS_COMPILER_COMMAND_RV32 ${GCC_BIN_RV64})
```
arichardson, line 58:
```suggestion
        set(CROSS_COMPILER_COMMAND_RV32 ${GCC_BIN_RV64} -m32)
```
Isn't this needed to target RV32?

jordancarlin: I believe `-m32` is specific to x86. For riscv, this is all handled by the march string.

arichardson: -m32 works for many architectures, but you are right using -march=rv32... will have the same effect. Please ignore.

test/first_party/src/common/nanoprintf.h
```diff
@@ -0,0 +1,1333 @@
...
+#endif
+#pragma GCC diagnostic ignored "-Wpadded"
+#pragma GCC diagnostic ignored "-Wfloat-equal"
+#if NANOPRINTF_CLANG
+#pragma GCC diagnostic ignored "-Wc++98-compat-pedantic"
+#pragma GCC diagnostic ignored "-Wcovered-switch-default"
+#pragma GCC diagnostic ignored "-Wdeclaration-after-statement"
+#pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
+#ifndef __APPLE__
+#pragma GCC diagnostic ignored "-Wunsafe-buffer-usage"
```
arichardson, line 169: This is triggering a warning and needs a version check...

Playing around with https://godbolt.org/z/rxxovj9nd, it seems it was added in clang 16

arichardson:
I like this, just a few minor comments (and haven't managed to review all the asm code)

Timmmm:
I figured out the `lga` thing - in the code I copied it from it was fine with just `la` but for some reason I was getting a relocation out of range error.

It was because I forgot to set `__global_pointer$` in the linker script, and I guess it defaults to 0 or something, but the code starts at like 0x8000000000 so it was out of the 2GB range limit on `la`. Setting `__global_pointer$` to be near `.text` means we can just use `la`.

I'll fix the other issues mentioned in a bit.

test/first_party/src/common/nanoprintf.h
Timmmm: I added this, hopefully it's ok...

```
#ifndef __APPLE__
#if defined(__clang__) && __clang_major__ >= 16
#pragma GCC diagnostic ignored "-Wunsafe-buffer-usage"
#endif
#endif
```

Timmmm: No wait that's not right! Should be `#if !defined(__clang__) || __clang_major__ >= 16`...

Timmmm:
I removed the max_pmp test. I'll add it with my PR that fixes the PMP code.

jordancarlin (approved):
LGTM at this point. There are still improvements to be made (especially with ISA strings), but those can be part of a follow up PR. This will be great to get in so new tests can easily be added for the model.

arichardson (approved):
LGTM, this is great and any improvements can be made in follow-up changes

PR #769, opened 2025-03-06
Fix missing initialisation for minstret[h]_write

Timmmm:
Fixes a warning about missing initialisation here.


jordancarlin (approved):
LGTM

PR #770, opened 2025-03-06
Expose VLEN/ELEN as CLI options.

pmundkur:
This is needed for some of the vector test suites.

### DIFF
--- a/c_emulator/riscv_platform_impl.cpp
+++ b/c_emulator/riscv_platform_impl.cpp
@@ -6,6 +6,7 @@
 uint64_t rv_pmp_count = 0;
 uint64_t rv_pmp_grain = 0;
 
+// These should be less than 0x10 (16) according to the "V" 1.0 spec.
 uint64_t rv_vector_vlen_exp = 0x9;
 uint64_t rv_vector_elen_exp = 0x6;
 
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -10,6 +10,7 @@
 #include <sys/time.h>
 #include <fcntl.h>
 #include <optional>
+#include <assert.h>
 
 #include "elf.h"
 #include "sail.h"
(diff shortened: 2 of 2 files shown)


jordancarlin:
Not sure if we want to just wait for json configuration support to be merged unless there is an urgent need for this.

pmundkur:
@nadime15 has been testing the Vector extension (see #790) and this is useful to enable wider testing instead of just the default VLEN/ELEN.

pmundkur:
@Alasdair any thoughts on how the config process would work for these settings?   We could discuss this on Monday.

pmundkur:
I suggest we move forward on reviewing and merging this PR to enable wider testing of the vector extension that is already in progress thanks to @nadime15.  We can port it over to the upcoming config system once we have the config for the defaults in place.

nadime15 (commented):
Since check_vector_lens() is always called, you could move it outside the if-else block, either before or after and consider exiting immediately instead of returning a boolean.

Edit: Sorry I forgot to highlight any lines, here is the code I mean:

```
static void set_vector_lens(uint64_t vlen, uint64_t elen)
{
  if (vlen > 0 && elen == 0) {
    assert(ilog2(rv_vector_elen_exp) > 0);
    elen = p2(rv_vector_elen_exp);
    if (!check_vector_lens(vlen, elen)) {
      exit(1);
    }
    rv_vector_vlen_exp = (uint64_t)ilog2(vlen);
  } else if (vlen == 0 && elen > 0) {
    assert(ilog2(rv_vector_vlen_exp) > 0);
    vlen = p2(rv_vector_vlen_exp);
    if (!check_vector_lens(vlen, elen)) {
      exit(1);
    }
    rv_vector_elen_exp = (uint64_t)ilog2(elen);
  } else {
    if (!check_vector_lens(vlen, elen)) {
      exit(1);
    }
    rv_vector_vlen_exp = (uint64_t)ilog2(vlen);
    rv_vector_elen_exp = (uint64_t)ilog2(elen);
  }
}
```

pmundkur:
True. As discussed above, this will change anyway once part of the config workflow.   I'd rather get something we can use merged quickly.

pmundkur:
Closing this as obsolete given the new config system.

PR #772, opened 2025-03-07
README updates

jordancarlin:
- Remove list of ISAs that have Sail implementations. This is already in the Sail repo README and doesn't contribute anything to the RISC-V specific README
- Refer to the [Extension Roadmap](https://github.com/riscv/sail-riscv/wiki/Extension-Roadmap) Wiki page for the list of features that is not yet supported. I've spent a bunch of time lately going through this Wiki and updating it with a list of (as far as I can tell) just about every ratified extension that we don't yet support. The only extensions I didn't bother to include are many of the profile defined extensions that are really just architectural options.
- Update sample Sail code to match the current implementation of the ITYPE instructions.

The other thing that is currently out of date with the README is the directory structure. I could update it, but I'd be inclined to just remove it because I don't think it provides much value and it will continue to fall out of date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,6 @@ interactive proof (in Isabelle, HOL4, and Coq), support proof about binary code 
   <img width="800" src="https://www.cl.cam.ac.uk/~pes20/sail/overview-sail.png?">
 <p>
 
-Sail is being used for multiple ISA descriptions, including
-essentially complete versions of the sequential behaviour of Arm-A
-(automatically derived from the authoritative Arm-internal
-specification, and released under a BSD Clear licence with Arm's
-permission), RISC-V, CHERI-RISC-V, CHERIoT, MIPS, and CHERI-MIPS; all these are complete
-enough to boot various operating systems.  There are also Sail models
-for smaller fragments of IBM POWER and x86, including a version of the ACL2 x86 model automatically translated from that.
-
 Getting started
 ---------------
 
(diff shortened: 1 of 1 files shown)


Timmmm:
LGTM, and I agree we can remove the directory structure.

Timmmm (approved):

pmundkur:
I'm not sure removing the directory structure is a good idea.  It's a useful orientation and pretty easy to keep up-to-date, unlike the rest of the documentation.  The other changes LGTM.

README.md
```diff
@@ -302,21 +287,6 @@ The files under `handwritten_support` provide library definitions for
 Coq, Isabelle and HOL4.
 
 
-Directory Structure
```
arichardson, line 305: Could keep this, I don't think it adds much maintenance burden

Timmmm: Yeah it's the sort of thing that people forget to update though. E.g. I didn't update it when adding the first party tests.

I also think it's of little value. The `doc` directory contains documentation. The `tests` directory contains tests, etc.

But if people really want to keep it I don't mind.

pmundkur: It's a bit like forgetting to update the list of supported extensions in the README: easy to fix after the fact, and useful for newcomers.

jordancarlin:
Added directory structure back and updated it to match the current state of the repo

pmundkur (approved):

arichardson (approved):

PR #773, opened 2025-03-07
Fix incorrect behaviour with INT_MAX pmpaddr

### LINKED ISSUE #739: Incorrect behaviour with INT_MAX pmpaddr
If you set `pmpaddr` to `ones()` then it will misbehave. For NAPOT it will return `(zeros(), zeros())`. For NA4 it will return `(ones(), zeros())`. In both cases it will match no addresses, whereas it should match all memory, or the top 4 bytes respectively.

I have a fix for this, and a test (I know!). The test is written in Zig... I was experimenting with that. I'll demo it in the next meeting I'm at (24th).

Timmmm:
This fixes a bug if you set `pmpaddr` to `ones()`. For NAPOT it would return `(zeros(), zeros())`. For NA4 it would return `(ones(), zeros())`. In both cases it will match no addresses, whereas it should match all memory, or the top 4 bytes respectively.

A test is included for NAPOT that fails before and passes afterwards.

Fixes #739

### DIFF
--- a/test/first_party/src/test_max_pmp.c
+++ b/test/first_party/src/test_max_pmp.c
@@ -1,0 +1,46 @@
+#include "common/runtime.h"
+
+#include <stdint.h>
+#include <stddef.h>
+
+#define PMPCFG_L (0b1 << 7)
+#define PMPCFG_NA4 (0b10 << 3)
+#define PMPCFG_NAPOT (0b11 << 3)
+#define PMPCFG_X (0b1 << 2)
+#define PMPCFG_W (0b1 << 1)
+#define PMPCFG_R (0b1 << 0)
+
+#define MSTATUS_MPP_MASK (0b11 << 11)
+#define MSTATUS_MPRV_MASK (0b1 << 17)
+
+volatile unsigned GLOBAL = 1;
+
+int main()
+{
+  printf("Testing 0xFF..FF PMP addresses\n");
+
+  uint_xlen_t ones = UINT_XLEN_MAX;
+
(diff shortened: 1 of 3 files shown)


pmundkur (approved):
LGTM.

The `it` in `it would return` in the commit message wasn't clear.  It may help to also clarify that the overflow problem was solved by converting to `nat` before doing range arithmetic.

test/first_party/src/test_max_pmp.c
```diff
@@ -0,0 +1,41 @@
...
+  asm volatile("csrc mstatus, %[mpp];"
+               "csrs mstatus, %[mprv];"
+               :
+               : [mpp] "r"(mpp), [mprv] "r"(mprv));
+
+  uint_xlen_t pmpcfg_napot = PMPCFG_NAPOT | PMPCFG_X | PMPCFG_W | PMPCFG_R;
+
+  asm volatile("csrw pmpcfg0, %[cfg]" : : [cfg] "r"(pmpcfg_napot));
+
+  // Access memory to test loads/stores.
```
arichardson, line 38: Add a volatile load of some global volatile variable here instead of relying on printf side effects?

arichardson (commented):
Did not review the sail change in detail but it looks reasonable. One small comment on the test, but overall this lgtm

test/first_party/src/test_max_pmp.c
Timmmm: Will do.

Timmmm: Done.

arichardson (approved):

PR #777, opened 2025-03-07
Implement Zawrs.

pmundkur:
This implements #398 by @ved-rivos  on top of the refactor in #755.

It should address the concerns raised in #398, and was one of the motivating reasons for #755.


ved-rivos (approved):

pmundkur:
Rebased on master; ready for review.

Makefile.old
```diff
@@ -69,6 +70,7 @@ SAIL_DEFAULT_INST += riscv_insts_zicbom.sail
 SAIL_DEFAULT_INST += riscv_insts_zicboz.sail
 SAIL_DEFAULT_INST += riscv_insts_zvbb.sail
 SAIL_DEFAULT_INST += riscv_insts_zvbc.sail
+SAIL_DEFAULT_INST += riscv_insts_zvknhab.sail
```
ved-rivos, line 73: This change seems unrelated to this PR.

Makefile.old
```diff
@@ -111,8 +113,9 @@ SAIL_ARCH_SRCS += riscv_extensions.sail riscv_types_common.sail riscv_types_ext.
 SAIL_ARCH_SRCS += riscv_vmem_types.sail $(SAIL_REGS_SRCS) $(SAIL_SYS_SRCS) riscv_platform.sail
 SAIL_ARCH_SRCS += riscv_sstc.sail
 SAIL_ARCH_SRCS += riscv_mem.sail $(SAIL_VM_SRCS)
-SAIL_ARCH_RVFI_SRCS = $(PRELUDE) rvfi_dii.sail riscv_extensions.sail riscv_types_common.sail riscv_types_ext.sail riscv_types.sail riscv_vmem_types.sail $(SAIL_REGS_SRCS) $(SAIL_SYS_SRCS) riscv_platform.sail riscv_mem.sail $(SAIL_VM_SRCS) riscv_types_kext.sail riscv_inst_retire.sail
+SAIL_ARCH_RVFI_SRCS = $(PRELUDE) rvfi_dii.sail riscv_extensions.sail riscv_types_common.sail riscv_types_ext.sail riscv_types.sail riscv_vmem_types.sail $(SAIL_REGS_SRCS) $(SAIL_SYS_SRCS) riscv_platform.sail riscv_mem.sail $(SAIL_VM_SRCS) riscv_types_kext.sail riscv_zvk_utils.sail riscv_inst_retire.sail
 SAIL_ARCH_SRCS += riscv_types_kext.sail    # Shared/common code for the cryptography extension.
+SAIL_ARCH_SRCS += riscv_zvk_utils.sail
```
ved-rivos, line 118: This change seems unrelated to this PR.

Makefile.old
```diff
@@ -111,8 +113,9 @@ SAIL_ARCH_SRCS += riscv_extensions.sail riscv_types_common.sail riscv_types_ext.
 SAIL_ARCH_SRCS += riscv_vmem_types.sail $(SAIL_REGS_SRCS) $(SAIL_SYS_SRCS) riscv_platform.sail
 SAIL_ARCH_SRCS += riscv_sstc.sail
 SAIL_ARCH_SRCS += riscv_mem.sail $(SAIL_VM_SRCS)
-SAIL_ARCH_RVFI_SRCS = $(PRELUDE) rvfi_dii.sail riscv_extensions.sail riscv_types_common.sail riscv_types_ext.sail riscv_types.sail riscv_vmem_types.sail $(SAIL_REGS_SRCS) $(SAIL_SYS_SRCS) riscv_platform.sail riscv_mem.sail $(SAIL_VM_SRCS) riscv_types_kext.sail riscv_inst_retire.sail
+SAIL_ARCH_RVFI_SRCS = $(PRELUDE) rvfi_dii.sail riscv_extensions.sail riscv_types_common.sail riscv_types_ext.sail riscv_types.sail riscv_vmem_types.sail $(SAIL_REGS_SRCS) $(SAIL_SYS_SRCS) riscv_platform.sail riscv_mem.sail $(SAIL_VM_SRCS) riscv_types_kext.sail riscv_zvk_utils.sail riscv_inst_retire.sail
```
ved-rivos, line 116: This change seems unrelated to this PR.

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,28 @@
...
+
+/* ****************************************************************** */
+union clause ast = WRS : (wrsop)
+
+mapping clause encdec = WRS(WRS_STO)
+  <-> 0b000000011101 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
+  when currentlyEnabled(Ext_Zawrs)
+
+function clause execute (WRS(WRS_STO)) =
+  return RETIRE_FAIL(Enter_Wait(WAIT_WRS_STO))
```
ved-rivos, line 21: Why always RETIRE_FAIL?

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,28 @@
...
+
+mapping clause encdec = WRS(WRS_NTO)
+  <-> 0b000000001101 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
+  when currentlyEnabled(Ext_Zawrs)
+
+function clause execute (WRS(WRS_NTO)) =
+  return RETIRE_FAIL(Enter_Wait(WAIT_WRS_NTO))
```
ved-rivos, line 28: Why always RETIRE_FAIL?

Makefile.old
pmundkur: This `Makefile.old` is not really used for building (hence it's name); its only use is to manually run SMT property checks.   I wanted to make sure these still passed so I updated it.

Makefile.old
pmundkur: See above.

Makefile.old
pmundkur: See above.

model/riscv_insts_zawrs.sail
pmundkur: Yeah, this is a confusing name.  The `RETIRE_FAIL` is the return code to the stepper, which handles the embedded reason.   This PR could be a good place to rename it.

model/riscv_insts_zawrs.sail
pmundkur: See above.

model/riscv_insts_zawrs.sail
ved-rivos: I think so. Perhaps something like WAIT_INTERRUPTED

model/riscv_step.sail
```diff
@@ -37,36 +38,74 @@ union Step = {
...
+      // less-privileged mode, and it does not complete within an
+      // implementation-specific, bounded time limit, the WFI
+      // instruction causes an illegal-instruction exception."
+      if   (cur_privilege == Machine | mstatus[TW] == 0b0)
+      then (Step_Execute(RETIRE_OK(), instbits), true)
+      else (Step_Execute(RETIRE_FAIL(Illegal_Instruction()), instbits), true)
+    },
+    (_, WAIT_WRS_STO, _, true) => {
+      if   get_config_print_instr()
+      then print_instr("timed-out exit from WRS-STO state at PC " ^ BitStr(PC));
```
ved-rivos, line 86: Nit: It may be either timeout or the reservation set becoming invalid or an interrupt.

model/riscv_step.sail
```diff
@@ -37,36 +38,74 @@ union Step = {
...
+    (_, WAIT_WRS_STO, false, _) => {
+      if   get_config_print_instr()
+      then print_instr("reservation invalid exit from WRS-STO state at PC " ^ BitStr(PC));
+
+      hart_state = HART_ACTIVE();
+      (Step_Execute(RETIRE_OK(), instbits), true)
+    },
+    (_, WAIT_WRS_NTO, false, _) => {
+      if   get_config_print_instr()
+      then print_instr("reservation invalid exit from WRS-NTO state at PC " ^ BitStr(PC));
```
ved-rivos, line 65: Nit: It may be reservation set becoming invalid or an interrupt.

model/riscv_step.sail
```diff
@@ -37,36 +38,74 @@ union Step = {
...
-    if   get_config_print_instr()
-    then print_instr("forced exit from WAIT state at PC " ^ BitStr(PC));
+      // The waiting instruction retires successfully.  The
+      // pending interrupts will be handled in the next step.
+      hart_state = HART_ACTIVE();
+      (Step_Execute(RETIRE_OK(), instbits), true)
+    },
+    (_, WAIT_WRS_STO, false, _) => {
+      if   get_config_print_instr()
+      then print_instr("reservation invalid exit from WRS-STO state at PC " ^ BitStr(PC));
```
ved-rivos, line 58: Nit: It may be reservation set becoming invalid or an interrupt.

ved-rivos:
@pmundkur lgtm. Noted a few comments.

model/riscv_step.sail
pmundkur: The interrupt case is handled in the match expression's first clause `(true, _, _, _)`; it's common to WFI, and WRS_{STO, NTO}.  The invalid reservation set is handled in the `(_, WAIT_WRS_STO, false, _)` and `(_, WAIT_WRS_NTO, false, _)` clauses for `STO` and `NTO` respectively.

model/riscv_step.sail
pmundkur: See above.

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,28 @@
...
+
+/* This file specifies the instructions in the 'Zawrs' extension.  */
+
+function clause currentlyEnabled(Ext_Zawrs) = hartSupports(Ext_Zawrs)
+
+/* ****************************************************************** */
+union clause ast = WRS : (wrsop)
+
+mapping clause encdec = WRS(WRS_STO)
+  <-> 0b000000011101 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
```
jordancarlin, line 17: Any reason not to do something like this instead:
```
mapping encdec_wrsop : wrsop <-> bits(12) = {
  WRS_STO <-> 0b000000011101,
  WRS_NTO <-> 0b000000001101,
}

mapping clause encdec = WRS(op)
  <-> encdec_wrsop(op) @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
```

That is how a lot of other extensions are set up and reduces duplicate code between the two. Similar for the execute clause.

model/riscv_insts_zawrs.sail
pmundkur: How about `EXEC_INCOMPLETE` instead of `RETIRE_FAIL` (and `EXEC_COMPLETE` for `RETIRE_SUCCESS`)?  Or less loudly, `Exec_Incomplete` and `Exec_complete`?  It'll affect a lot of files: I could put it in a separate commit in this PR.

model/riscv_insts_zawrs.sail
pmundkur: Good idea, done.

config/default.json
```diff
@@ -36,7 +36,15 @@
             "size" : 786432
         },
         "instructions_per_tick" : 100,
-        "wfi_is_nop" : true
+        "wfi_is_nop" : true,
+        "wrs" : {
+            "nto" : {
```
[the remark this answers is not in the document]
pmundkur: Good idea, done.

model/riscv_insts_zawrs.sail
jordancarlin: That seems like a much better match for what these actually signify.

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,33 @@
...
+}
+
+mapping clause encdec = WRS(op)
+  <-> encdec_wrsop(op) @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
+  when currentlyEnabled(Ext_Zawrs)
+
+function clause execute (WRS(WRS_STO)) =
+  return RETIRE_FAIL(Enter_Wait(WAIT_WRS_STO))
+
+mapping clause encdec = WRS(WRS_NTO)
```
jordancarlin, line 28: I don't think this is needed now that you have the `WRS(op)` approach above.

model/riscv_insts_zawrs.sail
ved-rivos: I am not sure we need to change all RETIRE_FAIL/RETIRE_SUCCESS. Since this is a special case of WAIT completed - it could have been WAIT_COMPLETE or WAIT_DONE or WAIT_INTERRUPTED - and then based on reason for the WAIT completion it may transform into successful retirement or the instruction or raising an exception (e.g., based on TW).

model/riscv_insts_zawrs.sail
pmundkur: Hmm, it's not in my local version.  I guess I removed it after pushing.

model/riscv_insts_zawrs.sail
pmundkur: We can't do that easily since we have a generic `RETIRE_FAIL` variant that wraps the specific cases.   I'll push a version with the big change isolated to a single commit to get feedback.

model/riscv_insts_zawrs.sail
```diff
@@ -0,0 +1,29 @@
...
+  WRS_STO <-> 0b000000011101,
+  WRS_NTO <-> 0b000000001101,
+}
+
+mapping clause encdec = WRS(op)
+  <-> encdec_wrsop(op) @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
+  when currentlyEnabled(Ext_Zawrs)
+
+function clause execute (WRS(WRS_STO)) =
+  return Enter_Wait(WAIT_WRS_STO)
```
jordancarlin, line 26: Any reason for the `return`? I don't think it's necessary here.

jordancarlin (commented):
One minor nit, but otherwise LGTM.

model/riscv_insts_zawrs.sail
pmundkur: Fixed.

jordancarlin (approved):

pmundkur:
I'd like to merge this early next week.  @ved-rivos 's issue with the `RETIRE_FAIL` wrapper was addressed by #884.

model/riscv_step.sail
```diff
@@ -15,36 +15,74 @@ union Step = {
...
-    // instruction causes an illegal-instruction exception."
-    if   (cur_privilege == Machine | mstatus[TW] == 0b0)
-    then Step_Execute(Retire_Success(), instbits)
-    else Step_Execute(Illegal_Instruction(), instbits)
-  } else {
-    if   get_config_print_instr()
-    then print_instr("remaining in WAIT state at PC " ^ BitStr(PC));
-    Step_Waiting()
+function run_hart_waiting(step_no : int, wr: WaitReason, instbits : instbits, exit_wait : bool) -> Step = {
+  match (shouldWakeForInterrupt(), wr, valid_reservation(), exit_wait) {
```
Timmmm, line 22: IMO this style of match is much more confusing than `if`/`else`.

```
match (boolean, .....) {
   (true, _, _, _) => then,
   ...
}
```

is literally an `if then` but ... odder. I guess we can refactor it later if you like.

model/riscv_step.sail
```diff
@@ -15,36 +15,74 @@ union Step = {
...
+
+      hart_state = HART_ACTIVE();
+      Step_Execute(Retire_Success(), instbits)
+    },
+    (_, WAIT_WRS_NTO, _, true) => {
+      if   get_config_print_instr()
+      then print_instr("timed-out exit from " ^ wait_name(WAIT_WRS_NTO) ^ " state at PC " ^ BitStr(PC));
+
+      hart_state = HART_ACTIVE();
+      if   cur_privilege == Machine
```
Timmmm, line 73: This is the same logic as line 57 but written in a more complex way. We should copy lines 57-59 here.

model/riscv_types.sail
```diff
@@ -300,6 +300,15 @@ enum extop_zbb = {SEXTB, SEXTH, ZEXTH}
...
 
+enum wrsop = {WRS_STO, WRS_NTO}
+
+enum WaitReason = {WAIT_WFI, WAIT_WRS_STO, WAIT_WRS_NTO}
+mapping wait_name : WaitReason <-> string = {
+  WAIT_WFI     <-> "WAIT-WFI",
+  WAIT_WRS_STO <-> "WAIT-WRS-STO",
+  WAIT_WRS_NTO <-> "WAIT-WRS-NTO",
+}
+
```
Timmmm, line 311: Might be worth an

```
overload to_str = {wait_name}
```

It's quite handy for debugging.

Timmmm (changes requested):
I think we should probably fix the "This is the same logic..." comment (it's very easy/minor) but other than that LGTM.

model/riscv_step.sail
pmundkur: Ok, I copied 57-59 for now, but as the TODO comment for the `else` case stated, it will have to be changed for VS/VU modes, and then will differ again.  WFI and WRS.NTO do behave differently with respect to causing `Illegal-Instruction`, as the spec explicitly notes.

model/riscv_types.sail
pmundkur: Good idea; added.

model/riscv_step.sail
pmundkur: Actually, that portion will change for Sdext anyway, so we could refactor it then.

pmundkur: I went ahead and refactored it, should make Sdext easier.

model/riscv_step.sail
```diff
@@ -15,36 +15,74 @@ union Step = {
...
 }
 
-function run_hart_waiting(step_no : nat, exit_wait : bool, instbits : instbits) -> Step = {
-  if shouldWakeForInterrupt() then {
-    if   get_config_print_instr()
-    then print_instr("interrupt exit from WAIT state at PC " ^ BitStr(PC));
+function run_hart_waiting(step_no : int, wr: WaitReason, instbits : instbits, exit_wait : bool) -> Step = {
+  // successfully interrupted waits
+  if  shouldWakeForInterrupt() then {
+    if  get_config_print_instr()
```
Timmmm, line 24: Slightly odd spacing here btw.

model/riscv_step.sail
```diff
@@ -92,6 +130,16 @@ function run_hart_active(step_no: nat) -> Step = {
 
 function wfi_is_nop() -> bool = config platform.wfi_is_nop
 
+function wrs_nto_is_nop() -> bool = config extensions.Zawrs.nto.is_nop
+function wrs_sto_is_nop() -> bool = config extensions.Zawrs.sto.is_nop
```
Timmmm, line 134: Btw you can use these directly. Not sure what style we want to go for (and no need to change it for now).

```
function wait_is_nop(wr : WaitReason) -> bool =
  match wr {
    WAIT_WFI     => wfi_is_nop(),
    WAIT_WRS_STO => config extensions.Zawrs.nto.is_nop,
    WAIT_WRS_NTO => config extensions.Zawrs.sto.is_nop,
  }
```

Might need `(config extensions.Zawrs.sto.is_nop : bool)` - Sail's type inference isn't great for `match`.

Timmmm (approved):
Thanks, this is definitely better. LGTM.

I added a few comments but they're optional.

model/riscv_step.sail
pmundkur: fixed

model/riscv_step.sail
pmundkur: nice; it worked without the annotation, thanks!

jordancarlin: Great! This is something probably worth cleaning up throughout the model to get rid of a lot of these unnecessary functions.

ved-rivos (approved):

PR #778, opened 2025-03-08
Mandatory vsetvli and vsetivli operands

### LINKED ISSUE #489: `vsetvli` vector tail agnostic and vector mask agnostic operands are mandatory
In the RVV 1.0 spec, there is this text (in 3.4.3):
> The assembly syntax adds two mandatory flags to the vsetvli instruction [...]

(I'm not sure why "adds" was used there.) The above text is complemented by a note:
> Prior to v0.9, when these flags were not specified on a `vsetvli`, they defaulted to mask-undisturbed/tail-undisturbed. The use of `vsetvli` without these flags is deprecated, however, and specifying a flag setting is now mandatory. The default should perhaps be tail-agnostic/mask-agnostic, so software has to specify when it cares about the non-participating elements, but given the historical meaning of the instruction prior to introduction of these flags, it was decided to always require them in future assembly code.

Note that the vector tail agnostic and vector mask agnostic operands are mandatory.

I cannot find an analogous statement about the `lmul` operand, but given its position in the middle of the list of operands, it would seem reasonable to infer that it is also mandatory.

Unfortunately, there are examples of `vsetvli` without these "mandatory" operands. In (6):
```
vsetvli rd, rs1, vtypei # rd = new vl, rs1 = AVL, vtypei = new vtype setting
```
in (6.1):
```
vsetvli t0, a0, e8          # SEW= 8, LMUL=1
vsetvli t0, a0, e8, m2      # SEW= 8, LMUL=2
vsetvli t0, a0, e32, mf2    # SEW=32, LMUL=1/2
```
in (7.9):
```
vsetvli t1, x0, e8, m8     # Maximum VLMAX
vlm.v v0, (a0)             # Load mask register
vsetvli x0, t0, <new type> # Restore vl (potentially already present)
```

In `model/riscv_insts_vext_vset.ml`, these operands are supported as optional:
```
mapping maybe_lmul_flag : string <-> bits(3) = {
  ""              <-> 0b000, /* m1 by default */
  sep() ^ "mf8"   <-> 0b101,
  sep() ^ "mf4"   <-> 0b110,
  sep() ^ "mf2"   <-> 0b111,
  sep() ^ "m1"    <-> 0b000,
  sep() ^ "m2"    <-> 0b001,
  sep() ^ "m4"    <-> 0b010,
  sep() ^ "m8"    <-> 0b011
}

mapping maybe_ta_flag : string <-> bits(1) = {
  ""           <-> 0b0, /* tu by default */
  sep() ^ "ta" <-> 0b1,
  sep() ^ "tu" <-> 0b0
}

mapping maybe_ma_flag : string <-> bits(1) = {
  ""           <-> 0b0, /* mu by default */
  sep() ^ "ma" <-> 0b1,
  sep() ^ "mu" <-> 0b0
}
[...]
mapping clause assembly = VSETVLI(ma, ta, sew, lmul, rs1, rd)
  <-> "vsetvli" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ sew_flag(sew) ^ maybe_lmul_flag(lmul) ^ maybe_ta_flag(ta) ^ maybe_ma_flag(ma)
```

I believe a spec-conforming Sail implementation should not have the "by default" cases in the "maybe" functions (and the "maybe" functions should be renamed to remove the "maybe").

(@XinlaiWan?)

jordancarlin:
Vector tail agnostic and vector mask agnostic operands are mandatory for `vsetvli` and `vsetivli`.

Resolves #489

### DIFF
--- a/model/riscv_insts_vext_vset.sail
+++ b/model/riscv_insts_vext_vset.sail
@@ -20,24 +20,23 @@ mapping sew_flag : string <-> bits(3) = {
 }
 
 mapping maybe_lmul_flag : string <-> bits(3) = {
-  ""              <-> 0b000, /* m1 by default */
   sep() ^ "mf8"   <-> 0b101,
   sep() ^ "mf4"   <-> 0b110,
   sep() ^ "mf2"   <-> 0b111,
   sep() ^ "m1"    <-> 0b000,
   sep() ^ "m2"    <-> 0b001,
   sep() ^ "m4"    <-> 0b010,
-  sep() ^ "m8"    <-> 0b011
+  sep() ^ "m8"    <-> 0b011,
+  forwards ""      => 0b000, /* m1 by default */
+
 }
 
-mapping maybe_ta_flag : string <-> bits(1) = {
-  ""           <-> 0b0, /* tu by default */
+mapping ta_flag : string <-> bits(1) = {
   sep() ^ "ta" <-> 0b1,
   sep() ^ "tu" <-> 0b0
 }
 
-mapping maybe_ma_flag : string <-> bits(1) = {
-  ""           <-> 0b0, /* mu by default */
+mapping ma_flag : string <-> bits(1) = {
   sep() ^ "ma" <-> 0b1,
   sep() ^ "mu" <-> 0b0
 }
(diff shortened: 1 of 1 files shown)


model/riscv_insts_vext_vset.sail
```diff
@@ -20,24 +20,22 @@ mapping sew_flag : string <-> bits(3) = {
...
-  ""              <-> 0b000, /* m1 by default */
   sep() ^ "mf8"   <-> 0b101,
   sep() ^ "mf4"   <-> 0b110,
   sep() ^ "mf2"   <-> 0b111,
   sep() ^ "m1"    <-> 0b000,
   sep() ^ "m2"    <-> 0b001,
   sep() ^ "m4"    <-> 0b010,
-  sep() ^ "m8"    <-> 0b011
+  sep() ^ "m8"    <-> 0b011,
+  ""              <-> 0b000, /* m1 by default */
```
Timmmm, line 30: This will change it to output `m1` instead of an empty string when disassembling. Is that intentional?

If so maybe use `forward "" -> 0b000,`?

jordancarlin: Good point, and I think we could go either way. I'm in favor of Sail being explicit and outputting the more verbose version of the assembly, but could be convinced either way.

Timmmm: Yeah I think I agree.

Timmmm: I think we should have this actually so it's clear that we never map 0b000 to "".

```
  forward ""            -> 0b000, /* m1 by default */
```

jordancarlin: Updated

pmundkur (approved):

Timmmm (approved):

PR #780, opened 2025-03-10
Increase bracket depth limit for Clang

Timmmm:
Since #617 we hit the default bracket depth limit of 256 with Clang (GCC's limit is hard-coded to 1024). This increases the limit on Clang so that it compiles.

### DIFF
--- a/c_emulator/CMakeLists.txt
+++ b/c_emulator/CMakeLists.txt
@@ -62,15 +62,16 @@ foreach (xlen IN ITEMS 32 64)
                 )
             endif()
 
-            # TODO: Enable warnings when we use the #include trick
-            # to include the generated Sail code. Currently it
-            # generates too many warnings to turn these on globally.
+            target_compile_options(riscv_sim_${arch} PRIVATE
+                $<$<CXX_COMPILER_ID:Clang,AppleClang>:-fbracket-depth=1024>
 
-            # target_compile_options(riscv_sim_${arch} PRIVATE
-            #     -Wall -Wextra
-            #     # Too annoying at the moment.
-            #     -Wno-unused-parameter
-            # )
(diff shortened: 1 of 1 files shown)


c_emulator/CMakeLists.txt
```diff
@@ -62,15 +62,16 @@ foreach (xlen IN ITEMS 32 64)
...
 
-            # target_compile_options(riscv_sim_${arch} PRIVATE
-            #     -Wall -Wextra
-            #     # Too annoying at the moment.
-            #     -Wno-unused-parameter
-            # )
+                # TODO: Enable warnings when we use the #include trick
+                # to include the generated Sail code. Currently it
+                # generates too many warnings to turn these on globally.
+                # -Wall -Wextra
```
arichardson, line 71: Ah I should have removed this in #762 now that is actually being added.

c_emulator/CMakeLists.txt
```diff
@@ -62,15 +62,16 @@ foreach (xlen IN ITEMS 32 64)
                 )
             endif()
 
-            # TODO: Enable warnings when we use the #include trick
-            # to include the generated Sail code. Currently it
-            # generates too many warnings to turn these on globally.
+            target_compile_options(riscv_sim_${arch} PRIVATE
+                $<$<CXX_COMPILER_ID:Clang,AppleClang>:-fbracket-depth=1024>
```
arichardson, line 66: Could also do a check_c_compiler_flag() and add it if it's supported?

arichardson (approved):
Seems fine to me although I'd have a slight preference for check_c_compiler_flag()

jordancarlin:
Seems like this might no longer be necessary after the latest changes to Sail? @Alasdair

PR #784, opened 2025-03-11
Move most of RVFI code into its own file.

pmundkur:
This unclutters the main simulator code considerably.

Tested with TestRig to ensure no regressions using 
```
./utils/scripts/runTestRIG.py -a manual -b manual  --implementation-A-port 8080 --implementation-B-port 8081
```

@arichardson Are there other RVFI invocations I should test?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -9,3 +9,4 @@ IndentWidth: 2
 PointerAlignment: Right
 SortIncludes: Never
 AlignTrailingComments: true
+AccessModifierOffset: -2
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -4,14 +4,12 @@
 #include <stdlib.h>
 #include <errno.h>
 #include <unistd.h>
-#include <arpa/inet.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/mman.h>
 #include <sys/time.h>
-#include <sys/socket.h>
-#include <netinet/ip.h>
 #include <fcntl.h>
+#include <optional>
 
 #include "elf.h"
 #include "sail.h"
@@ -22,6 +20,7 @@
 #include "riscv_platform.h"
 #include "riscv_platform_impl.h"
 #include "riscv_sail.h"
+#include "rvfi_dii.h"
 
 const char *RV64ISA = "RV64IMAC";
 const char *RV32ISA = "RV32IMAC";
@@ -63,14 +62,7 @@ FILE *trace_log = NULL;
 char *dtb_file = NULL;
 unsigned char *dtb = NULL;
 size_t dtb_len = 0;
-#ifdef RVFI_DII
-static bool rvfi_dii = false;
-/* Needs to be global to avoid the needed for a set-version packet on each trace
- */
-static unsigned rvfi_trace_version = 1;
-static int rvfi_dii_port;
-static int rvfi_dii_sock;
-#endif
+std::optional<rvfi_handler> rvfi;
 
 char *sig_file = NULL;
 uint64_t mem_sig_start = 0;
@@ -365,11 +357,11 @@ static int process_args(int argc, char **argv)
       print_usage(argv[0], 0);
       break;
 #ifdef RVFI_DII
-    case 'r':
-      rvfi_dii = true;
-      rvfi_dii_port = atoi(optarg);
-      fprintf(stderr, "using %d as RVFI port.\n", rvfi_dii_port);
+    case 'r': {
+      int rvfi_dii_port = atoi(optarg);
+      rvfi = rvfi_handler(rvfi_dii_port);
       break;
+    }
 #endif
     case 'V':
       set_config_print(optarg, false);
@@ -441,7 +433,7 @@ static int process_args(int argc, char **argv)
     }
   }
 #ifdef RVFI_DII
-  if (optind > argc || (optind == argc && !rvfi_dii))
+  if (optind > argc || (optind == argc && !rvfi))
     print_usage(argv[0], 0);
 #else
   if (optind >= argc) {
@@ -452,9 +444,7 @@ static int process_args(int argc, char **argv)
   if (dtb_file)
     read_dtb(dtb_file);
 
-#ifdef RVFI_DII
-  if (!rvfi_dii)
-#endif
+  if (!rvfi)
     fprintf(stdout, "Running file %s.\n", argv[optind]);
   return optind;
 }
@@ -544,8 +534,7 @@ void init_sail_reset_vector(uint64_t entry)
 void init_sail(uint64_t elf_entry)
 {
   zinit_model(UNIT);
-#ifdef RVFI_DII
-  if (rvfi_dii) {
+  if (rvfi) {
     rv_ram_base = UINT64_C(0x80000000);
     rv_ram_size = UINT64_C(0x800000);
     rv_rom_base = UINT64_C(0);
@@ -555,7 +544,6 @@ void init_sail(uint64_t elf_entry)
     rv_htif_tohost = UINT64_C(0);
     zPC = elf_entry;
   } else
-#endif
     init_sail_reset_vector(elf_entry);
 }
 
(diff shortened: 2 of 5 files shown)


c_emulator/riscv_sim.cpp
```diff
@@ -915,69 +760,15 @@ int main(int argc, char **argv)
...
-    printf("Connected\n");
-  } else
-    entry = load_sail(initial_elf_file, /*main_file=*/true);
-#else
-  uint64_t entry = load_sail(initial_elf_file, /*main_file=*/true);
-#endif
+  if (rvfi) {
+    int exit_code = rvfi->setup_socket(config_print_rvfi);
+    if (exit_code)
+      exit(exit_code);
```
arichardson, line 766:
```suggestion
    if (int exit_code = rvfi->setup_socket(config_print_rvfi))
      return exit_code;
```
We are in main, so can just return?

c_emulator/riscv_sim.cpp
```diff
@@ -991,17 +782,18 @@ int main(int argc, char **argv)
...
-#endif
+  } else {
+    do {
+      run_sail();
+      if (rvfi) {
+        /* Reset for next test */
+        reinit_sail(entry);
+      }
+    } while (rvfi);
+  }
```
arichardson, line 795: This should be the same as just the second part:

```
    do {
      run_sail();
      if (rvfi) {
        /* Reset for next test */
        reinit_sail(entry);
      }
    } while (rvfi);
```
no need for the if else anymore.

c_emulator/rvfi_dii.h
```diff
@@ -0,0 +1,31 @@
...
+typedef enum {
+  RVFI_prestep_continue,  // continue loop
+  RVFI_prestep_eof,       // Got EOF, delete rvfi and return
+  RVFI_prestep_end_trace, // Got EndOfTrace
+  RVFI_prestep_ok,        // Ready for step
+} rvfi_prestep_t;
+
+typedef void (*packet_reader_fn)(lbits *rop, unit);
+
+class rvfi_t {
```
arichardson, line 14:
```suggestion
class rvfi_handler {
```
Maybe something like this? But I don't really mind what name we choose.

c_emulator/riscv_sim.cpp
```diff
@@ -366,9 +359,8 @@ static int process_args(int argc, char **argv)
       break;
 #ifdef RVFI_DII
     case 'r':
-      rvfi_dii = true;
       rvfi_dii_port = atoi(optarg);
-      fprintf(stderr, "using %d as RVFI port.\n", rvfi_dii_port);
+      rvfi = new rvfi_t(rvfi_dii_port);
```
arichardson, line 363: We could avoid the need for heap allocation and the need to free if we use std::optional<rvfi_t>.
```suggestion
      rvfi = rvfi_t(rvfi_dii_port);
```

c_emulator/riscv_sim.cpp
```diff
@@ -63,14 +62,7 @@ FILE *trace_log = NULL;
...
 size_t dtb_len = 0;
-#ifdef RVFI_DII
-static bool rvfi_dii = false;
-/* Needs to be global to avoid the needed for a set-version packet on each trace
- */
-static unsigned rvfi_trace_version = 1;
-static int rvfi_dii_port;
-static int rvfi_dii_sock;
-#endif
+rvfi_t *rvfi = NULL;
```
arichardson, line 65:
```suggestion
std::optional<rvfi_t> rvfi;
```
Avoids the need for heap allocation and free.

c_emulator/rvfi_dii.cpp
```diff
@@ -0,0 +1,249 @@
...
+    fprintf(stderr, "Sending packet with length %zd... ", send_size);
+  }
+  if (send_size > 4096) {
+    fprintf(stderr, "Unexpected large packet size (> 4KB): %zd\n", send_size);
+    exit(1);
+  }
+  unsigned char bytes[send_size];
+  /* mpz_export might not write all of the null bytes */
+  memset(bytes, 0, sizeof(bytes));
+  mpz_export(bytes, NULL, -1, 1, 0, 0, *(packet.bits));
```
arichardson, line 120: Now that this code is C++ we could probably move most of the handling out of sail and into the callbacks. But not for this change.

c_emulator/riscv_sim.cpp
```diff
@@ -714,108 +643,22 @@ void run_sail(void)
...
-          fprintf(stderr, "Got EOF, exiting... ");
-        }
-        rvfi_dii = false;
+    if (rvfi) {
+      switch (rvfi->pre_step(config_print_rvfi)) {
+      case RVFI_prestep_continue:
+        continue;
+      case RVFI_prestep_eof:
+        delete rvfi;
+        rvfi = NULL;
```
arichardson, line 652:
```suggestion
        rvfi = std::nullopt;
```

arichardson (commented):
A few minor commments but this is definitely an improvement!

c_emulator/rvfi_dii.h
```diff
@@ -0,0 +1,32 @@
...
+  public:
+  rvfi_handler(int port);
+
+  int setup_socket(bool config_print);
+  uint64_t get_entry();
+  void send_trace(bool config_print);
+  rvfi_prestep_t pre_step(bool config_print);
+
+  private:
+  void get_and_send_packet(packet_reader_fn reader, bool config_print);
```
arichardson, line 24:
```suggestion
public:
  rvfi_handler(int port);

  int setup_socket(bool config_print);
  uint64_t get_entry();
  void send_trace(bool config_print);
  rvfi_prestep_t pre_step(bool config_print);

private:
  void get_and_send_packet(packet_reader_fn reader, bool config_print);
```

The clang format configuration is doing some weird indentation here. I wonder why the access modifiers are indented.

c_emulator/rvfi_dii.cpp
arichardson: I think `AccessModifierOffset: -2` in .clang-format should fix this.

arichardson (approved):
A few more minor things I noticed, otherwise LGTM.

arichardson (approved):

jordancarlin:
Generally looks like a great cleanup. Any thoughts on how this impacts #494 though?

c_emulator/rvfi_dii.h
```diff
@@ -0,0 +1,29 @@
...
+  RVFI_prestep_eof,       // Got EOF, delete rvfi and return
+  RVFI_prestep_end_trace, // Got EndOfTrace
+  RVFI_prestep_ok,        // Ready for step
+} rvfi_prestep_t;
+
+typedef void (*packet_reader_fn)(lbits *rop, unit);
+
+class rvfi_handler {
+public:
+  rvfi_handler(int port);
```
Timmmm, line 16: Should be `explicit`

c_emulator/rvfi_dii.h
```diff
@@ -0,0 +1,29 @@
...
+  int setup_socket(bool config_print);
+  uint64_t get_entry();
+  void send_trace(bool config_print);
+  rvfi_prestep_t pre_step(bool config_print);
+
+private:
+  void get_and_send_packet(packet_reader_fn reader, bool config_print);
+
+  unsigned trace_version = 1;
+  int dii_port;
```
Timmmm, line 27: Probably initialise this to -1 or 0 or something just to be safe.

c_emulator/riscv_sim.cpp
```diff
@@ -825,6 +665,8 @@ void run_sail(void)
         goto step_exception;
       flush_logs();
       KILL(sail_int)(&sail_step);
+      if (rvfi)
```
Timmmm, line 668: I don't think we have a clang-format rule for this yet but I strongly prefer requiring {} for single statement blocks. Less error prone.

c_emulator/rvfi_dii.cpp
```diff
@@ -0,0 +1,244 @@
...
+  fprintf(stderr, "using %d as RVFI port.\n", port);
+}
+
+uint64_t rvfi_handler::get_entry()
+{
+  return 0x80000000;
+}
+
+// returns zero on success
+int rvfi_handler::setup_socket(bool config_print)
```
Timmmm, line 34: Probably makes more sense to return a bool and return true on success?

c_emulator/rvfi_dii.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "sail.h"
+
+typedef enum {
```
Timmmm, line 5: I would use `enum class` here.

Timmmm (commented):
This is great. Have a few minor comments and obviously there's a load of C++-ification we can do in future.

I *think* this shouldn't make the state callback PR any harder.

.clang-format
```diff
@@ -9,3 +9,4 @@ IndentWidth: 2
 PointerAlignment: Right
 SortIncludes: Never
 AlignTrailingComments: true
+AccessModifierOffset: -2
```
jrtc27, line 12:
```suggestion
IndentAccessModifiers: false
```

jrtc27: Hm that's related but slightly different, never mind

Timmmm (approved):

jordancarlin (approved):
LGTM. #494 can be updated once this is merged.

PR #785, opened 2025-03-11
add enable-gates for Zvfh/Zvfhmin extensions.

wwwwwwOwO:
Zvfh/Zvfhmin and the V extension implement the same set of instructions, with the primary differences being in the handling of SEW.
This update addresses SEW handling as follows:
1. For basic instructions, Zvfh only supports SEW = 16, while the V extension supports SEW = 32/64. (`valid_fp_op` in `model/riscv_insts_vext_fp_utils.sail`)
2. For narrowing/widening instructions(`valid_fp_op_variable_width` in `model/riscv_insts_vext_fp_utils.sail`):
   - Both Zvfh/Zvfhmin and the V extension support SEW = 16.
   - SEW = 32 is only supported by the V extension.
   - Notably, Zvfh allows SEW = 8 for 8-bit integers conversion instructions.
3. Zvfhmin only supports `vfwcvt.f.f.v` and `vfncvt.f.f.w`. Therefore, the definition of these instructions has been separated from the other `cvt` instructions.

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -116,3 +116,8 @@ enum clause extension = Ext_Svpbmt
 
 // Cycle and Instret Privilege Mode Filtering
 enum clause extension = Ext_Smcntrpmf
+
+//  Vector Extension for Half-Precision Floating-Point
+enum clause extension = Ext_Zvfh
+// Vector Extension for Minimal Half-Precision Floating-Point
+enum clause extension = Ext_Zvfhmin
--- a/model/riscv_insts_vext_fp.sail
+++ b/model/riscv_insts_vext_fp.sail
@@ -27,8 +27,8 @@ mapping encdec_fvvfunct6 : fvvfunct6 <-> bits(6) = {
   FVV_VMUL       <-> 0b100100
 }
 
-mapping clause encdec = FVVTYPE(funct6, vm, vs2, vs1, vd)                                                       if extensionEnabled(Ext_V)
-  <-> encdec_fvvfunct6(funct6) @ vm @ encdec_vreg(vs2) @ encdec_vreg(vs1) @ 0b001 @ encdec_vreg(vd) @ 0b1010111 if extensionEnabled(Ext_V)
+mapping clause encdec = FVVTYPE(funct6, vm, vs2, vs1, vd)                                                       if have_vext_fp()
+  <-> encdec_fvvfunct6(funct6) @ vm @ encdec_vreg(vs2) @ encdec_vreg(vs1) @ 0b001 @ encdec_vreg(vd) @ 0b1010111 if have_vext_fp()
 
 function clause execute(FVVTYPE(funct6, vm, vs2, vs1, vd)) = {
   let rm_3b    = fcsr[FRM];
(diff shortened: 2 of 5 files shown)


jordancarlin:
If we’re already splitting up the floating-point components of the vector extension, I wonder if it makes sense to add explicit support for Zve32f and Zve64d (which add 32/64 bit vector floating point respectively) at the same time. V implies Zve64d.

jordancarlin:
> Yes and no. I can see a desire to implement a vector unit that doesn't do floating-point, or one that doesn't have double-precision (e.g. AI workloads that don't do fp arithmetic above 16 bits).
> 
> On the other hand, V requires F. Perhaps "Fmin/Dmin" are needed, and have V depend on those instead.

There are a number of different vector extensions other than just `V`. See section 32.18 of the unprivileged isa spec.

`Zve32x` and `Zve64x` have no vector floating point support, `Zve32f` and `Zve64f` have 32 bit floating point support, and `Zve64d` has 64 bit floating point support. The 32 vs 64 is based on supported EEW. These extensions are all targeted at smaller implementations.

The `V` extension depends on `Zve64d` and imposes additional constrains on VLEN and some multiplication instructions.

jordancarlin:
Now that the new config system is in place, this is a good one to revisit. The vector extensions need a somewhat significant overhaul to be able to support all of the different extensions instead of just V or nothing. @wwwwwwOwO, I know that is slightly beyond the scope of this original PR. Is it something you'd want to drive to completion?

wwwwwwOwO:
> Now that the new config system is in place, this is a good one to revisit. The vector extensions need a somewhat significant overhaul to be able to support all of the different extensions instead of just V or nothing. @wwwwwwOwO, I know that is slightly beyond the scope of this original PR. Is it something you'd want to drive to completion?

@jordancarlin Yes, I’m happy to take this on. Just to clarify, does this overhaul mean adding explicit support for the vector extensions listed in section 32.18 of the Unprivileged ISA spec (e.g., zve32x, zve32f, zve64x, zve64f, zve64d (V), zvfh, zvfhmin), rather than simply differentiating between integer and floating-point vector support?

jordancarlin:
> @jordancarlin Yes, I’m happy to take this on. Just to clarify, does this overhaul mean adding explicit support for the vector extensions listed in section 32.18 of the Unprivileged ISA spec (e.g., zve32x, zve32f, zve64x, zve64f, zve64d, V, zvfh, zvfhmin), rather than simply differentiating between integer and floating-point vector support?

@wwwwwwOwO Exactly. Now that the new config system is in place, we should have each of these be separate extensions that can be enabled/disabled so that the model can match any implementation that conforms to the spec.

sizn-sc:
Hope that this refactoring can be revived, because this also had a significant bugfix for narrowing converts with SEW=8, which trap with invalid-instruction on main. cc @nadime15 (this is similar to https://github.com/riscv/sail-riscv/pull/1141). Did it get caught with the vector tests?

nadime15:
As far as I can tell @wwwwwwOwO closed this and moved the code to #982. Like I have said [here](https://github.com/riscv/sail-riscv/pull/982#issuecomment-3089919354), it would be great to split the PR into smaller ones. So would be nice to reopen this again, its easier to review.

@sizn-sc Are we talking about an actual bug here, or just a feature that isnt implemented yet? Conversion instructions from float to integer like `vfncvt.[xu].f.w` arent defined for SEW=8 unless Zvfh is supported.

sizn-sc:
@nadime15

> Are we talking about an actual bug here, or just a feature that isnt implemented yet

Narrowing conversions *are* implemented for SEW=8, but are considered illegal. Best I can tell sail-riscv does mostly support Zvfh atm (though they are not gated behind configuration flags) and the only problem is that `illegal_fp_variable_width` does not take `Zvfh` into account. Modifying the check there to allow SEW=8 unbreaks that and the behavior matches that of spike. Is that a bug? Or is Zvfh not actually fully implemented?

nadime15:
@sizn-sc Yes you are right _some_ narrowing conversion instructions for SEW=8 are implemented but would trap.  

>  Modifying the check there to allow SEW=8 unbreaks that and the behavior matches that of spike. Is that a bug? Or is Zvfh not actually fully implemented?

Its not a bug, Zvfh just isnt supported yet, so narrowing conversions for SEW=8 are illegal, even if some of the code is already there. 

I just checked Spike’s code, and you need to enable Zvfh or Zvfhmin (or just Zvfh, since that includes Zvfhmin) to get SEW=8 support (at least in theory). Are you sure you ran Spike and Sail with the same extensions enabled? You might have turned on Zvfh in Spike.

nadime15:
@sizn-sc I just had a look at the vector test suite code and turns out they actually list `Zvfh` in the enabled extensions:

`rv${XLEN}gcv_zvbb_zvbc_zfh_zvfh_zvkg_zvkned_zvknha_zvksed_zvksh`

> Did it get caught with the vector tests?

I guess the right test case wasnt generated and thats how it slipped through!

PR #788, opened 2025-03-11
Bump required sail compiler version to 0.19 and remove zlib dependency

jordancarlin:
There are a whole lot of other small improvements to make once we bump to 0.19 (enabling the theorem provers in CI, when guards for encdec, etc.), but I figured I’d start with the dependency simplification since building the model has been such a big focus lately.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -230,7 +230,7 @@ foreach (xlen IN ITEMS 32 64)
                         # deprecated because it is error-prone.
                         --strict-var
                         # Minimum required Sail compiler version.
-                        --require-version 0.18
+                        --require-version ${SAIL_REQUIRED_VER}
                         # Optimisations.
                         -O --Oconstant-fold
                         # Cache Z3 results in z3_problems file to speed up incremental compilation.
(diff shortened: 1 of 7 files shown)


nadime15:
Would this be a good time to remove the old Makefile in this PR?

Timmmm (approved):

pmundkur (approved):

model/CMakeLists.txt
```diff
@@ -230,7 +230,7 @@ foreach (xlen IN ITEMS 32 64)
                         # deprecated because it is error-prone.
                         --strict-var
                         # Minimum required Sail compiler version.
-                        --require-version 0.18
+                        --require-version 0.19
```
arichardson, line 233: It's a bit awkward having to update this multiple times, could we add something like a `SAIL_REQUIRED_VER` variable in the toplevel CMakeLists.txt and use that here and in all the other cases?

If some targets have lower/higher minimum versions those can always use a custom definition but having a common variable seems like it will make maintainance easier.

arichardson:
Although I wonder if we should hold of on this one until the opam repository has 0.19? (https://github.com/ocaml/opam-repository/pull/27610)

model/CMakeLists.txt
jordancarlin: Good suggestion. Updated.

Timmmm:
It's in Opam now so I'll merge this :tada:

PR #791, opened 2025-03-14
[Ext_V] Simplily nf type constrains

trdthg:
Try to simplify some type constraints of ext_v

- Created a new type, `type nf_t = range(1, 8)`(any naming suggestions?), which can save a lot of type constraints like `0 < nf <= 8`
- Removed some type declarations that can be inferred automatically

There are still a few exceptions where nf_t cannot be used

1. One is that the spec restricts the optional values of `nf`
	
	> Only NFIELDS values of 1, 2, 4, 8 are supported, with other values reserved.
	> [31.7.9. Vector Load/Store Whole Register Instructions](https://riscv-specs.timhutt.co.uk/spec/20240411/unpriv-isa-asciidoc.html#_vector_loadstore_whole_register_instructions)

	```
	val process_vsre : forall 'f 'b 'n, ('f in {1, 2, 4, 8}) ...
	```

2. Another is that we need `'q` to mark the size

So I created a `type nf_v('q) = 'q > 0 & 'q <= 8` (looks bad)

```
val read_vreg_seg : forall 'n 'm 'p 'q, 'n >= 0 & 'm >= 0 & nf_v('q). (int('n), int('m), int('p), int('q), vregidx) -> vector('n, bits('q * 'm))
function read_vreg_seg(num_elem, SEW, LMUL_pow, nf, vrid) = {
  var vreg_list : vector('q, vector('n, bits('m))) = vector_init(vector_init(zeros()));
```

Any advice?

### DIFF
--- a/model/riscv_insts_vext_utils.sail
+++ b/model/riscv_insts_vext_utils.sail
@@ -10,6 +10,14 @@
 /* This file implements functions used by vector instructions.                     */
 /* ******************************************************************************* */
 
+/* NFIELDS Encoding
+ * typically, nf is in range 1 to 8, but for 'Vector Load/Store Whole Register Instructions',
+ * only NFIELDS values of 1, 2, 4, 8 are supported, with other values reserved.
+ */
+type nfields_range('q) = 'q > 0 & 'q <= 8
+type nfields_range_pow2('q) = 'q in { 1, 2, 4, 8 }
+type nfields = { 'q, nfields_range('q). int('q) }
+
 /* Vector mask mapping */
 mapping maybe_vmask : string <-> bits(1) = {
   ""              <-> 0b1, /* unmasked by default */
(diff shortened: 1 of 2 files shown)


model/riscv_insts_vext_utils.sail
```diff
@@ -10,6 +10,14 @@
 /* This file implements functions used by vector instructions.                     */
 /* ******************************************************************************* */
 
+/* NFIELDS Encoding
+ * typically, nf is in range 1 to 8, but for 'Vector Load/Store Whole Register Instructions',
+ * only NFIELDS values of 1, 2, 4, 8 are supported, with other values reserved.
+ */
+type nfields_range('q) = 'q > 0 & 'q <= 8
+type nfields_range_pow2('q) = 'q in { 1, 2, 4, 8 }
+type nfields = { 'q, nfields_range('q). int('q) }
```
Timmmm, line 19: Maybe `type nfields = range(1, 8)`? I guess it means you're duplicating the limit...

Timmmm (approved):
LGTM, one minor suggestion but I'd be happy without it.

model/riscv_insts_vext_utils.sail
trdthg: `nfields_range` is the range constraint, and it's also used in other place.

So I think `type nfields = nfields_range` is for reuse the limit?

pmundkur (approved):
Typo in commit (simplily -> simplify) but otherwise LGTM.

Issue #787, opened 2025-03-11
Extend access type to cache accesses


Timmmm:
For CHERI we need to extend `AccessType` to have entries for cache accesses. Here's what we have:

```
enum CacheAccessType = {
  CleanFlush,
  Inval,
  Zero,
}

union AccessType ('a : Type) = {
  Read            : 'a,
  Write           : 'a,
  // (Read, Write)
  ReadWrite       : ('a, 'a),
  Execute         : unit,
  Cache           : CacheAccessType,
}
```

Also the access type for Zicbom needs to be based on the original type of access, not what it is changed to by `menvcfg[CBIE]` etc.

Timmmm:
Completed in #1405

PR #792, opened 2025-03-14
Extend access type to cache accesses

KotorinMinami:
Added an implementation for #787, but I was wondering how to implement this TODO in the comment
> // TODO: This is incorrect since CHERI only requires at least one byte
> // to be in bounds here, whereas `ext_data_get_addr()` checks that all bytes
> // are in bounds. We will need to add a new function, parameter or access type.

### DIFF
--- a/model/riscv_types.sail
+++ b/model/riscv_types.sail
@@ -116,11 +116,18 @@ enum Retired = {RETIRE_SUCCESS, RETIRE_FAIL}
 
 /* memory access types */
 
+enum CacheAccessType = {
+  CleanFlush,
+  Inval,
+  Zero,
+}
+
 union AccessType ('a : Type) = {
   Read      : 'a,
   Write     : 'a,
   ReadWrite : ('a, 'a),
-  Execute   : unit
+  Execute   : unit,
+  Cache     : CacheAccessType,
 }
 
 enum word_width = {BYTE, HALF, WORD, DOUBLE}
--- a/model/riscv_vmem_pte.sail
+++ b/model/riscv_vmem_pte.sail
@@ -118,6 +118,8 @@ function check_PTE_permission(ac        : AccessType(ext_access_type),
                                        & (pte_W == 0b1)
                                        & ((pte_R == 0b1) | ((pte_X == 0b1) & mxr)),
       (Execute(),       User)       => (pte_U == 0b1) & (pte_X == 0b1),
+      (Cache(_),        User)       => (pte_U == 0b1)
+                                       & ((pte_R == 0b1) | ((pte_X == 0b1) & mxr)),
       (Read(_),         Supervisor) => ((pte_U == 0b0) | do_sum)
                                        & ((pte_R == 0b1) | ((pte_X == 0b1) & mxr)),
       (Write(_),        Supervisor) => ((pte_U == 0b0) | do_sum)
@@ -127,6 +129,8 @@ function check_PTE_permission(ac        : AccessType(ext_access_type),
                                        & ((pte_R == 0b1)
                                           | ((pte_X == 0b1) & mxr)),
       (Execute(),       Supervisor) => (pte_U == 0b0) & (pte_X == 0b1),
+      (Cache(_),        Supervisor) => ((pte_U == 0b0) | do_sum)
+                                       & ((pte_R == 0b1) | ((pte_X == 0b1) & mxr)),
       (_,               Machine)    => internal_error(__FILE__, __LINE__,
                                                       "m-mode mem perm check")};
   if success then PTE_Check_Success(())
@@ -147,7 +151,8 @@ function update_PTE_Bits forall 'pte_size, 'pte_size in {32, 64} . (
                              Execute()       => false,
                              Read(_)         => false,
                              Write(_)        => true,
-                             ReadWrite(_, _) => true
+                             ReadWrite(_, _) => true,
+                             Cache(_)        => false,
                            });
   // Update 'accessed'-bit?
   let update_a = (pte_flags[A] == 0b0);
--- a/model/riscv_vmem_types.sail
+++ b/model/riscv_vmem_types.sail
@@ -43,7 +43,8 @@ function accessType_to_str (a) =
     Read(_)            => "R",
     Write(_)           => "W",
     ReadWrite(_, _)    => "RW",
-    Execute()          => "X"
+    Execute()          => "X",
+    Cache(_)           => "C",
   }
 
 overload to_str = {accessType_to_str}
--- a/model/riscv_insts_zicbom.sail
+++ b/model/riscv_insts_zicbom.sail
@@ -32,6 +32,14 @@ mapping cbop_mnemonic : cbop_zicbom <-> string = {
   CBO_INVAL <-> "cbo.inval"
 }
 
+function cbop_to_access_type(cbop) : cbop_zicbom -> AccessType(ext_access_type) = {
+  match cbop {
+    CBO_CLEAN => Cache(CleanFlush),
+    CBO_FLUSH => Cache(CleanFlush),
+    CBO_INVAL => Cache(Inval),
+  }
+}
+
 mapping clause assembly = RISCV_ZICBOM(cbop, rs1)
   <-> cbop_mnemonic(cbop) ^ spc() ^ "(" ^ opt_spc() ^ reg_name(rs1) ^ opt_spc() ^ ")"
 
@@ -40,6 +48,7 @@ function process_clean_inval(rs1, cbop) = {
   let rs1_val = X(rs1);
   let cache_block_size_exp = plat_cache_block_size_exp();
   let cache_block_size = 2 ^ cache_block_size_exp;
+  let access_type = cbop_to_access_type(cbop);
 
   // Offset from rs1 to the beginning of the cache block. This is 0 if rs1
   // is aligned to the cache block, or negative if rs1 is misaligned.
@@ -48,10 +57,10 @@ function process_clean_inval(rs1, cbop) = {
   // TODO: This is incorrect since CHERI only requires at least one byte
   // to be in bounds here, whereas `ext_data_get_addr()` checks that all bytes
   // are in bounds. We will need to add a new function, parameter or access type.
-  match ext_data_get_addr(rs1, negative_offset, Read(Data), cache_block_size) {
+  match ext_data_get_addr(rs1, negative_offset, access_type, cache_block_size) {
     Ext_DataAddr_Error(e) => { ext_handle_data_check_error(e); RETIRE_FAIL },
     Ext_DataAddr_OK(vaddr) => {
-      let res : option(ExceptionType) = match translateAddr(vaddr, Read(Data)) {
+      let res : option(ExceptionType) = match translateAddr(vaddr, access_type) {
         TR_Address(paddr, _) => {
           // "A cache-block management instruction is permitted to access the
           // specified cache block whenever a load instruction or store instruction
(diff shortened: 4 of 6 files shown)


model/riscv_types.sail
```diff
@@ -116,11 +116,18 @@ enum Retired = {RETIRE_SUCCESS, RETIRE_FAIL}
 
 /* memory access types */
 
+enum CacheAccessType = {
+  CleanFlush,
+  Inval,
+  Zero,
```
arichardson, line 122: Isn't zeroing just a write?

Timmmm: It behaves slightly differently because the address of a CBO zero is aligned to the cache block size.

Timmmm:
Here's how we are doing it for CBOs:

https://github.com/CHERI-Alliance/sail-riscv/blob/de9279755a2fbce96314083bff884f21e6937992/model/riscv_insts_zicboz.sail#L34

https://github.com/CHERI-Alliance/sail-riscv/blob/de9279755a2fbce96314083bff884f21e6937992/model/riscv_insts_zicbom.sail#L48-L53

Basically you just have to make sure to pass in the correct access type, and for cbom it is based on the original access type rather than the modified one.

@tomaird that bit isn't super clear in the CHERI spec. Do you know where it came from?

tomaird:
Not sure I understand which bit isn't clear so I'll just explain the full rationale behind all this code...

The CHERI spec describes the following checks for the CBO instructions (amongst others):

```
CBO   | Bounds    | Permissions
--------------------------------
CLEAN | Any byte  | W+R
FLUSH | Any byte  | W+R
INVAL | Any byte  | W+R+ASR
ZERO  | All bytes | W
```

So for the bounds check, CLEN/FLUSH/INVAL all behave the same - it only needs a single byte to be in bounds.

But for the permissions checks it's a bit different, with `CBO.INVAL` requiring more permissions than the others.

The RISC-V spec describes that `CBO.INVAL` can also do a flush depending on `xenvcfg.cbie`, which in the model we treat as executing a `CBO.FLUSH`. But from the table above you can see that `CBO.FLUSH` has a looser permissions requirement than `CBO.INVAL`, so we need to make sure we do the permissions check based on what instruction actually executed, i.e. the `CBO.INVAL`, rather than what we're _effectively_ executing in the model. Hence the need to pass in the "original" access type.

Does that answer your question?

Timmmm:
Yeah I think my question is do you definitely still need ASR permission if INVAL executes as FLUSH. My understanding was that INVAL needs the extra permission because it's "unsafe" - you can resurrect capabilities that should have died. But if it's actually doing a FLUSH then it's safe.

Although now that I check the spec again it does actually explicitly mention this:

> The CBIE bit in [menvcfg](https://riscv.github.io/riscv-cheri/#menvcfg) and [senvcfg](https://riscv.github.io/riscv-cheri/#senvcfg) indicates whether CBO.INVAL performs cache block flushes instead of invalidations for less privileged modes. The instruction checks shown in the table below remain unchanged regardless of the value of CBIE and the privilege mode.

jrtc27:
> Not sure I understand which bit isn't clear so I'll just explain the full rationale behind all this code...
> 
> The CHERI spec describes the following checks for the CBO instructions (amongst others):
> 
> ```
> CBO   | Bounds    | Permissions
> --------------------------------
> CLEAN | Any byte  | W+R
> FLUSH | Any byte  | W+R
> INVAL | Any byte  | W+R+ASR

This should be "All bytes":

> ifdef::cbo_inval[]
> | Permission violation  | It does not grant <<w_perm>>, <<r_perm>> or <<asr_perm>>
> | Length violation      | At least one byte accessed is outside the bounds
> endif::[]

tomaird:
@jrtc27 where are you quoting that bit of spec from? On latest main of the CHERI spec has `cbo.inval` only requiring any byte to be in bounds:

```
| Bounds violation      | None of the bytes accessed are within the bounds, or the capability has <<section_cap_malformed,malformed>> bounds
```

https://github.com/riscv/riscv-cheri/blob/main/src/cheri/insns/cbo_exceptions.adoc?plain=1#L23

jrtc27:
> @jrtc27 where are you quoting that bit of spec from? On latest main of the CHERI spec has `cbo.inval` only requiring any byte to be in bounds:
> 
> ```
> | Bounds violation      | None of the bytes accessed are within the bounds, or the capability has <<section_cap_malformed,malformed>> bounds
> ```
> 
> https://github.com/riscv/riscv-cheri/blob/main/src/cheri/insns/cbo_exceptions.adoc?plain=1#L23

I guess I still had an old version open somehow. But this is very wrong. See https://github.com/riscv/riscv-cheri/pull/221#discussion_r1592596594 :(

model/riscv_insts_aext.sail
```diff
@@ -108,7 +108,7 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
    * RMEM: SC is allowed to succeed (but might fail later)
    */
   let data = X(rs2)[width_bytes * 8 - 1 .. 0];
-  match vmem_write(rs1, zeros(), width_bytes, data, Write(Data), aq & rl, rl, true) {
+  match vmem_write(rs1, zeros(), width_bytes, data, StoreConditional(aq, rl, Data), aq & rl, rl, true) {
```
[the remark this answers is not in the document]
KotorinMinami: That way means we should also put aq and rl in AccessType Read() and Write()? Indeed it's a big work.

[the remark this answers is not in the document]
Timmmm: Normal loads/stores can't acquire or release (as far as I know; I'm not a memory expert though).

Timmmm:
It would be good to get this in - do you mind if I just update it to remove `ExecutableDataFetch`? Or do you think we should keep it even though it isn't used yet?

KotorinMinami:
Okay, I'll remove it.

Timmmm:
Hey, I rebased this and changed it to use the AMO PMA level in the AMO access type instead of the full AMO op since a) that's all we really need to know, and b) we don't have access to that type in `riscv_types.sail` any more. I maybe could have moved the types around to fix that but this seemed better.

Also I think we probably want to go the full way and remove all the `aq`/`rel`/`res`/`con` bools before merging this, otherwise it's just weird having both things. I'll have a go at that at some point.

KotorinMinami:
Thanks for the rebase @Timmmm . I'm also considering how to proceed. Perhaps we should first open an issue to document the removal of the `aq/rel/res/con` boolean parameters.

Timmmm:
I don't think we need to document anything, we just need to remove those parameters and instead source them from the access type. I did start working on this, it doesn't look too bad. The only minor issue I ran into is that a couple of functions have the bool flags but not the access type, but we can just add that.

jordancarlin:
@KotorinMinami @Timmmm This would be great to get in soon as it is blocking #328 (which was recently ratified and would be good to get merged). Are either of you able to rebase and finish this (after the holidays)?

Timmmm:
Yeah maybe we could take a look at each of our approaches in the next meeting and see which one is best (I haven't actually looked yet).

model/riscv_insts_zaamo.sail
```diff
@@ -31,6 +31,28 @@ function amo_encoding_valid(width : word_width_wide, op : amoop, Regidx(rs2) : r
     width <= xlen_bytes | (op == AMOCAS & width <= xlen_bytes * 2 & rs2[0] == bitzero & rd[0] == bitzero)
   ))
 
+function amo_level(op : amoop, width : word_width_wide) -> AmoLevel =
```
pmundkur, line 34: AMO width handling  is not handled in #1405.  In that PR, `pmaCheck` gets a more informative `access`, but doesn't do anything with it.  I think that's best left to a subsequent PR, which would also handle the width of cache block accesses.

model/riscv_insts_zalrsc.sail
```diff
@@ -68,7 +69,7 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
    * RMEM: SC is allowed to succeed (but might fail later)
    */
   let data = X(rs2)[width * 8 - 1 .. 0];
-  match vmem_write(rs1, zeros(), width, data, Write(Data), aq & rl, rl, true) {
+  match vmem_write(rs1, zeros(), width, data, StoreConditional(aq, rl, Data), aq & rl, rl, true) {
```
pmundkur, line 72: It's not clear whether to use `StoreConditional(aq, rl,..)` here or `StoreConditional(aq & rl, rl, ...)` as per the original code. This is why I did not fold in these flags in #1405.  It's probably better to have litmus tests running first to ensure we don't break things in the refactor.

model/riscv_types.sail
```diff
@@ -326,3 +308,42 @@ mapping width_mnemonic_wide : word_width_wide <-> string = {
   8  <-> "d",
   16 <-> "q",
 }
+
+/* memory access types */
+
+enum CacheAccessType = {
+  CleanFlush,
+  Inval,
+  Zero,
```
pmundkur, line 317: We've since added prefetch from Zicbop as well.

model/riscv_vmem_pte.sail
```diff
@@ -142,10 +158,13 @@ function update_PTE_Bits forall 'pte_size, 'pte_size in {32, 64} . (
...
+                             InstructionFetch()        => false,
+                             Read(_)                   => false,
+                             Write(_)                  => true,
+                             AMOAccess(_, _, _, _, _)  => true,
+                             LoadReserved(_, _, _)     => false,
+                             StoreConditional(_, _, _) => true,
+                             Cache(_)                  => false,
                            });
   // Update 'accessed'-bit?
   let update_a = (pte_flags[A] == 0b0);
```
pmundkur, line 170: This needs to be updated: "During address translation, the [cache-block management] instruction also checks the accessed bit and may either raise an exception or set the bit as required." Similarly for cache-block zero.

model/riscv_vmem_pte.sail
```diff
@@ -142,10 +158,13 @@ function update_PTE_Bits forall 'pte_size, 'pte_size in {32, 64} . (
...
-                             Read(_)         => false,
-                             Write(_)        => true,
-                             ReadWrite(_, _) => true
+                             InstructionFetch()        => false,
+                             Read(_)                   => false,
+                             Write(_)                  => true,
+                             AMOAccess(_, _, _, _, _)  => true,
+                             LoadReserved(_, _, _)     => false,
+                             StoreConditional(_, _, _) => true,
+                             Cache(_)                  => false,
```
pmundkur, line 167: This is incorrect: "During address translation, the [cache-block zero] instruction also checks the accessed and dirty bits and may either raise an exception or set the bits as required."

model/riscv_vmem_types.sail
```diff
@@ -42,8 +42,11 @@ function accessType_to_str (a) =
   match (a) {
     Read(_)            => "R",
     Write(_)           => "W",
-    ReadWrite(_, _)    => "RW",
+    AMOAccess(_)       => "A",
```
pmundkur, line 45: I left this as "RW" in #1405.  Is "A" preferable?

model/riscv_vmem_types.sail
```diff
@@ -42,8 +42,11 @@ function accessType_to_str (a) =
...
     InstructionFetch() => "X",
+    Cache(_)           => "C",
+    LoadReserved(_)    => "LR",
+    StoreConditional(_) => "SC",
```
pmundkur, line 49: I left these as "R" and "W", perhaps "LR" and "SC" are more informative?

pmundkur (commented):
Notes from comparing this to #1405.  Sorry, I should have checked this before.

model/riscv_vmem_types.sail
jordancarlin: Until we have other trace/logging formats, I would avoid changing what is printed in the log. It is currently the only way of parsing the output of the model, so switching from `RW` to `A` would likely break many tools. Once we have new machine-readable formats available, we can intentionally break the human-readable log, and then I think switching to `A`/`LR`/`SC` is worth considering.

pmundkur:
Should we close this, now that #1405 is merged?

PR #793, opened 2025-03-14
Fix vmsof.v condition check and vmxnor.mm bit mask

nadime15:
Running the vector test suite revealed failures in `vmsof.v` and `vmxnor.mm`:

 * vmsof.v: Condition check was incorrect
 * vmxnor.mm: Applied the wrong bitwise operation


jordancarlin (approved):
LGTM. Simple fixes for obvious bugs.

pmundkur (approved):

Issue #768, opened 2025-03-05
Update mappings to use the `when` construct for guards


pmundkur:
TODO after the next Sail compiler release.

See rems-project/sail#1106

arichardson:
This would make a lot of the clauses much easier to read - hopefully we can get a new release soon @Alasdair?

Alasdair:
Yes, fixing that and the outstanding warnings about incomplete patterns were some of the last few things I wanted to get done. I think it's just improving some of the configuration system error messages and it should be good.

PR #794, opened 2025-03-14
change all mapping guard ifs with when

KotorinMinami:
mentioned in #768

### DIFF
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -57,8 +57,9 @@ function clause extensionEnabled(Ext_Zalrsc) = misa[A] == 0b1
 
 union clause ast = LOADRES : (bool, bool, regidx, word_width, regidx)
 
-mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)                                                                        if extensionEnabled(Ext_Zalrsc) & lrsc_width_valid(size)
-  <-> 0b00010 @ bool_bits(aq) @ bool_bits(rl) @ 0b00000 @ encdec_reg(rs1) @ 0b0 @ size_enc(size) @ encdec_reg(rd) @ 0b0101111 if extensionEnabled(Ext_Zalrsc) & lrsc_width_valid(size)
+mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)
+  <-> 0b00010 @ bool_bits(aq) @ bool_bits(rl) @ 0b00000 @ encdec_reg(rs1) @ 0b0 @ size_enc(size) @ encdec_reg(rd) @ 0b0101111
+  when extensionEnabled(Ext_Zalrsc) & lrsc_width_valid(size)
 
 /* We could set load-reservations on physical or virtual addresses.
  * However most chips (especially multi-core) will use physical addresses.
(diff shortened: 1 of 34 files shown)


arichardson (commented):
This is a nice cleanup! Not reviewed this in detail, but it looks mostly mechanical, so should be safe to merge although I'd appreciate if someone could double check there aren't any functional changes here.

I agree that formatting improvements can be made later (ideally automatically once the sail auto-formatter works).

jordancarlin:
> This is a nice cleanup! Not reviewed this in detail, but it looks mostly mechanical, so should be safe to merge although I'd appreciate if someone could double check there aren't any functional changes here.

Unfortunately, I think it’ll be tricky to test until the new configuration support is merged. Once that’s in, we should add tests that disable various extensions and ensure that the relevant instructions don’t execute.

arichardson:
> > This is a nice cleanup! Not reviewed this in detail, but it looks mostly mechanical, so should be safe to merge although I'd appreciate if someone could double check there aren't any functional changes here.
> 
> Unfortunately, I think it’ll be tricky to test until the new configuration support is merged. Once that’s in, we should add tests that disable various extensions and ensure that the relevant instructions don’t execute.

I wasn't suggesting adding tests here just looking through the diff in detail to check it's equivalent since I don't have time to do that right now.

model/riscv_insts_aext.sail
```diff
@@ -57,8 +57,8 @@ function clause extensionEnabled(Ext_Zalrsc) = misa[A] == 0b1
 
 union clause ast = LOADRES : (bool, bool, regidx, word_width, regidx)
 
-mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)                                                                        if extensionEnabled(Ext_Zalrsc) & lrsc_width_valid(size)
-  <-> 0b00010 @ bool_bits(aq) @ bool_bits(rl) @ 0b00000 @ encdec_reg(rs1) @ 0b0 @ size_enc(size) @ encdec_reg(rd) @ 0b0101111 if extensionEnabled(Ext_Zalrsc) & lrsc_width_valid(size)
+mapping clause encdec = LOADRES(aq, rl, rs1, size, rd)
+  <-> 0b00010 @ bool_bits(aq) @ bool_bits(rl) @ 0b00000 @ encdec_reg(rs1) @ 0b0 @ size_enc(size) @ encdec_reg(rd) @ 0b0101111 when extensionEnabled(Ext_Zalrsc) & lrsc_width_valid(size)
```
pmundkur, line 61:
```suggestion
  <-> 0b00010 @ bool_bits(aq) @ bool_bits(rl) @ 0b00000 @ encdec_reg(rs1) @ 0b0 @ size_enc(size) @ encdec_reg(rd) @ 0b0101111 
        when extensionEnabled(Ext_Zalrsc) & lrsc_width_valid(size)
```

It would be good to put the `when` on a different line (and appropriately indented) to make it more visible,

KotorinMinami: Perhaps we should discuss a more uniform formatting scheme. As this problem I didn't notice because I use the wrap option on my local IDE

pmundkur: This might need to go into  CODE_STYLE, with examples for clarity.

KotorinMinami: So we may need to discuss this in a separate Issue.

pmundkur: Since this affects so many files, it would be best to do this formatting once in this PR instead of a separate one.  Could the updated guidelines in CODE_STYLE  be used here?

KotorinMinami: It might take some time to review and revise, but I'll do it as quickly as possible. I may also update the guidelines in CODE_STYLE.

Timmmm (approved):
LGTM, thanks! Must have been tedious :-D

pmundkur (approved):
Thanks for this!

jordancarlin (approved):
Looks great!

PR #797, opened 2025-03-16
Add unratified Svrsw60t59b extension

ved-rivos:
Spec: https://github.com/riscv/riscv-isa-manual/pull/1907

### DIFF
--- a/model/riscv_vmem_pte.sail
+++ b/model/riscv_vmem_pte.sail
@@ -27,7 +27,9 @@ bitfield PTE_Ext : pte_ext_bits = {
   N         : 9,
   // Page based memory types
   PBMT      : 8 .. 7,
-  reserved  : 6 .. 0,
+  // Svrsw60t59b: Additional reserved-for-software bits
+  EXT_RSW   : 6 .. 5,
+  reserved  : 4 .. 0,
 }
 
 /*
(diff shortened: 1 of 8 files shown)


jordancarlin:
@ved-rivos this would be good to get merged now that it is ratified. Are you planning to rebase/update it? If not, I can also handle updating it.

ved-rivos:
@jordancarlin its ratified now . Would be glad if you can help update. Thanks much !

jordancarlin:
Updated to use the new config system. Should be ready to go now.

nadime15 (approved):
It would be nice, if you could add a TODO somewhere: 
> // TODO: Implement G-stage PTE validation when Hypervisor extension is added 

or at least create an issue to track the missing pieces that need to be added once Hypervisor support is available.

jordancarlin:
> It would be nice, if you could add a TODO somewhere:
> 
> > // TODO: Implement G-stage PTE validation when Hypervisor extension is added
> 
> or at least create an issue to track the missing pieces that need to be added once Hypervisor support is available.

I think this is part of adding G-stage translation in general. There are many things that influence that, and I don’t see a reason to add a comment for just this one.

model/riscv_vmem_pte.sail
```diff
@@ -27,7 +27,9 @@ bitfield PTE_Ext : pte_ext_bits = {
   N         : 9,
   // Page based memory types
   PBMT      : 8 .. 7,
-  reserved  : 6 .. 0,
+  // Svrsw60t59b: Additional reserved-for-software bits
+  EXT_RSW   : 6 .. 5,
```
pmundkur: Should we use a better name than `EXT_RSW`?  I'm not sure what the `EXT_` is adding. Perhaps `RSW_60t59` or just `RSW`?  I'd prefer the former, but just `RSW` also works for me since we could change `RSW` later if there is a later extension that needs disambiguation.

ved-rivos: RSW - Reserved for software - this is the term used in Priv. ISA for bits 9:8 that are reserved for software use.
EXT_RSW - extended reserved for software. 

RSW_60t59b is also fine. The only reason the field has a name (and not 9:8) is because the bits 60:59 must be 0 when Svrsw60t59b is not implemented.

jordancarlin: Updated to use `RSW_60t59b` to remove any ambiguity.

Timmmm (approved):

pmundkur (approved):

PR #801, opened 2025-03-17
Print compiler used for tests

Timmmm:
This is quite useful information for debugging. Note this will print `...clang;--target=riscv32` (i.e. semicolon separated instead of space separated), but I think it's not worth adding the code to get "correct" output.

### DIFF
--- a/test/first_party/CMakeLists.txt
+++ b/test/first_party/CMakeLists.txt
@@ -66,6 +66,9 @@ if (NOT CROSS_COMPILER_COMMAND_RV32 OR NOT CROSS_COMPILER_COMMAND_RV64)
     message(FATAL_ERROR "No suitable cross-compiler found. We recommend downloading Clang from https://github.com/llvm/llvm-project/releases")
 endif()
 
+message(STATUS "Compiling RV32 tests with: ${CROSS_COMPILER_COMMAND_RV32}")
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):
LGTM. Simple improvement that will make debugging easier.

test/first_party/CMakeLists.txt
```diff
@@ -66,6 +66,9 @@ if (NOT CROSS_COMPILER_COMMAND_RV32 OR NOT CROSS_COMPILER_COMMAND_RV64)
     message(FATAL_ERROR "No suitable cross-compiler found. We recommend downloading Clang from https://github.com/llvm/llvm-project/releases")
 endif()
 
+message(STATUS "Compiling RV32 tests with: ${CROSS_COMPILER_COMMAND_RV32}")
+message(STATUS "Compiling RV64 tests with: ${CROSS_COMPILER_COMMAND_RV64}")
```
arichardson, line 70: I think `$<JOIN:${CROSS_COMPILER_COMMAND_RV32}, >` Should work, but probably not worth fixing.

Timmmm: Ah didn't know about that. I'll do it in a future PR if I remember...

PR #805, opened 2025-03-18
Fix --inst-limit

trdthg:
caused by #748
 
Forgot to assign `val` to `inst_limit` after parsing cli params


arichardson (approved):

jordancarlin (approved):

Timmmm (approved):
Oops :-)

I guess we should have tests for the CLI... I don't think it's super critical though.

Issue #796, opened 2025-03-15
Configuration of supported SATP Virtual Memory modes


jordancarlin:
Right now, all virtual memory modes are always available (ie sv39, sv48, and sv57 in rv64). Each of these should have `extensionEnabled` clauses and appropriate guards.

PR #807, opened 2025-03-21
Add satp mode guard

KotorinMinami:
From #796 , as the spec say
>Implementations are not required to support all MODE settings, and if satp is written with an unsupported MODE, the entire write has no effect; no fields in satp are modified.

so adding extensionenabled and applying guard in legalize_satp I think might work.

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -202,6 +202,16 @@ function clause hartSupports(Ext_Svnapot) = false // Not supported yet
 enum clause extension = Ext_Svpbmt
 function clause hartSupports(Ext_Svpbmt) = false // Not supported yet
 
+// Supervisor-Level Address Translation Modes
+enum clause extension = Ext_Svbare
+enum clause extension = Ext_Sv32
+function clause hartSupports(Ext_Sv32) = config extensions.Sv32.supported : bool & (xlen == 32)
+enum clause extension = Ext_Sv39
+function clause hartSupports(Ext_Sv39) = config extensions.Sv39.supported : bool & (xlen == 64)
+enum clause extension = Ext_Sv48
+function clause hartSupports(Ext_Sv48) = config extensions.Sv48.supported : bool & (xlen == 64)
+enum clause extension = Ext_Sv57
+function clause hartSupports(Ext_Sv57) = config extensions.Sv57.supported : bool & (xlen == 64)
 // Cycle and Instret Privilege Mode Filtering
 enum clause extension = Ext_Smcntrpmf
 function clause hartSupports(Ext_Smcntrpmf) = config extensions.Smcntrpmf.supported
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -140,6 +140,16 @@ function clause write_CSR(0x301, value) = { misa = legalize_misa(misa, value); m
 function clause currentlyEnabled(Ext_U) = hartSupports(Ext_U) & misa[U] == 0b1
 function clause currentlyEnabled(Ext_S) = hartSupports(Ext_S) & misa[S] == 0b1
 
+function clause currentlyEnabled(Ext_Svbare) = currentlyEnabled(Ext_S)
+function clause currentlyEnabled(Ext_Sv32) = hartSupports(Ext_Sv32) &  currentlyEnabled(Ext_S)
+function clause currentlyEnabled(Ext_Sv39) = hartSupports(Ext_Sv39) &  currentlyEnabled(Ext_S)
+function clause currentlyEnabled(Ext_Sv48) = hartSupports(Ext_Sv48) &  currentlyEnabled(Ext_S)
+function clause currentlyEnabled(Ext_Sv57) = hartSupports(Ext_Sv57) &  currentlyEnabled(Ext_S)
+
+function virtual_memory_supported() -> bool = {
+  currentlyEnabled(Ext_Sv32) | currentlyEnabled(Ext_Sv39) | currentlyEnabled(Ext_Sv48) | currentlyEnabled(Ext_Sv57)
+}
+
 /*
  * Illegal values legalized to least privileged mode supported.
  * Note: the only valid combinations of supported modes are M, M+U, M+S+U.
@@ -235,7 +245,7 @@ function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
     TW = if currentlyEnabled(Ext_U) then v[TW] else 0b0,
     TVM = if currentlyEnabled(Ext_S) then v[TVM] else 0b0,
     MXR = if currentlyEnabled(Ext_S) then v[MXR] else 0b0,
-    SUM = if currentlyEnabled(Ext_S) then v[SUM] else 0b0, // TODO: Should also be disabled if satp.MODE is read-only 0
+    SUM = if virtual_memory_supported() then v[SUM] else 0b0,
     MPRV = if currentlyEnabled(Ext_U) then v[MPRV] else 0b0,
     /* We don't have any extension context yet. */
     XS = extStatus_to_bits(Off),
@@ -851,13 +861,26 @@ function legalize_satp(
   written_value : xlenbits,
 ) -> xlenbits = {
   if xlen == 32 then {
-    /* all 32-bit satp modes are valid */
-    written_value
+    let s = Mk_Satp32(written_value);
+    match satpMode_of_bits(arch, 0b000 @ s[Mode]) {
+      None()  => prev_value,
+      Some(Sv_mode) => match Sv_mode {
+        Bare if currentlyEnabled(Ext_Svbare) => s.bits,
+        Sv32 if currentlyEnabled(Ext_Sv32) => s.bits,
+        _ => prev_value,
+      }
+    }
   } else if xlen == 64 then {
     let s = Mk_Satp64(written_value);
     match satpMode_of_bits(arch, s[Mode]) {
       None()  => prev_value,
-      Some(_) => s.bits
+      Some(Sv_mode) =>  match Sv_mode {
+        Bare if currentlyEnabled(Ext_Svbare) => s.bits,
+        Sv39 if currentlyEnabled(Ext_Sv39) => s.bits,
+        Sv48 if currentlyEnabled(Ext_Sv48) => s.bits,
+        Sv57 if currentlyEnabled(Ext_Sv57) => s.bits,
+        _ => prev_value,
+      }
     }
   } else {
     internal_error(__FILE__, __LINE__, "Unsupported xlen" ^ dec_str(xlen))
--- a/config/default.json
+++ b/config/default.json
@@ -195,6 +195,18 @@
     },
     "Svinval": {
       "supported": true
+    },
+    "Sv32": {
+      "supported": true
+    },
+    "Sv39": {
+      "supported": true
+    },
+    "Sv48": {
+      "supported": true
+    },
+    "Sv57": {
+      "supported": true
     }
   }
 }


model/riscv_extensions.sail
```diff
@@ -116,3 +116,9 @@ enum clause extension = Ext_Svpbmt
 
 // Cycle and Instret Privilege Mode Filtering
 enum clause extension = Ext_Smcntrpmf
+
+// supervisor-level address translation modes
+enum clause extension = Ext_Sv32
```
jordancarlin, line 121: These should be with the other Sv extensions a bit higher up the list.

model/riscv_extensions.sail
```diff
@@ -116,3 +116,9 @@ enum clause extension = Ext_Svpbmt
 
 // Cycle and Instret Privilege Mode Filtering
 enum clause extension = Ext_Smcntrpmf
+
+// supervisor-level address translation modes
```
jordancarlin, line 120: For consistency with the formatting of other extensions in the list:
```suggestion
// Supervisor-Level Address Translation Modes
```

jordancarlin (changes requested):
Need to think about whether `extensionEnabled()` makes sense for this. Technically only one of these extensions is ever enabled at a time (whichever is set in `satp`), but this may be the simplest way to implement this.

jrtc27:
> Need to think about whether `extensionEnabled()` makes sense for this. Technically only one of these extensions is ever enabled at a time (whichever is set in `satp`), but this may be the simplest way to implement this.

No? The extension just means it's a supported option for satp.MODE, not that it's the currently active one.

jordancarlin:
> > Need to think about whether `extensionEnabled()` makes sense for this. Technically only one of these extensions is ever enabled at a time (whichever is set in `satp`), but this may be the simplest way to implement this.
> 
> No? The extension just means it's a supported option for satp.MODE, not that it's the currently active one.

The way it's handled in the model right now, we use `extensionEnabled()` to mean an extension is active and `sys_enable_XXX()` to indicate it's a supported extension. For example, `extensionEnabled(Ext_F)` is only true when `mstatus.FS` is not set to off. It seems like the analogue would be for the `extensionEnabled()` clauses to be dependent on `satp.MODE`, but I'm not sure that scheme totally works for the VM mode extensions for the same reason you just mentioned.

jrtc27:
> > > Need to think about whether `extensionEnabled()` makes sense for this. Technically only one of these extensions is ever enabled at a time (whichever is set in `satp`), but this may be the simplest way to implement this.
> > 
> > 
> > No? The extension just means it's a supported option for satp.MODE, not that it's the currently active one.
> 
> The way it's handled in the model right now, we use `extensionEnabled()` to mean an extension is active and `sys_enable_XXX()` to indicate it's a supported extension. For example, `extensionEnabled(Ext_F)` is only true when `mstatus.FS` is not set to off. It seems like the analogue would be for the `extensionEnabled()` clauses to be dependent on `satp.MODE`, but I'm not sure that scheme totally works for the VM mode extensions for the same reason you just mentioned.

The extension is always active, there's no way to turn it off. You just may not currently be using that specific mode in satp. It makes total sense for extensionEnabled to be true for each of the SvXX that the model is configured to support, because that's what the architecture says happens.

KotorinMinami:
If I understand correctly, my current implementation should be fine, right? Going forward, we just need to add the `sys_enable_svxx()` function and use them in `isextensionEnable()`.

model/riscv_sys_regs.sail
```diff
@@ -898,13 +904,21 @@ function legalize_satp(
...
   } else if xlen == 64 then {
     let s = Mk_Satp64(written_value);
     match satpMode_of_bits(arch, s[Mode]) {
       None()  => prev_value,
-      Some(_) => s.bits
+      Some(Sv_mode) =>  match Sv_mode {
+        Sv39 => if extensionEnabled(Ext_Sv39) then s.bits else prev_value,
+        Sv48 => if extensionEnabled(Ext_Sv48) then s.bits else prev_value,
+        Sv57 => if extensionEnabled(Ext_Sv57) then s.bits else prev_value,
+        _    => prev_value
```
jordancarlin, line 920: I think prevents you from ever writing bare (0x0) to mode. Same issue for rv32 above.

jordancarlin (changes requested):

model/riscv_sys_regs.sail
```diff
@@ -898,13 +904,25 @@ function legalize_satp(
   written_value : xlenbits,
 ) -> xlenbits = {
   if xlen == 32 then {
-    /* all 32-bit satp modes are valid */
-    written_value
+    let s = Mk_Satp32(written_value);
+    match satpMode_of_bits(arch, 0b000 @ s[Mode]) {
+      None()  => prev_value,
+      Some(Sv_mode) => match Sv_mode {
+        Bare => s.bits,
```
jrtc27, line 911: Svbare also exists

KotorinMinami: Currently, Sail supports only the following SATP modes: Bare, Sv32, Sv39, Sv48, and Sv57. Have I missed any?

jordancarlin: We should gate Bare on the `Svbare` extension. From https://github.com/riscv/riscv-profiles/blob/main/src/profiles.adoc:

> Svbare: The satp mode Bare must be supported.

jordancarlin (approved):
LGTM now

arichardson (approved):
LGTM, but I think minor simplification is possible.

model/riscv_sys_regs.sail
```diff
@@ -898,13 +905,26 @@ function legalize_satp(
...
   if xlen == 32 then {
-    /* all 32-bit satp modes are valid */
-    written_value
+    let s = Mk_Satp32(written_value);
+    match satpMode_of_bits(arch, 0b000 @ s[Mode]) {
+      None()  => prev_value,
+      Some(Sv_mode) => match Sv_mode {
+        Bare => if extensionEnabled(Ext_Svbare) then s.bits else prev_value,
+        Sv32 => if extensionEnabled(Ext_Sv32) then s.bits else prev_value,
+        _   => prev_value,
```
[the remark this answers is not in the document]
KotorinMinami: Oh yes, and it definitely makes the code cleaner.

arichardson (approved):

jordancarlin (changes requested):
From section 3.1.6.4 of the privileged spec:

> SUM is read-only 0 if S-mode is not supported or if satp.MODE is read-only 0.

Right now `mstatus.SUM` is guarded on S, but now that we are adding a way to disable writes to `satp` it needs to be guarded on that as well. Same for `sstatus.SUM`.

jordancarlin (approved):
LGTM

model/riscv_extensions.sail
```diff
@@ -113,6 +113,11 @@ enum clause extension = Ext_Svinval
 enum clause extension = Ext_Svnapot
 // Page-Based Memory Types
 enum clause extension = Ext_Svpbmt
-
+// Supervisor-Level Address Translation Modes
+enum clause extension = Ext_Svbare
+enum clause extension = Ext_Sv32
```
jordancarlin, line 118: Can you update this to add the `hartSupports` clauses and appropriate config keys in the default config json.
```suggestion
enum clause extension = Ext_Sv32
function clause hartSupports(Ext_Sv32) = config extensions.Sv32.supported
```

model/riscv_sys_regs.sail
```diff
@@ -160,6 +160,19 @@ function sys_enable_supervisor() -> bool = true
 function clause extensionEnabled(Ext_U) = misa[U] == 0b1
 function clause extensionEnabled(Ext_S) = misa[S] == 0b1
 
+// TODO: Make configurable
+function clause extensionEnabled(Ext_Svbare) = extensionEnabled(Ext_S)
+function clause extensionEnabled(Ext_Sv32) = extensionEnabled(Ext_S) & (xlen == 32)
```
jordancarlin, line 165: Update
```suggestion
function clause currentlyEnabled(Ext_Sv32) = currentlyEnabled(Ext_S) & (xlen == 32) & hartSupports(Ext_Sv32)
```

jordancarlin (changes requested):
Just needs to be updated for the new config system. Then should be good to go.

model/riscv_sys_regs.sail
```diff
@@ -140,6 +140,16 @@ function clause write_CSR(0x301, value) = { misa = legalize_misa(misa, value); m
 function clause currentlyEnabled(Ext_U) = hartSupports(Ext_U) & misa[U] == 0b1
 function clause currentlyEnabled(Ext_S) = hartSupports(Ext_S) & misa[S] == 0b1
 
+function clause currentlyEnabled(Ext_Svbare) = currentlyEnabled(Ext_S)
+function clause currentlyEnabled(Ext_Sv32) = currentlyEnabled(Ext_S) & (xlen == 32) & hartSupports(Ext_Sv32)
```
jordancarlin, line 144: For the other extensions, we pushed the `xlen` check down into the `hartSupports` clause.

jordancarlin: To clarify, the xlen component should be in `hartSupports` only and not `currentlyEnabled`. It looks like you've duplicated it both places now.

model/riscv_extensions.sail
```diff
@@ -196,6 +196,16 @@ function clause hartSupports(Ext_Svnapot) = false // Not supported yet
...
+function clause hartSupports(Ext_Sv32) = config extensions.Sv32.supported : bool & (xlen == 32)
+enum clause extension = Ext_Sv39
+function clause hartSupports(Ext_Sv39) = config extensions.Sv39.supported
+enum clause extension = Ext_Sv48
+function clause hartSupports(Ext_Sv48) = config extensions.Sv48.supported
+enum clause extension = Ext_Sv57
+function clause hartSupports(Ext_Sv57) = config extensions.Sv57.supported
```
arichardson, line 208: Should these have an xlen check (or a configuration sanity check that gives a fatal error if you try to enable them for RV32?)? That would allow us to remove the checks lower down.

arichardson (approved):
LGTM with my comment and https://github.com/riscv/sail-riscv/pull/807/files#r2042691915 addressed

model/riscv_extensions.sail
jordancarlin: They should definitely check xlen, similar to sv32.

We probably need a whole new file/function that checks for illegal configurations since there a number of other extensions that have the same issue. I’d hold off on that part in this PR.

model/riscv_sys_regs.sail
```diff
@@ -140,6 +140,16 @@ function clause write_CSR(0x301, value) = { misa = legalize_misa(misa, value); m
 function clause currentlyEnabled(Ext_U) = hartSupports(Ext_U) & misa[U] == 0b1
 function clause currentlyEnabled(Ext_S) = hartSupports(Ext_S) & misa[S] == 0b1
 
+function clause currentlyEnabled(Ext_Svbare) = currentlyEnabled(Ext_S)
+function clause currentlyEnabled(Ext_Sv32) = currentlyEnabled(Ext_S) & hartSupports(Ext_Sv32)
```
jordancarlin, line 144: We've generally been ordering these with the `hartSupports` part first. So
```suggestion
function clause currentlyEnabled(Ext_Sv32) = hartSupports(Ext_Sv32) & currentlyEnabled(Ext_S)
```

arichardson (approved):

jordancarlin (approved):
Looks great!

Timmmm (commented):
LGTM. The only other thing I wonder is if we should enforce that Sv48 depends on Sv39 and Sv57 depends on Sv48? E.g. you can't *just* enable Sv57.

jordancarlin:
> LGTM. The only other thing I wonder is if we should enforce that Sv48 depends on Sv39 and Sv57 depends on Sv48? E.g. you can't _just_ enable Sv57.

True, and maybe that is the easiest way for now.

Timmmm:
Discussed in the meeting. We think we're going to eventually have a config validation function to check you don't enable Sv57 without Sv48 etc. so this is fine.

Timmmm (approved):

PR #809, opened 2025-03-21
Update code-style to add suggestion about formatting mapping clauses and the use of `when`.

pmundkur:
Not sure if this sounds too prescriptive, open to suggestions.


jordancarlin (approved):
Seems reasonable to me. Hopefully at some point the Sail auto-formatter will be ready to go, but for now I don’t think it hurts to be this specific.

Timmmm (approved):

PR #790, opened 2025-03-13
Add Support for Zvbb and Zvkb

nadime15:
Add support for Zvbb and Zvkb with following instructions:

```
vandn.[vv,vx]
vbrev.v
vbrev8.v
vrev8.v
vclz.v
vctz.v
vcpop.v
vrol.[vv,vx]
vror.[vv,vx,vi]
vwsll.[vv,vx,vi]
```

CI will fail due to the use of `when`, which is unsupported in Sail versions below 0.19.

Based on #558 and #236

The instruction set for the Zvbb (this extension is a superset of the Zvkb extension) and Zvkb extensions has evolved over time. 
* This PR makes sure that the instructions are assigned to their corresponding extension
* Adds a missing instruction
* Simplifies the code

Code was tested with the risc-v vector test suite with (VLEN=512 and ELEN=64)

Once we can configure Sail, we can add the Zvbb and Zvbk option to enable or disable accordingly.

PR #811, opened 2025-03-23
fix lean support after #790

tobiasgrosser:


Timmmm:
How do you use Lean? Can we add it to CI now we have updated Sail?

tobiasgrosser:
We currently us it to build this model. I am happy to move ci over to the sail riscv model. Would it be possible to merge this pr to unbreakable our external build. I will then prepare a pr next week.

tobiasgrosser:
Yes, adding CI would be helpful. We already have a GitHub action that does sth similar in this repo: https://github.com/opencompl/sail-riscv-lean/blob/main/.github/workflows/update-sail-build.yml. We are looking into cleaning it up and you should have a PR sometime soon.

Issue #812, opened 2025-03-23
Simulator no longer builds on macOS


Alasdair:
```
/Users/alasdair/repos/sail-riscv/c_emulator/riscv_sim.cpp:66:6: error: no template named 'optional' in namespace 'std'
   66 | std::optional<rvfi_handler> rvfi;
      | ~~~~~^
/Users/alasdair/repos/sail-riscv/c_emulator/riscv_sim.cpp:658:21: error: no member named 'nullopt' in namespace 'std'
  658 |         rvfi = std::nullopt;
      |                ~~~~~^
3 warnings and 2 errors generated.
```
I think we need to specify C++17 somewhere in CMake. I tried setting it in what felt like the obvious place, but that caused issues when building the `.c` files.

jrtc27:
Set CMAKE_CXX_STANDARD to 17 and CMAKE_CXX_STANDARD_REQUIRED to On.

Alasdair:
Thanks, that fixed my problem!

arichardson:
Ah that should have been caught by CI!

arichardson:
https://github.com/riscv/sail-riscv/pull/813

jordancarlin:
> Ah that should have been caught by CI!

We don’t run CI on macOS right now.

edolnx:
I'll look into this and see what I can come up with. Might as well get a Mac Mini setup sooner rather than later to get this working. I can at least start with the older Intel devices I have and then get a Apple Silicon device by end of week.

PR #813, opened 2025-03-24
Explicitly request C++17 support

arichardson:
Not all compilers default to C++17, so we need to tell CMake to request c++17 (for std::optional).

Fixes https://github.com/riscv/sail-riscv/issues/812
Needed since https://github.com/riscv/sail-riscv/pull/784

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.20)
 
 project(sail_riscv)
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED  ON)
 set(SAIL_REQUIRED_VER 0.19)
 
 # Make users explicitly pick a build type so they don't get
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.20)
 
 project(sail_riscv)
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED  ON)
```
arichardson, line 6:
```suggestion
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

Timmmm (approved):

CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.20)
 
 project(sail_riscv)
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
Timmmm, line 6: I prefer `TRUE` but :shrug:

jordancarlin (approved):

CMakeLists.txt
arichardson:
```suggestion
set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
```

arichardson:
I can clean up the history, but since this is trivial squash and merge should be fine?

pmundkur (approved):

PR #814, opened 2025-03-25
Lean: refine sizeOf instance for extension type

javra:
Fixes a termination error in the Lean translation of the function `extensionEnabled`.

### DIFF
--- a/handwritten_support/RiscvExtras.lean
+++ b/handwritten_support/RiscvExtras.lean
@@ -141,18 +141,57 @@ axiom extern_f32roundToInt : BitVec 3 → BitVec 32 → Bool → Unit
 axiom extern_f64roundToInt : BitVec 3 → BitVec 64 → Bool → Unit
 
 -- Termination of extensionEnabled
-
-
 instance : SizeOf extension where
   sizeOf x :=
     match x with
-    | .Ext_Zihpm => 0
-    | .Ext_B => 0
-    | .Ext_C => 0
-    | .Ext_D => 0
-    | .Ext_F => 0
-    | .Ext_Zfh => 0
-    | .Ext_Zca => 1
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):

Timmmm:
Damn that looks like a pain to maintain! Sail does generate `num_of_<enum>` functions now - maybe you could use that somehow?

jordancarlin:
I think a lot of the handwritten support files have similar, very tedious aspects. It would be good for most of this to be autogenerated at some point.

javra:
Yes, I'm working on a general solution to this issue on the Lean side that would equip all enums with a more refined termination strategy. Apart from that, I was made aware that there actually *is* a function like that autogenerated in Lean. Will push in a second.

Timmmm (approved):

handwritten_support/RiscvExtras.lean
```diff
@@ -141,18 +141,7 @@ axiom extern_f32roundToInt : BitVec 3 → BitVec 32 → Bool → Unit
...
-    | .Ext_B => 0
-    | .Ext_C => 0
-    | .Ext_D => 0
-    | .Ext_F => 0
-    | .Ext_Zfh => 0
-    | .Ext_Zca => 1
-    | _ => 2
+  sizeOf := extension.toCtorIdx
 
 macro_rules | `(tactic| decreasing_trivial) => `(tactic| simp [sizeOf])
```
ineol, line 147:
```suggestion
macro_rules | `(tactic| decreasing_trivial) => `(tactic| decide)
```
I need this change to make this work, does it work for you without it @javra?

javra: You're right

PR #815, opened 2025-03-25
Install libstdc++-static for release builds

Timmmm:
Now that we use C++ we need this installed to compile the release.


jordancarlin (approved):

.github/workflows/release.yml
```diff
@@ -51,6 +51,7 @@ jobs:
             gcc \
             gcc-c++ \
             cmake \
+            libstdc++-static \
```
jordancarlin, line 54: Maybe put it after `gmp-static` so all the libraries are listed together, but either way is fine with me.

Timmmm: Done

PR #816, opened 2025-03-25
Ensure V context is dirtied in Zvbb instructions.

pmundkur:
While here, also fix an indentation for vstart.

It looks like this is a hole in the vector test suite.


pmundkur:
@nadime15 could you review?

nadime15 (approved):
Looks good, thanks for catching that!

Timmmm (approved):

PR #822, opened 2025-03-31
Improvements to prelude

pmundkur:
Make the `reverse` function more generic; add a helper for carryless multiply and use it in the appropriate scalar and vector instructions.

Tested against upstream `riscv-tests` and the vector test suite.

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -191,14 +191,15 @@ function rotatel (v, n) =
 overload operator >>> = {rotate_bits_right, rotater}
 overload operator <<< = {rotate_bits_left, rotatel}
 
-function reverse_bits_in_byte (xs : bits(8)) -> bits(8) = {
-  var ys : bits(8) = zeros();
-  foreach (i from 0 to 7)
-    ys[i] = xs[7-i];
+val reverse_bits : forall 'n, 'n > 0. bits('n) -> bits('n)
+function reverse_bits (xs)  = {
+  var ys : bits('n) = zeros();
+  foreach (i from 0 to ('n - 1))
+    ys[i] = xs['n - 1 - i];
   ys
 }
 
-overload reverse = {reverse_bits_in_byte}
+overload reverse = {reverse_bits}
 
 overload operator / = {quot_positive_round_zero, quot_round_zero}
 overload operator * = {mult_atom, mult_int}
--- a/model/riscv_insts_zbc.sail
+++ b/model/riscv_insts_zbc.sail
@@ -20,12 +20,8 @@ mapping clause assembly = RISCV_CLMUL(rs2, rs1, rd)
   <-> "clmul" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
 
 function clause execute (RISCV_CLMUL(rs2, rs1, rd)) = {
-  let rs1_val = X(rs1);
-  let rs2_val = X(rs2);
-  var result : xlenbits = zeros();
-  foreach (i from 0 to (xlen_val - 1))
-    if rs2_val[i] == bitone then result = result ^ (rs1_val << i);
-  X(rd) = result;
+  let prod = carryless_mul(X(rs1), X(rs2));
+  X(rd) = prod[xlen - 1 .. 0];
   RETIRE_SUCCESS
 }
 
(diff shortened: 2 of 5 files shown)


model/prelude.sail
```diff
@@ -221,6 +222,20 @@ function log2(n) = {
...
+function carryless_mult(a, b) = {
+  var result : bits(2 * 'n) = zeros();
+  foreach (i from 0 to ('n - 1)) {
+    if   a[i] == bitone
+    then result = result ^ (zero_extend(b) << i);
+  };
+  result
+}
+
+infix 7 *_c
```
arichardson, line 236: Do we really need this? It's convenient, but when looking at the sail code integrated into the main I would find this much more confusing than a "carryless_mul(a, b)" function call.

with `*_c` I'd have no idea what it is, whereas a well-named function makes it obvious.

arichardson (commented):
The new code is much more readable, but I'd avoid the overloaded operator to make it more clear even if it's longer.

Since the goal is to integrate the code from model into specs, it should be easy to read for someone with limited experience and not use too much clever stuff.

pmundkur:
Agreed; I removed the overload.

model/prelude.sail
```diff
@@ -191,14 +191,15 @@ function rotatel (v, n) =
 overload operator >>> = {rotate_bits_right, rotater}
 overload operator <<< = {rotate_bits_left, rotatel}
 
-function reverse_bits_in_byte (xs : bits(8)) -> bits(8) = {
-  var ys : bits(8) = zeros();
-  foreach (i from 0 to 7)
-    ys[i] = xs[7-i];
+val reverse_bits : forall 'n, 'n > 0. bits('n) -> bits('n)
```
Timmmm, line 194: I think you don't need the 'n > 0 with the latest Sail compiler. IIRC it implicitly adds it since 'n is used in bits().

model/riscv_insts_zbc.sail
```diff
@@ -20,12 +20,8 @@ mapping clause assembly = RISCV_CLMUL(rs2, rs1, rd)
...
 
 function clause execute (RISCV_CLMUL(rs2, rs1, rd)) = {
-  let rs1_val = X(rs1);
-  let rs2_val = X(rs2);
-  var result : xlenbits = zeros();
-  foreach (i from 0 to (xlen_val - 1))
-    if rs2_val[i] == bitone then result = result ^ (rs1_val << i);
-  X(rd) = result;
+  let prod = carryless_mul(X(rs1), X(rs2));
+  X(rd) = truncate(prod, xlen);
```
Timmmm, line 24: I think `prod[xlen - 1 .. 0]` is a bit clearer and also more type safe IIRC. Ditto other places.

model/prelude.sail
```diff
@@ -191,14 +191,15 @@ function rotatel (v, n) =
...
+function reverse_bits (xs)  = {
+  var ys : bits('n) = zeros();
+  foreach (i from 0 to ('n - 1))
+    ys[i] = xs['n - 1 - i];
   ys
 }
 
-overload reverse = {reverse_bits_in_byte}
+overload reverse = {reverse_bits}
```
Timmmm, line 202: Is `reverse` overloaded with anything else?

model/prelude.sail
```diff
@@ -221,6 +222,17 @@ function log2(n) = {
   result
 }
 
+/* Carry-less multiply. */
```
Timmmm, line 225: I wonder if it would make sense to put this in a new file, `carryless_multiply.sail` or whatever. I dunno if it really belongs in `prelude.sail`.

jordancarlin: Maybe something like `arithmetic.sail` or something along those lines. Then we can add the `brev8` function from #819 and likely others as we continue to clean up the existing code.

model/prelude.sail
pmundkur: It's not, and I  was wondering why it was defined this way, but I left it alone for now.

nadime15 (approved):

model/riscv_insts_zbc.sail
```diff
@@ -60,11 +52,7 @@ mapping clause assembly = RISCV_CLMULR(rs2, rs1, rd)
   <-> "clmulr" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ reg_name(rs2)
 
 function clause execute (RISCV_CLMULR(rs2, rs1, rd)) = {
-  let rs1_val = X(rs1);
-  let rs2_val = X(rs2);
-  var result : xlenbits = zeros();
-  foreach (i from 0 to (xlen_val - 1))
-    if rs2_val[i] == bitone then result = result ^ (rs1_val >> (xlen_val - i - 1));
-  X(rd) = result;
+  let prod = carryless_mul(reverse(X(rs1)), reverse(X(rs2)));
```
Timmmm, line 55: I wonder if it would make sense to have a clmulr function to match the pseudocode in the spec? Otherwise the note in the spec about it being equivalent is going to read kind of weirdly.

The would be a really good application for an SMT property to verify they really are equivalent.

pmundkur: I'm leaning towards leaving this as it is for now, since the code is quite clear.  We could deal with the feedback from spec integration when we actually get the Sail into the spec.

Timmmm: I guess it makes me a little nervous since we don't have any tests for it and it's definitely not an obvious transformation...

pmundkur: The upstream `riscv-software-src/riscv-tests` do have tests for these `cmul` variants, as I said in my PR.  I tweaked  `carryless_mul` and the tests did fail.  A property test would indeed be nice, but perhaps in a separate PR?

Timmmm: Ah ok fair enough. Yeah I'll open an issue.

Timmmm: #824

Timmmm (approved):

PR #823, opened 2025-04-02
Rename pte_is_ptr to pte_is_non_leaf for clarity

david-henley:
This has caught us (Codasip) out before. One could make the argument that a leaf PTE is a pointer to a physical address. Renaming it to `pte_is_non_leaf` removes any ambiguity.


jordancarlin (approved):
Seems reasonable to me.

Timmmm (approved):

arichardson (approved):

PR #826, opened 2025-04-02
Fix PMP test

Timmmm:
Somehow the config changes got through CI without this breakage being detected. Alasdair added a `jq` based system to allow tests to have modified configs, but it wasn't being used for the PMP test which needs PMPs to be enabled (by default they weren't).

I'm not totally convinced by the `jq` approach and to avoid the immediate need to add `jq` everywhere and learn how its filters work I just change the default config to enable PMPs.

I also fixed a couple of missing `DEPENDS`.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -11,7 +11,7 @@
     "memory" : {
         "pmp" : {
             "grain" : 0,
-            "count" : 0
+            "count" : 16
         },
         "misaligned" : {
             "enabled" : false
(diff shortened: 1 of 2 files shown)


jordancarlin:
I don't know that I love the jq approach either, but I'm hesitant to change the default config. Especially before we have a good flow for users to generate their own custom configs.

pmundkur (approved):
This is a reasonable stopgap measure.

Timmmm:
I've been using squash merges where we don't want to preserve the history of a PR. It avoid a pointless merge commit - see how this one looks compared to all the previous ones:

image

Also when we *do* want to preserve history (e.g. when we merged Vector), I edit the merge commit message, because `Merged pull request #234234 from big_dave/wip2` is a terrible default.

I don't think there's any way to enforce any of that on Github unfortunately.

jordancarlin:
See https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches. It probably makes sense to at least enable “Require branches to be up to date before merging” and “Require status checks before merging” to avoid issues like this.

PR #828, opened 2025-04-02
Don't print the clang warning message if a gcc cross-compiler is found

jordancarlin:
Don't print a warning message indicating the clang compiler does not support riscv if a riscv gcc cross-compiler is found. This is common on RHEL systems that have clang but don't support riscv with the system clang.


Timmmm (approved):

test/first_party/CMakeLists.txt
```diff
@@ -63,7 +61,11 @@ if (GCC_BIN_RV64)
 endif()
 
 if (NOT CROSS_COMPILER_COMMAND_RV32 OR NOT CROSS_COMPILER_COMMAND_RV64)
-    message(FATAL_ERROR "No suitable cross-compiler found. We recommend downloading Clang from https://github.com/llvm/llvm-project/releases")
+    if (CLANG_BIN)
+        message(WARNING "Your Clang compiler does not support RISC-V (see '${CLANG_BIN} -print-targets'). Please download one from https://github.com/llvm/llvm-project/releases")
+    else()
+        message(FATAL_ERROR "No suitable cross-compiler found. We recommend downloading Clang from https://github.com/llvm/llvm-project/releases")
+    endif()
```
arichardson, line 68:
```suggestion
    if (CLANG_BIN)
        message(WARNING "Your Clang compiler does not support RISC-V (see '${CLANG_BIN} -print-targets'). Please download one from https://github.com/llvm/llvm-project/releases")
    endif()
    message(FATAL_ERROR "No suitable cross-compiler found. We recommend downloading Clang from https://github.com/llvm/llvm-project/releases")
```

I believe this should still be an unconditional FATAL_ERROR since we didn't find a suitable cross compiler.

arichardson (changes requested):
While removing the warning makes sense, I believe this is wrong.

test/first_party/CMakeLists.txt
jordancarlin: Oops. Good catch. Fixed.

test/first_party/CMakeLists.txt
```diff
@@ -63,7 +61,11 @@ if (GCC_BIN_RV64)
 endif()
 
 if (NOT CROSS_COMPILER_COMMAND_RV32 OR NOT CROSS_COMPILER_COMMAND_RV64)
-    message(FATAL_ERROR "No suitable cross-compiler found. We recommend downloading Clang from https://github.com/llvm/llvm-project/releases")
+    if (CLANG_BIN)
+        message(FATAL_ERROR "Your Clang compiler does not support RISC-V (see '${CLANG_BIN} -print-targets') and another suitable cross-compiler was not found. Please download one from https://github.com/llvm/llvm-project/releases")
```
Timmmm, line 65:
```suggestion
        message(FATAL_ERROR "Your Clang compiler does not support RISC-V (see '${CLANG_BIN} -print-targets') and another suitable cross-compiler was not found.")
```

Otherwise we're repeating ourselves I think.

arichardson (approved):

test/first_party/CMakeLists.txt
jordancarlin: It uses `FATAL_ERROR` so will exit immediately after either message.

Timmmm: Ah yes sorry I didn't see that changed.

Timmmm (approved):

Issue #495, opened 2024-06-14
Refactor `haveExt()` & `sys_enable_ext()` functions


Timmmm:
I think the `haveExtension()` functions are a bit confusingly named. It's unclear whether they mean that the CPU *supports* the extension, or if it is currently *enabled* (which are not necessarily the same if you have writeable `misa`). I believe the answer is the latter, and there are a few `sys_enable_extension()` functions which actually dictate whether the extension is *supported*. That function only exists for some extensions though.

This is confusing. I think we should do the following:

1. Rename all `haveExtension()` functions to `extensionEnabled()`.
2. Rename all `sys_enable_extension()` functions to `sys_support_extension()` or `sys_extension_supported()`.
3. Add `sys_extension_supported()` functions corresponding to all the `haveExtension()` functions, so they are 1:1. Initially we can just hard-code the result rather than adding CLI arguments.
4. Update `legalize_misa()` to use the `sys_extension_supported()` functions (even if they are hard-coded to `true`).

arichardson:
I agree that this separation makes a lot of sense. Not sure it really matters but I wonder if we can use a common prefix for the enabled checks instead of a suffix? `enable<Ext>` isn't particularly nice but at least more explicit that `haveXXX`. I guess we could also use something like `canUse<Ext>`, `is<Ext>Enabled`

Timmmm:
Yeah I was thinking exactly the same thing - common prefix is definitely nicer but I couldn't think of a good way to word it that makes sense.

jrtc27:
I'd rather see the extensions become a (scattered) enum that gets passed to haveExt/isExtEnabled/whatever (scattered) functions.

jordancarlin:
Now that #758 is merged, this is worth revisiting. 

We have the scattered `extensionEnabled(Ext_*)` function for when an extension is currently enabled. I think it makes sense to replace all of the `sys_enable_extension()` functions with another scattered function (maybe `extensionSupported(Ext_*)`) that returns if an extension is supported by the current configuration. Each extension would have a clause in this scattered function that points to the config file (`function clause extensionSupported(Ext_Zaamo) = config extensions.Zaamo.enabled`).

@Alasdair @Timmmm @pmundkur, any thoughts on this?

Timmmm:
I think we have to be really careful about the language since this has the potential to be very confusing. Also RISC-V has a special "implemented" meaning that we might want to use... or avoid:

[...]

PR #830, opened 2025-04-03
Refactor extension support logic

jordancarlin:
- Update config file to use `supported` instead of `enabled`
- Change `extensionEnabled(...)` scattered function to `currentlySupported(...)` see #495 
- Create `hartSupports(...)` scattered function and replace many of the existing `sys_enable_...()` functions with it. So far this was only done for the simple extensions that already had an entry in the config file. Once we agree upon the approach I can update this PR or open a follow up PR to change the rest of the extensions. This also skips F/D for now because of complications with FLEN.

I recommend reviewing each of the commits separately and keeping the history when we merge this. I can break it out into separate PRs, but it all seemed related enough that I thought it would probably be fine as one PR.

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -6,7 +6,20 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
+// Enum with all extensions possibly supported by the model.
 scattered enum extension
+
+// Function used to determine if an extension is supported in the current configuration.
+val hartSupports : extension -> bool
+scattered function hartSupports
+
+// Function used to determine if an extension is currently enabled in the model.
+// This means an extension is supported, *and* any necessary bits are set in the
+// relevant CSRs (misa, mstatus, etc.) to enable its use. It is possible for some
+// extensions to be supported in hardware, but temporarily disabled via a CSR, in
+// which case this function should return false.
+val currentlyEnabled : extension -> bool
+scattered function currentlyEnabled
 
 // Note, these are sorted according to the canonical ordering vaguely described
 // in the `Subset Naming Convention` section of the unprivileged spec.
@@ -15,14 +28,19 @@ scattered enum extension
 enum clause extension = Ext_M
 // Single-Precision Floating-Point
 enum clause extension = Ext_F
+function clause hartSupports(Ext_F) = config extensions.FD.supported
 // Double-Precision Floating-Point
 enum clause extension = Ext_D
+function clause hartSupports(Ext_D) = config extensions.FD.supported // TODO: Separate F and D supported flags
 // Compressed Instructions
 enum clause extension = Ext_C
+function clause hartSupports(Ext_C) = config extensions.C.supported
 // Bit Manipulation
 enum clause extension = Ext_B
+function clause hartSupports(Ext_B) = config extensions.B.supported
 // Vector Operations
 enum clause extension = Ext_V
+function clause hartSupports(Ext_V) = config extensions.V.supported
 // Supervisor
 enum clause extension = Ext_S
 // User
@@ -30,8 +48,10 @@ enum clause extension = Ext_U
 
 // Cache-Block Management Instructions
 enum clause extension = Ext_Zicbom
+function clause hartSupports(Ext_Zicbom) = config extensions.Zicbom.supported
 // Cache-Block Zero Instructions
 enum clause extension = Ext_Zicboz
+function clause hartSupports(Ext_Zicboz) = config extensions.Zicboz.supported
 // Base Counters and Timers
 enum clause extension = Ext_Zicntr
 // Integer Conditional Operations
@@ -61,6 +81,7 @@ enum clause extension = Ext_Zfh
 enum clause extension = Ext_Zfhmin
 // Floating-Point in Integer Registers (single precision)
 enum clause extension = Ext_Zfinx
+function clause hartSupports(Ext_Zfinx) = config extensions.Zfinx.supported
 
 // Floating-Point in Integer Registers (double precision)
 enum clause extension = Ext_Zdinx
@@ -69,6 +90,7 @@ enum clause extension = Ext_Zdinx
 enum clause extension = Ext_Zca
 // Code Size Reduction: additional 16-bit aliases
 enum clause extension = Ext_Zcb
+function clause hartSupports(Ext_Zcb) = config extensions.Zcb.supported
 // Code Size Reduction: compressed double precision floating point loads and stores
 enum clause extension = Ext_Zcd
 // Code Size Reduction: compressed single precision floating point loads and stores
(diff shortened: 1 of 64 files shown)


jordancarlin:
Looks like CI is failing on the smt output. The C tests all passed locally. I'll look into this soon.
```
Error:
riscv_extensions.sail:141.44-80:
141 |function clause cpuSupports(Ext_Sscofpmf) = config extensions.Sscofpmf.supported
    |                                            ^----------------------------------^
    | Runtime configuration is not supported when generating code for target 'smt'
    | An explicit configuration should be provided with the --config option.
```

model/riscv_fdext_control.sail
```diff
@@ -15,18 +15,18 @@
 
 /* **************************************************************** */
 
-function clause extensionEnabled(Ext_F) = (misa[F] == 0b1) & (mstatus[FS] != 0b00)
-function clause extensionEnabled(Ext_D) = (misa[D] == 0b1) & (mstatus[FS] != 0b00) & flen >= 64
-function clause extensionEnabled(Ext_Zfinx) = sys_enable_zfinx()
+function clause currentlyEnabled(Ext_F) = (misa[F] == 0b1) & (mstatus[FS] != 0b00)
```
Timmmm, line 18: So should this have `& cpuSupports(Ext_F)`? Or do we rely on `misa[F]` being read-only zero if that is the case.

Maybe we could do something like

```
function currentlyEnabled(ext : Extension) = cpuSupports(ext) & currentlyEnabledImpl(ext)

scattered function currentlyEnabledImpl...
```

(Better name than `currentlyEnabledImpl` needed!)

radimkrcmar: misa[F] must be zero when F is not implemented, but it would be nice to express it somewhere.

> `function currentlyEnabled(ext : Extension) = cpuSupports(ext) & currentlyEnabledImpl(ext)`

`currentlyEnabledImpl(ext)` without `cpuSupports(ext)` seems like a bug.
I think the function would be better with a guard, so the non-sensical case throws an internal error.

jordancarlin: I'd rather avoid yet another layer of enabled/supported functions for the model. As we've been discussing, 2 is already hard enough for people to follow.

I think it doesn't hurt to explicitly add `& cpuSupports(Ext_F)` for the sake of readability and clearly understanding what the gating conditions are, but relying on `misa[F]` is probably fine from a purely correctness standpoint.

I've also avoided touching `cpuSupports(Ext_F)` (and D) in this first version of this PR because of issues with FLEN. They really need FLEN to be an abstract type that is runtime configurable.

pmundkur: I think it would be clearer to add ` & cpuSupports()` to the `currentlyEnabled()` definitions.

pmundkur (changes requested):
Other than adding `& cpuSupports()` this looks good to me.
The cleanup of `riscv_sys_regs` is nice.

model/riscv_fdext_control.sail
jordancarlin: In this version F/D/etc. don't have clauses defined for `cpuSupports()` because I didn't want to add all of the new config options in the same PR. I could add the clauses and just hardcode them to true for now? Or I could add another commit to this PR that actually adds the rest of these configuration options. Not sure what would be better. @pmundkur @Timmmm

pmundkur: What do you think about adding the `cpuSupports()` clauses for F and D and setting them both to `config extensions.FD.supported`  for now?   We need to eventually separate them in the config file (and somehow handle dependencies like D implies F) through a configuration validator.

jordancarlin: Good solution for now. Updated.

jordancarlin:
Updated `cpuSupports()` to `hartSupports()` per suggestion from Alasdair at the meeting.

arichardson (approved):

model/riscv_fdext_regs.sail
```diff
@@ -103,18 +103,18 @@ register f30 : fregtype
 register f31 : fregtype
 
 function dirty_fd_context() -> unit = {
-  assert(sys_enable_fdext());
+  assert(hartSupports(Ext_F));
```
pmundkur, line 106: Can this assert be strengthened to `currentlyEnabled(Ext_F)`?  If the previous asserts were too weak, we could correct them in this PR.

model/riscv_sys_control.sail
```diff
@@ -309,22 +309,22 @@ function handle_interrupt(i : InterruptType, del_priv : Privilege) -> unit =
...
-
-  if   sys_enable_fdext() & sys_enable_zfinx()
+  misa[A]   = 0b1;                                     /* atomics */
+  misa[C]   = bool_to_bits(hartSupports(Ext_C));        /* RVC */
+  misa[B]   = bool_to_bits(hartSupports(Ext_B));        /* Bit-manipulation */
+  misa[I]   = 0b1;                                     /* base integer ISA */
+  misa[M]   = 0b1;                                     /* integer multiply/divide */
+  misa[U]   = bool_to_bits(sys_enable_user());         /* user-mode */
+  misa[S]   = bool_to_bits(sys_enable_supervisor());   /* supervisor-mode */
+  misa[V]   = bool_to_bits(hartSupports(Ext_V));        /* vector extension */
```
pmundkur, line 319: It would be good to retain the comment alignment on the right.

model/riscv_fdext_regs.sail
```diff
@@ -103,18 +103,18 @@ register f30 : fregtype
...
 function dirty_fd_context_if_present() -> unit = {
-  assert(sys_enable_fdext() != sys_enable_zfinx());
-  if sys_enable_fdext() then dirty_fd_context()
+  assert(hartSupports(Ext_F) != hartSupports(Ext_Zfinx));
+  if hartSupports(Ext_F) then dirty_fd_context()
 }
 
 function rF (Fregno(r) : fregno) -> flenbits = {
-  assert(sys_enable_fdext());
+  assert(hartSupports(Ext_F));
```
pmundkur, line 117: Again, perhaps this should be 
```suggestion
  assert(currentlyEnabled(Ext_F));
```
here and in the various read/write accessors below.

model/riscv_fdext_regs.sail
```diff
@@ -103,18 +103,18 @@ register f30 : fregtype
...
   mstatus[FS] = extStatus_to_bits(Dirty);
   mstatus[SD] = 0b1;
 }
 
 function dirty_fd_context_if_present() -> unit = {
-  assert(sys_enable_fdext() != sys_enable_zfinx());
-  if sys_enable_fdext() then dirty_fd_context()
+  assert(hartSupports(Ext_F) != hartSupports(Ext_Zfinx));
+  if hartSupports(Ext_F) then dirty_fd_context()
```
pmundkur, line 113: Should it be
```suggestion
  if currentlyEnabled(Ext_F) then dirty_fd_context()
```

pmundkur (changes requested):
Perhaps we could strengthen the suggested conditions and asserts while we're here?  Otherwise LGTM.

model/riscv_fdext_regs.sail
jordancarlin: That feels like a separate change that should be part of its own PR. This PR has no functional changes and is just a refactor.

It’s probably worth going through all of the asserts related to extensions and checking if they are necessary/tightening the constraints if possible.

model/riscv_sys_control.sail
jordancarlin: Oops. I had it lined up at one point and then made more changes. Fixed.

pmundkur (approved):

model/riscv_extensions.sail
```diff
@@ -6,32 +6,52 @@
...
 // in the `Subset Naming Convention` section of the unprivileged spec.
 
 // Integer Multiplication and Division; not Machine!
 enum clause extension = Ext_M
 // Single-Precision Floating-Point
 enum clause extension = Ext_F
+function clause hartSupports(Ext_F) = config extensions.FD.supported
 // Double-Precision Floating-Point
 enum clause extension = Ext_D
+function clause hartSupports(Ext_D) = config extensions.FD.supported // TODO: Separate F and D supported flags
```
Timmmm, line 34: We might just have an `flen` option. But yeah there something todo anyway!

PR #836, opened 2025-04-03
Define a reverse carry-less multiply function for cmulr instructions.

### LINKED ISSUE #824: Improve code of reverse carryless multiply
See [this conversation](https://github.com/riscv/sail-riscv/pull/822#discussion_r2023714077). Basically it's a PR that changes the reversed carryless multiple to actually use `reverse()`. The original code and [the current spec](https://riscv-specs.timhutt.co.uk/spec/riscv-isa-release-b143ec8-2025-04-02/riscv-unprivileged.html#insns-clmulr) use the direct implementation:

```
let rs1_val = X(rs1);
let rs2_val = X(rs2);
let output : xlenbits = 0;

foreach (i from 0 to (xlen - 1) by 1) {
   output = if   ((rs2_val >> i) & 1)
            then output ^ (rs1_val >> (xlen - i - 1));
            else output;
}
```

and then the spec also notes

> The clmulr instruction is used to accelerate CRC calculations. The r in the instruction’s mnemonic stands for reversed, as the instruction is equivalent to bit-reversing the inputs, performing a clmul, then bit-reversing the output.

That note is going go be kind of weird if the actual pseudocode is just reversing the inputs and outputs anyway. What is it then equivalent to?

IMO we should add a `clmulr()` function that uses the above direct implementation. This would be faster, a bit simpler, and most importantly would allow embedding the Sail into the spec.

For bonus points we can add an SMT property to prove it is equivalent to the `reverse()` version.

pmundkur:
Also define a SMT property to prove equivalence with the previous implementation.

Update the Makefile.old to use the check_properties target to test the property with Z3, pending CMake integration.

Fixes #824.

### DIFF
--- a/Makefile.old
+++ b/Makefile.old
@@ -67,6 +67,10 @@ SAIL_DEFAULT_INST += riscv_insts_vext_red.sail
 SAIL_DEFAULT_INST += riscv_insts_vext_fp_red.sail
 SAIL_DEFAULT_INST += riscv_insts_zicbom.sail
 SAIL_DEFAULT_INST += riscv_insts_zicboz.sail
+SAIL_DEFAULT_INST += riscv_insts_zvbb.sail
+SAIL_DEFAULT_INST += riscv_insts_zvbc.sail
+SAIL_DEFAULT_INST += riscv_insts_zimop.sail
+SAIL_DEFAULT_INST += riscv_insts_zcmop.sail
 
 SAIL_SEQ_INST  = $(SAIL_DEFAULT_INST) riscv_jalr_seq.sail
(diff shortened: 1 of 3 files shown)


Timmmm (approved):
This is great! Does the property prove successfully?

After this is merged I'll add the SMT check to CMake so it's checked in CI.

pmundkur:
Yes, `make -f Makefile.old check_properties` says:
```
Checking counterexample: cmulr_equivalence.smt2
Solver could not find counterexample
Solver output:
unsat
```
which I think proves the property.

Timmmm:
I reckon it's fine for now. Maybe if it gets a lot bigger...

nadime15 (approved):

jordancarlin (approved):

Issue #580, opened 2024-10-07
Update Linux images


Timmmm:
The Linux images in `os-boot` a) are very out of date, awkward to use because they require patching the model, and don't really belong there anyway.

We should

1. Create up-to-date images and put them in another repo, probably with the ELF's attached to Github releases.
2. Remove the images from `os-boot`, and replace them with instructions (or maybe a script) to use the ELFs from the other repo.

It would be really great if the images could run on all of the major emulators:

1. Sail
2. Spike
3. QEMU
4. Whisper
5. Imperas

(I think that's it?)

Timmmm:
I have got Sail and Spike working, and Imperas partly working with a newer Linux (with no userspace admittedly but I don't know if that really matters). The hard bit is getting console output working on all the emulators, and how to supply the DTB.

1. Sail and Spike support HTIF for console output. OpenSBI supports this and can expose this to Linux, so they're pretty easy. Both support supplying a DTB on the command line.
2. Imperas doesn't support HTIF or supplying DTBs on the command line. OpenSBI can embed the DTB but I'm not sure how to get console output yet.
3. For some reason my Linux image doesn't boot with the default Sail memory limit. Changing it to match Spike makes it work. Probably a good idea anyway.

I haven't tried QEMU or Whisper yet.

PR #837, opened 2025-04-04
More closely align address translation with the specification.

pmundkur:
The specification has a numbered list of steps in the algorithm. The Sail version had implemented access permission checks (Steps 6,8) before the superpage alignment check (Step 5).  This commit fixes the ordering.

There is currently no functional difference between the two orders: all the steps involved raise page-faults corresponding to the access type; the only difference would be the error printed in the execution log.

However, when shadow stack is implemented, it inserts a Step 7 that could raise an access-fault exception that would make the difference in step order visible.

While here, mark in comments the steps of the algorithm for easier correlation with the prose spec.

Also tested with a successful Linux boot.

### DIFF
--- a/model/riscv_vmem.sail
+++ b/model/riscv_vmem.sail
@@ -11,6 +11,11 @@
 // including PTWs (Page Table Walks) and TLBs (Translation Look-aside Buffers)
 // Supported VM modes: Sv32, Sv39, Sv48, Sv57
 
+// The code below implements the steps in the "Virtual Address
+// Translation Process" (abbreviated here to VATP) section of the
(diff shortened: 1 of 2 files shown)


jordancarlin:
> Also tested with a successful Linux boot.

Was this done using the existing images in the `os-boot` directory? Or an alternative approach? Asking because of discussions that we've had about things like #580.

pmundkur:
Yeah, I tested with the `os-boot` image, which worked without needing to patch the model.  #580 would be very worthwhile.

Timmmm (approved):

Issue #808, opened 2025-03-21
Rename Execute access type to InstructionFetch


jordancarlin:
As discussed in https://github.com/riscv/sail-riscv/pull/751#discussion_r2007049652, rename the `Execute` access type to `InstructionFetch`. There are multiple access types that require execute permission to the target memory address, so a more specific name is helpful here. Once we add Zcmt and CLIC, we will need to add another separate access type that obeys things like endianness but still requires execute permission.

PR #839, opened 2025-04-07
Rename the Execute access type to InstructionFetch

Arielfoever:
Do you put a period after your commit messages?

for closing https://github.com/riscv/sail-riscv/issues/808

### DIFF
--- a/model/riscv_pmp_control.sail
+++ b/model/riscv_pmp_control.sail
@@ -15,7 +15,7 @@ function pmpCheckRWX(ent, acc) =
     Read(_)      => ent[R] == 0b1,
     Write(_)     => ent[W] == 0b1,
     ReadWrite(_) => ent[R] == 0b1 & ent[W] == 0b1,
-    Execute()    => ent[X] == 0b1,
+    InstructionFetch() => ent[X] == 0b1,
   }
 
 
@@ -99,7 +99,7 @@ function accessToFault(acc : AccessType(ext_access_type)) -> ExceptionType =
     Read(_)      => E_Load_Access_Fault(),
     Write(_)     => E_SAMO_Access_Fault(),
     ReadWrite(_) => E_SAMO_Access_Fault(),
-    Execute()    => E_Fetch_Access_Fault(),
+    InstructionFetch() => E_Fetch_Access_Fault(),
   }
 
 function pmpCheck forall 'n, 'n > 0. (addr: physaddr, width: int('n), acc: AccessType(ext_access_type), priv: Privilege)
--- a/model/riscv_vmem_pte.sail
+++ b/model/riscv_vmem_pte.sail
@@ -119,7 +119,7 @@ function check_PTE_permission(ac        : AccessType(ext_access_type),
       (ReadWrite(_, _), User)       => (pte_U == 0b1)
                                        & (pte_W == 0b1)
                                        & ((pte_R == 0b1) | ((pte_X == 0b1) & mxr)),
-      (Execute(),       User)       => (pte_U == 0b1) & (pte_X == 0b1),
+      (InstructionFetch(),    User) => (pte_U == 0b1) & (pte_X == 0b1),
       (Read(_),         Supervisor) => ((pte_U == 0b0) | do_sum)
                                        & ((pte_R == 0b1) | ((pte_X == 0b1) & mxr)),
       (Write(_),        Supervisor) => ((pte_U == 0b0) | do_sum)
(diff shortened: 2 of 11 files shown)


Arielfoever:
https://github.com/Arielfoever/sail-riscv/actions/runs/14301895964 looks in a good shape.

model/riscv_vmem_ptw.sail
```diff
@@ -60,7 +60,7 @@ function translationException(a : AccessType(ext_access_type),
     (Read(_), _)                 => E_Load_Page_Fault(),
     (Write(_), PTW_Access())     => E_SAMO_Access_Fault(),
     (Write(_), _)                => E_SAMO_Page_Fault(),
-    (Execute(), PTW_Access())    => E_Fetch_Access_Fault(),
-    (Execute(), _)               => E_Fetch_Page_Fault()
+    (InstructionFetch(), PTW_Access())    => E_Fetch_Access_Fault(),
```
Timmmm, line 63: Probably wouldn't update the alignment of the lines above but at least remove the extra spaces before `=>`.

model/riscv_vmem_types.sail
```diff
@@ -43,7 +43,7 @@ function accessType_to_str (a) =
     Read(_)            => "R",
     Write(_)           => "W",
     ReadWrite(_, _)    => "RW",
-    Execute()          => "X"
+    InstructionFetch() => "X"
```
Timmmm, line 46: While we're changing this line can you add a trailing comma?

Timmmm (commented):
LGTM. I think we should do this. Just some minor formatting nits.

nadime15:
I think it would make more sense to merge #751 first, then rebase this branch before merging it into Main.

nadime15 (commented):
@Timmmm Would it make sense to add InstructionFetch() as an additional type and replace all instances of Execute() with it, instead of deleting Execute()? We could use the Execute() Type then for Zcmt etc.

Arielfoever:
> @Timmmm Would it make sense to add InstructionFetch() as an additional type and replace all instances of Execute() with it, instead of deleting Execute()? We could use the Execute() Type then for Zcmt etc.

good idea and I will deal with it now ~~with my ice cream~~.

model/riscv_vmem_types.sail
Arielfoever: Added.

model/riscv_pmp_control.sail
```diff
@@ -15,7 +15,7 @@ function pmpCheckRWX(ent, acc) =
     Read(_)      => ent[R] == 0b1,
     Write(_)     => ent[W] == 0b1,
     ReadWrite(_) => ent[R] == 0b1 & ent[W] == 0b1,
-    Execute()    => ent[X] == 0b1,
+    InstructionFetch()    => ent[X] == 0b1,
```
Timmmm, line 18:
```suggestion
    InstructionFetch() => ent[X] == 0b1,
```

model/riscv_pmp_control.sail
```diff
@@ -99,7 +99,7 @@ function accessToFault(acc : AccessType(ext_access_type)) -> ExceptionType =
     Read(_)      => E_Load_Access_Fault(),
     Write(_)     => E_SAMO_Access_Fault(),
     ReadWrite(_) => E_SAMO_Access_Fault(),
-    Execute()    => E_Fetch_Access_Fault(),
+    InstructionFetch()    => E_Fetch_Access_Fault(),
```
Timmmm, line 102:
```suggestion
    InstructionFetch() => E_Fetch_Access_Fault(),
```

model/riscv_vmem_pte.sail
```diff
@@ -144,7 +144,7 @@ function update_PTE_Bits forall 'pte_size, 'pte_size in {32, 64} . (
   // Update 'dirty' bit?
   let update_d : bool = (pte_flags[D] == 0b0)
                         & (match a {
-                             Execute()       => false,
+                             InstructionFetch()       => false,
```
Timmmm, line 147:
```suggestion
                             InstructionFetch() => false,
```

Timmmm (changes requested):
Couple of more alignment things that were missed but otherwise LGTM.

I think it's better to just rename this rather than have an `Execute()` we aren't currently using. Any other PRs that need it can add it when it is used.

Since we're now going to try and merge #467 before #751 and that will take a while I vote we just merge this first.

Timmmm:
> Should all the other match clauses be realigned?

I probably wouldn' t bother slavishly realigning all the other ones... Maybe one day the auto-formatter will handle this anyway.

> I guess another option to reduce the indentation would be IFetch?

I think `InstructionFetch` is probably worth the extra length for clarity given it doesn't occur that often.

arichardson:
> > Should all the other match clauses be realigned?
> 
> I probably wouldn' t bother slavishly realigning all the other ones... Maybe one day the auto-formatter will handle this anyway.
> 
> > I guess another option to reduce the indentation would be IFetch?
> 
> I think `InstructionFetch` is probably worth the extra length for clarity given it doesn't occur that often.

Sounds good to me.

arichardson (approved):
Formatting is a bit odd, but I don't really mind. Hopefully will be fixed by the future sail autoformatter anyway.

jordancarlin (approved):

Arielfoever:
> Formatting is a bit odd, but I don't really mind. Hopefully will be fixed by the future sail autoformatter anyway.

I think we need to look forward to https://github.com/rems-project/sail/issues/158

Timmmm:
There is a code formatter - see #264

Seems like there were some cases where it didn't produce the nicest formatting though...

Timmmm (approved):

PR #840, opened 2025-04-07
Use correct effective privilege for cbo.clean/inval

Timmmm:
These permission checks were not respecting `mstatus[MPRV]` (modify privilege).

### DIFF
--- a/model/riscv_insts_zicbom.sail
+++ b/model/riscv_insts_zicbom.sail
@@ -90,8 +90,14 @@ function process_clean_inval(rs1, cbop) = {
           // instruction is permitted to access the cache block is UNSPECIFIED."
           //
           // In this implementation we currently don't allow access for fetches.
-          let exc_read = phys_access_check(Read(Data), cur_privilege, paddr, cache_block_size);
-          let exc_write = phys_access_check(Write(Data), cur_privilege, paddr, cache_block_size);
+
+          // effectivePrivilege() only cares if access type is Execute or not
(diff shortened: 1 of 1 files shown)


model/riscv_insts_zicbom.sail
```diff
@@ -90,8 +90,9 @@ function process_clean_inval(rs1, cbop) = {
           // instruction is permitted to access the cache block is UNSPECIFIED."
           //
           // In this implementation we currently don't allow access for fetches.
-          let exc_read = phys_access_check(Read(Data), cur_privilege, paddr, cache_block_size);
-          let exc_write = phys_access_check(Write(Data), cur_privilege, paddr, cache_block_size);
+          let ep = effectivePrivilege(Read(Data), mstatus, cur_privilege);
```
pmundkur, line 93: It would be helpful to have a comment here explaining why the effective privilege computed for `Read(Data)` here can also be re-used for the `Write(Data)` access check below.

pmundkur (changes requested):
LGTM other than adding a comment.

model/riscv_insts_zicbom.sail
Timmmm: Makes sense. We will also change it to `Cache(CleanInval)` or something when that is available so it will be less weird...

Timmmm: Comment added.

jordancarlin (approved):

pmundkur (approved):

PR #842, opened 2025-04-07
Add CI to produce and compile the Lean backend

jprider63:
This is based on the [CI workflow](https://github.com/opencompl/sail-riscv-lean/blob/103c41c5111a9c03d43a3e785f6f6b4e742227c8/.github/workflows/update-sail-build.yml) by @tobiasgrosser.


.github/workflows/update-sail-build.yml
```diff
@@ -0,0 +1,77 @@
...
+        path: ~/.opam
+        key: ${{ matrix.os }}-${{ matrix.version }}
+
+    - name: Setup opam
+      if: steps.cache-opam-restore.outputs.cache-hit != 'true'
+      run: |
+        opam init --yes --no-setup --shell=sh --compiler=${{ matrix.version }}
+        opam install dune
+
+    - name: Save cached opam
```
jordancarlin, line 39: Any reason not to have this step after building Sail so that it gets cached too? I think this was copied from the sail CI where it obviously wouldn’t make sense to do that, but for this project the CI is primarily for changes to the model and not for changes to Sail.

jordancarlin (changes requested):

.github/workflows/update-sail-build.yml
```diff
@@ -0,0 +1,77 @@
...
+        patch -p1 < .github/workflows/build-riscv-lean-workaround-riscv.patch
+        cmake --build build/ --target generated_lean_rv64d
+
+    - name: Use Lean to build the model and report errors
+      run: |
+        source ~/.profile
+        cd build/model/Lean_RV64D/
+        echo "Set ulimit"
+        ulimit -s unlimited
+        ulimit -s
```
tobiasgrosser, line 76: AFAIU these three are not needed anymore.

.github/workflows/build-riscv-lean-workaround-riscv.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/model/riscv_step.sail b/model/riscv_step.sail
+index 7a6d9450..dd38bdc3 100644
+--- a/model/riscv_step.sail
++++ b/model/riscv_step.sail
+@@ -118,6 +118,7 @@ function loop () : unit -> unit = {
+         /* for now, we drive the platform i/o at every clock tick. */
+         tick_platform();
+         i = 0;
++	()
```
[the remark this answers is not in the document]
Alasdair: Is this still needed?

.github/workflows/compile-lean.yml
```diff
@@ -0,0 +1,71 @@
...
+      run: |
+        wget -q https://raw.githubusercontent.com/leanprover-community/mathlib4/master/scripts/install_debian.sh
+        bash install_debian.sh ; rm -f install_debian.sh
+
+    - name: Build the Sail RISCV Model for Lean
+      run: |
+        eval $(opam config env)
+        cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
+        # This patch is needed to work around https://github.com/rems-project/sail/issues/1003.
+        patch -p1 < .github/workflows/build-riscv-lean-workaround-riscv.patch
```
jordancarlin, line 64: This line should be removed now that the patch is integrated into the main repo.

.github/workflows/compile-lean.yml
```diff
@@ -0,0 +1,71 @@
...
+        cmake --build build/ --target generated_lean_rv64d
+
+    - name: Use Lean to build the model and report errors
+      run: |
+        source ~/.profile
```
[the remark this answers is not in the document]
jprider63: Lean's installation does. It failed trying to use `lake` before I added this.

arichardson:
Maybe the best way to reduce that time would be to have another manual workflow that builds a docker container with sail HEAD and lean and then use that in CI?

jordancarlin:
The slow part is taking the Lean output of the model and then running that with Lean (which is the final step). That part is 30 minutes. So I don't think a Docker image would help.

.github/workflows/build-riscv-lean-workaround-riscv.patch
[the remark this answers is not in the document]
tobiasgrosser: This is not needed anymore. I just dropped this from https://github.com/opencompl/sail-riscv-lean/blob/main/.github/workflows/build-riscv-lean-workaround-riscv.patch and the CI still goes through without any issues during compilation. So this can be removed here.

tobiasgrosser (approved):
From my side this looks great. I don't think there is a meaningful way to reduce the CI time. I would suggest to merge this as is. I am happy to see merges take place without waiting for this CI to finish, especially if it is unlikely to be affected.

arichardson (approved):
40 minutes isn't too bad, especially since it runs in parallel with the other checks.

IMO these ~30 minutes of added CI time will not slow down PRs being merged noticeably.

.github/workflows/compile-lean.yml
```diff
@@ -0,0 +1,69 @@
...
+      run: |
+        eval $(opam config env)
+        cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
+        cmake --build build/ --target generated_lean_rv64d
+
+    - name: Use Lean to build the model and report errors
+      run: |
+        source ~/.profile
+        cd build/model/Lean_RV64D/
+        lake build
```
arichardson, line 69: Can this make use of multiple CPU cores with some kind of -j flag?

tobiasgrosser: This is already run in parallel by default (on the two cores available at GitHub). Even on a 16-core system, this is not a lot faster as the main compilation thread is largely parallel for now.

jordancarlin (approved):
LGTM

Timmmm (approved):

nadime15:
See #848, which fails during the build and test of the Lean backend.

Timmmm:
Good question. In theory it *should*, but I guess if its unstable and the breakages are likely due to bugs in the Sail -> Lean backend, rather than bugs in the PR, then maybe we should change it to allow failure of that workflow.

PR #844, opened 2025-04-08
Use C header generated by Sail

Timmmm:
This switches from manually copy & pasting model declarations to using Sail to generate the header.

The horrible `RISCV_MODEL_HEADER` will hopefully go away when we have runtime-configurable xlen and flen.

Also I think we may end up not using this if we do the C++ `struct Model { #include "model.c" };` trick, but I figured we may as well use it in the mean time.


c_emulator/riscv_sail.h
```diff
@@ -1,79 +1,23 @@
...
-extern bool zrvfi_mem_data_present;
-void zrvfi_get_mem_data(sail_bits *rop, unit);
-void zrvfi_get_v2_support_packet(sail_bits *rop, unit);
-
-// Debugging prints
-unit zprint_rvfi_exec(unit);
-unit zprint_instr_packet(uint64_t);
+// The generated header does not include these functions.
+void model_init();
+void model_fini();
```
arichardson: Is this included from C? Do we need to add (void) here?

Timmmm: Good question. I'll double check.

Timmmm: Ah yeah it is included in `riscv_softfloat.c`. I guess whatever compiler we use in CI defaults to C20 or whatever version removes the need for `void`. I added it anyway just in case.

pmundkur (approved):
Nice!

Issue #841, opened 2025-04-07
Test failed on newer riscv-tests


Arielfoever:
I bump riscv-tests to https://github.com/riscv-software-src/riscv-tests/commit/799d6e807fa1e5b71cc4dace9d75f264c4ff224c and get 5 failed.

You can see my workload on https://github.com/Arielfoever/sail-riscv/actions/runs/14312771316

https://github.com/Arielfoever/riscv-tests/actions/runs/14311988616 is the CI who produced the ELFs thanks to https://github.com/riscv-collab/riscv-gnu-toolchain/releases/tag/2025.01.20

```
99% tests passed, 5 tests failed out of 622                                                     
                                                                                                
Total Test time (real) = 167.17 sec                                                             
                                                
The following tests FAILED:                                                                     
        106 - rv32d_rv32ui-p-ma_data.elf (Failed)                                               
        148 - rv32d_rv32ui-v-ma_data.elf (Failed)                                               
        286 - rv64d_rv64mzicbo-p-zero.elf (Failed)                                              
        406 - rv64d_rv64ui-p-ma_data.elf (Failed)                           
        460 - rv64d_rv64ui-v-ma_data.elf (Timeout)                                              
Errors while running CTest                                                                      
Output from these tests are in: /home/ariel/sail/sail-riscv/build/Testing/Temporary/LastTest.log                                                                                                
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.   
```

for reproduce this

```
ctest --test-dir build --output-junit ~/sail/tests.xml --timeout 3
```

pmundkur:
You will need to enable the `Zicboz` extension and set `memory.misaligned` to true in the configuration.   

It's perhaps best to enable all (compatible) options and allow misaligned accesses in the default config to avoid these kinds of testing issues.  This issue also came up in the ACT meeting today.

pmundkur:
We should probably update the default config after #830 makes it in, since it touches the config as well.

jordancarlin:
> We should probably update the default config after [#830](https://github.com/riscv/sail-riscv/pull/830) makes it in, since it touches the config as well.

I have another PR ready to go after #830 makes it in that adds configuration options for almost all of the existing extensions (skips some tricky dependencies like F/D and C/Zca for now). When I open that I can have all of those new options default to on.

PR #845, opened 2025-04-08
Add configuration options for (almost) all supported extensions

jordancarlin:
This adds `hartSupports` clauses for all of the remaining extensions and connects them to corresponding `supported` keys in the configuration json. All extensions that were hardcoded to true have been updated to use the new configuration system. 

This also enables all of the (non-conflicting) extensions that were previously disabled (as discussed in #841).

There are a handful of strange cases that will require some discussion and might need to be cleaned up in a future PR:
- The interaction between F/D (and Zfinx/Zdinx) and `FLEN`. This probably can't be properly sorted out until `FLEN` becomes an abstract type that can be set at runtime.
- C/Zca/Zcf/Zcd, A/Zaamo/Zalrsc, and B/Zba/Zbb/Zbs: these all have interesting dependencies between the single letter extension and the smaller sub-extensions.
  - C is the most well defined and there is a clear mapping between it and the sub-extensions. We just need to decide if we want to have configuration keys for just C, just the sub-extensions, or both (which would just result in a bunch of ors).
  - The B `misa` bit means that Zba, Zbb, and Zbs are supported, but the lack of it doesn't mean anything since it was defined after the sub-extensions. My vote for this one would be to have a options for each of the sub-extensions and an additional configuration key for the `misa.B` bit itself.
  - Atomics seem to be the least documented case. I couldn't find anything about the relationship between the `misa.A` bit and the Zaamo/Zalrsc extensions. The A extension is made up of those two sub-extensions, but does that mean turning off the A bit should disable them? How does that work if you only implement one of the two (in which case the A bit cannot be set)? We made need to open an issue against the manual to clarify this one.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -38,7 +38,10 @@
         "instructions_per_tick" : 100
     },
     "extensions" : {
-        "C" : {
+        "M" : {
+            "supported" : true
+        },
+        "A" : {
             "supported" : true
         },
         "FD" : {
@@ -51,22 +54,121 @@
             "vl_use_ceil" : false
         },
         "B" : {
+            "supported" : true
+        },
+        "S" : {
+            "supported" : true
+        },
+        "U" : {
+            "supported" : true
+        },
+        "Zicbom" : {
+            "supported" : true
+        },
+        "Zicboz" : {
+            "supported" : true
+        },
+        "Zicond" : {
+            "supported" : true
+        },
+        "Zicsr" : {
+            "supported" : true
+        },
+        "Zicntr" : {
+            "supported" : true
+        },
+        "Zifencei" : {
+            "supported" : true
+        },
+        "Zihpm" : {
+            "supported" : true
+        },
+        "Zimop" : {
+            "supported" : true
+        },
+        "Zmmul" : {
             "supported" : false
         },
-        "Svinval" : {
+        "Zaamo" : {
             "supported" : false
         },
-        "Zcb" : {
+        "Zabha" : {
+            "supported" : true
+        },
+        "Zalrsc" : {
+            "supported" : false
+        },
+        "Zfa" : {
+            "supported" : true
+        },
+        "Zfh" : {
+            "supported" : true
+        },
+        "Zfhmin" : {
             "supported" : false
         },
         "Zfinx" : {
             "supported" : false
         },
-        "Zicbom" : {
+        "Zca" : {
+            "supported" : true
+        },
+        "Zcf" : {
+            "supported" : true
+        },
+        "Zcd" : {
+            "supported" : true
+        },
+        "Zcb" : {
+            "supported" : true
+        },
+        "Zcmop" : {
+            "supported" : true
+        },
+        "Zba" : {
             "supported" : false
         },
-        "Zicboz" : {
+        "Zbb" : {
             "supported" : false
+        },
+        "Zbs" : {
+            "supported" : false
+        },
+        "Zbc" : {
+            "supported" : true
+        },
+        "Zbkb" : {
+            "supported" : true
+        },
+        "Zbkc" : {
+            "supported" : true
+        },
+        "Zbkx" : {
+            "supported" : true
+        },
+        "Zknd" : {
+            "supported" : true
+        },
+        "Zkne" : {
+            "supported" : true
+        },
+        "Zknh" : {
+            "supported" : true
+        },
+        "Zkr" : {
+            "supported" : true
+        },
+        "Zksed" : {
+            "supported" : true
+        },
+        "Zksh" : {
+            "supported" : true
+        },
+        "Zhinx" : {
+            "supported" : false
+        },
+        "Zvbb" : {
+            "supported" : true
         },
         "Zvkb" : {
             "supported" : false
@@ -74,8 +176,17 @@
         "Zvbc" : {
             "supported" : true
         },
+        "Sscofpmf" : {
+            "supported" : true
+        },
+        "Smcntrpmf" : {
+            "supported" : true
+        },
         "Sstc" : {
-            "supported" : false
+            "supported" : true
+        },
+        "Svinval" : {
+            "supported" : true
         }
     }
 }
--- a/config/default.json
+++ b/config/default.json
@@ -71,9 +71,6 @@
         "Zicond" : {
             "supported" : true
         },
-        "Zicsr" : {
-            "supported" : true
-        },
         "Zicntr" : {
             "supported" : true
         },
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -113,7 +113,7 @@ enum clause extension = Ext_Zcd
 function clause hartSupports(Ext_Zcd) = config extensions.Zcd.supported
 // Code Size Reduction: compressed single precision floating point loads and stores
 enum clause extension = Ext_Zcf
-function clause hartSupports(Ext_Zcf) = config extensions.Zcf.supported
+function clause hartSupports(Ext_Zcf) = config extensions.Zcf.supported : bool & (xlen == 32)
 // Compressed May-Be-Operations
 enum clause extension = Ext_Zcmop
 function clause hartSupports(Ext_Zcmop) = config extensions.Zcmop.supported
@@ -123,7 +123,7 @@ function clause hartSupports(Ext_Zcmop) = config extensions.Zcmop.supported
 //   - Zcf is supported if F is supported (on rv32)
 //   - Zcd is supported if D is supported
 enum clause extension = Ext_C
-function clause hartSupports(Ext_C) = hartSupports(Ext_Zca) & (hartSupports(Ext_Zcf) | ~(hartSupports(Ext_F)) | xlen != 32) & (hartSupports(Ext_Zcd) | ~(hartSupports(Ext_D)))
+function clause hartSupports(Ext_C) = hartSupports(Ext_Zca) & (hartSupports(Ext_Zcf) | not(hartSupports(Ext_F)) | xlen != 32) & (hartSupports(Ext_Zcd) | not(hartSupports(Ext_D)))
 
 // Bit Manipulation: Address generation
 enum clause extension = Ext_Zba
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -26,15 +26,16 @@ scattered function currentlyEnabled
 
 // Integer Multiplication and Division; not Machine!
 enum clause extension = Ext_M
+function clause hartSupports(Ext_M) = config extensions.M.supported
+// Atomic Instructions
+enum clause extension = Ext_A
+function clause hartSupports(Ext_A) = config extensions.A.supported
 // Single-Precision Floating-Point
 enum clause extension = Ext_F
 function clause hartSupports(Ext_F) = config extensions.FD.supported
 // Double-Precision Floating-Point
 enum clause extension = Ext_D
 function clause hartSupports(Ext_D) = config extensions.FD.supported // TODO: Separate F and D supported flags
-// Compressed Instructions
-enum clause extension = Ext_C
-function clause hartSupports(Ext_C) = config extensions.C.supported
 // Bit Manipulation
 enum clause extension = Ext_B
 function clause hartSupports(Ext_B) = config extensions.B.supported
@@ -43,8 +44,10 @@ enum clause extension = Ext_V
 function clause hartSupports(Ext_V) = config extensions.V.supported
 // Supervisor
 enum clause extension = Ext_S
+function clause hartSupports(Ext_S) = config extensions.S.supported
 // User
 enum clause extension = Ext_U
+function clause hartSupports(Ext_U) = config extensions.U.supported
 
 // Cache-Block Management Instructions
 enum clause extension = Ext_Zicbom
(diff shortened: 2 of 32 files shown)


Timmmm:
> C is the most well defined and there is a clear mapping between it and the sub-extensions. We just need to decide if we want to have configuration keys for just C, just the sub-extensions, or both (which would just result in a bunch of ors).

I wouldn't say it's super well defined, but my reading is:

* If `misa[C]` is set then
    * Zca is supported and enabled
    * Zcf is supported and enabled if F is supported and enabled
    * Zcd is supported and enabled if D is supported and enabled
    * Other Zc*'s are enabled if they are supported.
* If `misa[C]` is clear then no compressed instructions are enabled.

So I *think* we only need sub-extensions, since there's no way to have `misa[C]` unset but still enable any compressed instructions like you can with B.

> The B misa bit means that Zba, Zbb, and Zbs are supported, but the lack of it doesn't mean anything since it was defined after the sub-extensions. My vote for this one would be to have a options for each of the sub-extensions and an additional configuration key for the misa.B bit itself.

Yes I agree. Have an option for B, Zba, Zbb and Zbs, and then

```
function clause currentlyEnabled(Ext_Zba) = hartSupports(Ext_Zba) | (hartSupports(Ext_B) & misa[B] == 0b1)
```

> Atomics seem to be the least documented case. I couldn't find anything about the relationship between the misa.A bit and the Zaamo/Zalrsc extensions.

Yeah I agree, it's unspecified, but I think it should work the same as A and M. There was a big discussion about M [here](https://github.com/riscv/riscv-isa-manual/issues/869) and they totally didn't clarify it in the spec (just added `M implies Zmmul` which is... not clear) but anyway based on the discussion I think it should be:

```
function clause currentlyEnabled(Ext_Zaamo) = hartSupports(Ext_Zaamo) | (hartSupports(Ext_A) & misa[A] == 0b1)

function clause currentlyEnabled(Ext_Zmmul) = hartSupports(Ext_Zmmul) | (hartSupports(Ext_M) & misa[M] == 0b1)
```

jordancarlin:
Works for me for now. Will push an update for those extensions soon and then it's just F/D which will need to be deferred until later.

pmundkur:
We should also update `legalize_misa`, perhaps in a followup PR.  It doesn't look right (and probably hasn't been for a while).

jordancarlin:
> > C is the most well defined and there is a clear mapping between it and the sub-extensions. We just need to decide if we want to have configuration keys for just C, just the sub-extensions, or both (which would just result in a bunch of ors).
> 
> I wouldn't say it's super well defined, but my reading is:
> 
> * If `misa[C]` is set then
>   
>   * Zca is supported and enabled
>   * Zcf is supported and enabled if F is supported and enabled
>   * Zcd is supported and enabled if D is supported and enabled
>   * Other Zc*'s are enabled if they are supported.
> * If `misa[C]` is clear then no compressed instructions are enabled.
> 
> So I _think_ we only need sub-extensions, since there's no way to have `misa[C]` unset but still enable any compressed instructions like you can with B.

Unfortunately, I don't think this is actually quite right. There is nothing stopping a core from implementing (for example) `rv64if_Zca`. In that case, `misa.C` would not be set (since the processor has `F` but not `Zcf`), but compressed instructions from `Zca` would be allowed.

jordancarlin:
> Unfortunately, I don't think this is actually quite right. There is nothing stopping a core from implementing (for example) `rv64if_Zca`. In that case, `misa.C` would not be set (since the processor has `F` but not `Zcf`), but compressed instructions from `Zca` would be allowed.

Ended up being more subtle than I initially thought, but I think I came up with something that should work:
```
function clause hartSupports(Ext_C) = hartSupports(Ext_Zca) & (hartSupports(Ext_Zcf) | ~(hartSupports(Ext_F)) | xlen != 32) & (hartSupports(Ext_Zcd) | ~(hartSupports(Ext_D)))
function clause currentlyEnabled(Ext_C) = misa[C] == 0b1 & hartSupports(Ext_C)

function clause currentlyEnabled(Ext_Zca) = currentlyEnabled(Ext_C) | (hartSupports(Ext_Zca) & ~(hartSupports(Ext_C)))
function clause currentlyEnabled(Ext_Zcf) = currentlyEnabled(Ext_F) & (currentlyEnabled(Ext_C) | (hartSupports(Ext_Zcf) & currentlyEnabled(Ext_Zca) & ~(hartSupports(Ext_C)))) & xlen == 32
function clause currentlyEnabled(Ext_Zcd) = currentlyEnabled(Ext_D) & (currentlyEnabled(Ext_C) | (hartSupports(Ext_Zcd) & currentlyEnabled(Ext_Zca) & ~(hartSupports(Ext_C)))) & (xlen == 32 | xlen == 64)
```

jordancarlin:
> We should also update `legalize_misa`, perhaps in a followup PR. It doesn't look right (and probably hasn't been for a while).

Added another commit to this PR to take care of `legalize_misa`. Seems similar enough since it is based on the refactored extension handling, but I can break it out into a separate PR if you'd prefer.

model/riscv_insts_zcd.sail
```diff
@@ -6,7 +6,7 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_Zcd) = currentlyEnabled(Ext_Zca) & currentlyEnabled(Ext_D) & (xlen == 32 | xlen == 64)
+function clause currentlyEnabled(Ext_Zcd) = currentlyEnabled(Ext_D) & (currentlyEnabled(Ext_C) | (hartSupports(Ext_Zcd) & currentlyEnabled(Ext_Zca) & ~(hartSupports(Ext_C)))) & (xlen == 32 | xlen == 64)
```
Timmmm, line 9:
```suggestion
function clause currentlyEnabled(Ext_Zcd) = misa[C] == 0b1 & hartSupports(Ext_Zcd) & currentlyEnabled(Ext_D)
```

model/riscv_insts_zcf.sail
```diff
@@ -15,7 +15,7 @@
 
 /* ****************************************************************** */
 
-function clause currentlyEnabled(Ext_Zcf) = currentlyEnabled(Ext_Zca) & currentlyEnabled(Ext_F) & xlen == 32
+function clause currentlyEnabled(Ext_Zcf) = currentlyEnabled(Ext_F) & (currentlyEnabled(Ext_C) | (hartSupports(Ext_Zcf) & currentlyEnabled(Ext_Zca) & ~(hartSupports(Ext_C)))) & xlen == 32
```
Timmmm, line 18: function clause currentlyEnabled(Ext_Zcd) = misa[C] == 0b1 & hartSupports(Ext_Zcf) & currentlyEnabled(Ext_F)

model/riscv_smcntrpmf.sail
```diff
@@ -1,4 +1,4 @@
-function clause currentlyEnabled(Ext_Smcntrpmf) = true
+function clause currentlyEnabled(Ext_Smcntrpmf) = hartSupports(Ext_Smcntrpmf)
```
Timmmm, line 1: Probably depends on `Ext_Zicntr` too?

model/riscv_insts_fext.sail
```diff
@@ -25,11 +25,9 @@
 
 /* **************************************************************** */
 
-// TODO: Add config flags to control Zfh and Zfhmin
-function clause currentlyEnabled(Ext_Zfh) = (misa[F] == 0b1) & (mstatus[FS] != 0b00)
+function clause currentlyEnabled(Ext_Zfh) = (misa[F] == 0b1) & (mstatus[FS] != 0b00) & hartSupports(Ext_Zfh)
 
-// Zfhmin is a subset of Zfh. This can be changed to currentlyEnabled(Ext_Zfh) | sys_enable_zfhmin() when more configuration is implemented.
-function clause currentlyEnabled(Ext_Zfhmin) = currentlyEnabled(Ext_Zfh)
+function clause currentlyEnabled(Ext_Zfhmin) = currentlyEnabled(Ext_Zfh) | ((misa[F] == 0b1) & (mstatus[FS] != 0b00) & hartSupports(Ext_Zfhmin))
```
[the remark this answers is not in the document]
jordancarlin: I'd rather use `currentlyEnabled` instead of `hartSupports` unless there is a good reason to do otherwise. How about this version?
```
function clause currentlyEnabled(Ext_Zfhmin) = currentlyEnabled(Ext_Zfh) | (currentlyEnabled(Ext_F) & hartSupports(Ext_Zfhmin))
```

model/riscv_insts_base.sail
```diff
@@ -10,7 +10,7 @@
 /* This file specifies the instructions in the base integer set.      */
 
 function clause currentlyEnabled(Ext_C) = misa[C] == 0b1 & hartSupports(Ext_C)
-function clause currentlyEnabled(Ext_Zca) = currentlyEnabled(Ext_C)
+function clause currentlyEnabled(Ext_Zca) = currentlyEnabled(Ext_C) | (hartSupports(Ext_Zca) & ~(hartSupports(Ext_C)))
```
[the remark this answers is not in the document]
jordancarlin: I don't think that's quite right. From section 29.4 of the unprivileged spec:

> MISA.C is set if the following extensions are selected:
> 
> - Zca and not F
> - Zca, Zcf and F is specified (RV32 only)
> - Zca, Zcf and Zcd if D is specified (RV32 only)
> - Zca, Zcd if D is specified (RV64 only)

A core can implement Zca and F, and therefore not set `misa.C` while still supporting Zca.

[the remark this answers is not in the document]
Timmmm: Ah yes I think you're right. So setting `misa.C` to 0 only disables `Zca` if `misa.C` is writable. You could potentially make it clearer like this?

```
= hartSupports(Ext_Zca) & (misa[C] == 0b1 | not(hartSupports(Ext_C))
```

Also off-topic but I think we should use `not` instead of `~` for booleans. Less confusing. We should probably remove the overload entirely I reckon.

config/default.json
```diff
@@ -51,31 +54,139 @@
...
+            "supported" : false
+        },
+        "Zfa" : {
+            "supported" : true
+        },
+        "Zfh" : {
+            "supported" : true
+        },
+        "Zfhmin" : {
             "supported" : false
```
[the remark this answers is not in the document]
Timmmm: Ah fair enough.

model/riscv_insts_fext.sail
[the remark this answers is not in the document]
Timmmm: Yeah that sounds good to me.

model/riscv_insts_base.sail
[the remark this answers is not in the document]
jordancarlin: Also replaced uses of `~` with `not` in this PR. We can open a separate issue to look at changing it elsewhere.

model/riscv_insts_zcmop.sail
```diff
@@ -6,7 +6,7 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_Zcmop) = true & currentlyEnabled(Ext_Zca) // TODO: add configuration
+function clause currentlyEnabled(Ext_Zcmop) = currentlyEnabled(Ext_Zca) & hartSupports(Ext_Zcmop)
```
pmundkur, line 9:
```suggestion
function clause currentlyEnabled(Ext_Zcmop) = hartSupports(Ext_Zcmop) & currentlyEnabled(Ext_Zca)
```

model/riscv_insts_zfa.sail
```diff
@@ -6,7 +6,7 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_Zfa) = true
+function clause currentlyEnabled(Ext_Zfa) = hartSupports(Ext_Zfa)
```
pmundkur, line 9:
```suggestion
function clause currentlyEnabled(Ext_Zfa) = hartSupports(Ext_Zfa) & currentlyEnabled(Ext_F)
```
Chapter 26 says "The Zfa extension depends on the F extension."

config/default.json
```diff
@@ -51,31 +54,139 @@
...
+        "Zicbom" : {
+            "supported" : true
+        },
+        "Zicboz" : {
+            "supported" : true
+        },
+        "Zicond" : {
+            "supported" : true
+        },
+        "Zicsr" : {
```
pmundkur, line 74: There doesn't seem to be a `config extensions.Zicsr.supported` in the model?

model/riscv_insts_zfh.sail
```diff
@@ -6,7 +6,7 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_Zhinx) = hartSupports(Ext_Zfinx)
+function clause currentlyEnabled(Ext_Zhinx) = hartSupports(Ext_Zhinx)
```
pmundkur, line 9:
```suggestion
function clause currentlyEnabled(Ext_Zhinx) = hartSupports(Ext_Zhinx) & currentlyEnabled(Ext_finx)
```
Section 27.4 says "The Zhinx extension depends upon the Zfinx extension."

model/riscv_insts_zvbb.sail
```diff
@@ -6,7 +6,7 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_Zvbb) = true
+function clause currentlyEnabled(Ext_Zvbb) = hartSupports(Ext_Zvbb)
```
pmundkur, line 9:
```suggestion
function clause currentlyEnabled(Ext_Zvbb) = hartSupports(Ext_Zvbb) & currentlyEnabled(Ext_V)
```
I couldn't find this explicitly in the spec, but is probably implied?

model/riscv_sys_regs.sail
```diff
@@ -108,35 +108,36 @@ function clause currentlyEnabled(Ext_Sstc) = hartSupports(Ext_Sstc)
...
-  then m
-  else {
-    /* Suppress enabling C if C was disabled at boot (i.e. not supported) */
-    let m = if not(hartSupports(Ext_C)) then m else [m with C = v[C]];
-    /* Suppress updates to misa.{f,d} if disabled at boot */
-    if   not(hartSupports(Ext_F))
-    then m
-    else [m with F = v[F], D = v[D] & v[F]]
-  }
+  if   not(sys_enable_writable_misa()) | (v[C] == 0b0 & (nextPC[1] == bitone | ext_veto_disable_C())) then m
```
pmundkur, line 113:
```suggestion
  if   not(sys_enable_writable_misa()) | (v[C] == 0b0 & (nextPC[1] == bitone | ext_veto_disable_C()))
  then m
```
I find it easier to read when the `then` is broken out from a long predicate like this one

model/riscv_insts_zcd.sail
```diff
@@ -6,7 +6,7 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_Zcd) = currentlyEnabled(Ext_Zca) & currentlyEnabled(Ext_D) & (xlen == 32 | xlen == 64)
+function clause currentlyEnabled(Ext_Zcd) = currentlyEnabled(Ext_D) & hartSupports(Ext_Zcd) & currentlyEnabled(Ext_Zca) & (currentlyEnabled(Ext_C) | not(hartSupports(Ext_C))) & (xlen == 32 | xlen == 64)
```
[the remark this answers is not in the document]
jordancarlin: Agreed. I'll update the order throughout.

config/default.json
jordancarlin: Oops. You're right. We should probably add that since it is an optional extension, but I'll remove it from this PR.

model/riscv_insts_zvbb.sail
jordancarlin: This is a trickier one. All of the vector extensions are a bit mixed up. This really depends on one of the smaller vector extensions, but we don’t have that granularity of support yet. For now we can add this, but we’ll need to refactor soon.

jordancarlin:
Updated all `currentlyEnabled` clauses to have the relevant `hartEnabled` clause as the first component.

model/riscv_extensions.sail
```diff
@@ -54,83 +57,122 @@ enum clause extension = Ext_Zicboz
...
+function clause hartSupports(Ext_Zca) = config extensions.Zca.supported
 // Code Size Reduction: additional 16-bit aliases
 enum clause extension = Ext_Zcb
 function clause hartSupports(Ext_Zcb) = config extensions.Zcb.supported
 // Code Size Reduction: compressed double precision floating point loads and stores
 enum clause extension = Ext_Zcd
+function clause hartSupports(Ext_Zcd) = config extensions.Zcd.supported
 // Code Size Reduction: compressed single precision floating point loads and stores
 enum clause extension = Ext_Zcf
+function clause hartSupports(Ext_Zcf) = config extensions.Zcf.supported
```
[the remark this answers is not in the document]
jordancarlin: I moved it, but note that the type checker isn't able to handle this without an explicit type:
```
function clause hartSupports(Ext_Zcf) = config extensions.Zcf.supported : bool & (xlen == 32)
```

pmundkur (approved):

Timmmm (approved):

Timmmm:
I think we should do another release after this is merged. I'll try and write a nice changelog in the release notes.

nadime15:
Sorry I missed the discussion and maybe the following point was already addressed but I just noticed that Zcd and Zcf have following checks:

`.. & currentlyEnabled(Ext_Zca) & (currentlyEnabled(Ext_C) | not(hartSupports(Ext_C)))`

but `Ext_Zca` is already checking if `Ext_C` is enabled or not supported.

`function clause currentlyEnabled(Ext_Zca) = hartSupports(Ext_Zca) & (currentlyEnabled(Ext_C) | not(hartSupports(Ext_C)))`

Could we remove these checks for `Zcd` and `Zcf`?

pmundkur:
That makes sense.

PR #846, opened 2025-04-09
Update Isabelle and Rocq support

bacam:
This updates the support for generating RISC-V models for the Isabelle/HOL interactive theorem prover (and the intermediate Lem format), primarily to use cmake and deal with various bitrot that's accumulated.

There's also a commit to use the default configuration with the Rocq theorem prover target.


model/CMakeLists.txt
```diff
@@ -487,6 +503,92 @@ foreach (xlen IN ITEMS 32 64)
...
+                        echo
+                        "declare {isabelle} rename field sync_exception_ext = sync_exception_ext_exception"
+                        >>
+                        ${CMAKE_CURRENT_BINARY_DIR}/${arch}_types.lem
+
+                )
+                add_custom_target(generated_lem_${arch} DEPENDS "${arch}.lem")
+            endif()
+
+            # Build Lem definitions, en route to Isabelle or HOL4.
```
jordancarlin, line 540: Copy/paste issue with comment

model/CMakeLists.txt
```diff
@@ -1,6 +1,21 @@
+execute_process(
```
jordancarlin, line 1: Doesn't look like this function is being used anywhere?

jordancarlin (changes requested):

model/CMakeLists.txt
bacam: The `sail_dir` variable it sets is used by the Lem to Isabelle rule.  This code is actually the same as one in the `sail_runtime` directory, so I think I'll try moving it to the top-level `CMakeLists.txt` so that it can be shared.

jordancarlin: Ah. Missed that somehow. Sounds good.

model/CMakeLists.txt
bacam: Fixed, thanks

handwritten_support/riscv_extras.lem
```diff
@@ -153,17 +99,57 @@ val sys_enable_fdext : unit -> bool
 let sys_enable_fdext () = true
 declare ocaml target_rep function sys_enable_fdext = `Platform.enable_fdext`
 
+val sys_enable_svinval : unit -> bool
```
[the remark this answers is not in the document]
bacam: I've removed these, and a few more dead definitions.

Timmmm (approved):
LGTM. Shame about all the `sed`ing. Is there some kind of plan to remove the need for that?

bacam:
> LGTM. Shame about all the `sed`ing. Is there some kind of plan to remove the need for that?

No, unfortunately most of this stuff isn't worth the effort just for this.

jordancarlin (commented):
Looks like the README needs to be updated to mention Rocq instead of Coq. Otherwise LGTM.

bacam:
Thanks.  I took the opportunity to mention Lean too.

jordancarlin:
Looks like there are also some TODO comments at the bottom of the root CMakeLists file that can be removed now.

jordancarlin (approved):
LGTM

PR #847, opened 2025-04-10
Revert "Fix minstreth behaviour"

### LINKED ISSUE #832: Revert minstreth change
Turns out the fix in #760 *wasn't* the intended behaviour. See [here](https://github.com/riscv/riscv-isa-manual/issues/1255#issuecomment-2774065062). So we probably need to revert that. Maybe worth waiting a bit to see if it actually gets clarified in the spec.

Timmmm:
This reverts commit cd54a3780178d8cf3ca0fa29009fa7f14059e014.

It hasn't been clarified in the ISA manual, but it has been clarified in [this Github issue](1) and in a new test in `riscv-tests` that the original behaviour was expected.

Essentially there are two different concepts: architectural CSRs (the instruction counter) and CSR addresses `minstret`, `minstreth`. The ISA manual conflates these concepts. When it says

> an explicit CSR write suppresses and overrides any implicit writes or modifications to the same CSR by the same instruction.

When it says CSR here it means the architectural CSR, not the CSR address.

I tested this against the new `instret_overflow` test in `riscv-tests`.

[1]: https://github.com/riscv/riscv-isa-manual/issues/1255

Fixes #832

### DIFF
--- a/model/riscv_step.sail
+++ b/model/riscv_step.sail
@@ -16,13 +16,12 @@ function step(step_no : int) -> bool = {
   /*
    * This records whether or not minstret should be incremented when
    * the instruction is retired. Since retirement occurs before CSR
-   * writes we initialise it based on mcountinhibit here, before
-   * mcountinhibit is potentially changed. We also mark minstret(h) as not
-   * having been explicitly written yet.
+   * writes we initialise it based on mcountinhibit here, before it is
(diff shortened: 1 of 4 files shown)


jordancarlin (approved):
Really wish this was more clear in the manual, but this LGTM.

Issue #779, opened 2025-03-10
Small improvements to new regidx newtypes


Timmmm:
Now that #617 is merged I think there are a few little things we could do to improve the code:

1. Add operators for `regidx` and remove `regidx_offset`.
2. Add `overload bits_of = {...}` for `regidx_bits`, `physaddr_bits`, etc.
3. Maybe use the same case for `newtype regidx : Regidx`, i.e. `newtype regidx : regidx`. I think it's a bit less to think about. `physaddr`/`virtaddr` does that.

Timmmm:
3 was discussed in the meeting. @Alasdair prefers to change `physaddr`/`virtaddr` case to be `Physaddr` etc. so we'll do that instead.

PR #857, opened 2025-04-15
Add overload bits_of for virtaddr_bits and physaddr_bits

trdthg:
part of #779

Currently I retained other similar functions (`bool_to_bits`, (int) `to_bits`, `exceptionType_to_bits` ...). They seems good for the readability, and also make it easier to goto definitions.


jordancarlin:
The title indicates this is for both regidx and phys/virtaddr, but I only see the change for the addresses.

Timmmm (approved):
Great improvement, thanks! I would also maybe consider renaming `physaddr_bits` to `bits_of_physaddr` so it's a bit more congruent with `bits_of` and less similar to `physaddrbits` (and same for `virt`).

jordancarlin (approved):

arichardson (approved):

PR #858, opened 2025-04-15
Add cmake_policy

### LINKED ISSUE #818: Warning when using DOWNLOAD_GMP option
The following warning is generated when compiling with the `DOWNLOAD_GMP` option. Seems like we need to make a few tweaks because of newer CMake behavior.

```
CMake Warning (dev) at /usr/local/share/cmake-3.31/Modules/ExternalProject/shared_internal_commands.cmake:1276 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
Call Stack (most recent call first):
  /usr/local/share/cmake-3.31/Modules/ExternalProject.cmake:3041 (_ep_add_download_command)
  CMakeLists.txt:77 (ExternalProject_Add)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Timmmm:
This tells CMake that we've tested it up to version 3.31 so it can enable newer policies that are available in versions up to that point and not warn us about them.

Fixes #818. Tested with CMake 4 and `./build_simulators.sh`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.20)
+cmake_policy(VERSION 3.20...3.31)
 
 project(sail_riscv)
 


jordancarlin:
Any reason not to set this to version 4.0 (the current latest version) instead of 3.31?

arichardson (approved):

jordancarlin:
Yeah. As far as I can tell it's meaningless. Not sure why they bumped to 4.0.

Timmmm:
Weird. I'll just merge this for now. We can update to 4 later. I don't think it actually makes any difference since it doesn't look like they added any new policies in 4.0.

jordancarlin (approved):

Issue #860, opened 2025-04-15
Configuration validation


jordancarlin:
We need some way of ensuring the configuration passed to the model is legal. Right now (almost) all of the `hartSupports` clauses blindly accept the values from the json config file without any validation regarding extensions that require other extensions. We have some of this validation baked into the `currentlyEnabled` clauses, but this will just not enable extensions if the dependencies aren't enabled, which probably isn't very intuitive.

My current thought on this is having a `validateConfiguration` function (or something along those lines) that gets called as part of the init function. It would have a list of asserts that check for valid configurations based on the `hartSupports` clauses (and other relevant configuration parameters). That way the model could exit with a descriptive error that guides the user to fixing their configuration file.

I know we are eventually planning to have a tool that generates these configuration files, but people will inevitably end up modifying them by hand so I think this is still necessary.

@Timmmm @pmundkur @Alasdair any thoughts on this or a better approach?

Timmmm:
I think one thing to consider is the formal backends. I may be misremembering this but as I recall they treat the `sys_enable()` callbacks as free functions. I'm not sure if they will for `config` too... If so we might want to have the code legalise the config. For example instead of just blindly believing that the user enabled Sv39 if they enabled Sv48, we should have the `hartSupports(Ext_Sv48)` check both. So that no matter what values the `config` has it will result in a legal configuration.

The big downside there is that it's going to be really confusing when you accidentally enable Sv48 but not Sv39, and then the model silently disables both. So we probably *also* want the validation function. I think your proposal to call it at init time and fill it with asserts is a good one.

@Alasdair @bacam can you let us know if I've totally misremembered how the free functions stuff works?

jordancarlin:
I believe we pass a configuration file to Sail when compiling for one of the theorem provers and these values are all set at compile time.

pmundkur:
I'd prefer to keep minimal dependency checking in the Sail code itself, to roughly what is there now, and to reflect what's in the ISA spec.  The rest of the configuration validation (or automatic enabling of extension dependencies etc) should be in the `c_emulator` or a helper script called by the `c_emulator` if needed.  This script can also generate the RVFI version of the config, so that RVFI could be made to work again that way.

It also helps to avoid duplicating validation logic in the Sail code and in the helper script. The  script could also be more helpful about explaining what's wrong with the config than the Sail model could.  There should be no way of using a config without it going through the script.

One reason to have a script outside of `c_emulator` is so that it can be reused for the other backends.

[...]

PR #861, opened 2025-04-15
 Handle address translation for misaligned loads and stores better

pmundkur:
Refactor the LOAD and STORE instruction so they split misaligned
accesses into multiple sub-accesses and perform address translation
separately. This means we should handle the case where a misaligned
access straddles a page boundary in a sensible way, even if we don't
yet cover the full range of possibilities allowed for any RISC-V
implementation.

There are options for the order in which misaligned happen, i.e. from
high-to-low or from low-to-high as well as the granularity of the splitting,
either all the way to bytes or to the largest aligned size. The splitting
can also be disabled if an implementation supports misaligned accesses in hardware.

In addition tidy up the implementation in a few ways:

- Very long lines on the LOAD encdec were fixed by adding a helper

- Add some linebreaks in the code so it reads as less claustrophobic

- Ensure we use the same names for arguments in encdec/execute/assembly.
  Previously we used 'size' and 'width'. I opted for 'width' consistently.

Co-authored-by: Alasdair Armstrong <alasdair.armstrong@cl.cam.ac.uk>

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -14,7 +14,10 @@
       "count": 16
     },
     "misaligned": {
-      "supported": true
+      "supported": true,
+      "byte_by_byte": false,
+      "order_decreasing": false,
+      "allowed_within_exp": 0
     },
     "translation": {
       "dirty_update": false
--- a/doc/ReadingGuide.md
+++ b/doc/ReadingGuide.md
@@ -80,9 +80,15 @@ such as the platform memory map.
   appropriate access fault. This file also contains definitions that
   are used in the weak memory concurrency model.
 
-- The `riscv_vmem_*.sail` files describe the S-mode address
-  translation. See the [Virtual Memory Notes](./notes_Virtual_Memory.adoc)
-  for details.
+- The `riscv_vmem_{types,pte,ptw,tlb}.sail` and `riscv_vmem.sail`
+  files describe the S-mode address translation.
+  See the [Virtual Memory Notes](./notes_Virtual_Memory.adoc) for
+  details.
+
+- The `riscv_vmem_utils.sail` file provides a higher level interface
+  to virtual memory for load/store style instructions that handles
+  address translation and accesses to misaligned addresses taking
+  platform configuration options into account.
 
 - The `riscv_addr_checks_common.sail` and `riscv_addr_checks.sail`
   contain extension hooks to support the checking and transformation
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -107,7 +107,8 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
   /* normal non-rmem case
    * RMEM: SC is allowed to succeed (but might fail later)
    */
-  match vmem_write(rs1, zeros(), width_bytes, rs2, aq & rl, rl, true) {
+  let data = X(rs2)[width_bytes * 8 - 1 .. 0];
+  match vmem_write(rs1, zeros(), width_bytes, data, aq & rl, rl, true) {
     Ok(b) => {
       X(rd) = zero_extend(bool_bits(~(b)));
       cancel_reservation();
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -70,7 +70,7 @@ function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
   // This is checked during decoding.
   assert(width_bytes <= xlen_bytes);
 
-  match vmem_read(rs1, zeros(), width_bytes, aq, aq & rl, true) {
+  match vmem_read(rs1, zeros(), width_bytes, Read(Data), aq, aq & rl, true) {
     Ok(data) => {
       X(rd) = sign_extend(data);
       RETIRE_SUCCESS
@@ -108,7 +108,7 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
    * RMEM: SC is allowed to succeed (but might fail later)
    */
   let data = X(rs2)[width_bytes * 8 - 1 .. 0];
-  match vmem_write(rs1, zeros(), width_bytes, data, aq & rl, rl, true) {
+  match vmem_write(rs1, zeros(), width_bytes, data, Write(Data), aq & rl, rl, true) {
     Ok(b) => {
       X(rd) = zero_extend(bool_bits(~(b)));
       cancel_reservation();
--- a/model/riscv_insts_aext.sail
+++ b/model/riscv_insts_aext.sail
@@ -70,26 +70,12 @@ function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
   // This is checked during decoding.
   assert(width_bytes <= xlen_bytes);
 
-  /* Get the address, X(rs1) (no offset).
-    * Extensions might perform additional checks on address validity.
-    */
-  match ext_data_get_addr(rs1, zeros(), Read(Data), width_bytes) {
-    Ext_DataAddr_Error(e)  => Ext_DataAddr_Check_Failure(e),
-    Ext_DataAddr_OK(vaddr) => {
-      /* "LR faults like a normal load, even though it's in the AMO major opcode space."
-        * - Andrew Waterman, isa-dev, 10 Jul 2018.
-        */
-      if not(is_aligned(bits_of(vaddr), width))
-      then Memory_Exception(vaddr, E_Load_Addr_Align())
-      else match translateAddr(vaddr, Read(Data)) {
-        TR_Failure(e, _)    => Memory_Exception(vaddr, e),
-        TR_Address(addr, _) =>
-          match mem_read(Read(Data), addr, width_bytes, aq, aq & rl, true) {
-            Ok(result) => { load_reservation(bits_of(addr)); X(rd) = sign_extend(result); RETIRE_SUCCESS },
-            Err(e)     => Memory_Exception(vaddr, e),
-          },
-      }
-    }
+  match vmem_read(rs1, zeros(), width_bytes, aq, aq & rl, true) {
+    Ok(data) => {
+      X(rd) = sign_extend(data);
+      RETIRE_SUCCESS
+    },
+    Err(e) => e,
   }
 }
 
@@ -110,50 +96,24 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
   // This is checked during decoding.
   assert(width_bytes <= xlen_bytes);
 
-  if speculate_conditional () == false then {
-    /* should only happen in rmem
-     * rmem: allow SC to fail very early
+  if speculate_conditional() == false then {
+    /* should only happen in RMEM
+     * RMEM: allow SC to fail very early
      */
-    X(rd) = zero_extend(0b1); RETIRE_SUCCESS
-  } else {
-    /* normal non-rmem case
-      * rmem: SC is allowed to succeed (but might fail later)
-      */
-    /* Get the address, X(rs1) (no offset).
-      * Extensions might perform additional checks on address validity.
-      */
-    match ext_data_get_addr(rs1, zeros(), Write(Data), width_bytes) {
-      Ext_DataAddr_Error(e)  => Ext_DataAddr_Check_Failure(e),
-      Ext_DataAddr_OK(vaddr) => {
-        if not(is_aligned(bits_of(vaddr), width))
-        then Memory_Exception(vaddr, E_SAMO_Addr_Align())
-        else {
-          match translateAddr(vaddr, Write(Data)) {  /* Write and ReadWrite are equivalent here:
-                                                      * both result in a SAMO exception */
-            TR_Failure(e, _) => Memory_Exception(vaddr, e),
-            TR_Address(addr, _) => {
-              // Check reservation with physical address.
-              if not(match_reservation(bits_of(addr))) then {
-                /* cannot happen in rmem */
-                X(rd) = zero_extend(0b1); cancel_reservation(); RETIRE_SUCCESS
-              } else {
-                match mem_write_ea(addr, width_bytes, aq & rl, rl, true) {
-                  Err(e) => Memory_Exception(vaddr, e),
-                  Ok(_)  => {
-                    let rs2_val = X(rs2);
-                    match mem_write_value(addr, width_bytes, rs2_val[width_bytes * 8 - 1 .. 0], aq & rl, rl, true) {
-                      Ok(true)  => { X(rd) = zero_extend(0b0); cancel_reservation(); RETIRE_SUCCESS },
-                      Ok(false) => { X(rd) = zero_extend(0b1); cancel_reservation(); RETIRE_SUCCESS },
-                      Err(e)    => Memory_Exception(vaddr, e),
-                    }
-                  }
-                }
-              }
-            }
-          }
-        }
-      }
-    }
+    X(rd) = zero_extend(0b1);
+    return RETIRE_SUCCESS
+  };
+
+  /* normal non-rmem case
+   * RMEM: SC is allowed to succeed (but might fail later)
+   */
+  match vmem_write(rs1, zeros(), width_bytes, rs2, aq & rl, rl, true) {
+    Ok(b) => {
+      X(rd) = zero_extend(bool_bits(~(b)));
+      cancel_reservation();
+      RETIRE_SUCCESS
+    },
+    Err(e) => e,
   }
 }
 
@@ -195,7 +155,7 @@ function clause execute (AMO(op, aq, rl, rs2, rs1, width, rd)) = {
   match ext_data_get_addr(rs1, zeros(), ReadWrite(Data, Data), width_bytes) {
     Ext_DataAddr_Error(e)  => Ext_DataAddr_Check_Failure(e),
     Ext_DataAddr_OK(vaddr) => {
-      if not(is_aligned(bits_of(vaddr), width))
+      if not(is_aligned_addr(bits_of(vaddr), width))
       then Memory_Exception(vaddr, E_SAMO_Addr_Align())
       else match translateAddr(vaddr, ReadWrite(Data, Data)) {
         TR_Failure(e, _) => Memory_Exception(vaddr, e),
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -276,24 +276,15 @@ union clause ast = LOAD : (bits(12), regidx, regidx, bool, word_width, bool, boo
 
 /* unsigned loads are only present for widths strictly less than xlen,
    signed loads also present for widths equal to xlen */
-mapping clause encdec = LOAD(imm, rs1, rd, is_unsigned, size, false, false)
-  <-> imm @ encdec_reg(rs1) @ bool_bits(is_unsigned) @ size_enc(size) @ encdec_reg(rd) @ 0b0000011
-  when (size_bytes(size) < xlen_bytes) | (not(is_unsigned) & size_bytes(size) <= xlen_bytes)
+function valid_load_encdec(width : word_width, is_unsigned : bool) -> bool =
+  (size_bytes(width) < xlen_bytes) | (not(is_unsigned) & size_bytes(width) <= xlen_bytes)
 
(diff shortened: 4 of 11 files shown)


model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,231 @@
...
+    region_width
+  };
+
+  if width > region_width then return false;
+
+  access_within(vaddr, width, region_width)
+}
+
+val split_misaligned : forall 'width, 'width > 0.
+  (virtaddr, int('width)) -> {'n 'bytes, 'width == 'n * 'bytes & 'bytes > 0. (int('n), int('bytes))}
```
nadime15, line 90: I am not sure  if I overlooked that Sail feature, but I have not encountered it so far (is this new?). So essentially, it allows validating that the return values satisfy certain conditions?

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/* This property demonstrates that when bytes == region_bytes, the access_within check above is
+ * equivalent to a regular alignment check (for a constrained set of inputs to help the SMT solver).
+ */
+$[property]
+function prop_access_within_is_aligned(addr : bits(32), bytes : bits(4)) -> bool = {
+  let bytes = unsigned(zero_extend(32, 0b1) << unsigned(bytes));
+  if bytes > 0 then {
+    access_within(addr, bytes, bytes) == (fmod_int(unsigned(addr), bytes) == 0)
```
nadime15, line 60: What is `fmod_int()` doing?

nadime15 (commented):
Looks awesome! Just a few comments

model/riscv_insts_base.sail
```diff
@@ -370,31 +376,32 @@ function clause execute (STORE(imm, rs2, rs1, width, aq, rl)) = {
...
+      };
+
+      let data = X(rs2)[width_bytes * 8 - 1 .. 0];
+
+      match vmem_write(vaddr, width_bytes, data, aq, rl, false) {
+        Ok(_) => RETIRE_SUCCESS,
+        Err(vaddr, e) => {
+          handle_mem_exception(vaddr, e);
+          return RETIRE_FAIL
         }
```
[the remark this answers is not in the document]
pmundkur: This will change anyway once I rebase on top of #755.

model/riscv_insts_fext.sail
```diff
@@ -301,16 +301,11 @@ function clause execute(LOAD_FP(imm, rs1, rd, width)) = {
...
-          match mem_read(Read(Data), addr, width_bytes, aq, rl, res) {
-            Ok(result) => { F(rd) = nan_box(result); RETIRE_SUCCESS },
-            Err(e)     => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
-          }
-        },
+      then { handle_mem_exception(vaddr, E_Load_Addr_Align()); return RETIRE_FAIL };
+      let (aq, rl, res) = (false, false, false);
+      match vmem_read(vaddr, width_bytes, aq, rl, res) {
+        Err(vaddr, e) => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
+        Ok(result)    => { F(rd) = nan_box(result); RETIRE_SUCCESS }
```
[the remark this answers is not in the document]
pmundkur: Thanks, done.

model/riscv_vmem_utils.sail
pmundkur: Yes, I think it's an existential type; it's mentioned in passing in the Sail manual (see the `val div1` example).  It's been around for a while I think.

model/riscv_vmem_utils.sail
pmundkur: Good question.  It's integer modulus.  I now notice there are two of these: `emod_int` and `fmod_int`. The first always returns a  positive value, the second ('f' stands for 'floor' according to GMP docs) retains the sign of the divisor.  Not sure the difference matters here.  @Alasdair?

model/riscv_vmem_utils.sail
Timmmm: It's not new. Not sure exactly what it's called (refinement type, liquid type, existential type, etc. take a guess!) but yeah it lets you add conditions to a type. I think we use them in some type aliases too; it doesn't have to be in the return position.

You can use them in argument positions too but we generally use the `forall` syntax instead.

model/riscv_vmem_utils.sail
Alasdair: The only reason we really have euclidian division is it's the definition used in the SMT integer theory

jordancarlin: Seems like since these are both positive, they’d all be the same? In which case we might as well use `%`, which is already defined in the prelude to mean `emod`.

pmundkur: The SMT properties check out in both cases, so I'll switch it to '%' when I rebase.

pmundkur:
The last push rebased on master and refactored #467 to pull the `ext_data_get_addr` calls into the `vmem_utils` helpers.  This cleans up the calling code significantly, and (hopefully) isolates almost all the RVWMO-relevant pieces for usual loads/stores into `vmem_utils`.  

As before, AMOs, fetch, and CBOs are untouched.

There is a slight difference in alignment checks for LOADRES/STORECON in the read/write helpers.  I'll try to clean it up before final merge.

@Alasdair could you check if this breaks RVWMO modelling in any way?  There's a comment in `vmem_utils` that I retained from your #467 that may need adjusting?

pmundkur:
Hmm, the CI failure is odd.  It's building locally (and passing tests) with Sail 0.19.

pmundkur:
> Hmm, the CI failure is odd. It's building locally (and passing tests) with Sail 0.19.

Drat, I think this is triggering a bug in the smt backend of Sail, both 0.19 and the latest master.

pmundkur:
@bacam Could you take a look at this Rocq failure?  Perhaps it's being triggered by the `match` being inside a `repeat`?  Adding a `let _ : unit = match { ... };` annotation did not help.

jordancarlin (commented):
Overall looking like an amazing simplification!

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,348 @@
...
+
+/* This is an external option that, when true, causes all misaligned accesses
+ * to be split into single byte operations.
+ */
+function sys_misaligned_to_byte() -> bool = config memory.misaligned.to_byte
+
+/* This is an external option that returns an integer N, such that
+ * when N is greater than zero, misaligned accesses to physical memory
+ * (as atomic events) are allowed provided the access occurs within a N
+ * byte region. N must be a power of two.
```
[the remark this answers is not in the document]
pmundkur: The type constraint cannot be captured in the JSON schema, so Sail throws an error:
```
Failed to generate JSON Schema for configuration type {('n : Int) ('m : Int), ('m >= 0 & 'n == 2 ^ 'm). int('n)}
```
This check should go into the config validation checker instead of being checked every time in calls to these functions.  We should start making a list of the things to check.

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,348 @@
...
+}
+
+val allowed_misaligned : forall 'width, 'width > 0. (xlenbits, int('width)) -> bool
+
+function allowed_misaligned(vaddr, width) = {
+  let region_width = min_int(2 ^ pagesize_bits, sys_misaligned_allowed_within());
+
+  // If the Zama16b extension is enabled, the region_width must be at least 16
+  let region_width = if currentlyEnabled(Ext_Zama16b) then {
+    max_int(16, region_width)
```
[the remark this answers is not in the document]
pmundkur: Fixed.

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,348 @@
...
+    (0, n - 1, 1)
+  }
+}
+
+val vmem_write_addr : forall 'width, 'width in {1, 2, 4, 8} & ('width <= xlen_bytes | 'width <= flen_bytes).
+  (virtaddr, int('width), bits(8 * 'width), bool, bool, bool) -> result(bool, Retire_Failure)
+
+function vmem_write_addr(vaddr, width, data, aq, rl, res) = {
+  /* If the store is misaligned, split into `n` (single-copy-atomic) memory operations,
+     each of `bytes` width. If the store is aligned, then `n` = 1 and bytes will remain
```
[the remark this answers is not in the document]
pmundkur: Fixed.

doc/ReadingGuide.md
```diff
@@ -80,9 +80,15 @@ such as the platform memory map.
...
 
-- The `riscv_vmem_*.sail` files describe the S-mode address
-  translation. See the [Virtual Memory Notes](./notes_Virtual_Memory.adoc)
-  for details.
+- The `riscv_vmem_{types,pte,ptw,tlb}.sail` and `riscv_vmem.sail`
+  files describe the S-mode address translation.
+  See the [Virtual Memory Notes](./notes_Virtual_Memory.adoc) for
+  details.
+
+- The `riscv_vmem_utils.sail` file provides a higher level interface
```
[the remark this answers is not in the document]
pmundkur: I'm not sure what can go in there in addition to what's already described in comments in `riscv_vmem_utils.sail`.

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,352 @@
...
+    then return Err(Memory_Exception(vaddr, E_Load_Addr_Align()));
+  } else {
+    if check_misaligned(vaddr, size_bytes(width))
+    then return Err(Memory_Exception(vaddr, E_Load_Addr_Align()));
+  };
+
+  /* If the load is misaligned or an allowed misaligned access, split into `n`
+     (single-copy-atomic) memory operations, each of `bytes` width. If the load is
+     aligned, then `n` = 1 and bytes will remain unchanged. */
+  let ('n, bytes) = split_misaligned(vaddr, width);
```
bacam, line 223: We can work around the Rocq problem by changing `bytes` to `'bytes` - this gives `data` a printable type that can be used in rewrites.  (Rocq also needs a couple of termination measures, but I'll give those separately.)

bacam:
For Rocq output we also need to add two termination measures to `riscv_termination.sail`:
```
termination_measure vmem_write_addr repeat n
termination_measure vmem_read repeat n
```
Although it would be helpful if someone could check that `n` is the correct limit in both cases.

pmundkur:
> For Rocq output we also need to add two termination measures to `riscv_termination.sail`:
> 
> ```
> termination_measure vmem_write_addr repeat n
> termination_measure vmem_read repeat n
> ```
> 
> Although it would be helpful if someone could check that `n` is the correct limit in both cases.

Yeah, I think `n` is the correct limit since the loop should repeat at most `n` times in both cases.

Your suggestions fixed the build.  Thanks!

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,270 @@
...
+  access_within(addr, 1, 1)
+}
+
+val allowed_misaligned : forall 'width, 'width > 0. (xlenbits, int('width)) -> bool
+
+function allowed_misaligned(vaddr, width) = {
+  let region_width = min(2 ^ pagesize_bits, sys_misaligned_allowed_within());
+
+  // If the Zama16b extension is enabled, the region_width must be at least 16
+  let region_width = if currentlyEnabled(Ext_Zama16b) then {
```
jordancarlin, line 88: The interaction between `Zama16b` and `allowed_within` still needs to be resolved: https://github.com/riscv/sail-riscv/issues/860#issuecomment-2818649589. As discussed in that issue, I think Zama16b should not directly impact this function. Once we have configuration validation set up, we can have an assertion that if Zama16b is supported then allowed_within >= 16. For this PR, I think I would be in favor of just dropping the Zama16b configuration option altogether.

jordancarlin (changes requested):

nadime15 (approved):
Looks good!

It would be great to prioritize and get this merged so work on the other PRs can move forward!

jordancarlin (approved):
LGTM now that the configuration issue has been resolved. This will be great to get in!

pmundkur:
I'll merge this tomorrow.

model/riscv_insts_aext.sail
```diff
@@ -70,26 +70,12 @@ function clause execute(LOADRES(aq, rl, rs1, width, rd)) = {
...
-      else match translateAddr(vaddr, Read(Data)) {
-        TR_Failure(e, _)    => Memory_Exception(vaddr, e),
-        TR_Address(addr, _) =>
-          match mem_read(Read(Data), addr, width_bytes, aq, aq & rl, true) {
-            Ok(result) => { load_reservation(bits_of(addr)); X(rd) = sign_extend(result); RETIRE_SUCCESS },
-            Err(e)     => Memory_Exception(vaddr, e),
-          },
-      }
-    }
+  match vmem_read(rs1, zeros(), width_bytes, aq, aq & rl, true) {
```
Timmmm, line 73: Isn't this missing `Read(Data)`?

model/riscv_insts_aext.sail
```diff
@@ -110,50 +96,25 @@ function clause execute (STORECON(aq, rl, rs2, rs1, width, rd)) = {
...
+    return RETIRE_SUCCESS
+  };
+
+  /* normal non-rmem case
+   * RMEM: SC is allowed to succeed (but might fail later)
+   */
+  let data = X(rs2)[width_bytes * 8 - 1 .. 0];
+  match vmem_write(rs1, zeros(), width_bytes, data, aq & rl, rl, true) {
+    Ok(b) => {
+      X(rd) = zero_extend(if b then 0b0 else 0b1);
```
Timmmm, line 113:
```suggestion
      X(rd) = zero_extend(bool_bits(b));
```

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,261 @@
...
+ *   vmem_write       - writes a specified data payload
+ */
+
+/* This is a external option that controls the order in which the model
+ * performs misaligned accesses.
+ */
+function sys_misaligned_order_decreasing() -> bool = config memory.misaligned.order_decreasing
+
+/* This is an external option that, when true, causes all misaligned accesses
+ * to be split into single byte operations.
```
Timmmm, line 36: Worth adding a comment to say what happens if it is false.

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,261 @@
...
+ */
+function sys_misaligned_to_byte() -> bool = config memory.misaligned.to_byte
```
Timmmm, line 38: I think `byte_by_byte` would be a better name?

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,261 @@
...
+
+/* This is an external option that returns an integer N, such that
+ * when N is greater than zero, misaligned accesses to physical memory
+ * (as atomic events) are allowed provided the access occurs within a N
+ * byte region. N must be a power of two.
+ *
+ * This option cannot be unlimited (i.e. allowing all misaligned accesses
+ * to be atomic) as this would be incorrect above the maximum supported
+ * page size.
```
Timmmm, line 47: The `nat` type says that it *is* unlimited though. I think we should change the comment to say this can be a maximum of the page size (4096 bytes), and then change line 83 to be an `assert` rather than just silently modifying the setting.

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,261 @@
...
+
+/* This is an external option that, when true, causes all misaligned accesses
+ * to be split into single byte operations.
+ */
+function sys_misaligned_to_byte() -> bool = config memory.misaligned.to_byte
+
+/* This is an external option that returns an integer N, such that
+ * when N is greater than zero, misaligned accesses to physical memory
+ * (as atomic events) are allowed provided the access occurs within a N
+ * byte region. N must be a power of two.
```
Timmmm, line 43: Doesn't it have to be with a *naturally aligned* N byte region?

Also the power of 2 isn't verified anywhere. In fact we should probably just have the exponent as the setting. It would greatly simplify `access_within()`. Something like this I think?

```
function access_within(addr, bytes, region_bytes_exp) = {
  let mask : bits('width) = zero_extend(ones(region_bytes_exp));
  (addr + bytes) & mask == addr & mask
}
```

model/riscv_vmem_utils.sail
```diff
@@ -0,0 +1,261 @@
...
+ * to be atomic) as this would be incorrect above the maximum supported
+ * page size.
+ */
+function sys_misaligned_allowed_within() -> nat = config memory.misaligned.allowed_within
+
+/* Check if an 'n byte access for an address is within an aligned 'r byte region */
+val access_within : forall 'width 'n 'r, 1 <= 'n <= 'r. (bits('width), int('n), int('r)) -> bool
+
+function access_within(addr, bytes, region_bytes) = {
+  let low = addr & ~(get_slice_int(length(addr), region_bytes - 1, 0));
```
Timmmm, line 55: I think `to_bits()` is exactly what this `get_slice_int` is doing?

Timmmm (commented):
Do you mind waiting a day for me to take a more detailed look? Sorry I've been putting it off but I'll look tomorrow. I'd like to make sure it works for us and also with #866.

Also I think we really need to clean up the `AccessType` system because the behaviour is kind of half split between `AccessType` and the `res`/`aq`/`con` flags. I kind of think we should remove those flags and move everything into `AccessType`. (In future of course.)

model/riscv_insts_aext.sail
pmundkur: Since `vmem_read` is supposed to be a high-level utility API, it doesn't require the caller to specify `Read(Data)` (and similarly for `vmem_write`).

model/riscv_vmem_utils.sail
pmundkur: I made the config a range of exponents, which should ensure the power of two condition as well as the maximum.

model/riscv_vmem_utils.sail
pmundkur: See the other comment.

model/riscv_vmem_utils.sail
pmundkur: This is not needed after switching to the exponent-based version.

config/default.json
```diff
@@ -14,7 +14,10 @@
       "count": 16
     },
     "misaligned": {
-      "supported": true
+      "supported": true,
+      "byte_by_byte": false,
+      "order_decreasing": false,
+      "allowed_within": 0
```
jordancarlin, line 20: The convention in keeping with the other parameters would be to call this `allowed_within_exp`.

model/riscv_insts_aext.sail
Timmmm: I think the access type *is* part of the high level API though. For example we use a different value for loading/storing capabilities in CHERI.

Ultimately I think we will end up with `AccessType` something like this:

```
  Read            : 'a,
  Write           : 'a,
  // (Read, Write, operation)
  AMO           : ('a, 'a, amoop),
  // (acquire, release)
  LoadReserved : ('a, bool, bool),
  StoreConditional : ('a, bool, bool),
  InstructionFetch : unit,
  Cache         : CacheAccessType,
  JumpTableFetch : 'a,
}
```

and then remove the `aq`, `rel`, etc. flags. We don't have to do that now but I think we should have the parameter at least so it is easier to fix later.

model/riscv_vmem_utils.sail
pmundkur: Added.  We'll eventually need to collect all these comments into a documentation page explaining the settings in the config file.

model/riscv_insts_aext.sail
pmundkur: Yeah, we can address those issues later.  The API may also need tweaking for pointer masking, etc.  Let's get the simple version in first.

config/default.json
pmundkur: fixed.

model/riscv_insts_aext.sail
Timmmm: I think it's easy enough to add back now - I can do it if you like? It's just lifting the parameter.

pmundkur: Added.

Timmmm (approved):
Thanks, LGTM!

I think we will probably want to make this more configurable in future, but it at least fixes the page straddling problem which is currently totally broken... :rocket:

Issue #745, opened 2025-02-17
Test build_emulators.sh and DOWNLOAD_GMP in CI


Timmmm:
We have a couple of features that aren't tested in CI and could easily bitrot:

* `-DDOWNLOAD_GMP` to download GMP and compile it from source.
* The `./build_emulators.sh` script.

The latter uses the former, so we can just add `./build_emulators.sh` to CI.

jordancarlin:
Took a look at this, and apparently gmplib.org blocks GitHub Actions IP addresses, so this can't currently be tested in CI: https://github.com/actions/runner-images/issues/7901. Seems like one solution is to download from https://ftp.gnu.org/gnu/gmp/ instead.

Arielfoever:
Suggest doing these:

- add a ci to build everything with ./build_emulators.sh
- add some mirrors of GMP

I wiil do the first one now

PR #863, opened 2025-04-16
Add URL_HASH to GMP download

### LINKED ISSUE #859: Add URL_HASH to GMP download
I noticed when you build with DOWNLOAD_GMP you can get a warning like this:

```
-- File already exists but no hash specified (use URL_HASH):
  file='/home/me/sail-riscv/build/gmp-prefix/src/gmp-6.3.0.tar.xz'
Old file will be removed and new file downloaded from URL.
```

Seems like it probably redownloads that file more than it needs to because it doesn't know the hash. We should add it - see `URL_HASH` [here](https://cmake.org/cmake/help/latest/module/ExternalProject.html).

trdthg:
resolve #859

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,7 @@ if (DOWNLOAD_GMP)
     ExternalProject_Add(
         gmp
         URL https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):
We should really add this to CI so that we know changes like this don't break the GMP download option (#745). Seems fine for now though.

PR #864, opened 2025-04-16
Update Rocq termination information and add CI

bacam:
The termination measures should be fine (although I wish I had a nice way to generate most of these...)

The CI is based on a standard Rocq Docker image and uses the Sail release binary like the main CI job.  I suppose it could be made a job within the main workflow, but I think a separate workflow is better because it can be easily disabled if it causes trouble.


Timmmm (approved):

jordancarlin (approved):
This file seems like it’ll be a pain to maintain, but LGTM.

PR #865, opened 2025-04-16
Add termination measures to the Lean backend

ineol:
This commit enables the use of termination measures by the Lean backend.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -201,7 +201,7 @@ foreach (xlen IN ITEMS 32 64)
                 "riscv_step.sail"
             )
 
-            if (variant STREQUAL "rocq")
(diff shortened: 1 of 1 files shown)


bacam:
This overlaps quite a lot with #864, which includes updating all the termination measures.  Also, I think using `$iftarget` to restrict the `loop` limit to Rocq would probably be better than adding another termination file.

ineol:
Ah I didn't see your PR sorry! Maybe it's easiest if you added the `$iftarget` in your PR to avoid creating dependencies between PRs?

bacam:
Sure, I'll do that now.

jordancarlin (approved):

ineol:
@Timmmm I think it can safely be merged?

PR #867, opened 2025-04-16
Fix macOS compilation and add to CI

urbanij:
Unfortunately since a Sail compiler binary for Mac is not yet available this uses OPAM. Once one is available we'll switch to that.

This also fixes a dependency ordering error on all platforms - `sail_runtime` depends on GMP but this wasn't declared so it could fail if you didn't have GMP already available (i.e. you were building it from source).

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -3,7 +3,7 @@ name: CI
 on: [push, pull_request, workflow_dispatch, merge_group]
 
 jobs:
-  build:
+  build_matrix:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -75,3 +75,13 @@ jobs:
         with:
           name: event.json
           path: ${{ github.event_path }}
+
+  # This is here so that the "Status checks that are required" Github
+  # option doesn't need to list all of the matrix jobs (it doesn't seem
+  # to work if you just specify the job ID of the matrix itself).
+  build:
+    runs-on: ubuntu-latest
+    needs: [build_matrix]
+    steps:
+      - name: Dummy
+        run: echo 'Jobs need at least one step.'
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -4,20 +4,49 @@ on: [push, pull_request, workflow_dispatch, merge_group]
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-22.04, macos-latest]
+    env:
+      SAIL_VERSION: "0.19"
+
     steps:
-      - name: Install packages
-        run: sudo apt install -y --no-install-recommends pkg-config libgmp-dev curl ninja-build jq
-
-      - name: Install sail from binary
-        run: |
-          sudo mkdir -p /usr/local
-          curl --location https://github.com/rems-project/sail/releases/download/0.19-linux-binary/sail.tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
-
       - name: Check out repository code
         uses: actions/checkout@v4
 
+      - name: Install packages (Linux)
+        if: startsWith(matrix.os, 'ubuntu-')
+        run: |
+          sudo apt install -y --no-install-recommends pkg-config libgmp-dev curl ninja-build jq
+
+      - name: Install packages (macOS)
+        if: startsWith(matrix.os, 'macos-')
+        run: |
+          brew install opam llvm lld
+          echo "$(brew --prefix llvm)/bin" >> $GITHUB_PATH
+
+      - name: Install Sail (Linux)
+        if: startsWith(matrix.os, 'ubuntu-')
+        run: |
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -97,7 +97,7 @@ if (DOWNLOAD_GMP)
             INTERFACE_INCLUDE_DIRECTORIES "${INSTALL_DIR}/include")
     add_dependencies(GMP::GMP gmp)
 else()
-    find_package(GMP)
+    find_package(GMP REQUIRED)
```
arichardson, line 100: Required is missing here on purpose to provide a nicer error message, please drop this change.

sail_runtime/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ target_include_directories(sail_runtime
     PUBLIC "${sail_dir}/lib"
 )
 target_compile_options(sail_runtime PRIVATE -Wno-extra -Wno-unused)
+target_link_libraries(sail_runtime PUBLIC GMP::GMP)
```
Timmmm, line 21: I think with this here you can remove `GMP::GMP` from here:

```
    target_link_libraries(riscv_sim_${arch}
        PRIVATE softfloat sail_runtime default_config GMP::GMP
    )
```

.github/workflows/compile.yml
```diff
@@ -4,20 +4,57 @@ on: [push, pull_request, workflow_dispatch]
...
+        path: ~/.opam
+        key: opam-${{ runner.os }}-${{ hashFiles('*.opam', 'opam.locked') }}
+        restore-keys: |
+          opam-${{ runner.os }}-
+
+    - name: Set up OPAM if cache miss (macOS)
+      if: startsWith(matrix.os, 'macos-') && steps.opam-cache.outputs.cache-hit != 'true'
+      run: |
+        opam init --auto-setup --bare
+        opam switch create default 4.14.1
```
Timmmm, line 47: Probably could use OCaml 5 here?

.github/workflows/compile.yml
```diff
@@ -4,20 +4,57 @@ on: [push, pull_request, workflow_dispatch]
...
+        eval $(opam env)
+        opam update
+        opam install -y sail
```
Timmmm, line 50: I think we should lock the version (don't ask me how!).

Timmmm (changes requested):
Looks pretty good to me. Just a few minor comments.

.github/workflows/compile.yml
```diff
@@ -4,20 +4,47 @@ on: [push, pull_request, workflow_dispatch]
...
-      - name: Check out repository code
-        uses: actions/checkout@v4
+      - name: Install Sail (macOS)
+        if: startsWith(matrix.os, 'macos-')
+        run: |
+          opam init --auto-setup --bare
+          opam switch create default 5.3.0
+          eval $(opam env)
+          opam update
+          opam install -y sail.0.19
```
jordancarlin, line 40: Maybe make the version an environment variable (I think you can specify it at the top of the workflow file) so that Linux and macOS can easily download the same version and they don't accidentally get out of sync.

jordancarlin (approved):
Would be nice to eventually have a binary release of Sail for macOS that we could use here instead, but for now this LGTM.

@pmundkur We probably need to change the CI jobs that are required to pass when merging this since it renames them to include the OS now.

Timmmm (approved):
Yeah I agree, but I guess we can live with it for now. If anyone wants to buy me a mac... :-D

arichardson (approved):

Timmmm:
Something slightly odd is going on. Apparently closing and reopening this might fix it so I'm going to try it.

Timmmm:
Nope, that didn't work. I suspect the issue is that we have our setting to require the `build` job to pass but internally Github is renaming it to `build (macos-latest)` and `build (ubuntu-22.04)`.

image

@pmundkur do you think you could give me admin so I can try to figure it out?

We might want to change it to depend on `Test Results` and then have that job depend on all `build` jobs.

jordancarlin:
> We might want to change it to depend on `Test Results` and then have that job depend on all `build` jobs.

Maybe not `Test Results` since that tries to write to a pull request. It probably makes sense to define a `check` job or something that depends on all of the CI jobs that we want PRs to pass. We could also reenable the Lean and Rocq CI actions then but not include them as dependencies of `check` so that we can see if there are new failures without them blocking PRs.

.github/workflows/compile.yml
```diff
@@ -33,16 +62,26 @@ jobs:
...
+        if: startsWith(matrix.os, 'ubuntu-')
         uses: actions/upload-artifact@v4
         with:
           name: event.json
           path: ${{ github.event_path }}
+
+  # This is here so that the "Status checks that are required" Github
+  # option doesn't need to list all of the matrix jobs (it doesn't seem
+  # to work if you just specify the job ID of the matrix itself).
+  build:
```
jordancarlin, line 82: What about naming this something like "All-jobs-succeeded" or "CI-Successful"? Build is sort of misleading at this point.

PR #868, opened 2025-04-17
Add prettier pre-commit hook

jordancarlin:
Primarily added for formatting json files, but also formats markdown and yaml.

I noticed #866 makes significant formatting changes to the `default.json` config file. It probably makes sense to have an auto formatter that is enforced in CI for JSON files (especially as I expect we will add more preset configs as time goes on) to make changes easier to review going forward, similar to what we already do with `clang-format`. Prettier seems like one of the more popular JSON formatters, so I went with that. By default, it also formats Markdown and YAML files, but if we'd rather, I can disable that and only do the JSON files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,7 @@
-RISCV Sail Model
-================
+# RISCV Sail Model
 
 This repository contains a formal specification of the RISC-V architecture, written in
-[Sail](https://github.com/rems-project/sail).   It has been adopted by the RISC-V Foundation.
+[Sail](https://github.com/rems-project/sail). It has been adopted by the RISC-V Foundation.
 
 The model specifies
 assembly language formats of the instructions, the corresponding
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -11,28 +10,26 @@ A [reading guide](doc/ReadingGuide.md) to the model is provided in the
...
-
-What is Sail?
--------------
+## What is Sail?
 
 [Sail](https://github.com/rems-project/sail) is a language for describing the instruction-set architecture
 (ISA) semantics of processors: the architectural specification of the behaviour of machine instructions. Sail is an
 engineer-friendly language, much like earlier vendor pseudocode, but more precisely defined and with tooling to support a wide range of use-cases.
+
 <p>
```
Timmmm, line 22: Maybe delete these `<p>`s while we're at it.

README.md
```diff
@@ -87,8 +84,8 @@ Information on other options for the simulator is available from
 For booting operating system images, see the information under the
 [os-boot/](os-boot/) subdirectory.
 
-Supported RISC-V ISA features
------------------------------
+## Supported RISC-V ISA features
+
 #### The Sail specification currently captures the following ISA extensions and features:
```
Timmmm, line 89:
```suggestion
### The Sail specification currently captures the following ISA extensions and features:
```

Timmmm (approved):
This is a great improvement IMO and I think it gets all of the styles pretty much bang on. I'm not a fan of `-` for Markdown lists - I think most people use `*` but [you can't configure it](https://github.com/prettier/prettier/issues/4251) and I can live with it.

pmundkur (approved):
LGTM.  This is a useful pre-commit hook!  Quite meta in re-formatting its own config.

PR #870, opened 2025-04-17
Change XLEN and FLEN to be configure-time options

Timmmm:
This means you only need to compile the emulator once in order to support RV32F, RV32D, RV64F and RV64D.

This compiles doesn't work yet due to some issues in Sail:

* https://github.com/rems-project/sail/issues/1245 (just a warning so maybe it doesn't matter)
* https://github.com/rems-project/sail/issues/1246

There are also various TODOs but I thought I'd push this to prevent duplicated effort in case someone else starts working on it.


pmundkur:
This is awesome!  It looks like the upstream issues are fixed.  Are there any other issues preventing this coming out of draft?

test/riscv-tests/CMakeLists.txt
```diff
@@ -3,19 +3,19 @@
 # On success or failure they write to the `tohost` symbol. See this code:
 # https://github.com/riscv/riscv-test-env/blob/4fabfb4e0d3eacc1dc791da70e342e4b68ea7e46/p/riscv_test.h#L200
 
-file(GLOB elfs_rv32d CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/rv32*.elf")
-file(GLOB elfs_rv64d CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/rv64*.elf")
+# file(GLOB elfs_rv32d CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/rv32*.elf")
```
arichardson, line 6: So I guess here we would need to edit the json?

Maybe the easiest way would be to make it a template and use configure_file() to replace `@XLEN_LOG2@`, etc. to generate a config_rv32d.json, config_rv64d.json in the build dir?

model/riscv_flen.sail
```diff
@@ -6,6 +6,10 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
+type flen_bytes : Int = config base.flen_bytes
+
```
jordancarlin, line 10: Is it possible to automatically set this based on if F or D are supported instead of adding flen as a configuration option?

Timmmm: Yeah, or the other way around. Not sure which is better. I agree we shouldn't have both.

Timmmm:
> This is awesome! It looks like the upstream issues are fixed. Are there any other issues preventing this coming out of draft?

Getting closer but I hit another issue: https://github.com/rems-project/sail/issues/1251

Timmmm:
Also I need to:

1. Re-enable/fix the tests.
2. Remove `flen` or `F`/`D` as @jordancarlin suggested.
3. Maybe make the config option be `xlen` instead of `log2_xlen_bytes` which is not very intutive! Since there are only 2 options I think we can just go backwards like: `type xlen_bytes : Int = if xlen == 32 then 4 else 8`. Hopefully that will work.
4. Update the readme.
5. Ideally benchmark the performance hit. I suspect it will be relatively minor but it might not be.

And we'll need to wait for a new Sail compiler release after it's working.

Timmmm:
Ok I got the RV64 first party test passing! Still need to do all the todos I mentioned.

Timmmm:
I did have a crazy idea - what if we just take the xlen setting from the ELF file? :-D

On balance, probably a bad idea. Explicitness is good. Just thought I'd mention it!

jordancarlin:
> I did have a crazy idea - what if we just take the xlen setting from the ELF file? :-D
> 
> On balance, probably a bad idea. Explicitness is good. Just thought I'd mention it!

Something to think about down the road maybe. I think we can technically even find out which extensions an elf was compiled with if we really wanted to. Might be nice to be able to just run an elf without thinking about the config.

For verification purposes though (which I think is one of the primary uses of the Sail model), you want the model to exactly match an implementation no matter what the elf file being passed in is. Maybe there could be two different modes in the future: use the values from a config file if one is given and parse the elf if not.

Timmmm:
This basically works now. All the tests pass. Remaining todos:

1. Benchmark.
2. Come up with a better system for the config files than copy/pasting them. Though tbh if we only have 2/3 files that's maybe the best way anyway.
3. Ideally get rid of the weird `_rv` suffix on `riscv_sim_rv`, but I think that might be best deferred until we merge the RVFI version so there's only a single emulator.
4. Wait for new Sail compiler release.

CMakeLists.txt
```diff
@@ -167,8 +165,8 @@ endif()
 include(CPack)
 
 # Convenience targets.
-add_custom_target(csim DEPENDS riscv_sim_rv32d riscv_sim_rv64d)
-add_custom_target(check DEPENDS generated_model_rv32d generated_model_rv64d)
+add_custom_target(csim DEPENDS riscv_sim_rv)
```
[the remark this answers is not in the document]
arichardson: Could also keep the old names around as scripts that run with a specific config file.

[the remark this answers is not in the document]
arichardson: If we make the config json a template, we could install some common default config files such as config_rv32f.json, config_rv64d.json.
And we'd be able to use them for the tests without jq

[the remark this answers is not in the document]
Timmmm: @jordancarlin I agree, though it is a little awkward to do with how `${arch}` currently works. If it is empty or something like `_rvfi` we'll get weird filenames in places, e.g. search for `-o "${arch}"`.

If the simulator name could end with `rv` that would sneakily avoid the issue! `sim_rv`? :-D Probably not a name I would have chosen given a totally free choice tbf.

@arichardson Yeah... I dunno, it means you can't easily use them as starter configs. 

Perhaps we could have a script that generates a wide variety of config files (like, 20 of them), and add them to the repository. CI can check they are up-to-date. Then it means they're easily visible in the source repo and users don't need whatever scripting thing we use to generate them.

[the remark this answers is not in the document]
jordancarlin: @Timmmm I took another look at this now that runtime RVFI has been merged, and as far as I can tell it shouldn’t be too difficult to remove `${arch}` now. Happy to push an updated version with that change if you want.

That brings us back to the original question of what to actually name the simulator. My vote would still be `riscv_sail_sim`.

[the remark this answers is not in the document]
Timmmm: Yeah go for it if you have time! I just rebased it.

I'm working on making vlen configurable too, but that pretty much just affects `.sail` files so it shouldn't conflict. Also it's a bigger change since the vector code is a bit of a mess, so I'll probably do it in another PR.

Timmmm:
This also needs changes to the types of the callback interface now (`fbits` -> `sbits`). I'll do it at some point.

jordancarlin:
@Timmmm I updated the types for the callbacks to get this building again and then removed `${arch}` from the c simulator. As discussed during the meeting, I've renamed the simulator `sail_riscv_sim`.

I also realized that we need to maintain separate targets for the theorem provers since they take a compile time config file, so I added those back in.

c_emulator/CMakeLists.txt
```diff
@@ -16,56 +16,46 @@ set(EMULATOR_COMMON_SRCS
...
+                            PROPERTIES COMPILE_OPTIONS "${_generated_c_warning_opt_out}")
 
-        add_dependencies(riscv_sim_${arch} generated_model_${arch})
+add_dependencies(sail_riscv_sim generated_sail_riscv_model)
 
-        target_link_libraries(riscv_sim_${arch}
-            PRIVATE softfloat sail_runtime default_config
-        )
+target_link_libraries(sail_riscv_sim
+    PRIVATE softfloat sail_runtime default_config GMP::GMP
```
Timmmm, line 40: I think `GMP::GMP` shouldn't be here? Possibly a rebase mistake. IIRC it was moved to `sail_runtime` to fix Mac builds.

c_emulator/CMakeLists.txt
```diff
@@ -16,56 +16,46 @@ set(EMULATOR_COMMON_SRCS
...
 
-foreach (xlen IN ITEMS 32 64)
-    foreach (flen IN ITEMS 32 64)
-        set(arch "rv${xlen}")
-        if (flen EQUAL 32)
-            string(APPEND arch "f")
-        else()
-            string(APPEND arch "d")
-        endif()
+set(arch "rv")
```
Timmmm, line 19: Can remove this?

test/first_party/CMakeLists.txt
```diff
@@ -88,12 +88,15 @@ set(tests
     "test_pmp_access.c"
 )
 
-foreach (xlen IN ITEMS 32 64)
+foreach (arch IN ITEMS "rv32d" "rv64d")
```
Timmmm, line 91: I think I preferred this as it was, i.e.

```
foreach (xlen IN ITEMS 32 64)
        set(arch "rv${xlen}d")
        if (xlen EQUAL 32)
            set(mabi "ilp32")
        else()
            set(mabi "lp64")
        endif()
        set(default_config "${CMAKE_SOURCE_DIR}/config/${arch}.json")

        foreach (test_source IN LISTS tests)
...
```

Timmmm (commented):
Thanks! LGTM - just a few minor nits.

c_emulator/CMakeLists.txt
jordancarlin: Removed

c_emulator/CMakeLists.txt
jordancarlin: Yes. Rebase issue. Fixed.

test/first_party/CMakeLists.txt
jordancarlin: Changed this earlier when it was causing problems for a different reason and forgot to change it back. Reverted.

jordancarlin:
Other than waiting on the new sail release, I still need to sort out why all of the theorem provers (including the SMT target) are giving the following error:
```
Error:
riscv_vmem_utils.sail:47.63-106:
47 |function sys_misaligned_allowed_within_exp() -> range(0, 11) = config memory.misaligned.allowed_within_exp
   |                                                               ^-----------------------------------------^
   | SMT solver returned unexpected status 1
```

@Alasdair @Timmmm any ideas?

Timmmm:
Does it give any output after that line? I think SMT error reporting isn't super robust (mainly because the Z3 SMT2 interface is kind of bonkers), so I expect you may have to hack the Sail compiler a bit to find out the actual error (search for that error message; it's in `src/lib/constraint.ml`.

jordancarlin:
There wasn't any other output initially, but after adding a bunch of other debugging statements I found the following:

The SMT solver is failing with the following error message `(error "line 4 column 29: unknown constant zxlen")`, and the input to the SMT solver that generates that error is

```
(push)
(declare-const v0 Int)

(assert (and (= 0 v0) (or (= zxlen 32) (= zxlen 64))))
(check-sat)
(pop)
```

Seems like this might be an issue with how abstract types are being handled because it looks like it is indeed not declaring `zxlen`. @Alasdair

jordancarlin:
Seems I spoke too soon. It was working, but a rebase seems to have hit yet another bug in the type checker.

jordancarlin:
@ineol @tobiasgrosser @Alasdair looks like Sail itself is working now, but the Lean code that it outputs fails when Lean tries to compile it:
```
✖ [5/107] Building LeanRV64D.Defs
trace: .> LEAN_PATH=/home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean /home/runner/.elan/toolchains/leanprover--lean4-nightly---nightly-2025-05-26/bin/lean --tstack=400000 -Dweak.linter.style.nameCheck=false /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/LeanRV64D/Defs.lean -R /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D -o /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D/Defs.olean -i /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D/Defs.ilean -c /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D/Defs.c --json
error: /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/LeanRV64D/Defs.lean:50:32: unknown identifier 'ext_d_supported'
error: build failed
```

ineol:
I found the root of the issue, I have [a PR](https://github.com/rems-project/sail/pull/1332) ready for Sail in a few minutes.

Another issue that revealed itself when that error was fixed is the termination measure for `hartSupport`. The following change in `riscv_termination.sail` fixes it:

```
function hartSupports_measure(ext : extension) -> int =
  match ext {
    Ext_C => 1,
    _ => 0,
  }
```
into 
```
function hartSupports_measure(ext : extension) -> int =
  match ext {
    Ext_C => 2,
    Ext_D => 1,
    _ => 0,
  }
```

jordancarlin:
> I found the root of the issue, I have [a PR](https://github.com/rems-project/sail/pull/1332) ready for Sail in a few minutes.
> 
> Another issue that revealed itself when that error was fixed is the termination measure for `hartSupport`. The following change in `riscv_termination.sail` fixes it:

Thanks the quick fix. Termination measure updated.

config/rv64d.json
```diff
@@ -0,0 +1,244 @@
+{
```
arichardson, line 1: These two files look extremely similar, can we generate them at CMake time?

c_emulator/riscv_callbacks.cpp
```diff
@@ -21,12 +21,12 @@ void print_lbits_hex(lbits val, int length = 0)
 
 // Implementations of default callbacks for trace printing and RVFI.
 // The model assumes that these functions do not change the state of the model.
-unit mem_write_callback(const char *type, uint64_t paddr, uint64_t width,
+unit mem_write_callback(const char *type, sbits paddr, uint64_t width,
```
jrtc27, line 24: Hm, it's a bit sad that Sail doesn't apply the same optimisation despite the explicit constraint on xlen being 32 or 64.

Timmmm: It's because it needs to know the bit width at runtime now. I think there are some other things you could do but nothing I can think of that's easy and elegant.

It's not too bad though - it knows not to use `lbits` (GMP) and that's the real performance killers.

Alasdair: We might need to make sure enough primitives have specialised implementations for that type. I think the danger is it will upcast to a GMP-based bitvector to call primitives that don't have optimised implementations for more specific types.

One thing you can do (but not in this PR) is split the configuration in two so you can partially apply it at build time when you really want a specialised implementation for performance.

Timmmm:
I'd be wary of just wholesale renaming of `xlen` to `mxlen`. As it stands the model obviously doesn't support `u/sxlen != mxlen`. If we just rename things to `mxlen` without actually going through the model and properly supporting this I think it will give the wrong impression that we've actually done all of that implementation work. It will also make it harder to find all the places that reference `xlen` when we *do* do that implementation work.

I think we should defer it until we properly support mutable u/sxlen.

radimkrcmar:
Right, some `xlen` must stay `xlen`, and others must be changed to `mxlen`, `sxlen`, or `uxlen`.
I was thinking about a transition period where we start using the proper `*xlens` in places where the specification says *XLEN.  We will eventually have to revisit all occurrences of `xlen`, but we could be preparing for the work even now.

jordancarlin:
Either way, I think that change should happen as part of a separate PR. This one changes enough already without also modifying every file to rename `xlen`.

Timmmm:
Rebased. Hopefully I didn't screw it up (wasn't the most pleasant rebase).

As far as I can tell this has the following issues outstanding:

* [ ] Need to benchmark
* [ ] Come up with a better system for the config files than copy/pasting them. Maybe. Tbh I'm ok with copy/paste to start with.

arichardson:
> Rebased. Hopefully I didn't screw it up (wasn't the most pleasant rebase).
> 
> As far as I can tell this has the following issues outstanding:
> 
> * [ ]  Need to benchmark
> * [ ]  Come up with a better system for the config files than copy/pasting them. Maybe. Tbh I'm ok with copy/paste to start with.

nice! I think the config deduplication should be simple: change `config/default.json` with a few entries such as `"xlen": @CONFIG_XLEN@,`, `"D": { "supported": @CONFIG_D_SUPPORTED@ }` to generate rv32f.json, rv32d.json, rv64f.json, rv64d.json with configure_file()? And for the default builtin sail we could use rv64d.json?

Timmmm:
@Alasdair I might be being dumb but this is giving the following error when building the SMT target:

```
Error:
riscv_device_tree.sail:190.25-56:
190 |  let clock_freq : int = config platform.clock_frequency;
    |                         ^-----------------------------^
    | Runtime configuration is not supported when generating code for target 'smt'
    | An explicit configuration should be provided with the --config option.
```

Reasonable, except it *does* use the `--config` option:

```
cd /home/me/workspace/sail-riscv/model && /home/me/workspace/sail-riscv/build/sail/bin/sail --smt -o /home/me/workspace/sail-riscv/build/model/model --require-version 0.19.1 --config /home/me/workspace/sail-riscv/config/rv32d.json prelude.sail ... lots of sail files ... main.sail
```

jordancarlin:
> I might be being dumb but this is giving the following error when building the SMT target:
> 
> ```
> Error:
> riscv_device_tree.sail:190.25-56:
> 190 |  let clock_freq : int = config platform.clock_frequency;
>     |                         ^-----------------------------^
>     | Runtime configuration is not supported when generating code for target 'smt'
>     | An explicit configuration should be provided with the --config option.
> ```
> 
> Reasonable, except it _does_ use the `--config` option:
> 
> ```
> cd /home/me/workspace/sail-riscv/model && /home/me/workspace/sail-riscv/build/sail/bin/sail --smt -o /home/me/workspace/sail-riscv/build/model/model --require-version 0.19.1 --config /home/me/workspace/sail-riscv/config/rv32d.json prelude.sail ... lots of sail files ... main.sail
> ```

It happens when one of the config options is missing from the json file. Looks like part of one of my patches was also lost during the rebase. Pushed an updated version that should resolve these issues.

>  Come up with a better system for the config files than copy/pasting them. Maybe. Tbh I'm ok with copy/paste to start with.

Not sure that there is a better option than copy/pasting because there are various extensions that should be different between RV32/RV64. For example, Zcf should be on in RV32 but not RV64. Similar for the Sv* extensions.

arichardson:
> > I might be being dumb but this is giving the following error when building the SMT target:
> > ```
> > Error:
> > riscv_device_tree.sail:190.25-56:
> > 190 |  let clock_freq : int = config platform.clock_frequency;
> >     |                         ^-----------------------------^
> >     | Runtime configuration is not supported when generating code for target 'smt'
> >     | An explicit configuration should be provided with the --config option.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Reasonable, except it _does_ use the `--config` option:
> > ```
> > cd /home/me/workspace/sail-riscv/model && /home/me/workspace/sail-riscv/build/sail/bin/sail --smt -o /home/me/workspace/sail-riscv/build/model/model --require-version 0.19.1 --config /home/me/workspace/sail-riscv/config/rv32d.json prelude.sail ... lots of sail files ... main.sail
> > ```
> 
> It happens when one of the config options is missing from the json file. Looks like part of one of my patches was also lost during the rebase. Pushed an updated version that should resolve these issues.
> 
> > Come up with a better system for the config files than copy/pasting them. Maybe. Tbh I'm ok with copy/paste to start with.
> 
> Not sure that there is a better option than copy/pasting because there are various extensions that should be different between RV32/RV64. For example, Zcf should be on in RV32 but not RV64. Similar for the Sv* extensions.

These should be quite easy to handle with a simple CMake substition:

`"Zcf" : { supported: @CONFIG_IS_RV32@ }`
`"Sv32" : { supported: @CONFIG_IS_RV32@ }` (or True if it works for RV64).
`"Sv48" : { supported: @CONFIG_IS_RV64@ }`
and then just call `configure_file()` with those variables set appropriately.

Timmmm:
I'm still slightly partial to just copy & pasting. It's only two files and I think it's unlikely to grow... and I think there's a lot of value in having the file in the source tree that people can easily copy & paste and modify without having to dig it out from the build directory.

Would you be ok with merging this as-is and using `configure_file()` later if it becomes a problem (e.g. if they get out of sync)?

Timmmm:
I do think we could delete `default.json` and just use the rv64d one as the default though. There's no need for three files...

arichardson:
> I do think we could delete `default.json` and just use the rv64d one as the default though. There's no need for three files...

Yes if there are only two files I'd be happy with this.

Timmmm:
Ok I think this is ready to go. I *finally* got around to benchmarking it. First 3 million instructions of booting Linux (this is not very many - it actually doesn't get through OpenSBI). With a recent `sail2` commit:

* `master`: 53 kIPS
* This PR: 48 kIPS

Considering with the 0.19.1 compiler it's like 9 kIPS and nobody noticed, I think that's fine. It would be nice to work back up to ~500 kIPS which is what Sail was capable of a while ago, but we'll have to do some work for that. I have some ideas.

pmundkur (approved):
This is awesome!  We should merge this soon to avoid more painful rebasing.

arichardson (approved):
This is great!

jordancarlin:
Rebased again. Would definitely be nice to get merged soon since just about any change to the CMake or config files forces this be rebased.

Timmmm:
Didn't notice your rebase so I also rebased, and we both made slightly different rebase mistakes which I have fixed. :-D

I will merge this today if nobody objects!

Issue #851, opened 2025-04-14
RV32 memory addresses in log are not very readable


Timmmm:
Since we fixed RV32 so that physical addresses are actually 34 bits, they are printed in the log like this:

```
mem[X,0b0010000000000000000010000101010000] -> 0x0000
```

We probably need to add some function that can print long non-multiple-of-4 bit vectors a bit nicer. Some options:

1. Zero-extend to a multiple of 4 bits. Simple. Unsurprising. But you lose the length information which is a shame.
2. `0b00 @ 0x800....` but I think that's just confusing and still fairly verbose.
3. SV-style `34'h800...`. I hate it.
4. Something like `34'0x800...`. I feel like this is probably the best option if we *must* preserve length information.

I vote option 1.

PR #871, opened 2025-04-18
make 34 bit physaddr log readable

KotorinMinami:
As noted in #851, this PR adds a helper function to improve string readability. To address the loss of length information, I propose adding a comment or updating the documentation. If this approach is approved, I can include it in this PR.

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -24,6 +24,12 @@ val elf_entry = pure {
   c: "elf_entry"
 } : unit -> int
 
+function readable_addr_bits (addr : physaddrbits) -> string = {
+  if physaddrbits_len == 34
+  then BitStr(0b00 @ addr)
+  else BitStr(addr)
+}
+
 // Cache block size is 2^cache_block_size_exp. Max is `max_mem_access` (4096)
 // because this model performs `cbo.zero` with a single write, and the behaviour
 // with cache blocks larger than a page is not clearly defined.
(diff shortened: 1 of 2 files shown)


model/riscv_platform.sail
```diff
@@ -24,6 +24,12 @@ val elf_entry = pure {
   c: "elf_entry"
 } : unit -> int
 
+function readable_addr_bits (addr : physaddrbits) -> string = {
```
pmundkur, line 27:
```suggestion
function PhysaddrStr(addr : physaddr) -> string = {
```
Perhaps change the signature to ensure it can't be accidentally called on a bitvector, and the name to match the existing `RegStr` and `FRegStr`.   `PAddrStr` could also work as a shorter name.

jordancarlin:
How does this interact with #494? That moves most of the printing from Sail to C.

KotorinMinami:
> How does this interact with #494? That moves most of the printing from Sail to C.

I believe #494 may have addressed this issue, as C functions typically zero-extend arguments by default. However, there's been no progress in the last three weeks. This is a minor adjustment to improve output readability before that PR is merged.

jordancarlin:
> > How does this interact with #494? That moves most of the printing from Sail to C.
> 
> I believe #494 may have addressed this issue, as C functions typically zero-extend arguments by default. However, there's been no progress in the last three weeks. This is a minor adjustment to improve output readability before that PR is merged.

#494 is fairly high on my priority list, but I’m fine merging this in the interim to clean up the output.

model/riscv_platform.sail
```diff
@@ -24,6 +24,12 @@ val elf_entry = pure {
   c: "elf_entry"
 } : unit -> int
 
+function PhysaddrStr (Physaddr(addr) : physaddr) -> string = {
+  if physaddrbits_len == 34
+  then BitStr(0b00 @ addr)
+  else BitStr(addr)
+}
```
Timmmm, line 31: I think it would be nicer to use a generic function. Also `BitsStr()` is some old alias; the native function is `bits_str()` so I think we should use that style. How about this?

```
// Print a bit vector in hex. If it is not a multiple of 4
// bits in length zero extend it so it is (bits_str() prints
// bit vectors that aren't a multiple of 4 bits in binary).
```

The name combines the existing `bits_str` and `hex_str`.

I have definitely wanted this function for other things while debugging!

Finally it should go in `prelude.sail` rather than here.

Timmmm (approved):
LGTM, thanks!

pmundkur (approved):

PR #872, opened 2025-04-21
Use result type instead of PTW_Result

trdthg:
part of #433 

I'm not quite sure about the functionality of `ext_ptw` here

### DIFF
--- a/model/riscv_vmem.sail
+++ b/model/riscv_vmem.sail
@@ -39,10 +39,7 @@ struct PTW_Output('v : Int), is_sv_mode('v) = {
 }
 
 // PRIVATE
-union PTW_Result('v : Int), is_sv_mode('v) = {
-  PTW_Success : (PTW_Output('v), ext_ptw),
-  PTW_Failure : (PTW_Error, ext_ptw),
-}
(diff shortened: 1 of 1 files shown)


Timmmm:
LGTM. `ext_ptw` is used for implementing page table extensions for CHERI - [the current implementation is here](https://github.com/CHERI-Alliance/sail-riscv/blob/4f1dae4d0d38348a83dd8f3d670264330c767a22/model/cheri/cheri_riscv_types.sail#L38).

In theory it can be used for other out-of-tree extensions but as far as I know it's only used for CHERI, and it's kind of closely coupled to CHERI even though it seems generic. I had to change some of the extension types when updating the CHERI PTE code to the latest spec... but fortunately not this bit so this LGTM.

One future improvement I noticed - it doesn't really make sense to use `result((ok_type, T), (err_type, T))` - we should just use `(result(ok_type, err_type), T)` instead. We can fix that in future though.

At some point we may end up remove these extension types anyway since we're planning to eventually fully integrate CHERI into this model (or at least maintain it as a branch).

Timmmm (approved):

pmundkur (approved):

PR #873, opened 2025-04-21
Fix precommit failure

jordancarlin:
#864 was merged after #868, but without the changes from #868 merged into it, so it passed the CI from that branch but failed CI once merged into master. This fixes the precommit errors with it.


Timmmm (approved):
Thanks. Merge queue!

PR #876, opened 2025-04-22
CI: test ./build_emulators.sh

Arielfoever:
Add a build by build_emulators.sh and I believe there's no need to test.


Arielfoever:
See https://github.com/riscv/sail-riscv/issues/745

and somehow we need a mirror of gmp.

Arielfoever:
Go and see https://github.com/Arielfoever/sail-riscv/actions/runs/14591059685/job/40926223320 if it works.

.github/workflows/scripts.yml
```diff
@@ -0,0 +1,29 @@
+name: CI with script
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install packages
+        run: sudo apt install -y --no-install-recommends pkg-config libgmp-dev curl ninja-build jq
```
jordancarlin, line 10:
```suggestion
        run: sudo apt install -y --no-install-recommends pkg-config curl
```
This should download and build gmp, so no need to install it. It also uses Make by default, so ninja shouldn't be needed either.

.github/workflows/scripts.yml
```diff
@@ -0,0 +1,29 @@
...
+          curl --location https://github.com/rems-project/sail/releases/download/0.19-linux-binary/sail.tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
+
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
+      - name: Build simulator
+        run: ./build_simulators.sh
+        # Build with ./build_emulators.sh
+      
+      - name: Upload simulator
```
jordancarlin, line 24: Why are you uploading the built simulator?

jordancarlin (changes requested):
Would be great to get this in, but I don't think this CI job will work until the issue with downloading GMP is addressed (https://github.com/riscv/sail-riscv/issues/745#issuecomment-2764824346).

.github/workflows/scripts.yml
Arielfoever: Remove it if you think upload is not needed(

.github/workflows/scripts.yml
Arielfoever: Resolved with resolve.

.github/workflows/scripts.yml
Timmmm: I'd say instead we should just run `ctest` here.

jordancarlin:
That seems reasonable. If we want to make sure that the build works with both static and dynamic gmp, we could also have the build with the shell script happen on a weekly schedule instead of for every PR. Most code changes are unlikely to break it, so that is probably more than sufficient to make sure the script continues working.

Arielfoever:
> Would be great to get this in, but I don't think this CI job will work until the issue with downloading GMP is addressed ([#745 (comment)](https://github.com/riscv/sail-riscv/issues/745#issuecomment-2764824346)).

Fix via https://github.com/riscv/sail-riscv/pull/882

Arielfoever:
@Timmmm What is the problem? This is nothing to do with #882. This is CI not mirror.

jordancarlin:
I think it was closed automatically by GitHub when #882 was merged because you put "fix #876" in the PR description (which GitHub picks up as a keyword to link  PRs/issues). Reopened.

.github/workflows/scripts.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          sudo mkdir -p /usr/local
+          curl --location https://github.com/rems-project/sail/releases/download/0.19-linux-binary/sail.tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
+
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
+      - name: Build simulator
+        run: ./build_simulators.sh
+        # Build with ./build_emulators.sh
```
Timmmm, line 22:
```suggestion
```

.github/workflows/scripts.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Test simulator
+        run: ctest --test-dir build --output-junit tests.xml --output-on-failure
+
+      - name: Upload test results
```
Timmmm, line 27: I think we don't need to upload the test results/event payload since the other workflow will already do this.

.github/workflows/scripts.yml
```diff
@@ -0,0 +1,40 @@
+name: CI with script
```
Timmmm, line 1:
```suggestion
name: Test build_emulators.sh script
```

Timmmm (changes requested):
LGTM. I think it probably makes sense to make it be another job in `compile.yml` instead of a new file.

arichardson (changes requested):
This is duplicating quite a lot - can we just add another matrix combination to the main CI build?

I think something like this could possibly work?

```
    strategy:
      matrix:
        os: [macos-latest, ubuntu-latest]
        include:
          - os: ubuntu-latest
            variant: build-emulators
```

Timmmm:
I think a variant might get complicated. We could potentially use [composite actions](https://www.jameskerr.blog/posts/sharing-steps-in-github-action-workflows/) to add it as a job without duplicating everything though. I've never used them before so I'm not really sure...

jordancarlin:
Rather than adding a whole new job, what about updating the matrix in #952 so that one of the CMake versions uses the script for installation? That way, we are getting the benefit of testing the script without adding an ever-growing list of jobs to the CI.

I think it should be possible with something like this in place of the existing "Build and test simulators" step:
```
      - name: Build simulators
        if: ${{ ! matrix.use_script }}
        run: |
          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DFIRST_PARTY_TESTS=TRUE
          ninja -C build all riscv_sim_rv32f riscv_sim_rv64f generated_smt_rv64f generated_docs_rv64d
      - name: Build simulators (with build_simulators.sh)
        if: ${{ matrix.use_script }}
        run: ./build_simulators.sh
      - name: Run Tests
        run: ctest --test-dir build --output-junit tests.xml --output-on-failure
```

Arielfoever:
I suggest that this pr should add to tgmm to discuss about the version we want to test.

for instance,

- different cmake?
- different system like macos/windows/ubuntu2204/ubuntu2404
- remove old CI?

I mentioned that there are a lot of diversions.

PR #880, opened 2025-04-22
Expose platform ids to config.

pmundkur:
Pretty basic configuration.

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -427,6 +427,11 @@ function mmio_write forall 'n, 0 <'n <= max_mem_access . (paddr : physaddr, widt
 /* Platform initialization and ticking. */
 
 function init_platform() -> unit = {
+  mvendorid = to_bits(32,   (config platform.vendorid : int));
+  marchid   = to_bits(xlen, (config platform.archid : int));
+  mimpid    = to_bits(xlen, (config platform.impid : int));
+  mhartid   = to_bits(xlen, (config platform.hartid : int));
+
   htif_tohost = zeros();
   htif_done   = false;
   htif_exit_code = zeros();
--- a/config/default.json
+++ b/config/default.json
@@ -21,6 +21,10 @@
     }
   },
   "platform": {
+    "vendorid": 0,
+    "archid": 0,
+    "impid": 0,
+    "hartid": 0,
     "reset_vector": 4096,
     "cache_block_size_exp": 6,
     "ram": {


model/riscv_platform.sail
```diff
@@ -426,7 +426,15 @@ function mmio_write forall 'n, 0 <'n <= max_mem_access . (paddr : physaddr, widt
 
 /* Platform initialization and ticking. */
 
+function plat_vendorid() -> bits(32) = to_bits(32, config platform.vendorid : int)
+function plat_archid() -> xlenbits = to_bits(xlen, config platform.archid : int)
+function plat_impid() -> xlenbits = to_bits(xlen, config platform.impid : int)
+
 function init_platform() -> unit = {
+  mvendorid = plat_vendorid();
```
Timmmm, line 434: Maybe just use them directly?

```suggestion
  mvendorid = config platform.vendorid
  marchid = config platform.archid
  mimpid = config platform.impid
```

@Alasdair I might have mentioned this already but it would be nice if the config system had some way of zero-extending bit vectors (either from integers or from "0x123" strings. The downside of the above code is that you have to add/remove a load of extra zeros when you change XLEN for `archid` and `impid`.

pmundkur: Right, wouldn't the advantage of the `plat_*()` calls be that they could use whatever `xlen` had been initialized to once we have #870?

pmundkur: So directly assigning
```
mvendorid = config platform.vendorid;
```
results in
```
Pattern match failure in init_platform
```
while
```
mvendorid = plat_vendorid();
```
works.  I'll leave the `plat_*` functions in for now.
```

Comment (Timmmm) [reply model/riscv_platform.sail:435]: You need to change the config to be `"0x00000000"` instead of `0`. Or leave it as 0 but then use `to_bits(config ...)`.

The config error messages are not especially clear yet!

Comment (pmundkur) [reply model/riscv_platform.sail]: The second option works, thanks, that's nicer than what I had.  I couldn't get the first one to work.

Comment (Timmmm) [reply model/riscv_platform.sail]: Actually you have to use

```
"impid": {
   "len": 64,
   "value": "0xF..."
 }
 ```
 
 See `writable_hpm_counters` for an example. This kind of sucks because you have to change all of the `len`s when you change XLEN, but I just discovered that after #870 I *think* you can do this:
 
```
 "impid": {
   "len": "xlen",
   "value": "34" // It can also be hex or binary and it gets zero extended (and truncated).
 }
```

I'll try it, but I think let's go with what you've got for now at least. Just having an integer is a bit nicer for the config file anyway.

Timmmm (approved):

jordancarlin (approved):
Great to start getting these simpler configuration parameters into the model!

PR #881, opened 2025-04-23
Change default instructions_per_tick to 2

### LINKED ISSUE #193: mcycle value does not increment in sail
I am writing test that checks mcycle value increments. For Spike, the value of mcycle increases, whereas for Sail, the value of mcycle remains constant. Here is the test result screenshot and the assembly code for the test.

   LI(x1,0)
    	LI(a5,-1)		// ALL 1's
	LI(x5, 1)
	csrw mcycle, a5
	nop
	csrr a4, mcycle		
	nop				// nop in case of trap
	RVTEST_SIGUPD(x13, a4)		// Putting updated value in signature file
	
	add t0, x5, x0
	csrr a4, mcycle		
	nop				// nop in case of trap
	RVTEST_SIGUPD(x13, a4)		// Putting updated value in signature file	
	
	addi t0, x5, 3
	csrr a4, mcycle		
	nop				// nop in case of trap
	RVTEST_SIGUPD(x13, a4)		// Putting updated value in signature file	
	
	csrr a4, mcycle		
	nop				// nop in case of trap
	RVTEST_SIGUPD(x13, a4)		// Putting updated value in signature file

image

jordancarlin:
Fixes #193. This parameter controls how many instructions need to complete for the cycle CSR to increment. While this is implementation defined (and likely not constant even within a given implementation), a default of 2 seems a lot more natural than 100.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -35,7 +35,7 @@
       "base": 33554432,
       "size": 786432
     },
-    "instructions_per_tick": 100,
+    "instructions_per_tick": 2,
     "wfi_is_nop": true
   },
   "extensions": {


Timmmm:
The only downside I can see to this is that people might get confused and write tests where e.g. they expect it to increment every instruction. You might say that's unlikely but that's exactly what happened in #193.

I guess ideally we'd have it increment by some pseudo-random number each time (which could be 0). However that's complicated... 100 does seem very unrealistically high though. Why don't we set it to 2? Dual issue is pretty common in real life too.

jordancarlin:
Seems reasonable. Updated to 2 instructions.

Timmmm (approved):

PR #882, opened 2025-04-23
CMakeLists: Add mirrors of gmp.

Arielfoever:
See https://www.gnu.org/server/mirror.html

You can see the expected result at https://github.com/Arielfoever/sail-riscv/actions/runs/14609337684/job/40984351177

This will fix https://github.com/riscv/sail-riscv/pull/876#pullrequestreview-2783448510

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ if (DOWNLOAD_GMP)
     include(ExternalProject)
     ExternalProject_Add(
         gmp
-        URL https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz
+        URL https://ftp.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz https://ftpmirror.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz
         URL_HASH SHA256=a3c2b80201b89e68616f4ad30bc66aee4927c3ce50e33929ca819d5c43538898
         INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/gmp
         CONFIGURE_COMMAND "<SOURCE_DIR>/configure" "--prefix=<INSTALL_DIR>" --disable-shared --enable-static


CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ if (DOWNLOAD_GMP)
     include(ExternalProject)
     ExternalProject_Add(
         gmp
-        URL https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz
+        URL https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz https://ftp.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz http://ftp.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz http://ftpmirror.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz
```
Timmmm, line 80: Should we put ftpmirror first so CI doesn't have to fail downloading from gmplib.org? It's also the one gnu recommends.  Also I would use https.

Timmmm: Right, I meant it takes some time to fail though so it would make sense to use the best one first.

Any reason not to put the `https://ftpmirror.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz` first? The GNU site says:

> First, for users/downloaders: the address http://ftpmirror.gnu.org/ multiplexes between the mirrors, trying to choose one that is nearby and up to date. E.g., http://ftpmirror.gnu.org/emacs/ goes to a mirror's directory of [GNU Emacs](https://www.gnu.org/software/emacs/). **We recommend using this generic ftpmirror.gnu.org address wherever possible** in links, documentation, etc., to reduce load on the main GNU server.

So I think we should probably use it first. We've got the hash there so I don't think there are any security concerns.

Arielfoever: I think determinism is more important. The ftp mirror sometimes direct you to a bad mirror and gen some trouble (on my computer) so I put it at the second.

Timmmm: Ah ok that seems reasonable.

Timmmm (approved):

jordancarlin (approved):

PR #883, opened 2025-04-25
Refactor try_step() some more

Timmmm:
* Simplify the `stepped` code. It is equivalent to whether or not the model ends in an active state, so we can just return that.
* Simplify handling of `minstret_increment`.
* Change `step_no` to `nat`.
* Remove unused `exit_wait` parameter from `run_hart_active()`.
* Change `instbits` to 32-bit instead of xlenbits.

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -446,9 +446,9 @@ function tick_platform() -> unit = {
 
 /* Platform-specific handling of instruction faults */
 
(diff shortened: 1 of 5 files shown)


pmundkur (approved):
This is nice refactor, with clearer comments.  Thanks!

jordancarlin (commented):
LGTM

jordancarlin (approved):

arichardson (approved):

PR #888, opened 2025-04-25
Fix the '--help' option.

pmundkur:
It was not handled correctly (unlike '-h'), resulting in a confusing `No elf file provided.` message before the help text.


arichardson (approved):
Looks like this was accidentally introduced in https://github.com/riscv/sail-riscv/commit/17cfdec13b99e61da9fdf8c32d21b33e008bd82d

Could you update the commit message to add that it fixes a regression from 17cfdec13b99e61da9fdf8c32d21b33e008bd82d?

jordancarlin (approved):

PR #890, opened 2025-04-27
CI: change CI condition.

Arielfoever:
Change some CI condition to avoid wrong dispatch.

Separate build and commit checks.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -4,7 +4,7 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     steps:
       - name: Install packages
         run: sudo apt install -y --no-install-recommends pkg-config libgmp-dev curl ninja-build jq
@@ -16,9 +16,6 @@ jobs:
 
       - name: Check out repository code
         uses: actions/checkout@v4
-
-      - name: Ensure pre-commit checks pass
-        run: python3 -m pip install pre-commit && pre-commit run --all-files --show-diff-on-failure --color=always
 
       - name: Build and test simulators
         run: |
@@ -33,7 +30,7 @@ jobs:
           ctest --test-dir build --output-junit tests.xml --output-on-failure
 
       - name: Upload test results
-        if: always()
+        if: ${{ ! cancelled() }}
         uses: actions/upload-artifact@v4
         with:
           name: tests.xml
--- a/.github/workflows/precheck.yml
+++ b/.github/workflows/precheck.yml
@@ -1,0 +1,16 @@
+name: Pre-commit Check
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install packages
+        run: python3 -m pip install pre-commit
+
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
+      - name: Ensure pre-commit checks pass
+        run: pre-commit run --all-files --show-diff-on-failure --color=always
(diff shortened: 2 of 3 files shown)


.github/workflows/precheck.yml
```diff
@@ -0,0 +1,16 @@
+name: Pre-commit Check
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  build:
```
jordancarlin, line 6: It doesn’t make sense for this job to be named build. Maybe pre-commit.

.github/workflows/precheck.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install packages
+        run: python3 -m pip install pre-commit
+
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
+      - name: Ensure pre-commit checks pass
+        run: python3 -m pip install pre-commit && pre-commit run --all-files --show-diff-on-failure --color=always
```
jordancarlin, line 16: You installed pre-commit in a previous step already. No need to do it twice.

.github/workflows/test-results.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   publish-test-results:
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion != 'skipped'
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
jordancarlin, line 12: Won’t this prevent a comment indicating failure from being created?

jordancarlin (changes requested):
I like the idea of splitting pre-commit checks into a separate workflow. Just a few comments to address first.

.github/workflows/test-results.yml
Arielfoever: At least when you cancel the CI build workflow, it won't generate results workflow.

.github/workflows/compile.yml
```diff
@@ -17,9 +17,6 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v4
 
-      - name: Ensure pre-commit checks pass
```
arichardson, line 20: This only takes a few seconds, do we really need to split it?

Also how about just adding another entry under jobs instead of a completely new file?

jordancarlin: I think splitting it into a separate job is advantageous for things like #867. Agreed that a separate job in the same file would work well though.

Arielfoever: I really hate precommit check stops my hope of seeing the result of build. This make me angry.

jordancarlin: We could also think about enabling pre-commit using the pre-commit ci app if we're changing this anyway: https://pre-commit.ci.

arichardson: I'd suggest running `pre-commit install` locally, but I can see why you'd want to see the build result. However I don't think this should be a seaprate build file, if anything we want a matrix build and a condition.

It's unfortunate the githubs `continue-on-error` doesn't quite seem to do the right thing, I believe it will mark the whole job as a success rather than just continuing and marking it as failure.

I'm not sure I'd want the application that automatically applies the fixes since then we have to teach everyone how to pull the latest changes.

Arielfoever: They can run it locally but put it in a seprate file makes it easy to disable it for fork users

Timmmm: I still think we don't want this in a separate file. Just install pre-commit locally. You can avoid the mess of `pip` now too:

```
curl -LsSf https://astral.sh/uv/install.sh | sh
uv tool install pre-commit
cd sail-riscv
pre-commit [--hook-type pre-push]
```

arichardson:
Is this still relevant or can we close it?

jordancarlin:
Closing this since there seemed to be agreement that this is not the direction we want to take CI.

PR #891, opened 2025-04-27
Fix termination measure for `currentlyEnabled`

ineol:
This fixes the Lean build, at least.


jordancarlin (approved):
Would be really great to somehow automate this in the future, but for now, LGTM.

PR #892, opened 2025-04-27
Disable htif when tohost symbol not found

trdthg:
If htif is disabled

- cpp doesn't enforce check on the elf `tohost` symbol
- sail disallow reading and writing to tohost

If the elf does not meet the requirements, sail can at least try to run it instead of exit directly

### DIFF
--- a/c_emulator/riscv_platform.h
+++ b/c_emulator/riscv_platform.h
@@ -16,6 +16,7 @@ unit cancel_reservation(unit);
 unit plat_term_write(mach_bits);
 
 mach_bits plat_htif_tohost(unit);
+bool plat_enable_htif();
 
 bool sys_enable_experimental_extensions(unit);
 
--- a/c_emulator/riscv_platform_impl.cpp
+++ b/c_emulator/riscv_platform_impl.cpp
@@ -18,6 +18,7 @@ uint64_t rv_16_random_bits(void)
 }
 
 uint64_t rv_htif_tohost = UINT64_C(0x80001000);
+bool rv_enable_htif = true;
 
 bool rv_enable_experimental_extensions = false;
 
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -352,10 +352,11 @@ uint64_t load_sail(char *f, bool main_file)
   fprintf(stdout, "ELF Entry @ 0x%" PRIx64 "\n", entry);
   /* locate htif ports */
   if (lookup_sym(f, "tohost", &rv_htif_tohost) < 0) {
-    fprintf(stderr, "Unable to locate htif tohost port.\n");
-    exit(1);
+    fprintf(stderr, "Unable to locate tohost symbol; disabling HTIF.\n");
+    rv_enable_htif = false;
+  } else {
+    fprintf(stdout, "HTIF located at 0x%0" PRIx64 "\n", rv_htif_tohost);
   }
-  fprintf(stderr, "tohost located at 0x%0" PRIx64 "\n", rv_htif_tohost);
   /* locate test-signature locations if any */
   if (!lookup_sym(f, "begin_signature", &begin_sig)) {
     fprintf(stdout, "begin_signature: 0x%0" PRIx64 "\n", begin_sig);
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -53,6 +53,10 @@ function plat_rom_size() -> physaddrbits = to_bits(physaddrbits_len, config plat
 function plat_clint_base() -> physaddrbits = to_bits(physaddrbits_len, config platform.clint.base : int)
 function plat_clint_size() -> physaddrbits = to_bits(physaddrbits_len, config platform.clint.size : int)
 
+// Whether HTIF (Host Target InterFace) is enabled. This is used for
+// console output and signalling the end of tests.
+val plat_enable_htif = pure "plat_enable_htif" : unit -> bool
+
 /* Location of HTIF ports */
 val plat_htif_tohost = pure {c: "plat_htif_tohost", lem: "plat_htif_tohost"} : unit -> physaddrbits
 function plat_htif_tohost () = to_bits(physaddrbits_len, elf_tohost ())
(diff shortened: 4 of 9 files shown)


jrtc27:
Do we really need the flag, or would it be better to just make the missing section/symbol(s) non-fatal, as is the case for the signature?

c_emulator/riscv_sim.cpp
```diff
@@ -480,6 +489,9 @@ void reinit_sail(uint64_t elf_entry)
 
 void write_signature(const char *file)
 {
+  if (!sys_enable_htif()) {
```
jrtc27, line 492: This is unrelated (other than that you've made load_sail exit early and not parse the signature location so have broken that combination)

arichardson:
> Do we really need the flag, or would it be better to just make the missing section/symbol(s) non-fatal, as is the case for the signature?

I agree it would be nicer for users if we just don't enable htif if the symbol is missing instead of exiting with an error.

c_emulator/riscv_sim.cpp
```diff
@@ -353,9 +359,11 @@ uint64_t load_sail(char *f, bool main_file)
   /* locate htif ports */
   if (lookup_sym(f, "tohost", &rv_htif_tohost) < 0) {
     fprintf(stderr, "Unable to locate htif tohost port.\n");
-    exit(1);
+    if (plat_enable_htif()) {
+      exit(1);
+    }
   }
-  fprintf(stderr, "tohost located at 0x%0" PRIx64 "\n", rv_htif_tohost);
```
jrtc27, line 358: ?

c_emulator/riscv_sim.cpp
```diff
@@ -353,9 +359,11 @@ uint64_t load_sail(char *f, bool main_file)
   /* locate htif ports */
   if (lookup_sym(f, "tohost", &rv_htif_tohost) < 0) {
     fprintf(stderr, "Unable to locate htif tohost port.\n");
-    exit(1);
+    if (plat_enable_htif()) {
```
jrtc27, line 362: Why not just delete this code and set rv_enable_htif to false here? That's what we meant by removing the flag entirely.

c_emulator/riscv_sim.cpp
```diff
@@ -480,6 +488,9 @@ void reinit_sail(uint64_t elf_entry)
 
 void write_signature(const char *file)
 {
+  if (!plat_enable_htif()) {
```
jrtc27, line 491: Still no. Signature should not be tied to HTIF.

c_emulator/riscv_sim.cpp
trdthg: it's stderr -> stdout
Isn't this just a regular log?

c_emulator/riscv_sim.cpp
trdthg: understand now, removed

c_emulator/riscv_sim.cpp
```diff
@@ -353,9 +353,10 @@ uint64_t load_sail(char *f, bool main_file)
   /* locate htif ports */
   if (lookup_sym(f, "tohost", &rv_htif_tohost) < 0) {
     fprintf(stderr, "Unable to locate htif tohost port.\n");
```
Timmmm, line 355: Maybe
```suggestion
    fprintf(stderr, "Unable to locate tohost symbol; disabling HTIF.\n");
```

c_emulator/riscv_sim.cpp
```diff
@@ -353,9 +353,10 @@ uint64_t load_sail(char *f, bool main_file)
...
-    exit(1);
+    rv_enable_htif = false;
+  } else {
+    fprintf(stdout, "tohost located at 0x%0" PRIx64 "\n", rv_htif_tohost);
```
Timmmm, line 358:
```suggestion
    fprintf(stdout, "HTIF located at 0x%0" PRIx64 "\n", rv_htif_tohost);
```

model/riscv_platform.sail
```diff
@@ -410,7 +413,7 @@ $endif
 function mmio_read forall 'n, 0 < 'n <= max_mem_access . (t : AccessType(ext_access_type), paddr : physaddr, width : int('n)) -> MemoryOpResult(bits(8 * 'n)) =
   if   within_clint(paddr, width)
   then clint_load(t, paddr, width)
-  else if within_htif_readable(paddr, width) & (1 <= 'n)
+  else if (plat_enable_htif() & within_htif_readable(paddr, width) & (1 <= 'n))
```
Timmmm, line 416: This isn't quite right because it will give you access faults when you try to read where HTIF was.

I would instead just add a check for `plat_enable_htif()` to `within_htif_readable/writable()`.

Timmmm (changes requested):
This is good. Just a small issue...

model/riscv_platform.sail
trdthg: You mean move `plat_enable_htif` inside `within_htif_readable`?

```
function within_htif_writable ... =
    plat_enable_htif() & (origin_checks())
```

trdthg:
> As you're adding a new built-in, could you add it to the handwritten Rocq, Lem, and Lean files?

Of cource, thanks for you help!

c_emulator/riscv_sim.cpp
arichardson: This is consistent with the  "ELF Entry @ 0x%" print so that seems fine. I'd view all of these as debug prints that should go to stderr, but don't really mind.

c_emulator/riscv_platform.h
```diff
@@ -19,6 +19,8 @@ mach_bits plat_htif_tohost(unit);
 
 bool sys_enable_experimental_extensions(unit);
 
+bool plat_enable_htif();
+
```
arichardson, line 23: Nit: probably best to move it immediately after plat_htif_tohost()

c_emulator/riscv_platform_impl.cpp
```diff
@@ -20,6 +20,7 @@ uint64_t rv_16_random_bits(void)
 uint64_t rv_htif_tohost = UINT64_C(0x80001000);
 
 bool rv_enable_experimental_extensions = false;
+bool rv_enable_htif = true;
```
arichardson, line 23: same here, I'd group it with the other htif variable.

arichardson (commented):
This looks sensible to me now.

model/riscv_platform.sail
Timmmm: Exactly

model/riscv_platform.sail
```diff
@@ -53,6 +53,9 @@ function plat_rom_size() -> physaddrbits = to_bits(physaddrbits_len, config plat
 function plat_clint_base() -> physaddrbits = to_bits(physaddrbits_len, config platform.clint.base : int)
 function plat_clint_size() -> physaddrbits = to_bits(physaddrbits_len, config platform.clint.size : int)
 
+// Whether htif enabled
```
Timmmm, line 56:
```suggestion
// Whether HTIF (Host Target InterFace) is enabled. This is used for
// console output and signalling the end of tests.
```

Timmmm (approved):
LGTM. One minor comment improvement (I can commit it if you like).

pmundkur (approved):

jordancarlin (approved):

PR #894, opened 2025-04-28
Fix read_vreg for large num_elem when LMUL_pow < 0

wanghuibin0:
This patch fixes a logic error in the execute function for vector widen instructions.
Previously, the number of elements (num_elem) was obtained via get_num_elem(LMUL_pow, SEW).
When LMUL_pow < 0, this call returns the number of elements across the full register based on SEW, which is convenient for unified tail handling.

However, using this num_elem to perform read_vreg can cause an out-of-bounds access when reading a widened destination register, because a single vreg may not contain enough elements after widening.

For example, in vwadd.vv with VLEN=128, SEW=8, and LMUL_pow=-3, num_elem=16.
Since vd has a widened SEW of 16 bits, using read_vreg(num_elem, SEW_widen, LMUL_pow_widen, vd) would access beyond the valid range.

The fix is to compute the element count using SEW_widen and LMUL_pow_widen instead.

### DIFF
--- a/model/riscv_vext_regs.sail
+++ b/model/riscv_vext_regs.sail
@@ -290,11 +290,20 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
     let 'num_elem_single : int = VLEN / SEW;
     assert('num_elem_single >= 0);
     if LMUL_pow < 0 then {
-      /* Use num_elem_single to call read_single_vreg.
-       * When n is greater than num_elem_single, we will get a out-of-bounds access if we try to read n elements.
-       * Take vwadd.vv (VLEN=128, SEW=8, and LMUL_pow=-3) as an example, we have num_elem=16. When we perform read_vreg
-       * on the destination reg, we have SEW = 16, and then num_elem_single = 8.
-       * (See https://github.com/riscv/sail-riscv/pull/894) */
+      /* Use num_elem_single with read_single_vreg instead of n to prevent out-of-bounds access.
+       * 
+       * For widening instructions with LMUL < 1, the destination has EEW = 2×SEW and EMUL = 2xLMUL, which reduces
+       * the number of elements that fit in a single vector register compared to the source operands.
+       * Using the source's element count (n) when accessing the destination register can cause
+       * out-of-bounds access since fewer elements fit due to the doubled element width.
+       *
+       * Example: vwadd.vv with VLEN=128, SEW=8, LMUL=1/8 (LMUL_pow=-3):
+       * - Source operands: EEW=8, EMUL=1/8, 128/8=16 elements per source register
+       * - Destination: EEW=16, EMUL=1/4, 128/16=8 elements per destination register
+       * - Using n=16 for destination access would exceed the 8-element capacity
+       *
+       * See: https://github.com/riscv/sail-riscv/pull/894
+       */```
       let single_result : vector('num_elem_single, bits('m)) = read_single_vreg('num_elem_single, SEW, vrid);
       foreach (i from 0 to ('num_elem_single - 1)) {
         assert(0 <= i & i < 'n);


nadime15 (commented):
Hey, this is actually a really interesting catch but I have a follow up questions.

1. You are now doing this right?
`let num_elem = get_num_elem(LMUL_pow_widen, SEW_widen);`

But doesn't this halve the number of elements you're processing? Since you're doubling the SEW, you end up with fewer elements.
In your example, you'd get 8 elements of 16 bits each, even though originally SEW was 8 bits, and you should have 16 elements of 8 bits.

nadime15:
I think the problem is that `read_vreg` (when LMUL_pow < 0) directly calls:
```
result = read_single_vreg('n, SEW, vrid);
```

without verifying whether 'n and SEW are within valid boundaries.
This becomes a valid edge case during a widening operation, given how it's currently implemented.

nadime15:
I tried to run the example with the vector-test suite (VLEN=128, SEW = 8, LMUL=mf8) and its working for me. Do you can provide a testcase and show us an failing example?

So after thinking for a while, I think the current implementation is correct. 

> However, using this num_elem to perform read_vreg can cause an out-of-bounds access when reading a widened destination register, because a single vreg may not contain enough elements after widening.

The number of elements is not changed in case of a widening.

> For example, in vwadd.vv with VLEN=128, SEW=8, and LMUL_pow=-3, num_elem=16.
> Since vd has a widened SEW of 16 bits, using read_vreg(num_elem, SEW_widen, LMUL_pow_widen, vd) would access beyond the valid range.

This shouldn't be an issue because when LMUL < 1 (m1), a widening operation can at most use the entire register. It won't exceed the register size, since even a 2× (widening) would still fit within one vector register.

wanghuibin0:
I'll take a look at the [vector test suite](https://github.com/chipsalliance/riscv-vector-tests) you mentioned and see if it works on my side. If it does turn out to be a bug, I'll share my test case. 

It might take a little time, so please bear with me.

nadime15:
Ok I had another look! I think I finally get what you mean, you are right! :smile: 

Having a look at read_vreg() reveals that in case of LMUL_pow() the code enters the second if statement and then `read_single_vreg()`and without checking if `'n * SEW <= VLEN` and, as @wanghuibin0 has said, repeatedly calls `slice()` and has out-of-bounds access.

This part needs a rewrite, and for now, we should at least catch it with an assertion!

```
/* The general vreg reading operation with num_elem as max(VLMAX,VLEN/SEW)) */
val read_vreg : forall 'n 'm 'p, 'n >= 0 & 'm >= 0. (int('n), int('m), int('p), vregidx) -> vector('n, bits('m))
function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
  let vrid_val = unsigned(vregidx_bits(vrid));
  var result : vector('n, bits('m)) = vector_init(zeros());

  /* Check for valid vrid */
  if vrid_val + 2 ^ LMUL_pow_reg > 32 then {
    /* vrid would read past largest vreg (v31) */
    assert(false, "invalid register group: vrid overflow the largest number")
  } else if vrid_val % (2 ^ LMUL_pow_reg) != 0 then {
    /* vrid must be a multiple of emul */
    assert(false, "invalid register group: vrid is not a multiple of EMUL")
  } else {
    if LMUL_pow < 0 then {
      result = read_single_vreg('n, SEW, vrid); // <-----------------------------------------------
    } else {
      let 'num_elem_single : int = VLEN / SEW;
      assert('num_elem_single >= 0);
      foreach (i_lmul from 0 to (2 ^ LMUL_pow_reg - 1)) {
        let r_start_i : int = i_lmul * 'num_elem_single;
        let r_end_i   : int = r_start_i + 'num_elem_single - 1;
        let vrid_lmul     : vregidx = vregidx_offset(vrid, to_bits(5, i_lmul));
        let single_result : vector('num_elem_single, bits('m)) = read_single_vreg('num_elem_single, SEW, vrid_lmul);
        foreach (r_i from r_start_i to r_end_i) {
          let s_i : int = r_i - r_start_i;
          assert(0 <= r_i & r_i < num_elem);
          assert(0 <= s_i & s_i < 'num_elem_single);
          result[r_i] = single_result[s_i];
        }
      }
    }
  };

  result
}
```
Here `read_signle_vreg()`


```
/* Reads a single vreg into multiple elements */
val read_single_vreg : forall 'n 'm, 'n >= 0 & 'm >= 0. (int('n), int('m), vregidx) -> vector('n, bits('m))
function read_single_vreg(num_elem, SEW, vrid) = {
  let bv     : vregtype                  = V(vrid);
  var result : vector('n, bits('m)) = vector_init(zeros());

  assert(8 <= SEW & SEW <= 64);
    let start_index = i * SEW;
    result[i] = slice(bv, start_index, SEW);
  };

  result
}
```

@Alasdair I think this is probably quite important, I guess that could be used as an exploit?

wanghuibin0:
Yes, that's exactly what I meant!

What I'm not sure about now is how best to fix it. It seems like there are two possible approaches. One is to fix the callers of `read_vreg` — they should make sure not to pass in a `num_elem` that's too large when `LMUL_pow < 0`. That’s the approach I took in my patch.

The other option is to modify `read_vreg` itself to handle out-of-range `num_elem` values properly. Of course, `write_vreg` would need a similar change too. That way, we wouldn’t have to touch the caller code at all.

@nadime15 @Alasdair What do you think?

nadime15:
Yeah, I agree `read_vreg()` and `write_vreg()` need to be modified. It might confuse the reader if we go with the first solution because it seems counterintuitive to use `SEW_widen` instead of `SEW` for `get_num_elem()` .

wanghuibin0:
OK, I'll try to fix this issue by the second approach.

wanghuibin0:
I’ve rethought the second approach and made a new commit.
This time, I mainly updated the `read_vreg` logic for the `LMUL_pow < 0` case — now it checks if `num_elem` is too large, and if so, it fills the out-of-bound elements in the result vector with zeros.
As for `write_vreg`, I don’t think any changes are needed there, since its current logic doesn’t seem to be affected by an oversized num_elem.

Could you help me review it? @nadime15

pmundkur (changes requested):
Good catch!  This looks good to me.  

It would be good to have a clarifying comment in the `if LMUL_pow < 0 then` block explaining the situation with `num_elem` and `SEW_widen` under widening.

It looks like `write_vreg` is indeed okay in this case, since tail handling (for writing `num_elem_single  > num_elem` elements) has already been done by `init_masked_result`.

model/riscv_vext_regs.sail
```diff
@@ -288,7 +288,20 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
     assert(false, "invalid register group: vrid is not a multiple of EMUL")
   } else {
     if LMUL_pow < 0 then {
```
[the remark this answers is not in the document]
nadime15: I forgot to click submit... :smile:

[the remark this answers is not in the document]
wanghuibin0: You're right — the unhandled section is already zero-initialized.
I'll submit an updated patch with comments explaining this.

wanghuibin0:
Hi, I have resolved it and made a new commit.
Could you help me review it? @nadime15 @pmundkur

model/riscv_vext_regs.sail
```diff
@@ -287,21 +287,19 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
...
+       * Take vwadd.vv (VLEN=128, SEW=8, and LMUL_pow=-3) as an example, we have num_elem=16. When we perform read_vreg
+       * on the destination reg, we have SEW = 16, and then num_elem_single = 8.
+       * (See https://github.com/riscv/sail-riscv/pull/894) */
+      let single_result : vector('num_elem_single, bits('m)) = read_single_vreg('num_elem_single, SEW, vrid);
+      foreach (i from 0 to ('num_elem_single - 1)) {
+        assert(0 <= i & i < 'n);
+        result[i] = single_result[i];
+      };
     } else {
       let 'num_elem_single : int = VLEN / SEW;
```
nadime15, line 304: Did you forget to remove this line? Because you have the same at line number 290

wanghuibin0: done

nadime15:
I tested the fix with the [vector test suite](https://github.com/chipsalliance/riscv-vector-tests) (VLEN=512, ELEN=64), lots of failures suddenly. Stopped after ~650 tests. Here's the list:

```
❌ Failed on vluxei16_v-0 [596/2021]
❌ Failed on vluxei16_v-1 [596/2021]
❌ Failed on vluxei32_v-0 [596/2021]
❌ Failed on vluxei8_v-0 [598/2021]
❌ Failed on vluxei8_v-1 [598/2021]
❌ Failed on vluxseg2ei16_v-0 [599/2021]
❌ Failed on vluxseg2ei16_v-1 [599/2021]
❌ Failed on vluxseg2ei32_v-0 [600/2021]
❌ Failed on vluxseg2ei8_v-0 [603/2021]
❌ Failed on vluxseg2ei8_v-1 [603/2021]
❌ Failed on vluxseg2ei8_v-2 [603/2021]
❌ Failed on vluxseg3ei16_v-0 [603/2021]
❌ Failed on vluxseg3ei16_v-1 [603/2021]
❌ Failed on vluxseg3ei32_v-0 [603/2021]
❌ Failed on vluxseg3ei8_v-0 [605/2021]
❌ Failed on vluxseg3ei8_v-1 [605/2021]
❌ Failed on vluxseg4ei16_v-0 [605/2021]
❌ Failed on vluxseg4ei16_v-1 [605/2021]
❌ Failed on vluxseg4ei16_v-2 [605/2021]
❌ Failed on vluxseg4ei32_v-0 [605/2021]
❌ Failed on vluxseg4ei32_v-1 [605/2021]
❌ Failed on vluxseg4ei8_v-0 [607/2021]
❌ Failed on vluxseg4ei8_v-1 [607/2021]
❌ Failed on vluxseg4ei8_v-2 [607/2021]
❌ Failed on vluxseg5ei16_v-0 [607/2021]
❌ Failed on vluxseg5ei16_v-1 [607/2021]
❌ Failed on vluxseg5ei32_v-0 [607/2021]
❌ Failed on vluxseg5ei32_v-1 [607/2021]
❌ Failed on vluxseg5ei8_v-0 [608/2021]
❌ Failed on vluxseg5ei8_v-1 [608/2021]
❌ Failed on vluxseg6ei16_v-0 [608/2021]
❌ Failed on vluxseg6ei16_v-1 [608/2021]
❌ Failed on vluxseg6ei32_v-0 [608/2021]
❌ Failed on vluxseg6ei32_v-1 [608/2021]
❌ Failed on vluxseg6ei8_v-0 [609/2021]
❌ Failed on vluxseg6ei8_v-1 [609/2021]
❌ Failed on vluxseg6ei8_v-2 [609/2021]
❌ Failed on vluxseg7ei16_v-0 [609/2021]
❌ Failed on vluxseg7ei16_v-1 [609/2021]
❌ Failed on vluxseg7ei16_v-2 [609/2021]
❌ Failed on vluxseg7ei32_v-0 [609/2021]
❌ Failed on vluxseg7ei32_v-1 [609/2021]
❌ Failed on vluxseg7ei8_v-0 [610/2021]
❌ Failed on vluxseg7ei8_v-1 [610/2021]
❌ Failed on vluxseg7ei8_v-2 [610/2021]
❌ Failed on vluxseg8ei16_v-0 [610/2021]
❌ Failed on vluxseg8ei16_v-1 [610/2021]
❌ Failed on vluxseg8ei16_v-2 [610/2021]
❌ Failed on vluxseg8ei32_v-0 [610/2021]
❌ Failed on vluxseg8ei32_v-1 [610/2021]
...
```

There were many failures with different types of loads. At the moment I don't know what the cause is, but we need to investigate. I will look into that as well.

model/riscv_vext_regs.sail
```diff
@@ -287,11 +287,20 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
...
   } else {
+    let 'num_elem_single : int = VLEN / SEW;
+    assert('num_elem_single >= 0);
     if LMUL_pow < 0 then {
-      result = read_single_vreg('n, SEW, vrid);
+      /* Use num_elem_single to call read_single_vreg.
+       * When n is greater than num_elem_single, we will get a out-of-bounds access if we try to read n elements.
+       * Take vwadd.vv (VLEN=128, SEW=8, and LMUL_pow=-3) as an example, we have num_elem=16. When we perform read_vreg
+       * on the destination reg, we have SEW = 16, and then num_elem_single = 8.
+       * (See https://github.com/riscv/sail-riscv/pull/894) */
```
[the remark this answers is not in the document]
wanghuibin0: That' ok. I think our project is benefiting from your "nitpicking"

model/riscv_vext_regs.sail
```diff
@@ -287,11 +287,29 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
     /* vrid must be a multiple of emul */
     assert(false, "invalid register group: vrid is not a multiple of EMUL")
   } else {
+    let 'num_elem_single : int = VLEN / SEW;
```
Timmmm, line 290:
```suggestion
    let 'num_elem_single = VLEN / SEW;
```

Generally it's better to avoid adding explicit type annotations like this because it reduces the amount of information Sail knows about the type. You can probably remove the `assert` on the next line without this `int`.

model/riscv_vext_regs.sail
```diff
@@ -287,11 +287,29 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
...
+       * Using the source's element count (n) when accessing the destination register can cause
+       * out-of-bounds access since fewer elements fit due to the doubled element width.
+       *
+       * Example: vwadd.vv with VLEN=128, SEW=8, LMUL=1/8 (LMUL_pow=-3):
+       * - Source operands: EEW=8, EMUL=1/8, 128/8=16 elements per source register
+       * - Destination: EEW=16, EMUL=1/4, 128/16=8 elements per destination register
+       * - Using n=16 for destination access would exceed the 8-element capacity
+       *
+       * See: https://github.com/riscv/sail-riscv/pull/894
+       */```
```
Timmmm, line 306:
```suggestion
       */
```

Also I think just use `//`. I don't see the point of using `/*` if you're just going to prefix every line anyway...

model/riscv_vext_regs.sail
```diff
@@ -287,11 +287,29 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
...
+       * the number of elements that fit in a single vector register compared to the source operands.
+       * Using the source's element count (n) when accessing the destination register can cause
+       * out-of-bounds access since fewer elements fit due to the doubled element width.
+       *
+       * Example: vwadd.vv with VLEN=128, SEW=8, LMUL=1/8 (LMUL_pow=-3):
+       * - Source operands: EEW=8, EMUL=1/8, 128/8=16 elements per source register
+       * - Destination: EEW=16, EMUL=1/4, 128/16=8 elements per destination register
+       * - Using n=16 for destination access would exceed the 8-element capacity
+       *
+       * See: https://github.com/riscv/sail-riscv/pull/894
```
Timmmm, line 305: Probably remove this - you can get back to the PR via `git blame` and I think you've included plenty of information!

model/riscv_vext_regs.sail
```diff
@@ -287,11 +287,29 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
...
+       */```
+      let single_result : vector('num_elem_single, bits('m)) = read_single_vreg('num_elem_single, SEW, vrid);
+      foreach (i from 0 to ('num_elem_single - 1)) {
+        assert(0 <= i & i < 'n);
```
Timmmm, line 309: You shouldn't need `0 <= i`.

Timmmm:
Since this is a bug that would have been caught if the vector code had proper types, could we maybe try improving them around this a bit. E.g. can you change this:

```
    result[i] = slice(bv, start_index, SEW);
```

to use normal `bv[..]` indexing? I assume `slice()` only does runtime index checking and that's why it was sneakily used. Also there are a load of other uses of `slice()` that we should remove.

Timmmm:
I'm kind of confused now - I tested the new configurable `vlen` against the vector tests [here](https://github.com/riscv/sail-riscv/actions/runs/16343464307) and it all passed, but now I'm running into this issue and it's causing an assertion failure because of the out-of-bounds slice. I guess it didn't cause an issue before because the vector registers were actually always 65536 bits, but I don't understand why my run of the vector tests passed. Unless these are very recently added tests or something?

Anyway this reproduces the failure on `master`. It would be good to fix it.

```
cd build
c_emulator/sail_riscv_sim --config test/config_rv64d_v128_e64.json test/2025-07-16/riscv-vector-tests-v128x64/rv64vloxei16_v-0
```

pmundkur:
See #1179.  The better fix would be to take `vlen` into account in the type constraints on `read_single_vreg` so that the buggy call would have a type error, but that would be a much bigger PR.  That should also allow removing the use of `slice` in `read_single_vreg`.

Timmmm:
> but I don't understand why my run of the vector tests passed.

Ah, so it turns out the test I posted passes if you build in release mode, but not in debug mode. CMake defines `-DNDEBUG` by default for release modes so the assertion is skipped.

I think we should:

1. Change the `slice()` to a `[]` with a native Sail `assert()` (this doesn't depend on `NDEBUG`). I'll add a commit to your PR with that.
2. Probably either use debug mode for CI or force assertions to be enabled in CI. I'll make an issue for this.

PR #895, opened 2025-04-28
riscv_types cleanup

jordancarlin:
- Convert enum-bits covervsions to mappings where possible
- Remove unused types

### DIFF
--- a/model/riscv_insts_zbb.sail
+++ b/model/riscv_insts_zbb.sail
@@ -247,7 +247,7 @@ mapping clause assembly = CPOP(rs1, rd)
 function clause execute (CPOP(rs1, rd)) = {
   let rs1_val = X(rs1);
   var result : nat = 0;
-  foreach (i from 0 to (xlen_val - 1))
+  foreach (i from 0 to (xlen - 1))
     if rs1_val[i] == bitone then result = result + 1;
   X(rd) = to_bits(xlen, result);
   RETIRE_SUCCESS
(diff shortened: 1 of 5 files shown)


model/riscv_insts_zbb.sail
```diff
@@ -247,7 +247,7 @@ mapping clause assembly = CPOP(rs1, rd)
 function clause execute (CPOP(rs1, rd)) = {
   let rs1_val = X(rs1);
   var result : nat = 0;
-  foreach (i from 0 to (xlen_val - 1))
+  foreach (i from 0 to (xlen - 1))
```
pmundkur, line 250: I think you forgot to remove `xlen_val` after this.

pmundkur (changes requested):
LGTM other than a missed cleanup.

model/riscv_insts_zbb.sail
jordancarlin: I left it in deliberately. Per the comment currently there, it is for use in the C emulator but shouldn't be used in the Sail.

```
/* this value is only defined for the runtime platform for ELF loading
 * checks, and not used in the model.
 */
let xlen_val = xlen
```

I don't see a reason not to use `xlen` directly in the C, but I figured there must have been a reason this choice was made and didn't want to change it.

pmundkur (approved):

model/riscv_insts_zbb.sail
jordancarlin: Took a closer look at the C side, and it seems like there are no issues with switching to using `xlen` directly, so I switched it and removed `xlen_val`.

Timmmm (approved):
Nice!

PR #896, opened 2025-04-29
Inline machine info register initialization

jordancarlin:
Not sure why I didn't think of this when reviewing #880. This inlines the initialization of the machine info registers, similar to what we do for several other registers throughout the model. These values are set at model initialization time.


pmundkur (approved):
Indeed, this is simpler.

Timmmm (approved):
LGTM. I think given we are going to be using `to_bits` a lot we should probably prioritise #614 and also #898 would help.

I might have a go now.

PR #899, opened 2025-04-29
Make truncate infer its length

### LINKED ISSUE #898: Make to_bits and truncate infer their length
`zero_extend()` and `zeros()` can infer the length from their context, but `truncate()` and `to_bits()` can't, which is a bit weird. We should probably change the latter so that they do.

See also #614

Timmmm:
Unfortunately I had to rename it to trunc, but I think this is still better.

Partially fixes #898 

Better name suggestions welcome! @Alasdair FYI; I don't know if you have a plan for breaking changes to the standard library at some point but maybe add this to the list!

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -93,6 +93,11 @@ function zeros (n) = sail_zeros(n)
 val ones : forall 'n, 'n >= 0 . implicit('n) -> bits('n)
 function ones (n) = sail_ones(n)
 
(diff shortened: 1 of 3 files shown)


jordancarlin (commented):
Seems like a nice improvement! There are several more calls to `truncate` in `rvfi_dii.sail`.

Timmmm:
Yeah that code looked like it needed a full work over so I left it for now...

jordancarlin (approved):

PR #900, opened 2025-04-29
Make to_bits infer its length

### LINKED ISSUE #898: Make to_bits and truncate infer their length

Timmmm:
Make to_bits infer the length of its return value, and remove many of the values that can now be inferred.

I also simplified MULW a bit - since `to_bits()` truncates there's no need for a separate truncation step.

Partially fixes #898.

On reflection we should to #614 before merging this since this kind of loses some information (e.g. the `MULW` `to_bits()`) deliberately truncates, whereas some of the others don't.

### DIFF
--- a/model/main.sail
+++ b/model/main.sail
@@ -15,7 +15,7 @@ $ifdef SYMBOLIC
 
 $include <elf.sail>
 
-function get_entry_point() = to_bits(xlen, elf_entry())
+function get_entry_point() = to_bits(elf_entry())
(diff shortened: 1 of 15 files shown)


Timmmm:
Completed via other PRs.

PR #902, opened 2025-04-29
Minor cleanups to M extension

Timmmm:
Various minor improvement:

1. Instead of an `s` bool to indicate that DIV and REM are *not* the U(nsigned) variants, I inverted it. This removes the need for `bool_not_bits` and `maybe_not_u`. I also used a more descriptive variable name than `s`.

2. Change `operator %` to use truncated division instead of Euclidian division. We already had `/` using Euclidian division so this is more consistent. It also matches what RISC-V uses for `div`/`rem` so we can then use the operators in those functions' `execute` clauses.

3. Inline `xlen_max_unsigned` etc. which are only used once.

4. Rename `rs1_val` to `rs1_bits` which I think is a better name.

5. Removed some unnecessary `int` type annotations (these actually make the code worse since they discard more specific type information).

6. Simplify MULW 32-bit truncation.

7. Explicitly use `to_bits_truncate` which is intentional in this case.

I think we could potentially do the signed overflow detection more like a real chip would do it (just detect the specific values that can cause it rather than using a comparison), but I left it for now.

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -49,12 +49,10 @@ val sub_vec_int = pure {c: "sub_bits_int", lean: "BitVec.subInt", _: "sub_vec_in
 overload operator - = {sub_vec, sub_vec_int}
 
 val quot_positive_round_zero = pure {interpreter: "quot_round_zero", lem: "hardware_quot", lean: "Int.tdiv", c: "tdiv_int", coq: "Z.quot"} : forall 'n 'm, 'n >= 0 & 'm > 0. (int('n), int('m)) -> int(div('n, 'm))
-
 val quot_round_zero = pure {interpreter: "quot_round_zero", lem: "hardware_quot", c: "tdiv_int", coq: "Z.quot", lean: "Int.tdiv"} : (int, int) -> int
-val rem_round_zero = pure {interpreter: "rem_round_zero", lem: "hardware_mod", c: "tmod_int", coq: "Z.rem", lean: "Int.tmod"} : (int, int) -> int
 
-/* The following defines % as euclidean modulus */
-overload operator % = {emod_int}
+val rem_positive_round_zero = pure {interpreter: "rem_round_zero", lem: "hardware_mod", c: "tmod_int", coq: "Z.rem", lean: "Int.tmod"} : forall 'n 'm, 'n >= 0 & 'm > 0. (int('n), int('m)) -> int(mod('n, 'm))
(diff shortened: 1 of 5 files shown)


jordancarlin:
> * Slightly unrelated but remove use of `xlen_val` by Zbb which was probably a mistake. In future we will remove `xlen_val`.

This is removed in #895.

jordancarlin (approved):
Nice cleanup. Not sure about using `/` instead of `quot_round_zero` since we still use `rem_round_zero`, but I'm fine with it either way.

Timmmm:
Yeah I didn't like that either. I found a way to make % and / much more regular so we can use % and / in all cases.

This does change the semantics of % for negative values. I tried removing the version that allowed negative values to see if it still compiled, but unfortunately due to the very loose types in the vector extension there are places where it can't prove they are positive, even though I think they all are.

jordancarlin:
> Yeah I didn't like that either. I found a way to make % and / much more regular so we can use % and / in all cases.
> 
> This does change the semantics of % for negative values. I tried removing the version that allowed negative values to see if it still compiled, but unfortunately due to the very loose types in the vector extension there are places where it can't prove they are positive, even though I think they all are.

Seems nicer this way and I think that's probably ok for vector for now. Eventually that all needs a refactor to clean up the types anyway.

Timmmm:
Ok I'm not quite sure why this is failing. Works for me locally but I am using a newer version of the Sail compiler so maybe it's that. I'll try again when we upgrade it.

Timmmm:
@Alasdair - Possibly a Sail bug here too. It type checks for the C backend but not the SMT backend?

The issue is with this:

```
```

Because I changed the definition of `%` to be truncated instead of Euclidian I think it's now getting confused because `('n + 3) % 4` can be negative or something?

So it works fine if I add this constraint (which I'll do for now):

```
function hex_bits_str forall 'n, 'n >= 0 . (x : bits('n)) -> string =
```

but it still seems weird that it works fine for the C backend but not the SMT one. I vaguely recall you mentioned something about adding an implicit >= 0 constraint when type variables are used in `bits()`. Maybe that isn't being applied for SMT or something?

pmundkur:
There are too many unrelated changes in this one PR for my taste.  I'd prefer if this was split out into one each. Perhaps 3 and 4 could go together into one.

Timmmm:
Ok I'll split it up.

Timmmm:
This is all covered by other PRs now.

Issue #849, opened 2025-04-11
Separate Zicsr extension


jordancarlin:
Currently the Zicsr extension ins mixed into the base ISA and does not have a configuration option to enable/disable it. It is an optional extension in the spec so we need a way of turning it off in the model. 

This might be a trickier one because a lot of functions interact with the CSRs, but it may be sufficient to just disable the read/write instructions to them and it won't matter if the model is updating them in the background.

PR #904, opened 2025-04-29
Make Zicsr extension configurable

KotorinMinami:
As mentioned in #849, the spec states that the Zicsr extension is Extension for Control and Status Register (CSR) Instructions. Therefore, the encdec sections of the AST CSRReg and CSRImm are guard with a `when` condition to achieve this purpose.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -82,6 +82,9 @@
     "Zicntr": {
       "supported": true
     },
+    "Zicsr": {
+      "supported": true
+    },
     "Zifencei": {
       "supported": true
     },
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -64,6 +64,9 @@ function clause hartSupports(Ext_Zicntr) = config extensions.Zicntr.supported
 // Integer Conditional Operations
 enum clause extension = Ext_Zicond
 function clause hartSupports(Ext_Zicond) = config extensions.Zicond.supported
+// Extension for Control and Status Register (CSR) Instructions
+enum clause extension = Ext_Zicsr
+function clause hartSupports(Ext_Zicsr) = config extensions.Zicsr.supported
 // Instruction-Fetch Fence
 enum clause extension = Ext_Zifencei
 function clause hartSupports(Ext_Zifencei) = config extensions.Zifencei.supported
--- a/model/riscv_fdext_control.sail
+++ b/model/riscv_fdext_control.sail
@@ -15,9 +15,9 @@
 
 /* **************************************************************** */
 
-function clause currentlyEnabled(Ext_F) = hartSupports(Ext_F) & (misa[F] == 0b1) & (mstatus[FS] != 0b00)
-function clause currentlyEnabled(Ext_D) = hartSupports(Ext_D) & (misa[D] == 0b1) & (mstatus[FS] != 0b00) & flen >= 64
-function clause currentlyEnabled(Ext_Zfinx) = hartSupports(Ext_Zfinx)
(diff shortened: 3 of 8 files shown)


Alasdair:
I think the Zicsr extension only gates the CSR reading and writing instructions.

KotorinMinami:
I think it would be better to discuss the corresponding solution for this dependency issue in #860 before adapting it in this PR？

jordancarlin:
According to the spec Zicntr, F, Zfinx, Zcmt, V, and "the privileged architecture" require the Zicsr extension. While the config file validation can be deferred to #860, we have been capturing simple extension dependencies in the `currentlyEnabled` function. For example, 
```
function clause currentlyEnabled(Ext_Zihpm) = hartSupports(Ext_Zihpm) & currentlyEnabled(Ext_Zicntr)
```

I think at least adding something similar for Zicntr, F, Zfinx, Zcmt, and V in this PR would make sense.

Timmmm:
Sorry I didn't read your comment @jordancarlin before approving.

I guess the alternative is we add that validation function we were talking about:

```
function is_config_valid() -> bool =
  hartSupports(Ext_Zihpm) ==> hartSupports(Ext_Zicsr) &
  hartSupports(Ext_F) ==> hartSupports(Ext_Zicsr)
...
```

(with `==>` being an implication operator)

Maybe we should stick with adding it to the `currentlyEnabled()`s like you say for now.

jordancarlin (approved):
LGTM with the latest changes to the `currentlyEnabled` clauses.

pmundkur (changes requested):
Zihpm also depends on Zicsr.  Once that's fixed, LGTM.

jordancarlin:
> Zihpm also depends on Zicsr. Once that's fixed, LGTM.

Zihpm depends on Zicntr (which in turns depends on Zicsr), so I think we are good without explicitly adding it there.

pmundkur:
> > Zihpm also depends on Zicsr. Once that's fixed, LGTM.
> 
> Zihpm depends on Zicntr (which in turns depends on Zicsr), so I think we are good without explicitly adding it there.

Perhaps I'm missing something, but I don't see this in the spec.  The counters are split between Zicntr and Zihpm; Zihpm counters supplement Zicntr, but don't depend on it.

jordancarlin:
> > > Zihpm also depends on Zicsr. Once that's fixed, LGTM.
> > 
> > 
> > Zihpm depends on Zicntr (which in turns depends on Zicsr), so I think we are good without explicitly adding it there.
> 
> Perhaps I'm missing something, but I don't see this in the spec. The counters are split between Zicntr and Zihpm; Zihpm counters supplement Zicntr, but don't depend on it.

I was basing this off of what we currently have in the model, but looking at the spec I agree that Zihpm only depends on Zicsr.

Right now we have
```
function clause currentlyEnabled(Ext_Zihpm) = hartSupports(Ext_Zihpm) & currentlyEnabled(Ext_Zicntr)
```
So really the `Ext_Zicntr` requirement needs to be changed to `Ext_Zicsr`.

jordancarlin (changes requested):
The priv spec also says
>  The privileged architecture requires the Zicsr extension

Presumably that means S and U should depend on Zicsr as well.

pmundkur:
S makes sense, not sure about U.

jordancarlin (approved):

jordancarlin:
> S makes sense, not sure about U.

Not sure that it would be possible to have multiple privilege modes without CSRs.

allenjbaum:
Also all FP extensions require CSRs

jordancarlin:
> Also all FP extensions require CSRs

I think we’ve captured that already. F and Zfinx are set to require Zicsr and all other FP extensions require one of those two.

pmundkur (approved):

PR #905, opened 2025-04-30
Add Decompress result for compressed instructions

Alasdair:
The rationale for this is something we encountered when doing CHERIoT verification using the generated SystemVerilog from Sail. By having the compressed instructions directly invoke execute for the decompressed variant, we end up with (for example) all the logic for load and store instructions duplicated in the output.

It's also a bit awkward because it makes the main `execute` function recursive, which inhibits some useful Sail optimisations.

This resolves this by adding a new return option to the execute function via the `ExecutionResult` type (as enabled by some other recent commits).
```
Decompress : ast
```
which allows the top-level step function to handle executing the decompressed variant.

As part of this PR that type is moved to where the `ast` type is declared rather than being in a separate file.

In the actual CHERIoT verification we solved this slightly differently by adding a separate `decompress` function, but I think that solution was a bit more inelegant because it meant that some `ast` members could not be executed, only decompressed, and vice versa without any real way to distinguish them. This solution is a bit nicer in that regard.

Not 100% sure I'm handling it in the correct way in the step function, would be good to get some input there.


model/riscv_insts_zca.sail
```diff
@@ -54,7 +54,7 @@ function clause execute (C_LW(uimm, rsc, rdc)) = {
   let imm : bits(12) = zero_extend(uimm @ 0b00);
   let rd = creg2reg_idx(rdc);
   let rs = creg2reg_idx(rsc);
-  execute(LOAD(imm, rs, rd, false, WORD, false, false))
+  Decompress(LOAD(imm, rs, rd, false, WORD, false, false))
 }
```
pmundkur, line 58: It might be awkward to explain this when using `execute` snippets in ISA documentation.

pmundkur (commented):
The step handling looks good to me.  I'm wondering whether we should remove the redundant check for `Ext_Zca` in this PR.

model/riscv_insts_zca.sail
Alasdair: maybe change the name to something like `ExecuteAs` so it reads the same?

model/riscv_insts_begin.sail
```diff
@@ -13,6 +13,38 @@
...
+  Memory_Exception               : (virtaddr, ExceptionType),
+
+  // Did not retire for custom reason.
+  Ext_CSR_Check_Failure          : unit,
+  Ext_ControlAddr_Check_Failure  : ext_control_addr_error,
+  Ext_DataAddr_Check_Failure     : ext_data_addr_error,
+  Ext_XRET_Priv_Failure          : unit,
+}
+
+// To ease the introduction of `Retire_Failure`, this global
```
jordancarlin, line 41: Unrelated to this PR, but this comment needs to be updated. It is no longer for sake of `Retire_Failure`.

model/riscv_insts_zca.sail
jordancarlin: Agreed that `ExecuteAs` would be a more clear name for this.

jordancarlin (commented):
Seems like nice refactor.

model/riscv_insts_begin.sail
pmundkur: I can fix it in #777 since I'm touching that code there.

jordancarlin: It was being moved here anyway so I figured it could be changed here. Either way works though.

model/riscv_insts_zca.sail
Timmmm: Heh I wrote this comment but forgot to hit submit:

> Seems reasonable to me. I would just suggest renaming `Decompress` since it's already been decoded by that point so there's not really any decompression. Maybe something like `ExecuteAs()`? 

Clearly a great name. :-D

model/riscv_step.sail
```diff
@@ -68,8 +68,10 @@ function run_hart_active(step_no: nat) -> Step = {
         /* check for RVC once here instead of every RVC execute clause. */
         if currentlyEnabled(Ext_Zca) then {
           nextPC = PC + 2;
-          let r = execute(ast);
-          Step_Execute(r, instbits)
+          match execute(ast) {
+            Decompress(ast) => Step_Execute(execute(ast), instbits),
```
arichardson, line 72:
```suggestion
            Decompress(expanded) => Step_Execute(execute(expanded), instbits),
```
I feel like using a different name makes this slightly easier to read?

pmundkur:
It would be good to update this for merging.

Timmmm:
I guess @Alasdair wouldn't mind if someone else did the final bit?

pmundkur:
Closing since this got rebased and merged via #1382.

PR #906, opened 2025-04-30
Fix Svbare warning.

pmundkur:
Also add Svbare to the config and README as a supported extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ For booting operating system images, see the information under the
 - Sstc extension for Supervisor-mode Timer Interrupts, v1.0
 - Svinval extension for fine-grained address-translation cache invalidation, v1.0
 - Sv32, Sv39, Sv48 and Sv57 page-based virtual-memory systems
+- Svbare extension for Bare mode virtual-memory translation
 - Physical Memory Protection (PMP)
 
 <!-- Uncomment the following section when unratified extensions are added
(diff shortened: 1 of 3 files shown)


nadime15 (approved):

README.md
```diff
@@ -114,6 +114,7 @@ For booting operating system images, see the information under the
 - Sstc extension for Supervisor-mode Timer Interrupts, v1.0
 - Svinval extension for fine-grained address-translation cache invalidation, v1.0
 - Sv32, Sv39, Sv48 and Sv57 page-based virtual-memory systems
+- Svbare extension for the Bare mode of satp
```
jordancarlin, line 117: Maybe use the name from the profiles doc:
> Bare mode virtual-memory translation supported

arichardson (approved):

jordancarlin (approved):
Seems like this was missed in #807. Simple fix.

README.md
pmundkur: I used the language from 
https://github.com/riscv/riscv-profiles/blob/main/src/profiles.adoc
```
Svbare The satp mode Bare must be supported.
```

jordancarlin: That’s the description of what the extension does, but lower down (https://github.com/riscv/riscv-profiles/blob/main/src/profiles.adoc#7-new-isa-extensions) there is a section that lists the extension names.

pmundkur: I missed that; fixed.

jordancarlin (approved):

PR #907, opened 2025-05-01
Fix termination measure again, try to make it more rubust to change

ineol:
Change the structure of the measure for `currentlyEnabled` to make the base extensions (A, B, C, D, M, ...) have measure 0 and the generic case be 1, so that adding an extension is hopefuly less likely to break terminiation checking if it only depends on these basic extensions.

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -19,7 +19,7 @@ scattered function hartSupports
 // extensions to be supported in hardware, but temporarily disabled via a CSR, in
 // which case this function should return false.
 // Note: when adding a new extension, adjust the associated termination measure
-// in the file riscv_terminiation_measure.sail, as explained in the comment in
+// in the file riscv_terminiation.sail, as explained in the comment in
 // that file.
 val currentlyEnabled : extension -> bool
 scattered function currentlyEnabled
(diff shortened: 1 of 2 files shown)


Timmmm:
Maybe worth adding a comment explaining how to update this. I think most people adding extensions have no idea what a termination measure is or why these values have to be used. If there was a comment we could remind people to update it...

Timmmm (approved):

ineol:
@Timmmm I tried adding a comment in the termination file as well as in the `riscv_extensions.sail` file to remind people to update the measure. I hope it's somewhat clear, but unfortunately the only way to test it is to build the Lean model../

model/riscv_extensions.sail
```diff
@@ -18,6 +18,9 @@ scattered function hartSupports
 // relevant CSRs (misa, mstatus, etc.) to enable its use. It is possible for some
 // extensions to be supported in hardware, but temporarily disabled via a CSR, in
 // which case this function should return false.
+// Note: when adding a new extension, adjust the associated termination measure
+// in the file riscv_terminiation_measure.sail, as explained in the comment in
```
jordancarlin, line 22: Name of the file doesn't match.
```suggestion
// in the file riscv_terminiation.sail, as explained in the comment in
```

jordancarlin (approved):

Timmmm (approved):

PR #889, opened 2025-04-26
Update Linux build scripts

Arielfoever:
Remove old pre-built Linux images and add a Makefile to build one from scratch, including downloading all the tools.

This Makefile also includes a target for profiling using gperftools.

Fixes #580

PR #911, opened 2025-05-01
Add missing riscv_sstc.sail to CMake build

arichardson:
This was not included when moving over to the CMake build system and resulted in a kernel panic when trying to boot linux (#889)


jordancarlin (approved):
Yikes. We really need to get updated versions of tests running to avoid things like this.

Timmmm (approved):

jrtc27:
I've verified this was the only file missing (with ``diff -u <(sed -n 's/.*"\(.*\.sail\)".*/\1/p' model/CMakeLists.txt | sort -u) <(cd model; ls *.sail)``)

PR #912, opened 2025-05-01
build_simulators.sh: use /bin/sh

arichardson:
This file does not depend on any bash features, we can use POSIX sh.

### DIFF
--- a/build_simulators.sh
+++ b/build_simulators.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 : "${DOWNLOAD_GMP:=TRUE}"


jordancarlin (approved):
Can you update the commit message? Seems unrelated to the actual change.

arichardson:
> Can you update the commit message? Seems unrelated to the actual change.

I uploaded this using the spr tool, landing it via that tool or squash merging will pick the PR description. I can force push a new version though next week if you prefer

jordancarlin:
> > Can you update the commit message? Seems unrelated to the actual change.
> 
> I uploaded this using the spr tool, landing it via that tool or squash merging will pick the PR description. I can force push a new version though next week if you prefer

Probably fine to change when merging.

Timmmm (approved):

PR #914, opened 2025-05-02
Fix compilation of the Lean backend

ineol:
Sail needs to give some of the implicit arguments to Lean because they are not deducible from the type of the arguments.

### DIFF
--- a/model/riscv_insts_vext_utils.sail
+++ b/model/riscv_insts_vext_utils.sail
@@ -191,9 +191,11 @@ function write_velem_quad(vd, SEW, input, i) = {
     write_single_element(SEW, 4 * i + j, vd, slice(input, j * SEW, SEW));
 }
 
-/* Extracts 8 consecutive vector elements starting from index 8*i and returns a vector */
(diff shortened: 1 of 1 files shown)


tobiasgrosser (approved):
LGTM.

Timmmm:
Should Sail be able to handle this for us? Feels kind of weird that it is required. I mean Sail can deduce it, why can't Lean?

Also definitely add a comment otherwise someone is going to come along and remove it later...

bacam:
Sail's unification is a bit stronger than the provers and can figure out that dividing by 8 gives the right size.  This special case isn't usually present in general purpose unification solvers, and may not be possible in general without recreating enough of Sail's type information to spot that the division is safe.  Ideally we'd somehow spot that the unifier isn't obvious and add it explicitly in the back-end, but adding an implicit argument provides Sail's solution as a workaround.

That said, does Lean have support for user-defined unification hints?  I searched around a little and it seems like it does, but I didn't find any proper documentation.  Perhaps those could be used instead.

ineol:
It does have unification hints but they're undocumented and I never managed to do anything with them, so I don't think it's a short/medium term solution.

ineol:
I added comments to explain the implicit arguments.

tobiasgrosser:
Thank you, @ineol.

PR #916, opened 2025-05-05
Update README to reflect the cleanup in 587294254.

pmundkur:
Simple PR to test the merge queue.


jordancarlin (approved):
Necessary and a good test of the queue.

jordancarlin:
Seems like this still isn't working quite right. It has the build job as a pending check and gives the option to "merge when ready", but not to add it to the actual queue.

jordancarlin:
Looks like this merged without going through the queue after you force pushed because auto-merge was still on from when I tried adding it to the queue.

jordancarlin:
And it used a merge commit instead of squashing like we'd set.

PR #917, opened 2025-05-05
Remove unneeded diff file from os-boot.

pmundkur:
This diff is obsolete and is not needed to boot the linux image.

Another test of the merge queue.


jordancarlin (approved):
Removed in #889 anyway. Let’s try this again.

pmundkur:
This one got squashed, but the commit message got rewritten.

jordancarlin:
> This one got squashed, but the commit message got rewritten.

Looks like it used the PR message for the commit message. Makes sense since if it is squashing multiple commits it needs to pick something as the message.

Timmmm:
Hmm yeah it seems like you can't edit the commit message if using the merge queue unfortunately. It seems like it will use the default message which is set here:

image

I guess it means we can just edit the PR title/message before adding to the merge queue if we want to change it. It does also add the PR number to the subject and reflow everything to 80 columns, both of which are pretty pointless and ugly IMO, but there's no way to turn those off so I guess we'll have to live with it.

I found [this feature request](https://github.com/orgs/community/discussions/15925) for it but it's 3 years old so I guess we shouldn't hold our breaths!

pmundkur:
We could pick a different option.  Any suggestions?
squash_options

jordancarlin:
I think default does the commit title and message if it is a single commit and the PR title and description if it is multiple commits. That could work well.

pmundkur:
Sounds good, I set it to default message.

Timmmm:
That means we can't easily edit it though? I'd rather not have to hassle people to write good commit messages...

jrtc27:
LLVM uses "Pull request title and description" and that seems to work for them.

jordancarlin:
That's a good point regarding editing. Maybe we stick with "Pull request title and description" and just need to get into the flow of editing the PR message if needed before merging.

pmundkur:
Ok, set it back to PR title and description.

PR #918, opened 2025-05-05
Add the config schema to the build.

pmundkur:
The schema is generated in the default (c_emulator) build target variant.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -218,6 +218,8 @@ foreach (variant IN ITEMS "" "smt" "rocq" "rmem" "lean" "lem" "isabelle")
         set(c_model_no_ext "${CMAKE_BINARY_DIR}/sail_riscv_model")
         set(c_model "${c_model_no_ext}.c")
         set(c_model_header "${c_model_no_ext}.h")
+        # JSON schema for configuration file.
+        set(schema_file "${CMAKE_BINARY_DIR}/sail_riscv_config_schema.json")
 
         if (COVERAGE)
             set(branch_info_file "${c_model_no_ext}.branch_info")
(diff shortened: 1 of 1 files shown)


jordancarlin (commented):
We should probably generate the schema as part of the default target, since I imagine we will want to validate against it before loading a config file in the future.

The xlen/flen situation is a little weird with this one, but that should be resolved once #870 is merged, so not a big deal for now.

c_emulator/CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ foreach (xlen IN ITEMS 32 64)
 
             install(TARGETS riscv_sim_${arch}
                 OPTIONAL
-                RUNTIME DESTINATION "bin"
+                RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
```
Timmmm, line 79: Actually I think we only need

```suggestion
                RUNTIME
```

It uses `CMAKE_INSTALL_BINDIR` by default.

Timmmm (changes requested):
Yeah I agree with @jordancarlin - we can probably just add this flag to the C build command.

We could either do that now with `--output-schema "riscv_config_schema_${arch}.json"` or just wait for #870. I'd probably vote for the latter given Alasdair said they will probably do another release soon.

jordancarlin:
This would be a good one to revisit now that #870 is merged.

Timmmm (approved):

jordancarlin (commented):
Looks like installing the generated schema was dropped in the rebase.

pmundkur:
> Looks like installing the generated schema was dropped in the rebase.

Yes, that was intentional, since that installation was being only used for the schema file.  I thought we should handle installation of all the build artifacts together, like default configs and the simulator binary?

PR #919, opened 2025-05-06
Enable match_bv for Lean

tobiasgrosser:
This will make matching on bitvector pattern a lot more readable.

This is still in draft stage, as one last sail bug prevents us from using this option.


arichardson:
Is this still needed or can we close it?

jordancarlin:
@tobiasgrosser any update? Is this still needed? If not, can you close it?

PR #920, opened 2025-05-06
Lean: add stubs for the new callback functions

ineol:
This fixes the Lean build.


tobiasgrosser (approved):
LGTM. Thank you, @ineol

jordancarlin:
Would defining default empty implementations in the Sail work too? Seems like it has the advantage of fixing all of the backends instead of needing to do it separately for each one.

bacam:
Yes, I think default Sail implementations would be preferable, especially as these functions can't actually do anything.

PR #921, opened 2025-05-07
Add option to print abi or architectural register names in trace and disassembly

### LINKED ISSUE #909: Add option to print registers using ABI names in trace
As mentioned here (https://github.com/riscv/sail-riscv/pull/494#discussion_r2070268477), printing the ABI names for the registers is often more useful than the x names. Once #494 is merged, this should be added as another logging option.

jordancarlin:
Fixes #909

### DIFF
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -62,11 +62,17 @@ unit mem_exception_callback(uint64_t paddr, uint64_t num_of_exception)
   return UNIT;
 }
 
-unit xreg_write_callback(unsigned reg, uint64_t value)
+unit xreg_full_write_callback(const_sail_string abi_name, unsigned reg,
+                              uint64_t value)
 {
   if (config_print_reg) {
-    fprintf(trace_log, "x%d <- 0x%0*" PRIX64 "\n", reg,
-            static_cast<int>(zxlen / 4), value);
+    if (config_use_abi_names) {
+      fprintf(trace_log, "%s <- 0x%0*" PRIX64 "\n", abi_name,
+              static_cast<int>(zxlen / 4), value);
+    } else {
+      fprintf(trace_log, "x%d <- 0x%0*" PRIX64 "\n", reg,
+              static_cast<int>(zxlen / 4), value);
+    }
   }
   if (config_enable_rvfi) {
     zrvfi_wX(reg, value);
--- a/model/riscv_regs.sail
+++ b/model/riscv_regs.sail
@@ -11,6 +11,101 @@
 register PC       : xlenbits
 register nextPC   : xlenbits
 
+/* mappings for assembly */
+
+mapping reg_abi_name_raw : bits(5) <-> string = {
+    0b00000 <-> "zero",
+    0b00001 <-> "ra",
+    0b00010 <-> "sp",
+    0b00011 <-> "gp",
+    0b00100 <-> "tp",
+    0b00101 <-> "t0",
+    0b00110 <-> "t1",
+    0b00111 <-> "t2",
+    0b01000 <-> "fp",
+    0b01001 <-> "s1",
+    0b01010 <-> "a0",
+    0b01011 <-> "a1",
+    0b01100 <-> "a2",
+    0b01101 <-> "a3",
+    0b01110 <-> "a4",
+    0b01111 <-> "a5",
+    0b10000 <-> "a6",
+    0b10001 <-> "a7",
+    0b10010 <-> "s2",
+    0b10011 <-> "s3",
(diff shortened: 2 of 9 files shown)


c_emulator/riscv_callbacks.cpp
```diff
@@ -68,10 +68,15 @@ unit mem_exception_callback(uint64_t paddr, uint64_t num_of_exception)
   return UNIT;
 }
 
-unit xreg_write_callback(unsigned reg, uint64_t value)
+unit xreg_full_write_callback(const_sail_string reg_name, unsigned reg,
```
Timmmm, line 71: Use `abi_name` instead of `reg_name` (which could easily be `"x12"` etc).

model/riscv_regs.sail
```diff
@@ -11,6 +11,101 @@
...
+    0b11010 <-> "x26",
+    0b11011 <-> "x27",
+    0b11100 <-> "x28",
+    0b11101 <-> "x29",
+    0b11110 <-> "x30",
+    0b11111 <-> "x31"
+}
+
+mapping reg_name : regidx <-> string = {
+  Regidx(i) if get_config_print_abi_names()      <-> reg_abi_name_raw(i),
```
Timmmm, line 87: This isn't exactly printing... I would call it `get_config_use_abi_names()`.

model/riscv_regs.sail
```diff
@@ -11,6 +11,101 @@
...
+  Regidx(i) if not(get_config_print_abi_names()) <-> reg_arch_name_raw(i)
```
Timmmm, line 88:
```suggestion
  Regidx(i) if not(get_config_print_abi_names()) <-> reg_arch_name_raw(i),
```

model/riscv_regs.sail
```diff
@@ -11,6 +11,101 @@
...
+    0b10110 <-> "x22",
+    0b10111 <-> "x23",
+    0b11000 <-> "x24",
+    0b11001 <-> "x25",
+    0b11010 <-> "x26",
+    0b11011 <-> "x27",
+    0b11100 <-> "x28",
+    0b11101 <-> "x29",
+    0b11110 <-> "x30",
+    0b11111 <-> "x31"
```
Timmmm, line 83:
```suggestion
    0b11111 <-> "x31",
```

model/riscv_regs.sail
```diff
@@ -11,6 +11,101 @@
...
+    0b10110 <-> "s6",
+    0b10111 <-> "s7",
+    0b11000 <-> "s8",
+    0b11001 <-> "s9",
+    0b11010 <-> "s10",
+    0b11011 <-> "s11",
+    0b11100 <-> "t3",
+    0b11101 <-> "t4",
+    0b11110 <-> "t5",
+    0b11111 <-> "t6"
```
Timmmm, line 48:
```suggestion
    0b11111 <-> "t6",
```

model/riscv_regs.sail
```diff
@@ -11,6 +11,101 @@
...
+
+mapping creg_name_raw : bits(3) <-> string = {
+  0b000 <-> "s0",
+  0b001 <-> "s1",
+  0b010 <-> "a0",
+  0b011 <-> "a1",
+  0b100 <-> "a2",
+  0b101 <-> "a3",
+  0b110 <-> "a4",
+  0b111 <-> "a5"
```
Timmmm, line 99:
```suggestion
  0b111 <-> "a5",
```

Timmmm (commented):
Seems reasonable to me. Some minor nits.

pmundkur (commented):
Any thoughts of doing the mapping entirely in C++ by directly indexing an array of names using the reg index, which should be faster than the Sail generated C which compares the index one by one until it finds a match?  It would really help when booting an OS with tracing on.

pmundkur:
Yes, `entirely` was too strong; I meant the prints for the `trace_log`.

pmundkur:
> That would require us to duplicate the mapping in both C++ and Sail, which I don't love, but if we think there will be a significant performance degradation then that seems reasonable. 

I think the benefit of the C++ array will be relatively minor given the inefficiency of the disassembly, but probably still worth it given how slow traced OS boots are.

> If we go with the C++ route, do we want the log to always show both versions or keep that as a configurable option?

The registers in trace log should match the disassembly, so `config_use_abi_names` should apply to the C++ as well as the Sail.

pmundkur:
That's fair.  Let's keep it in Sail then.

jordancarlin:
Keeping it in Sail for now. If there ends up being a significant slowdown then we can revisit this later.

pmundkur (approved):

pmundkur (approved):

PR #924, opened 2025-05-07
Lean: fix termination measures

ineol:
Lean does not use termination measures for loops, so we need to guard their definitions.

The problematic measures was introduced by #861.


bacam:
If you don't need the termination measures for loops, could you just ignore them?

ineol:
The issue is that they're rewritten by a Sail pass, so we'd have to add an option to that rewriting pass, which is doable but a bit of work.

bacam:
I think you should be able to easily split that rewrite into two separate rewrites, one for functions and one for loops.

tobiasgrosser (approved):
LGTM. Let's get this merged! :-)

PR #926, opened 2025-05-08
Remove tick_platform() and htif_tick()

Timmmm:
Originally htif_tick() write `htif_tohost = zeros()` - see 6305b6a8c3b219286d1892940f59eddcb67be210 - but now it just prints a log message that we probably don't need.

Removing it means we also don't need to disable self assignment warnings.


c_emulator/CMakeLists.txt
```diff
@@ -34,13 +34,10 @@ foreach (xlen IN ITEMS 32 64)
                 ${EMULATOR_COMMON_SRCS}
             )
             # The generated model is not warnings-clean, silence them.
-            # -Wno-self-assing is needed for `zhtif_tohost = zhtif_tohost`
```
Timmmm, line 37: Didn't catch this spelling mistake @jordancarlin :-D

jordancarlin: Haha. I used codespell and I guess this one isn't in the index.

jordancarlin (approved):
Seems like this function is essentially doing nothing at this point, so seems fine to remove.

pmundkur (approved):
LGTM

PR #928, opened 2025-05-11
Add config option to trap on sfence.vma if virt mem not supported

jordancarlin:
From section 12.2.1 of the priv spec:
> For implementations that make satp.MODE read-only zero (always Bare), attempts to execute an SFENCE.VMA instruction might raise an illegal-instruction exception.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -215,7 +215,8 @@
       "supported": true
     },
     "Svbare": {
-      "supported": true
+      "supported": true,
+      "sfence_vma_illegal_if_svbare_only": true
     },
     "Sv32": {
       "supported": true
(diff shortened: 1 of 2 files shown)


config/default.json
```diff
@@ -203,7 +203,8 @@
       "supported": true
     },
     "Svbare": {
-      "supported": true
+      "supported": true,
+      "trap_sfence_vma_no_virtmem": true
```
Timmmm, line 207: I wonder if it would be conceptually simpler if we just had a top level `sfence_vma_is_illegal` option and then verified it?

jordancarlin: Yeah. I went back and forth on this. It only has an impact when virtual memory isn't supported, but presenting it this way is a little confusing. I'd be fine with a more general `sfence_vma_is_illegal` option, but then we really need to get that config validation system we've been discussing in #860 implemented.

jordancarlin: Updated name to `sfence_vma_illegal_if_svbare_only` per TGMM meeting discussion.

Timmmm (approved):

pmundkur (approved):

PR #931, opened 2025-05-11
Adding support for Zvbc32e (and eventually Zvkgs)

nibrunieAtSi5:
This pull request aims at adding support for a couple of on-going fast track extensions, namely Zvbc32e and Zvkgs.

This is very much a work in progress at this point.


jordancarlin:
Looking good so far. If you add the config option to the default.json config file, it should resolve the CI failure.

nibrunieAtSi5:
> Looking good so far. If you add the config option to the default.json config file, it should resolve the CI failure.

Thank you for the piece of advice @jordancarlin

PR #932, opened 2025-05-11
Make RVFI a runtime option

### LINKED ISSUE #642: Make RVFI a runtime option
I don't think there's any reason to have RVFI as a compile time option. We can always build support in and just enable it at runtime. The benefit is halved compilation time and better developer experience (because you don't have to remember to compile twice to check your code).

This will be easier after #494.

jordancarlin:
Instead of compiling separate models/emulators for RVFI, compile a single one with a run-time flag to enable RVFI.

The RVFI implementation is still in Sail, but it is now a runtime option so an additional simulator doesn't need to be built.

Tested with TestRIG comparing to the previous, compile time RVFI version of the model.


model/riscv_platform.sail
```diff
@@ -390,19 +390,15 @@ function htif_store(Physaddr(paddr), width, data) = {
 }
 
 /* Top-level MMIO dispatch */
-$ifndef RVFI_DII
 function within_mmio_readable forall 'n, 0 < 'n <= max_mem_access . (addr : physaddr, width : int('n)) -> bool =
-  within_clint(addr, width) | (within_htif_readable(addr, width) & 1 <= 'n)
-$else
-function within_mmio_readable forall 'n, 0 < 'n <= max_mem_access . (addr : physaddr, width : int('n)) -> bool = false
-$endif
+  if not(get_config_rvfi())
```
arichardson, line 394: Maybe swap the order here?

model/riscv_platform.sail
```diff
@@ -390,19 +390,15 @@ function htif_store(Physaddr(paddr), width, data) = {
...
+  then within_clint(addr, width) | (within_htif_readable(addr, width) & 1 <= 'n)
+  else false
 
-$ifndef RVFI_DII
 function within_mmio_writable forall 'n, 0 < 'n <= max_mem_access . (addr : physaddr, width : int('n)) -> bool =
-  within_clint(addr, width) | (within_htif_writable(addr, width) & 'n <= 8)
-$else
-function within_mmio_writable forall 'n, 0 < 'n <= max_mem_access . (addr : physaddr, width : int('n)) -> bool = false
-$endif
+  if not(get_config_rvfi())
```
arichardson, line 399: Same here to avoid the not()?

arichardson (approved):
Thanks, this looks great. On extra boolean in the callbacks check should not add any noticeable performance overhead and greatly reduces the complexity.

model/riscv_platform.sail
jordancarlin: Updated

model/riscv_fetch.sail
```diff
@@ -16,36 +16,40 @@ function fetch() -> FetchResult =
...
-                match ext_fetch_check_pc(PC, PC_hi) {
-                  Ext_FetchAddr_Error(e)      => F_Ext_Error(e),
-                  Ext_FetchAddr_OK(use_pc_hi) => {
-                    match translateAddr(use_pc_hi, InstructionFetch()) {
-                      TR_Failure(e, _)     => F_Error(e, PC_hi),
-                      TR_Address(ppchi, _) => {
-                        match mem_read(InstructionFetch(), ppchi, 2, false, false, false) {
-                          Err(e)  => F_Error(e, PC_hi),
-                          Ok(ihi) => F_Base(append(ihi, ilo))
+  if (get_config_rvfi())
```
Timmmm, line 19:
```suggestion
  if get_config_rvfi()
```

model/riscv_step.sail
```diff
@@ -184,7 +184,12 @@ function try_step(step_no : nat, exit_wait : bool) -> bool = {
       // mcountinhibit[IR] or minstretcfg.
       if retired & minstret_increment then minstret = minstret + 1;
 
-      /* for step extensions */
+      // Record the next PC for RVFI.
```
Timmmm, line 187: Maybe add

```
// TODO: Remove this and implement it in C.
```

model/riscv_fetch.sail
```diff
@@ -16,36 +16,40 @@ function fetch() -> FetchResult =
...
+  then rvfi_fetch()
```
Timmmm, line 20: Maybe add:

```
// TODO: Add a parameter to try_step() to allow forcing the fetch result and use that instead.
```

I assume that's where we want to go eventually... RVFI definitely isn't the only verification system that will want to do this.

Timmmm (commented):
This is brilliant!

model/riscv_fetch.sail
jordancarlin: That would be a great improvement longer term. TODO added.

model/riscv_fetch.sail
jordancarlin: Updated.

model/riscv_fetch.sail
```diff
@@ -16,36 +16,41 @@ function fetch() -> FetchResult =
...
-                  Ext_FetchAddr_OK(use_pc_hi) => {
-                    match translateAddr(use_pc_hi, InstructionFetch()) {
-                      TR_Failure(e, _)     => F_Error(e, PC_hi),
-                      TR_Address(ppchi, _) => {
-                        match mem_read(InstructionFetch(), ppchi, 2, false, false, false) {
-                          Err(e)  => F_Error(e, PC_hi),
-                          Ok(ihi) => F_Base(append(ihi, ilo))
+  // TODO: Add a parameter to try_step() to allow forcing the fetch result and use that instead.
+  if get_config_rvfi()
+  then rvfi_fetch()
```
pmundkur, line 21:
```suggestion
  if get_config_rvfi()
  then return rvfi_fetch();
```
This way the else block below doesn't need to be indented inside an `else`, which should make the diff smaller and easier to review?

pmundkur (changes requested):
Great improvement!  LGTM other than a minor nit.

model/riscv_fetch.sail
pmundkur: You might need to open a `{` at the top:
```
function fetch() -> FetchResult = {
```

jordancarlin: Ah. Good point. Not sure why I didn't realize that. Seems to work now.

pmundkur (approved):

PR #935, opened 2025-05-12
Don't require makeinfo to build GMP

### LINKED ISSUE #934: gmp build thinks it requires makeinfo
When I `./build_emulators.sh` on WSL I get

```
/home/tim/workspace/sail-riscv/build/gmp-prefix/src/gmp/missing: line 81: makeinfo: command not found
WARNING: 'makeinfo' is missing on your system.
         You should only need it if you modified a '.texi' file, or
         any other file indirectly affecting the aspect of the manual.
         You might want to install the Texinfo package:
         <http://www.gnu.org/software/texinfo/>
         The spurious makeinfo call might also be the consequence of
         using a buggy 'make' (AIX, DU, IRIX), in which case you might
         want to install GNU make:
         <http://www.gnu.org/software/make/>
make[6]: *** [Makefile:428: /home/tim/workspace/sail-riscv/build/gmp-prefix/src/gmp/doc/gmp.info] Error 127
```

Now, I don't have `makeinfo` but I shouldn't need it. Unfortunately this is due to a confluence of errors:

1. There's no `./configure --without-docs` to just not build the docs in gmp.
2. There doesn't seem to be a sensible `make` target we can use to not build the docs.
3. CMake prior to 3.24 use the archive timestamp when unzipping files so this isn't an issue, because the timestamps mean that Make doesn't try to rebuild the docs.
4. There is an option `DOWNLOAD_EXTRACT_TIMESTAMP TRUE` that fixes it on CMake >3.24 (I tested it), but we can't just add it because we support CMake 3.20.

Possible solutions:

1. Add `DOWNLOAD_EXTRACT_TIMESTAMP TRUE` and change the minimum requirement to 3.24. I think we picked 3.20 because it was in some distro we're using though.
2. Set CMP0135 to OLD, though they say they will remove that eventually.
3. Ask Claude to write a minimal replacement for GMP that just supports what is needed by Sail.

I guess we should go with 2 for now.

Timmmm:
On CMake >= 3.24 with the `DOWNLOAD_GMP` the build would fail if you didn't have `makeinfo` installed. This fixes that.

Fixes #934.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,15 @@
 cmake_minimum_required(VERSION 3.20)
 cmake_policy(VERSION 3.20...3.31)
 
+# CMake 3.24 and newer don't use the mtimes from archives when extracting
+# GMP in ExternalProject_Add(). This reverts that behaviour so that the
+# mtimes in the archive are used. This is necessary because otherwise
+# Make tries to rebuild the documentation that is already built, and
+# this requires `makeinfo` which we don't want to require.
+# TODO: Remove this when we bump the minimum CMake version to 3.24 and
+# instead add DOWNLOAD_EXTRACT_TIMESTAMP TRUE to ExternalProject_Add().
+cmake_policy(SET CMP0135 OLD)
+
 project(sail_riscv)
 
 set(CMAKE_CXX_STANDARD 17)


jordancarlin (approved):

CMakeLists.txt
```diff
@@ -1,6 +1,15 @@
...
 cmake_policy(VERSION 3.20...3.31)
 
+# CMake 3.24 and newer don't use the mtimes from archives when extracting
+# GMP in ExternalProject_Add(). This reverts that behaviour so that the
+# mtimes in the archive are used. This is necessary because otherwise
+# Make tries to rebuild the documentation that is already built, and
+# this requires `makeinfo` which we don't want to require.
+# TODO: Remove this when we bump the minimum CMake version to 3.24 and
+# instead add DOWNLOAD_EXTRACT_TIMESTAMP TRUE to ExternalProject_Add().
+cmake_policy(SET CMP0135 OLD)
```
arichardson, line 11: Is there some kind of --disable-docs configure flag that we could pass instead/in addition? Anyway this looks fine to me

Timmmm: Unfortunately not as far as I can see.

PR #942, opened 2025-05-13
Fix vreg write print output

nadime15:
Currently, `vreg_write_callback()` and `print_lbits_hex()` uses `vlenmax` (65536) as the output length for vector register writes. This fix accounts for the actual VLEN and prints the corresponding number of bytes and not `vlenmax` (65536) bytes.

### DIFF
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -103,6 +103,9 @@ unit vreg_write_callback(unsigned reg, lbits value)
 {
   if (config_print_reg) {
     fprintf(trace_log, "v%d <- ", reg);
+    // TODO: `value` is currently `vlenmax` which can be much greater than
+    // VLEN. In future we will remove `vlenmax`, then we can
+    // remove the `zVLEN / 8` argument here.
     print_lbits_hex(value, zVLEN / 8);
   }
   return UNIT;
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -98,11 +98,11 @@ unit csr_full_read_callback(const_sail_string csr_name, unsigned reg,
   return UNIT;
 }
 
-unit vreg_write_callback(unsigned reg, lbits value)
+unit vreg_write_callback(unsigned reg, lbits value, uint64_t vlen)
 {
   if (config_print_reg) {
     fprintf(trace_log, "v%d <- ", reg);
-    print_lbits_hex(value);
+    print_lbits_hex(value, vlen / 8);
   }
   return UNIT;
 }
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -1,6 +1,7 @@
 #include "riscv_callbacks.h"
 #include "riscv_config.h"
 #include "riscv_platform_impl.h"
+#include "riscv_sail.h"
 #include <stdlib.h>
 #include <vector>
 #include <inttypes.h>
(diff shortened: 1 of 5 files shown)


jordancarlin (commented):
Nice improvement. Could this just use `vlenb` to simplify the logic?

c_emulator/riscv_callbacks.h
```diff
@@ -18,7 +18,7 @@ unit csr_full_write_callback(const_sail_string csr_name, unsigned reg,
                              uint64_t value);
 unit csr_full_read_callback(const_sail_string csr_name, unsigned reg,
                             uint64_t value);
-unit vreg_write_callback(unsigned reg, lbits value);
+unit vreg_write_callback(unsigned reg, lbits value, uint64_t vlen);
```
jordancarlin, line 21:
```suggestion
unit vreg_write_callback(unsigned reg, lbits value, uint64_t vlenb);
```
To make it clear that this is vlen in bytes for any other callback implementations.

jordancarlin (changes requested):

c_emulator/riscv_callbacks.h
jordancarlin: Looks like it was changed in `riscv_callbacks.cpp` but not `riscv_callbacks.h`.

nadime15: Ups, my bad, thanks!

jordancarlin (approved):
LGTM

pmundkur (approved):

Timmmm:
If this isn't urgent maybe we should wait until the new compiler release, then we can remove vlenmax and make VLEN a configure time option like XLEN and FLEN and this change won't be necessary at all.

Or we can revert it later.

nadime15:
I would merge it now, otherwise you basically cant run any vector instructions because **each** vreg write will print 65536 Bytes and the system becomes unresponsive (atleast mine).

c_emulator/riscv_sail.h
```diff
@@ -68,6 +68,8 @@ extern struct zMcause zmcause, zscause;
 
 extern mach_bits zminstret;
 
+extern mach_bits zVLEN;
```
jordancarlin, line 71: Move this up to be with `zxlen` and `zflen`.

jordancarlin (changes requested):
Minor nit. Otherwise LGTM.

jordancarlin (approved):

c_emulator/riscv_callbacks.cpp
```diff
@@ -102,7 +103,7 @@ unit vreg_write_callback(unsigned reg, lbits value)
 {
   if (config_print_reg) {
     fprintf(trace_log, "v%d <- ", reg);
-    print_lbits_hex(value);
+    print_lbits_hex(value, zVLEN / 8);
```
Timmmm, line 106: Ok let's just add a TODO:

```suggestion
    // TODO: `value` is currently `vlenmax` which can be much greater than
    // VLEN. In future we will remove `vlenmax`, then we can
    // remove the `zVLEN / 8` argument here.
    print_lbits_hex(value, zVLEN / 8);
```

c_emulator/riscv_callbacks.cpp
```diff
@@ -102,7 +103,10 @@ unit vreg_write_callback(unsigned reg, lbits value)
 {
   if (config_print_reg) {
     fprintf(trace_log, "v%d <- ", reg);
-    print_lbits_hex(value);
+    // TODO: `value` is currently `vlenmax` which can be much greater than
```
pmundkur, line 106:
```suggestion
    // TODO: the width of `value` is currently `vlenmax` bits which can be much greater than
```

pmundkur (changes requested):
just a minor nit

c_emulator/riscv_callbacks.cpp
```diff
@@ -102,7 +103,10 @@ unit vreg_write_callback(unsigned reg, lbits value)
 {
   if (config_print_reg) {
     fprintf(trace_log, "v%d <- ", reg);
-    print_lbits_hex(value);
+    // TODO: the witdh of `value` is currently `vlenmax` bits which can be much
```
pmundkur, line 106:
```suggestion
    // TODO: the width of `value` is currently `vlenmax` bits which can be much
```

pmundkur (changes requested):

pmundkur (approved):

Timmmm (approved):

PR #943, opened 2025-05-13
Fix x and f register write print output  

nadime15:
Only print 32 bits when XLEN/FLEN is 32.

### DIFF
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -62,10 +62,13 @@ unit mem_exception_callback(uint64_t paddr, uint64_t num_of_exception)
   return UNIT;
 }
 
-unit xreg_write_callback(unsigned reg, uint64_t value)
+unit xreg_write_callback(unsigned reg, uint64_t value, uint64_t xlen)
 {
   if (config_print_reg) {
-    fprintf(trace_log, "x%d <- 0x%016" PRIX64 "\n", reg, value);
+    if (xlen == 32)
+      fprintf(trace_log, "x%d <- 0x%08" PRIX32 "\n", reg, (uint32_t)value);
+    else
+      fprintf(trace_log, "x%d <- 0x%016" PRIX64 "\n", reg, value);
   }
   if (config_enable_rvfi) {
     zrvfi_wX(reg, value);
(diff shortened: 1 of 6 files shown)


c_emulator/riscv_callbacks.cpp
```diff
@@ -58,22 +58,29 @@ unit mem_exception_callback(uint64_t paddr, uint64_t num_of_exception)
   return UNIT;
 }
 
-unit xreg_write_callback(unsigned reg, uint64_t value)
+unit xreg_write_callback(unsigned reg, uint64_t value, uint64_t xlen)
 {
   if (config_print_reg) {
-    fprintf(trace_log, "x%d <- 0x%016" PRIX64 "\n", reg, value);
+    if (xlen == 32)
```
jordancarlin, line 64: Could this just use `zxlen` which is already accessed several other places in the c emulator? That way we're not passing additional arguments? The same thing could be done for flen and glen (in #942).

jordancarlin (commented):
Thanks for catching all of these. I'm realizing I think I did all of my testing on rv64d when looking at the new logs, so I missed this issue.

c_emulator/riscv_callbacks.cpp
nadime15: Sure! How does this work with flen and vlen? 

Do I have to define something like:

```
extern mach_bits zflen
extern mach_bits zvlenb

```

Timmmm: Yeah - check the generated code to see what type they are, then copy it. In future we will use a header generated by Sail so you don't need to do this, but we're waiting for a new compiler release.

jordancarlin: For `flen` that should work fine. `vlenb` isn't a type, so it would probably be easiest to use `zvlen` and just divide by 8 to get to bytes. Technically you could call the `zget_vlenb()` function, but that would probably be more complicated.

c_emulator/riscv_callbacks.cpp
```diff
@@ -58,22 +58,29 @@ unit mem_exception_callback(uint64_t paddr, uint64_t num_of_exception)
...
+      fprintf(trace_log, "x%d <- 0x%08" PRIX32 "\n", reg, (uint32_t)value);
+    else
+      fprintf(trace_log, "x%d <- 0x%016" PRIX64 "\n", reg, value);
```
jrtc27, line 67:
```suggestion
    fprintf(trace_log, "x%d <- 0x%0*" PRIX64 "\n", reg, (int)xlen, value);
```
is a bit nicer, no?

arichardson (approved):

jordancarlin (approved):

pmundkur (approved):

jordancarlin:
Not necessary to change, but I just realized we also have `xlen_bytes` as a variable you could use without needing to do the /8. And the same for flen.

jordancarlin:
Ah wait. Never mind. It’s the number of hex digits, which would be twice that. No need to change in that case.

c_emulator/riscv_callbacks.cpp
```diff
@@ -73,7 +74,8 @@ unit freg_write_callback(unsigned reg, uint64_t value)
 {
   // TODO: will only print bits; should we print in floating point format?
   if (config_print_reg) {
-    fprintf(trace_log, "f%d <- 0x%016" PRIX64 "\n", reg, value);
+    // TODO: once the "Q" extension is supported change PRIX64 to PRIX128
```
Timmmm, line 77: Probably not - if FLEN is >64 bits the callback will be `sail_int` or whatever instead of `uint64_t`. It's going to get awkward for sure...

PR #945, opened 2025-05-14
Fix memory write/read print output

nadime15:
This should be the last PR related to printing. I added the access type to `mem_write_callback()`.

### DIFF
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -26,8 +26,7 @@ unit mem_write_callback(const char *type, uint64_t paddr, uint64_t width,
 {
   if (config_print_mem_access) {
     fprintf(trace_log, "mem[%s,0x%0*" PRIX64 "] <- 0x", type,
-            (int)zphysaddrbits_len / 4 + ((int)zphysaddrbits_len % 4 != 0),
-            paddr);
+            static_cast<int>((zphysaddrbits_len + 3) / 4), paddr);
     print_lbits_hex(value, width);
   }
   if (config_enable_rvfi) {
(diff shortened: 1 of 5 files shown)


pmundkur (approved):

c_emulator/riscv_callbacks.cpp
```diff
@@ -20,10 +21,12 @@ void print_lbits_hex(lbits val, int length = 0)
 
 // Implementations of default callbacks for trace printing and RVFI.
 // The model assumes that these functions do not change the state of the model.
-unit mem_write_callback(uint64_t paddr, uint64_t width, lbits value)
+unit mem_write_callback(const char *type, uint64_t paddr, uint64_t width,
+                        lbits value)
 {
   if (config_print_mem_access) {
-    fprintf(trace_log, "mem[0x%016" PRIX64 "] <- 0x", paddr);
+    fprintf(trace_log, "mem[%s,0x%0*" PRIX64 "] <- 0x", type,
```
jordancarlin, line 28: Could you just use `physaddrbits_len`?

nadime15: You mean instead of doing `zxlen == 64 ? 16 : 9`? If so, no we cant because we want to have the number of hex digits and `(int)zphysaddrbits_len / 4` wont work either because in case of xlen == 32 the number of physical bits is 34 and not 32 and will round down to 8.

jordancarlin: Ah. Good point about the rounding issue. I guess technically we could do something like  `(int)zphysaddrbits_len/4 + ((int)zphysaddrbits_len % 4 != 0)`, but that's probably more complicated.

jordancarlin (approved):

c_emulator/riscv_callbacks.cpp
nadime15: I added it. I think its ok, at least we got rid of the hardcoded numbers.

c_emulator/riscv_callbacks.cpp
```diff
@@ -20,10 +21,13 @@ void print_lbits_hex(lbits val, int length = 0)
...
+            (int)zphysaddrbits_len / 4 + ((int)zphysaddrbits_len % 4 != 0),
```
Timmmm, line 29: This is very implicit and not obviously not a mistake. Lets do this:

```
int hex_chars = (zphysaddrbits_len + 3) / 4;
```

Timmmm (changes requested):
I have to leave a comment to request changes....

Timmmm (approved):
I think at some point we'll have to change `const char* type` to something structural rather than stringy because this is intended to be used for things like logging to machine readable formats.... however `mem_read_callback()` already has it and I guess you can always omit that information from the machine readable log, so lgtm, thanks!

PR #946, opened 2025-05-14
Add mstatush to csr_name_map

nadime15:
Add missing `mstatush` to the `csr_name_map` mapping. This was causing incorrect CSR log messages.


jordancarlin (approved):
Surprised we didn’t have this already.

pmundkur (approved):

Timmmm:
It's odd that this gave incorrect results rather than a match failure. We should investigate that.

Issue #948, opened 2025-05-14
Test with multiple versions of CMake


jordancarlin:
Between #934 and #947 recently, it seems like we should be testing compilation on multiple versions of CMake. Probably just the oldest version that we support and the latest release.

PR #949, opened 2025-05-14
Prevent build failures with CMake versions < 3.24.

### LINKED ISSUE #947: Policy "CMP0135" is not known to this version of CMake for versions before 3.24
While #935 fixed #934, it seems to have inadvertently broken older versions of CMake. With CMake 3.22 I now get the following error message:
```
CMake Error at CMakeLists.txt:11 (cmake_policy):
  Policy "CMP0135" is not known to this version of CMake.
```

CMake then terminates and there is no way to build the model. Seems like we need a way to set this policy only on newer versions of CMake.

pmundkur:
Check if CMP0135 is supported by the version of CMake being used before setting it. If it isn't supported, the old behavior is used anyway.

Fixes #947.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,9 @@ cmake_policy(VERSION 3.20...3.31)
 # this requires `makeinfo` which we don't want to require.
 # TODO: Remove this when we bump the minimum CMake version to 3.24 and
 # instead add DOWNLOAD_EXTRACT_TIMESTAMP TRUE to ExternalProject_Add().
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):
Tested this with CMake 3.22 in #950 and can confirm that it works.

pmundkur:
I just saw #947 and #948 after pushing this.

jordancarlin:
> I just saw #947 and #948 after pushing this.

Yeah. Seems like we both stumbled across this at the same time. Let's go ahead and merge this version.

PR #950, opened 2025-05-14
Only set CMP0135 if it is supported

jordancarlin:
Check if CMP0135 is supported by the version of CMake being used before setting it. If it isn't supported the old behavior is used anyway.


jordancarlin:
Seems like @pmundkur beat me by a few seconds. Closing this in favor of #949.

PR #953, opened 2025-05-15
Fix xepc interaction with compressed extensions

jordancarlin:
When any compressed extension is supported (whether enabled or not), `xepc[1]` can be written.
When compressed instructions are disabled, `xepc[1]` is read as 0.

Both of these were checked based on `misa[C]`, but compressed instructions can be supported without setting `misa[C]` (e.g. an `RV32IF_Zca` hart).


Timmmm (approved):
Looks correct to me. I wish the spec was less vague about this stuff!

pmundkur (approved):
Good catch.

PR #573, opened 2024-10-04
Add Debug Native Trigger Support

YazanHussnain-10x:
## Debug native trigger support

This PR adds the debug [native trigger](https://github.com/riscv-admin/dev-partners/issues/8) support in sail RISC-V golden model. [RISC-V Debug Specification](https://github.com/riscv/riscv-debug-spec)

| **Parameter**                             | **Details**                                                                                         |
|-------------------------------------------|-----------------------------------------------------------------------------------------------------|
| **CSRs Implemented**                      | Trigger Select, Trigger Data 1, Trigger Data 2, Trigger Data 3, Trigger Info, Trigger Control        |
| **No. of Supported Triggers**             | 7                                                                                                   |
| **Supported Actions**                     | 0 - Currently Sail-RISC-V does not have support for debug mode, therefore, only breakpoint exception action is supported |
| **Value of dmode**                        | 0 - Both Debug and M-mode can write the `tdata` registers                                            |
| **Supported Trigger Types**               | Instruction Count, Interrupt Trigger, Exception Trigger, Match Control Type 6 Trigger, Disabled      |
| **Re-entrancy Solution**                  | Both solutions supported - Use `--reent-opt1` flag to enable the second solution. Only one solution can be enabled at a time |
| **Supported Match Options in MControl6**  | 0: Equal, 1: NAPOT, 2: GE, 3: LT                                                                    |

### Implemented Functions

- **tdata1_write**: Writes the `tdata1` register at the selected `tselect`.
- **legalize_tdata1**: Legalizes the values in `tdata1` for the following triggers: `icount`, `itrigger`, `etrigger`, and `mcontrol6`.
- **tdata1_read**: Reads the `tdata1` register at the selected `tselect`.
- **tdata2_write**: Writes the `tdata2` register at the selected `tselect`.
- **tdata2_read**: Reads the `tdata2` register at the selected `tselect`.
- **tdata3_write**: Writes the `tdata3` register at the selected `tselect`.
- **tdata3_read**: Reads the `tdata3` register at the selected `tselect`.
- **legalize_tcontrol**: Legalizes the `tcontrol` register: only `mte` and `mpte` fields are writable.
- **TriggerFireMatch**: Checks re-entrancy and returns `true` if trigger matching and firing are valid.
- **icount_trigger_match**: Checks for a match in the instruction count trigger.
- **IETrigger_match**: Checks for a match in the interrupt and exception triggers.
[...]

PR #954, opened 2025-05-15
WIP: Implement the Sdext debug mode extension.

pmundkur:
The `try_step` API was changed to reflect the current debug mode state and to provide a mechanism for the external harness to request a change in that state.

This does not provide a model-external harness to connect to external debugger infrastructure; that is left for the future. For now, two callbacks indicate entry and exit from debug mode, and the `try_step()` API has been augmented to allow requests to the model to halt and resume (i.e. enter and exit debug mode).

Hook functions have been defined for behavior that is UNSPECIFIED in the debug specification or depends on the debugger's memory map (in `riscv_sdext_hook_insts`).

StopCount is implemented for the single-hart case, with a note marking the change needed for multi-hart emulation.

This moves `should_inc_minstret` out of an extension-specific file into `riscv_platform` in the core model, which fixes 783.

Some notes have been added under `doc/` about the sdext implementation.


pmundkur:
Marked as ready to review.  This may provide a better base to implement Sdtrig than what #573 had.

model/riscv_callbacks.sail
```diff
@@ -31,6 +31,12 @@ function csr_full_read_callback(_) = ()
 val trap_callback = pure {c: "trap_callback"} : unit -> unit
 function trap_callback(_) = ()
 
+val debug_entry_callback = pure {c: "debug_entry_callback"} : unit -> unit
+function debug_entry_callback(_) = ()
+
+val debug_exit_callback = pure {c: "debug_exit_callback"} : unit -> unit
+function debug_exit_callback(_) = ()
+
```
nadime15: Thats probably unrelated to this PR more about the API itself, but seeing these functions (which I assume are placeholders meant to communicate with the Debug Module) got me thinking again. The more I deal with the Debug Module, the more it seems like it would actually make sense to implement it in Sail as well.

The DM exposes a bunch of (mostly optional) registers, each with tons of bits and configuration options. Sails support for bitfields would come in handy for modeling all that.

From what I understand so far, the data exchange between the Debug Module and the hart mainly involves these (1-bit) signals:

> allnonexistent, anynonexistent, allunavail, anyunavail, allrunning, anyrunning, allhalted, and anyhalted.

They are observed by the DM but actually driven by the hardware.

On top of that, we have abstract commands (optionally access memory and quick access too), which operate in 32-bit chunks but the argument itself can go up 128 bits (the argument is split into four 32-bit chunks for communications). 

So we need to be able to transmit/receive between 1-4 32-bit chunks

pmundkur: The DM<->hart interface seems relatively straightforward to do in Sail. But the DM needs to connect to the outside world somehow.   Right now, the Sail model has a pretty narrow interface to the external C++ harness.  The DM would add another interface.  It might be worth thinking through how big that is, and how it would be driven.

model/riscv_insts_base.sail
```diff
@@ -64,6 +69,8 @@ function clause execute (JAL(imm, rd)) = {
   match ext_control_check_pc(target) {
     Ext_ControlAddr_Error(e) => Ext_ControlAddr_Check_Failure(e),
     Ext_ControlAddr_OK(target) => {
+      if   in_debug_abstract_command() & ~(debug_is_legal_ct_target(target))
+      then return Illegal_Instruction();
```
tomaird: I think these illegal instruction checks should come before the `match ext_control_check_pc(target)`. In general an illegal instruction will have a higher priority than any exception arising from actually executing the instruction. Any extension that places additional checks on the address checks (e.g. CHERI), using `ext_control_check_pc`, will require the instruction to be actually legal before those checks are done.

pmundkur: Hmm, the `ext_control_check_pc(target)` could return a transformed `target`, so `debug_is_legal_ct_target(target)` is checking the transformed `target`.  Changing the order would require `debug_is_legal_ct_target(target)` work on the value before transformation, i.e. it would need to be aware of any out-of-tree extension (e.g. CHERI currently) that can transform the target in an unknown way, or am I missing something?

c_emulator/riscv_sim.cpp
```diff
@@ -581,7 +581,12 @@ void run_sail(void)
       sail_int sail_step;
       CREATE(sail_int)(&sail_step);
       CONVERT_OF(sail_int, mach_int)(&sail_step, step_no);
-      is_waiting = ztry_step(sail_step, exit_wait);
+
+      // If the model is in Debug mode, request an exit, else request
+      // no change.
+      enum zdebug_request dr
```
```suggestion
      zdebug_request dr
```

Don't need the `enum` in C++.

model/riscv_pmp_control.sail
```diff
@@ -96,6 +96,15 @@ function pmpCheck forall 'n, 0 < 'n <= max_mem_access . (
   priv : Privilege,
 ) -> option(ExceptionType) = {
 
+  // In debug mode, allow all access to debug memory range.
```
Timmmm: Is this normal? Don't most implementations allow access to all memory?

model/riscv_insts_base.sail
```diff
@@ -25,11 +25,16 @@ mapping clause encdec = UTYPE(imm, rd, op)
...
   let off : xlenbits = sign_extend(imm @ 0x000);
-  X(rd) = match op {
-    LUI   => off,
-    AUIPC => get_arch_pc() + off
-  };
-  RETIRE_SUCCESS
+  match op {
+    LUI   => { X(rd) = off; RETIRE_SUCCESS },
+    AUIPC => {
+      if   in_debug_abstract_command() & ~(debug_is_pc_instr_legal())
```
Timmmm: We did this a bit differently by wrapping the entire `execute` function in another function like this:

```
// If true then control transfer functions are illegal in debug mode.
function debugIllegalCtrlTransferInstr() -> bool = true

// Which instructions are illegal in debug mode?
function illegalInDebugMode(ast : ast) -> bool = match ast {
  RISCV_JAL(_, _) => debugIllegalCtrlTransferInstr(),
  RISCV_JALR(_, _, _) => debugIllegalCtrlTransferInstr(),
  BTYPE(_, _, _, _) => debugIllegalCtrlTransferInstr(),
  ECALL() => debugIllegalCtrlTransferInstr(),
  MRET() => debugIllegalCtrlTransferInstr(),
  SRET() => debugIllegalCtrlTransferInstr(),
  UTYPE(_, _, RISCV_AUIPC) => debugIllegalCtrlTransferInstr(),
  C_BNEZ(_, _) => debugIllegalCtrlTransferInstr(),
  C_JAL(_) => debugIllegalCtrlTransferInstr(),
  C_J(_) => debugIllegalCtrlTransferInstr(),
  C_BEQZ(_,_) => debugIllegalCtrlTransferInstr(),
  C_JR(_) => debugIllegalCtrlTransferInstr(),
  C_JALR(_) => debugIllegalCtrlTransferInstr(),
  _ => false,
}

// Execute an instruction, as long as it isn't one that is illegal in debug mode.
function execute_instruction(ast : ast) -> Retired = {
    if debug_mode_active & illegalInDebugMode(ast) then {
        handle_illegal();
        RETIRE_FAIL
    } else execute(ast)
}
```

I think that's probably cleaner.

model/riscv_sdext_control.sail
```diff
@@ -0,0 +1,81 @@
...
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+/* Debug mode extension Sdext: control */
+
+// The specification denotes a hart's entry into Debug mode as the hart "halting".
+function enter_debug_mode(cause : debug_entry_cause, extcause : bits(3), pc : xlenbits) -> unit = {
+  assert(~(debug_mode_active), "enter_debug_mode: already in Debug mode.");
```
Timmmm: Btw I normally prefer `not()` for bools. A bit clearer I think.

model/riscv_sdext_regs.sail
```diff
@@ -0,0 +1,167 @@
...
+function init_sdext() -> unit =
+  if hartSupports(Ext_Sdext) then {
+    debug_mem_base = to_bits_checked(config extensions.Sdext.mem_base : nat);
+    debug_trap_vector = to_bits_checked(config extensions.Sdext.trap_vector : int);
+  }
+
+// Other configuration elements.
+let writable_dcsr_stepie : bool =
+  if   hartSupports(Ext_Sdext)
+  then config extensions.Sdext.writable_stepie
```
Timmmm: `hartSupports(Ext_Sdext) & config extensions.Sdext.writable_stepie`

model/riscv_sdext_regs.sail
```diff
@@ -0,0 +1,167 @@
...
+  else false
+let preset_dcsr_stopcount : bool =
```
Timmmm: Minor nit but this reads kind of weirdly - I'd go for `dcsr_stopcount_exists`, `dcsr_stopcount_writable`.

model/postlude/step.sail
```diff
@@ -135,16 +136,55 @@ function wait_is_nop(wr : WaitReason) -> bool =
...
+// buffer).
+// `.debug_cmd_error` indicates whether an exception was encountered
+// when executing an abstract command.
+
+struct step_result = {
+  in_wait         : bool,
+  in_debug_mode   : bool,
+  in_debug_cmd    : bool,
+  debug_cmd_error : bool,
+}
```
Timmmm, line 176: It's not very clear what the valid combinations here are. Could we use enum/unions to [make invalid states unrepresentable](https://geeklaunch.io/blog/make-invalid-states-unrepresentable/)?

model/extensions/Zicntr/zicntr_control.sail
```diff
@@ -23,12 +23,20 @@ function clause is_CSR_defined(0xC81) = currentlyEnabled(Ext_Zicntr) & (xlen ==
 function clause is_CSR_defined(0xC82) = currentlyEnabled(Ext_Zicntr) & (xlen == 32) // instreth
 
 function clause read_CSR(0xC00) = mcycle[(xlen - 1) .. 0]
-function clause read_CSR(0xC01) = mtime[(xlen - 1) .. 0]
-function clause read_CSR(0xC02) = minstret[(xlen - 1) .. 0]
 function clause read_CSR(0xC80 if xlen == 32) = mcycle[63 .. 32]
-function clause read_CSR(0xC81 if xlen == 32) = mtime[63 .. 32]
+function clause read_CSR(0xC02) = minstret[(xlen - 1) .. 0]
```
Timmmm, line 26: Did you mean to move this?

model/riscv_pmp_control.sail
pmundkur: This is for memory associated with the debug module, and handles the following in the debug spec:
`The PMP must not disallow fetches, loads, or stores in the address range associated with the Debug Module when the hart is in Debug Mode, regardless of how the PMP is configured.`

model/postlude/step.sail
pmundkur: There is some initial documentation of the combinations in `doc/notes_Debug_mode`, which could no doubt be improved.  Since this info crosses over to C as the return value of `try_step()`, I left it as a simple struct, since dealing with Sail unions in C is a bit painful.

model/extensions/Zicntr/zicntr_control.sail
pmundkur: yes, it keeps the `mcycle[h]` and `minstret[h]` clauses together, but also true, it is unrelated to `Sdext`.

model/riscv_pmp_control.sail
Timmmm: I think that only applies if you happen to use that implementation though, which is really just an example that you might use to implement this on real hardware. I don't think we need to use that rather convoluted implementation because Sail has way fewer constraints than real hardware.

> Below are two possible implementations. A designer could choose one, mix and match, or come up with their own design.

model/postlude/step.sail
Timmmm: Yeah fair enough.

model/extensions/Zicntr/zicntr_control.sail
Timmmm: That's fine, just checking it was intentional!

model/riscv_pmp_control.sail
pmundkur: It's probably best to keep this implementation-agnostic for now.   Let's see what is needed by the forthcoming Debug Module from @nadime15.

model/riscv_insts_base.sail
pmundkur: Hmm, that's an interesting approach.  Whether the control transfer is illegal does depend on the destination though, so this might need to be tweaked.  Also, `mret,sret,ecall` are  `UNSPECIFIED` in their behaviour, which is different from the rest.  I'm leaning towards keeping the current hooks for now.

model/riscv_insts_base.sail
tomaird: It is a bit tricky because `ext_control_check_pc` can return both a modified target _and_ an exception (that is most likely lower priority than `Illegal Instruction`)
Are there any examples of `ext_control_check_pc(target)` transforming a target in any existing extensions? i.e. are we worrying about a problem that doesn't currently exist? 
CHERI uses `ext_control_check_pc` to do additional checks on the PC, which may raise a lower priority exception [(see here)](https://github.com/CHERI-Alliance/sail-riscv/blob/codasip/model/cheri/cheri_addr_checks.sail#L100C10-L100C30) , but with your current code, the CHERI exception would take priority over any illegal instruction exception raised from `debug_is_legal_ct_target` failing.
I also think it might be common to want to make _all_ control transfer instructions illegal in debug mode, in which case the target is irrelevant anyway.
A simple alternative would be to have an additional flag that indicates _all_ CT instructions are illegal in debug, which is checked at the start of these instructions. e.g. `all_ct_illegal_in_debug`. Or do something similar to what @Timmmm suggested below (but possibly in addition to what you have here)

model/riscv_insts_base.sail
Timmmm: If they are unspecified we can just add `config`s for them I think.

I think until we actually have a program buffer it probably makes sense to just make all these illegal. What do you think @tomaird ?

I don't think it makes any sense to *only* have special behaviour for AUIPC anyway. If we do something for that we should do it for all of the control transfer instructions or instructions that depend on PC (e.g. JAL).

tomaird: Technically the debug spec does allow you to specify differently the behaviour for jumps to inside or outside the program buffer:

> 10. All control transfer instructions may act as illegal instructions if their destination is **in** the Program Buffer. If one such instruction acts as an illegal instruction, all such instructions must act as illegal instructions.
> 11. All control transfer instructions may act as illegal instructions if their destination is **outside** the Program Buffer. If one such instruction acts as an illegal instruction, all such instructions must act as illegal instructions.

But I agree that in general it makes more sense here to treat them all the same regardless of target. 
There's so much "unspecified" and "implementation defined" in the debug spec, that I think it's prudent in certain places to just pick an implementation for the Sail model, and people can add other options later if needed, rather than trying to cater for every single combination of options.

For the UNSPECIFIED ones (mret,sret,ecall) I do agree with @pmundkur that we should leave those as hooks. It's reasonable to either execute them as NOPs or as illegal instructions, on a case-by-case basis. (although I personally would combine mret and sret into xret, but leave ecall separate).

For the PC-relative instructions and control transfer instructions (auipc, jalr, branches, etc.) - I think it makes sense to treat them all the same. So either have a flag to make them illegal in debug mode, or have a single hook (as for mret etc.) that could default to raising an exception.

I'm not sure how you could really execute a PC-relative instruction in debug mode when there's not a PC, so the only reasonably options there are executing as a NOP or as an illegal instruction.

Timmmm: > I'm not sure how you could really execute a PC-relative instruction in debug mode when there's not a PC

I think they're imagining that you might do a loop within the program buffer using `bne` etc. Those instructions are relative to the address of the instruction, which you could imagine to be well-defined as long as it stays within the program buffer (which is probably why they distinguish those two cases).

I don't know how you would interpret a jump to *outside* the program buffer though. It's very vague.

model/riscv_sdext_control.sail
```diff
@@ -0,0 +1,81 @@
...
+
+// This is called from the external harness, so there is no assert.
+function clear_debug_abstract_command_error() -> unit = {
+  debug_abstract_command_error = false
+}
+
+function reset_sdext() -> unit = {
+  init_sdext();
+  dcsr = reset_Dcsr();
+}
```
nadime15: I think we need to reset dpc to the reset vector (or just dpc = pc) when the hart experiences a reset. Otherwise, when leaving debug mode, although pc was correctly set to the reset vector during reset(), it gets overwritten by the stale dpc value from before the reset. This causes execution to resume from the wrong location.

(This depends on whether `enter_debug_mode` or `exit_debug_mode` is called first after a reset.)

I am not sure if this happens because I have not implemented the correct event order yet, or if dpc should always be set to the reset vector regardless but I think we should play it safe and set dpc to the reset vector as well ~~because dpc is set to the next instruction to be executed.~~

nadime15: I think this answers the question: https://github.com/riscv/riscv-debug-spec/issues/1094#issuecomment-2533607680 

It turned out a bit more complicated than I expected. It depends on whether we jump straight back into halt after resetting the hart and setting halt-on-reset to 1. But even without halt-on-reset, section 3.2 says

> "Otherwise, if the hart was initially running it will execute normally (running state) and if the hart was initially halted it should now be running but may be halted."

But even if we continue execution immediately after reset, dpc should be set to the reset vector (?). Its confusing

pmundkur: If the intent is to `halt-on-reset`, I think the sequence would be to `reset` the hart, which should call `reset_sdext()`,  and then `enter_debug_mode`, which should set the `dpc` (See sections 4.6 and 4.7). `exit_debug_mode()` should not be called without an earlier `enter_debug_mode()`.

nadime15: Yeah, I think you are right, I called `exit_debug_mode()` after the reset, which is causing the issue.

 The part I dont fully understand is what happens if the hart is in debug mode and then experiences a reset. For example, the hart could have been halted for single-stepping, or it could have been generally halted before the reset occurs.

But if the core is already in debug mode when the reset happens, whats the correct behavior? Should I call `exit_debug_mode()` before the reset, and then after the reset just switch to normal execution? And if halt-on-reset (or a similar flag) is set to 1, should I switch back to debug mode after the reset?

pmundkur: If the hart is reset, it loses its earlier state, including that it was in debug mode before reset.  It's not clear to me where the reset is coming from.  Let's talk ...

tomaird:
@nadime15 @pmundkur @Timmmm 

I see the changes to the `ext_*` functions have been merged. Is there anything else blocking this PR from progressing? This is quite a big PR and blocks other things being added, such as Sdtrig, so it would be good to keep it moving forward.

nadime15:
We discussed debug yesterday in our weekly meeting and how to proceed. Our current work is here (https://github.com/nadime15/sail-riscv/tree/debug_all) and includes all the plumbing for OpenOCD (bit bang protocol), the debug module, and the sdext extension. I am currently rebasing the branch and will slice it into smaller pieces. I will update this PR as soon as I am ready, since there are a few things I still want to change/have not been added (dealing with Zicfilp). I hope to have the updated PR by the end of the week.

Feel free to check out the implementation in the meantime. It already works, and you can connect GDB, etc.. If you need guidance on how to get started just let me know @tomaird.

nadime15:
I have updated this PR and added a few missing pieces. It’s not perfect and still needs some work, but the basic functionality has been added and it works! Let me know what you think.

This PR obviously won’t work without #1121 and #1414. It would be great to get a review and get #1121 (I need to rebase it again) merged as soon as possible, otherwise, constantly rebasing is a huge pain.

Additionally, this won’t work until issue https://github.com/rems-project/sail/issues/1555 is fixed.

PR #369, opened 2023-12-07
Smcsrindpr

adlr:
Draft PR for Smcdeleg/Ssccfg support. Includes Smcntrpmf, S*csrind as well.

PR #956, opened 2025-05-18
Add support for smcsrind/sscsrind

KotorinMinami:
To avoid complexity in reviewing multiple extensions within a single PR like #369 and #491, this extension is submitted separately as a foundation for subsequent extension implementations.

### DIFF
--- a/model/extensions/Csrind/csrind_regs.sail
+++ b/model/extensions/Csrind/csrind_regs.sail
@@ -1,0 +1,193 @@
+// =======================================================================================
+// This Sail RISC-V architecture model, comprising all files and
+// directories except where otherwise noted is subject the BSD
+// two-clause license in the LICENSE file.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+// =======================================================================================
+
+function clause currentlyEnabled(Ext_Smcsrind) = hartSupports(Ext_Smcsrind) & currentlyEnabled(Ext_Zicsr)
+function clause currentlyEnabled(Ext_Sscsrind) = (hartSupports(Ext_Sscsrind) | currentlyEnabled(Ext_Smcsrind)) & currentlyEnabled(Ext_Zicsr) & currentlyEnabled(Ext_S)
+
+bitfield Iselect : xlenbits = {
+  value : 11 .. 0,
+}
+
+register miselect  : Iselect
+register siselect  : Iselect
+register vsiselect : Iselect
+
+enum IregType = { Mireg, Sireg, Vsireg }
+
+function legalize_iselect(v : xlenbits) -> Iselect = Mk_Iselect(zero_extend(v[11 .. 0]))
+
+// True when sel maps to a register reachable via *ireg*.
+// NOTE: no dependent extension uses the window yet, so this is always false.
+function indirect_reg_implemented(_ireg : IregType, _sel : Iselect, _i : range(1, 6)) -> bool =
+  false
+
+// Access the selected register, dispatched on sel. Unreachable until a select
+// is implemented. set_ireg returns Err(()) for an illegal access.
+function get_ireg(_ireg : IregType, _sel : Iselect, _i : range(1, 6)) -> xlenbits =
+  zeros()
+
+function set_ireg(_ireg : IregType, _sel : Iselect, _i : range(1, 6), _value : xlenbits) -> result(xlenbits, unit) =
+  Err(())
+
+function read_siselect() -> xlenbits = if inVirtualPrivilege() then vsiselect.bits else siselect.bits
+
+function write_siselect(value : xlenbits) -> result(xlenbits, unit) =
+  if inVirtualPrivilege()
+  then { vsiselect = legalize_iselect(value); Ok(vsiselect.bits) }
+  else { siselect  = legalize_iselect(value); Ok(siselect.bits) }
+
+function read_sireg(i : range(1, 6)) -> xlenbits =
+  if inVirtualPrivilege()
+  then get_ireg(Vsireg, vsiselect, i)
+  else get_ireg(Sireg, siselect, i)
+
+function write_sireg(i : range(1, 6), value : xlenbits) -> result(xlenbits, unit) =
+  if inVirtualPrivilege()
+  then set_ireg(Vsireg, vsiselect, i, value)
+  else set_ireg(Sireg, siselect, i, value)
+
+function sireg_accessible(i : range(1, 6)) -> bool =
+  if inVirtualPrivilege()
+  then indirect_reg_implemented(Vsireg, vsiselect, i)
+  else indirect_reg_implemented(Sireg, siselect, i)
+
+// vsireg* is reachable only from M/HS.
+function vsireg_accessible(i : range(1, 6)) -> bool =
+  not(inVirtualPrivilege()) & indirect_reg_implemented(Vsireg, vsiselect, i)
+
+// Picks virtual vs illegal for a failed sireg* access from VS/VU.
+// NOTE: "the hypervisor extension's usual rules for when to raise a
+// virtual-instruction exception (based on whether an instruction is
+// HS-qualified) are not applicable" here.
+// TODO: VS access should also be virtual when vsiselect "holds a value that is
+// implemented at HS level but not at VS level". Unreachable for now as no select
+// is implemented at either level.
+function is_sireg_exception_virtual(priv : Privilege) -> bool =
+  match priv {
+    VirtualUser       => mstateen0_csrind(),
+    VirtualSupervisor => mstateen0_csrind() & not(hstateen0_csrind()),
+    _                 => false,
+  }
+
+// Machine indirect register select and alias CSRs
+mapping clause csr_name_map = 0x350  <-> "miselect"
+mapping clause csr_name_map = 0x351  <-> "mireg"
+mapping clause csr_name_map = 0x352  <-> "mireg2"
+mapping clause csr_name_map = 0x353  <-> "mireg3"
+mapping clause csr_name_map = 0x355  <-> "mireg4"
+mapping clause csr_name_map = 0x356  <-> "mireg5"
+mapping clause csr_name_map = 0x357  <-> "mireg6"
+
+// Supervisor indirect register select and alias CSRs
+mapping clause csr_name_map = 0x150  <-> "siselect"
+mapping clause csr_name_map = 0x151  <-> "sireg"
+mapping clause csr_name_map = 0x152  <-> "sireg2"
+mapping clause csr_name_map = 0x153  <-> "sireg3"
+mapping clause csr_name_map = 0x155  <-> "sireg4"
+mapping clause csr_name_map = 0x156  <-> "sireg5"
+mapping clause csr_name_map = 0x157  <-> "sireg6"
+
+// Virtual-supervisor indirect register select and alias CSRs
+mapping clause csr_name_map = 0x250  <-> "vsiselect"
+mapping clause csr_name_map = 0x251  <-> "vsireg"
+mapping clause csr_name_map = 0x252  <-> "vsireg2"
+mapping clause csr_name_map = 0x253  <-> "vsireg3"
+mapping clause csr_name_map = 0x255  <-> "vsireg4"
+mapping clause csr_name_map = 0x256  <-> "vsireg5"
+mapping clause csr_name_map = 0x257  <-> "vsireg6"
+
+// *iselect always exists. The *ireg* aliases exist only when the select is implemented.
+function clause is_CSR_accessible(0x350, _, _) = currentlyEnabled(Ext_Smcsrind)
+function clause is_CSR_accessible(0x351, _, _) = currentlyEnabled(Ext_Smcsrind) & indirect_reg_implemented(Mireg, miselect, 1)
+function clause is_CSR_accessible(0x352, _, _) = currentlyEnabled(Ext_Smcsrind) & indirect_reg_implemented(Mireg, miselect, 2)
+function clause is_CSR_accessible(0x353, _, _) = currentlyEnabled(Ext_Smcsrind) & indirect_reg_implemented(Mireg, miselect, 3)
+function clause is_CSR_accessible(0x355, _, _) = currentlyEnabled(Ext_Smcsrind) & indirect_reg_implemented(Mireg, miselect, 4)
+function clause is_CSR_accessible(0x356, _, _) = currentlyEnabled(Ext_Smcsrind) & indirect_reg_implemented(Mireg, miselect, 5)
+function clause is_CSR_accessible(0x357, _, _) = currentlyEnabled(Ext_Smcsrind) & indirect_reg_implemented(Mireg, miselect, 6)
+
+function clause is_CSR_accessible(0x150, _, _) = currentlyEnabled(Ext_Sscsrind)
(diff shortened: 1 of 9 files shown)


model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,98 @@
...
+
+function clause is_CSR_defined(0x150) = currentlyEnabled(Ext_Sscsrind)
+function clause is_CSR_defined(0x151) = currentlyEnabled(Ext_Sscsrind)
+function clause is_CSR_defined(0x152) = currentlyEnabled(Ext_Sscsrind)
+function clause is_CSR_defined(0x153) = currentlyEnabled(Ext_Sscsrind)
+function clause is_CSR_defined(0x155) = currentlyEnabled(Ext_Sscsrind)
+function clause is_CSR_defined(0x156) = currentlyEnabled(Ext_Sscsrind)
+function clause is_CSR_defined(0x157) = currentlyEnabled(Ext_Sscsrind)
+
+function clause read_CSR(0x350) = miselect.bits[xlen - 1 .. 0]
```
jordancarlin, line 68: `miselect` is `xlenbits`, so I think this should work fine:
```suggestion
function clause read_CSR(0x350) = miselect.bits
```

Same for `siselect` below.

model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,98 @@
...
+
+function clause currentlyEnabled(Ext_Smcsrind) = hartSupports(Ext_Smcsrind)
+function clause currentlyEnabled(Ext_Sscsrind) = hartSupports(Ext_Sscsrind) & currentlyEnabled(Ext_S)
+
+bitfield Iselect : xlenbits = {
+    csr : 11 .. 0
+}
+
+register miselect : Iselect
+register sirelect : Iselect
```
jordancarlin, line 17: Typo throughout all instances of this CSR.
```suggestion
register siselect : Iselect
```

model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,98 @@
...
+
+/*
+ * Different from miselect, An extension may be considered not to be implemented at supervisor level
+ * if machine level has disabled the extension for S-mode,
+ * such as by the settings of certain fields in CSR menvcfg, for example.
+ * So this may need to be checked in the future.
+ */
+function legalize_sirelect(i: Iselect, v: xlenbits) -> Iselect = { Mk_Iselect(zeros()) }
+
+function get_ireg(ireg : IregType, i : int) -> xlenbits = zeros()
```
jordancarlin, line 31: You should be able to restrict the type of `i` to `range(0, 6)` instead of the more general `int`.

model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,98 @@
...
+
+function clause is_CSR_defined(0x350) = currentlyEnabled(Ext_Smcsrind)
+function clause is_CSR_defined(0x351) = currentlyEnabled(Ext_Smcsrind)
+function clause is_CSR_defined(0x352) = currentlyEnabled(Ext_Smcsrind)
+function clause is_CSR_defined(0x353) = currentlyEnabled(Ext_Smcsrind)
+function clause is_CSR_defined(0x355) = currentlyEnabled(Ext_Smcsrind)
+function clause is_CSR_defined(0x356) = currentlyEnabled(Ext_Smcsrind)
+function clause is_CSR_defined(0x357) = currentlyEnabled(Ext_Smcsrind)
+
+function clause is_CSR_defined(0x150) = currentlyEnabled(Ext_Sscsrind)
```
jordancarlin, line 60: From the spec:
> The machine-level extension Smcsrind encompasses all added CSRs and all behavior modifications for a hart, over all privilege levels. For a supervisor-level environment, extension Sscsrind is essentially the same as Smcsrind except excluding the machine-level CSRs and behavior not directly visible to supervisor level.

Seems like these CSRs should be accessible based on `currentlyEnabled(Ext_Smcsrind)` as well. Or maybe we set `currentlyEnabled(Ext_Sscsrind)` to `(hartSupports(Ext_Sscsrind) | currentlyEnabled(Ext_Sscsrind)) & currentlyEnabled(Ext_S)`.

model/riscv_csrind_regs.sail
jordancarlin, line 31: Not sure that this should just be returning 0. From the spec:

> The behavior upon accessing mireg* from M-mode, while miselect holds a value that is not implemented, is UNSPECIFIED.
>> It is expected that implementations will typically raise an illegal instruction exception for such accesses, so that, for example, they can be identified as software bugs. Platform specs, profile specs, and/or the Privileged ISA spec may place more restrictions on behavior for such accesses.

Seems like we at least need a configuration option for this to raise an illegal instruction exception. Maybe we define an `is_ireg_accessible` function that is part of the `is_CSR_defined` clause for each of the `mireg`/`sireg` CSRs. That way, if `miselect` does not hold a value that allows for accessing a particular `mireg`, it will fall through to the existing default case of an illegal instruction exception.

jordancarlin (changes requested):

model/riscv_csrind_regs.sail
KotorinMinami: Actually, I already implemented it in the previous push. Are there still any problem with the current implementation?

model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,108 @@
...
+
+function legalize_miselect(i: Iselect, v: xlenbits) -> Iselect = { Mk_Iselect(zeros()) }
+
+/*
+ * Different from miselect, An extension may be considered not to be implemented at supervisor level
+ * if machine level has disabled the extension for S-mode,
+ * such as by the settings of certain fields in CSR menvcfg, for example.
+ * So this may need to be checked in the future.
+ */
+function legalize_siselect(i: Iselect, v: xlenbits) -> Iselect = { Mk_Iselect(zeros()) }
```
radimkrcmar: ```The siselect register will support the value range 0..0xFFF at a minimum.```
This means that the legalization must keep the value in the register even if the selected feature is not implemented.

We can have a config option to increase the range above this value.

model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,108 @@
...
+bitfield Iselect : xlenbits = {
+    csr : 11 .. 0
+}
+
+register miselect : Iselect
+register siselect : Iselect
+
+enum IregType = { Mireg, Sireg }
+
+function legalize_miselect(i: Iselect, v: xlenbits) -> Iselect = { Mk_Iselect(zeros()) }
```
radimkrcmar: ```The miselect register implements at least enough bits to support all implemented miselect values.```
I think that it would make sense to have a configuration option for the supported values.

model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,108 @@
...
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function clause currentlyEnabled(Ext_Smcsrind) = hartSupports(Ext_Smcsrind) & currentlyEnabled(Ext_Zicsr)
+function clause currentlyEnabled(Ext_Sscsrind) = (hartSupports(Ext_Sscsrind) | currentlyEnabled(Ext_Smcsrind)) & currentlyEnabled(Ext_Zicsr) & currentlyEnabled(Ext_S)
+
+bitfield Iselect : xlenbits = {
+    csr : 11 .. 0
```
radimkrcmar: Why hardcode the 11 here?
The width of miselect and siselect can differ and we already have the legalization functions.

model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,108 @@
...
+ * Different from miselect, An extension may be considered not to be implemented at supervisor level
+ * if machine level has disabled the extension for S-mode,
+ * such as by the settings of certain fields in CSR menvcfg, for example.
+ * So this may need to be checked in the future.
+ */
+function legalize_siselect(i: Iselect, v: xlenbits) -> Iselect = { Mk_Iselect(zeros()) }
+
+function is_ireg_accessible(i: Iselect, ireg : IregType) -> bool =
+    match ireg {
+        Mireg => match i[csr] {
```
radimkrcmar: The csr is a global variable?  And what is this code even trying to do?

(I'd remove this function for now.)

model/riscv_csrind_regs.sail
```diff
@@ -0,0 +1,108 @@
...
+function clause read_CSR(0x150) = siselect.bits
+function clause read_CSR(0x151) = get_ireg(Sireg, 1)
+function clause read_CSR(0x152) = get_ireg(Sireg, 2)
+function clause read_CSR(0x153) = get_ireg(Sireg, 3)
+function clause read_CSR(0x155) = get_ireg(Sireg, 4)
+function clause read_CSR(0x156) = get_ireg(Sireg, 5)
+function clause read_CSR(0x157) = get_ireg(Sireg, 6)
+
+function clause write_CSR(0x350, value) = { miselect = legalize_miselect(miselect, value); miselect.bits }
+function clause write_CSR(0x351, value) = set_ireg(Mireg, 1, value)
```
radimkrcmar: The split between Mireg and Sireg is not important -- if the indirect access is allowed, the behavior should be the same.
I think we should pass which CSR is being accessed instead. i.e. miselect or siselect.

radimkrcmar (changes requested):

nadime15:
@KotorinMinami do you have the time and interest to continue working on this PR? If so, would you be willing to address the comments and rebase it?

KotorinMinami:
I may try to rebase it. @nadime15

ibvqeibob:
@nadime15 @KotorinMinami I ported the `miselect` CSR (0x350) implementation onto the #956 layout and gated it behind the Smcsrind/Sscsrind config flags.
Branch: `miselect-on-pr956`, commit: `f92a05b` — feel free to cherry-pick.

nadime15:
We discussed today in the golden model meeting that it would be great to get this merged, since the extension in #1457 depends on it. @KotorinMinami, could we rebase this PR? In the meantime, I am going to review it. 🙂

model/riscv_extensions.sail
```diff
@@ -210,6 +210,14 @@ function clause hartSupports(Ext_Zvksh) = config extensions.Zvksh.supported
 // Count Overflow and Mode-Based Filtering
 enum clause extension = Ext_Sscofpmf
 function clause hartSupports(Ext_Sscofpmf) = config extensions.Sscofpmf.supported
+
+// Machine-mode view of the Indirect CSR Access
+enum clause extension = Ext_Smcsrind
+function clause hartSupports(Ext_Smcsrind) = config extensions.Smcsrind.supported
+// Supervisor-mode view of the Indirect CSR Access
+enum clause extension = Ext_Sscsrind
+function clause hartSupports(Ext_Sscsrind) = config extensions.Sscsrind.supported
```
nadime15: Can you add these extensions to the ISA string below (extensions_ordered_for_isa_string)?

model/riscv_extensions.sail
```diff
@@ -210,6 +210,14 @@ function clause hartSupports(Ext_Zvksh) = config extensions.Zvksh.supported
 // Count Overflow and Mode-Based Filtering
 enum clause extension = Ext_Sscofpmf
 function clause hartSupports(Ext_Sscofpmf) = config extensions.Sscofpmf.supported
+
+// Machine-mode view of the Indirect CSR Access
+enum clause extension = Ext_Smcsrind
+function clause hartSupports(Ext_Smcsrind) = config extensions.Smcsrind.supported
```
nadime15: We can reorganize this part and move the M-Mode Extension down to the other M-Mode extensions.

nadime15 (commented):
I will continue the review later, but we should already add the Virtual Supervisor CSRs as well. Also, since we now support Smstateen, we need to take Section 5.1.5 ("Access control by the state-enable CSRs") into account.

nadime15:
I took a copy of your branch @KotorinMinami and added the remaining pieces (the Hypervisor and Stateen related parts). Would it be okay if I pushed this version instead? You can find the changes here:

https://github.com/riscv/sail-riscv/compare/master...nadime15:sail-riscv:ext_Sxscsrind

KotorinMinami:
I apologize. Of course, I've been busy with my exams recently, so feel free to push

nadime15:
@jordancarlin If you have some bandwidth, could you review this PR? Its quite small extension, and once its merged, we can move forward with #1457.

model/extensions/Csrind/csrind_regs.sail
```diff
@@ -0,0 +1,193 @@
...
+// two-clause license in the LICENSE file.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+// =======================================================================================
+
+function clause currentlyEnabled(Ext_Smcsrind) = hartSupports(Ext_Smcsrind) & currentlyEnabled(Ext_Zicsr)
+function clause currentlyEnabled(Ext_Sscsrind) = (hartSupports(Ext_Sscsrind) | currentlyEnabled(Ext_Smcsrind)) & currentlyEnabled(Ext_Zicsr) & currentlyEnabled(Ext_S)
+
+bitfield Iselect : xlenbits = {
+  value : 11 .. 0,
```
radimkrcmar, line 13: The bit width could be a config parameter.

model/extensions/Stateen/stateen_regs.sail
```diff
@@ -119,6 +121,7 @@ function legalize_hstateen0(h : Hstateen0, v : bits(64)) -> Hstateen0 = {
   [h with
     SE0    = if config extensions.Stateen.SE0_readonly_zero then 0b0 else v[SE0],
     ENVCFG = if currentlyEnabled(Ext_S) then v[ENVCFG] else 0b0,
+    CSRIND = if currentlyEnabled(Ext_Smcsrind) then v[CSRIND] else 0b0,
```
radimkrcmar, line 124: This would probably be slightly more precise as:
```suggestion
    CSRIND = if currentlyEnabled(Ext_Sscsrind) then v[CSRIND] else 0b0,
```

radimkrcmar (commented):
I only gave it a quick look to finish before the meeting...
A subset of the indirect CSR extensions is also defined as a part of AIA, but the current code looks well structured for that extension.

christian-herber-nxp:
I just discovered this PR, and had drafted indirect CSR support in parallel (see https://github.com/riscv/sail-riscv/pull/1874).

There is an architectural difference between our approaches. In my PR, the indirect-register dispatch is scattered, mirroring direct CSRs. Thus, extensions do not have to touch the Smcsrind files, and registers are fully contained within the extension itself. This is in my opinion the better architecture, and is more aligned to what is already in place for direct CSRs.

I have already built an initial version of AIA on my indirect CSR support. This is a demonstration of how Smcsrind is integrated with: https://github.com/NXP/sail-riscv/pull/1

nadime15:
Hmm, it's been a while since I last looked at this code, but nothing upstream is actually using these extensions yet, so the final approach isn't settled anyway.
 
The AIA registers won't be defined in csrind_regs.sail anyway. csrind_regs.sail would only handle the routing/dispatching (see indirect_reg_implemented) to the relevant state registers, such as those introduced by AIA, so adding AIA would really just add a few routing lines there. The entire AIA implementation would still remain self contained in its own file.

Either way it's easy to change once there is a real use and I am fine with both.

PR #957, opened 2025-05-20
Use existing bit counting functions

Timmmm:
Use the existing bit counting functions, and add `count_ones()`.

One downside is that `count_leading_zeros()` is not implemented in Sail, so it makes the inclusion in the ISA manual less helpful.

However given the difficulties of actually getting Sail code into the ISA manual it's probably ok for now. Also `count_leading_zeros()` is a standard operation and has obvious semantics.

### DIFF
--- a/model/riscv_insts_vext_fp_utils.sail
+++ b/model/riscv_insts_vext_fp_utils.sail
@@ -476,7 +476,7 @@ function rsqrt7 (v, sub) = {
 
   let (normalized_exp, normalized_sig) =
       if sub then {
-        let nr_leadingzeros = count_leadingzeros(sig, s);
+        let nr_leadingzeros = count_leading_zeros(sig[s - 1 .. 0]);
         assert(nr_leadingzeros >= 0);
         (to_bits_unsafe(64, (0 - nr_leadingzeros)), zero_extend(64, sig[(s - 1) .. 0] << (1 + nr_leadingzeros)))
       } else {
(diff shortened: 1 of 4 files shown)


model/riscv_insts_vext_fp_utils.sail
```diff
@@ -544,12 +544,11 @@ function riscv_f64Rsqrte7 (rm, v) = {
 
 val recip7 : forall 'm, 'm in {16, 32, 64}. (bits('m), bits(3), bool) -> (bool, bits_D)
 function recip7 (v, rm_3b, sub) = {
-  let (sig, exp, sign, e, s) : (bits(64), bits(64), bits(1), nat, nat) = match 'm {
```
jordancarlin, line 547: I think this is slightly weaker than the existing assert since it allows any combination of these values instead of just specific pairs. Not sure that's actually an issue though and I generally think it's better to push this up into the type constraints.

jordancarlin (approved):
Nice cleanup. If documentation in the ISA manual for these functions becomes an issue, we can always include the library version of them as an appendix or something.

model/riscv_insts_vext_fp_utils.sail
Timmmm: True, though the combinations are just hardcoded in the next 3 lines anyway. I think the assert was there just to make type checking pass, rather than to actually verify the combination of these values.

tobiasgrosser (approved):
This is nice. Thank you, @Timmmm!

nadime15 (approved):

PR #959, opened 2025-05-20
Fix to_bits_unsafe() in crypto extension

Timmmm:
There was no need for using `to_bits()` in the first place - we can do it simpler just by accessing the bits directly.

### DIFF
--- a/model/riscv_types_kext.sail
+++ b/model/riscv_types_kext.sail
@@ -213,8 +213,8 @@ function aes_subword_inv(x) = {
 val sm4_sbox : bits(8) -> bits(8)
 function sm4_sbox(x) = sbox_lookup(x, sm4_sbox_table)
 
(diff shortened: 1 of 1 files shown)


model/riscv_types_kext.sail
```diff
@@ -250,8 +250,8 @@ function aes_apply_inv_sbox_to_each_byte(x) = {
  * AES full-round transformation functions.
  */
 
-val getbyte : (bits(64), int) -> bits(8)
-function getbyte(x, i) = (x >> to_bits_unsafe(6, i * 8))[7..0]
+function getbyte(x : bits(64), i : range(0, 7)) -> bits(8) =
```
jordancarlin, line 253: Might be worth having this as a more generic function in `prelude.sail` at some point.

jordancarlin (approved):

PR #960, opened 2025-05-20
Explicitly use get_bits_truncate() for multiplications

Timmmm:
This is one place where we explicitly do want to truncate the result.

### DIFF
--- a/model/riscv_insts_mext.sail
+++ b/model/riscv_insts_mext.sail
@@ -30,9 +30,9 @@ mapping clause encdec = MUL(rs2, rs1, rd, mul_op)
 function clause execute (MUL(rs2, rs1, rd, mul_op)) = {
   let rs1_val = X(rs1);
   let rs2_val = X(rs2);
-  let rs1_int : int = if mul_op.signed_rs1 then signed(rs1_val) else unsigned(rs1_val);
-  let rs2_int : int = if mul_op.signed_rs2 then signed(rs2_val) else unsigned(rs2_val);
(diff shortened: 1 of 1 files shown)


pmundkur (approved):

model/riscv_insts_mext.sail
```diff
@@ -110,8 +110,7 @@ function clause execute (MULW(rs2, rs1, rd)) = {
   let rs2_val = X(rs2)[31..0];
   let rs1_int : int = signed(rs1_val);
   let rs2_int : int = signed(rs2_val);
-  /* to_bits_unsafe requires expansion to 64 bits followed by truncation */
-  let result32 = to_bits_unsafe(64, rs1_int * rs2_int)[31..0];
+  let result32 : bits(32) = to_bits_truncate(rs1_int * rs2_int);
```
jordancarlin, line 113: This is the only w instruction that has intermediate `result32` and `result` variables. Any reason to keep it? `divw` and `remw` directly assign to `X(rd)`.

Timmmm: I've changed this in #902 (which is waiting for a new Sail compiler) so I think I'll just resolve it there. I think this is probably clearest:

```
let result32 : bits(32) = to_bits_truncate(rs1_int * rs2_int);
X(rd) = sign_extend(result32);
```

Timmmm: Applied that change in #902

Timmmm:
This is now a subset of #902 so I'll close it in favour of that.

PR #961, opened 2025-05-20
Fix to_bits_unsafe() in PMP code

Timmmm:
Also reformat it slightly.


arichardson (approved):

pmundkur (approved):
We should add something to CODE_STYLE for formatting long argument lists to improve consistency in the code base.

PR #964, opened 2025-05-20
Fix uses of to_bits_unsafe for bit rotate and Zvbb

Timmmm:
This delegates the bits versions to the integer versions. This also identified a dangerous looseness in the types - previously you could use the bit version to shift by any amount, and it would give the wrong results if the shift was more than `'n`.

In Zvbb we have to use `2 ^ SEW_pow` instead of `get_sew()` so that Sail knows the link between `SEW` and `SEW_pow`.

### DIFF
--- a/model/riscv_insts_zvbb.sail
+++ b/model/riscv_insts_zvbb.sail
@@ -233,7 +233,7 @@ function clause execute (VCLZ_V(vm, vs2, vd)) = {
   foreach (i from 0 to (num_elem - 1)) {
     if mask[i] == bitone then {
       let clz = count_leading_zeros(vs2_val[i]);
-      result[i] = to_bits_unsafe('m, clz);
+      result[i] = to_bits(clz);
     };
   };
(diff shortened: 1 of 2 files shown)


model/riscv_insts_zvbb.sail
```diff
@@ -404,7 +406,8 @@ mapping clause assembly = VROR_VV(vm, vs1, vs2, vd)
  <-> "vror.vv" ^ spc() ^ vreg_name(vd) ^ sep() ^ vreg_name(vs2) ^ sep() ^ vreg_name(vs1) ^ maybe_vmask(vm)
 
 function clause execute (VROR_VV(vm, vs1, vs2, vd)) = {
-  let SEW      = get_sew();
+  let SEW_pow  = get_sew_pow();
```
arichardson, line 409: Unfortunate that we have to make this change but IMO the other cleanups are worth it.

arichardson (approved):

jordancarlin (approved):

pmundkur (approved):

PR #966, opened 2025-05-21
os-boot/opensbi: add opensbi.

Arielfoever:


Arielfoever:
The number should change.

os-boot/opensbi/sail.dts
```diff
@@ -14,10 +17,16 @@
       reg = <0>;
       status = "okay";
       compatible = "riscv";
-      riscv,isa = "rv64imac";
+      riscv,isa-base = "rv64i";
+      riscv,isa-extensions = "i", "m", "a", "f", "d", "c", "v", "sstc", "svinval", "zba", "zbb", "zbc", "zbs", "zca", "zcb", "zcd", "zfa", "zfh", "zkn", "zkt", "zicbom", "zicboz", "zicntr", "zicond", "zicsr", "zifencei", "zihpm";
```
arichardson, line 21: Not for this PR, but it would be great if we could have the sail emulator output a dts file.

Arielfoever: I had appied this changes.

os-boot/opensbi/sail.dts
```diff
@@ -14,10 +17,16 @@
       reg = <0>;
       status = "okay";
       compatible = "riscv";
-      riscv,isa = "rv64imac";
+      riscv,isa-base = "rv64i";
```
jrtc27, line 20: This can't boot FreeBSD, it currently only recognises riscv,isa.

jrtc27: Also, why is the FDT within opensbi/?

os-boot/opensbi/sail.dts
```diff
@@ -5,6 +5,9 @@
   #size-cells = <2>;
   compatible = "ucbbar,spike-bare-dev";
   model = "ucbbar,spike-bare";
+  chosen {
+    bootargs = "console=hvc0 earlycon=sbi";
```
jrtc27, line 9: These are Linux-specific, do they really belong here?

Arielfoever: That doesn't mater I think(

os-boot/opensbi/sail.dts
Arielfoever: > This can't boot FreeBSD, it currently only recognises riscv,isa.

I just test openSBI here, let FreeBSD to another pr(

pmundkur: Since the model can now generate a device tree, perhaps we could use that instead?

PR #967, opened 2025-05-21
Remove the use of function wrappers to initialize config parameters.

pmundkur:
This was a legacy of the configuration via cli option approach.
    
Parameters initialized via `to_bits_checked()` are set to `register` values, since top-level `let` bindings seem to require pure initializers and the assert in `to_bits_checked()` is not pure.

PMP config parameters were moved to`riscv_pmp_regs.sail` from `riscv_sys_regs.sail`, and a now redundant assert was removed from `reset_pmp()`.

Rename the vector config parameters since they are not `get_` functions anymore. This results in a slight simplification of vector code using these parameters.

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -28,30 +28,30 @@ val elf_entry = pure {
 // Cache block size is 2^cache_block_size_exp. Max is `max_mem_access` (4096)
 // because this model performs `cbo.zero` with a single write, and the behaviour
 // with cache blocks larger than a page is not clearly defined.
-function plat_cache_block_size_exp() -> range(0, 12) = config platform.cache_block_size_exp
+let plat_cache_block_size_exp : range(0, 12) = config platform.cache_block_size_exp
 
 /* Main memory */
-function plat_ram_base() -> physaddrbits = to_bits_checked(config platform.ram.base : int)
-function plat_ram_size() -> physaddrbits = to_bits_checked(config platform.ram.size : int)
+register plat_ram_base : physaddrbits = to_bits_checked(config platform.ram.base : int)
+register plat_ram_size : physaddrbits = to_bits_checked(config platform.ram.size : int)
 
 /* whether the MMU should update dirty bits in PTEs */
-function plat_enable_dirty_update() -> bool = config memory.translation.dirty_update
+let plat_enable_dirty_update : bool = config memory.translation.dirty_update
 
 /* whether the platform supports misaligned accesses without trapping to M-mode. if false,
  * misaligned loads/stores are trapped to Machine mode.
  */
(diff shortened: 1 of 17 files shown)


jordancarlin (commented):
Great cleanup. We should probably add a comment about the register vs let issue because that’s not immediately obvious. Otherwise LGTM.

pmundkur:
There's no single place to put that comment, so I put it in the commit message.  Currently, this only affects `riscv_platform` but for #954 it also occurs in another file.  It should really go in the docs.

jordancarlin (approved):

model/riscv_platform.sail
```diff
@@ -28,38 +28,38 @@ val elf_entry = pure {
 // Cache block size is 2^cache_block_size_exp. Max is `max_mem_access` (4096)
 // because this model performs `cbo.zero` with a single write, and the behaviour
 // with cache blocks larger than a page is not clearly defined.
-function plat_cache_block_size_exp() -> range(0, 12) = config platform.cache_block_size_exp
+let plat_cache_block_size_exp : range(0, 12) = config platform.cache_block_size_exp
 
 /* Main memory */
-function plat_ram_base() -> physaddrbits = to_bits_checked(config platform.ram.base : int)
-function plat_ram_size() -> physaddrbits = to_bits_checked(config platform.ram.size : int)
+register plat_ram_base : physaddrbits = to_bits_checked(config platform.ram.base : int)
```
Timmmm, line 34: Why not `let`?

model/riscv_platform.sail
```diff
@@ -28,38 +28,38 @@ val elf_entry = pure {
...
+let plat_enable_misaligned_access : bool = config memory.misaligned.supported
 
 /* whether mtval stores the bits of a faulting instruction on illegal instruction exceptions */
-function plat_mtval_has_illegal_inst_bits() -> bool = config base.mtval_has_illegal_instruction_bits
+let plat_mtval_has_illegal_inst_bits : bool = config base.mtval_has_illegal_instruction_bits
 
 /* ROM holding reset vector and device-tree DTB */
-function plat_rom_base() -> physaddrbits = to_bits_checked(config platform.rom.base : int)
-function plat_rom_size() -> physaddrbits = to_bits_checked(config platform.rom.size : int)
+register plat_rom_base : physaddrbits = to_bits_checked(config platform.rom.base : int)
```
Timmmm, line 49: Ditto

Timmmm (commented):
This is great. I reckon we might want to rename some of these things at some point, but we can do that later. Also this slightly conflicts with my change to make vlen a configure time option but I haven't actually pushed that yet so I can't really complain! I'll deal with the conflicts there.

arichardson (approved):

jordancarlin (approved):

PR #970, opened 2025-05-22
Fix printf in first party test

### LINKED ISSUE #944: Printf in first party tests not working
I definitely had this working at some point but now the hello world test prints `Hello �0 0 0 0.000000` whereas it's meant to print `Hello world 1 2 3.00000`. Haven't debugged it at all.

To test, `make` and then

    c_emulator/riscv_sim_rv64d -V test/first_party/rv64d_test_hello_world.c.elf

trdthg:
fix #944 

Passing va to ... is invalid, but I didn't find an offical doc

### DIFF
--- a/test/first_party/src/common/runtime.c
+++ b/test/first_party/src/common/runtime.c
@@ -6,7 +6,7 @@ int printf(const char *fmt, ...)
 {
   va_list ap;
   va_start(ap, fmt);
-  int rc = npf_pprintf(&htif_putc, NULL, fmt, ap);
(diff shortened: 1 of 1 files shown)


Timmmm (approved):
Ah nice! I could have sworn this worked once but I guess not... :-D

jordancarlin (approved):

arichardson (approved):
Yeah that is implementation defined behaviour that may or may not work. Always need to use the va_list version.

PR #971, opened 2025-05-22
Make vlen a configure-time option

Timmmm:
Rather than having vlenmax (64kb) sized vector registers and only using part of them, just allow configuring vlen directly.


model/riscv_sys_regs.sail
```diff
@@ -915,7 +915,7 @@ register vstart : bits(16) /* use the largest possible length of vstart */
 register vl     : xlenbits
 
 function get_vlenb() -> xlenbits =
-  to_bits(2 ^ (get_vlen_pow()) / 8)
+  to_bits(xlen, vlen / 8)
```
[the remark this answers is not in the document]
Timmmm: Ah yeah it probably isn't - I think I just wrote this before the `to_bits` changes to support inference. I'll update it.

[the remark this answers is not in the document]
Timmmm: Removed

model/riscv_vlen.sail
```diff
@@ -6,15 +6,21 @@
...
-function get_vlen_pow() -> range(3, 16) = config extensions.V.vlen_exp
+let vlen_exp = sizeof(vlen_exp)
+let elen_exp = sizeof(elen_exp)
 
-type vlenmax : Int = 65536
+type vlen : Int = 2 ^ vlen_exp
+type elen : Int = 2 ^ elen_exp
 
-let VLEN = 2 ^ get_vlen_pow()
+let vlen = sizeof(vlen)
```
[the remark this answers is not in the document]
Timmmm: Done

model/riscv_vlen.sail
```diff
@@ -6,15 +6,21 @@
...
-let VLEN_pow : range(3, 16) = config extensions.V.vlen_exp
+let vlen_exp = sizeof(vlen_exp)
+let elen_exp = sizeof(elen_exp)
 
-type vlenmax : Int = 65536
+type vlen : Int = 2 ^ vlen_exp
+type elen : Int = 2 ^ elen_exp
 
-let VLEN = 2 ^ VLEN_pow
+let vlen = sizeof(vlen)
```
nadime15, line 25: Out of curiosity, #1005 added config validation, but we are now basically double checking the same condition twice. Does this really need to stay? Couldnt we just revert to the original version and instead of:

```
type elen_exp : Int = config extensions.V.elen_exp
constraint 3 <= elen_exp <= 16 & elen_exp <= vlen_exp
```

go back to

`type elen_exp : range(3, 16) = config extensions.V.elen_exp`

I think you could still call sizeof(elen_exp) in this case (?)

Timmmm: Hmm yeah we might be able to get rid of the `3 <= elen_exp <= 16` . I think `elen_exp <= vlen_exp` is still needed. The config validation function doesn't expose that information to the type system.

I'll try it.

Timmmm: Oh yeah actually you can't do that, `Int` is the type of a type (AKA a "kind"), but `range(3, 16)` is just a type.

jordancarlin: We could probably go the other direction and drop it from the config validation function now that it is enforced by the type system. The validation function might give a better error though that makes it worth keeping around.

Timmmm: Yeah I wouldn't be too surprised if Sail dead-code eliminates the check. Ideally Sail would generate runtime code to verify constraints but I expect that's a bit of work...

nadime15: > We could probably go the other direction and drop it from the config validation function now that it is enforced by the type system. The validation function might give a better error though that makes it worth keeping around.

Yeah, makes sense, but would be nice if we could keep all the validation logic in one file and not have it scattered across multiple files. But yeah, thats definitely an option.

model/riscv_vlen.sail
```diff
@@ -6,15 +6,21 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-let ELEN_pow : range(3, 16) = config extensions.V.elen_exp
+type vlen_exp : Int = config extensions.V.vlen_exp
```
pmundkur, line 9: Won't this need calls to `sail_set_abstract_vlen_exp()` in `riscv_sim.cpp`?  Did the vector tests pass without this?

model/riscv_vlen.sail
```diff
@@ -6,15 +6,21 @@
...
+constraint 3 <= vlen_exp <= 16
 
 /* Note: ELEN=32 requires a different encoding of the CSR vtype.
  * The current version of vtype implementation corresponds to the ELEN=64 configuration.
  * TODO: the configurarion of ELEN and its corresponding vtype implementations.
  */
+type elen_exp : Int = config extensions.V.elen_exp
```
pmundkur, line 16: Similarly here for `sail_set_abstract_elen_exp()`.

model/riscv_vlen.sail
Timmmm: Ah yes you're right! I haven't run the vector tests with this yet actually.

Timmmm: Added `sail_set_abstract...` and ran a few vector tests which pass. I'll see if I can run the whole suite on it.

Timmmm:
Full test suite run: https://github.com/riscv/sail-riscv/actions/runs/16343464307

Timmmm:
Success! However it can confirm that Sail dead-code eliminates the elen < vlen check:

```
bool zcheck_vlen_elen(unit z3zE27830)
{
  __label__ end_function_22946, end_block_exception_22947;

  bool z6zE1585;
  bool z2zE18081;
  z2zE18081 = (zvlen_exp < zelen_exp);
  unit z3zE27831;
  z3zE27831 = UNIT;
  z6zE1585 = true;
end_function_22946: ;
  return z6zE1585;
end_block_exception_22947: ;

  return false;
}
```

I think the best solution here would be for the C backend to generate code to actually check the constraints and types of `config` values when they are loaded. I'll make an issue. I think we should just accept that for now though and merge this anyway.

Timmmm:
I made [this issue](https://github.com/rems-project/sail/issues/1387) for the Sail compiler. I think it's quite a lot of work so lets not hold our breath, but I think we can live without it for a while.

nadime15 (approved):
Looks good to me. Great to get this in!

model/riscv_vlen.sail
pmundkur: Hmm, this PR breaks the `check_vlen_elen()` check in the config validator, perhaps due to this Sail bug https://github.com/rems-project/sail/issues/1383#issuecomment-3079418924
```
$ ./build/c_emulator/sail_riscv_sim --config config/rv64d.json --validate-config
VLEN set to 2^9   [added debug logging]
ELEN set to 2^11  [added debug logging]
Configuration in config/rv64d.json is valid.
```
Changing the `if vlen_exp < elen_exp` in `check_vlen_elen` to `if (vlen_exp : nat) < (elen_exp : nat)` fixes that check for some reason (unrelated to the above issue?).

Since the type-level range constraints are not checked, they would need to be checked in the `check_vlen_elen` function, to avoid this:
```
$ ./build/c_emulator/sail_riscv_sim --config config/rv64d.json --validate-config
VLEN set to 2^22
ELEN set to 2^22
Configuration in config/rv64d.json is valid.
```
It's unfortunate but we'd need to repeat the constraints twice.

pmundkur (changes requested):

model/riscv_vlen.sail
Timmmm: I think that bug is a separate (but kind of related) issue. The problem here is that the `constraint` is a type level assertion that tells the compiler "you can assume `elen_exp <= vlen_exp`". Unlike the value-level assertion `assert(elen_exp <= vlen_exp)` there isn't actually anything that checks that this is true.

Because it's a type level assertion the check has to be done at compile time (or "configure" time). You can't really do it at runtime because by that point Sail is already assuming it is true (hence why it is dead-code eliminated).

Your `: nat` trick works because it tells Sail to disregard the type information, but it doesn't *have* to do that so it's technically relying on UB. It would be equally valid for Sail to say "well, they've told me they're expecting `elen_exp` to be any `nat` but actually I do know that it is less than `vlen_exp` so I can still eliminate this check". (Correct me if I'm wrong @Alasdair !)

I think the proper fix is: https://github.com/rems-project/sail/issues/1387

However, the `: nat` trick does work today and it's probably better than nothing so I'll go with that and add a comment explaining the situation.

pmundkur: I'm wondering how we should address the range constraint validation, since the below is still possible:
```
$ ./build/c_emulator/sail_riscv_sim --config config/rv64d.json --validate-config
elen_exp : 22
Configuration in config/rv64d.json is valid.
```

Until we get a solution for https://github.com/rems-project/sail/issues/1387, we could either handle it now in `check_vlen_elen()` or later (hopefully soon) by JSON schema validation.

Timmmm: I added an extra range change, now you get:

```
ELEN set to 2^22 but must be within [2^3, 2^16].
```

pmundkur (approved):

jordancarlin (approved):

PR #972, opened 2025-05-22
Print the device-tree for the model configuration.

### LINKED ISSUE #734: Will Sail support generating DTS ?
It seems that this feature is only related to debugging osboot?

> We should probably open an issue about this so we can decide if we want to build support for this into the model or just drop the feature entirely and remove the flag.

_Originally posted by @jordancarlin in https://github.com/riscv/sail-riscv/pull/732#discussion_r1952199747_

pmundkur:
Reduce the default ram size to 64MB as that should be enough for most small test cases.  This makes the DTS mostly compatible with the handwritten 64mb version in os-boot.  This can be improved on to reflect more of the model configuration in the future.

Fixes #734.


c_emulator/dts.cpp
```diff
@@ -0,0 +1,97 @@
...
+      = get_config_uint64({"platform", "instructions_per_tick"});
+  uint64_t pmp_count = get_config_uint64({"memory", "pmp", "count"});
+  uint64_t pmp_grain = get_config_uint64({"memory", "pmp", "grain"});
+  uint64_t ram_base = get_config_uint64({"platform", "ram", "base"});
+  uint64_t ram_size = get_config_uint64({"platform", "ram", "size"});
+  uint64_t clint_base = get_config_uint64({"platform", "clint", "base"});
+  uint64_t clint_size = get_config_uint64({"platform", "clint", "size"});
+
+  // Default, should be adjusted based on config.
+  const char *isa_string = "rv64imafdc_zicntr_zihpm";
```
jordancarlin, line 22: This should at least depend on `xlen` even if we aren't adding full configuration yet.

jordancarlin (commented):
Seems like a great feature. Not sure how useful it is until it more closely matches the configured model (especially the ISA string), but seems like a good starting point.

c_emulator/config_utils.cpp
```diff
@@ -0,0 +1,36 @@
+#include "config_utils.h"
+
+#include <iostream>
+#include <sstream>
```
arichardson, line 4:
```suggestion
```

c_emulator/config_utils.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <vector>
+#include "sail.h"
+#include "sail_config.h"
+
+uint64_t get_config_uint64(std::vector<const char *> keypath);
```
arichardson, line 7:
```suggestion
uint64_t get_config_uint64(const std::vector<const char *> &keypath);
```

c_emulator/config_utils.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+uint64_t get_config_uint64(std::vector<const char *> keypath)
```
arichardson, line 6: Const ref

c_emulator/config_utils.h
pmundkur: Making the const ref changes in the .h and .cpp gives:
```
sail-riscv/c_emulator/config_utils.cpp: In function ‘uint64_t get_config_uint64(const std::vector<const char*>&)’:
sail-riscv/c_emulator/config_utils.cpp:7:71: error: invalid conversion from ‘const char* const*’ to ‘const char**’ [-fpermissive]
    7 |   sail_config_json json = sail_config_get(keypath.size(), keypath.data());
      |                                                           ~~~~~~~~~~~~^~
      |                                                                       |
      |                                                                       const char* const*
In file included from sail-riscv/c_emulator/config_utils.h:5,
                 from sail-riscv/c_emulator/config_utils.cpp:1:
sail/lib/sail_config.h:80:68: note:   initializing argument 2 of ‘sail_json* sail_config_get(size_t, const char**)’
   80 | sail_config_json sail_config_get(const size_t n, const_sail_string key[]);
      |                                                  ~~~~~~~~~~~~~~~~~~^~~~~

```

Timmmm: Ah yeah that's a bug in Sail I think. Apparently it should be `const_sail_string key[const]`! Can you try hacking that header and seeing if it works?

@Alasdair FYI.

Timmmm (commented):
Nice. With #866 it's going to be a lot more complex to read information from JSON, since it's an array of structs.

I wonder if a better approach is to implement this function in Sail, then you can easily access the config data without having to mess around with JSON.

```
function generate_dts() -> string = {
   "/dts-v1/;\n" ^
...
   "riscv,isa = \"rv" ^ dec_str(xlen) ^ ...
```

I think that would probably end up a fair bit simpler. And better if we ever use any other backends for simulation.

c_emulator/config_utils.h
arichardson: C++ doesn't allow that syntax, I think it would be better to just use `const_sail_string const *key`: https://godbolt.org/z/fnszqncMr

arichardson: https://github.com/rems-project/sail/pull/1317

arichardson:
> Nice. With #866 it's going to be a lot more complex to read information from JSON, since it's an array of structs.
> 
> I wonder if a better approach is to implement this function in Sail, then you can easily access the config data without having to mess around with JSON.
> 
> ```
> function generate_dts() -> string = {
>    "/dts-v1/;\n" ^
> ...
>    "riscv,isa = \"rv" ^ dec_str(xlen) ^ ...
> ```
> 
> I think that would probably end up a fair bit simpler. And better if we ever use any other backends for simulation.

Generating it from sail sounds like a great suggestion since then it's easy to determine which extensions are enabled and generate the right ISA string, etc.

arichardson (approved):
This looks good as an initial version, we can make follow-up changes to fix the actual ISA string?

c_emulator/riscv_sim.cpp
```diff
@@ -244,6 +255,9 @@ static int process_args(int argc, char **argv)
     case OPT_PRINT_CONFIG:
       printf("%s", DEFAULT_JSON);
       exit(0);
+    case OPT_PRINT_DTS:
+      do_print_dts = true;
```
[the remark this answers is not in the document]
pmundkur: `print_dts()` needs to be called after `model_init()` to get an initialized config to use for the DTS.  `model_init()` is not needed for printing the JSON config.

model/riscv_device_tree.sail
```diff
@@ -0,0 +1,63 @@
+function generate_dts() -> string = {
+  let hart_hz : int = config platform.hart_hz;
+  let pmp_grain : int = 2 ^ (sys_pmp_grain + 2);
```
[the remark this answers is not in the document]
pmundkur: I put `pmpregions` and `pmpgranularity` to mimic what Spike did, but as far as I can tell, these entries are not defined in any device-tree bindings.  It appears that they were once defined for a Sifive chip on Zephyr, but not used:
https://github.com/zephyrproject-rtos/zephyr/commit/dbe783f5c4428835eba3c0e48ada4f84062f0c3c
I can't find any evidence them elsewhere.  I'm tempted to just remove them from the printed DTS.  Any objections?

model/riscv_device_tree.sail
```diff
@@ -0,0 +1,63 @@
...
+^ "    device_type = \"memory\";\n"
+^ "    reg = <" ^ hex_str(ram_base_hi) ^ " " ^ hex_str(ram_base_lo)
+^ " " ^ hex_str(ram_size_hi) ^ " " ^ hex_str(ram_size_lo) ^ ">;\n"
+^ "  };\n"
+^ "  soc {\n"
+^ "    #address-cells = <2>;\n"
+^ "    #size-cells = <2>;\n"
+^ "    compatible = \"ucbbar,spike-bare-soc\", \"simple-bus\";\n"
+^ "    ranges;\n"
+^ "    clint@" ^ string_drop(hex_str(unsigned(plat_clint_base)), 2) ^ " {\n"
```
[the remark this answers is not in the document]
pmundkur: `hex_bits_str` prints leading zeros, and node addresses in the DTSes I've seen omit those, so I'll stick with the current approach.

model/riscv_device_tree.sail
```diff
@@ -0,0 +1,63 @@
...
+  let pmp_grain : int = 2 ^ (sys_pmp_grain + 2);
+  let ram_base_hi = unsigned(plat_ram_base >> 32);
+  let ram_base_lo = unsigned(plat_ram_base[31 .. 0]);
+  let ram_size_hi = unsigned(plat_ram_size >> 32);
+  let ram_size_lo = unsigned(plat_ram_size[31 .. 0]);
+  let clint_base_hi = unsigned(plat_clint_base >> 32);
+  let clint_base_lo = unsigned(plat_clint_base[31 .. 0]);
+  let clint_size_hi = unsigned(plat_clint_size >> 32);
+  let clint_size_lo = unsigned(plat_clint_size[31 .. 0]);
+  let isa_string : string = "imafdc_zicntr_zihpm"; // Default, should be adjusted based on config.
```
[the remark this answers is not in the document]
arichardson: For reference here is the schema from the linux kernel: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings/riscv/extensions.yaml

model/riscv_device_tree.sail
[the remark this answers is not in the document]
arichardson: Shouldn't really matter since everyone consumes the binary version and I imagine `dtc` accepts both.

If we need this in more than one place in the future it may be worth adding a local helper function to print an unprefixed hex value. Or even with just one use since it makes it easier to read.

model/riscv_device_tree.sail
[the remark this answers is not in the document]
jordancarlin: One argument in favor of keeping the `pmp` properties is configuration of Spike. The `pmp` properties in the device tree are the only way of setting the pmp grain in Spike (see https://github.com/riscv-software-src/riscv-isa-sim/issues/703#issuecomment-827223405). Retaining them means Spike can easily be configured to match a configuration of Sail by having Sail spit out the device tree and using that as the input to Spike.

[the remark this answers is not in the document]
jordancarlin: Strange. I was looking for those recently and somehow got to the conclusion that the device tree was the only way to change them. Seems reasonable to defer connections between configs for now.

pmundkur:
It might be better to just fix the Lean backend instead of suppressing the Lean failures.  The point of adding Lean to the CI was to catch such bugs in the first place.

jordancarlin:
> It might be better to just fix the Lean backend instead of suppressing the Lean failures. The point of adding Lean to the CI was to catch such bugs in the first place.

Fair enough. @tobiasgrosser @ineol

model/riscv_device_tree.sail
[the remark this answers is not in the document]
Timmmm: Ah fair enough. Maybe add a comment saying dtc requires no leading zeros because that's kind of surprising!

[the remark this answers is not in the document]
pmundkur: Comment added.

ineol:
This [Sail PR](https://github.com/rems-project/sail/pull/1328) fixes the issue. Thank you for the bug report!

tobiasgrosser:
The Lean CI will work if eitehr https://github.com/riscv/sail-riscv/pull/987 is merged or the Lean CI cache is deleted.

c_emulator/config_utils.h
jordancarlin: Can you add a TODO to change this after we update the compiler version?

pmundkur: PR #999.

jordancarlin (approved):

model/riscv_device_tree.sail
```diff
@@ -0,0 +1,73 @@
...
+^ "    #address-cells = <1>;\n"
+^ "    #size-cells = <0>;\n"
+^ "    timebase-frequency = <" ^ dec_str(clock_freq / plat_insns_per_tick) ^ ">;\n"
+^ "    CPU0: cpu@0 {\n"
+^ "      device_type = \"cpu\";\n"
+^ "      reg = <0>;\n"
+^ "      status = \"okay\";\n"
+^ "      compatible = \"riscv\";\n"
+^ "      riscv,isa = \"" ^ "rv" ^ dec_str(xlen) ^ isa_string ^ "\";\n"
+^ "      mmu-type = \"riscv," ^ (if xlen == 32 then "sv32" else "sv39") ^ "\";\n"
```
nadime15, line 42: I think the mmu-type can be sv48 as well. I am not 100% sure how to verify that but I found some evidence here:

[https://github.com/riscv-non-isa/riscv-device-tree-doc/blob/master/bindings/riscv/cpus.txt](https://github.com/riscv-non-isa/riscv-device-tree-doc/blob/master/bindings/riscv/cpus.txt).


and here:

[https://android.googlesource.com/kernel/common/+/07e6124a1a46/Documentation/devicetree/bindings/riscv/cpus.yaml](https://android.googlesource.com/kernel/common/+/07e6124a1a46/Documentation/devicetree/bindings/riscv/cpus.yaml)

I checked the latest device-tree specification as well but as far as I can tell risc-v is not mentioned.

[https://github.com/devicetree-org/devicetree-specification/releases/tag/v0.4](https://github.com/devicetree-org/devicetree-specification/releases/tag/v0.4).

nadime15 (commented):
Sorry I am a bit late!

model/riscv_device_tree.sail
radimkrcmar: Yeah, the mmu-type is interpreted as the highest supported paging mode and implies all standard lower ones (e.g. "riscv,sv57" is {Sv57, Sv48, and Sv39}).
In Sail, we should have a condition chain with hartSupports to pick the correct one.

The design is not great and hopefully we'll move to listing of supported paging extensions in the future --  there is no good way to describe both rv32 and rv64 modes (when we have variable xlen) and mmu-type doesn't really work with custom modes either.

PR #973, opened 2025-05-22
fix: maybe_vmask() should not be prefixed with sep()

ThinkOpenly:
maybe_vmask() will emit its own comma when needed.


ThinkOpenly:
> Nice catch!

First fix from me that can be attributed to utilizing Sail-in-JSON. :-)

nadime15 (approved):

pmundkur (approved):

PR #975, opened 2025-05-22
Fix lean currentlyEnabled termination issue introduced by #904

jordancarlin:
@ineol @tobiasgrosser @bacam It would be great for this list to be something Sail can output so that this doesn't keep happening. Any chance that would be possible?


pmundkur (approved):

pmundkur:
Merging this now so that we have clean test results again.

PR #976, opened 2025-05-23
Simplify PMP control code by removing pmpMatchEntry

Timmmm:
I think this extra function was unnecessary and just made the code harder to follow.

### DIFF
--- a/model/riscv_pmp_control.sail
+++ b/model/riscv_pmp_control.sail
@@ -79,19 +79,6 @@ function pmpMatchAddr(
   }
 }
 
-enum pmpMatch = {PMP_Success, PMP_Continue, PMP_Fail}
-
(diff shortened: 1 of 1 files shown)


arichardson (approved):
Since there is only one call the extra indirection and new type don't add any value. easier to understand now.

arichardson:
CI is not happy with this: 
```
Type error:
riscv_pmp_control.sail:108.11-74:
108 |        if pmpCheckRWX(cfg, acc) | (priv == Machine & not(pmpLocked(cfg)))
    |           ^-------------------------------------------------------------^ checking function argument has type Pmpcfg_ent
    | Identifier cfg is unbound
```

Timmmm:
> CI is not happy with this:

Oops, fixed.

> One advantage of the current code is pmpMatchEntry is a pure function for which it is easier to write SMT property tests.

That is true. Not sure I have a strong opinion on that.

PR #977, opened 2025-05-23
Fix to_bits_unsafe() in Zbb

Timmmm:
We could potentially add functions for like `min_unsigned_bits()`, `max_signed_bits()` etc, but I think this is fine for now.


jordancarlin (approved):
Agreed this is fine for now. We can add the other functions later if there are other places they would be useful.

PR #982, opened 2025-05-24
Refactor vector extension support based on SPEC.

wwwwwwOwO:
In order to support the  fine-grained vector extension instead of V or nothing, the changes include:

1. Add the extensions Zvl*, Zve*, Zvfhmin, Zvfh in the new config system.
2. `currentlyEnabled()` for each extension based on dependencies described in V-spec 18.(`riscv_vext_control.sail` ). Some of them depend on F/D/Zfhmin extensions, so `riscv_vext_control.sail` needs to be added after `riscv_fdext_control.sail`.
3. Add switches to each support instruction strictly. 
  - For Load/Store, check `width` segment which implies EEW.
  - For others, check supported SEW. 
4. Merge `riscv_insts_vext_fp_red.sail`, `riscv_insts_vext_fp_vm.sail` into `riscv_insts_vext_fp.sail`. The 2 files  were generated when separated floating-point operations from integer, the purpose is to allow importing required files when F is not supported. So there is no need to keep the 2 files independent. 
5. Separate `Single-Width/Widening Floating-Point Reduction`. The old version implemented the Single-Width and  Widening  together, I separate them because they have different supported SEW.

When I am creating the PR, it says "Can’t automatically merge". I don't know how to handle it.
Please let me know if there are problems, thanks!


wwwwwwOwO:
I see the conflict. The V extension also depends on Zicsr. Maybe it needs to put them together.

jrtc27:
> Refactor vector extension support based on SPEC.

I assume you mean "spec" as in specification? That's not something to put in all-caps, and in fact is actively misleading in this case since "SPEC" in all-caps normally refers to the SPEC family of benchmarks.

config/rv32d.json
```diff
@@ -235,6 +235,45 @@
...
+    },
+    "Zve64d": {
+      "supported": true
+    },
+    "Zvfhmin": {
+      "supported": true
+    },
+    "Zvfh": {
+      "supported": true
+    },
```
nadime15: Extensions are usually ordered by category and then alphabetically if possible (this applies to rv64d.json and riscv_extensions.sail as well)

model/riscv_vext_control.sail
```diff
@@ -6,7 +6,24 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_V) = hartSupports(Ext_V) & (misa[V] == 0b1) & (mstatus[VS] != 0b00) & currentlyEnabled(Ext_Zicsr)
+function clause currentlyEnabled(Ext_Zvl32b)   = hartSupports(Ext_Zvl32b)   & (VLEN >= 32)
+function clause currentlyEnabled(Ext_Zvl64b)   = hartSupports(Ext_Zvl64b)   & (VLEN >= 64)
+function clause currentlyEnabled(Ext_Zvl128b)  = hartSupports(Ext_Zvl128b)  & (VLEN >= 128)
+function clause currentlyEnabled(Ext_Zvl256b)  = hartSupports(Ext_Zvl256b)  & (VLEN >= 256)
+function clause currentlyEnabled(Ext_Zvl512b)  = hartSupports(Ext_Zvl512b)  & (VLEN >= 512)
+function clause currentlyEnabled(Ext_Zvl1024b) = hartSupports(Ext_Zvl1024b) & (VLEN >= 1024)
```
nadime15: Dont you double check conditions now with (`VLEN>= x`), you already checked for those in riscv_extensions.sail. Since VLEN is fixed this should not be necessary.

model/riscv_vext_control.sail
```diff
@@ -6,7 +6,24 @@
...
+
+function clause currentlyEnabled(Ext_Zve32x)   = hartSupports(Ext_Zve32x)
+function clause currentlyEnabled(Ext_Zve32f)   = hartSupports(Ext_Zve32f)   & currentlyEnabled(Ext_Zve32x)   & currentlyEnabled(Ext_F)
+function clause currentlyEnabled(Ext_Zve64x)   = hartSupports(Ext_Zve64x)   & currentlyEnabled(Ext_Zve32x)
+function clause currentlyEnabled(Ext_Zve64f)   = hartSupports(Ext_Zve64f)   & currentlyEnabled(Ext_Zve64x)   & currentlyEnabled(Ext_Zve32f)
+function clause currentlyEnabled(Ext_Zve64d)   = hartSupports(Ext_Zve64d)   & currentlyEnabled(Ext_Zve64f)   & currentlyEnabled(Ext_D)
```
nadime15: You dont check for minimum VLEN (32 or 64), maybe something like:

`function clause currentlyEnabled(Ext_Zve32x)   = hartSupports(Ext_Zve32x) & currentlyEnabled(Ext_Zvl32b)`

nadime15 (commented):
Could it be that you accidentally changed the permissions for a lof of files? If so, it would be great if you could undo that, it would make the PR easier to review.

config/rv32d.json
wwwwwwOwO: > Extensions are usually ordered by category and then alphabetically if possible (this applies to rv64d.json and riscv_extensions.sail as well)

Thanks! I followed the spec’s order, since later extensions (like Zve64d) depend on earlier ones (e.g., Zve64f, Zve64x). Could you let me know how it should be ordered excatly？

nadime15:
Hi @wwwwwwOwO, thanks for the update! Do you think it would be possible to break this PR into smaller ones, each focusing on just one change? I feel like this changes too many things at once, and splitting it up would really help with the review process.

config/rv32d.json
nadime15: I think you can still keep the intended order (as described above) both in the JSON file and in the `extensions_ordered_for_isa_string()` function. The only place where you actually need to reorder is in riscv_extensions.sail, right at the beginning when you define:

```
// Vector Extensions for Embedded Processors
enum clause extension = Ext_Zve32x
mapping clause extensionName = Ext_Zve32x <-> "zve32x"
function clause hartSupports(Ext_Zve32x) = config extensions.Zve32x.supported
...
```

If you cant use the intended order there, just use the dependency order and leave a comment for now.

model/riscv_extensions.sail
```diff
@@ -332,6 +332,50 @@ mapping clause extensionName = Ext_Zvksg <-> "zvksg"
 function clause hartSupports(Ext_Zvksg) = hartSupports(Ext_Zvks) & hartSupports(Ext_Zvkg)
 function clause currentlyEnabled(Ext_Zvksg) = hartSupports(Ext_Zvksg)
 
+// Minimum Vector Length Standard Extensions
+enum clause extension = Ext_Zvl32b
+mapping clause extensionName = Ext_Zvl32b <-> "zvl32b"
+function clause hartSupports(Ext_Zvl32b) = config extensions.Zvl32b.supported : bool & (config extensions.V.vlen_exp : int >= 5)
```
jordancarlin: I'm not sure that it makes sense to expose the `Zvl*b` extensions in the config file. We already have `vlen_exp`, so we can just have the `hartSupports` clause set to true if `vlen_exp` meets the minimum required size. 

If we do keep them as options in the config file, then `riscv_validate_config.sail` needs to be updated to throw an error if one of them is enabled with a vlen that is too short. Right now the config option will be silently ignored, which is sure to be confusing for users.

model/riscv_extensions.sail
```diff
@@ -332,6 +332,50 @@ mapping clause extensionName = Ext_Zvksg <-> "zvksg"
...
+enum clause extension = Ext_Zvl1024b
+mapping clause extensionName = Ext_Zvl1024b <-> "zvl1024b"
+function clause hartSupports(Ext_Zvl1024b) = config extensions.Zvl1024b.supported : bool & (config extensions.V.vlen_exp : int >= 10)
+// Vector Extensions for Embedded Processors
+enum clause extension = Ext_Zve32x
+mapping clause extensionName = Ext_Zve32x <-> "zve32x"
+function clause hartSupports(Ext_Zve32x) = config extensions.Zve32x.supported
+enum clause extension = Ext_Zve32f
+mapping clause extensionName = Ext_Zve32f <-> "zve32f"
+function clause hartSupports(Ext_Zve32f) = config extensions.Zve32f.supported : bool & hartSupports(Ext_Zve32x) & hartSupports(Ext_F)
```
jordancarlin: These dependencies between extensions should be included in `riscv_validate_config.sail` so that extensions are not silently disabled if a config file enables Zve32f but not F. This applies to all of the dependencies added here.

model/riscv_extensions.sail
```diff
@@ -332,6 +332,50 @@ mapping clause extensionName = Ext_Zvksg <-> "zvksg"
...
+function clause hartSupports(Ext_Zve64f) = config extensions.Zve64f.supported : bool & hartSupports(Ext_Zve32f) & hartSupports(Ext_Zve64x)
+enum clause extension = Ext_Zve64d
+mapping clause extensionName = Ext_Zve64d <-> "zve64d"
+function clause hartSupports(Ext_Zve64d) = config extensions.Zve64d.supported : bool & hartSupports(Ext_Zve64f) & hartSupports(Ext_D)
+// Vector Extension for Minimal Half-Precision Floating-Point
+enum clause extension = Ext_Zvfhmin
+mapping clause extensionName = Ext_Zvfhmin <-> "zvfhmin"
+function clause hartSupports(Ext_Zvfhmin) = config extensions.Zvfhmin.supported : bool & hartSupports(Ext_Zve32f)
+// Vector Extension for Half-Precision Floating-Point
+enum clause extension = Ext_Zvfh
```
jordancarlin: All of these new extensions need to be added to the list in the README.

model/riscv_vext_control.sail
```diff
@@ -6,7 +6,24 @@
...
+function clause currentlyEnabled(Ext_Zvl512b)  = hartSupports(Ext_Zvl512b) 
+function clause currentlyEnabled(Ext_Zvl1024b) = hartSupports(Ext_Zvl1024b)
+
+function clause currentlyEnabled(Ext_Zve32x)   = hartSupports(Ext_Zve32x)   & currentlyEnabled(Ext_Zvl32b)
+function clause currentlyEnabled(Ext_Zve32f)   = hartSupports(Ext_Zve32f)   & currentlyEnabled(Ext_Zve32x)   & currentlyEnabled(Ext_F)
+function clause currentlyEnabled(Ext_Zve64x)   = hartSupports(Ext_Zve64x)   & currentlyEnabled(Ext_Zvl64b)   & currentlyEnabled(Ext_Zve32x)
+function clause currentlyEnabled(Ext_Zve64f)   = hartSupports(Ext_Zve64f)   & currentlyEnabled(Ext_Zve64x)   & currentlyEnabled(Ext_Zve32f)
+function clause currentlyEnabled(Ext_Zve64d)   = hartSupports(Ext_Zve64d)   & currentlyEnabled(Ext_Zve64f)   & currentlyEnabled(Ext_D)
+
+function clause currentlyEnabled(Ext_V) = hartSupports(Ext_V) & (misa[V] == 0b1) & (mstatus[VS] != 0b00) & currentlyEnabled(Ext_Zvl128b) & currentlyEnabled(Ext_Zve64d) & currentlyEnabled(Ext_Zicsr)
```
jordancarlin: All vector extensions are dependent on `mstatus[VS] != 0b00`, not just `V`. This condition should be moved to `currentlyEnabled(Ext_Zve32x)`.

jordancarlin: Handling `misa[V]` is going to be trickier than this. Most of the instructions are gated on the sub extensions now, but when the full V extension is supported, setting `misa[V] = 0` should presumably disable all vector instructions.

One potential solution is adding `misa[V] == 0b1` as a requirement for `currentlyEnabled(Ext_Zve32x)` if `hartSupports(Ext_V)`.

config/rv32d.json
wwwwwwOwO: Got it! I'll use the order you mentioned. And I'll do the dependency check in `riscv_validate_config.sail`.

model/riscv_extensions.sail
wwwwwwOwO: Gotcha, makes sense! I’ll remove it from the config file. (as the specification indicates that Zvl* extensions are primarily meant for stating hardware capabilities)

wwwwwwOwO:
> Hi @wwwwwwOwO, thanks for the update! Do you think it would be possible to break this PR into smaller ones, each focusing on just one change? I feel like this changes too many things at once, and splitting it up would really help with the review process.

Got it! I‘ll split it into smaller ones and fix things based on the latest comments. Thanks for all the helpful feedback!

wwwwwwOwO:
I've created the PR #1163 which only contains configuration aspect changes, and I’m planning to submit a follow-up PR focusing on instruction-level support (e.g., SEW/EEW gating). Do you think it’s a good time to open it now, or should I wait until this one is merged?

nadime15:
I think its ok to open it now and just say that @1163 needs to be merge first.

PR #990, opened 2025-05-30
add overload for regidx_offset

KotorinMinami:
For the first part of #779

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -282,7 +282,7 @@ function rF_or_X_D(i) = {
   else {
     let ridx = fregidx_to_regidx(i);
     assert(regidx_bits(ridx)[0] == bitzero);
-    if ridx == zreg then zeros() else X(regidx_offset(ridx, 0b00001)) @ X(ridx)
+    if ridx == zreg then zeros() else X(ridx + 0b00001) @ X(ridx)
   }
 }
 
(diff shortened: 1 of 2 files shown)


model/riscv_fdext_regs.sail
```diff
@@ -317,7 +317,7 @@ function wF_or_X_D(i, data) = {
     assert (regidx_bits(ridx)[0] == bitzero);
     if ridx != zreg then {
       X(ridx)                         = data[31..0];
```
Timmmm: Probably should reformat this... (One of the reasons I generally don't like aligning things.)

model/riscv_types.sail
```diff
@@ -28,7 +28,7 @@ type csreg = bits(12)  /* CSR addressing */
 
 function regidx_offset(Regidx(r) : regidx, o : bits(5)) -> regidx = Regidx(r + o)
```
Timmmm, line 29: I wonder if it would be a bit nicer using `range(0, 31)` here:

```
function regidx_offset(Regidx(r) : regidx, o : range(0, 31)) -> regidx = Regidx(r + o)

X(ridx + 1) = data[...
```

What do you think?

jordancarlin: We could also add both as overloads, so either can be used depending on the specific context.

Timmmm: Yeah although I think the only thing that this is ever used for is to add 1.

PR #991, opened 2025-05-30
Executable version of the Lean model

ineol:
This adds a new target `generated_lean_rv64d_executable` that does not use axioms nor any non-computable constructs so that it can be used as a library for the forthcoming elf binary runner.


tobiasgrosser (approved):
Yes, this is super helpful. Thank you, @ineol .

ineol:
@jordancarlin  thank you for the feedback! I changed the target name and added it to the CI (but without the lean build because it's very slow at the moment)

jordancarlin (changes requested):

ineol:
Done!

jordancarlin (approved):

model/CMakeLists.txt
```diff
@@ -478,28 +501,25 @@ foreach (xlen IN ITEMS 32 64)
...
-                        --lean-non-beq-type ast
-                        --lean-import-file ../handwritten_support/RiscvExtras.lean
-                        --config ${CMAKE_SOURCE_DIR}/config/default.json
-                        -o "Lean_${arch_uppercase}"
+                        ${lean_sail_common}
+                        ${lean_sail_default}
+                        ${sail_srcs}
+                )
+                add_custom_command(
+                    DEPENDS ${sail_srcs}
```
Timmmm, line 509: This is missing some dependencies - at least `config/default.json` and `handwritten_support/RiscvExtrasExecutable.lean` and `handwritte_supports/RiscvExtras.lean`. It means incremental builds won't work properly if you edit those files. Same for line 497.

Existing problem though so don't feel like you need to fix it in this PR - just pointing it out.

model/CMakeLists.txt
```diff
@@ -470,6 +470,29 @@ foreach (xlen IN ITEMS 32 64)
...
+                set(lean_sail_default
+                    --lean-noncomputable
+                    --lean-noncomputable-function encdec_forwards
+                    --lean-noncomputable-function encdec_backwards
+                    --lean-noncomputable-function encdec_forwards_matches
+                    --lean-noncomputable-function encdec_backwards_matches
+                    --lean-noncomputable-function encdec_compressed_forwards
+                    --lean-noncomputable-function encdec_compressed_backwards
+                    --lean-noncomputable-function encdec_compressed_forwards_matches
+                    --lean-noncomputable-function encdec_compressed_backwards_matches
```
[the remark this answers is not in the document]
tobiasgrosser: Actually, we do.

model/CMakeLists.txt
ineol: I fixed it in this PR since it's a quick change. Thank you for the remark!

tobiasgrosser:
@ineol, there seem to be some trailing white spaces that break the CI.

tobiasgrosser:
OK, this is good-to-go as well. Thank you all!

PR #994, opened 2025-05-30
Run Lean CI on push in addition to PRs

jordancarlin:
Caches are only reused if they are from the same branch or the master branch. Currently this never gets run on master, so the opam cache is not shared across PRs.


Timmmm (approved):

tobiasgrosser:
lgtm

tobiasgrosser (approved):

PR #996, opened 2025-05-30
Fix handling of .aqrl in assembly mapping

### LINKED ISSUE #995: Assembly clauses with maybe_aq, maybe_rl
In riscv_insts_base.sail:
```
val maybe_aq : bool <-> string
mapping maybe_aq = {
  true  <-> ".aq",
  false <-> ""
}

val maybe_rl : bool <-> string
mapping maybe_rl = {
  true  <-> ".rl",
  false <-> ""
}

mapping clause assembly = LOAD(imm, rs1, rd, is_unsigned, size, aq, rl)
  <-> "l" ^ size_mnemonic(size) ^ maybe_u(is_unsigned) ^ maybe_aq(aq) ^ maybe_rl(rl) ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_signed_12(imm) ^ "(" ^ reg_name(rs1) ^ ")"
```
When both `aq` and `rl` flags are true for `LOAD`, the assembly output incorrectly shows `.aq.rl` instead of the correct `.aqrl`.

pmundkur:
This would previously have printed `.aq.rl` whereas the correct assembly is `.aqrl`.

Fixes #995.

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -302,26 +302,20 @@ function clause execute (LOAD(imm, rs1, rd, is_unsigned, width, aq, rl)) = {
   }
 }
 
-val maybe_aq : bool <-> string
-mapping maybe_aq = {
-  true  <-> ".aq",
-  false <-> ""
+mapping maybe_aqrl : (bool, bool) <-> string = {
+  (true, true)   <-> ".aqrl",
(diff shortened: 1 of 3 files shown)


arichardson (approved):

jordancarlin (approved):
Simple fix.

PR #1001, opened 2025-06-03
Generate more accurate ISA information for device tree.

### LINKED ISSUE #992: More accurate ISA extension information in generated device tree
#972 uses a default hard-coded minimal ISA string in the device tree that doesn't reflect the extensions enabled in the model configuration.

A more accurate string would need to follow this specification (from https://github.com/riscv/sail-riscv/pull/972#discussion_r2114415756):
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings/riscv/extensions.yaml

pmundkur:
- Add the `riscv,isa-base` and `riscv,isa-extensions` properties, and make the value of `riscv,isa` more accurate.

- Introduce an extensionName mapping for extensions, and use it in ISA string generation.

- Add a formatting enum for generation that could be extended for other formats of the ISA string.

- While here, do a minor cleanup of `riscv_extensions.sail` to fix ordering and comments.

Fixes #992.

### DIFF
--- a/model/riscv_device_tree.sail
+++ b/model/riscv_device_tree.sail
@@ -19,6 +19,156 @@ function mmu_type() -> string = {
   }
 }
 
+// Generators for Device-Tree source corresponding to the model
+// configuration.
+// NOTE: These assume a valid model configuration.
+
+// Generates the value for the "riscv,isa-base" property.
+function generate_isa_base() -> string = {
+  "rv" ^ dec_str(xlen) ^ "i"
+}
+
+/* Some notes from the "ISA Extension Naming Convections" section of
+   the manual:
+
+  . The first letter following the "Z" conventionally indicates the
+    most closely related alphabetical extension category,
+    IMAFDQLCBKJTPVH.  If multiple "Z" extensions are named, they
+    should be ordered first by category, then alphabetically within a
+    category—for example, "Zicsr_Zifencei_Ztso".
+
+  . All multi-letter extensions, including those with the "Z" prefix, must be
+    separated from other multi-letter extensions by an underscore, e.g.,
+    "RV32IMACZicsr_Zifencei".
+
+  . Standard supervisor-level extensions (prefixed with either 'Ss' or
+    'Sv') should be listed after standard unprivileged extensions, and
+    like other multi-letter extensions, must be separated from other
+    multi-letter extensions by an underscore. If multiple
+    supervisor-level extensions are listed, they should be ordered
+    alphabetically.
+
+  . Standard machine-level extensions (prefixed with the letters "Sm")
+    should be listed after standard lesser-privileged extensions, and
+    like other multi-letter extensions, must be separated from other
+    multi-letter extensions by an underscore. If multiple
+    machine-level extensions are listed, they should be ordered
+    alphabetically.
+
+  Once the Sail configuration supports extension version numbers, we could
+  add them to the generated isa string.
+
+  The device-tree specification for the ISA string is at:
+  https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings/riscv/extensions.yaml
+
+  As the above document uses lower-case, the generator below does too.
+*/
+
+// Generates the value for the "riscv,isa-extensions" property
+// (if `as_isa_ext == false`) or the "riscv,isa" property (otherwise).
+// NOTE: The "riscv,isa" property is marked deprecated but is probably
+// still widely used.
+
+function dt_wrap(s : string, as_isa_ext : bool) -> string = {
+  if   as_isa_ext
+  then ", \"" ^ s ^ "\""
+  // In "riscv,isa", single letter extensions are concatenated
+  // without a separator.
+  else if string_length(s) == 1
+  then s
+  else "_" ^ s
+}
+
+function generate_isa_string(as_isa_ext : bool) -> string = {
+  (if   as_isa_ext
+   then "\"i\""
+   else "rv" ^ dec_str(xlen) ^ "i")
+
+// First append the single letter extensions.
+
+^ (if hartSupports(Ext_M) then dt_wrap("m", as_isa_ext) else "")
+^ (if hartSupports(Ext_A) then dt_wrap("a", as_isa_ext) else "")
+^ (if hartSupports(Ext_F) then dt_wrap("f", as_isa_ext) else "")
+^ (if hartSupports(Ext_D) then dt_wrap("d", as_isa_ext) else "")
+^ (if hartSupports(Ext_C) then dt_wrap("c", as_isa_ext) else "")
+  // 'b' does not appear in the above device-tree spec for "riscv,isa-extensions".
+^ (if hartSupports(Ext_B) & not(as_isa_ext) then "b"   else "")
+^ (if hartSupports(Ext_V) then dt_wrap("v", as_isa_ext) else "")
+// S and U are not valid extensions in the device-tree, whereas H is.
+
+// Append the Z extensions, ordered by category and then alphabetically.
+
+^ (if hartSupports(Ext_Zicbom) then dt_wrap("zicbom", as_isa_ext) else "")
+^ (if hartSupports(Ext_Zicboz) then dt_wrap("zicboz", as_isa_ext) else "")
+^ (if hartSupports(Ext_Zicntr) then dt_wrap("zicntr", as_isa_ext) else "")
+^ (if hartSupports(Ext_Zicond) then dt_wrap("zicond", as_isa_ext) else "")
+^ (if hartSupports(Ext_Zicsr)  then dt_wrap("zicsr", as_isa_ext)  else "")
+^ (if hartSupports(Ext_Zifencei) then dt_wrap("zifencei", as_isa_ext)  else "")
+^ (if hartSupports(Ext_Zihpm)  then dt_wrap("zihpm", as_isa_ext)  else "")
+^ (if hartSupports(Ext_Zimop)  then dt_wrap("zimop", as_isa_ext)  else "")
+
+// Zmmul is not a device-tree extension.
+
+^ (if hartSupports(Ext_Zaamo)  then dt_wrap("zaamo", as_isa_ext)  else "")
+^ (if hartSupports(Ext_Zabha)  then dt_wrap("zabha", as_isa_ext)  else "")
+^ (if hartSupports(Ext_Zalrsc) then dt_wrap("zalrsc", as_isa_ext) else "")
+^ (if hartSupports(Ext_Zawrs)  then dt_wrap("zawrs", as_isa_ext)  else "")
+
+^ (if hartSupports(Ext_Zfa)    then dt_wrap("zfa", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zfh)    then dt_wrap("zfh", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zfhmin) then dt_wrap("zfhmin", as_isa_ext) else "")
+^ (if hartSupports(Ext_Zfinx)  then dt_wrap("zfinx", as_isa_ext)  else "")
+
+// Zdinx and Zhinx{,min} are not extensions in device-tree.
+
+^ (if hartSupports(Ext_Zca)    then dt_wrap("zca", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zcb)    then dt_wrap("zcb", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zcd)    then dt_wrap("zcd", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zcf)    then dt_wrap("zcf", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zcmop)  then dt_wrap("zcmop", as_isa_ext)  else "")
+
+^ (if hartSupports(Ext_Zba)    then dt_wrap("zba", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zbb)    then dt_wrap("zbb", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zbc)    then dt_wrap("zbc", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zbkb)   then dt_wrap("zbkb", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zbkc)   then dt_wrap("zbkc", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zbkx)   then dt_wrap("zbkx", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zbs)    then dt_wrap("zbs", as_isa_ext)    else "")
+
+^ (if hartSupports(Ext_Zknd)   then dt_wrap("zknd", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zkne)   then dt_wrap("zkne", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zknh)   then dt_wrap("zknh", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zkr)    then dt_wrap("zkr", as_isa_ext)    else "")
+^ (if hartSupports(Ext_Zksed)  then dt_wrap("zksed", as_isa_ext)  else "")
+^ (if hartSupports(Ext_Zksh)   then dt_wrap("zksh", as_isa_ext)   else "")
+
+^ (if hartSupports(Ext_Zvbb)   then dt_wrap("zvbb", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zvbc)   then dt_wrap("zvbc", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zvkb)   then dt_wrap("zvkb", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zvkg)   then dt_wrap("zvkg", as_isa_ext)   else "")
+^ (if hartSupports(Ext_Zvkned) then dt_wrap("zvkned", as_isa_ext) else "")
+^ (if hartSupports(Ext_Zvknha) then dt_wrap("zvknha", as_isa_ext) else "")
+^ (if hartSupports(Ext_Zvknhb) then dt_wrap("zvknhb", as_isa_ext) else "")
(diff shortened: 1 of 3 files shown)


model/riscv_device_tree.sail
```diff
@@ -6,6 +6,233 @@
...
+^ (if hartSupports(Ext_Zicbom) then dt_prefix("zicbom") else "")
+^ (if hartSupports(Ext_Zicboz) then dt_prefix("zicboz") else "")
+^ (if hartSupports(Ext_Zicntr) then dt_prefix("zicntr") else "")
+^ (if hartSupports(Ext_Zicond) then dt_prefix("zicond") else "")
+^ (if hartSupports(Ext_Zicsr)  then dt_prefix("zicsr")  else "")
+^ (if hartSupports(Ext_Zifencei) then dt_prefix("zifencei")  else "")
+^ (if hartSupports(Ext_Zihpm)  then dt_prefix("zihpm")  else "")
+^ (if hartSupports(Ext_Zimop)  then dt_prefix("zimop")  else "")
+
+// Zmmul is not a device-tree extension.
```
arichardson: I wonder if this matters? I'd expect the kernel to ignore unknown extensions?

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,233 @@
...
+^ (if hartSupports(Ext_F) then "f" else "")
+^ (if hartSupports(Ext_D) then "d" else "")
+^ (if hartSupports(Ext_C) then "c" else "")
+^ (if hartSupports(Ext_B) then "b" else "")
+^ (if hartSupports(Ext_V) then "v" else "")
+// S and U are not valid extensions in the device-tree, whereas H is.
+
+// Append the Z extensions, ordered by category and then alphabetically.
+
+^ (if hartSupports(Ext_Zicbom) then dt_prefix("zicbom") else "")
```
arichardson:
```suggestion
^ (if hartSupports(Ext_Zicbom) then "_zicbom" else "")
```
Not sure this extra function makes things clearer? underscore is the standard separator in the RISC-V ISA string, so no need for `dt_*`.

arichardson (commented):
This is great, just a few questions.

model/riscv_device_tree.sail
pmundkur: Probably true.  Spike refuses unrecognized extensions (but it does handle "Zmmul"), so it might be best to stick to the spec?

model/riscv_device_tree.sail
pmundkur: `dt_wrap` is a bit painful to inline due to the escaped quotes.  `dt_prefix` is easier to manually inline, I can remove it if you prefer.

model/riscv_device_tree.sail
pmundkur: https://github.com/riscv-software-src/riscv-isa-sim/blob/4c870d063dbbaeb4dc7007fe5c2a1bf8b00a767e/disasm/isa_parser.cc#L405

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,233 @@
...
+// Append the Machine mode extensions, ordered alphabetically.
+
+// Smcntrpmf is not a device-tree extension.
+}
+
+// Generates the value for the "riscv,isa-extensions" property.
+
+function dt_wrap(s : string) -> string = ", \"" ^ s ^ "\""
+
+function generate_isa_extensions() -> string = {
```
jordancarlin: Overall, this seems great. I'm not a fan of maintaining this list in two different functions though. Is there any way we could have a single function that takes an input to determine what delimiter is used in the output string?

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,233 @@
...
+// Handle svade and svadu here when ready.
+^ (if hartSupports(Ext_Svinval)   then dt_wrap("svinval")   else "")
+^ (if hartSupports(Ext_Svnapot)   then dt_wrap("svnapot")   else "")
+^ (if hartSupports(Ext_Svpbmt)    then dt_wrap("svpbmt")    else "")
+
+// Append the Hypervisor extensions, ordered alphabetically.
+
+// Append the Machine mode extensions, ordered alphabetically.
+
+// Smcntrpmf is not a device-tree extension.
```
radimkrcmar: It's an ISA extension, so I think it belongs in DT.

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,233 @@
...
+// Generates the value for the "riscv,isa-extensions" property.
+
+function dt_wrap(s : string) -> string = ", \"" ^ s ^ "\""
+
+function generate_isa_extensions() -> string = {
+  "\"i\""
+
+// First append the single letter extensions.
+
+^ (if hartSupports(Ext_M) then dt_wrap("m") else "")
```
radimkrcmar: This pattern makes me quite sad.  Can't we define a list with `extension, string` tuples and do a loop over it both in generate_isa and generate_isa_extensions?
(Or at least generate one from the other.)

We would ideally also be checking that the list is exhaustive.

model/riscv_device_tree.sail
radimkrcmar: Zmmul is an ISA extension, so it belongs in DT.  Sail does support it (it's an subset of M), so I think we should put it here (similarly like we do with Zca/... that are a subset of C).

model/riscv_device_tree.sail
Timmmm: Yeah I think what we could do is have a mapping from extension to extension name:

```
mapping clause extension_name = Ext_Zibcom <-> "zicbom"
```

and then a function like

```
function extension_isa_string(ext : Extension, prefix : string) -> string = if hartSupports(ext) then prefix ^ extension_name(ext) else ""
```

Then the function can simply be

```
    ^ extension_isa_string(Ext_A, prefix)
    ^ extension_isa_string(Ext_F, prefix)
    ^ extension_isa_string(Ext_D, prefix)
    ^ extension_isa_string(Ext_C, prefix)
...
```

This would also solve @jordancarlin's point about not duplicating the function.

A slightly unsatisfying thing here is that we're generating the lowercase string, whereas the ISA string is supposed to be Title_Case. It seems like GCC and device tree didn't get that memo and require lowercase.

If we could generate it as uppercase and then just lowercase the whole string that would be great but I would guess Sail doesn't have a built-in lowercase function.

model/riscv_device_tree.sail
pmundkur: I've pushed a single-function version.

model/riscv_device_tree.sail
radimkrcmar: The mapping sounds much better than a tuple!  The manual says that the ISA string is case-insensitive, so at least we're not too incorrect.

Now if Sail could list all members of an enum, we could do something like:
```
reduce("rv64", join_underscore,
       map(extension_name,
           filter(hartSupports,
                  enum_to_list(extension))))
```
And be sure that we never miss any extension.

model/riscv_device_tree.sail
pmundkur: How definitive is the RISC-V device-tree extensions spec in the Linux Documentation?   `Smctrpmf` does not appear in it, nor does `Zmmul`.

model/riscv_device_tree.sail
pmundkur: That doesn't quite work due to the difference in handling the single-letter extensions and the prefix.  The "riscv,isa" and "riscv,isa-extensions" are annoyingly different.  

One option is to handle the _deprecated_ "riscv,isa" property with a list of supported single-letter extensions only and then fill out the "isa-extensions" list fully.  
For e.g.: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/riscv/boot/dts/sifive/fu740-c000.dtsi#n60

But I'm not sure that will work for other kernels like Zephyr, which doesn't seem to be using "riscv,isa-extensions" currently.

radimkrcmar: Yeah...  You already handle single and multi letter by looking at the string length, so if you abstract the `if` as well:
```
function short_and_descriptive(ext : extension) -> string {
  if hartSupports(ext) then dt_wrap(ext, true)  else "" # a second function would have false there
}
```
Then `dt_wrap` function can use a `mapping extention -> bool` whether the extension should be emitted based on the `as_isa_ext`.

(I'll have to look again at the OSes and flexibility that DT gives us, because I think we could just put everything there and all DT consumers should accept it.)

model/riscv_device_tree.sail
radimkrcmar: DT as a format already outgrew Linux -- I think the Linux documentation is mostly a best effort and others are free to extend it.   (e.g. Sail could be supplying the DT to M-mode opensbi.)

Adding ISA extensions should be harmless, because the extension format is defined and the list will continuously expand anyway. Old software should not break when seeing extensions that didn't exist at the compilation time.

model/riscv_device_tree.sail
pmundkur: There is also the ordering issue for single-letter extensions: they need to come first for "riscv,isa".   A general `enum_to_list` function cannot guarantee that ordering from the Sail `extension` type.

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,156 @@
...
+}
+
+function generate_isa_string(as_isa_ext : bool) -> string = {
+  (if   as_isa_ext
+   then "\"i\""
+   else "rv" ^ dec_str(xlen) ^ "i")
+
+// First append the single letter extensions.
+
+^ (if hartSupports(Ext_M) then dt_wrap("m", as_isa_ext) else "")
```
arichardson:
```suggestion
^ dt_wrap(hartSupports(Ext_M), "m", as_isa_ext)
```

How about passing a boolean as the first argument to dt_wrap  and push the if into that function? Then we can avoid all the if/else in this function?

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,156 @@
...
+
+  As the above document uses lower-case, the generator below does too.
+*/
+
+// Generates the value for the "riscv,isa-extensions" property
+// (if `as_isa_ext == false`) or the "riscv,isa" property (otherwise).
+// NOTE: The "riscv,isa" property is marked deprecated but is probably
+// still widely used.
+
+function dt_wrap(s : string, as_isa_ext : bool) -> string = {
```
arichardson:
```suggestion
function dt_wrap(supported : bool, s : string, as_isa_ext : bool) -> string = {
  if not(supported) then return "";
```
Something like this?

model/riscv_device_tree.sail
Timmmm: Actually you can iterate through Sail enums. It auto-generates `num_of_....` and `..._of_num` functions. So this probably could be reduced to a for loop.

radimkrcmar: Yeah, I assumed that sail will capture the order in the source code.  All the other extensions need to be ordered as well, so we would just have carefully place them, like we do in this functions.

I agree that it might be putting too many assumptions on sail, so it's probably better to proceed with the current explicit approach.

pmundkur:
@Alasdair the Sail compiler (both 0.19 and the current master) is taking a very long time (longer than an hour for current master) to compile this.   I didn't have time to run either compile to completion.

pmundkur:
@Alasdair the CI failed after 6 hours.  Is this triggering an infinite loop?

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,158 @@
...
+  add them to the generated isa string.
+
+  The device-tree specification for the ISA string is at:
+  https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings/riscv/extensions.yaml
+
+  As the above document uses lower-case, the generator below does too.
+*/
+
+// Generates the value for the "riscv,isa-extensions" property
+// (if `as_isa_ext == false`) or the "riscv,isa" property (otherwise).
```
Timmmm: Isn't this the wrong way round?

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,158 @@
...
+// NOTE: The "riscv,isa" property is marked deprecated but is probably
+// still widely used.
+
+function dt_wrap(ext : extension, as_isa_ext : bool) -> string = {
```
Timmmm: I think I would probably add an enum for `as_isa_ext` - we have a similar function and it already has two variants for GCC and riscv-config which expect different things. Something like

```
function wrap_ext_name(ext : extension, target : isa_string_target) -> string = {
  if not(hartSupports(ext)) then return "";
  let s = extensionName(ext);
  match target {
     DeviceTree_IsaExtensions => ", \"" ^ s ^ "\"",
     _ => if string_length(s) == 1 then s else "_" ^ s,
  }
}
```

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,158 @@
...
+}
+
+function generate_isa_string(as_isa_ext : bool) -> string = {
+  (if   as_isa_ext
+   then "\"i\""
+   else "rv" ^ dec_str(xlen) ^ "i")
+
+// First append the single letter extensions.
+
+^ dt_wrap(Ext_M, as_isa_ext)
```
Timmmm: I think I'd still indent these.

arichardson (approved):

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,175 @@
...
+// Generates the value for the "riscv,isa-base" property.
+function generate_isa_base() -> string = {
+  "rv" ^ dec_str(xlen) ^ "i"
+}
+
+// Generates an ISA string in the appropriate format.
+
+// NOTE: The "riscv,isa" property is marked deprecated but is probably
+// still widely used.
+enum ISA_Format = {DeviceTree_ISA, DeviceTree_ISA_Extensions}
```
jordancarlin: What about naming the `DeviceTree_ISA` format something more generic like `Lowercase_Canonical`? It's not a device tree specific format and has uses elsewhere (would be nice for Sail to be able to spit out the `march` string for a given config, for example).

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,175 @@
...
+  ^ ext_wrap(Ext_Zalrsc, fmt)
+  ^ ext_wrap(Ext_Zawrs, fmt);
+
+  let zf_exts = ""
+  ^ ext_wrap(Ext_Zfa, fmt)
+  ^ ext_wrap(Ext_Zfh, fmt)
+  ^ ext_wrap(Ext_Zfhmin, fmt)
+  ^ ext_wrap(Ext_Zfinx, fmt);
+
+  // Zdinx and Zhinx{,min} are not extensions in device-tree.
```
jordancarlin: Unless there are known issues with including certain extensions in the device tree, I'd be tempted to just include all extensions that the model supports (even if they aren't in the Linux document). As things continue to evolve, existing extensions might get added to the device tree, and it would be easy to forget to keep an eye on that and never add them here.

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,175 @@
...
+  ^ ext_wrap(Ext_Zvbb, fmt)
+  ^ ext_wrap(Ext_Zvbc, fmt)
+  ^ ext_wrap(Ext_Zvkb, fmt)
+  ^ ext_wrap(Ext_Zvkg, fmt)
+  ^ ext_wrap(Ext_Zvkned, fmt)
+  ^ ext_wrap(Ext_Zvknha, fmt)
+  ^ ext_wrap(Ext_Zvknhb, fmt)
+  ^ ext_wrap(Ext_Zvksh, fmt);
+
+  // Append the Supervisor extensions, ordered alphabetically.
```
jordancarlin, line 165: These comments are probably unnecessary now that we have the extension groups. `s_exts` implies these are the supervisor extensions.

pmundkur: I kept the comment for the reminder to keep it alphabetically ordered.

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,173 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
+// Generators for Device-Tree source corresponding to the model
+// configuration.
+// NOTE: These assume a valid model configuration.
+
+// Generates the value for the "riscv,isa-base" property.
+function generate_isa_base() -> string = {
+  "rv" ^ dec_str(xlen) ^ "i"
```
radimkrcmar, line 28: We'll eventually have `E` too, so I think that treating `I` as a normal extension is fine.

radimkrcmar (approved):
My comment about `I` and `E` can be ignored.  I think this PR is good to go.

In the future, I would like to add the `Svbare`,` Sv32`, `Sv39`, `Sv48`, and `Sv57` extensions too.  (And maybe even `S` and `U`.)
Do you have a list of other ISA extensions that are missing in ISA string?

pmundkur:
> Do you have a list of other ISA extensions that are missing in ISA string?

Other than `S` and `U`, it should now cover the extensions in `riscv_extensions.sail` (i.e. all currently supported extensions).

model/riscv_device_tree.sail
```diff
@@ -19,6 +19,173 @@ function mmu_type() -> string = {
...
+
+  . The first letter following the "Z" conventionally indicates the
+    most closely related alphabetical extension category,
+    IMAFDQLCBKJTPVH.  If multiple "Z" extensions are named, they
+    should be ordered first by category, then alphabetically within a
+    category—for example, "Zicsr_Zifencei_Ztso".
+
+  . All multi-letter extensions, including those with the "Z" prefix, must be
+    separated from other multi-letter extensions by an underscore, e.g.,
+    "RV32IMACZicsr_Zifencei".
```
jrtc27, line 46: Doesn't this implementation put an underscore before the first multi-letter extension though? It's valid to do that, you can put redundant underscores between extensions if you want to, but it's odd to document what the manual says when you don't quite follow it and don't document how it differs.

Also P is special (due to being both the major/minor separator and a single-letter extension), which isn't noted here. Fine for now, but that will need handling.

radimkrcmar: We're not versioning the extensions, and ISA has wisely abandoned that idea of versioned extensions, so I think we don't need to worry about special handling of the P extension.

Improving the comments makes sense.

pmundkur: Good catch, thanks.  I'll document the difference in a comment.

pmundkur: #1025

PR #1003, opened 2025-06-04
Add risc-v vector extension tests

nadime15:
This commit adds a submodule containing precompiled tests (around ~1800 tests) for the risc-v vector extension and integrates them into the CMake testing workflow.

Due to their long runtime, these tests are executed weekly via scheduled CI runs or can be triggered manually. Developers can also run these tests locally by enabling the `RVV_TESTS` flag during the CMake configuration.

These tests were compiled with:
https://github.com/chipsalliance/riscv-vector-tests/tree/c093610

and with the following ISA options:
`  --isa=rv64gcv_zvl128b_zve64d_zfh_zfhmin_zvfh`

Configurated with:
  VLEN=128 and ELEN=64

This is still a work in progress, but I would appreciate early feedback and would like to discuss a few points. For now, I have hosted the binaries myself, but the idea is to move them to an official RISC-V repository soon.

Discussion points:

1. Should we rename the binaries to include a .elf suffix, similar to what we do for riscv-tests?
2. Should we also generate disassembly dumps (*.dump) like in riscv-tests?
3. I set the default VLEN in config.json to 128 (7) to avoid maintaining multiple JSON files.
4. Personally, I don't think it makes sense to run these tests on every push, so I added a weekly cron job instead. They can be manually triggered by developers or by us in case someone is pushing some changes that effect the vector extension.
5. I excluded the tests for vector crypto extension because I though I would save much more time but I think at the end I might have saved 2-3 minutes. I think it might makes sense to add them back to the tests?
6. I reuse existing workflows, which makes no use of caches etc. Should I create a whole seperate workflow or are people ok with the idea to reuse the existing workflow.

The whole workflow takes around ~40 Minutes, here is an example:
[https://github.com/nadime15/sail-riscv/actions/runs/15445714226/job/43474657958](https://github.com/nadime15/sail-riscv/actions/runs/15445714226/job/43474657958)

### DIFF
--- a/.github/workflows/run-riscv-tests.yml
+++ b/.github/workflows/run-riscv-tests.yml
@@ -12,14 +12,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - TEST_SUITE: RISCV_TESTS
-          - TEST_SUITE: RISCV_VECTOR_TESTS_V128_E32
-          - TEST_SUITE: RISCV_VECTOR_TESTS_V128_E64
-          - TEST_SUITE: RISCV_VECTOR_TESTS_V256_E32
-          - TEST_SUITE: RISCV_VECTOR_TESTS_V256_E64
-          - TEST_SUITE: RISCV_VECTOR_TESTS_V512_E32
-          - TEST_SUITE: RISCV_VECTOR_TESTS_V512_E64
+        TEST_SUITE:
+          [
+            RISCV_TESTS,
+            RISCV_VECTOR_TESTS_V128_E32,
+            RISCV_VECTOR_TESTS_V128_E64,
+            RISCV_VECTOR_TESTS_V256_E32,
+            RISCV_VECTOR_TESTS_V256_E64,
+            RISCV_VECTOR_TESTS_V512_E32,
+            RISCV_VECTOR_TESTS_V512_E64,
+          ]
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4


arichardson:
Since this already set up as a weekly run would it be possible to build it as part of the CI flow? Using something like ExternalProject_Add() with a fixed git commit hash?

nadime15:
I think we decided to add the tests as a git submodule instead of using an external project. I purposely didn’t fix it to a specific commit hash so it tracks master. But honestly, I don't mind either way.

Alasdair:
I think the idea was we could create a separate repository that would pull these tests in as a submodule, then that repository would have a weekly job that would build the tests how we need them. That job could upload them as a release that we could pull into a job on this repo to test more regularly.

arichardson:
> I think the idea was we could create a separate repository that would pull these tests in as a submodule, then that repository would have a weekly job that would build the tests how we need them. That job could upload them as a release that we could pull into a job on this repo to test more regularly.

That's also fine. Since git is not a great way of storing binaries, I'd much prefer having it stored as a github release.

nadime15:
> That's also fine. Since git is not a great way of storing binaries, I'd much prefer having it stored as a github release.

Sure sounds good!

> I think the idea was we could create a separate repository that would pull these tests in as a submodule, then that repository would have a weekly job that would build the tests how we need them. That job could upload them as a release that we could pull into a job on this repo to test more regularly.


I think I dont understand the part with the submodule then. What do you mean by "...a separate repository that would pull these tests in as a submodule, then that repository would have a weekly job that would build the tests how we need them." 

The tests are all auto generated by a script and not manually written, so what exactly do you wanna pull in as a submodule and then build? 

From what I understand, we create a separate repository that builds different versions of the vector tests using the vector test suite (with varying VLEN, XLEN, etc.), and publishes weekly releases (probably rather monthly releases because the rvv test suite is basically never changing).  

And finally, we use these releases as part of a CI job in this project.

pmundkur:
The separate repo will pull in `riscv-tests`, `riscv-vector-tests` and perhaps other test repos as submodules.

nadime15:
Couldn't I just install them via Github Actions and then build and release them? Whats the advantage of having them as submodules?

arichardson:
> Couldn't I just install them via Github Actions and then build and release them? Whats the advantage of having them as submodules?

I agree there is no need to have them as submodules. In my experience submodules just create lots of pain

jordancarlin:
The benefit of submodules in the testing repo instead of just cloning with GitHub Actions during the CI jobs is that it pins the version of the tests we are using to a specific commit. We probably don't want the tests updating randomly because that could introduce unexpected failures. We're better off controlling when we update the sources.

Technically we could pin the version the CI job installs, but if we do it with submodules we could use something like Dependabot to automatically open a PR when there is an updated version of any of the repos available.

pmundkur:
> Discussion points:
> 
>     1. Should we rename the binaries to include a .elf suffix, similar to what we do for riscv-tests?

Not sure why those had an `.elf` suffix; perhaps easier to glob from a script maybe?  As long as the files are easily listed in CMake or a script, it shouldn't matter.

>     2. Should we also generate disassembly dumps (*.dump) like in riscv-tests?

That's not needed.

>     3. I set the default VLEN in config.json to 128 (7) to avoid maintaining multiple JSON files.

The repository should have a script to generate the appropriate configs, and a script/CMakeLists to run the simulator (or _simulators_ until we have a unified build) with the appropriate configs for the tests.

>     4. Personally, I don't think it makes sense to run these tests on every push, so I added a weekly cron job instead. They can be manually triggered by developers or by us in case someone is pushing some changes that effect the vector extension.

Perhaps a daily job is also ok for the vector tests, depending on how long they take.  A week might be too long.

>     5. I excluded the tests for vector crypto extension because I though I would save much more time but I think at the end I might have saved 2-3 minutes. I think it might makes sense to add them back to the tests?

Agreed, it makes sense to add back.

>     6. I reuse existing workflows, which makes no use of caches etc. Should I create a whole seperate workflow or are people ok with the idea to reuse the existing workflow.
> 
> 
> The whole workflow takes around ~40 Minutes, here is an example: https://github.com/nadime15/sail-riscv/actions/runs/15445714226/job/43474657958

Don't know enough to comment.

pmundkur:
Here are some test repos that would be useful to integrate under a unified sail-riscv-test repo.  I'm not sure which of these are useful/maintained:

https://github.com/josecm/riscv-hyp-tests
https://github.com/avpatel/xvisor-next/tree/master/tests/riscv
https://github.com/tenstorrent/riscv_arch_tests
https://github.com/riscv-ovpsim/imperas-riscv-tests
https://gitlab.com/kvm-unit-tests/kvm-unit-tests/-/tree/master/riscv?ref_type=heads
https://github.com/luisccc/riscv-iopmp-tests
https://github.com/ucb-bar/riscv-torture

jordancarlin:
If the whole workflow only takes 40 minutes, I'd be tempted to have it run on all PRs. The Lean workflow is 30-45+ minutes (depending on the cache), so it wouldn't even extend the total CI time for a PR.

nadime15:
I have reworked the flow, made a few changes and followed @Alasdair description. (Still needs some work, like clearer step names and general cleanup etc.). I think it doesn’t really make sense to recompile the tests weekly, so instead I followed @jordancarlin  idea to use Dependabot. I am not fully there yet, but I have at least enabled Dependabot. Now, Dependabot will monitor changes in the underlying submodules and open a pull request if one of them updates (so far, riscv-test and riscv-vector-test).

A feature I am working on, but have not fully implemented is to auto-merge Dependabots PR if an additional workflow successfully recompiles all tests with the updated commit. If that passes, we automatically merge the PR (which updates the submodule hash) and create a new release based on the latest riscv-test or riscv-vector-test commit.

I created a new [repository](https://github.com/nadime15/riscv-test-generator) that compiles tests from riscv-test and riscv-vector-test for VLEN = [512, 256, 128] and ELEN = [64, 32]. Thanks to parallel execution, this works pretty well (it takes around 3h in total). See the image: 

Bildschirmfoto vom 2025-06-08 18-35-08

I needed to upload all binaries as separate chunks in the [release](https://github.com/nadime15/riscv-test-generator/releases/tag/2025-06-08) because bundling them into a single tarball exceeds GitHub's 2GB per file limit.

This commit replaces the old workflow I had with a new one that always fetches the latest release and runs (in theory) the full test suite. For now, to save time, I’m only testing riscv-vector-test with VLEN=512 and ELEN=64.

Here is the [link](https://github.com/nadime15/sail-riscv/actions/runs/15523078830/job/43698891041) to one workflow, which adds about ~2,500 tests.

I was wondering how to handle different VLEN and ELEN values? Should I dynamically create new config files directly in the workflow with the corresponding parameters? We can also simply use the default config file and stick to VLEN = 512 and ELEN = 64 without testing different VLEN/ELEN values.

Edit: I just noticed that the [workflow](https://github.com/nadime15/sail-riscv/actions/runs/15523078830/job/43698891041) has failed due to:
`
System.IO.IOException: No space left on device : '/home/runner/runners/2.325.0/_diag/Worker_20250608-221729-utc.log' Unhandled exception. System.IO.IOException: No space left on device : ...`

It stopped at around `687/3097 Test`

jordancarlin:
This is definitely moving in the right direction. A few thoughts/comments:

> I needed to upload all binaries as separate chunks in the [release](https://github.com/nadime15/riscv-test-generator/releases/tag/2025-06-08) because bundling them into a single tarball exceeds GitHub's 2GB per file limit.

I think it makes more sense to upload each suite as a separate release artifact. So one tarball for riscv-tests and one for each of the riscv-vector-test variants that you are compiling. Then we can have a matrix workflow in the sail repo where a separate job runs each set of tests to cut down the runtime. It should also help avoid file size issues.

> I was wondering how to handle different VLEN and ELEN values? Should I dynamically create new config files directly in the workflow with the corresponding parameters? We can also simply use the default config file and stick to VLEN = 512 and ELEN = 64 without testing different VLEN/ELEN values.

What about using `jq` to modify VLEN/ELEN in the config file directly in the workflow? Seems like the simplest solution. I think testing multiple VLEN/ELEN combinations has value.

>Edit: I just noticed that the [workflow](https://github.com/nadime15/sail-riscv/actions/runs/15523078830/job/43698891041) has failed due to:
> `System.IO.IOException: No space left on device : '/home/runner/runners/2.325.0/_diag/Worker_20250608-221729-utc.log' > Unhandled exception. System.IO.IOException: No space left on device : ...`
> 
> It stopped at around 687/3097 Test

I have a script that I've been using in other GitHub CI jobs that increases the available space on the runner from ~24GB to ~61GB by deleting lots of unneeded things. I think that should be plenty of space for this. https://github.com/openhwgroup/cvw/blob/main/.github/scripts/cli-space-cleanup.sh

nadime15:
> I think it makes more sense to upload each suite as a separate release artifact. So one tarball for riscv-tests and one for each of the riscv-vector-test variants that you are compiling. Then we can have a matrix workflow in the sail repo where a separate job runs each set of tests to cut down the runtime. It should also help avoid file size issues.

Yeah, this would kinda work too, the only issue is that once you set VLEN ≥ 1028 (which we might do at some point), you hit the same problem again. No idea if that really matters or if people even care about those cases, but I think it is worth mentioning.

I was thinking if we could have one job that grabs all the parts, reconstructs them into a single tarball, unpacks it, uploads the artifact, and then launches multiple jobs. It’s basically what you’re describing, just with a slightly different starting point.

>   What about using jq to modify VLEN/ELEN in the config file directly in the workflow? Seems like the simplest solution. I think testing multiple VLEN/ELEN combinations has value.

Ahh ok, I was not familiar with that tool, makes sense!

> I have a script that I've been using in other GitHub CI jobs that increases the available space on the runner from ~24GB to ~61GB by deleting lots of unneeded things. I think that should be plenty of space for this. https://github.com/openhwgroup/cvw/blob/main/.github/scripts/cli-space-cleanup.sh

That is helpful! Not 100% sure, but I think ctest (?) stores logs for each test, and that might be why I am running out of space. The vector test logs are insanely long, it might make sense to disable logging if that is really what is causing it.

nadime15:
I have updated a few things as @jordancarlin  suggested. Each test set now gets its own tarball, and the new workflow pulls them all in parallel and runs them:

[See](https://github.com/nadime15/riscv-test-generator/releases/tag/2025-06-10): 
Bildschirmfoto vom 2025-06-10 23-06-40

and one example [run](https://github.com/nadime15/sail-riscv/actions/runs/15572728949): 
Bildschirmfoto vom 2025-06-10 23-07-31

As discussed on monday I also added the option to run the tests locally. By default, it fetches the riscv-tests and riscv-vector-tests. It checks vlen and elen dynamically and downloads the matching test set if the values are valid.

One thought: might be worth adding a CMake option to enable/disable running vector tests locally (by default should be off). The default config (VLEN=512, ELEN=64) takes over 3h, while something like VLEN=128, ELEN=32 finishes in ~45min. For GitHub workflows on push, we could tweak vlen/elen on the fly with jq if needed.

test/CMakeLists.txt
```diff
@@ -1,4 +1,108 @@
...
+# This option is required for the GitHub CI to prevent parallel
+# jobs from building both riscv-tests and riscv-vector-tests.
+# For example, if the `run-riscv-tests` job runs and DOWNLOAD_LOCALLY
+# is not set to FALSE, it would automatically download and run the
+# vector tests as well, because they are missing locally.
+option(DOWNLOAD_LOCALLY "Download test files locally if needed" ON)
+
+# Function to download and extract test files
+function(download_riscv_tests DOWNLOAD_PATH ARCHIVE_NAME DOWNLOAD_URL)
+    if(NOT EXISTS "${DOWNLOAD_PATH}/${ARCHIVE_NAME}" AND DOWNLOAD_LOCALLY)
```
jordancarlin: Seems like this won't update the downloaded tests if there is a newer version available. You'd need to delete the directory manually to get the new ones and there is no indicator that they are outdated. I think @Timmmm has dealt with this?

test/CMakeLists.txt
```diff
@@ -1,4 +1,108 @@
...
+            message(FATAL_ERROR "Failed to extract ${ARCHIVE_NAME}")
+        endif()
+
+        # Clean up the tar.gz file
+        file(REMOVE "${DOWNLOAD_PATH}/${ARCHIVE_NAME}.tar.gz")
+        message(STATUS "${ARCHIVE_NAME} downloaded and extracted successfully")
+    endif()
+endfunction()
+
+find_program(JQ_EXECUTABLE jq REQUIRED)
```
jordancarlin: I don't think we should be requiring `jq` to compile the model (which this will do in its current form).

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          sudo apt update
+          sudo apt install -y --no-install-recommends pkg-config libgmp-dev curl ninja-build jq
+
+      - name: Install CMake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: "${{ env.CMAKE_VERSION }}"
+
+      - name: Download latest release
```
jordancarlin: What is the benefit of the action? Why not just do
```
wget "https://github.com/nadime15/riscv-test-generator/releases/latest/download/riscv-vector-tests-v${{ matrix.VLEN }}x${{ matrix.XLEN }}.tar.gz"
```

jordancarlin: Or just let CMake handle the download (see below for suggestions on that).

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,106 @@
...
+          jq --argjson vlen "$VLEN" --argjson elen "$ELEN" \
+          '.extensions.V.vlen_exp = $vlen | .extensions.V.elen_exp = $elen' \
+          config/default.json > tmp.json && mv tmp.json config/default.json
+
+      - name: Install Sail (Linux)
+        run: |
+          sudo mkdir -p /usr/local
+          curl --location https://github.com/rems-project/sail/releases/download/$SAIL_VERSION-linux-binary/sail.tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
+
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DDOWNLOAD_LOCALLY=FALSE
```
jordancarlin: What about adding `RUN_VECTOR_TESTS` and `RUN_RISCV_TESTS` options that control which tests are downloaded and run. `RUN_RISCV_TESTS` can default to on so that people have basic testing locally, but then we can disable it for the vector tests in CI so we don't need to manually download the tests above.

test/CMakeLists.txt
nadime15: Yes I agree, after reading a bit, I could remove `jq` altogether and extract `vlen_exp` and `elen_exp` with pure CMake regex. 

I could replace it with:

```
string(REGEX MATCH "\"vlen_exp\"[ \t]*:[ \t]*([0-9]+)" _ "${json}")
set(vlen_exp "${CMAKE_MATCH_1}")
```

nadime15: I removed the jq dependency.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,74 @@
+name: Run risc-v tests
```
Timmmm: Can we avoid having yet another workflow file?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Install CMake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: "${{ env.CMAKE_VERSION }}"
+
+      - name: Set VLEN and ELEN
+        run: |
+          echo "VLEN=$(echo ${{ matrix.VLEN }} | awk '{print int(log($1)/log(2))}')" >> $GITHUB_ENV
+          echo "ELEN=$(echo ${{ matrix.XLEN }} | awk '{print int(log($1)/log(2))}')" >> $GITHUB_ENV
```
Timmmm: This does not seem right? Also I think `jq` supports `log2` natively so you can probably skip this step. Something like

```
jq --argjson vlen "$VLEN" --argjson elen "$ELEN" \
          '.extensions.V.vlen_exp = log2($vlen) | .extensions.V.elen_exp = log2($elen)' \
          config/default.json > tmp.json && mv tmp.json config/default.json
```

This feels like quite a hacky solution.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Modify default Sail config file
+        run: |
+          jq --argjson vlen "$VLEN" --argjson elen "$ELEN" \
+          '.extensions.V.vlen_exp = $vlen | .extensions.V.elen_exp = $elen' \
+          config/default.json > tmp.json && mv tmp.json config/default.json
+
+      - name: Install Sail (Linux)
```
Timmmm: Name is wrong.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: |
+          sudo mkdir -p /usr/local
+          curl --location https://github.com/rems-project/sail/releases/download/$SAIL_VERSION-linux-binary/sail.tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
+
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DRUN_RISCV_TESTS=FALSE -DRUN_RISCV_VECTOR_TESTS=TRUE
+          ninja -C build all riscv_sim_rv32f riscv_sim_rv64f generated_smt_rv64f generated_docs_rv64d
```
Timmmm: Only need `riscv_sim_rv32f riscv_sim_rv64f` here?

test/CMakeLists.txt
```diff
@@ -1,4 +1,103 @@
-add_subdirectory("riscv-tests")
+option(RUN_RISCV_VECTOR_TESTS "Run the riscv-vector-tests" OFF)
+option(RUN_RISCV_TESTS "Run the riscv-tests" ON)
```
Timmmm: These options don't technically *run* the tests. Maybe just use `ENABLE_RISCV_[VECTOR_]TESTS`?

test/CMakeLists.txt
```diff
@@ -1,4 +1,103 @@
...
+        # Clean up the tar.gz file
+        file(REMOVE "${DOWNLOAD_PATH}/${ARCHIVE_NAME}.tar.gz")
+        message(STATUS "${ARCHIVE_NAME} downloaded and extracted successfully")
+    endif()
+endfunction()
+
+# Extracting vlen_exp and elen_exp
+file(READ "${CMAKE_CURRENT_SOURCE_DIR}/../config/default.json" json)
+
+string(REGEX MATCH "\"vlen_exp\"[ \t]*:[ \t]*([0-9]+)" _ "${json}")
```
Timmmm: Veeeery hacky. :-/

Timmmm (changes requested):
I feel like there must be a better way to handle testing multiple vlen/elen. This way seems very awkward.

I think the best option would be to have each VLEN/ELEN as an option. Something like this:

```
foreach (vlen IN ITEMS 128 256 512)
  foreach (elen IN ITEMS 32 64)
    option(ENABLE_RISCV_VECTOR_TESTS_V${vlen}_E{$elen} "Enable the riscv-vector-tests with vlen=${vlen}, elen=${elen}")
    if (ENABLE_RISCV_VECTOR_TESTS_V${vlen}_E{$elen})
       download_riscv_tests(...)
       file(GLOB elfs_rv32d ... )
       file(GLOB elfs_rv32d ... )

        foreach (arch IN ITEMS "rv32d" "rv64d")
            add_custom_command(... call `jq` to generate config JSON ...)
   
           foreach(elf IN LISTS elfs_${arch})
             file(RELATIVE_PATH elf_name "${CMAKE_CURRENT_SOURCE_DIR}" ${elf})

               add_test(
                  NAME "${arch}_${elf_name}"
                    COMMAND
                        $<TARGET_FILE:riscv_sim_${arch}>
                       --config "${CMAKE_CURRENT_BINARY_DIR}/config_${arch}_${vlen}_${elen}.json"
                ${elf}
```

Then I think you don't even need to change CI at all; just enable those CMake options.

.github/workflows/run-tests.yml
nadime15: I still think it makes sense to have one. That doesn’t mean it has to run on every push (which would be way too much anyway, VLEN=512, ELEN=64 takes ~3h), but something like a weekly or monthly run could make sense. Otherwise, you just have to fully trust that the developer actually tested it, which might not be the case. Especially now when a lot of changes are about to happen regarding the vector extension. 

But as always I go with the majority and what they think

test/CMakeLists.txt
nadime15: I tried to avoid anything external and tried to keep it within CMakes capabilities

.github/workflows/run-tests.yml
nadime15: Hmm I am pretty sure I had a look into that and it did not work, I will have another look!

nadime15:
About your change, I like it. The only thing I am unsure about is using `jq`, since not everyone has it installed. And even if they do, they might have a version that does not support certain features (mine cant do log2, jq-1.6).

Timmmm:
Yeah and now that I'm thinking about it our JSON format supports comments, which `jq` presumably doesn't. We don't have any comments yet but it would be nice to have them.

Maybe we should instead use `configure_file()` with a template like @arichardson was suggesting for #870

.github/workflows/run-tests.yml
nadime15: @jordancarlin @pmundkur

pmundkur: I know too little about github CI idioms to have an opinion on this.

nadime15: Ok maybe I misunderstood @Timmmm comment, do you want to keep this job (workflow) and just move it into another workflow file, or do you actually wanna get rid of it completely? 

@pmundkur My question and what I took from this comment was more about whether it even makes sense to have a workflow that runs the vector tests and not about CI idioms.

pmundkur: I'd expect `run-riscv-tests` to run on every push, and `run-riscv-vector-tests` to run on request or every couple of days.  From what little I know, that would mean `run-riscv-tests` should move into the `compile.yml` workflow, but `run-riscv-vector-tests` should have its own file (this one, but perhaps suitably renamed to reflect the extra on-demand tests?).

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,74 @@
...
+          cmake-version: "${{ env.CMAKE_VERSION }}"
+
+      - name: Install Sail (Linux)
+        run: |
+          sudo mkdir -p /usr/local
+          curl --location https://github.com/rems-project/sail/releases/download/$SAIL_VERSION-linux-binary/sail.tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
+
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo
+          ninja -C build all riscv_sim_rv32f riscv_sim_rv64f generated_smt_rv64f generated_docs_rv64d
+          ctest --test-dir build --output-junit tests.xml --quiet
```
pmundkur: Isn't `run-riscv-tests` doing the same thing as `compile.yml::Build and test simulators`?  If so,  do we need it here as well?

nadime15: Yes I agree. Me personally, I think it makes sense to have a separate workflow (file) thats only triggered by us, either manually or through a weekly/monthly cron job.

.github/workflows/run-tests.yml
Timmmm: I think we can have one workflow file and use workflow parameters to control exactly which vector tests are run.

nadime15: I am a bit confused, do you want to have one dedicated workflow file only for the vector tests or should I add this job to the already existing one and do you still wanna run multiple vector test with different ELENs and VLENs at the same time (via matrix strategy)?

I think it makes more sense to have a dedicated workflow just for the vector tests, which can be triggered manually and accepts external parameters to fine-tune what tests you wanna run (i.e., choice(s) of VLEN and ELEN).

nadime15: A quick check with ChatGPT would give something like:

1. Main Workflow:
```
on:
  workflow_dispatch:
    inputs:
      vlen:
        description: "Comma-separated list of VLENs"
        required: false
        default: "128,256,512"
      xlen:
        description: "Comma-separated list of XLENs (== ELEN)"
        required: false
        default: "32,64"
```

which would trigger (manually):

```
jobs:
  run-riscv-vector-tests:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        VLEN: ${{ fromJSON(format('["{0}"]', join('","', split(inputs.vlen, ',')))) }}
        XLEN: ${{ fromJSON(format('["{0}"]', join('","', split(inputs.xlen, ',')))) }}
    steps:
      - name: Echo config
        run: echo "Running with VLEN=${{ matrix.VLEN }}, XLEN=${{ matrix.XLEN }}"
```

Timmmm: Yeah I think one workflow where we can tweak the `inputs`. I think to start with I would just have a bool for each CMake option (`ENABLE_RISCV_TESTS`, `ENABLE_RISCV_VECTOR_TESTS_VLEN128_ELEN64` etc.)

Then we don't need a matrix for vlen/elen.

How long does it take to run all the vector tests for one config? If it's like 10 mins we can probably just run them all for every PR. If not we can select a subset of them to run via CTest's labels feature (you can mark some tests as being slow in the CMakeLists.txt and then exclude them when you run CTest.

Timmmm: I can sketch out the code if you like...

nadime15: I have a successful run here:

https://github.com/nadime15/sail-riscv/actions/runs/15838755386

### Run Time

| Job                               | Run time     |
|----------------------------------|--------------|
| run-riscv-vector-tests (128, 32) | 40m 10s      |
| run-riscv-vector-tests (128, 64) | 45m 36s      |
| run-riscv-vector-tests (256, 32) | 1h 13m 19s   |
| run-riscv-vector-tests (256, 64) | 1h 22m 31s   |
| run-riscv-vector-tests (512, 32) | 2h 43m 56s   |
| run-riscv-vector-tests (512, 64) | 3h 16m 15s   |
| run-riscv-tests                  | 16m 44s      |
| **Total**                        | **10h 18m 31s (cumulative)**|


it stills fails from time to time because according to github there is a [resource starvation issue](https://github.com/nadime15/sail-riscv/actions/runs/15850420179).

> Yeah I think one workflow where we can tweak the inputs. I think to start with I would just have a bool for each CMake option (ENABLE_RISCV_TESTS, ENABLE_RISCV_VECTOR_TESTS_VLEN128_ELEN64 etc.)

So if I got this right, you want to add different VLEN and ELEN variants to the same tests? That’s gonna make the test runs super long.

If you start stacking test runs (like enabling ENABLE_RISCV_VECTOR_TESTS_VLEN128_ELEN64 and ENABLE_RISCV_VECTOR_TESTS_VLEN256_ELEN64), you will probably hit the same resource limits I ran into, plus the tests will drag on even more.

I think for CI it makes sense to run them in Parallel? 

> I can sketch out the code if you like...

Yeah this might be better so I understand what is requested

Timmmm: > I think for CI it makes sense to run them in Parallel?

Yes, CTest *does* run them in parallel. See [`-j`](https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-j).

To avoid that resource starvation issue we might be able to use [`--test-load`](https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-test-load).

IIRC someone offered us a more powerful machine for CI (@jjscheel was it you?). I think that is probably the best way to make it faster.

Alternatively we can keep the matrix and just control the `-DENABLE_..` flags.

> Yeah this might be better so I understand what is requested

Ok I'll make a separate PR so I don't step on your toes.

jjscheel: >IIRC someone offered us a more powerful machine for CI (@jjscheel was it you?). I think that is probably the best way to make it faster.
@Timmmm, it sounds like something I might have said.  :-O

Seriously, I don't have anything in my back pocket, but if that's a problem we need to solve, please reach out directly and we can actively work on it.  Please start with @pmundkur and @nadime15, refine the request and then they can help me solve it.

Does this sound like an acceptable plan (for a plan)?

Timmmm: Hmm maybe it was someone else... someone definitely said they had a beefy machine available. I'll ask at the next meeting I guess.

pmundkur: It might have been @edolnx ?

nadime15:
@Timmmm I gave this another try and incorporated all your suggestions. I think the code is cleaner now and easier to understand.

I explicitly disabled tracing for Sail (I assumed the `--quiet` option for ctest would stop ctest from logging the simulators output, but it does not). Disabling it had a big impact and saved a lot of time and probably resources too, since the tests are now running without any errors. 

I have only tried one run so far, but I’ll rerun it to be sure. [Here](https://github.com/nadime15/sail-riscv/actions/runs/16079813034) is the workflow.

Updated RISC-V Vector Test Runtimes (New without logging vs Previous)

| Job                              | New Time   | Prev Time   |
|----------------------------------|------------|-------------|
| run-riscv-vector-tests (128, 32) | 23m 27s    | 40m 10s     |
| run-riscv-vector-tests (128, 64) | 28m 33s    | 45m 36s     |
| run-riscv-vector-tests (256, 32) | 41m 28s    | 1h 13m 19s  |
| run-riscv-vector-tests (256, 64) | 50m 08s    | 1h 22m 31s  |
| run-riscv-vector-tests (512, 32) | 1h 31m 38s | 2h 43m 56s  |
| run-riscv-vector-tests (512, 64) | 1h 58m 02s | 3h 16m 15s  |
| run-riscv-tests                  | 10m 13s    | 16m 44s     |
| **Total**                        | **6h 3m 29s** | **10h 18m 31s** |

test/CMakeLists.txt
```diff
@@ -1,4 +1,131 @@
...
+
+if(ENABLE_RISCV_TESTS)
+    download_riscv_tests(
+        "${CMAKE_CURRENT_SOURCE_DIR}"
+        "riscv-tests"
+        "https://github.com/nadime15/riscv-sail-tests/releases/download/2025-06-22/riscv-tests.tar.gz"
+    )
+endif()
+
+file(GLOB elfs_rv32d CONFIGURE_DEPENDS LIST_DIRECTORIES false
```
Timmmm: Shouldn't this code all be in the `if(ENABLE_RISCV_TESTS)`?

Timmmm (commented):
Thanks, this looks *much* better! I think the build matrix is ok until we have a faster machine available.

The only things I would change are:

1. Am I misreading it or does it now not run *any* tests by default in CI? I think we want to run `riscv-tests` and a subset of the vector tests.
2. To run that subset we need to add CTest labels to the tests that we want to exclude from CI. Like this:

```
set_tests_properties(
    <tests>...
    PROPERTIES LABELS "too_slow_for_ci"
)
```

And then when testing:

```
ctest --label-exclude too_slow_for_ci ...
```

3. We should download and unzip the tests in the build directory, not the source directory.

And then a couple of things I'm not keen on but I guess we can deal with later:

4. Rather than adding a dependency on `jq`, we should probably do what @arichardson was suggesting and template the JSON using `configure_file`.

5. I suspect that downloading the tests in the configure stage will not be especially robust... but I can't think of a better way to do it so :shrug: .

nadime15:
@Timmmm I added the following label:

>  set_tests_properties(
>      <tests>...
>      PROPERTIES LABELS "too_slow_for_ci"
>  )

For all vector tests where either `VLEN` > 128 or `ELEN` > 64 ( only `VLEN` = 128 and `ELEN` = 32 are exempt).

I know we just talked about the use case for this label today (sorry!), but is it really needed here? I added `-DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE` to `compile.yml`, so `ctest` will only run the `VLEN`=128 and `ELEN`=32 tests on top of the riscv-tests.

Regarding the new `run-riscv-tests` workflow (which, so far, can only be triggered manually), I don’t think it makes sense to include the label there either unless we start triggering this workflow on each `push` or `pull_request`?

Timmmm:
Yeah good point, I think if we just run `(128, 32)` and it only takes 23 minutes then that's probably fine and we can skip the labelling.

I wonder if it will run faster with #971...

.github/workflows/compile.yml
```diff
@@ -64,7 +64,7 @@ jobs:
         run: |
           # Ninja is used because the CMake Makefile generator doesn't
           # build top-level targets in parallel unfortunately.
-          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DFIRST_PARTY_TESTS=TRUE
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DFIRST_PARTY_TESTS=TRUE -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE
```
Timmmm: And `-DENABLE_RISCV_TESTS=TRUE`?

Timmmm: Oh wait, never mind - it's on by default.

jordancarlin:
Based on your table above, it seems like VLEN=128, ELEN=64 takes about the same amount of time (28 minutes) as ELEN=32. If we run both of those in parallel it should have minimal impact on CI time and might be worthwhile.

.github/workflows/compile.yml
jordancarlin: I wonder if it makes sense to split the build and test aspects of our PR CI into separate jobs (within the same file). The build can happen for various OS/CMake versions, but there is no reason to run all of the tests on each of those. We could run a small subset of tests just to ensure the simulator is working as expected (maybe just the first party ones).

Then a separate job could run the full riscv-tests and selected vector tests on just a single OS/CMake combo (probably ubuntu with the latest CMake). This separate test job should also be split so that the riscv-tests and vector tests are run on separate runners to speed it up since the GitHub runners are restricted to just a few cores and won't be able to run many tests in parallel.

.github/workflows/run-riscv-tests.yml
```diff
@@ -0,0 +1,64 @@
+name: Run risc-v tests
+on:
+  workflow_dispatch:
```
jordancarlin, line 3: Should we have the full vector test suite run weekly in addition to on demand?

build_simulators.sh
```diff
@@ -2,6 +2,6 @@
 
 set -e
 : "${DOWNLOAD_GMP:=TRUE}"
-cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DDOWNLOAD_GMP="${DOWNLOAD_GMP}"
+cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DDOWNLOAD_GMP="${DOWNLOAD_GMP}" -DENABLE_RISCV_VECTOR_TESTS_V256_E32=TRUE
```
jordancarlin: I don't think we should enable the vector tests in the `build_simulators.sh` script. This will likely be used primarily by end users who won't even run the test suite, so it just adds an unnecessary and large download to the build process for them.

test/CMakeLists.txt
```diff
@@ -1,4 +1,138 @@
-add_subdirectory("riscv-tests")
+# Function to download and extract test files
+function(download_riscv_tests DOWNLOAD_PATH ARCHIVE_NAME DOWNLOAD_URL)
+    if(NOT EXISTS "${DOWNLOAD_PATH}/${ARCHIVE_NAME}")
```
jordancarlin: Isn't this back to the problem of not updating the downloaded tests if we bump the release of the test repo we are using?

test/CMakeLists.txt
```diff
@@ -1,4 +1,138 @@
...
+                        )
+                    endif()
+                endforeach()
+            endforeach()
+        endif()
+    endforeach()
+endforeach()
+
+# Download and add RISC-V tests.
+option(ENABLE_RISCV_TESTS "Enable riscv-tests" ON)
```
jordancarlin: I'm somewhat hesitant to make the default build require downloading a test suite. If someone tries to build without internet it will now fail by default.

build_simulators.sh
nadime15: Ups my bad, I accidentally pushed that!

test/CMakeLists.txt
nadime15: Yes, but I think we had agreed to just keep it static for now, which is fine I think, probably better than CI suddenly failing when new tests get added.

.github/workflows/run-riscv-tests.yml
nadime15: Yes, I think it would make sense
@Timmmm @pmundkur

test/CMakeLists.txt
nadime15: I might miss something but if you dont have an internet connection you wouldn be able to build Sail anyway because you need to download GMP? But I can also change the exit condition and instead of 

```
    if(NOT extract_result EQUAL 0)
            message(FATAL_ERROR "Failed to extract ${ARCHIVE_NAME}")
        endif()
```

I can just print a warning?

.github/workflows/compile.yml
nadime15: > Then a separate job could run the full riscv-tests and selected vector tests on just a single OS/CMake combo (probably ubuntu with the latest CMake). This separate test job should also be split so that the riscv-tests and vector tests are run on separate runners to speed it up since the GitHub runners are restricted to just a few cores and won't be able to run many tests in parallel.

Sounds like this could just be done by calling the new `run-riscv-tests.yml` from `compile.yml`.

test/CMakeLists.txt
```diff
@@ -1,4 +1,130 @@
...
+    foreach(elen IN ITEMS 32 64)
+        option(ENABLE_RISCV_VECTOR_TESTS_V${vlen}_E${elen}
+            "Enable riscv-vector-tests with vlen=${vlen}, elen=${elen}"
+            OFF)
+
+        if(ENABLE_RISCV_VECTOR_TESTS_V${vlen}_E${elen})
+            download_riscv_tests(
+                "${CMAKE_CURRENT_BINARY_DIR}"
+                "riscv-vector-tests-v${vlen}x${elen}"
+                "https://github.com/nadime15/riscv-sail-tests/releases/download/2025-06-22/riscv-vector-tests-v${vlen}x${elen}.tar.gz"
```
pmundkur: Could the urls in this CMakefile be factored out into `.url` files the way it was done for `os-boot/linux/*.url`?  It would make it easier to run tests against forked test repos.

.github/workflows/compile.yml
jordancarlin: I don't think so. `run-riscv-tests.yml` runs all of the tests. `compile.yml` would still only run a subset (riscv-tests and one or two of the vector suites).

.github/workflows/run-riscv-tests.yml
pmundkur: Agreed, we should run it weekly as well.

test/CMakeLists.txt
jordancarlin: Most systems will already have GMP installed.

pmundkur: The GMP download defaults to OFF.  I'd rather the tests download default to ON, but can go either way.  The `README` and `tests/README` discuss disabling tests so a default of ON sounds reasonable to me.

nadime15: I constantly rebuild the model and delete the build folder, and GMP gets downloaded every time (only when i delete the build folder but somehow CMake struggles with the existing build folder when I exit my code editor, I dont know why). Looks like the script explicitly defaults to downloading GMP unless you override it, instead of using a local system install.

`: "${DOWNLOAD_GMP:=TRUE}"`

```
option(DOWNLOAD_GMP "Download libgmp and build the library locally instead of using a system installation" OFF)
if (DOWNLOAD_GMP)
    include(ExternalProject)
    ExternalProject_Add(
    ...
   )
```

test/CMakeLists.txt
jordancarlin: To clarify, I agree that we should keep it static so CI doesn't start failing when new tests are added. The issue is right now changing the statically set version will not update the tests that are locally downloaded if an older version has already been downloaded.

test/CMakeLists.txt
jordancarlin: The script defaults to on to make it easy for someone to just clone the repo and run the build script without needing dependencies. The CMake flow itself defaults to it being off though.

nadime15: I can't find the discussion right now, but @Timmmm and I discussed this and agreed to set `ENABLE_RISCV_TESTS` to true.

An alternative could be to use the same approach we do right now with:

`: "${DOWNLOAD_GMP:=TRUE}"`

i just set `ENABLE_RISCV_TESTS` to OFF (in CMake file itself) and we enable it in the script via:

`: "${ENABLE_RISCV_TESTS:=TRUE}"`

> The script defaults to on to make it easy for someone to just clone the repo and run the build script without needing dependencies.

Yeah, but I think thats how most people use Sail, they just run the script and build it. So even if its set to OFF in CMake, it would still fail for most users.

test/CMakeLists.txt
nadime15: Im not sure its that straightforward in this case, since the corresponding links include the vlen and elen values, so I would need to fetch the URL and do some string manipulation to replace those placeholders.

Or do you mean I provide links for each tarball (in total 7 links, 6 links to vector tests and 1 to riscv-test)?

pmundkur: Yeah, the latter: have all the available links from the upstream release tarballs under `tests/*.url`, and then select the appropriate `.url` file based on the setting of `vlen` and `elen` in `CMakeLists.txt`.

test/CMakeLists.txt
Timmmm: Yeah I think the easiest way to fix that is to include the version in the downloaded path.

test/CMakeLists.txt
Timmmm: I don't think that helps in this case because we need to GLOB the files in the configure step, so we need to download them in the configure step too. Slight limitation of CMake - it doesn't work too well with dynamic dependencies (probably because it targets build systems that also don't). It's pretty rare in my experience for build systems to support build graphs that aren't fully known in advance (though [some do](https://buck2.build/docs/rule_authors/dynamic_dependencies/)).

test/CMakeLists.txt
Timmmm: Hmm yeah I forgot about it requiring the download though. Tbh I don't have a strong opinion either way. Sorry!

nadime15: No problem! I changed it back and followed the same approach as we have for `DOWNLOAD_GMP`

test/CMakeLists.txt
nadime15: Done, let me now if this is fine.

test/CMakeLists.txt
nadime15: What I did not is, I extract the tag/version from the URL and add it to the downloaded path /build/test/VERSION/riscv-tests etc. 

I hope this works out and is not to hacky, let me know what you think

nadime15:
There is something wrong with [CI](https://github.com/nadime15/sail-riscv/actions/runs/16153055646/job/45588851351). 

Some of the tests from https://github.com/riscv-software-src/riscv-tests are failing. The naming is weird, no clue what they are even supposed to test.

```
1983/2098 Test #1983: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh1add_uw ....................***Failed    0.01 sec
Running file /home/runner/work/sail-riscv/sail-riscv/build/test/2025-06-22/riscv-tests/rv64uzba-p-sh1add_uw.
ELF Entry @ 0x80000000
HTIF located at 0x80001000
begin_signature: 0x80002000
end_signature: 0x80002000
FAILURE: 669
          Start 1984: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh2add
1984/2098 Test #1984: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh2add .......................   Passed    0.01 sec
          Start 1985: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh2add_uw
1985/2098 Test #1985: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh2add_uw ....................***Failed    0.01 sec
Running file /home/runner/work/sail-riscv/sail-riscv/build/test/2025-06-22/riscv-tests/rv64uzba-p-sh2add_uw.
ELF Entry @ 0x80000000
HTIF located at 0x80001000
begin_signature: 0x80002000
end_signature: 0x80002000
FAILURE: 669
          Start 1986: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh3add
1986/2098 Test #1986: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh3add .......................   Passed    0.01 sec
          Start 1987: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh3add_uw
1987/2098 Test #1987: rv64d_2025-06-22/riscv-tests/rv64uzba-p-sh3add_uw ....................***Failed    0.01 sec
Running file /home/runner/work/sail-riscv/sail-riscv/build/test/2025-06-22/riscv-tests/rv64uzba-p-sh3add_uw.
ELF Entry @ 0x80000000
HTIF located at 0x80001000
begin_signature: 0x80002000
end_signature: 0x80002000
FAILURE: 669
          Start 1988: rv64d_2025-06-22/riscv-tests/rv64uzba-p-slli_uw
1988/2098 Test #1988: rv64d_2025-06-22/riscv-tests/rv64uzba-p-slli_uw ......................   Passed    0.01 sec
          Start 1989: rv64d_2025-06-22/riscv-tests/rv64uzba-v-add_uw
1989/2098 Test #1989: rv64d_2025-06-22/riscv-tests/rv64uzba-v-add_uw .......................   Passed    0.17 sec
          Start 1990: rv64d_2025-06-22/riscv-tests/rv64uzba-v-sh1add
1990/2098 Test #1990: rv64d_2025-06-22/riscv-tests/rv64uzba-v-sh1add .......................   Passed    0.16 sec
          Start 1991: rv64d_2025-06-22/riscv-tests/rv64uzba-v-sh1add_uw
```

jordancarlin:
> There is something wrong with [CI](https://github.com/nadime15/sail-riscv/actions/runs/16153055646/job/45588851351).
> 
> Some of the tests from https://github.com/riscv-software-src/riscv-tests are failing. The naming is weird, no clue what they are even supposed to test.

Fixed with #1132

test/CMakeLists.txt
```diff
@@ -1,4 +1,158 @@
...
+        option(ENABLE_RISCV_VECTOR_TESTS_V${vlen}_E${elen}
+            "Enable riscv-vector-tests with vlen=${vlen}, elen=${elen}"
+            OFF)
+
+        if(ENABLE_RISCV_VECTOR_TESTS_V${vlen}_E${elen})
+            set(TARBALL_NAME "riscv-vector-tests-v${vlen}x${elen}")
+            set(URL_FILE "${CMAKE_CURRENT_SOURCE_DIR}/${TARBALL_NAME}.url")
+
+            # Read URL and extract version, then build path
+            file(READ "${URL_FILE}" DOWNLOAD_URL)
```
Timmmm: I think just have the base URL as a CMake option. There's no point having it in a separate file. Just do something like this:

```
set(TEST_DOWNLOAD_URL "https://github.com/nadime15/riscv-sail-tests/releases/download" CACHE STRING "Base URL to download precompiled riscv-tests and riscv-vector-tests from")
set(TEST_DOWNLOAD_VERSION "2025-06-22" CACHE STRING "Version of precompiled tests to download")

...
set(download_path "${CMAKE_CURRENT_BINARY_DIR}/${TEST_DOWNLOAD_VERSION}")
set(download_url "${TEST_DOWNLOAD_URL}/${TEST_DOWNLOAD_VERSION}")
```

Timmmm (commented):
I think this is getting pretty close. We should probably ask for an official repo under the `riscv` github org. @pmundkur can you do that?

test/CMakeLists.txt
nadime15: > I think just have the base URL as a CMake option. There's no point having it in a separate file.

@pmundkur asked for this, just checking before I start making changes again. Would be great to agree on whether we want to keep it or not!

pmundkur: Yes, I'd prefer having it in a `.url` file to make it easy to point to forked repos, as I mentioned earlier.  This is simpler than mucking around inside `CMakeList` files. It's also more consistent with `os-boot/linux/*.url`.  I'd like to keep it the way you have it now, thanks @nadime15 for doing it!

pmundkur:
> I think this is getting pretty close. We should probably ask for an official repo under the `riscv` github org. @pmundkur can you do that?

Sure, I can do that, let's first merge this using @nadime15's repo and give it a few days to work out any issues.

pmundkur (approved):
LGTM.  This is an amazing improvement for our testing situation!

test/CMakeLists.txt
Timmmm: I have to disagree here - this way seems significantly more complex (a whole extra hacky regex function) for no benefit. If forks want to use a different URL they can easily pass `-DTEST_DOWNLOAD_URL=https://something_else` - that doesn't even require modifying the code unlike a `.url` file.

> It's also more consistent with os-boot/linux/*.url

Yeah but there's actually a benefit to doing it like that there. Here it just makes things more complicated.

What do you think @arichardson ?

pmundkur:
Agreed: "-DTEST_DOWNLOAD_URL=https://something_else" seems preferable and nicer.

test/CMakeLists.txt
nadime15: Done

.github/workflows/compile.yml
jordancarlin: I’d still like to see this happen, but it can be part of a separate PR at this point.

.github/workflows/run-riscv-tests.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Build and test simulator
+        run: |
+          CMAKE_OPTS=""
+          if [[ "${{ matrix.test_suite }}" == "riscv-vector" ]]; then
+            CMAKE_OPTS="-DENABLE_RISCV_VECTOR_TESTS_V${{ matrix.VLEN }}_E${{ matrix.XLEN }}=TRUE"
+          else
+            CMAKE_OPTS="-DENABLE_RISCV_TESTS=TRUE"
+          fi
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo $CMAKE_OPTS
```
jordancarlin: This could be simplified significantly by changing the matrix to just have a `TEST_SUITE` entry and having it contain `RISCV_TESTS`, `RISCV_VECTOR_TESTS_V128_E32`, etc.

Then instead of needing the whole conditional, this section becomes 

```
cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_${TEST_SUITE}=TRUE
```

jordancarlin (approved):
One minor suggestion, but LGTM! This will be a great improvement.

.github/workflows/run-riscv-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+      fail-fast: false
+      matrix:
+        include:
+          - TEST_SUITE: RISCV_TESTS
+          - TEST_SUITE: RISCV_VECTOR_TESTS_V128_E32
+          - TEST_SUITE: RISCV_VECTOR_TESTS_V128_E64
+          - TEST_SUITE: RISCV_VECTOR_TESTS_V256_E32
+          - TEST_SUITE: RISCV_VECTOR_TESTS_V256_E64
+          - TEST_SUITE: RISCV_VECTOR_TESTS_V512_E32
+          - TEST_SUITE: RISCV_VECTOR_TESTS_V512_E64
```
jordancarlin:
```suggestion
      matrix:
        TEST_SUITE: [RISCV_TESTS, RISCV_VECTOR_TESTS_V128_E32, RISCV_VECTOR_TESTS_V128_E64, RISCV_VECTOR_TESTS_V256_E32, RISCV_VECTOR_TESTS_V256_E64, RISCV_VECTOR_TESTS_V512_E32, RISCV_VECTOR_TESTS_V512_E64]
```

Timmmm:
Commit message should have been edited before merging this. :-/

PR #1004, opened 2025-06-04
Fix mmu-type in device tree.

pmundkur:
This does a better job of ensuring that the `mmu-type` is actually derived from the model configuration.

### DIFF
--- a/model/riscv_device_tree.sail
+++ b/model/riscv_device_tree.sail
@@ -6,6 +6,21 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
+// Identifies the largest MMU address translation mode supported by this hart.
+function mmu_type() -> string = {
+  if xlen == 32 then {
+    assert(hartSupports(Ext_Sv32), "Cannot generate RV32 device-tree without Sv32 support.");
+    return "sv32"
+  };
+  if hartSupports(Ext_Sv57) then return "sv57"
+  else if hartSupports(Ext_Sv48) then return "sv48"
+  else if hartSupports(Ext_Sv39) then return "sv39"
+  else {
+    assert(false, "No MMU types supported for RV64 device-tree.");
+    "" // to satisfy the type-checker
+  }
+}
+
 function generate_dts() -> string = {
   let clock_freq : int = config platform.clock_frequency;
   let ram_base_hi = unsigned(plat_ram_base >> 32);
@@ -39,7 +54,7 @@ function generate_dts() -> string = {
 ^ "      status = \"okay\";\n"
 ^ "      compatible = \"riscv\";\n"
 ^ "      riscv,isa = \"" ^ "rv" ^ dec_str(xlen) ^ isa_string ^ "\";\n"
-^ "      mmu-type = \"riscv," ^ (if xlen == 32 then "sv32" else "sv39") ^ "\";\n"
+^ "      mmu-type = \"riscv," ^ mmu_type() ^ "\";\n"
 ^ "      clock-frequency = <" ^ dec_str(clock_freq) ^ ">;\n"
 ^ "      CPU0_intc: interrupt-controller {\n"
 ^ "        #address-cells = <2>;\n"


model/riscv_device_tree.sail
```diff
@@ -6,6 +6,21 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
+// Identifies the largest MMU address translation mode supported by this hart.
+function mmu_type() -> string = {
+  if xlen == 32 then {
+    assert(hartSupports(Ext_Sv32), "Cannot generate RV32 device-tree without Sv32 support.");
+    return "sv32"
+  };
+  if hartSupports(Ext_Sv57) then return "sv57"
```
nadime15: I know it makes no sense but the [links](https://github.com/riscv/sail-riscv/pull/972#discussion_r2126544308) I have checked only defined it up to `sv48`. So we allow sv57 as well?

pmundkur: Yes, sv57 is a valid mmu-type.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings/riscv/cpus.yaml#n78

nadime15: Ahh ok great!

pmundkur:
> To follow up on the discussion we had [here](https://github.com/riscv/sail-riscv/pull/972#discussion_r2126757307)
> 
> and @radimkrcmar comment
> 
> > Yeah, the mmu-type is interpreted as the highest supported paging mode and implies all standard lower ones (e.g. "riscv,sv57" is {Sv57, Sv48, and Sv39}).
> > In Sail, we should have a condition chain with hartSupports to pick the correct one.
> > The design is not great and hopefully we'll move to listing of supported paging extensions in the future -- there is no good way to describe both rv32 and rv64 modes (when we have variable xlen) and mmu-type doesn't really work with custom modes either.
> 
> Does this mean this [PR](https://github.com/riscv/sail-riscv/pull/807/files) needs to be redone as well?
> 
> ```
> // Supervisor-Level Address Translation Modes
> enum clause extension = Ext_Svbare
> function clause hartSupports(Ext_Svbare) = config extensions.Svbare.supported
> enum clause extension = Ext_Sv32
> function clause hartSupports(Ext_Sv32) = config extensions.Sv32.supported : bool & (xlen == 32)
> enum clause extension = Ext_Sv39
> function clause hartSupports(Ext_Sv39) = config extensions.Sv39.supported : bool & (xlen == 64)
> enum clause extension = Ext_Sv48
> function clause hartSupports(Ext_Sv48) = config extensions.Sv48.supported : bool & (xlen == 64)
> enum clause extension = Ext_Sv57
> function clause hartSupports(Ext_Sv57) = config extensions.Sv57.supported : bool & (xlen == 64)
> ```

I don't think that needs changing. The hart could support multiple translation modes; the device-tree only reports the "largest MMU translation mode".

nadime15:
What I mean is, what if Sv57 is enabled but you disabled Sv48 in the config file? In this case Sv48 should be supported (implicitly) as well. I think this will "fail" later too in `riscv_sys_regs.sail`:

```
match satpMode_of_bits(arch, s[Mode]) {
      None()  => prev_value,
      Some(Sv_mode) =>  match Sv_mode {
        Bare if currentlyEnabled(Ext_Svbare) => s.bits,
        Sv39 if currentlyEnabled(Ext_Sv39) => s.bits,
        Sv48 if currentlyEnabled(Ext_Sv48) => s.bits,
        Sv57 if currentlyEnabled(Ext_Sv57) => s.bits,
        _ => prev_value,
      }
    }
```

Something like:
```
enum clause extension = Ext_Sv48
function clause hartSupports(Ext_Sv48) = ((config extensions.Sv48.supported : bool) | (config extensions.Sv57.supported : bool)) & (xlen == 64)
enum clause extension = Ext_Sv57
function clause hartSupports(Ext_Sv57) = config extensions.Sv57.supported : bool & (xlen == 64)
```

radimkrcmar:
The ISA requires that all smaller Sv modes are supported as well, so Sv57 without Sv48 would not be a valid RISC-V configuration.  Where we want to check and enforce it is another decision...

pmundkur:
See #1005.

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,21 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
+// Identifies the largest MMU address translation mode supported by this hart.
+function mmu_type() -> string = {
+  if xlen == 32 then {
+    assert(hartSupports(Ext_Sv32), "Cannot generate RV32 device-tree without Sv32 support.");
```
jordancarlin: The device tree supports `riscv,none` as the mmu type. We should output `none` rather than failing when virtual memory isn't supported. Same for RV64 below.

jordancarlin (approved):
LGTM

model/riscv_device_tree.sail
```diff
@@ -6,6 +6,19 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
+// Identifies the largest MMU address translation mode supported by this hart.
+function mmu_type() -> string = {
+  if xlen == 32 then {
+    if hartSupports(Ext_Sv32) then "sv32" else "none"
+  } else {
+    assert(xlen == 64);
```
Timmmm, line 14: Could you use `switch xlen { 32 => ... 64 => ... }` to avoid this?

Timmmm (approved):

model/riscv_device_tree.sail
pmundkur: Trying that gives:
```
Warning: Incomplete pattern match statement at riscv_device_tree.sail:11.2-7:
11 |  match xlen {
   |  ^---^
   | 
The following expression is unmatched: 0
```
Explicitly typing the `xlen` definition in `riscv_xlen.sail` with
```
let xlen : int(xlen) = sizeof(xlen)
```
gives the same warning.

Timmmm: That feels like a bug @Alasdair ?

Timmmm: Maybe lets check it again after #870

PR #1005, opened 2025-06-04
Add a configuration validation function to the model.

pmundkur:
This is a start and can be extended to check more invariants of a valid configuration.


pmundkur:
This is a start at addressing #860.

jordancarlin:
I was looking into something similar recently so that we could support extensions like Zic64b (they would just be a check in the config validation function). Now that we output a device tree, it makes sense to support these kinds of extensions since many of them are valid device tree entries.

What's the benefit of pulling this up into the C model? I think it would be simpler if it stayed within the Sail and was called as part of `init_model()`.

I was thinking about this as a bunch of asserts in one big config validation function:
```
assert (hartSupports(Ext_U) | not(hartSupports(Ext_S)), "Supervisor (S) mode requires User (U) mode");
...
```

I think the if/else structure you set up is more intuitive, but I'm worried that having separate functions for all of this will get unwieldy because there is going to end up being a lot of validation that makes its way into this.

pmundkur:
One aspect is the user-experience.  I initially had asserts as well, and then testing showed only one error message when there could be several errors in the config.  The current approach tries to show all the config errors before erroring out.

model/riscv_validate_config.sail
```diff
@@ -0,0 +1,117 @@
...
+    then {
+      valid = false;
+      print_endline("Sv57 is enabled: Sv57 is not supported on RV32.");
+    };
+  };
+  valid
+}
+
+function check_vlen_elen() -> bool = {
+  // Should this be conditioned on a vector extension being enabled?
```
Timmmm, line 59: I think it's simpler just to require a valid config anyway. We can make the defaults valid so it's no effort.

model/riscv_validate_config.sail
```diff
@@ -0,0 +1,117 @@
...
+  };
+  true
+}
+
+function has_overlap(a_lo : nat, a_hi : nat, b_lo : nat, b_hi : nat) -> bool = {
+  not((a_lo < b_lo & a_hi < b_lo) | (b_lo < a_lo & b_hi < a_lo))
+}
+
+// This will change when we have a more general memory layout.
+function check_mem_layout() -> bool = {
```
Timmmm, line 73: Maybe skip this for now until #866 is done?

c_emulator/riscv_sim.cpp
```diff
@@ -145,6 +146,19 @@ static void report_arch(void)
...
 }
 
+static void validate_config(const char *conf_file)
+{
+  if (!zvalidate_config(UNIT)) {
+    if (conf_file) {
+      fprintf(stderr, "Configuration in %s is invalid.\n", conf_file);
+    } else {
+      fprintf(stderr, "Default configuration is invalid.\n");
+      // Should abort here?
```
Timmmm: Probably don't need to because you're `exit(1)`ing immediately anyway. I guess you could add something like `(this is a bug; please report it to the sail-riscv developers)` to make it clearer.

model/riscv_validate_config.sail
```diff
@@ -0,0 +1,117 @@
...
+  then {
+    valid = false;
+    print_endline("The Clint and RAM regions overlap.");
+  };
+
+  // Should memory regions also be 4K aligned?
+  valid
+}
+
+function validate_config() -> bool = {
```
Timmmm: Maybe rename this to `config_is_valid()`? At some point we'll have `assert(config_is_valid())`.

Timmmm (commented):
Looks good to me - just some minor comments.

model/riscv_validate_config.sail
pmundkur: As the comment there says, we can always change this once we update the memory layout. So keeping it doesn't hurt?

Timmmm:
> One aspect is the user-experience. I initially had asserts as well, and then testing showed only one error message when there could be several errors in the config. The current approach tries to show all the config errors before erroring out.

I think you can still do that if you call it in `init_model()` though? I think the only thing you can't do is repeat the config file name, but I think that's worth the extra simplicity.

pmundkur:
> > One aspect is the user-experience. I initially had asserts as well, and then testing showed only one error message when there could be several errors in the config. The current approach tries to show all the config errors before erroring out.
> 
> I think you can still do that if you call it in `init_model()` though? I think the only thing you can't do is repeat the config file name, but I think that's worth the extra simplicity.

It leads to complexity elsewhere: `init_model()` is in `riscv_step.sail` and this would make `riscv_step.sail` depend on `riscv_validate_config.sail`, which would be odd.  It would be better to pull out `init_model()` from `riscv_step.sail` into a separate file that then depends on `riscv_validate_config.sail`  but then we're getting into more file restructuring than the current commit.   If you'd really prefer that, we could do that restructure in a follow up PR?

c_emulator/riscv_sim.cpp
```diff
@@ -430,9 +432,9 @@ void init_sail_reset_vector(uint64_t entry)
   zPC = rom_base;
 }
 
-void init_sail(uint64_t elf_entry)
+void init_sail(uint64_t elf_entry, const char *config_file)
 {
-  zinit_model(UNIT);
+  zinit_model(config_file ? config_file : "");
```
```suggestion
  zinit_model(config_file != nullptr ? config_file : "");
```

(I generally try to avoid implicit bool conversions - even though C++ lets you do it it's bad style.)

model/riscv_model.sail
```diff
@@ -0,0 +1,25 @@
...
+function reset() -> unit = {
+  reset_sys();
+  reset_vmem();
+
+  // To allow model extensions (code outside this repo) to perform additional reset.
+  ext_reset();
+}
+
+// Initialize model state. This is only called once; not for every chip reset.
+function init_model(src: string) -> unit = {
```
Timmmm:
```suggestion
function init_model(config_filename : string) -> unit = {
```

Slightly clearer?

model/riscv_model.sail
```diff
@@ -0,0 +1,25 @@
...
+  assert(config_is_valid(), (if string_length(src) == 0 then "Default config" else "Config in " ^ src)
```
Timmmm: Could also do this but I dunno which is preferable. `string_length` is so verbose!

```suggestion
  assert(config_is_valid(), (if src == "" then "Default config" else "Config in " ^ src)
```

model/riscv_validate_config.sail
Timmmm: Just means more rebase work but ok.

Timmmm (commented):
This looks a lot better! Just some minor style comments and then LGTM.

c_emulator/riscv_sail.h
```diff
@@ -18,10 +18,11 @@ extern struct zMisa zmisa;
 void model_init(void);
 void model_fini(void);
 
-unit zinit_model(unit);
+unit zinit_model(const_sail_string);
 bool ztry_step(sail_int, bool);
 unit ztick_clock(unit);
 
+bool zconfig_is_valid(unit);
```
jordancarlin, line 25: Not needed anymore.

model/CMakeLists.txt
```diff
@@ -268,6 +274,7 @@ foreach (xlen IN ITEMS 32 64)
                         --c-preserve try_step
                         --c-preserve tick_clock
                         --c-preserve generate_dts
+                        --c-preserve config_is_valid
```
jordancarlin, line 277: Not needed anymore.

model/riscv_validate_config.sail
```diff
@@ -0,0 +1,117 @@
...
+    then {
+      valid = false;
+      print_endline("Sv39 is enabled: Sv39 is not supported on RV32.");
+    };
+    if hartSupports(Ext_Sv48)
+    then {
+      valid = false;
+      print_endline("Sv48 is enabled: Sv48 is not supported on RV32.");
+    };
+    if hartSupports(Ext_Sv57)
```
jordancarlin: Can probably combine these three. Something like this.
```sail
if (hartSupports(Ext_Sv57) | hartSupports(Ext_Sv48) | hartSupports(Ext_Sv39))
     then {
       valid = false;
       print_endline("Sv39/Sv48/Sv57 is enabled: Sv39/Sv48/Sv57 are not supported on RV32.");
     };
```

jordancarlin (changes requested):

c_emulator/riscv_sail.h
pmundkur: I decided to keep these and add back the cli option; these helps validate a config without running an elf.

Timmmm (approved):
LGTM!

model/riscv_validate_config.sail
```diff
@@ -0,0 +1,107 @@
...
+      print_endline("Sv48 is enabled but Sv39 is disabled: supporting Sv48 requires supporting Sv39.");
+    };
+    if hartSupports(Ext_Sv32)
+    then {
+      valid = false;
+      print_endline("Sv32 is enabled: Sv32 is not supported on RV64.");
+    };
+  } else {
+    assert(xlen == 32);
+    if not(hartSupports(Ext_S)) & hartSupports(Ext_Sv32)
```
jordancarlin, line 44: Could probably combine this with the `if not(hartSupports(Ext_S)) & (hartSupports(Ext_Sv57) | hartSupports(Ext_Sv48) | hartSupports(Ext_Sv39))` clause and move them before the rv32/64 conditional, but fine with me either way.

jordancarlin (approved):

model/riscv_validate_config.sail
```diff
@@ -0,0 +1,107 @@
...
+
+function check_vlen_elen() -> bool = {
+  // Should this be conditioned on a vector extension being enabled?
+  if VLEN_pow < ELEN_pow
+  then {
+    print_endline("VLEN (set to 2^" ^ dec_str(VLEN_pow) ^ ") cannot be less than ELEN (set to 2^" ^ dec_str(ELEN_pow) ^ ").");
+    return false;
+  };
+  true
+}
```
nadime15, line 66: Should we check here that `VLEN ≤ 2^16`, and similarly that `ELEN` and `VLEN` are `≥ 2^3`?

> 32.2. Implementation-defined Constant Parameters
> Each hart supporting a vector extension defines two parameters:
> 1. The maximum size in bits of a vector element that any operation can produce or consume, ELEN ≥
> 8, which must be a power of 2.
> 2. The number of bits in a single vector register, VLEN ≥ ELEN, which must be a power of 2, and must
> be no greater than 2^16.

pmundkur: Those are enforced by the range constraints in `riscv_vlen.sail`.

nadime15: Ahh true, my bad!

PR #1010, opened 2025-06-06
Support generating assembly for reserved vtypes

Timmmm:
The `vset[i]vli` instructions take an immediate that sets the `vtype` CSR. The normal way to generate the immediate value is by giving values for all of `vtype`'s fields (`ma`, `ta`, `sew` and `lmul`).

`sew` and `lmul` have reserved values, however that does *not* mean that instructions that use those reserved values are necessarily reserved. Implementations are allowed to legalise the values (see the note in `handle_illegal_vtype()`).

Therefore we should support disassembling these instructions. The syntax for a `vset[i]vli` with a reserved `vtype` is just to use an integer immediate instead of the mnemonics.

### DIFF
--- a/model/riscv_insts_vext_vset.sail
+++ b/model/riscv_insts_vext_vset.sail
@@ -16,7 +16,7 @@ mapping sew_flag : string <-> bits(3) = {
   "e8"    <-> 0b000,
   "e16"   <-> 0b001,
   "e32"   <-> 0b010,
-  "e64"   <-> 0b011
+  "e64"   <-> 0b011,
 }
 
 mapping maybe_lmul_flag : string <-> bits(3) = {
@@ -28,17 +28,25 @@ mapping maybe_lmul_flag : string <-> bits(3) = {
   sep() ^ "m4"    <-> 0b010,
   sep() ^ "m8"    <-> 0b011,
   forwards ""      => 0b000, /* m1 by default */
-
 }
 
 mapping ta_flag : string <-> bits(1) = {
   sep() ^ "ta" <-> 0b1,
-  sep() ^ "tu" <-> 0b0
+  sep() ^ "tu" <-> 0b0,
 }
 
 mapping ma_flag : string <-> bits(1) = {
   sep() ^ "ma" <-> 0b1,
-  sep() ^ "mu" <-> 0b0
+  sep() ^ "mu" <-> 0b0,
+}
+
+mapping vtype_assembly : string <-> (bits(1), bits(1), bits(3), bits(3)) = {
+  // Mnemonics are preferable but there's no mnemonic for lmul=0b100.
+  sew_flag(sew) ^ maybe_lmul_flag(lmul) ^ ta_flag(ta) ^ ma_flag(ma) <-> (ma, ta, sew, lmul)
+    when sew[2] != bitone & lmul != 0b100,
+  // However that is still potentially a legal instruction so you can
+  // use an integer for the whole vtype in that case.
+  hex_bits_8(ma @ ta @ sew @ lmul)  <-> (ma, ta, sew, lmul),
 }
 
 val handle_illegal_vtype : unit -> unit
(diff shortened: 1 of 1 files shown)


jordancarlin:
Interesting one. Do the assemblers accept mnemonics for the valid arguments and integers for the invalid ones instead of making the whole thing a single integer?

Also, @tobiasgrosser @ineol, seems like we hit another Lean bug:
```
✖ [66/96] Building LeanRV64D.RiscvInstsVextVset
trace: .> LEAN_PATH=/home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean /home/runner/.elan/toolchains/leanprover--lean4-pr-releases---pr-release-8577/bin/lean --tstack=400000 -Dweak.linter.style.nameCheck=false /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/LeanRV64D/RiscvInstsVextVset.lean -R /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D -o /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D/RiscvInstsVextVset.olean -i /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D/RiscvInstsVextVset.ilean -c /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D/RiscvInstsVextVset.c --json
error: LeanRV64D/RiscvInstsVextVset.lean:278:11: function expected at
  some throw
term has type
  Option (?m.2679 → ?m.2680 ?m.2682)
error: LeanRV64D/RiscvInstsVextVset.lean:279:4: invalid dotted identifier notation, expected type is not of the form (... → C ...) where C is a constant
  ?m.2838
error: LeanRV64D/RiscvInstsVextVset.lean:289:11: function expected at
  some throw
term has type
  Option (?m.3188 → ?m.3189 ?m.3191)
error: LeanRV64D/RiscvInstsVextVset.lean:290:4: invalid dotted identifier notation, expected type is not of the form (... → C ...) where C is a constant
  ?m.3347
```

Timmmm:
> Interesting one. Do the assemblers accept mnemonics for the valid arguments and integers for the invalid ones instead of making the whole thing a single integer?

[Doesn't look like it.](https://godbolt.org/z/8b6nPGrMY)

jordancarlin:
> > Interesting one. Do the assemblers accept mnemonics for the valid arguments and integers for the invalid ones instead of making the whole thing a single integer?
> 
> [Doesn't look like it.](https://godbolt.org/z/8b6nPGrMY)

Unfortunate. Then this seems fine to me.

ineol:
As far as I can see, the mapping generated by sail for `vtype_assembly` never matches because of the `exit` call:
```
function vtype_assembly_forwards arg# = let head_exp# = arg# in
  $[complete] $[mapping_match] match let g__3 = head_exp# in Some(exit(())) {
    Some(result) => result,
    _ => internal_plet _ =
      assert(false, "Pattern match failure at unknown location")
    in
      exit(())
  }
```
Is it an issue in the sail compiler or in the mapping?

Timmmm:
Hmm that's weird. What happens if you remove this line?

```
  sew_flag(sew) ^ maybe_lmul_flag(lmul) ^ ta_flag(ta) ^ ma_flag(ma) <-> (ma, ta, sew, lmul)
    when sew[2] != bitone & lmul != 0b100,
```

ineol:
Then it does work (and the `exit` disappears)

Timmmm:
Ok that's weird. Feels like a Sail compiler bug to me unless I've missed something... @Alasdair ?

jordancarlin:
Seems like this is worth merging and hopefully @ineol or @tobiasgrosser can look into a fix for the Lean issue separately.

PR #1012, opened 2025-06-06
Handle reserved nreg values for `vmv` in encdec

Timmmm:
The `vmvNr.v` instruction only supports N of 1, 2, 4, or 8 (encoded as 0b00000, 0b00001, 0b00011, and 0b00111 respectively). Other values are reserved.

This changes the code to raise an illegal instruction exception directly in `encdec` instead of in the `execute` clause. Without this change it's possible to decode an opcode to `VMVRTYPE` but then not be able to disassemble it, because the `assembly` clause does not support the reserved values.

This also makes the code a bit simpler since it doesn't need to do bit twiddling tricks to get `N` - it comes directly from the decoder mapping.

### DIFF
--- a/model/riscv_insts_vext_arith.sail
+++ b/model/riscv_insts_vext_arith.sail
@@ -1248,22 +1248,32 @@ mapping clause assembly = MOVETYPEI(vd, simm)
   <-> "vmv.v.i" ^ spc() ^ vreg_name(vd) ^ sep() ^ hex_bits_5(simm)
 
 /* ********************* OPIVI (Whole Vector Register Move) ********************** */
-union clause ast = VMVRTYPE : (vregidx, bits(5), vregidx)
+union clause ast = VMVRTYPE : (vregidx, {1, 2, 4, 8}, vregidx)
 
-mapping clause encdec = VMVRTYPE(vs2, simm, vd)
-  <-> 0b100111 @ 0b1 @ encdec_vreg(vs2) @ simm @ 0b011 @ encdec_vreg(vd) @ 0b1010111
+// "The number of vector registers to copy is encoded in the low three
+// bits of the simm field (simm[2:0]) using the same encoding as the nf[2:0]
+// field for memory instructions, i.e., simm[2:0] = NREG-1.
+//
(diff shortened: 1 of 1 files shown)


model/riscv_insts_vext_arith.sail
```diff
@@ -1248,22 +1248,32 @@ mapping clause assembly = MOVETYPEI(vd, simm)
...
+// than 0, 1, 3, and 7 are reserved."
+mapping encdec_nreg : bits(5) <-> {1, 2, 4, 8} = {
+  0b00000 <-> 1,
+  0b00001 <-> 2,
+  0b00011 <-> 4,
+  0b00111 <-> 8,
+}
+
+mapping clause encdec = VMVRTYPE(vs2, nreg, vd)
+  <-> 0b100111 @ 0b1 @ encdec_vreg(vs2) @ encdec_nreg(nreg) @ 0b011 @ encdec_vreg(vd) @ 0b1010111
```
jordancarlin, line 1267: What about this to simplify the mapping:
```suggestion
  <-> 0b100111 @ 0b1 @ encdec_vreg(vs2) @ 0b00 @ unsigned(nregm1) @ 0b011 @ encdec_vreg(vd) @ 0b1010111
```

Then lower down change `EMUL` to be `let EMUL = nregm1 + 1`.

Might also be able to do something like this:
```suggestion
  <-> 0b100111 @ 0b1 @ encdec_vreg(vs2) @ 0b00 @ (unsigned(nregm1) + 1) @ 0b011 @ encdec_vreg(vd) @ 0b1010111
```

Timmmm: That would allow reserved nreg value to be decided though. Also `unsigned` isn't a `mapping` so you can't use it there.

Timmmm: *decoded

jordancarlin: I didn't look closely enough and missed that it wasn't legal for all values of the bottom 3 bits. This is fine as is.

Issue #879, opened 2025-04-22
Fix handling of fractions in zvk_check_encdec() function for Crypto Vector Extensions


nadime15:
This is a low-priority issue related to the Vector Crypto Extensions. Currently, all Vector Crypto Extensions rely on the following function:

`function zvk_check_encdec(EGW: int, EGS: int) -> bool = (unsigned(vl) % EGS == 0) & (unsigned(vstart) % EGS == 0) & (2 ^ get_lmul_pow() * VLEN) >= EGW`

As noted in this [discussion](https://github.com/riscv/sail-riscv/pull/752#discussion_r2035501261), the internal Sail pow function does not handle fractions, so:

`2 ^ get_lmul_pow() * VLEN`

will not work in certain cases.

Once we merge all Crypto Vector Extensions, the `zvk_check_encdec()` function will be updated accordingly. I have opened the issue to ensure we don’t forget it.

PR #1022, opened 2025-06-10
Fix exponential types with fractional LMUL

Alasdair:
In some places we would compute 2 ^ LMUL_pow, where LMUL_pow can be less than zero intending to support fractional LMUL. However, this will not work as expected because the `2^x` function would always return an integer.

This fixes the issue by dividing by the denominator of the fractional LMUL where needed, so `2 ^ LMUL_pow * VLEN` becomes `VLEN / 2 ^ abs_int(LMUL_pow)` when LMUL_pow is negative.

A new Sail option `--strict-exponentials` enforces the constraint that `2 ^ N` in types always has a positive `N`. This required some further changes, as certain LMUL and SEW values would result in a fractional VLMAX which seems unintended. My reading of the spec is that VLMAX is clamped at 1 in these cases, but it is unclear.

Note that this PR will likely not pass CI until the new Sail release is used, because of the new option. I think it would work with the option removed from the CMakeLists file on the older release though.


nadime15:
Ok cool! That resolves #879.

pmundkur (commented):
I'm unsure about the clamping of VLMAX to 1.  That implies `vl` could be set to 1, but I'm not sure how one element could be processed with these settings.   Better to assert that the exponent is non-negative?

nadime15:
> I'm unsure about the clamping of VLMAX to 1. That implies vl could be set to 1, but I'm not sure how one element could be processed with these settings. Better to assert that the exponent is positive?

I think a vector unit would internally masks the tail:

`tail(x) = (vl <= x < max(VLMAX,VLEN/SEW)) 
`

Edit: I am confused, I think in case of VLMAX == vl == 1, there is simply no tail (?)

pmundkur:
But what if `vstart = x = 0`?

nadime15:
I might miss something but thats not a problem. The element at index `vl` is not part of the computation, so in case of VLMAX = 1 you would operate only on the element at index 0:

`body(x) = (vstart <= x < vl)`

pmundkur:
The question is how that element at index 0 can be processed when `SEW_pow > LMUL_pow + VLEN_pow`.
I don't think this can happen since SEW_pow <= LMUL_pow + ELEN_pow is ensured in  `vsetvl*`, and the config ensures ELEN <= VLEN.  So we could just assert non-negativity of that exponent instead of taking a `max`?

model/riscv_vmem_utils.sail
```diff
@@ -47,7 +47,7 @@ let sys_misaligned_byte_by_byte : bool = config memory.misaligned.byte_by_byte
 let sys_misaligned_allowed_within_exp : range(0, 11) = config memory.misaligned.allowed_within_exp
 
 /* Check if an 'n byte access for an address is within an aligned 2^'r byte region */
-val access_within : forall 'width 'n 'r, 0 <= 'r <= 'width & 1 <= 'n <= 2^'r.
+val access_within : forall 'width 'n ('r : Nat), 'r <= 'width & 1 <= 'n <= 2^'r.
```
Timmmm, line 50: Huh is `Nat` a new kind or has that always existed?

pmundkur:
I rebased this to the latest master, and changed the uses of `max` into asserts.  This still needs #1179 though to pass tests.

model/riscv_zvk_utils.sail
```diff
@@ -14,7 +14,11 @@ function zvk_valid_reg_overlap(rs, rd, emul_pow) = {
   (rs_int + reg_group_size <= rd_int) | (rd_int + reg_group_size <= rs_int)
 }
 
-function zvk_check_encdec(EGW : nat, EGS : nat1) -> bool = (unsigned(vl) % EGS == 0) & (unsigned(vstart) % EGS == 0) & (2 ^ get_lmul_pow() * vlen) >= EGW
+function zvk_check_encdec(EGW: nat, EGS: nat1) -> bool = {
+  let LMUL_pow = get_lmul_pow();
+  let LMUL_times_VLEN : int = if LMUL_pow < 0 then vlen / (2 ^ abs_int(LMUL_pow)) else (2 ^ LMUL_pow) * vlen in
```
Timmmm:
```suggestion
  let LMUL_times_VLEN = if LMUL_pow < 0 then vlen / (2 ^ abs_int(LMUL_pow)) else (2 ^ LMUL_pow) * vlen;
```

model/riscv_vmem_utils.sail
pmundkur: It was introduced in 0.19.

PR #612, opened 2024-11-06
Hypervisor extension

defermelowie:
This is a version of the hypervisor extension ported on a more recent version of the model. Feel welcome to provide any feedback

In order to verify the modifications & additions to this version of the model, the following tests are used:

- [x] The compiled tests in [./test/riscv-tests/](https://github.com/riscv/sail-riscv/tree/master/test/riscv-tests) in order to check for regressions
- [x] By booting Linux on top of OpenSBI (w/o virtualization) in order to check for regressions
- [x] A collection of test written by me for the H-extension: https://github.com/defermelowie/riscv-hext-asm-tests
- [x] A [fork](https://github.com/defermelowie/riscv-hyp-tests/tree/sail) of a testsuite written by  José Martins for the H-extension: https://github.com/josecm/riscv-hyp-tests
- [ ] A [fork](https://github.com/defermelowie/riscv-hs-tests) of https://github.com/dramforever/riscv-hs-tests
- [ ] ~By [booting Linux on top of Xvisor on top of OpenSBI](https://github.com/defermelowie/riscv-hext-xvisor-linux)~ (My current setup requires terminal input which is not already implemented nor part of the hypervisor spec.)

PS I closed #334 in favor of this pull request.

PR #1023, opened 2025-06-10
First step to support hypervisor extension

trdthg:
First step to support hypervisor extension

- Add config option extensions.H
- Add enum Ext_H
- Add hartSupports(Ext_H)
- Add currentlyEnabled(Ext_H), currently disabled
- Add to device_tree ext list
- Update legalize_misa to handle misa[H]

---

Once you begin, things get much easier. (Spilit and rebase from #612)

currentlyEnabled came from https://riscv.github.io/riscv-isa-manual/snapshot/privileged/#hypervisor chapter 21.(0)


model/riscv_extensions.sail
```diff
@@ -65,6 +65,10 @@ function clause hartSupports(Ext_S) = config extensions.S.supported
 enum clause extension = Ext_U
 mapping clause extensionName = Ext_U <-> "u"
 function clause hartSupports(Ext_U) = config extensions.U.supported
+// Hypervisor
+enum clause extension = Ext_H
+mapping clause extensionName = Ext_H <-> "h"
```
radimkrcmar, line 70: We can also add them to the DT.

arichardson (approved):

model/riscv_extensions.sail
trdthg: added

trdthg:
Not sure why lean ci failed. I couldn't find things related to `Ext_xxx` in _*.lean_

jrtc27:
I guess Lean isn't smart enough for the currentlySupported(Ext_H) -> virtual_memory_supported() -> currentlySupported(Ext_SvXX) mutual recursion to automatically note that it clearly terminates, and needs some kind of termination hint to see why that's the case. Perhaps because Ext_SvXX has a higher value than Ext_H so it sees the value as increasing.

trdthg:
> Perhaps because Ext_SvXX has a higher value than Ext_H

Let me test it locally.

trdthg:
The reason should be recursive, generated lean code is as follows

```lean
def currentlyEnabled (merge_var : extension) : SailM Bool := do
  | Ext_H =>
    (pure (false && ((hartSupports Ext_H) && (((_get_Misa_H (← readReg misa)) == (0b1 : (BitVec 1))) && (← (virtual_memory_supported
                ()))))))

def virtual_memory_supported (_ : Unit) : SailM Bool := do
  (pure ((← (currentlyEnabled Ext_Sv32)) || ((← (currentlyEnabled Ext_Sv39)) || ((← (currentlyEnabled
              Ext_Sv48)) || (← (currentlyEnabled Ext_Sv57))))))
end
```

Need a lean expert to help :D

model/riscv_extensions.sail
```diff
@@ -65,6 +65,10 @@ function clause hartSupports(Ext_S) = config extensions.S.supported
...
+function clause hartSupports(Ext_H) = config extensions.H.supported
```
jordancarlin: I think until the extension is fully supported, we shouldn't have a flag in the config file and instead should just hardcode the `hartSupports` clause to false. Otherwise, someone will inevitably try to enable it in the config and be confused when it doesn't work.

jordancarlin:
> The reason should be recursive, generated lean code is as follows
> 
> ```lean
> def currentlyEnabled (merge_var : extension) : SailM Bool := do
>   | Ext_H =>
>     (pure (false && ((hartSupports Ext_H) && (((_get_Misa_H (← readReg misa)) == (0b1 : (BitVec 1))) && (← (virtual_memory_supported
>                 ()))))))
> 
> def virtual_memory_supported (_ : Unit) : SailM Bool := do
>   (pure ((← (currentlyEnabled Ext_Sv32)) || ((← (currentlyEnabled Ext_Sv39)) || ((← (currentlyEnabled
>               Ext_Sv48)) || (← (currentlyEnabled Ext_Sv57))))))
> end
> ```
> 
> Need a lean expert to help :D

You might be able to fix it by adding `Ext_H` to the `currentlyEnabled_measure` function in `riscv_termination.sail` with a value of 3, which is higher than the Sv* extensions. If that doesn't work, hopefully @ineol or @tobiasgrosser can help out.

model/riscv_termination.sail
```diff
@@ -71,6 +71,7 @@ function currentlyEnabled_measure(ext : extension) -> int =
     Ext_M => 1,
     Ext_S => 1,
     Ext_V => 1,
+    Ext_H => 3,
```
trdthg: Doesn't seem to work. :(

bacam:
> You might be able to fix it by adding `Ext_H` to the `currentlyEnabled_measure` function in `riscv_termination.sail` with a value of 3, which is higher than the Sv* extensions. If that doesn't work, hopefully @ineol or @tobiasgrosser can help out.

With a value of 4 and the addition of
```
termination_measure virtual_memory_supported(_) = 3
```
I got it working in Rocq and Lean.  (Just up to that part of the model for Rocq, there's an unrelated issue in the Rocq backend that still needs fixed for the current RISC-V model.)

Timmmm:
This seems like a good way to make progress. I'll give it a couple of days before merging...

jordancarlin (approved):

PR #1025, opened 2025-06-10
Clarify the comments about ISA string generation.

pmundkur:
Mention the use of "_" to separate the single-letter extensions from the first multi-letter extension, and add a comment about the relevance to the `DeviceTree_ISA_Extensions` generator.


model/riscv_device_tree.sail
```diff
@@ -43,7 +44,12 @@ enum ISA_Format = {Canonical_Lowercase, DeviceTree_ISA_Extensions}
 
   . All multi-letter extensions, including those with the "Z" prefix, must be
     separated from other multi-letter extensions by an underscore, e.g.,
-    "RV32IMACZicsr_Zifencei".
+    "RV32IMACZicsr_Zifencei" (but see the next point).
+
+  . Underscores "_" may be used to separate ISA extensions, e.g.,
```
Timmmm: This makes it sound like the `_` is optional just after we've said it is required. I would say something like

> An underscore between the single letter extensions and the first mult-letter extension is optional. The canonical generator below includes it.

Annoyingly `riscv-config` doesn't allow it.

Issue #1002, opened 2025-06-04
Some assembly clauses missing explicit SP operand


s-khechnev:
While the RISC-V Instruction Set Manual is not really an assembly language reference, there are some examples here and there which I presume show preferred implementations of assembly syntax. The assembly definitions of instructions `c.ldsp`, `c.sdsp` are not defined with `(sp)`. But in LLVM and RISC-V ISA (Version 20250508) there are examples which include it.

Section 34.5.3
```
c.ldsp rd, offset(sp)
```

Section 34.5.4
```
c.sdsp rs2, offset(sp)
```

Assembly clauses:
```
mapping clause assembly = C_LDSP(uimm, rd)
  <-> "c.ldsp" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_6(uimm)
  when rd != zreg & xlen == 64

mapping clause assembly = C_SDSP(uimm, rs2)
  <-> "c.sdsp" ^ spc() ^ reg_name(rs2) ^ sep() ^ hex_bits_6(uimm)
  when xlen == 64
```

Also.
[...]

PR #1029, opened 2025-06-11
fix: Correct a few compressed load/store assembly

ThinkOpenly:
Assembly for load/store instructions need a base register in the form:
```
instruction destination,offset(base)
```

Within that set, stack-based instructions need to explicitly specify
the stack pointer (even though it is ignored in encoding).

Also, `fsd`, `fsw` have reversed operands.

### DIFF
--- a/model/riscv_insts_zca.sail
+++ b/model/riscv_insts_zca.sail
@@ -418,7 +418,7 @@ function clause execute (C_LWSP(uimm, rd)) = {
 }
 
 mapping clause assembly = C_LWSP(uimm, rd)
-  <-> "c.lwsp" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name(Regidx(0b00010)) ^ ")"
+  <-> "c.lwsp" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name_func(sp) ^ ")"
   when rd != zreg
 
 /* ****************************************************************** */
@@ -434,7 +434,7 @@ function clause execute (C_LDSP(uimm, rd)) = {
 }
 
 mapping clause assembly = C_LDSP(uimm, rd)
-  <-> "c.ldsp" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name(Regidx(0b00010)) ^ ")"
+  <-> "c.ldsp" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name_func(sp) ^ ")"
   when rd != zreg & xlen == 64
 
 /* ****************************************************************** */
@@ -450,7 +450,7 @@ function clause execute (C_SWSP(uimm, rs2)) = {
 }
 
 mapping clause assembly = C_SWSP(uimm, rs2)
-  <-> "c.swsp" ^ spc() ^ reg_name(rs2) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name(Regidx(0b00010)) ^ ")"
+  <-> "c.swsp" ^ spc() ^ reg_name(rs2) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name_func(sp) ^ ")"
 
 /* ****************************************************************** */
 union clause ast = C_SDSP : (bits(6), regidx)
@@ -465,7 +465,7 @@ function clause execute (C_SDSP(uimm, rs2)) = {
 }
 
 mapping clause assembly = C_SDSP(uimm, rs2)
-  <-> "c.sdsp" ^ spc() ^ reg_name(rs2) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name(Regidx(0b00010)) ^ ")"
+  <-> "c.sdsp" ^ spc() ^ reg_name(rs2) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name_func(sp) ^ ")"
   when xlen == 64
 
 /* ****************************************************************** */
--- a/model/riscv_insts_zca.sail
+++ b/model/riscv_insts_zca.sail
@@ -418,7 +418,7 @@ function clause execute (C_LWSP(uimm, rd)) = {
 }
 
 mapping clause assembly = C_LWSP(uimm, rd)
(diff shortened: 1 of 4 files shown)


jordancarlin (approved):
LGTM. Fixes #1002

jrtc27:
Should these not be using reg_name rather than hardcoding sp? Otherwise they'll always be using the ABI name, never the ISA name, and won't be configurable for CHERI either.

Timmmm:
Ah yeah we should probably use `reg_name` as @jrtc27 said (sorry should have read the comments first...).

ThinkOpenly:
Is there a better implementation than `reg_name(Regidx(0b00010))`? I tried `reg_name(2)`, `reg_name(sp)`, `reg_name(0b00010)`.

Timmmm:
`reg_name(sp)` should work I would have though, using this from `riscv_types.sail`:

```
let sp   : regidx = Regidx(0b00010) /* x2, stack pointer  */
```

ThinkOpenly:
> `reg_name(sp)` should work I would have though

I thought it would, too. It complained about `sp` being present on the right-hand side, and not on the left-hand side.

Timmmm:
Oh because it interprets it as a pattern. That's awkward. Hmm yeah I don't see any better way than this then. :-/

Unless... maybe it lets you put a function call there?

```
function sp_name() = reg_name(so)
```

ThinkOpenly:
> ```
> function sp_name() = reg_name(so)
> ```

Would something like that be preferred over the `reg_name(Regidx(X))` approach? (It's not clear that it is obviously better. :wink: )

It would be nice to force Sail to treat a mapping as a function in some cases, like (making this up, obviously) `func(reg_name(sp))`.

Timmmm:
Yeah it's kind of a universal flaw in ML languages that pattern bindings are conflated with variable names. Rust even copied this issue. Apparently it's been known for [a very long time](https://users.rust-lang.org/t/why-does-match-allow-arbitrary-identifiers/106476/11).

I always thought there should just be some different syntax for patterns... Ah well.

> Would something like that be preferred

Yeah I'm not sure I have a strong opinion but I'd say if it works it is slightly better. Happy to give way to stronger opinions though if anyone has them.

pmundkur:
Using `sp_name()` or `sp_reg_name()` seems preferable to me too.

ThinkOpenly:
My Sail-fu is not highly refined, but what I tried did not work:
```
riscv_insts_zca.sail:421.73-90:
421 |  <-> "c.lwsp" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_6(uimm) ^ "(" ^ reg_name_func(sp) ^ ")"
    |                                                                         ^---------------^
    | reg_name_func is not a union constructor or mapping in mapping-pattern reg_name_func(sp)
```

Timmmm:
I meant like `"(" ^ sp_name() ^ ")"` but unfortunately that gives the same error. :-/ Guess we'll have to go with this.

Or alternatively I feel like we could just hard-code `"sp"` like it was originally. It's definitely clearer and I think `get_config_use_abi_names()` is mostly useful for the register names that can actually vary - this is *always* `sp`.

What do you think @jordancarlin ?

pmundkur:
Due to the restrictions on `mapping` definitions, I think your initial solution of `reg_name(Regidx(0b00010))` is probably good for now.

pmundkur:
Or hardcoding `sp` is fine too.

ThinkOpenly:
@jrtc27, any opinions? Reviews seem to have circled back to hardcoded "sp", to which you originally objected. Is `reg_name(Regidx(0b00010))` preferred?

jrtc27:
> I meant like `"(" ^ sp_name() ^ ")"` but unfortunately that gives the same error. :-/ Guess we'll have to go with this.
> 
> Or alternatively I feel like we could just hard-code `"sp"` like it was originally. It's definitely clearer and I think `get_config_use_abi_names()` is mostly useful for the register names that can actually vary - this is _always_ `sp`.
> 
> What do you think @jordancarlin ?

But sp *can* vary even here:

```
% echo 'c.lwsp a0, 0(sp)' | clang -target riscv64 -march=rv64gc -x assembler - -o - -c | llvm-objdump -d - -M numeric -M no-aliases | tail -n1
       0: 02 45        	c.lwsp	x10, 0(x2)
```

jrtc27:
> @jrtc27, any opinions? Reviews seem to have circled back to hardcoded "sp", to which you originally objected. Is `reg_name(Regidx(0b00010))` preferred?

Yes, that would be preferable to me (or as a decimal number, given nobody really thinks of the register indices in binary)

Timmmm:
> But sp can vary even here:

I meant the actual register varying...

But anyway, I found a way to do it nicely. I pushed a commit. Although it has been compiling for quite a while.... hopefully this isn't hitting some compiler bug!

Timmmm:
> Although it has been compiling for quite a while

Actually it just takes a really long time to compiler the model now. :-/ Almost 5 minutes on my desktop. I'm sure it wasn't that bad in the past...

pmundkur:
> Actually it just takes a really long time to compiler the model now. :-/ Almost 5 minutes on my desktop. I'm sure it wasn't that bad in the past...

I suspect the string concatenation in the ISA string generation from #1001.

Timmmm:
Yeah 3m30 before that one.

Alasdair:
Is that with the `0.19.1` release I just made, or with `0.19`? I made quite a few performance fixes and improvements.

Timmmm:
0.19.1 I'm afraid.

Alasdair:
Is that in CI then? I think that commit is still adding 25s on my machine (from 20s total to 45s), so maybe worth changing how it's written. I think the operator overloading is causing it. We could add a helper function `vector(string) -> string` to construct the intermediate results.

Alasdair:
See https://github.com/riscv/sail-riscv/pull/1037

Timmmm:
No I tested it locally with 0.19.1, but I presume the issue affects 0.19 too?

#1037 looks good, assuming it works. Alternatively if it's just the fact that `^` is overloaded we could use another operator for string concatenation in this function that isn't overloaded? `++` or something? I always though `^` was a weird choice for concatenation anyway (and `@`  for that matter!)

Alasdair:
Both OCaml and Isabelle at least use `@` for appending lists/vectors. I do think it's quite nice having it as a single character operator. I think using `^` for string concatenation was probably a mistake, OCaml also uses it for that but I don't know of anything else.

jordancarlin (approved):
`sp_reg_name()` seems like the best solution here. LGTM

pmundkur (approved):

PR #1035, opened 2025-06-11
fix: unify naming of compressed source registers

ThinkOpenly:
Compressed source registers have two conventions:
- "rsc1"/"rsc2"
- "rs1c"/"rs2c"

Given that the latter convention is confined to just `model/riscv_insts_zcb.sail`, change it there to match the rest.


Timmmm (approved):
Seems reasonable to me. The ISA manual doesn't seem to mention these apart from a few `rs1c`s in some pseudo-Sail snippets that I don't think we need to worry about.

pmundkur (approved):

PR #1037, opened 2025-06-11
Improve type checking performance for generate_isa_string

Alasdair:
Currently Sail is can be quite slow when resolving the type for expressions containing many overloaded operators, particularly when the existing heuristics for speeding this process up do not fire. `The generate_isa_string` function hit this issue, increasing the time it takes to check the model significantly.

This PR resolves this by introducing an `append_strings` helper that allows writing `"a" ^ "b" ^ "c"` as `append_strings(["a", "b", "c"])`.

### DIFF
--- a/model/prelude.sail
+++ b/model/prelude.sail
@@ -263,3 +263,15 @@ val sys_enable_experimental_extensions = pure "sys_enable_experimental_extension
 // bit vectors that aren't a multiple of 4 bits in binary).
 function hex_bits_str forall 'n . (x : bits('n)) -> string =
   bits_str(zero_extend((3 - (('n + 3) % 4)) + 'n, x))
+
+val concat_strs : forall 'n. vector('n, string) -> string
+
+function concat_strs(strs) = {
+  var result = "";
+
+  foreach (n from 0 to (length(strs) - 1)) {
+    result = concat_str(strs[n], result)
+  };
+
+  result
+}
--- a/model/riscv_device_tree.sail
+++ b/model/riscv_device_tree.sail
@@ -81,107 +81,98 @@ function ext_wrap(ext : extension, fmt : ISA_Format) -> string = {
 }
 
 function generate_isa_string(fmt : ISA_Format) -> string = {
-  // The Sail compiler blows up if this string is constructed in one go.
-  // Build it up in pieces.
   let prefix : string = match fmt {
     Canonical_Lowercase => "rv" ^ dec_str(xlen) ^ "i",
     DeviceTree_ISA_Extensions => "\"i\"",
   };
 
-  // First append the single letter extensions.
+  let extensions = concat_strs([
+    // First the single letter extensions.
+    ext_wrap(Ext_M, fmt),
+    ext_wrap(Ext_A, fmt),
+    ext_wrap(Ext_F, fmt),
+    ext_wrap(Ext_D, fmt),
+    ext_wrap(Ext_C, fmt),
+    ext_wrap(Ext_B, fmt),
+    ext_wrap(Ext_V, fmt),
 
-  let singles = ""
-  ^ ext_wrap(Ext_M, fmt)
-  ^ ext_wrap(Ext_A, fmt)
-  ^ ext_wrap(Ext_F, fmt)
-  ^ ext_wrap(Ext_D, fmt)
-  ^ ext_wrap(Ext_C, fmt)
-  ^ ext_wrap(Ext_B, fmt)
-  ^ ext_wrap(Ext_V, fmt);
-  // S and U are not valid extensions in the device-tree, whereas H is.
+    // S and U are not valid extensions in the device-tree, whereas H will be.
 
-  // Append the Z extensions, ordered by category and then alphabetically.
-  let zi_exts = ""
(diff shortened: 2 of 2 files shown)


model/prelude.sail
```diff
@@ -263,3 +263,15 @@ val sys_enable_experimental_extensions = pure "sys_enable_experimental_extension
 // bit vectors that aren't a multiple of 4 bits in binary).
 function hex_bits_str forall 'n . (x : bits('n)) -> string =
   bits_str(zero_extend((3 - (('n + 3) % 4)) + 'n, x))
+
+val append_strings : forall 'n. vector('n, string) -> string
+
+function append_strings(strs) = {
```
Timmmm: `concat_strings` surely? :-)

Timmmm: Or maybe `concat_strs` to fit with the other `str` uses (`hex_bits_str` etc)

model/riscv_device_tree.sail
```diff
@@ -90,94 +90,101 @@ function generate_isa_string(fmt : ISA_Format) -> string = {
...
   // Append the Supervisor extensions, ordered alphabetically.
 
-  let s_exts = ""
-  ^ ext_wrap(Ext_Sscofpmf, fmt)
-  ^ ext_wrap(Ext_Sstc, fmt)
-  // Handle svade and svadu here when ready.
-  ^ ext_wrap(Ext_Svinval, fmt)
-  ^ ext_wrap(Ext_Svnapot, fmt)
-  ^ ext_wrap(Ext_Svpbmt, fmt);
+  let s_exts = append_strings([
```
arichardson: I assume we no longer need the intermediate variables now and could just use one big list?

model/prelude.sail
pmundkur: Nice, this does help to reduce the compile time.

model/riscv_device_tree.sail
pmundkur: The comment "// The Sail compiler blows up if this string is constructed in one go.
  // Build it up in pieces." up above can also be deleted.

model/prelude.sail
Alasdair: Updated

model/riscv_device_tree.sail
Alasdair: I left the `prefix` but consolidated everything else into `extensions`.

model/riscv_device_tree.sail
```diff
@@ -81,107 +81,98 @@ function ext_wrap(ext : extension, fmt : ISA_Format) -> string = {
 }
 
 function generate_isa_string(fmt : ISA_Format) -> string = {
-  // The Sail compiler blows up if this string is constructed in one go.
-  // Build it up in pieces.
```
arichardson, line 85: I think keeping a comment on why we don't use the string concatenation operator here might be useful for future readers. But feel free to ignore this nit.

arichardson (approved):

model/riscv_device_tree.sail
Alasdair: I do plan to fix it, so hopefully that comment would become outdated soon... At least as best as I can, I suspect if one is sufficiently determined it will always be possible to define something so ambiguous the compiler blows up, and that's just the price we pay for allowing fairly arbitrary ad-hoc overloading.

pmundkur (approved):

jordancarlin (approved):

model/prelude.sail
```diff
@@ -263,3 +263,15 @@ val sys_enable_experimental_extensions = pure "sys_enable_experimental_extension
 // bit vectors that aren't a multiple of 4 bits in binary).
 function hex_bits_str forall 'n . (x : bits('n)) -> string =
   bits_str(zero_extend((3 - (('n + 3) % 4)) + 'n, x))
+
+val concat_strs : forall 'n. vector('n, string) -> string
+
+function concat_strs(strs) = {
+  var result = "";
+
+  foreach (n from 0 to (length(strs) - 1)) {
```
radimkrcmar, line 272: Oh, I just sent a PR with something very very similar, see: https://github.com/riscv/sail-riscv/pull/1039

An issue I hit during development was that `default Order dec`, so this patch would likely print the dt in reverse order.

radimkrcmar: Ah, disregard that, I didn't notice that this patch is prepending the rest.

model/prelude.sail
```diff
@@ -263,3 +263,15 @@ val sys_enable_experimental_extensions = pure "sys_enable_experimental_extension
...
+    result = concat_str(strs[n], result)
```
radimkrcmar, line 273: Does sail first unroll the loop and then typecheck, so concat_str is beneficial even in this case?

Timmmm: Good point I guess you could just use `^` here and it wouldn't affect compile time badly?

Timmmm:
I think we should merge #1039  instead - it achieves the same thing but I think the code is a bit nicer. Any thoughts anyone?

jordancarlin:
Agreed. #1039 seems nicer and easier to maintain.

jordancarlin:
Closing since #1039 was merged.

PR #1039, opened 2025-06-12
Refactor ISA string generation and improve type checking performance

radimkrcmar:
Simplify the ISA string generation code using a loop over a vector. This also avoids slow type checking performance in the Sail compiler when expressions contain many overloaded operations, particularly when the existing heuristics for speeding this process up do not fire.

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -324,3 +324,98 @@ function clause hartSupports(Ext_Sv57) = config extensions.Sv57.supported : bool
 enum clause extension = Ext_Smcntrpmf
 mapping clause extensionName = Ext_Smcntrpmf <-> "smcntrpmf"
 function clause hartSupports(Ext_Smcntrpmf) = config extensions.Smcntrpmf.supported
+
+let extensions_ordered_for_isa_string = [
+  // Single letter extensions.
+  Ext_M,
+  Ext_A,
+  Ext_F,
+  Ext_D,
+  // Ext_G,
+  // Ext_Q,
+  Ext_C,
+  Ext_B,
+  // Ext_P,
+  Ext_V,
+  Ext_H,
+  // S and U are not valid extensions in the device-tree, whereas H is.
+
+  // Z extensions, ordered by category and then alphabetically.
+  // Zi
+  Ext_Zicbom,
+  Ext_Zicboz,
+  Ext_Zicntr,
+  Ext_Zicond,
+  Ext_Zicsr,
+  Ext_Zifencei,
+  Ext_Zihpm,
+  Ext_Zimop,
+
+  // Zm
+  Ext_Zmmul,
+
+  // Za
+  Ext_Zaamo,
+  Ext_Zabha,
+  Ext_Zalrsc,
+  Ext_Zawrs,
+
+  // Zf, Zd, and Zq
+  Ext_Zfa,
+  Ext_Zfh,
+  Ext_Zfhmin,
+  Ext_Zfinx,
+  Ext_Zdinx,
+  Ext_Zhinx,
+  Ext_Zhinxmin,
+
+  // Zc
+  Ext_Zca,
+  Ext_Zcb,
+  Ext_Zcd,
+  Ext_Zcf,
+  Ext_Zcmop,
+
+  // Zb
+  Ext_Zba,
+  Ext_Zbb,
+  Ext_Zbc,
+  Ext_Zbkb,
+  Ext_Zbkc,
+  Ext_Zbkx,
+  Ext_Zbs,
+
+  // Zk
+  Ext_Zknd,
+  Ext_Zkne,
+  Ext_Zknh,
+  Ext_Zkr,
+  Ext_Zksed,
+  Ext_Zksh,
+  Ext_Zkt,
+
+  // Zv
+  Ext_Zvbb,
+  Ext_Zvbc,
+  Ext_Zvkb,
+  Ext_Zvkg,
+  Ext_Zvkned,
+  Ext_Zvknha,
+  Ext_Zvknhb,
+  Ext_Zvksh,
+  Ext_Zvkt,
+
+  // Supervisor extensions, ordered alphabetically.
+  Ext_Sscofpmf,
+  Ext_Sstc,
+  // Ext_Svade,
+  // Ext_Svadu,
+  Ext_Svinval,
+  Ext_Svnapot,
+  Ext_Svpbmt,
+
+  // Hypervisor extensions, ordered alphabetically.
+
+  // Machine mode extensions, ordered alphabetically.
+  Ext_Smcntrpmf,
+]
--- a/model/riscv_device_tree.sail
+++ b/model/riscv_device_tree.sail
@@ -86,110 +86,16 @@ function ext_wrap(ext : extension, fmt : ISA_Format) -> string = {
 }
 
 function generate_isa_string(fmt : ISA_Format) -> string = {
-  // The Sail compiler blows up if this string is constructed in one go.
-  // Build it up in pieces.
-  let prefix : string = match fmt {
+  var isa_string : string = match fmt {
     Canonical_Lowercase => "rv" ^ dec_str(xlen) ^ "i",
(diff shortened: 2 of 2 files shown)


Timmmm:
I think this is slightly nicer than #1037 ... my only worry is that with this style it's quite hard to conditionally include extensions.

I've been using the output of this string for `-march` and it needs some tweaks to work with Clang and GCC:

* For `-march` it shouldn't include the `S*` extensions.
* Clang doesn't know about the `B` extension for some reason.
* GCC doesn't know about `Z[c]mop`
* Clang requires a version number for `Z[c]mop`, i.e. `_Zmop0p1_Zcmop0p2` (and also 

Might be easier to do those tweaks if it's not in a fixed array?

On the other hand we could probably just change `extensions_ordered_for_isa_string` to a function if/when we add support for printing the `-march` string.

So yeah I think we should merge this instead of #1037.

radimkrcmar:
> Might be easier to do those tweaks if it's not in a fixed array?

I guess we could handle the special cases by defining a mapping of extension and fmt that tells dt_wrap what to do in each use-case -- we could still have a single clean list of all extensions.

radimkrcmar:
> So yeah I think we should merge this instead of https://github.com/riscv/sail-riscv/pull/1037.

@Alasdair made different decisions that I'd like to understand, and possibly amend my code:
* is it better to prepend a shorter string to a longer string, than to append?
* is counting from 0 upwards to N is better than downwards from N?

Thanks.

Timmmm:
> is it better to prepend a shorter string to a longer string, than to append?

It ultimately ends at [`concat_str`](https://github.com/rems-project/sail/blob/3944e71e9dee76f8b8d88033a7cd541379bf19b9/lib/sail.c#L206-L238)

```
void concat_str(sail_string *stro, const_sail_string str1, const_sail_string str2)
{
  sail_string in1;
  sail_string in2;
  size_t in1_len = strlen(str1);
  size_t in2_len = strlen(str2);
  bool in1_free = false;
  bool in2_free = false;

  if (*stro == str1) {
    in1 = (sail_string)sail_malloc(in1_len + 1);
    strcpy(in1, str1);
    in1_free = true;
  } else {
    in1 = (sail_string)str1;
  }

  if (*stro == str2) {
    in2 = (sail_string)sail_malloc(in2_len + 1);
    strcpy(in2, str2);
    in2_free = true;
  } else {
    in2 = (sail_string)str2;
  }

  *stro = (sail_string)realloc(*stro, in1_len + in2_len + 1);
  (*stro)[0] = '\0';
  strcat(*stro, in1);
  strcat(*stro, in2);

  if (in1_free) sail_free(in1);
  if (in2_free) sail_free(in2);
}
```

So it makes no difference. That function's actually fairly suboptimal but I don't suppose it really matters.

I think this is optimal and a bit simpler (but this is suuuuper error-prone C so no guarantees I haven't missed an off-by-one or whatever).

```
void concat_str(sail_string *stro, const_sail_string str1, const_sail_string str2)
{
  size_t in1_len = strlen(str1);
  size_t in2_len = strlen(str2);

  // Resize the output so there's enough space for both strings.
  *stro = (sail_string)realloc(*stro, in1_len + in2_len + 1);

  // Special case for appending str2 to stro.
  if (*stro == str1) {
    strcat(*stro, str2);
    return;
  }

  // Write str2 to the end of stro (including null terminator).
  // Don't use memcpy here because the source and destination
  // might overlap if `*stro == str2`, e.g.
  // 
  //   var x = "bcd";
  //   x = "a" ^ x;
  //
  memmove(*stro + in2_len + 1, str2, in2_len + 1);
  // Copy str1 to the start of stro (not including null terminator).
  memcpy(*str, str1, in1_len);
}
```

For that code appending is slightly more optimal, though in practice we don't care at all.

> is counting from 0 upwards to N is better than downwards from N?

Nah that's going to make no measurable difference.

Timmmm:
Actually, even simpler and more optimal (`strcat` has to redundantly `strlen` again):

```
void concat_str(sail_string *stro, const_sail_string str1, const_sail_string str2)
{
  size_t in1_len = strlen(str1);
  size_t in2_len = strlen(str2);

  // Resize the output so there's enough space for both strings.
  *stro = (sail_string)realloc(*stro, in1_len + in2_len + 1);

  // Write str2 to the end of stro (including null terminator).
  // Don't use memcpy here because the source and destination
  // might overlap if `*stro == str2`, e.g.
  // 
  //   var x = "bcd";
  //   x = "a" ^ x;
  //
  memmove(*stro + in2_len + 1, str2, in2_len + 1);
  // Copy str1 to the start of stro (not including null terminator).
  // We can skip this for the special case of appending str2 to stro
  // when it should already be there.
  if (*stro != str1) {
    memcpy(*str, str1, in1_len);
  }
}
```

radimkrcmar:
Thanks a lot!  Your code for concat_str look much better.  I see a minor bug in memmove:
```
memmove(*stro + in1_len + 1, str2, in2_len + 1);
```
And a typo in the memcpy.  A minor optimization on top could be to detect the overlap:
```
if (*stro == str2 && in1_len < in2_len)
  memmove(*stro + in1_len + 1, str2, in2_len + 1);
else
  memcpy(*stro + in1_len + 1, str2, in2_len + 1);
```

radimkrcmar:
Ugh, off by one, of course... It's best to never touch C. :)
```
if (*stro == str2 && in1_len < in2_len)
  memmove(*stro + in1_len, str2, in2_len + 1);
else
  memcpy(*stro + in1_len, str2, in2_len + 1);
```

Timmmm:
> A minor optimization on top could be to detect the overlap:

Yeah I thought about that but in practice good implementations of `memmove` should detect this case themselves and fall back to `memcpy` so it's better to keep it simple.

> `memmove(*stro + in1_len + 1, str2, in2_len + 1);`

Oops! Good spot! I'll make a PR in case Alasdair wants it.

Timmmm:
Actually I asked Claude about this just to check and it noticed there is a use-after-free here, which just goes to show you should never write C in any circumstances. :-D

Anyway, getting off-topic. This PR looks good.

radimkrcmar:
Right, the pointers can be aliased. :/

I've updated the PR, because I noticed that the second patch fixes something that the first patch shouldn't have introduced.  The final code is unchanged, although I also accidentally rebased, so it's not that easy to see.

Timmmm:
We squash commits when merged anyway so I wouldn't worry about that.

model/riscv_extensions.sail
```diff
@@ -310,3 +310,96 @@ function clause hartSupports(Ext_Sv57) = config extensions.Sv57.supported : bool
...
+    Ext_F,
+    Ext_D,
+    // Ext_G,
+    // Ext_Q,
+    Ext_C,
+    Ext_B,
+    // Ext_P,
+    Ext_V,
+    //Ext_H,
+  // S and U are not valid extensions in the device-tree, whereas H is.
```
pmundkur: Perhaps we could have uniform indentation in this block?  The `CODE_STYLE.md` recommends a 2-space block-level indent.

pmundkur (changes requested):
This does improve on #1037.  Other than indentation fixes, this seems preferable to me.

model/riscv_extensions.sail
radimkrcmar: I'll align this one, thanks.

And when we're discussing indentation, which one do you prefer the most?
1.
```
let xs = [ 
// Zi
  Ext_Zifoo
  Ext_Zibar

// Zm
```
2.
```
let xs = [
  // Zi
  Ext_Zifoo
  Ext_Zibar

  // Zm
```
3.
```
let xs = [
  // Zi
    Ext_Zifoo
    Ext_Zibar

  // Zm
```
Actually thinking about it, I'm leaning towards 2, with newlines providing enough visual separation between categories.

pmundkur: Yes, I'd also go with 2.

pmundkur (approved):

Timmmm:
Rebased (added H, Zkt, Zvkt), and I slightly tweaked the comments (removed `Append the `).

PR #1043, opened 2025-06-13
Fix extensionName regression introduced in #1028.

pmundkur:
This caused a pattern match failure when using `--print-device-tree`.


pmundkur:
My bad for not reviewing carefully enough.  We should have a test in CI for this.

PR #1044, opened 2025-06-13
Add an cli option to print the canonical ISA string for the model configuration.

pmundkur:
This allows feeding into consumers of these strings, such as `spike --isa=<isa_string>`.  

Tested with `spike`.

Also remove the `--report-arch` option as it is now subsumed by `--print-isa-string`.


Timmmm (approved):
Nice. I think the output does need tweaking to work with GCC/Clang (e.g. GCC doesn't recognise `B` for some reason & Clang thinks `Z[c]mop` are still experimental), but we probably shouldn't try to handle all those issues, which also depend on the compiler versions.

pmundkur:
I was wondering whether I should remove `--report-arch` in this PR since it's subsumed by `--print-isa-string`?

Timmmm (approved):

jordancarlin:
> Nice. I think the output does need tweaking to work with GCC/Clang (e.g. GCC doesn't recognise `B` for some reason & Clang thinks `Z[c]mop` are still experimental), but we probably shouldn't try to handle all those issues, which also depend on the compiler versions.

Looks like the latest versions of GCC (15.1) and Clang (20.1) work fine with B/Zcmop/Zimop.

PR #1046, opened 2025-06-13
Fix scaling of compressed instruction immediates

Timmmm:
Clang expects these to be in bytes, not words.


pmundkur (changes requested):
Should we add comments explaining the different scalings used, and why the scaling in the first place?  Perhaps in the commit if its too painful for every affected clause.

Timmmm:
I dunno I think it's kind of arbitrary and there were lots of instructions that do this correctly already so I'm not sure where to put the comment (not on every such instruction surely?).

pmundkur (approved):

PR #1047, opened 2025-06-13
Respect get_config_use_abi_names() for floating point register names

Timmmm:
Previously floating point registers always used ABI names.

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -57,6 +57,7 @@ newtype fregno = Fregno : range(0, 31)
 
 function fregidx_bits(Fregidx(r) : fregidx) -> bits(5) = r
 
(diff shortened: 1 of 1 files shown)


model/riscv_fdext_regs.sail
```diff
@@ -57,6 +57,7 @@ newtype fregno = Fregno : range(0, 31)
 
 function fregidx_bits(Fregidx(r) : fregidx) -> bits(5) = r
 
+// TODO: We should have a separate Cfregidx type for compressed float registers.
```
jordancarlin, line 60: Might be worth opening an issue for this.

jordancarlin (approved):

model/riscv_fdext_regs.sail
Timmmm: #1052

pmundkur (approved):

PR #1048, opened 2025-06-13
Delegate compressed register name map

Timmmm:
Instead of having a totally new map, delegate to the main register name map. This also means that compressed register names will respect `get_config_use_abi_names()`.

### DIFF
--- a/model/riscv_regs.sail
+++ b/model/riscv_regs.sail
@@ -95,18 +95,7 @@ mapping sp_reg_name : unit <-> string = {
   () if not(get_config_use_abi_names()) <-> "x2",
 }
 
-mapping creg_name_raw : bits(3) <-> string = {
-  0b000 <-> "s0",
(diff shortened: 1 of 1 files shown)


Timmmm:
Ah good spot. I think that's nothing to do with compressed vs uncompressed; just that both names are accepted:

image

Binutils [seems to prefer `s0`](https://godbolt.org/z/T8PYex1os) so I think the correct thing to change this:

```
    0b00111 <-> "t2",
    0b01000 <-> "fp",
    0b01001 <-> "s1",
```

To this:

```
    0b00111 <-> "t2",
    0b01000 <-> "s0",
    0b01000 <-> "fp",
    0b01001 <-> "s1",
```

Mappings are checked in order so it will use `"s0"` first but this also allows decoding `"fp"` to `0b01000`.

I'll do that in another PR. I created #1053 so I don't forget.

PR #1051, opened 2025-06-13
Fix a few issues in the `vset{i}vl{i}` instructions.

pmundkur:
- Check for reserved values of `vsew` and `vlmul` in the proposed `vtype` before setting `vtype`.  If reserved values are present, set `vtype.vill` instead of asserting (as was the case for all three instructions).

- Ensure that the reserved bits of `vtype` are zeroed as required instead of blindly copying them from the proposed `vtype` (as was the case for `vsetvl`).

### DIFF
--- a/model/riscv_insts_vext_vset.sail
+++ b/model/riscv_insts_vext_vset.sail
@@ -81,12 +81,11 @@ function clause execute VSETVLI(ma, ta, sew, lmul, rs1, rd) = {
   let SEW_pow_ori   = get_sew_pow();
   let ratio_pow_ori = SEW_pow_ori - LMUL_pow_ori;
 
-  /* set vtype */
-  vtype.bits = 0b0 @ zeros(xlen - 9) @ ma @ ta @ sew @ lmul;
-
   /* check new SEW and LMUL are legal and calculate VLMAX */
-  let LMUL_pow_new = get_lmul_pow();
-  let SEW_pow_new  = get_sew_pow();
+  let (LMUL_pow_new, SEW_pow_new) : (LMUL_pow, SEW_pow) = match (lmul_pow_of_bits(lmul), sew_pow_of_bits(sew)) {
+    (Some(lmul_pow), Some(sew_pow)) => (lmul_pow, sew_pow),
+    (_, _) => { handle_illegal_vtype(); return RETIRE_SUCCESS },
+  };
   if SEW_pow_new > (LMUL_pow_new + ELEN_pow) then { handle_illegal_vtype(); return RETIRE_SUCCESS };
   let VLMAX = 2 ^ (VLEN_pow + LMUL_pow_new - SEW_pow_new);
 
@@ -105,6 +104,9 @@ function clause execute VSETVLI(ma, ta, sew, lmul, rs1, rd) = {
     let ratio_pow_new = SEW_pow_new - LMUL_pow_new;
     if (ratio_pow_new != ratio_pow_ori) then { handle_illegal_vtype(); return RETIRE_SUCCESS }
   };
+
+  /* set vtype */
+  vtype.bits = 0b0 @ zeros(xlen - 9) @ ma @ ta @ sew @ lmul;
 
   /* reset vstart to 0 */
   set_vstart(zeros());
@@ -131,12 +133,17 @@ function clause execute VSETVL(rs2, rs1, rd) = {
   let SEW_pow_ori   = get_sew_pow();
   let ratio_pow_ori = SEW_pow_ori - LMUL_pow_ori;
 
-  /* set vtype */
-  vtype.bits = X(rs2);
+  let vtype_new = Mk_Vtype(X(rs2));
+  let ma   = vtype_new[vma];
+  let ta   = vtype_new[vta];
+  let sew  = vtype_new[vsew];
+  let lmul = vtype_new[vlmul];
 
   /* check new SEW and LMUL are legal and calculate VLMAX */
-  let LMUL_pow_new = get_lmul_pow();
-  let SEW_pow_new  = get_sew_pow();
+  let (LMUL_pow_new, SEW_pow_new) : (LMUL_pow, SEW_pow) = match (lmul_pow_of_bits(lmul), sew_pow_of_bits(sew)) {
+    (Some(lmul_pow), Some(sew_pow)) => (lmul_pow, sew_pow),
+    (_, _) => { handle_illegal_vtype(); return RETIRE_SUCCESS },
+  };
   if SEW_pow_new > (LMUL_pow_new + ELEN_pow) then { handle_illegal_vtype(); return RETIRE_SUCCESS };
   let VLMAX = 2 ^ (VLEN_pow + LMUL_pow_new - SEW_pow_new);
 
@@ -155,6 +162,9 @@ function clause execute VSETVL(rs2, rs1, rd) = {
     let ratio_pow_new = SEW_pow_new - LMUL_pow_new;
     if (ratio_pow_new != ratio_pow_ori) then { handle_illegal_vtype(); return RETIRE_SUCCESS }
   };
+
+  /* set vtype */
+  vtype.bits = 0b0 @ zeros(xlen - 9) @ ma @ ta @ sew @ lmul;
 
   /* reset vstart to 0 */
   set_vstart(zeros());
@@ -177,12 +187,11 @@ mapping clause encdec = VSETIVLI(ma, ta, sew, lmul, uimm, rd)
   when currentlyEnabled(Ext_V)
 
 function clause execute VSETIVLI(ma, ta, sew, lmul, uimm, rd) = {
-  /* set vtype */
-  vtype.bits = 0b0 @ zeros(xlen - 9) @ ma @ ta @ sew @ lmul;
-
   /* check new SEW and LMUL are legal and calculate VLMAX */
-  let LMUL_pow_new = get_lmul_pow();
-  let SEW_pow_new  = get_sew_pow();
+  let (LMUL_pow_new, SEW_pow_new) : (LMUL_pow, SEW_pow) = match (lmul_pow_of_bits(lmul), sew_pow_of_bits(sew)) {
+    (Some(lmul_pow), Some(sew_pow)) => (lmul_pow, sew_pow),
+    (_, _) => { handle_illegal_vtype(); return RETIRE_SUCCESS },
+  };
   if SEW_pow_new > (LMUL_pow_new + ELEN_pow) then { handle_illegal_vtype(); return RETIRE_SUCCESS };
   let VLMAX = 2 ^ (VLEN_pow + LMUL_pow_new - SEW_pow_new);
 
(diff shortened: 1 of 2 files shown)


model/riscv_sys_regs.sail
```diff
@@ -920,15 +920,37 @@ bitfield Vtype  : xlenbits = {
 }
 register vtype : Vtype
 
+type SEW_pow = range(3, 6)
+
+function sew_pow_of_bits(sew : bits(3)) -> option(SEW_pow) =
```
Timmmm: Instead of mapping to `option` you can use `sew_pow_of_bits_forwards_matches(0b111)` which returns a bool.

model/riscv_insts_vext_vset.sail
```diff
@@ -147,15 +153,17 @@ function clause execute VSETVL(rs2, rs1, rd) = {
...
   } else if (rd != zreg) then { /* set vl to VLMAX */
-    let AVL = unsigned(ones(xlen));
     vl = to_bits_unsafe(xlen, VLMAX);
     X(rd) = vl;
-  } else { /* keep existing vl */
-    let AVL = unsigned(vl);
+  } else { /* keep existing vl if it is compatible and vtype.ill is valid */
     let ratio_pow_new = SEW_pow_new - LMUL_pow_new;
-    if (ratio_pow_new != ratio_pow_ori) then { handle_illegal_vtype(); return RETIRE_SUCCESS }
+    let is_legal_vset = ratio_pow_new == ratio_pow_ori & valid_vtype();
```
nadime15, line 174: I am wondering about the `valid_vtype()` check in the rs1=rd=x0 case. Is this necessary because a previous `vset{i}vl{i}` instruction could have failed and set `vtype.vill=1`, and we want to prevent the current instruction from accidentally bypassing that error state and appearing to succeed with the old (invalid) `vl=0` value?

nadime15 (commented):
Looks good! Removal of unnecessary variables and the flow aligns better with the spec description. I just have one question below for my understanding

model/riscv_insts_vext_vset.sail
pmundkur: Yes, I think that case is what the spec has in mind.  The spec says:
```
When rs1=x0 and rd=x0, ...  Use of the instruction with a new SEW/LMUL ratio that 
would result in a change of VLMAX is reserved. Use of the instruction is also reserved 
if vill was 1 beforehand. Implementations may set vill in either case.
```

nadime15: Ahh ok yeah, makes sense!

model/riscv_insts_vext_vset.sail
```diff
@@ -147,15 +153,17 @@ function clause execute VSETVL(rs2, rs1, rd) = {
...
+    if not(is_legal_vset) then { handle_illegal_vtype(); return RETIRE_SUCCESS }
   };
 
+  /* set vtype */
+  vtype.bits = 0b0 @ zeros(xlen - 9) @ ma @ ta @ sew @ lmul;
```
jordancarlin, line 179: Seems like this is still missing checking `vill` and the reserved bits in the new vtype:

> A vtype value with vill set is treated as an unsupported configuration.
> Implementations must consider all bits of the vtype value to determine if the configuration is supported. An unsupported value in any location within the vtype value must result in vill being set.

> In particular, all XLEN bits of the register vtype argument to the vsetvl instruction must be checked. Implementations cannot ignore fields they do not implement. All bits must be checked to ensure that new code assuming unsupported vector features in vtype traps instead of executing incorrectly on an older implementation.

If X(rs2) has anything beyond bit 7 set, we should call `handle_illegal_vtype()`.

model/riscv_insts_vext_vset.sail
```diff
@@ -133,6 +133,10 @@ function clause execute VSETVL(rs2, rs1, rd) = {
   let ratio_pow_ori = SEW_pow_ori - LMUL_pow_ori;
 
   let vtype_new = Mk_Vtype(X(rs2));
+
+  if   vtype_new[vill] == 0b1 | vtype[reserved] != zeros()
```
nadime15: I might be missing something, but if `vill` was set previously and you now try to change `vtype`, it basically wont work, you always exit here since `vtype_new[vill] == 0b1`, which makes it impossible to get a valid configuration again (at least by calling `VSETVL`).

I think it was good how it was before (?) except the additional check for the reserved bits

nadime15:
Should we clear `vstart` everytime even with invalid configurations? 

the spec says:

> All vector instructions are defined to begin execution with the element number given in the vstart
> CSR, leaving earlier elements in the destination vector undisturbed, and to reset the vstart CSR to zero
> at the end of execution.
> 
> All vector instructions, including vset{i}vl{i}, reset the vstart CSR to zero.

Does at the end of execution imply that execution was successful only? If so, should we clear `vstart`  when we return successfully?

model/riscv_sys_regs.sail
pmundkur: That's nice, but perhaps a bit arcane for a general Sail spec reader.  I don't see other usages of that idiom in the model now.  Perhaps it's easier to understand with the option?

model/riscv_insts_vext_vset.sail
pmundkur: If `vill` was set previously, it's not going to help if you try to change it by setting another illegal `vtype`?  It's possible to change it by setting a valid `vtype`, which this still allows.

pmundkur:
> Should we clear `vstart` everytime even with invalid configurations?

Good catch!  I think we should clear it, since the execute is not raising an illegal instruction exception.  This might be best put in `handle_illegal_vtype()` since 
```
 { handle_illegal_vtype(); return RETIRE_SUCCESS }
```
appears pervasively in the vector vset code.

model/riscv_insts_vext_vset.sail
nadime15: If vill was previously set and you now call `vsetvl` again with what should be a valid configuration, you would still exit here and `vill` would remain set.

> It's possible to change it by setting a valid vtype, which this still allows.

But how? If `vill` is set, you immediately exit the next time `vsetvl` is called, assuming the calls happen back to back.

nadime15:
Yeah I agree, seems to be the best approach!

model/riscv_insts_vext_vset.sail
pmundkur: Sorry, I missed what I think you were pointing out: `vtype[reserved]` should be `vtype_new[reserved]`?  Good catch!

nadime15: Oh my god, I think I mixed something up, I thought you used `vtype[vill] `and not `vtype_new[vill]`. Haha at least we caught something else!

model/riscv_sys_regs.sail
jordancarlin: I generally think the code is cleaner when we can avoid option and the forwards_matches name is pretty self explanatory, but I could go either way on this one.

pmundkur: Since forwards_matches is not documented in the Sail manual, I'm not sure it's so straightforward, whereas the option type is standard now in multiple languages. See the consensus on this comment
https://github.com/riscv/sail-riscv/pull/822#pullrequestreview-2730638630
where the context was an operator overload, which even has a full section in the Sail manual.

Timmmm: I don't think getting Sail into the ISA manual is going to happen any time soon. I'd prefer using `forwards_matches` and just documenting that in the Sail manual.

pmundkur: Let's discuss this today.   I'm wondering what the constraints are on these functions, e.g. why does Lean need this:
https://github.com/riscv/sail-riscv/blob/e7c85797ef848bc4153c782f912aea42071c3dd1/model/CMakeLists.txt#L504

pmundkur: I pushed a switch to using `forwards_matches` instead of `option`, and it's actually not as succinct as `option`.

nadime15: I have a question regarding the following line:

`assert(sew_pow_val_forwards_matches(sew_pow), "invalid vsew field in vtype");`

what's going on under the hood here? if `sew_pow_val_forwards_matches(sew_pow)` does not match anything, does the function just return False? or is it actually returning the match result? and if that’s the case, say the match is valid but returns 0, would that count as success or failure regarding the assert?

pmundkur: It returns `false` and does not return the match result.

 [These are precisely the kinds of questions which makes me want to avoid the use of these functions until better documentation is available, and they are better understood.  Also, their usage does not lead to as much succinctness as might have been expected, say as compared to just using an `option` type.]

model/riscv_sys_regs.sail
```diff
@@ -922,36 +922,32 @@ register vtype : Vtype
...
 /* the dynamic selected element width (SEW) */
 /* this returns the power of 2 for SEW */
-function get_sew_pow() -> SEW_pow =
-  match sew_pow_of_bits(vtype[vsew]) {
-    Some(p) => p,
-    None()  => {assert(false, "invalid vsew field in vtype"); 0}
-  }
+function get_sew_pow() -> SEW_pow = {
+  let sew_pow = vtype[vsew];
+  assert(sew_pow_val_forwards_matches(sew_pow), "invalid vsew field in vtype");
```
Timmmm: Can this actually happen? I think it's probably fine to rely on the `pattern match failure in sew_pow_val` error seeing as we shouldn't ever set an invalid value in the first place.

Up to you though.

model/riscv_sys_regs.sail
nadime15: Yeah, I think its just a bit unexpected when you grep for these functions and they are nowhere to be found or defined, might confuse the reader. Once the docs are in place, its probably not a big deal anymore.

pmundkur: I'd like to revert this and go back to using an option.  Reasons for:

- what @nadime15 said above (re: grepping in model/ and in sail/lib won't reveal definition)
- not a standard language idiom, and this would be its first use in the model

We could create an issue to revisit this when the Sail manual is updated, and then consider changing other uses of option or similar as well, so that this is not the only mysterious usage.  What do you think @Timmmm @jordancarlin ?

Timmmm: We could just update the manual? I don't think it will take long.

Alternatively you could add separate functions

```
function illegal_lmul_pow(v: bits(4)) -> bool = v == 0b111
function illegal_sew_pow(v: bits(3)) -> bool = v >u 0b011
```

I don't really like the `option` mapping - it adds a load of noise, and Sail mappings are implicitly allowed to be incomplete so it seems weird.

jordancarlin: I agree with @Timmmm. Even with the documentation, I remember finding `option` to be a confusing concept when first learning Sail and it adds lots of indirection to the code. I think with a couple sentences added to the manual this will be far easier to grasp.

pmundkur: Ok, I added separate functions as a compromise.

jordancarlin (approved):

PR #1055, opened 2025-06-16
Inline xlen_min_signed etc.

Timmmm:
These are only used once and `xlen_max_unsigned` as completely unused. The `0 -` is slightly unfortunate.


Alasdair:
Could do `negate(2 ^ (xlen - 1))`

Timmmm:
We have a `~` function - can we make a `-` function too? Or is that too weird?

```
-(2 ^ (xlen - 1))
```

Alasdair:
I improved the operator parsing logic a while ago by adding a separate infix expression step, so I can probably handle all infix operators better in general.

Timmmm:
> We have a ~ function - can we make a - function too?

Oh yeah we can't do that: `~` can be a function because it's a valid identifier, but `-` isn't. I'll use `negate` for now.

pmundkur (approved):

PR #1057, opened 2025-06-16
Add missing SEW_widen checks for Zvbb

Timmmm:
These instructions were missing checks that the EEW (`SEW_widen`) is not greater than 64. The spec states:

> The V extension supports EEW of 8, 16, and 32, and 64.

And also

> Note: If Zve32x is supported then Zvkb or Zvbb provide support for EEW of 8, 16, and 32. If Zve64x is supported then Zvkb or Zvbb also add support for EEW 64.

So eventually we will need to make this configurable, but it never supports >64 bits anyway.

### DIFF
--- a/model/riscv_insts_zvbb.sail
+++ b/model/riscv_insts_zvbb.sail
@@ -520,6 +520,10 @@ function clause execute (VWSLL_VV(vm, vs2, vs1, vd)) = {
   let SEW_widen      = SEW * 2;
   let LMUL_pow_widen = LMUL_pow + 1;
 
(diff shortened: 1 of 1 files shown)


pmundkur (approved):
LGTM.  It looks like some widening instructions in `_arith` are also missing the asserts on `SEW_widen`, but can't tell immediately if they are needed there as well.

Timmmm:
Are there any that I missed in #1058 ? Those were all the ones that I needed to add/move to get type checking to work when `/` and `%` require positive arguments and non-zero divisor.

pmundkur:
Great, it looks like #1058 covers the ones I found.

jordancarlin (approved):

PR #1058, opened 2025-06-16
Add/move assertions for SEW_widen size

Timmmm:
This is necessary for future improvements to the vector types.


Timmmm:
I didn't notice actually (this was a very mechanical change), but I think you're right - they can easily be consolidated. I created #1064 so we don't forget.

PR #1060, opened 2025-06-16
Remove speculate_conditional

Timmmm:
I've previously used this to inject sprurious Store-Conditional failures, however it required some tweaking because it's called very early before various exceptions can happen.

A better solution is to spuriously cancel the reservation directly in `load_reservation()` or by calling `cancel_reservation()`.

This change doesn't actually do that but I added notes explaining that you can. In future when we provide a library interface we can do it like that.

### DIFF
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -92,8 +92,6 @@ function check_CSR(csr : csreg, p : Privilege, isWrite : bool) -> bool =
  * where cancellation can be performed.
  */
 
-val speculate_conditional = impure {interpreter: "excl_res", c: "speculate_conditional", lem: "speculate_conditional_success"} : unit -> bool
-
 val load_reservation = impure {interpreter: "Platform.load_reservation", c: "load_reservation", lem: "load_reservation"} : physaddrbits -> unit
 val match_reservation = pure {interpreter: "Platform.match_reservation", lem: "match_reservation", c: "match_reservation"} : physaddrbits -> bool
 val cancel_reservation = impure {interpreter: "Platform.cancel_reservation", c: "cancel_reservation", lem: "cancel_reservation"} : unit -> unit
--- a/c_emulator/riscv_platform.cpp
+++ b/c_emulator/riscv_platform.cpp
@@ -24,17 +24,17 @@ mach_bits plat_get_16_random_bits(unit)
   return rv_16_random_bits();
 }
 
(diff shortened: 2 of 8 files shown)


model/riscv_sys_control.sail
```diff
@@ -92,10 +92,12 @@ function check_CSR(csr : csreg, p : Privilege, isWrite : bool) -> bool =
  * where cancellation can be performed.
  */
 
-val speculate_conditional = impure {interpreter: "excl_res", c: "speculate_conditional", lem: "speculate_conditional_success"} : unit -> bool
-
+// `match_reservation` is impure because Store-Conditional is allowed
+// to spuriously fail and the mechanism you can use to do that is
+// to return `false` from `match_reservation()` where it *could* have
+// returend `true`.
```
radimkrcmar: Why can't we use cancel_reservation if we want subsequent match_reservation to return false?

Timmmm: Because the SC can spuriously fail and then succeed on a later attempt. If you `cancel_reservation` it will fail all subsequent SCs until you do a LR again.

radimkrcmar: I don't think it can happen: `Regardless of success or failure, executing an SC.W instruction invalidates any reservation held by this hart.`
Software is always expected to redo the whole LR/SC loop.

Timmmm: Oh good point. Yeah I think you're right then. Ok I'll change the comments and make it pure again.

I guess formal environments could simulate this without having to call `cancel_reservation` just by having `load_reservation` sometimes not do anything?

radimkrcmar: Yes, that sounds like a good optimization.  There are guarantees that the environments must follow, but those guarantees allow a hart to get stuck if a device writes to the reserved address with "infinite" frequency, so even a load_reservation that does nothing is correct for some models. :)

Timmmm: Updated.

radimkrcmar (approved):
That's the kind of diff I like to see.

jordancarlin (approved):
Nice simplification

pmundkur:
@Alasdair @bacam Could you review this PR?  AFAIR, this affects the formal memory model for LR/SC.

Alasdair:
Right now in the axiomatic model, we allow speculate_conditional to nondeterministically return true or false, but I don't think moving the failure somewhere else would break anything there.

pmundkur (approved):

PR #1061, opened 2025-06-17
Add current virtualization mode register

trdthg:
Add current virtualization mode register

- Add new enum `Virtualization`
- Add new register `cur_virtualization`
- Add `virt_priv_to_str `and `two_stage_translation_enabled `function

---

I deleted the `type virt_mode = bits(1)` and related functions defined in the original commit, because they were not actually used

reference:

- spec from [privileged 21.1.(0)](https://riscv.github.io/riscv-isa-manual/snapshot/privileged/#_privilege_modes)
- splite from the first commit [Setup for hypervisor extension](https://github.com/riscv/sail-riscv/pull/612/commits/b2e56bbbc3216eedae0907308bab3fde5a014e8f) of #612 

> I will try to align the PR with the specific spec sections for easier review

### DIFF
--- a/model/riscv_types_hext.sail
+++ b/model/riscv_types_hext.sail
@@ -1,0 +1,30 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+// The current virtualization mode
+// indicates whether the hart is currently executing in a guest
+enum Virtualization = { V0, V1 }
+
+function virt_priv_to_str(priv : Privilege, virt: Virtualization) -> string = {
+  match (priv, virt) {
+    (User      , V0) => "U",  // User mode
+    (Supervisor, V0) => "HS", // Hypervisor-extended supervisor mode
+    (Machine   , V0) => "M",  // Machine mode
+    (User      , V1) => "VU", // Virtual user mode
+    (Supervisor, V1) => "VS", // Virtual supervisor mode
+    (Machine   , V1) => internal_error(__FILE__, __LINE__, "illegal privilege mode"),
+  }
+}
+overload to_str = {virt_priv_to_str}
+
+function two_stage_translation_enabled(priv : Privilege, virt : Virtualization) -> bool = {
+  match (priv, virt) {
+    (priv, V1) if priv != Machine => true,
+    _ => false,
+  }
+}
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -161,6 +161,7 @@ foreach (xlen IN ITEMS 32 64)
                 "riscv_regs.sail"
                 "riscv_pc_access.sail"
                 "riscv_sys_regs.sail"
+                "riscv_hext_regs.sail"
                 "riscv_pmp_regs.sail"
                 "riscv_pmp_control.sail"
                 "riscv_ext_regs.sail"
@@ -175,6 +176,7 @@ foreach (xlen IN ITEMS 32 64)
                 "riscv_types_common.sail"
                 "riscv_types_ext.sail"
                 "riscv_types.sail"
+                "riscv_types_hext.sail"
                 "riscv_vmem_types.sail"
                 ${sail_regs_srcs}
                 ${sail_sys_srcs}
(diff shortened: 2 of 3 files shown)


model/riscv_types_hext.sail
```diff
@@ -0,0 +1,30 @@
...
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+// The current virtualization mode
+// indicates whether the hart is currently executing in a guest
+enum Virtualization = { V0, V1 }
+
+function virt_priv_to_str(priv : Privilege, virt: Virtualization) -> string = {
```
Timmmm, line 13: This could be a mapping:

```
mapping virt_priv_str : (Privilege, Virtualization) <-> string = {
    (User      , V0) <-> "U",  // User mode
    (Supervisor, V0) <-> "HS", // Hypervisor-extended supervisor mode
    (Machine   , V0) <-> "M",  // Machine mode
    (User      , V1) <-> "VU", // Virtual user mode
    (Supervisor, V1) <-> "VS", // Virtual supervisor mode
}
```

You could include `forwards (Machine, V1) => ...` but it's not strictly necessary (it will just give you a slightly better error message).

model/riscv_types_hext.sail
```diff
@@ -0,0 +1,30 @@
...
+    (Supervisor, V0) => "HS", // Hypervisor-extended supervisor mode
+    (Machine   , V0) => "M",  // Machine mode
+    (User      , V1) => "VU", // Virtual user mode
+    (Supervisor, V1) => "VS", // Virtual supervisor mode
+    (Machine   , V1) => internal_error(__FILE__, __LINE__, "illegal privilege mode"),
+  }
+}
+overload to_str = {virt_priv_to_str}
+
+function two_stage_translation_enabled(priv : Privilege, virt : Virtualization) -> bool = {
```
Timmmm, line 25:
```
function two_stage_translation_enabled(priv : Privilege, virt : Virtualization) -> bool =
   virt == V1 & priv != Machine
```

Timmmm (changes requested):
I guess the alternative is to extend the `Privilege` enum...

```
enum Privilege = {
  User,
  VirtualUser,
  Supervisor,
  VirtualSupervisor,
  Machine,
}
```

I guess that might require a lot of code changes in places where we compare to `User` and have to change them all to `x == User | x == VirtualUser`... It probably is more explicit though. What do you think?

model/riscv_types_hext.sail
```diff
@@ -0,0 +1,30 @@
...
+  match (priv, virt) {
+    (priv, V1) if priv != Machine => true,
```
radimkrcmar, line 27: I think `(Machine, V1)` should be an internal error everywhere, but it doesn't seem necessary to check everywhere, so the condition to apply the G-stage mapping seems better just as `virt == V1`.

If we don't want the internal state to actually be a bool, maybe we could have a general function for `virt == V1`?

model/riscv_types_hext.sail
trdthg: `string -> (Privilege, Virtualization)` has never been used in the origin PR, So I think this is a bit redundant

trdthg:
>  guess that might require a lot of code changes in places where we compare to User and have to change them all to x == User | x == VirtualUser... It probably is more explicit though.

I thought about it again and feel that (maybe) it's good. Almost all places that use `cur_privilege` need to handle the VU and VS states separately, so this won't bring too much redundancy.

I also want to add `TODO: check again after H ext is completed` to these places

model/riscv_types_hext.sail
trdthg: > have a general function for virt == V1

Looks good, but how about delete this function for now (actually, I wrote it here just to cover all the spec ch21.1 said) and reconsider it when we actually implement two-stage address translation

radimkrcmar: I like that.

model/riscv_types_hext.sail
pmundkur: Yeah, leaving as a function should be fine.

Timmmm: I think a mapping is still better even if we don't use the other direction. It *is* a mapping so we should use Sail's native feature for it.

trdthg:
I'll try to submit another version of the PR later, let's see how it goes.

jordancarlin:
@trdthg Can this be closed now that #1067 is merged instead?

trdthg:
Of course

PR #1062, opened 2025-06-17
Remove explicit compressed instruction hints

Timmmm:
Remove the explicit hint decoding. This was unnecessary because currently the behaviour of hints is exactly the same as executing them normally. It also caused problems because it generated invalid assembly for those instructions.


model/riscv_insts_zca.sail
```diff
@@ -13,17 +13,6 @@
  * the check in one place in the fetch-execute logic.
  */
 
-/* ****************************************************************** */
-union clause ast = C_NOP : unit
-
-mapping clause encdec_compressed = C_NOP()
```
jordancarlin, line 19: `c.nop` is listed in the spec as a distinct instruction from `c.addi`. The Sail model should reflect that and keep this separate from `c.addi` below.

Timmmm: Well, sort of. It's deliberately assigned to naturally-nop instructions so we don't necessarily need to explicitly support it. Maybe it would be clearer though separately. I'll update this.

I think the spec needs a lot of clarification around these hint instructions anyway. Maybe I'll make a PR.

Timmmm: PR to clarify spec: https://github.com/riscv/riscv-isa-manual/pull/2088

I pushed an update with `C_NOP` restored, and I also fixed it so it matches the spec (`c.nop` actually has an immediate argument) and generates correct assembly for non-zero immediates.

model/riscv_insts_zca.sail
```diff
@@ -205,14 +215,15 @@ function clause execute (C_LUI(imm, rd)) = {
...
   <-> "c.lui" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_signed_6(imm)
-  when rd != zreg & rd != sp & imm != 0b000000
+  when rd != sp & imm != 0b000000
 
 /* ****************************************************************** */
 union clause ast = C_SRLI : (bits(6), cregidx)
 
+// c.srli with shamt=0 are hints.
 mapping clause encdec_compressed = C_SRLI(nzui5 @ nzui40, rsd)
   <-> 0b100 @ nzui5 : bits(1) @ 0b00 @ encdec_creg(rsd) @ nzui40 : bits(5) @ 0b01
```
pmundkur: Perhaps the `nz` can be removed from the variable prefix now
```suggestion
mapping clause encdec_compressed = C_SRLI(ui5 @ ui40, rsd)
  <-> 0b100 @ ui5 : bits(1) @ 0b00 @ encdec_creg(rsd) @ ui40 : bits(5) @ 0b01
```
and similarly for the `nzui*` using instructions below

model/riscv_insts_zca.sail
```diff
@@ -205,14 +215,15 @@ function clause execute (C_LUI(imm, rd)) = {
...
-  when nzui5 @ nzui40 != 0b000000 & (xlen == 64 | nzui5 == 0b0) & currentlyEnabled(Ext_Zca)
+  when (xlen == 64 | nzui5 == 0b0) & currentlyEnabled(Ext_Zca)
```
pmundkur:
```suggestion
  when (xlen == 64 | ui5 == 0b0) & currentlyEnabled(Ext_Zca)
```

model/riscv_insts_zca.sail
```diff
@@ -118,18 +123,20 @@ mapping clause assembly = C_SD(uimm, rsc1, rsc2)
 /* ****************************************************************** */
 union clause ast = C_ADDI : (bits(6), regidx)
 
-mapping clause encdec_compressed = C_ADDI(nzi5 @ nzi40, rsd)
-  <-> 0b000 @ nzi5 : bits(1) @ encdec_reg(rsd) @ nzi40 : bits(5) @ 0b01
-  when nzi5 @ nzi40 != 0b000000 & rsd != zreg & currentlyEnabled(Ext_Zca)
+// Code points with rd=x0 are c.nop.
```
pmundkur:
```suggestion
// Code points with rsd=x0 are c.nop.
```

model/riscv_insts_zca.sail
```diff
@@ -394,16 +404,16 @@ mapping clause assembly = C_BNEZ(imm, rs)
 /* ****************************************************************** */
 union clause ast = C_SLLI : (bits(6), regidx)
 
+// c.slli with shamt=0 or rd=x0 are hints.
```
pmundkur:
```suggestion
// c.slli with shamt=0 or rsd=x0 are hints.
```

model/riscv_insts_zca.sail
```diff
@@ -525,13 +537,15 @@ mapping clause assembly = C_EBREAK() <-> "c.ebreak"
 /* ****************************************************************** */
 union clause ast = C_ADD : (regidx, regidx)
 
+// Code points with rs2=x0 are c.jalr/c.ebreak.
+// c.add with rd=x0 are hints.
```
pmundkur:
```suggestion
// c.add with rsd=x0 are hints.
```

model/riscv_regs.sail
```diff
@@ -95,6 +95,11 @@ mapping sp_reg_name : unit <-> string = {
   () if not(get_config_use_abi_names()) <-> "x2",
 }
 
+mapping zero_reg_name : unit <-> string = {
```
pmundkur: This mapping doesn't seem to be used in this PR?

Timmmm: Ah oops, I thought I needed it but ended up not.

model/riscv_insts_zca.sail
jordancarlin: The isa manual uses `shamt` as the name for this immediate (for c.slli, c.srli, and c.srai), which might make more sense here as well for consistency.

model/riscv_regs.sail
Timmmm: Fixed.

model/riscv_insts_zca.sail
Timmmm: Yeah makes sense - the non-compressed ones used `shamt` too so I changed it to that.

jordancarlin (approved):

pmundkur (approved):

PR #1063, opened 2025-06-18
Split rvfi_dii v1 and v2

trdthg:
A small change to make the code cleaner

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -151,6 +151,8 @@ foreach (xlen IN ITEMS 32 64)
                 "prelude_mem.sail"
                 "arithmetic.sail"
                 "rvfi_dii.sail"
+                "rvfi_dii_v1.sail"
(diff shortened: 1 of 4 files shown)


Timmmm (approved):
LGTM. Do we actually need to support V1 any more? @PeterRugg @jrtc27

jordancarlin (approved):

PeterRugg:
> LGTM. Do we actually need to support V1 any more? @PeterRugg @jrtc27

Hmm, I'm not sure about that. I think we use V2 in testrig and it's able to just compare the V1 fields if the other implementation doesn't have V2. I think V1 is more along the lines of "standard" RVFI though, as opposed to TestRIG specific. It's all designed to be backwards compatible (@arichardson did this)

Timmmm:
Ah ok lets just leave it in for now then.

pmundkur (approved):

PR #1067, opened 2025-06-19
Add VU and VS privilege level

trdthg:
- Add `VirtualUser` and `VirtualSupervisor` to the existing `Privilege` enum
- Rename `privLevel_bits` to `nominalPrivLevel_bits` and add conversions for VS and VU to bits
- Add placeholder to all `match cur_privilege` statements, mark VS and VU branches as unimplemented

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -508,7 +508,7 @@ mapping clause encdec = ECALL()
   <-> 0b000000000000 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
 
 function clause execute ECALL() = {
-  let trap : ExceptionType = match (cur_privilege) {
+  let trap : ExceptionType = match cur_privilege {
     User              => E_U_EnvCall(),
     Supervisor        => E_S_EnvCall(),
     Machine           => E_M_EnvCall(),
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -508,14 +508,15 @@ mapping clause encdec = ECALL()
   <-> 0b000000000000 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
 
 function clause execute ECALL() = {
+  let trap : ExceptionType = match (cur_privilege) {
+    User              => E_U_EnvCall(),
+    Supervisor        => E_S_EnvCall(),
+    Machine           => E_M_EnvCall(),
+    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+  };
   let t : sync_exception = struct {
-    trap = match (cur_privilege) {
-      User              => E_U_EnvCall(),
-      Supervisor        => E_S_EnvCall(),
-      Machine           => E_M_EnvCall(),
-      VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-      VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    },
+    trap    = trap,
     excinfo = (None() : option(xlenbits)),
     ext     = None(),
   };
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -508,14 +508,17 @@ mapping clause encdec = ECALL()
   <-> 0b000000000000 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
 
 function clause execute ECALL() = {
-  let t : sync_exception =
-    struct { trap = match (cur_privilege) {
-                      User       => E_U_EnvCall(),
-                      Supervisor => E_S_EnvCall(),
-                      Machine    => E_M_EnvCall()
-                    },
-             excinfo = (None() : option(xlenbits)),
-             ext     = None() };
+  let t : sync_exception = struct {
+    trap = match (cur_privilege) {
+      User              => E_U_EnvCall(),
+      Supervisor        => E_S_EnvCall(),
+      Machine           => E_M_EnvCall(),
+      VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+      VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    },
+    excinfo = (None() : option(xlenbits)),
+    ext     = None(),
+  };
   Trap(cur_privilege, CTL_TRAP(t), PC)
 }
 
@@ -548,9 +551,11 @@ mapping clause encdec = SRET()
 
 function clause execute SRET() = {
   let sret_illegal : bool = match cur_privilege {
-    User       => true,
-    Supervisor => not(currentlyEnabled(Ext_S)) | mstatus[TSR] == 0b1,
-    Machine    => not(currentlyEnabled(Ext_S))
+    User              => true,
+    Supervisor        => not(currentlyEnabled(Ext_S)) | mstatus[TSR] == 0b1,
+    Machine           => not(currentlyEnabled(Ext_S)),
+    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
   };
   if   sret_illegal
   then Illegal_Instruction()
@@ -583,11 +588,13 @@ mapping clause encdec = WFI()
 
 function clause execute WFI() =
   match cur_privilege {
-    Machine    => Enter_Wait(WAIT_WFI),
-    Supervisor => if   mstatus[TW] == 0b1
-                  then Illegal_Instruction()
-                  else Enter_Wait(WAIT_WFI),
-    User       => Illegal_Instruction()
+    Machine           => Enter_Wait(WAIT_WFI),
+    Supervisor        => if   mstatus[TW] == 0b1
+                         then Illegal_Instruction()
+                         else Enter_Wait(WAIT_WFI),
+    User              => Illegal_Instruction(),
+    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
   }
 
 mapping clause assembly = WFI() <-> "wfi"
@@ -606,12 +613,14 @@ function clause execute SFENCE_VMA(rs1, rs2) = {
   // is 9 but we always set it to 16 for RV64.
   let asid = if rs2 != zreg then Some(X(rs2)[asidlen - 1 .. 0]) else None();
   match cur_privilege {
-    User       => Illegal_Instruction(),
-    Supervisor => match mstatus[TVM] {
-                    0b1 => Illegal_Instruction(),
-                    0b0 => { flush_TLB(asid, addr); RETIRE_SUCCESS },
-                  },
-    Machine    => { flush_TLB(asid, addr); RETIRE_SUCCESS }
+    User              => Illegal_Instruction(),
+    Supervisor        => match mstatus[TVM] {
+                           0b1 => Illegal_Instruction(),
+                           0b0 => { flush_TLB(asid, addr); RETIRE_SUCCESS },
+                         },
+    Machine           => { flush_TLB(asid, addr); RETIRE_SUCCESS },
+    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
   }
 }
 
--- a/model/riscv_insts_zicbom.sail
+++ b/model/riscv_insts_zicbom.sail
@@ -50,7 +50,7 @@ enum checked_cbop = {CBOP_ILLEGAL, CBOP_ILLEGAL_VIRTUAL, CBOP_INVAL_FLUSH, CBOP_
 // Select the cbop to perform based on the privilege.
 function cbop_priv_check(p: Privilege) -> checked_cbop = {
   let mCBIE : cbie = encdec_cbie(menvcfg[CBIE]);
-  // TODO: check henvcfg after hyppervisor is implemented
+  // TODO: check henvcfg after hypervisor is implemented
   let sCBIE : cbie = if   currentlyEnabled(Ext_S)
                      then encdec_cbie(senvcfg[CBIE])
                      else encdec_cbie(menvcfg[CBIE]);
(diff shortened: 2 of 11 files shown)


model/riscv_types.sail
```diff
@@ -58,23 +58,27 @@ mapping architecture : Architecture <-> arch_xlen = {
 /* privilege levels */
 
 type priv_level = bits(2)
```
Timmmm: Dunno how many places this is used but if it isn't many maybe we should rename this to `nominal_priv_level` or something (`nom_priv_level`? `nom_priv_bits`?).

Timmmm:
> I suspect there are places that are like:

I just did this search quickly and as far as I can see there aren't any, so don't worry about that. :tada:

trdthg:
> searching for Machine, Supervisor and User

I found a few places use things like `p == Machine`:

- `check_CSR(csr, p)` and related functions
- `execute SFENCE_W_INVAL`
- `pmpCheck`

I'll check again later and add TODO comments if necessary.

model/riscv_smcntrpmf.sail
```diff
@@ -15,8 +15,8 @@ function legalize_smcntrpmf(c : CountSmcntrpmf, value : bits(64)) -> CountSmcntr
     MINH  = v[MINH],
     SINH  = if currentlyEnabled(Ext_S) then v[SINH] else 0b0,
     UINH  = if currentlyEnabled(Ext_U) then v[UINH] else 0b0,
-    // VSINH = v[VSINH],
-    // VUINH = v[VUINH],
+    VSINH = if currentlyEnabled(Ext_S) & currentlyEnabled(Ext_H) then v[VSINH] else 0b0,
+    VUINH = if currentlyEnabled(Ext_U) & currentlyEnabled(Ext_H) then v[VUINH] else 0b0,
```
jordancarlin: H requires S (and therefore U), so this is redundant.

model/riscv_types.sail
```diff
@@ -58,23 +58,27 @@ mapping architecture : Architecture <-> arch_xlen = {
...
+function privLevel_of_bits(b : priv_level) -> Privilege = nominalPrivLevel_bits(b)
 
 function privLevel_to_str (p : Privilege) -> string =
   match (p) {
-    User       => "U",
-    Supervisor => "S",
-    Machine    => "M"
+    User              => "U",
+    VirtualUser       => "VU",
+    Supervisor        => "HS",
```
jordancarlin: I don't think we want to unconditionally print out Supervisor mode as "HS". Really it should be "S" when H is not enabled and "HS" only when it is.

model/riscv_types.sail
trdthg: I like `nom_priv_bits`.
I also added a `type virt_mode_bits = bits(1)`, which is used for the mapping `privilege <-> (nom_priv_bits, virt_mode_bits)`

model/riscv_types.sail
trdthg: done

model/riscv_smcntrpmf.sail
trdthg: removed

trdthg:
> add TODO comments if necessary.

Basically all the places use `privilege ==` directly has been added
- most places relying on new registers
- SFENCE has two new instructions
- the rest are trap and address translate related

model/riscv_types.sail
```diff
@@ -52,26 +52,32 @@ mapping architecture : Architecture <-> arch_xlen = {
...
 }
 
-/* privilege levels */
+/* nominal privilege levels */
+type nom_priv_bits = bits(2)
 
-type priv_level = bits(2)
-enum Privilege  = {User, Supervisor, Machine}
+/* virtualization modes */
+type virt_mode_bits = bits(1)
```
pmundkur: Can't this just be a `bit` instead of `bits(1)`?

model/riscv_types.sail
```diff
@@ -52,26 +52,32 @@ mapping architecture : Architecture <-> arch_xlen = {
...
-  Machine    <-> 0b11,
-  backwards 0b10 => internal_error(__FILE__, __LINE__, "Invalid privilege level: " ^ BitStr(0b10))
+enum Privilege  = {User, VirtualUser, Supervisor, VirtualSupervisor, Machine}
+
+mapping privLevel_bits : (nom_priv_bits, virt_mode_bits) <-> Privilege = {
+  (0b00, 0b0) <-> User,
+  (0b00, 0b1) <-> VirtualUser,
+  (0b01, 0b0) <-> Supervisor,
+  (0b01, 0b1) <-> VirtualSupervisor,
+  (0b11, 0b0) <-> Machine,
```
pmundkur: We should have an `internal_error()` for the other cases.  The default match failure message is not very helpful.

model/riscv_sys_exceptions.sail
```diff
@@ -20,10 +20,12 @@ function handle_trap_extension(p : Privilege, pc : xlenbits, u : option(unit)) -
...
   let tvec : Mtvec = match p {
-                       Machine    => mtvec,
-                       Supervisor => stvec,
-                       User       => internal_error(__FILE__, __LINE__, "Invalid privilege level"),
-                     };
+    Machine    => mtvec,
+    Supervisor => stvec,
+    User       => internal_error(__FILE__, __LINE__, "Invalid privilege level"),
+    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
```
pmundkur: Might as well indent these consistently.

model/riscv_sys_control.sail
```diff
@@ -230,54 +245,60 @@ function trap_handler(del_priv : Privilege, intr : bool, c : exc_code, pc : xlen
...
-       stval           = tval(info);
-       sepc            = pc;
+      mstatus[SPIE] = mstatus[SIE];
+      mstatus[SIE]  = 0b0;
+      mstatus[SPP]  = match cur_privilege {
+        User => 0b0,
+        Supervisor => 0b1,
+        Machine => internal_error(__FILE__, __LINE__, "invalid privilege for s-mode trap"),
+        VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+        VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
```
pmundkur: Might as well indent these consistently.

model/riscv_sys_control.sail
```diff
@@ -46,8 +50,13 @@ function check_Counteren(csr : csreg, p : Privilege) -> bool = {
...
   // Check if it is not stimecmp.
   if csr != 0x14D & csr != 0x15D then return true;
-
-  p == Machine | (p == Supervisor & mcounteren[TM] == 0b1 & menvcfg[STCE] == 0b1)
+  match p {
+    Machine => true,
+    Supervisor => mcounteren[TM] == 0b1 & menvcfg[STCE] == 0b1,
+    User => false,
+    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
```
pmundkur: Might as well indent these consistently.

model/riscv_types.sail
nadime15: Additionally, I think it would make sense to introduce some constants or use existing constants (if possible) in this case instead of using the raw bits. Otherwise its not very readable

trdthg: > introduce some constants or use existing constants

maybe impossible, sail doesn't have `const` keyword or `#define` macro

`type xlen : Int = 64` seems couldn't be used for `bits(2)`

`bitone` and `bitzero` are defined in sail compiler (since lexer)

jrtc27: `type` and `let` define immutable quantities. In many languages immutability is the default and you have to opt in to mutability (`var` in Sail), unlike something like C (which I guess is what you're familiar with given you mention `const` and `#define`) where it's opt out.

nadime15: > maybe impossible, sail doesn't have const keyword or #define macro

True, but in the worst case, you could use `let` with the following expression:

`let virt_mode_on : virt_mode_bit = bitone`

just as an idea, or an Enum + Mapping.

Others may also have an opinion on this.

Timmmm: The issue is you can't use those constants in patterns. This doesn't work like you'd want:

```
let machine = 0b11;
match priv {
  machine =>
```

This came up recently with sp-relative instruction assembly and the best solution we found is a mapping that maps unit to some value (`sp_reg_name()`) but it kind of sucks.

trdthg: > type and let define immutable quantities

Thank you for the explanation, but I also found the main issue here is that mapping doesn't trying to find the value of the identidier from the context.

```
let UserBits = 0b00

Type error:
riscv_types.sail:66.3-11:
66 |  (UserBits, bitzero) <-> User,
   |   ^------^
   | Identifier UserBits found on left hand side of mapping, but not on right
```

jrtc27: Oh that's the context, sure. You can do something like `match priv { x if x = machine` but that's kinda gross.

jrtc27: > Thank you for the explanation, but I also found the main issue here is that mapping doesn't trying to find the value of the identidier from the context.

Well that's a symptom of the wider problem, that for things that aren't keywords like bitzero/bitone that syntax means "bind a new identifier UserBits to the value in the first element of the tuple", it's the same as writing `(x, bitzero)`. For a forwards-only mapping that means it silently matches everything. For a bidirectional mapping the forwards direction will silently match everything, but the backwards direction isn't valid because it's expecting to be told what value the identifier should have in that direction (like `x <-> foo(x)`).

model/riscv_types.sail
```diff
@@ -52,26 +52,33 @@ mapping architecture : Architecture <-> arch_xlen = {
...
+  (0b01, bitzero) <-> Supervisor,
+  (0b01, bitone ) <-> VirtualSupervisor,
+  (0b11, bitzero) <-> Machine,
+  forwards _       => internal_error(__FILE__, __LINE__, "Invalid privilege level or virtual mode"),
 }
 
-function privLevel_to_bits(p : Privilege) -> priv_level = privLevel_bits(p)
-function privLevel_of_bits(b : priv_level) -> Privilege = privLevel_bits(b)
+function privLevel_to_bits(p : Privilege) -> nom_priv_bits = { let (p, _) = privLevel_bits(p); p }
+function privLevel_of_bits(p : nom_priv_bits, v : virt_mode_bit) -> Privilege = privLevel_bits((p, v))
```
pmundkur: These two functions are asymmetric, which is confusing.  Could they be made symmetric, or does that significantly increase the size of the PR?  I suspect making them symmetric will reveal places where you need to use `cur_virt_mode`.

model/riscv_types.sail
```diff
@@ -52,26 +52,33 @@ mapping architecture : Architecture <-> arch_xlen = {
...
 }
 
-/* privilege levels */
+/* nominal privilege levels */
+type nom_priv_bits = bits(2)
 
-type priv_level = bits(2)
-enum Privilege  = {User, Supervisor, Machine}
+/* virtualization modes */
+type virt_mode_bit = bit
```
pmundkur, line 68: This type is defined in this PR, but no state (e.g. `cur_virt_mode`) with that type.   I'd prefer to define the state in this PR; later PRs can handle modifying it.

model/riscv_types.sail
Timmmm: I think `privLevel_to_bits()` is deliberately used to drop the `virt_mode_bits` where it isn't needed. `privLevel_of_bits` isn't really needed.

I think I would remove `privLevel_of_bits`  at least. Maybe rename `privLevel_bits` to `privLevel_encdec`?

It is kind of awkward because Sail doesn't have a syntax to access tuple elements. Another option would be to use a `struct`:

```
struct PrivBits {
  priv : bits(2),
  virt : bit,
}
```

You can use that in the `mapping`, and then you can do something like this:

```
if privLevel_bits(deleg).priv <_u privLevel_bits(p).priv
```

Not sure that is better...

pmundkur: Yes, it looks like the one use of `privLevel_of_bits` in `effectivePrivilege` could be removed to directly use `privLevel_bits`.  To keep this PR small, we should preferably do any renaming in a different PR.

model/riscv_types.sail
trdthg: But `cur_privilege: Privilege` already includes both `nom_priv` and `virt` infomation, and I feel that adding another variable would increase redundancy, so I prefer to change `privLevel_to_bits` to the following two functions
- `privLevel_to_nom_bits`
- `privLevel_to_virt_bit`

KotorinMinami:
Are there any issues with this PR? If not, can we push forward? @Timmmm  @pmundkur  @jordancarlin @jrtc27 @nadime15

model/riscv_sys_control.sail
```diff
@@ -221,7 +229,7 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       let prev_priv = cur_privilege;
       mstatus[MIE]  = mstatus[MPIE];
       mstatus[MPIE] = 0b1;
-      cur_privilege = privLevel_of_bits(mstatus[MPP]);
+      cur_privilege   = privLevel_bits(mstatus[MPP], bitzero); // TODO: use mstatus[MPV] if hypervisor enabled
```
pmundkur: Minor alignment issue
```suggestion
      cur_privilege = privLevel_bits(mstatus[MPP], bitzero); // TODO: use mstatus[MPV] if hypervisor enabled
```

model/riscv_sys_regs.sail
```diff
@@ -152,7 +152,7 @@ function virtual_memory_supported() -> bool = {
 function lowest_supported_privLevel() -> Privilege =
   if currentlyEnabled(Ext_U) then User else Machine
 
-function have_privLevel(priv : priv_level) -> bool =
+function have_privLevel(priv : nom_priv_bits) -> bool =
```
pmundkur: This name is now a bit confusing, perhaps it should be 
```suggestion
function have_nominal_privLevel(priv : nom_priv_bits) -> bool =
```
to reflect the type?  It's called just in one place, so not a big change.

pmundkur (changes requested):
One minor rename might be good, but otherwise LGTM

model/riscv_insts_zicbom.sail
```diff
@@ -50,10 +50,13 @@ enum checked_cbop = {CBOP_ILLEGAL, CBOP_ILLEGAL_VIRTUAL, CBOP_INVAL_FLUSH, CBOP_
 // Select the cbop to perform based on the privilege.
 function cbop_priv_check(p: Privilege) -> checked_cbop = {
   let mCBIE : cbie = encdec_cbie(menvcfg[CBIE]);
+  // TODO: check henvcfg after hyppervisor is implemented
```
Timmmm:
```suggestion
  // TODO: check henvcfg after hypervisor is implemented
```

model/riscv_insts_base.sail
```diff
@@ -510,14 +510,18 @@ mapping clause encdec = ECALL()
...
 function clause execute ECALL() = {
-  let t : sync_exception =
-    struct { trap = match (cur_privilege) {
-                      User       => E_U_EnvCall(),
-                      Supervisor => E_S_EnvCall(),
-                      Machine    => E_M_EnvCall()
-                    },
-             excinfo = (None() : option(xlenbits)),
-             ext     = None() };
+  let trap : ExceptionType = match (cur_privilege) {
```
Timmmm:
```suggestion
  let trap : ExceptionType = match cur_privilege {
```

PR #1077, opened 2025-06-19
Invert signed bool for M extension

Timmmm:
Change `s` (signed) to `is_unsigned`. This removes the need for `maybe_not_u` and `bool_not_bits` and it's clearer.

Also add a `negate` that I missed.


pmundkur (approved):
Nice, this is simpler.

PR #1078, opened 2025-06-19
Fix a bug in `clint_dispatch()`.

pmundkur:
When `menvcfg.STCE` is unset, `mip.STIP` is only writable by M-mode software.   `menvcfg.STCE` needs to be set for `mip.STIP` to reflect the timer interrupt from `stimecmp`.

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -210,7 +210,7 @@ function clint_load(t, Physaddr(addr), width) = {
 
 function clint_dispatch() -> unit = {
   mip[MTI] = bool_to_bits(mtimecmp <=_u mtime);
-  if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then {
+  if currentlyEnabled(Ext_Sstc) then {
     mip[STI] = bool_to_bits(stimecmp <=_u mtime);
   };
   if get_config_print_platform()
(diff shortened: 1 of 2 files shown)


jordancarlin (commented):
I wonder if we should change `currentlyEnabled(Ext_Sstc)` to be dependent on `menvcfg[STCE]` instead. That might be more in line with the spec:

> When this extension is implemented and STCE in menvcfg is zero, an attempt to access stimecmp or vstimecmp in a mode other than M-mode raises an illegal-instruction exception, STCE in henvcfg is read-only zero, and STIP in mip and sip reverts to its defined behavior as if this extension is not implemented.

Then the writability of `menvcfg[STCE]` would depend on `hartSupports(Ext_Sstc)` and all of the other features added by the Sstc extension would be gated on `currentlyEnabled(Ext_Sstc)`.

Timmmm:
I dunno though because you can still access the CSRs from M mode. This change now breaks that...

Timmmm:
Also we don't do that for the other `menvcfg` "enable" flags. I think it's better to consider `currentlyEnabled()` as covering the *entire* extension. If you set `mstatus[FS]=0` it's as if the F extension doesn't exist even for M mode. `menvcfg[STCE]` only partially disables it.

pmundkur:
> I dunno though because you can still access the CSRs from M mode. This change now breaks that...

Good point.  I now prefer my initial minimal fix.

jordancarlin:
> > I dunno though because you can still access the CSRs from M mode. This change now breaks that...
> 
> Good point. I now prefer my initial minimal fix.

Good point. Agreed.

Timmmm (approved):

PR #1082, opened 2025-06-20
Improve typing of division and modulo

Timmmm:
This changes the definition of `%` and `/` to *only* work with positive numbers and also require the divisor to be non-zero.

The very few number of places that need signed division can call `quot_round_zero`/`rem_round_zero` directly. These don't require positive arguments, but they do still require a non-zero divisor.

Making this change required updating a load of other code which didn't have sufficiently accurate types, so it's a quite a large commit but not easy to split up.

### DIFF
--- a/model/riscv_insts_vext_fp_red.sail
+++ b/model/riscv_insts_vext_fp_red.sail
@@ -27,7 +27,7 @@ mapping clause encdec = RFVVTYPE(funct6, vm, vs2, vs1, vd)
   <-> encdec_rfvvfunct6(funct6) @ vm @ encdec_vreg(vs2) @ encdec_vreg(vs1) @ 0b001 @ encdec_vreg(vd) @ 0b1010111
   when currentlyEnabled(Ext_V)
 
-val process_rfvv_single: forall 'n 'm 'p, 'n > 0 & 'm in {8, 16, 32, 64}. (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, int('n), int('m), int('p)) -> ExecutionResult
+val process_rfvv_single : (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, nat1, sew_bitsize, range(-3, 3)) -> ExecutionResult
 function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_pow) = {
   let rm_3b = fcsr[FRM];
   let num_elem_vd = get_num_elem(0, SEW); /* vd regardless of LMUL setting */
@@ -70,7 +70,7 @@ function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_po
   RETIRE_SUCCESS
 }
 
-val process_rfvv_widen: forall 'n 'm 'p, 'n > 0 & 'm in {8, 16, 32, 64}. (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, int('n), int('m), int('p)) -> ExecutionResult
+val process_rfvv_widen : (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, nat1, sew_bitsize, range(-3, 3)) -> ExecutionResult
 function process_rfvv_widen(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_pow) = {
   let rm_3b          = fcsr[FRM];
   let SEW_widen      = SEW * 2;
--- a/model/riscv_insts_vext_utils.sail
+++ b/model/riscv_insts_vext_utils.sail
@@ -185,7 +185,7 @@ val get_velem_quad : forall 'n 'm 'p, 'n > 0 & 'm > 0 & 'p >= 0 & 4 * 'p + 3 < '
 function get_velem_quad(v, i) = v[4 * i + 3] @ v[4 * i + 2] @ v[4 * i + 1] @ v[4 * i]
 
 /* Divide the input bitvector into 4 equal slices and store them in vd starting at position 4*i */
-val write_velem_quad : forall 'p 'n 'm, 8 <= 'n <= 64 & 'm > 0 & 'p >= 0. (vregidx, int('n), bits('m), int('p)) -> unit
+val write_velem_quad : forall 'n, 'n > 0 . (vregidx, sew_bitsize, bits('n), nat) -> unit
 function write_velem_quad(vd, SEW, input, i) = {
   foreach(j from 0 to 3)
     write_single_element(SEW, 4 * i + j, vd, slice(input, j * SEW, SEW));
@@ -198,7 +198,7 @@ val get_velem_oct_vec : forall 'n 'm 'p, 'n > 0 & 8 <= 'm <= 64 & 'p >= 0 & 8 * 
 function get_velem_oct_vec(n, v, i) = [ v[8 * i + 7], v[8 * i + 6], v[8 * i + 5], v[8 * i + 4], v[8 * i + 3], v[8 * i + 2], v[8 * i + 1], v[8 * i] ]
 
 /* Writes each of the 8 elements from the input vector to the vector register vd, starting at position 8 * i */
-val write_velem_oct_vec : forall 'p 'n, 8 <= 'n <= 64 & 'p >= 0. (vregidx, int('n), vector(8, bits('n)), int('p)) -> unit
+val write_velem_oct_vec : forall 'n, 'n in {8, 16, 32, 64}. (vregidx, int('n), vector(8, bits('n)), nat) -> unit
 function write_velem_oct_vec(vd, SEW, input, i) = {
   foreach(j from 0 to 7)
     write_single_element(SEW, 8 * i + j, vd, input[j]);
@@ -209,7 +209,7 @@ val get_velem_quad_vec : forall 'n 'm 'p, 'n > 0 & 8 <= 'm <= 64  & 'p >= 0 & 4 
 function get_velem_quad_vec(v, i) = [ v[4 * i + 3], v[4 * i + 2], v[4 * i + 1], v[4 * i] ]
 
 /* Writes each of the 4 elements from the input vector to the vector register vd, starting at position 4 * i */
-val write_velem_quad_vec : forall 'p 'n, 8 <= 'n <= 64 & 'p >= 0. (vregidx, int('n), vector(4, bits('n)), int('p)) -> unit
+val write_velem_quad_vec : forall 'p 'n, 'n in {8, 16, 32, 64} & 'p >= 0. (vregidx, int('n), vector(4, bits('n)), int('p)) -> unit
 function write_velem_quad_vec(vd, SEW, input, i) = {
   foreach(j from 0 to 3)
     write_single_element(SEW, 4 * i + j, vd, input[j]);
@@ -243,8 +243,8 @@ function get_end_element() = unsigned(vl) - 1
  *   vector2 is a "mask" vector that is true for an element if the corresponding element
  *     in the result vector should be updated by the calling instruction
  */
-val init_masked_result : forall 'n 'm 'p, 'n >= 0 & 'm >= 0. (int('n), int('m), int('p), vector('n, bits('m)), bits('n)) -> result((vector('n, bits('m)), bits('n)), unit)
-function init_masked_result(num_elem, SEW, LMUL_pow, vd_val, vm_val) = {
+val init_masked_result : forall 'n 'm 'p, 'n >= 0 . (int('n), int('m), int('p), vector('n, bits('m)), bits('n)) -> result((vector('n, bits('m)), bits('n)), unit)
+function init_masked_result(num_elem, EEW, LMUL_pow, vd_val, vm_val) = {
   let start_element : nat = match get_start_element() {
     Ok(v)   => v,
     Err(()) => return Err(())
@@ -336,8 +336,8 @@ function init_masked_source(num_elem, LMUL_pow, vm_val) = {
 
 /* Mask handling for carry functions that use masks as input/output */
 /* Only prestart and tail elements are masked in a mask value */
-val init_masked_result_carry : forall 'n 'm 'p, 'n >= 0. (int('n), int('m), int('p), bits('n)) -> result((bits('n), bits('n)), unit)
-function init_masked_result_carry(num_elem, SEW, LMUL_pow, vd_val) = {
+val init_masked_result_carry : forall 'n, 'n >= 0 . (int('n), sew_bitsize, int, bits('n)) -> result((bits('n), bits('n)), unit)
+function init_masked_result_carry(num_elem, EEW, LMUL_pow, vd_val) = {
   let start_element : nat = match get_start_element() {
     Ok(v)   => v,
     Err(()) => return Err(())
@@ -375,8 +375,8 @@ function init_masked_result_carry(num_elem, SEW, LMUL_pow, vd_val) = {
 }
 
 /* Mask handling for cmp functions that use masks as output */
-val init_masked_result_cmp : forall 'n 'm 'p, 'n >= 0. (int('n), int('m), int('p), bits('n), bits('n)) -> result((bits('n), bits('n)), unit)
-function init_masked_result_cmp(num_elem, SEW, LMUL_pow, vd_val, vm_val) = {
+val init_masked_result_cmp : forall 'n, 'n >= 0 . (int('n), sew_bitsize, int, bits('n), bits('n)) -> result((bits('n), bits('n)), unit)
+function init_masked_result_cmp(num_elem, EEW, LMUL_pow, vd_val, vm_val) = {
   let start_element : nat = match get_start_element() {
     Ok(v)   => v,
     Err(()) => return Err(())
@@ -425,7 +425,7 @@ function init_masked_result_cmp(num_elem, SEW, LMUL_pow, vd_val, vm_val) = {
  *   Read multiple register groups and concatenate them in parallel
  *   The whole segments with the same element index are combined together
  */
-val read_vreg_seg : forall 'n 'm 'p 'q, 'n >= 0 & 'm >= 0 & nfields_range('q). (int('n), int('m), int('p), int('q), vregidx) -> vector('n, bits('q * 'm))
+val read_vreg_seg : forall 'n 'm 'p 'q, 'n >= 0 & 'm in {8, 16, 32, 64} & nfields_range('q). (int('n), int('m), int('p), int('q), vregidx) -> vector('n, bits('q * 'm))
 function read_vreg_seg(num_elem, SEW, LMUL_pow, nf, vrid) = {
   let LMUL_reg : int = if LMUL_pow <= 0 then 1 else 2 ^ LMUL_pow;
   var vreg_list : vector('q, vector('n, bits('m))) = vector_init(vector_init(zeros()));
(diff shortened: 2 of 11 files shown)


model/riscv_insts_vext_fp_red.sail
```diff
@@ -27,7 +27,7 @@ mapping clause encdec = RFVVTYPE(funct6, vm, vs2, vs1, vd)
   <-> encdec_rfvvfunct6(funct6) @ vm @ encdec_vreg(vs2) @ encdec_vreg(vs1) @ 0b001 @ encdec_vreg(vd) @ 0b1010111
   when currentlyEnabled(Ext_V)
 
-val process_rfvv_single: forall 'n 'm 'p, 'n > 0 & 'm in {8, 16, 32, 64}. (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, int('n), int('m), int('p)) -> ExecutionResult
+val process_rfvv_single: forall 'n 'p, 'n > 0 . (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, int('n), sew_bitsize, int('p)) -> ExecutionResult
```
jordancarlin: Might be worth changing `'p` to `'m` now that there are only two constraints.

jordancarlin: And I think the `int('n)` can be replaced with `nat1` so `'n` can be dropped entirely.

model/riscv_insts_vext_utils.sail
```diff
@@ -185,7 +185,7 @@ val get_velem_quad : forall 'n 'm 'p, 'n > 0 & 'm > 0 & 'p >= 0 & 4 * 'p + 3 < '
 function get_velem_quad(v, i) = v[4 * i + 3] @ v[4 * i + 2] @ v[4 * i + 1] @ v[4 * i]
 
 /* Divide the input bitvector into 4 equal slices and store them in vd starting at position 4*i */
-val write_velem_quad : forall 'p 'n 'm, 8 <= 'n <= 64 & 'm > 0 & 'p >= 0. (vregidx, int('n), bits('m), int('p)) -> unit
+val write_velem_quad : forall 'p 'n 'm, 'n in {8, 16, 32, 64} & 'm > 0 & 'p >= 0. (vregidx, int('n), bits('m), int('p)) -> unit
```
jordancarlin: Any reason not to use `sew_bitsize` instead of `int('n)` with `'n in {8, 16, 32, 64}`?

model/riscv_insts_vext_fp_red.sail
Timmmm: Yeah you're right, and we don't need `'p` either. Fixed.

model/riscv_insts_vext_utils.sail
Timmmm: Nope! I think I started with `{8...`, then added `sew_bitsize` and forgot to update this one I guess. Fixed. Also replaced `int('p)` with `nat`.

model/riscv_insts_vext_utils.sail
```diff
@@ -198,7 +198,7 @@ val get_velem_oct_vec : forall 'n 'm 'p, 'n > 0 & 8 <= 'm <= 64 & 'p >= 0 & 8 *
 function get_velem_oct_vec(n, v, i) = [ v[8 * i + 7], v[8 * i + 6], v[8 * i + 5], v[8 * i + 4], v[8 * i + 3], v[8 * i + 2], v[8 * i + 1], v[8 * i] ]
 
 /* Writes each of the 8 elements from the input vector to the vector register vd, starting at position 8 * i */
-val write_velem_oct_vec : forall 'p 'n, 8 <= 'n <= 64 & 'p >= 0. (vregidx, int('n), vector(8, bits('n)), int('p)) -> unit
+val write_velem_oct_vec : forall 'p 'n, 'n in {8, 16, 32, 64} & 'p >= 0. (vregidx, int('n), vector(8, bits('n)), int('p)) -> unit
```
jordancarlin:
```suggestion
val write_velem_oct_vec : forall 'n, 'n in {8, 16, 32, 64}. (vregidx, int('n), vector(8, bits('n)), nat) -> unit
```

Several other places in this file as well.

model/riscv_insts_vext_utils.sail
```diff
@@ -336,8 +336,8 @@ function init_masked_source(num_elem, LMUL_pow, vm_val) = {
 
 /* Mask handling for carry functions that use masks as input/output */
 /* Only prestart and tail elements are masked in a mask value */
-val init_masked_result_carry : forall 'n 'm 'p, 'n >= 0. (int('n), int('m), int('p), bits('n)) -> result((bits('n), bits('n)), unit)
-function init_masked_result_carry(num_elem, SEW, LMUL_pow, vd_val) = {
+val init_masked_result_carry : forall 'n 'm 'p, 'n >= 0 & 'm in {8, 16, 32, 64} . (int('n), int('m), int('p), bits('n)) -> result((bits('n), bits('n)), unit)
```
jordancarlin:
```suggestion
val init_masked_result_carry : forall 'n 'p, 'n >= 0 . (int('n), sew_bitsize, int('p), bits('n)) -> result((bits('n), bits('n)), unit)
```

Several other spots as well.

model/riscv_insts_vext_utils.sail
Timmmm: Ah oops. I fixed two of them. The others can't use `sew_bitsize` because they are used in the return type too.

There are *lots* of other ways the types can be simplified but I don't think we should do any more in this PR.

jordancarlin: Agreed. Just figured we should simplify the spots that were already being updated in this PR.

Timmmm:
@Alasdair something slightly weird here - the C model builds but the `generated_docs_rv64d` target gives this error:

```
Type error:
riscv_insts_vext_utils.sail:191.45-71:
191 |    write_single_element(SEW, 4 * i + j, vd, slice(input, j * SEW, SEW));
    |                                             ^------------------------^ checking function argument has type bitvector('ex21557#)
    | Could not resolve quantifiers for slice
    | * (0 <= 'n & 0 <= 'ex21557#)
```

Surely the type checking for all backends is the same?

It's something to do with the condition on `'n`

```
val write_velem_quad : forall 'n . (vregidx, sew_bitsize, bits('n), nat) -> unit
function write_velem_quad(vd, SEW, input, i) = {
  foreach(j from 0 to 3)
    write_single_element(SEW, 4 * i + j, vd, slice(input, j * SEW, SEW));
}
```

With no condition it only works in the C model. With `'n >= 0` or `'n > 0` it works in both cases.

I do vaguely recall that you added something that implicitly adds a `'n>0` constraint for `bits('n)` - maybe that was somehow only done for the C backend? Also deja vu... Sorry if I've asked this exact question before!

model/riscv_insts_vext_fp_red.sail
```diff
@@ -70,7 +70,7 @@ function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_po
   RETIRE_SUCCESS
 }
 
-val process_rfvv_widen: forall 'n 'm 'p, 'n > 0 & 'm in {8, 16, 32, 64}. (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, int('n), int('m), int('p)) -> ExecutionResult
+val process_rfvv_widen : (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, nat1, sew_bitsize, range(-3, 3)) -> ExecutionResult
```
pmundkur, line 73: Building on the `sew_bitsize` improvement, it might be also be worth defining an `lmul_pow` type as `range(-3,3)` and using it here.
```suggestion
val process_rfvv_widen : (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, nat1, sew_bitsize, lmul_pow) -> ExecutionResult
```

model/riscv_insts_vext_utils.sail
```diff
@@ -209,7 +209,7 @@ val get_velem_quad_vec : forall 'n 'm 'p, 'n > 0 & 8 <= 'm <= 64  & 'p >= 0 & 4
 function get_velem_quad_vec(v, i) = [ v[4 * i + 3], v[4 * i + 2], v[4 * i + 1], v[4 * i] ]
 
 /* Writes each of the 4 elements from the input vector to the vector register vd, starting at position 4 * i */
-val write_velem_quad_vec : forall 'p 'n, 8 <= 'n <= 64 & 'p >= 0. (vregidx, int('n), vector(4, bits('n)), int('p)) -> unit
+val write_velem_quad_vec : forall 'p 'n, 'n in {8, 16, 32, 64} & 'p >= 0. (vregidx, int('n), vector(4, bits('n)), int('p)) -> unit
```
pmundkur: Perhaps define `type sew_bitsize_range('n) = 'n in {8, 16, 32, 64}` and use it here (and use `nat` for `int('p)`)

```suggestion
val write_velem_quad_vec : forall 'p 'n, sew_bitsize_range('n) & 'p >= 0. (vregidx, int('n), vector(4, bits('n)), int('p)) -> unit
```

Even though it's a bit longer, I find `sew_bitsize_range('n)` more meaningful than a string of numbers.

Also, there is a possibility to similarly improve `get_velem_{quad,oct}_vec` above in this file.

model/riscv_insts_vext_utils.sail
```diff
@@ -198,7 +198,7 @@ val get_velem_oct_vec : forall 'n 'm 'p, 'n > 0 & 8 <= 'm <= 64 & 'p >= 0 & 8 *
 function get_velem_oct_vec(n, v, i) = [ v[8 * i + 7], v[8 * i + 6], v[8 * i + 5], v[8 * i + 4], v[8 * i + 3], v[8 * i + 2], v[8 * i + 1], v[8 * i] ]
 
 /* Writes each of the 8 elements from the input vector to the vector register vd, starting at position 8 * i */
-val write_velem_oct_vec : forall 'p 'n, 8 <= 'n <= 64 & 'p >= 0. (vregidx, int('n), vector(8, bits('n)), int('p)) -> unit
+val write_velem_oct_vec : forall 'n, 'n in {8, 16, 32, 64}. (vregidx, int('n), vector(8, bits('n)), nat) -> unit
```
pmundkur: See below suggestion of `sew_bitsize_range('n)`.

model/riscv_insts_vext_utils.sail
```diff
@@ -425,7 +425,7 @@ function init_masked_result_cmp(num_elem, SEW, LMUL_pow, vd_val, vm_val) = {
  *   Read multiple register groups and concatenate them in parallel
  *   The whole segments with the same element index are combined together
  */
-val read_vreg_seg : forall 'n 'm 'p 'q, 'n >= 0 & 'm >= 0 & nfields_range('q). (int('n), int('m), int('p), int('q), vregidx) -> vector('n, bits('q * 'm))
+val read_vreg_seg : forall 'n 'm 'p 'q, 'n >= 0 & 'm in {8, 16, 32, 64} & nfields_range('q). (int('n), int('m), int('p), int('q), vregidx) -> vector('n, bits('q * 'm))
```
pmundkur:
```suggestion
val read_vreg_seg : forall 'n 'm 'p 'q, 'n >= 0 & sew_bitsize_range('m) & nfields_range('q). (int('n), int('m), int('p), int('q), vregidx) -> vector('n, bits('q * 'm))
```
with above definition

model/riscv_sys_regs.sail
```diff
@@ -941,7 +944,7 @@ function get_sew_bytes() -> {1, 2, 4, 8} =
 
 /* the vector register group multiplier (LMUL) */
 /* this returns the power of 2 for LMUL */
-val get_lmul_pow : unit -> {-3, -2, -1, 0, 1, 2, 3}
+val get_lmul_pow : unit -> range(-3, 3)
```
pmundkur: See above on `type LMUL_pow = range(-3,3)` and then using `LMUL_pow` where needed.

model/riscv_vext_regs.sail
```diff
@@ -237,23 +237,22 @@ function ext_write_vcsr (vxrm_val, vxsat_val) = {
...
    * be handled in init_masked_result */
-  let num_elem = 2 ^ LMUL_pow_reg * VLEN / SEW;
+  let num_elem = (2 ^ LMUL_pow_reg) * VLEN / SEW;
   assert(num_elem > 0);
   num_elem
 }
 
 /* Reads a single vreg into multiple elements */
-val read_single_vreg : forall 'n 'm, 'n >= 0 & 'm >= 0. (int('n), int('m), vregidx) -> vector('n, bits('m))
+val read_single_vreg : forall 'n 'm, 'n >= 0 & 'm in {8, 16, 32, 64} . (int('n), int('m), vregidx) -> vector('n, bits('m))
```
pmundkur:
```suggestion
val read_single_vreg : forall 'n 'm, 'n >= 0 & sew_bitsize_range('m). (int('n), int('m), vregidx) -> vector('n, bits('m))
```

model/riscv_vext_regs.sail
```diff
@@ -263,11 +262,10 @@ function read_single_vreg(num_elem, SEW, vrid) = {
 }
 
 /* Writes multiple elements into a single vreg */
-val write_single_vreg : forall 'n 'm, 'n >= 0. (int('n), int('m), vregidx, vector('n, bits('m))) -> unit
+val write_single_vreg : forall 'n 'm, 'n >= 0 & 'm in {8, 16, 32, 64} . (int('n), int('m), vregidx, vector('n, bits('m))) -> unit
```
pmundkur:
```suggestion
val write_single_vreg : forall 'n 'm, 'n >= 0 & sew_bitsize_range('m) . (int('n), int('m), vregidx, vector('n, bits('m))) -> unit
```

model/riscv_vext_regs.sail
```diff
@@ -277,7 +275,7 @@ function write_single_vreg(num_elem, SEW, vrid, v) = {
 }
 
 /* The general vreg reading operation with num_elem as max(VLMAX,VLEN/SEW)) */
-val read_vreg : forall 'n 'm 'p, 'n >= 0 & 'm >= 0. (int('n), int('m), int('p), vregidx) -> vector('n, bits('m))
+val read_vreg : forall 'n 'm 'p, 'n >= 0 & 'm in {8, 16, 32, 64} . (int('n), int('m), int('p), vregidx) -> vector('n, bits('m))
```
pmundkur:
```suggestion
val read_vreg : forall 'n 'm 'p, 'n >= 0 & sew_bitsize_range('m). (int('n), int('m), int('p), vregidx) -> vector('n, bits('m))
```

model/riscv_insts_vext_fp_red.sail
```diff
@@ -27,7 +27,7 @@ mapping clause encdec = RFVVTYPE(funct6, vm, vs2, vs1, vd)
   <-> encdec_rfvvfunct6(funct6) @ vm @ encdec_vreg(vs2) @ encdec_vreg(vs1) @ 0b001 @ encdec_vreg(vd) @ 0b1010111
   when currentlyEnabled(Ext_V)
 
-val process_rfvv_single: forall 'n 'm 'p, 'n > 0 & 'm in {8, 16, 32, 64}. (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, int('n), int('m), int('p)) -> ExecutionResult
+val process_rfvv_single : (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, nat1, sew_bitsize, range(-3, 3)) -> ExecutionResult
```
pmundkur, line 30: Could just define `type LMUL_pow = range(-3,3)` (as I did in #1051) and use it here.

model/riscv_vext_regs.sail
```diff
@@ -315,11 +313,11 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
 }
 
 /* Single element reading operation */
-val read_single_element : forall 'm 'x, 8 <= 'm <= 128. (int('m), int('x), vregidx) -> bits('m)
+val read_single_element : forall 'm, 'm in {8, 16, 32, 64} . (int('m), nat, vregidx) -> bits('m)
```
pmundkur:
```suggestion
val read_single_element : forall 'm, sew_bitsize_range('m). (int('m), nat, vregidx) -> bits('m)
```

model/riscv_vext_regs.sail
```diff
@@ -328,7 +326,7 @@ function read_single_element(EEW, index, vrid) = {
 }
 
 /* The general vreg writing operation with num_elem as max(VLMAX,VLEN/SEW)) */
-val write_vreg : forall 'n 'm 'p, 'n >= 0 & 'm >= 0. (int('n), int('m), int('p), vregidx, vector('n, bits('m))) -> unit
+val write_vreg : forall 'n 'm 'p, 'n >= 0 & 'm in {8, 16, 32, 64} . (int('n), int('m), int('p), vregidx, vector('n, bits('m))) -> unit
```
pmundkur:
```suggestion
val write_vreg : forall 'n 'm 'p, 'n >= 0 & sew_bitsize_range('m). (int('n), int('m), int('p), vregidx, vector('n, bits('m))) -> unit
```

model/riscv_vext_regs.sail
```diff
@@ -350,10 +348,10 @@ function write_vreg(num_elem, SEW, LMUL_pow, vrid, vec) = {
 }
 
 /* Single element writing operation */
-val write_single_element : forall 'm 'x, 8 <= 'm <= 128. (int('m), int('x), vregidx, bits('m)) -> unit
+val write_single_element : forall 'm, 'm in {8, 16, 32, 64}. (int('m), nat, vregidx, bits('m)) -> unit
```
pmundkur:
```suggestion
val write_single_element : forall 'm, sew_bitsize_range('m). (int('m), nat, vregidx, bits('m)) -> unit
```

Alasdair:
> @Alasdair something slightly weird here - the C model builds but the `generated_docs_rv64d` target gives this error:
> 
> ```
> Type error:
> riscv_insts_vext_utils.sail:191.45-71:
> 191 |    write_single_element(SEW, 4 * i + j, vd, slice(input, j * SEW, SEW));
>     |                                             ^------------------------^ checking function argument has type bitvector('ex21557#)
>     | Could not resolve quantifiers for slice
>     | * (0 <= 'n & 0 <= 'ex21557#)
> ```
> 
> Surely the type checking for all backends is the same?

If they don't use the same combination of `--strict-XYZ` flags that could cause this, otherwise it should be the same.

Timmmm:
> If they don't use the same combination of --strict-XYZ flags that could cause this

Ah that was indeed the issue! I'll make a separate PR to fix that.

@pmundkur all good suggestions but do you mind if we do them in a future PR? I deliberately avoided fixing a lot of these (e.g. `{-3, ... 3}` -> `range(-3, 3)`) because this will balloon into a 5000 line PR if I don't. There's a ton of vector code that is not very well typed and very often if you improve one thing it causes a chain reaction and you end up having to rewrite the whole thing.

pmundkur:
Sounds good about deferring the `range(-3,3)` until later.  But could we at least use `sew_bitsize_range('n)` instead of `'m in {8, 16, 32, 64}`?

Timmmm:
> Sounds good about deferring the range(-3,3) until later. But could we at least use sew_bitsize_range('n) instead of 'm in {8, 16, 32, 64}?

Sure, done. I called it `is_sew_bitsize` though since it is a predicate.

pmundkur:
Oops, seems to have a minor conflict with #1051, easily fixed.

PR #1089, opened 2025-06-23
Add mseccfg CSR

Timmmm:
This is required if you support Zkr.

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -305,6 +305,42 @@ function in32BitMode() -> bool = {
   cur_architecture() == RV32
 }
 
+bitfield Seccfg : bits(64) = {
+  // Enable access to SEED in S-mode
+  SSEED  : 9,
+  // Enable access to SEED in U-mode
+  USEED  : 8,
+}
+
+function legalize_mseccfg(o : Seccfg, v : bits(64)) -> Seccfg = {
+  let v = Mk_Seccfg(v);
+  [o with
+    SSEED = if currentlyEnabled(Ext_Zkr) then v[SSEED] else 0b0,
+    USEED = if currentlyEnabled(Ext_Zkr) then v[USEED] else 0b0,
+  ]
+}
+
+register mseccfg : Seccfg = legalize_mseccfg(Mk_Seccfg(zeros()), zeros())
+
+mapping clause csr_name_map = 0x747  <-> "mseccfg"
+mapping clause csr_name_map = 0x757  <-> "mseccfgh"
+
+// "mseccfg exists if Zkr is implemented, or if it is required by other processor features."
+function clause is_CSR_defined(0x747) = currentlyEnabled(Ext_Zkr) // mseccfg
+function clause is_CSR_defined(0x757) = currentlyEnabled(Ext_Zkr) & xlen == 32 // mseccfgh
+
+function clause read_CSR(0x747) = mseccfg.bits[xlen - 1 .. 0]
+function clause read_CSR(0x757 if xlen == 32) = mseccfg.bits[63 .. 32]
+
+function set_mseccfg(v : bits(64)) -> unit = {
+  mseccfg = legalize_mseccfg(mseccfg, v);
+}
+
+function clause write_CSR((0x747, value) if xlen == 32) = { set_mseccfg(mseccfg.bits[63 .. 32] @ value); mseccfg.bits[31 .. 0] }
+function clause write_CSR((0x747, value) if xlen == 64) = { set_mseccfg(value); mseccfg.bits }
+function clause write_CSR((0x757, value) if xlen == 32) = { set_mseccfg(value @ mseccfg.bits[31 .. 0]); mseccfg.bits[63 .. 32] }
+
+
 // envcfg Resisters
 bitfield MEnvcfg : bits(64) = {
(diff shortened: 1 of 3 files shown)


radimkrcmar:
I assume Sail wants to model all RISC-V configurations, where some might have existed even before mseccfg was defined, so we want a way to keep the CSR reserved.

mseccfg must be present/implemented (the CSR access doesn't explode) if any extension furnishes the CSR fields.
I think we could add a configuration option that implements mseccfg (as read-only 0) even without any architectural extension using it.

Timmmm:
> mseccfg must be present/implemented (the CSR access doesn't explode) if any extension furnishes the CSR fields.

Where does it say this in the spec though?

> I assume Sail wants to model all RISC-V configurations, where some might have existed even before mseccfg was defined

Yes we'll have to make it configurable in future. I'd prefer to do that in a future PR though (I put a TODO in the code).

nadime15:
You probably saw this but the #969 defines the same reg and makes use of it. Does this resolve itself once PMM is added?

I wonder about the use of `set_mseccfg`, is there a specific reason to define that instead of just calling `legalize_mseccfg` directly?

radimkrcmar:
> > mseccfg must be present/implemented (the CSR access doesn't explode) if any extension furnishes the CSR fields.
> 
> Where does it say this in the spec though?

Unprivileged architecture, 31.5.3. Access Control to seed:
`mseccfg exists if Zkr is implemented, or if it is required by other processor features.`

Timmmm:
Ah good find. I updated the condition.

model/riscv_sys_regs.sail
```diff
@@ -305,6 +305,43 @@ function in32BitMode() -> bool = {
...
+  // Enable access to SEED in S-mode
+  SSEED  : 9,
+  // Enable access to SEED in U-mode
+  USEED  : 8,
+}
+
+function legalize_mseccfg(o : Seccfg, v : bits(64)) -> Seccfg = {
+  let v = Mk_Seccfg(v);
+  [
+    o with
```
pmundkur:
```suggestion
  [o with
```
is the style in the rest of the model

model/riscv_sys_regs.sail
```diff
@@ -305,6 +305,43 @@ function in32BitMode() -> bool = {
...
+function clause is_CSR_defined(0x757) = currentlyEnabled(Ext_Zkr) & xlen == 32 // mseccfgh
+
+function clause read_CSR(0x747) = mseccfg.bits[xlen - 1 .. 0]
+function clause read_CSR(0x757 if xlen == 32) = mseccfg.bits[63 .. 32]
+
+function set_mseccfg(v : bits(64)) -> unit = {
+  mseccfg = legalize_mseccfg(mseccfg, v);
+}
+
+function clause write_CSR((0x747, value) if xlen == 32) = { set_mseccfg(mseccfg.bits[63 .. 32] @ value); mseccfg.bits[31 .. 0] }
```
pmundkur: Other `write_CSR` clauses inline the `legalize_` call.  Perhaps we could do the same here for consistency?  It removes the need for a separate `set_` function.

pmundkur (changes requested):
LGTM, other than minor stylistic consistency fixes

model/riscv_sys_regs.sail
```diff
@@ -305,6 +305,43 @@ function in32BitMode() -> bool = {
...
+  SSEED  : 9,
+  // Enable access to SEED in U-mode
+  USEED  : 8,
+}
+
+function legalize_mseccfg(o : Seccfg, v : bits(64)) -> Seccfg = {
+  let v = Mk_Seccfg(v);
+  [
+    o with
+    SSEED = if currentlyEnabled(Ext_Zkr) then v[SSEED] else 0b0,
```
jordancarlin: > If S or U mode is not implemented, then the corresponding [s,u]seed bits of mseccfg must be hardwired to zero.

model/riscv_sys_regs.sail
```diff
@@ -305,6 +305,43 @@ function in32BitMode() -> bool = {
...
+    USEED = if currentlyEnabled(Ext_Zkr) then v[USEED] else 0b0,
```
jordancarlin: Probably need a config option for this: 
> Implementations may implement mseccfg such that [s,u]seed is a read-only constant value 0.

jordancarlin (changes requested):

model/riscv_sys_regs.sail
pmundkur: Good catch. Also, it needs to be added to reset
```
The [s,u]seed bits must have a defined reset value. The system must not allow them to be in an
undefined state after a reset.
```

model/riscv_sys_regs.sail
Timmmm: Done

model/riscv_sys_regs.sail
Timmmm: Good spot, added.

model/riscv_sys_regs.sail
Timmmm: Great spots! I added config values for read-only-zero and the reset values.

jordancarlin (approved):
LGTM now

pmundkur (approved):

PR #1090, opened 2025-06-23
Fix vror.vi immediate encoding

sizn-sc:
The encoding of `vror.vi` wasn't entirely correct, as it lacked the top bit of the 6 bit immediate.
This would lead to illegal instruction exceptions for immediates in range [32,64).

### DIFF
--- a/model/riscv_insts_zvbb.sail
+++ b/model/riscv_insts_zvbb.sail
@@ -466,14 +466,14 @@ function clause execute (VROR_VX(vm, vs2, rs1, vd)) = {
   RETIRE_SUCCESS
 }
 
-union clause ast = VROR_VI : (bits(1), vregidx, bits(5), vregidx)
+union clause ast = VROR_VI : (bits(1), vregidx, bits(6), vregidx)
 
-mapping clause encdec = VROR_VI(vm, vs2, uimm, vd)
-  <-> 0b010100 @ vm @ encdec_vreg(vs2) @ uimm @ 0b011 @ encdec_vreg(vd) @ 0b1010111
+mapping clause encdec = VROR_VI(vm, vs2, uimm5 @ uimm40, vd)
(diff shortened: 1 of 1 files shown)


nadime15 (approved):
Good catch, thank you!

sizn-sc:
For reference, this issue was found with the help of [llvm-snippy](https://github.com/syntacore/snippy/tree/main/llvm/tools/llvm-snippy), which is an LLVM-based random instruction sequence generator and supports most standard RISC-V extensions. The configuration for a Zkbb test containing VROR_VI instruction can be found here: https://github.com/syntacore/snippy/blob/b845839046c6d431eae31473de526aa16ca50fd6/llvm/test/tools/llvm-snippy/rvv-zvbb.yaml#L46

Timmmm (approved):

model/riscv_insts_zvbb.sail
```diff
@@ -466,14 +466,14 @@ function clause execute (VROR_VX(vm, vs2, rs1, vd)) = {
   RETIRE_SUCCESS
 }
 
-union clause ast = VROR_VI : (bits(1), vregidx, bits(5), vregidx)
+union clause ast = VROR_VI : (bits(1), vregidx, bits(6), vregidx)
 
-mapping clause encdec = VROR_VI(vm, vs2, uimm, vd)
-  <-> 0b010100 @ vm @ encdec_vreg(vs2) @ uimm @ 0b011 @ encdec_vreg(vd) @ 0b1010111
+mapping clause encdec = VROR_VI(vm, vs2, i5 @ uimm5, vd)
```
jordancarlin: This is more consistent with how we’ve named split immediate fields for other instructions.
```suggestion
mapping clause encdec = VROR_VI(vm, vs2, uimm5 @ uimm40, vd)
```

sizn-sc: Thanks, addressed.

jordancarlin (approved):

PR #1091, opened 2025-06-23
Move the functions guarding `mcycle` and `minstret` increments to `riscv_platform`.

### LINKED ISSUE #783: Fix dependency on `should_inc_minstret` in the stepper.
`should_inc_minstret` is defined in `riscv_smcntrpmf.sail` i.e. a file for an extension, but used in `riscv_step.sail`.   

This is unfortunate: core functionality should ideally not depend on functions defined only in extensions.  

Similarly for `should_inc_mcycle` and `riscv_platform.sail`.

Discovered while rebasing #755.

pmundkur:
These will be affected by the Sdext extension and hence need to belong to a common file.

This also fixes #783.

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -270,6 +270,14 @@ function clint_store(Physaddr(addr), width, data) = {
   }
 }
 
+/* Counters and timers are affected by Smcntrpmf (and the forthcoming Sdext extension). */
+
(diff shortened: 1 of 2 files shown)


pmundkur:
This is part of pulling out pieces of #954 to reduce the size of that PR.

jordancarlin (approved):

PR #1092, opened 2025-06-23
Move the reservation handlers to a separate file.

pmundkur:
They will need to be accessible to the Sdext extension, part of which needs to be sequenced before `riscv_sys_control`.


pmundkur:
This is part of pulling out pieces of https://github.com/riscv/sail-riscv/pull/954 to reduce the size of that PR.

model/CMakeLists.txt
```diff
@@ -130,6 +130,7 @@ foreach (xlen IN ITEMS 32 64)
                 "riscv_fdext_regs.sail"
                 "riscv_fdext_control.sail"
                 "riscv_smcntrpmf.sail"
+                "riscv_sys_reservations.sail"
```
Timmmm: Technically `reservation` because there can only ever be one.

PR #1093, opened 2025-06-23
Make `try_step` return whether the hart is waiting, instead of whether it stepped.

pmundkur:
For Sdext, `try_step` will also need to return whether the hart entered debug mode (i.e. halted) or experienced an error in debug mode.  The `stepped` case will then hold if nothing unusual happened (e.g. is not waiting and not in debug mode).

This is a step towards the API for Sdext.


pmundkur:
This is part of pulling out pieces of https://github.com/riscv/sail-riscv/pull/954 to reduce the size of that PR.

Issue #993, opened 2025-05-30
Configuration tooling


pmundkur:
The generated device tree for the Sail model (from #972) is different from one generated by Spike, which uses custom `cpu` properties like `riscv,pmpregions` and `riscv,pmpgranularity`.  Spike also supports different devices.  There is a need for tooling to convert device trees from Spike and other sources into a suitable valid configuration (#860) for the Sail model (and hence address the issue underlying #2).

PR #1096, opened 2025-06-24
Auto-enable Vector Crypto superset extensions (Zvkn, Zvknc, Zvkng, Zvks, Zvksc, Zvksg)

nadime15:
Add remaining vector crypto extensions and automatically enable superset extensions when all of the appropriate subset extensions are enabled.

### DIFF
--- a/model/riscv_extensions.sail
+++ b/model/riscv_extensions.sail
@@ -287,41 +287,49 @@ enum clause extension = Ext_Zvkt
 mapping clause extensionName = Ext_Zvkt <-> "zvkt"
 function clause hartSupports(Ext_Zvkt) = config extensions.Zvkt.supported
 function clause currentlyEnabled(Ext_Zvkt) = hartSupports(Ext_Zvkt)
-// The following extensions are superset/shorthand extensions that cannot be directly
-// configured in the config file and are not part of the device-tree. These extensions
-// are automatically enabled when all their required component extensions are enabled.
-// This approach simplifies checking for support of subset extensions
+// The following extensions are superset/shorthand extensions. They cannot be directly
+// configured in the config file, aren't part of the device-tree and shouldn't be
+// added to the extension list. These extensions are automatically enabled when all
+// their required subset extensions are enabled.
 enum clause extension = Ext_Zvkn
 mapping clause extensionName = Ext_Zvkn <-> "zvkn"
+function clause hartSupports(Ext_Zvkn) = true
 function clause currentlyEnabled(Ext_Zvkn) =
-currentlyEnabled(Ext_Zvkned) &
-currentlyEnabled(Ext_Zvknhb) &
-currentlyEnabled(Ext_Zvkb)   &
-currentlyEnabled(Ext_Zvkt)
+hartSupports(Ext_Zvkn)   &
+hartSupports(Ext_Zvkned) &
+hartSupports(Ext_Zvknhb) &
+hartSupports(Ext_Zvkb)   &
+hartSupports(Ext_Zvkt)
 // NIST Algorithm Suite with carryless multiply
 enum clause extension = Ext_Zvknc
 mapping clause extensionName = Ext_Zvknc <-> "zvknc"
-function clause currentlyEnabled(Ext_Zvknc) = currentlyEnabled(Ext_Zvkn) & currentlyEnabled(Ext_Zvbc)
+function clause hartSupports(Ext_Zvknc) = true
+function clause currentlyEnabled(Ext_Zvknc) = hartSupports(Ext_Zvknc) & currentlyEnabled(Ext_Zvkn) & hartSupports(Ext_Zvbc)
 // NIST Algorithm Suite with GCM
 enum clause extension = Ext_Zvkng
 mapping clause extensionName = Ext_Zvkng <-> "zvkng"
-function clause currentlyEnabled(Ext_Zvkng) = currentlyEnabled(Ext_Zvkn) & currentlyEnabled(Ext_Zvkg)
+function clause hartSupports(Ext_Zvkng) = true
+function clause currentlyEnabled(Ext_Zvkng) = hartSupports(Ext_Zvkng) & currentlyEnabled(Ext_Zvkn) & hartSupports(Ext_Zvkg)
 // ShangMi Algorithm Suite
 enum clause extension = Ext_Zvks
 mapping clause extensionName = Ext_Zvks <-> "zvks"
+function clause hartSupports(Ext_Zvks) = true
 function clause currentlyEnabled(Ext_Zvks) =
-currentlyEnabled(Ext_Zvksed) &
-currentlyEnabled(Ext_Zvksh)  &
-currentlyEnabled(Ext_Zvkb)   &
-currentlyEnabled(Ext_Zvkt)
+hartSupports(Ext_Zvks)   &
+hartSupports(Ext_Zvksed) &
+hartSupports(Ext_Zvksh)  &
+hartSupports(Ext_Zvkb)   &
+hartSupports(Ext_Zvkt)
 // ShangMi Algorithm Suite with carryless multiplication
 enum clause extension = Ext_Zvksc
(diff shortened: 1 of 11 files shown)


nadime15:
Does somebody know how things actually work with the device tree and shorthands like these:

[Vector Crypto Extensions](https://github.com/riscv/sail-riscv/wiki/Extension-Roadmap#vector-crypto-extensions)
| Extension | Description                                     |
|-----------|-------------------------------------------------|
| Zvkn      | Zvkned + Zvknhb + Zvkb + Zvkt                   |
| Zvknc     | Zvkn + Zvbc                                     |
| Zvkng     | Zvkn + Zvkg                                     |
| Zvks      | Zvksed + Zvksh + Zvkb + Zvkt                    |
| Zvksc     | Zvks + Zvbc                                     |
| Zvksg     | Zvks + Zvkg                                     |

So let’s say I enable Zvknc, which pulls in Zvkn + Zvbc, and those pull in Zvkned + Zvknhb + Zvkb + Zvkt. Should the final device tree only print `Zvknc` or `Zvbc_Zvkn_Zvknc` or even `Zvbc_Zvkb_Zvkn_Zvknc_Zvkned_Zvknhb_Zvkt`. 

As far as I can see now its actually printing `Zvbc_Zvkb_Zvkn_Zvknc_Zvkned_Zvknhb_Zvkt` but the desired behaviour (?) is probably only to print Zvknc.

@pmundkur

Timmmm:
Yeah I think we can print the minimum necessary if the others are implied. You don't have to always list all the sub-extensions.

jordancarlin:
I think the simplest approach is to not include these superset extensions in the config file and instead just have the smallest granularity extensions. These should still exist in the extensions enum and would get turned on automatically if all of the appropriate subset extensions are enabled.

In terms of what to print in the device tree/isa string, it might be the least ambiguous to just print the full list even if that is more verbose. I know some tools (like riscv-config) don't always handle the superset extensions well.

pmundkur:
Yes, I think for now we should stick to the smallest granularity extensions in the model and config file.  We could rely on other tooling (such as that in #993) to convert superset extensions into Sail model configurations.

Timmmm:
But note that in *some* cases it isn't as simple as extension 1 = extensions 2 + 3. IIRC think supporting B or C is different to supporting the extensions that they cover. In this case I believe it is though.

model/riscv_insts_zvkg.sail
```diff
@@ -6,7 +6,7 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_Zvkg) = hartSupports(Ext_Zvkg) & currentlyEnabled(Ext_V)
+function clause currentlyEnabled(Ext_Zvkg) = (hartSupports(Ext_Zvkg) | currentlyEnabled(Ext_Zvkng) | currentlyEnabled(Ext_Zvksg)) & currentlyEnabled(Ext_V)
```
jordancarlin: The superset extensions will only ever be enabled if the subset extensions are supported, so no need to include them here.

model/riscv_extensions.sail
```diff
@@ -285,6 +285,49 @@ enum clause extension = Ext_Zvkt
...
 function clause hartSupports(Ext_Zvkt) = config extensions.Zvkt.supported
 function clause currentlyEnabled(Ext_Zvkt) = hartSupports(Ext_Zvkt)
+// The following extensions are superset/shorthand extensions. They cannot be directly
+// configured in the config file, aren't part of the device-tree and shouldn't be
+// added to the extension list. These extensions are automatically enabled when all
+// their required subset extensions are enabled.
+enum clause extension = Ext_Zvkn
+mapping clause extensionName = Ext_Zvkn <-> "zvkn"
+function clause hartSupports(Ext_Zvkn) = true
+function clause currentlyEnabled(Ext_Zvkn) =
```
jordancarlin: Seems like `hartSupports` should depend on the subset extensions and `currentlyEnabled`can just depend on if the particular superset extension is supported.

model/riscv_extensions.sail
```diff
@@ -285,6 +285,49 @@ enum clause extension = Ext_Zvkt
 mapping clause extensionName = Ext_Zvkt <-> "zvkt"
 function clause hartSupports(Ext_Zvkt) = config extensions.Zvkt.supported
 function clause currentlyEnabled(Ext_Zvkt) = hartSupports(Ext_Zvkt)
+// The following extensions are superset/shorthand extensions. They cannot be directly
+// configured in the config file, aren't part of the device-tree and shouldn't be
```
jordancarlin: I’d be in favor of including them in the device tree/ISA string even if it is somewhat redundant. I think the Sail output for those should be exhaustive since there is limited downside to including them, and some tools might naively check for just the superset extension if they require all of the constituent extensions, for example.

jordancarlin (changes requested):

model/riscv_extensions.sail
nadime15: Should be addressed now!

model/riscv_extensions.sail
nadime15: Hmm yeah I added them back

jordancarlin:
`hartSupports_measure` in `riscv_termination.sail` needs to be updated with the dependency order to fix the Lean build.

model/riscv_extensions.sail
```diff
@@ -285,6 +285,46 @@ enum clause extension = Ext_Zvkt
...
+function clause hartSupports(Ext_Zvkng) = hartSupports(Ext_Zvkn) & hartSupports(Ext_Zvkg)
+function clause currentlyEnabled(Ext_Zvkng) = hartSupports(Ext_Zvkng)
+// ShangMi Algorithm Suite
+enum clause extension = Ext_Zvks
+mapping clause extensionName = Ext_Zvks <-> "zvks"
+function clause hartSupports(Ext_Zvks) =
+hartSupports(Ext_Zvksed) &
+hartSupports(Ext_Zvksh)  &
+hartSupports(Ext_Zvkb)   &
+hartSupports(Ext_Zvkt)
```
Timmmm: I think you dropped some of these: `      ` :-D

pmundkur:
I'm rethinking the choice to not expose superset extensions in the config file.  It necessitates extra tooling (and a degree of user-unfriendliness); is there a way of handling this without extra tooling?  

This would mean some extensions may not appear in the config file, e.g. if `Zvknc` is marked as supported, `Zvkn` and `Zvbc` may not appear since they would be redundant.  This would require some additional flexibility in the configuration mechanism, e.g. a check for whether a configuration option appears before trying to read its value.

model/riscv_termination.sail
```diff
@@ -91,8 +91,22 @@ termination_measure virtual_memory_supported(_) = 3
...
   match ext {
-    Ext_C => 2,
+    Ext_Zvkned => 0,
+    Ext_Zvknhb => 0,
+    Ext_Zvkb => 0,
+    Ext_Zvkt => 0,
+    Ext_Zvbc => 0,
+    Ext_Zvkg => 0,
+    Ext_Zvksed => 0,
+    Ext_Zvksh => 0,
```
jordancarlin: No need to explicitly include the extensions with a value of 0. That’s the default, and calling them out makes it seem like there is something different going on with them.

jordancarlin (changes requested):

model/riscv_termination.sail
nadime15: Good point, thanks!

jordancarlin (approved):
LGTM based on the decisions made in Monday's meeting.

Timmmm:
Isn't this going to put the superset extensions *and* the subset ones in the ISA string? That seems slightly odd... Or is that intentional?

jordancarlin:
> Isn't this going to put the superset extensions _and_ the subset ones in the ISA string? That seems slightly odd... Or is that intentional?

I think we decided that the generated ISA string should take the approach of being exhaustive to satisfy tools that might be looking for one or the other.

PR #1098, opened 2025-06-25
Use the same --strict flags for all Sail commands

Timmmm:
`--strict-bitvector` actually affects type checking so we need to use it for every command, otherwise we can get type errors with some backends and not others.


Alasdair:
Also just good to deduplicate any shared flags in general

pmundkur (approved):

Issue #933, opened 2025-05-11
Test RVFI in CI


jordancarlin:
Add TestRIG to CI to ensure RVFI implementation doesn't bitrot.

Issue #643, opened 2024-12-16
Move RVFI protocol implementation to C


Timmmm:
After #494 is done we should move the RVFI protocol implementation out of Sail and into C.

Timmmm:
So I was thinking

1. Move the struct definitions to C structs:

```
bitfield RVFI_DII_Execution_PacketV2 : bits(512) = {
  magic : 63 .. 0,         // must be set to 'trace-v2'
  trace_size : 127 .. 64,  // total size of the trace packet + extensions
...
}
```

To this (we can assume little endian):

```
#pragma pack(push, 1)
struct RVFI_DII_Execution_PacketV2 {
  uint64_t magic;
  uint64_t trace_size;
  ...
};
#pragma pack(pop)
```

2. Delete `rvfi_inst_data`, `rvfi_pc_data`, `rvfi_int_data_present`, `rvfi_mem_data`, `rvfi_write`, `rvfi_read`, etc.
3. In `riscv_callbacks.cpp`, wherever it calls `zrvfi_something`, create a similar function in `rvfi_dii.cpp` and call that instead.
4. Delete `riscv_fetch_rvfi.sail` and add a generic mechanism to force instructions into the model. This could be a parameter to `try_step()` for example. Or a separate `zforce_next_instruction()` call or something like that.

arichardson:
I think this is a good idea - CHERI QEMU already has it done in C. Maybe we can reuse some of that.

Some links to the relevant code:
https://github.com/CTSRD-CHERI/qemu/blob/967e7a86d8d7d1b0a730640354269afebaa0c3ed/target/riscv/cpu.c#L428

and https://github.com/CTSRD-CHERI/qemu/blob/967e7a86d8d7d1b0a730640354269afebaa0c3ed/target/riscv/rvfi_dii.h#L43

PR #1099, opened 2025-06-25
Move RVFI to C

trdthg:
- Move the struct definitions to C structs
- Move sail rvfi registers to C global variables
- Move callback functions to C functions
- Create sail api to access C global variables

---

resolve part of #643 
completely untested, waiting for #933.

- _rvfi_dii.sail_ has been completely rewritten. it's only used to declare some C functions that sail needs to access
- _rvfi_dii.h_ now contains all the struct and function definitions
- The current layout of _rvfi_dii.c_:
	1. Initialize rvfi related global variables
	2. Functions for creating empty structure. If you use the syntax `FOO foo = { 0 }`, there will be warnings
	3. Functions needed by `rvfi_handler` like `get_packet_v1/v2` ...
	5. Functions for callbacks
	6. Functions that need to be called by sail (these functions' headers are defined in _rvfi_dii_sail.h_ seperately to avoid C importing C++ class)
	7. Implementation of the original `Class rvfi_handler`

~The code is still a bit rough now, and the correctness still needs to wait for testing.~ TestRIG passed now

Currently, any suggestions for simplify, naming, file splitting, etc. are welcome


c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,376 @@
...
+    = Mk_RVFI_DII_Execution_Packet_Ext_MemAccess();
+bool rvfi_mem_data_present = false;
+bool rvfi_int_data_present = false;
+
+// ***************************************************************************
+
+RVFI_DII_Instruction_Packet Mk_RVFI_DII_Instruction_Packet()
+{
+  return RVFI_DII_Instruction_Packet {
+      .rvfi_insn = 0,
```
arichardson: Even though Clang has supported it for a while, this is C++20 syntax. I am not sure if all compilers are happy with it?

arichardson: Since it's C++ we can now add default initializers to the struct so there is no need for this.

c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,376 @@
...
+      .rvfi_mem_rmask = 0,
+      .rvfi_mem_wmask = 0,
+      .rvfi_rs1_addr = 0,
+      .rvfi_rs2_addr = 0,
+      .rvfi_rd_addr = 0,
+      .rvfi_trap = 0,
+      .rvfi_halt = 0,
+      .rvfi_intr = 0,
+  };
+}
```
arichardson:
```suggestion
```
These are not needed, we can use member ininitializers or constructors.

c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,376 @@
...
+// ****************************************************************************
+
+RVFI_DII_Instruction_Packet rvfi_instruction = Mk_RVFI_DII_Instruction_Packet();
+RVFI_DII_Execution_Packet_PC rvfi_pc_data = Mk_RVFI_DII_Execution_Packet_PC();
+RVFI_DII_Execution_Packet_InstMetadata rvfi_inst_data
+    = Mk_RVFI_DII_Execution_Packet_InstMetaData();
+RVFI_DII_Execution_Packet_Ext_Integer rvfi_int_data
+    = Mk_RVFI_DII_Execution_Packet_Ext_Integer();
+RVFI_DII_Execution_Packet_Ext_MemAccess rvfi_mem_data
+    = Mk_RVFI_DII_Execution_Packet_Ext_MemAccess();
```
arichardson:
```suggestion
static RVFI_DII_Instruction_Packet rvfi_instruction;
static RVFI_DII_Execution_Packet_PC rvfi_pc_data;
static RVFI_DII_Execution_Packet_InstMetadata rvfi_inst_data;
static RVFI_DII_Execution_Packet_Ext_Integer rvfi_int_data;
static RVFI_DII_Execution_Packet_Ext_MemAccess rvfi_mem_data;
```
And add member initializers/constructors.

c_emulator/rvfi_dii.h
```diff
@@ -23,7 +23,241 @@ class rvfi_handler {
...
+  uint8_t rvfi_trap;       // 687 .. 680, [85] Trap indicator: Invalid
+                           // decode, misaligned access or  misaligned address.
+  uint8_t rvfi_halt;       // 695 .. 688, [86] Halt indicator: Marks the
+                           // last instruction retired before halting execution.
+  uint8_t rvfi_intr;       // 703 .. 696, [87] Trap handler: Set for
+                           // first instruction in trap handler.
+} RVFI_DII_Execution_Packet_V1;
+
+typedef struct /* bits(512) */ {
+  uint64_t magic;      // 63 .. 0, must be set to 'trace-v2'
```
arichardson:
```suggestion
  char magic[8] = {'t', 'r', 'a', 'c', 'e', '-', 'v', '2'};      // 63 .. 0, must be set to 'trace-v2'
```

c_emulator/rvfi_dii.h
```diff
@@ -23,7 +23,241 @@ class rvfi_handler {
...
+  /// must always be zero.
+  uint8_t rvfi_rd_addr; // 263 .. 256
+  /// rvfi_rs1_addr and rvfi_rs2_addr are the decoded rs1 and rs1 register
+  /// addresses for the retired instruction. For an instruction that reads no
+  /// rs1/rs2 register, this output can have an arbitrary value. However, if
+  /// this output is nonzero then rvfi_rs1_rdata must carry the value stored
+  /// in that register in the pre-state.
+  uint8_t rvfi_rs1_addr; // 271 .. 264
+  uint8_t rvfi_rs2_addr; // 279 .. 272
+  uint8_t padding[5];
```
arichardson:
```suggestion
  uint64_t magic = char magic[8] = {'i', 'n', 't', '-', 'd', 'a', 't', 'a'}; // 63 .. 0, must be "int-data"
  /// rvfi_rd_wdata is the value of the x register addressed by rd after
  /// execution of this instruction. This output must be zero when rd is zero.
  uint64_t rvfi_rd_wdata = 0; // 127 ..  64
  /// rvfi_rs1_rdata/rvfi_rs2_rdata is the value of the x register addressed
  /// by rs1/rs2 before execution of this instruction. This output must be
  /// zero when rs1/rs2 is zero.
  uint64_t rvfi_rs1_rdata = 0; // 191 .. 128
  uint64_t rvfi_rs2_rdata = 0; // 255 .. 192
  /// rvfi_rd_addr is the decoded rd register address for the retired
  /// instruction. For an instruction that writes no rd register, this output
  /// must always be zero.
  uint8_t rvfi_rd_addr = 0; // 263 .. 256
  /// rvfi_rs1_addr and rvfi_rs2_addr are the decoded rs1 and rs1 register
  /// addresses for the retired instruction. For an instruction that reads no
  /// rs1/rs2 register, this output can have an arbitrary value. However, if
  /// this output is nonzero then rvfi_rs1_rdata must carry the value stored
  /// in that register in the pre-state.
  uint8_t rvfi_rs1_addr = 0; // 271 .. 264
  uint8_t rvfi_rs2_addr = 0; // 279 .. 272
  uint8_t padding[5] = {0};
```

c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,376 @@
...
+uint64_t rvfi_get_v2_trace_size()
+{
+  uint64_t trace_size = 512;
+  if (rvfi_int_data_present) {
+    trace_size = trace_size + 320;
+  }
+  if (rvfi_mem_data_present) {
+    trace_size = trace_size + 704;
+  }
+  return trace_size >> 3; // we have to return bytes not bits
```
arichardson: This is fragile, we could just use sizeof() now that the structs are in C.

c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,376 @@
...
+  };
+}
+
+void rvfi_zero_exec_packet()
+{
+  rvfi_instruction = Mk_RVFI_DII_Instruction_Packet();
+  rvfi_pc_data = Mk_RVFI_DII_Execution_Packet_PC();
+  rvfi_int_data = Mk_RVFI_DII_Execution_Packet_Ext_Integer();
+  rvfi_inst_data = Mk_RVFI_DII_Execution_Packet_InstMetaData();
+  rvfi_mem_data = Mk_RVFI_DII_Execution_Packet_Ext_MemAccess();
```
arichardson:
```suggestion
  rvfi_instruction = {};
  rvfi_pc_data = {};
  rvfi_int_data = {};
  rvfi_inst_data = {};
  rvfi_mem_data = {};
```

arichardson (commented):
I like this and it should make it easier to move away from my ugly binary format that was only designed this way since it was a straightfoward extension of the existing one, to something more flexible like CBOR, etc.

c_emulator/rvfi_dii_sail.h
```diff
@@ -0,0 +1,27 @@
...
+#include "sail.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// ****************************************************************************
+// Sail API
+
+mach_bits rvfi_get_cmd(unit);
```
Timmmm: You can use `uint64_t` here - `mach_bits` is an obsolete alias.

model/riscv_fetch_rvfi.sail
```diff
@@ -7,10 +7,10 @@
...
 
 function rvfi_fetch() -> FetchResult = {
-  rvfi_inst_data[rvfi_order] = minstret;
-  rvfi_pc_data[rvfi_pc_rdata] = zero_extend(get_arch_pc());
-  rvfi_inst_data[rvfi_mode] = zero_extend(privLevel_to_bits(cur_privilege));
-  rvfi_inst_data[rvfi_ixl] = zero_extend(misa[MXL]);
+  rvfi_set_pc_data_rdata(zero_extend(get_arch_pc()));
+  rvfi_set_inst_data_order(minstret);
+  rvfi_set_inst_data_mode(zero_extend(privLevel_to_bits(cur_privilege)));
+  rvfi_set_inst_data_ixl(zero_extend(misa[MXL]));
```
Timmmm: I was thinking that instead we could add handy accessors to the Sail model and preserve them, then call them from the RVFI code.

Like

```
function get_ixl() -> bits(2) = misa[MXL]
```

```
--c-preserve get_ixl
```

```
rvfi_inst_data.rvfi_ixl = zget_ixl();
```

Apart from that, and Alexander's comments this looks like almost exactly what I was thinking - nice one!

Eventually I think we could replace `rvfi_get_insn()` with a generic instruction forcing method but that can be in a future PR.

Timmmm (changes requested):

c_emulator/rvfi_dii.h
arichardson: I didn't add comments for the other structs but I would suggest adding the same default initializers there to simplify the code.

c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,242 @@
...
+  // the new trace format.
+  //  res = {};
+  return RVFI_DII_Execution_Packet_V1 {
+      .rvfi_halt = 0x03,
+  };
+}
+
+uint64_t rvfi_get_v2_trace_size()
+{
+  uint64_t trace_size = 512;
```
arichardson:
```suggestion
  uint64_t trace_size = sizeof(RVFI_DII_Execution_Packet_V2);
```

c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,242 @@
...
+  if (rvfi_int_data_present) {
+    trace_size = trace_size + sizeof(RVFI_DII_Execution_Packet_Ext_Integer);
+  }
+  if (rvfi_mem_data_present) {
+    trace_size = trace_size + sizeof(RVFI_DII_Execution_Packet_Ext_MemAccess);
+  }
+  return trace_size >> 3; // we have to return bytes not bits
```
arichardson:
```suggestion
  return trace_size >> 3; // we have to return bytes not bits
```
```suggestion
  return trace_size;
```

c_emulator/rvfi_dii.cpp
```diff
@@ -124,26 +360,59 @@ void rvfi_handler::get_and_send_packet(packet_reader_fn reader,
...
+                      config_print);
+    }
   } else {
     fprintf(stderr, "Sending v%d packets not implemented yet!\n",
             trace_version);
     abort();
   }
 }
 
+RVFI_DII_Instruction_Packet RVFI_DII_Instruction_Packet_from_u64(uint64_t value)
```
arichardson:
```suggestion
RVFI_DII_Instruction_Packet RVFI_DII_Instruction_Packet_from_u64(uint64_t value)
```
```suggestion
RVFI_DII_Instruction_Packet RVFI_DII_Instruction_Packet::from_u64(uint64_t value)
```

Might be nicer as a static function defined in the struct?

c_emulator/rvfi_dii.h
```diff
@@ -23,7 +23,222 @@ class rvfi_handler {
...
+  void send_packet_raw(const void *data, size_t len, bool config_print);
   unsigned trace_version = 1;
   int dii_port = -1;
   int dii_sock = -1;
 };
+
+// ****************************************************************************
+
+typedef struct {
+  uint32_t rvfi_insn = 0; // 31 ..  0, Instruction word: 32-bit instruction
```
arichardson: The formatting of the trailing comments is a bit odd, maybe better to just place them before the field definition instead?

c_emulator/rvfi_dii.h
```diff
@@ -23,7 +23,222 @@ class rvfi_handler {
...
 
+  void send_packet_raw(const void *data, size_t len, bool config_print);
   unsigned trace_version = 1;
   int dii_port = -1;
   int dii_sock = -1;
 };
+
+// ****************************************************************************
+
+typedef struct {
```
arichardson: Since we are using C++ there is no need for typedef struct, we can just use a named struct.
```suggestion
struct RVFI_DII_Execution_Packet_InstMetadata {
```

arichardson (commented):
Still not managed to get through the entire diff, but here are some further comments.

c_emulator/rvfi_dii.cpp
```diff
@@ -73,14 +88,14 @@ RVFI_DII_Execution_Packet_V1 rvfi_get_v2_support_packet()
 
 uint64_t rvfi_get_v2_trace_size()
 {
-  uint64_t trace_size = 512;
+  uint64_t trace_size = sizeof(RVFI_DII_Execution_Packet_V2);
   if (rvfi_int_data_present) {
     trace_size = trace_size + sizeof(RVFI_DII_Execution_Packet_Ext_Integer);
   }
   if (rvfi_mem_data_present) {
     trace_size = trace_size + sizeof(RVFI_DII_Execution_Packet_Ext_MemAccess);
```
arichardson:
```suggestion
    trace_size += sizeof(RVFI_DII_Execution_Packet_Ext_Integer);
  }
  if (rvfi_mem_data_present) {
    trace_size += sizeof(RVFI_DII_Execution_Packet_Ext_MemAccess);
```

c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,151 @@
...
+
+RVFI_DII_Instruction_Packet
+RVFI_DII_Instruction_Packet::from_u64(uint64_t value)
+{
+  return RVFI_DII_Instruction_Packet {
+      .rvfi_insn = (uint32_t)(value & 0xFFFFFFFF),
+      .rvfi_time = (uint16_t)((value >> 32) & 0xFFFF),
+      .rvfi_cmd = (uint8_t)((value >> 48) & 0xFF),
+      .padding = (uint8_t)((value >> 56) & 0xFF),
+  };
```
arichardson, line 32: Is GCC also happy with this C++20 syntax without extra flags?
If not
```suggestion
 RVFI_DII_Instruction_Packet pkt;
 pkt.rvfi_insn = (uint32_t)(value & 0xFFFFFFFF);
 pkt.rvfi_time = (uint16_t)((value >> 32) & 0xFFFF);
 pkt.rvfi_cmd = (uint8_t)((value >> 48) & 0xFF);
 pkt.padding = (uint8_t)((value >> 56) & 0xFF);
 return pkt;
```
This would be safer for older compilers, so we may want to use that instead?

arichardson: But if all compilers are fine with it I wouldn't bother making this change.

c_emulator/rvfi_dii.h
```diff
@@ -23,7 +23,238 @@ class rvfi_handler {
...
+  // 703 .. 640
+  // For memory operations (rvfi_mem_rmask and/or rvfi_mem_wmask are non-zero),
+  // rvfi_mem_addr holds the accessed memory location.
+  uint64_t rvfi_mem_addr = 0;
+};
+
+// ****************************************************************************
+
+struct RVFI_DII_Execution_Packet_V1 {
+  // [00 - 07] Instruction number: INSTRET value after completion.
```
arichardson: The numbers here are in bytes, the other structs use bits. Would be good to be consistent.

model/CMakeLists.txt
```diff
@@ -282,24 +281,6 @@ foreach (xlen IN ITEMS 32 64)
                         --c-preserve config_is_valid
                         --c-preserve generate_canonical_isa_string
                         # Preserve RVFI functions.
```
arichardson:
```suggestion
```

model/core/rvfi_dii.sail
```diff
@@ -6,260 +6,12 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-// "RISC-V Formal Interface - Direct Instruction Injection" support
-// For use with https://github.com/CTSRD-CHERI/TestRIG
```
arichardson, line 10: I'd keep this comment?

c_emulator/rvfi_dii.cpp
```diff
@@ -16,6 +16,151 @@
...
+// ****************************************************************************
+// Global Variables
+
+static RVFI_DII_Instruction_Packet rvfi_instruction = {};
+static RVFI_DII_Execution_Packet_PC rvfi_pc_data = {};
+static RVFI_DII_Execution_Packet_InstMetadata rvfi_inst_data = {};
+static RVFI_DII_Execution_Packet_Ext_Integer rvfi_int_data = {};
+static RVFI_DII_Execution_Packet_Ext_MemAccess rvfi_mem_data = {};
+static bool rvfi_mem_data_present = false;
+static bool rvfi_int_data_present = false;
```
arichardson: I feel like it would be cleaner to avoid global variables and make these members of the rvfi_handler struct.

c_emulator/rvfi_dii.h
trdthg: What about using bytes for all?

c_emulator/rvfi_dii.cpp
trdthg: I tested with `set(CMAKE_CXX_STANDARD 20)` and didn't encounter any issues with gcc 13.3 (ubuntu 2404)

c_emulator/rvfi_dii.cpp
trdthg: You can check this commit [`ccafb45` (#1099)](https://github.com/riscv/sail-riscv/pull/1099/commits/ccafb452bc8479beff51ecbd6b36c5b65bd17d07)

trdthg:
TestRIG passed now, using this branch and the master branch

model/core/rvfi_dii.sail
trdthg: move this to _rvfi_dii.h_?

c_emulator/rvfi_dii.h
arichardson: Sounds good to me.

c_emulator/rvfi_dii.cpp
Timmmm: We ran into a similar issue when compiling the Sail generated C as C++, because it includes designated initialisers. I have this in my `CMakeLists.txt`:

```
# We technically require C++20 since the C generated by Sail - which we compile
# as C++ - uses designated initialisers, a C++20 feature. However in practice
# much older compilers support this feature so everything does work with C++17,
# but you get a gazillion warnings on compilers that support C++20 if you use
# this feature without -std=c++20.
if (cxx_std_20 IN_LIST CMAKE_CXX_COMPILE_FEATURES)
    set(CMAKE_CXX_STANDARD 20)
else()
    set(CMAKE_CXX_STANDARD 17)
endif()
set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
```

Tbh I would almost say we just upgrade to C++20... it's fairly universal at this point. My only hesitation is that hardware people absolutely *LOVE* to use ancient Linux distros like Centos 7.

c_emulator/rvfi_dii.cpp
```diff
@@ -234,3 +413,107 @@ rvfi_prestep_t rvfi_handler::pre_step(bool config_print)
...
+    exit(1);
+  };
+}
+
+void rvfi_handler::rvfi_mem_exception(uint64_t paddr)
+{
+  /* Log only the memory address (without the value) if the write fails. */
+  rvfi_mem_data.rvfi_mem_addr = paddr;
+  rvfi_mem_data_present = true;
+  return;
```
Timmmm: You can delete all the `return;`s at the end of void functions - it's implicit.

c_emulator/rvfi_dii.cpp
```diff
@@ -234,3 +413,107 @@ rvfi_prestep_t rvfi_handler::pre_step(bool config_print)
...
+  rvfi_mem_data_present = true;
+  if (width <= 16) {
+    // TODO: report tag bit for capability writes and extend mask by one bit. */
+    rvfi_mem_data.clear_wdata();
+    convert_lbits_to_u8s(value, (uint8_t *)rvfi_mem_data.rvfi_mem_wdata);
+    rvfi_mem_data.rvfi_mem_wmask = rvfi_encode_width_mask(width);
+  } else {
+    fprintf(stderr, "Expected at most 16 bytes here!\n");
+    exit(1);
+  };
```
Timmmm: Extra semicolon, and on line 452.

c_emulator/sail_utils.cpp
```diff
@@ -0,0 +1,52 @@
...
+  result.len = num_bits;
+  result.bits = (mpz_t *)malloc(sizeof(mpz_t));
+  mpz_init(*result.bits);
+  mpz_import(*result.bits, num_bits, -1, 1, 0, 0, arr);
+  return result;
+}
+
+void convert_lbits_to_u8s(lbits val, uint8_t *arr)
+{
+  mpz_export(arr, NULL, -1, 1, 0, 0, *val.bits);
```
Timmmm: Pretty risky just to blindly write into an array of unknown size. It's probably fine for now, but can you add

```
// TODO: We should verify the size of `arr` to ensure we don't accidentally overflow.
```

c_emulator/rvfi_dii.cpp
```diff
@@ -234,3 +413,107 @@ rvfi_prestep_t rvfi_handler::pre_step(bool config_print)
...
+  return;
+}
+
+void rvfi_handler::rvfi_set_inst_data_ixl(uint8_t ixl)
+{
+  rvfi_inst_data.rvfi_ixl = ixl;
+  return;
+}
+
+uint32_t rvfi_handler::rvfi_get_insn(void)
```
Timmmm:
```suggestion
uint32_t rvfi_handler::rvfi_get_insn()
```

Not needed in C++; ditto in header.

c_emulator/rvfi_dii.cpp
```diff
@@ -234,3 +413,107 @@ rvfi_prestep_t rvfi_handler::pre_step(bool config_print)
...
+}
+
+void rvfi_handler::rvfi_write(uint64_t paddr, uint64_t width, lbits value)
+{
+  rvfi_mem_data.rvfi_mem_addr = paddr;
+  rvfi_mem_data_present = true;
+  if (width <= 16) {
+    // TODO: report tag bit for capability writes and extend mask by one bit. */
+    rvfi_mem_data.clear_wdata();
+    convert_lbits_to_u8s(value, (uint8_t *)rvfi_mem_data.rvfi_mem_wdata);
```
Timmmm: Minor nit but we should use C++ style casts - `reinterpret_cast<uint8_t*>()`... or `const_cast`? Why is the case needed?

c_emulator/rvfi_dii.h
```diff
@@ -1,6 +1,251 @@
...
+#define MAGIC_MEM_DATA {'m', 'e', 'm', '-', 'd', 'a', 't', 'a'}
+
+struct RVFI_DII_Execution_Packet_Ext_MemAccess {
+  // [07 - 00], magic must be "mem-data"
+  char magic[8] = MAGIC_MEM_DATA;
+  // [39 - 08]
+  // rvfi_mem_rdata is the pre-state data read from rvfi_mem_addr.
+  // rvfi_mem_rmask specifies which bytes are valid.
+  // CHERI-extension: widened to 32 bytes to allow reporting 129 bits
+  uint64_t rvfi_mem_rdata[4] = {};
```
Timmmm: Wouldn't `uint8_t rvfi_mem_rdata[32]` make more sense?

Timmmm (commented):
A few more comments. Getting pretty close I think.

c_emulator/rvfi_dii.h
```diff
@@ -1,6 +1,251 @@
...
+  char magic[8] = MAGIC_TRACE_V2;
+  // [15 - 00], total size of the trace packet + extensions
+  uint64_t trace_size = 0;
+  // [39 - 16]
+  RVFI_DII_Execution_Packet_InstMetadata basic_data = {};
+  // [55 - 39]
+  RVFI_DII_Execution_Packet_PC pc_data = {};
+  // [63 - 55]
+  uint64_t available_fields = 0;
+  // 448, Followed by RVFI_DII_Execution_Packet_Ext_Integer if set
```
arichardson: These are still in bits

c_emulator/rvfi_dii.h
```diff
@@ -20,10 +265,46 @@ class rvfi_handler {
...
+  void rvfi_halt_exec_packet();
+  void print_rvfi_exec();
 
-  unsigned trace_version = 1;
   int dii_port = -1;
   int dii_sock = -1;
+
+  // Trace States
+  unsigned trace_version = 1;
+
```
arichardson, line 308: I'd make these non-static members so they are only usable when rvfi is enabled. But can also be done as a cleanup later.

c_emulator/sail_utils.h
```diff
@@ -0,0 +1,39 @@
...
+unit prerr_sbits(const_sail_string str, const sbits op);
+
+void convert_lbits_to_u8s(lbits val, uint8_t *arr, size_t len);
+lbits convert_u8s_to_lbits(const uint8_t *arr, unsigned int num_bits);
+
+template <typename T> inline sbits make_sbits(T val)
+{
+  static_assert(sizeof(T) <= sizeof(uint64_t),
+                "sbits only accepts types that fit in uint64_t.");
+  return {sizeof(T) * 8, static_cast<unsigned int>(val)};
```
arichardson:
```suggestion
  return {sizeof(T) * 8, static_cast<uint64_t>(val)};
```
Shouldn't this be 64 bits rather than truncation?

c_emulator/sail_utils.h
```diff
@@ -0,0 +1,39 @@
...
+  return;
+}
+
+unit print_sbits(const_sail_string str, const sbits op);
+unit prerr_sbits(const_sail_string str, const sbits op);
+
+void convert_lbits_to_u8s(lbits val, uint8_t *arr, size_t len);
+lbits convert_u8s_to_lbits(const uint8_t *arr, unsigned int num_bits);
+
+template <typename T> inline sbits make_sbits(T val)
```
arichardson: This is only used for printing with print_sbits(), can't we avoid all this indirection and print the value directly?

c_emulator/rvfi_dii.cpp
```diff
@@ -14,8 +14,152 @@
...
+}
+
+void rvfi_handler::rvfi_halt_exec_packet()
+{
+  rvfi_inst_data.rvfi_halt = 0x01;
+}
+
+void rvfi_handler::print_rvfi_exec()
+{
+  print_sbits("rvfi_intr     : ", make_sbits(rvfi_inst_data.rvfi_intr));
```
arichardson:
```suggestion
  print_sbits("rvfi_intr     : %#016jx\n", (uintmax_t)rvfi_inst_data.rvfi_intr);
```

How about this instead? Uintmax_t and %j printf modifier to deal with different uint64_t representations. Or you could use the PRIx macros

c_emulator/rvfi_dii.h
trdthg: But they only have one bit.

c_emulator/rvfi_dii.h
trdthg: The problem is that，we have moved all the rvfi_xxx variables to class members, but from the sail side, it can only access regular C functions not class methods (maybe possibly? I don't know, please let me know if possible), so the associated variables/functions must to be static

c_emulator/sail_utils.cpp
trdthg: Added a "len"(byte) param to check the `len >= val.len`

c_emulator/rvfi_dii.cpp
trdthg: `print_sbits` currently decides whether to print in binary or hex based on the `sbits.len`, and it will only print the length of `sbits.len`. (The code is mainly modified from `print_bits` which is used for mpz_xxx.)

Using `uintmax_t` is obviously not very good? 👀

c_emulator/rvfi_dii.h
arichardson: Yeah you'd need to say bit xxx since all other numbers are in bytes

arichardson: Or something like byte 64 bit 1

c_emulator/rvfi_dii.cpp
arichardson: But all the callers here have a bit width that is a multiple of 8, so always print in hex. Printf with the PRI macros seems much easier

arichardson:
```suggestion
  printf("rvfi_intr     : %" PRIx8 "\n", rvfi_inst_data.rvfi_intr);
```

How about this instead?

c_emulator/rvfi_dii.h
arichardson: This should be fine we just have to dereference the global instance of the handler

Timmmm: Best thing would be to push the staticness back as far as possible, to a global variable. So the Sail C callbacks are something like:

```
void rvfi_set_inst_data_mode(uint8_t mode) {
  if (g_rvfi_handler != nullptr) {
    g_rvfi_handler->rvfi_set_inst_data_mode(mode);
  }
}
```

When we wrap the C code in a `struct` we can avoid the global entirely. I'm not super bothered about it at this stage though... especially if it is awkward. I'll see about getting that struct wrapping thing ready...

trdthg: Oh, I forget this. But now I have encountered another problem. The rvfi global variable is currently defined in _main.cpp_, which is a bit difficult to handle in the `CMakeLists.txt`

I am considering a solution to completely separate sail_model(lib, don't need rvfi) and c_simulator(exe, need rvfi), main_unit_tests(exe, don't need rvfi).

- set `callbacks` and `rvfi_sail_api` to empty impl. 
- introduce a register function, which could be called by c_simulator to replace them to any other implementation.

For sail-model, callbacks currently rely on the rvfi implementation, which is unnecessary. for rvfi_sail_apis, rvfi is not a part of riscv, but I haven't thought of a way to completely separate it. Setting the default implementation to empty can at least solve part of the dependency issues.

Anyway, the idea looks like this:

```cpp
using MemWriteCallback = unit (*)(const char *type, sbits paddr, uint64_t width, lbits value);

// default impl is empty
MemWriteCallback _mem_write_callback = nullptr;

// sail api wrapper
unit mem_read_callback(const char *type, sbits paddr, uint64_t width, lbits value)
{
  if (_mem_write_callback) {
    _mem_write_callback(type, paddr, width, value);
  }
  return UNIT;
}

// register custom impl
void register_callbacks(MemWriteCallback mem_write_callback = nullptr, ...)
{
  _mem_write_callback = mem_write_callback;
  _xxxxxxxxxxxxxxxxxx = xxx
}
```

then in _c_emulator.cpp_

```cpp

// real implementation
unit mem_write_callback(const char *type, sbits paddr, uint64_t width, lbits value)
{
  rvfi->rvfi_write(...)
}

// do register
register_callbacks(mem_write_callback, ...)
```

c_emulator/rvfi_dii.cpp
```diff
@@ -14,8 +14,141 @@
...
+}
+
+RVFI_DII_Execution_Packet_V1
+RVFI_DII_Execution_Packet_V1::rvfi_get_v2_support_packet()
+{
+  // Returning 0x3 (using the unused high bits) in halt instead of 0x1 means
+  // that we support the version 2 wire format. This is required to keep
+  // backwards compatibility with old implementations that do not support
+  // the new trace format.
+  //  res = {};
```
arichardson:
```suggestion
```

c_emulator/riscv_callbacks_rvfi.cpp
```diff
@@ -3,41 +3,41 @@
...
+#include "rvfi_dii.h"
+
+extern std::optional<rvfi_handler> rvfi;
 
 // Implementations of default callbacks for RVFI.
 // The model assumes that these functions do not change the state of the model.
 void rvfi_callbacks::mem_write_callback(const char *, sbits paddr,
                                         uint64_t width, lbits value)
 {
   if (config_enable_rvfi) {
```
arichardson: Isn't this always true by definition when the callbacks are registered now? I think we can drop the if.

c_emulator/rvfi_dii.cpp
```diff
@@ -88,40 +221,25 @@ bool rvfi_handler::setup_socket(bool config_print)
...
+  const size_t send_size = sizeof(T);
   if (config_print) {
-    print_bits("packet = ", packet);
+    print_raw_as_hex("packet = ", data);
     fprintf(stderr, "Sending packet with length %zd... ", send_size);
   }
   if (send_size > 4096) {
     fprintf(stderr, "Unexpected large packet size (> 4KB): %zd\n", send_size);
-    exit(EXIT_FAILURE);
+    exit(1);
```
arichardson: Any reason for this change? Symbolic names are generally cleaner.

c_emulator/rvfi_dii.cpp
```diff
@@ -88,40 +221,25 @@ bool rvfi_handler::setup_socket(bool config_print)
   return true;
 }
 
-void rvfi_handler::get_and_send_packet(packet_reader_fn reader,
-                                       bool config_print)
+template <typename T>
+void rvfi_handler::send_packet(const T *data, bool config_print)
```
arichardson:
```suggestion
void rvfi_handler::send_packet(const T &data, bool config_print)
```
Since we are writing only one element I would use a reference to show that it's not an array.

c_emulator/riscv_callbacks_rvfi.cpp
arichardson: And the same for the other functions in this file.

c_emulator/rvfi_dii.cpp
```diff
@@ -88,40 +221,25 @@ bool rvfi_handler::setup_socket(bool config_print)
...
-  unsigned char bytes[send_size];
-  /* mpz_export might not write all of the null bytes */
-  memset(bytes, 0, sizeof(bytes));
-  mpz_export(bytes, NULL, -1, 1, 0, 0, *(packet.bits));
-  /* Ensure that we can send a full packet */
-  if (write(dii_sock, bytes, send_size) != send_size) {
+  if (write(dii_sock, data, send_size) != send_size) {
     fprintf(stderr, "Writing RVFI DII trace failed: %s\n", strerror(errno));
-    exit(EXIT_FAILURE);
+    exit(1);
```
arichardson: Same here, keep EXIT_FAILURE?

c_emulator/rvfi_dii.h
```diff
@@ -9,8 +254,6 @@ enum rvfi_prestep_t {
   RVFI_prestep_ok,        // Ready for step
 };
 
-typedef void (*packet_reader_fn)(lbits *rop, unit);
-
 class rvfi_handler {
```
arichardson:
```suggestion
class rvfi_handler : public callbacks_if {
```

I think it would be simpler to implement the callback interface here, then you don't need all the delegation functions in riscv_callbacks_rvfi.

c_emulator/rvfi_dii.h
```diff
@@ -20,10 +263,45 @@ class rvfi_handler {
   void send_trace(bool config_print);
   rvfi_prestep_t pre_step(bool config_print);
 
+  // Callback API
+  void rvfi_write(uint64_t paddr, uint64_t width, lbits value);
+  void rvfi_read(uint64_t paddr, uint64_t width, lbits value);
+  void rvfi_mem_exception(uint64_t paddr);
+  void rvfi_wX(unsigned r, uint64_t v);
+  void rvfi_trap();
```
arichardson:
```suggestion
  void mem_write_callback(const char *type, sbits paddr, uint64_t width,
                          lbits value) override;
  void mem_read_callback(const char *type, sbits paddr, uint64_t width,
                         lbits value) override;
  void mem_exception_callback(sbits paddr, uint64_t num_of_exception) override;
  void xreg_full_write_callback(const_sail_string abi_name, sbits reg,
                                sbits value) override;
  void freg_write_callback(unsigned reg, sbits value) override;
  void csr_full_write_callback(const_sail_string csr_name, unsigned reg,
                               sbits value) override;
  void csr_full_read_callback(const_sail_string csr_name, unsigned reg,
                              sbits value) override;
  void vreg_write_callback(unsigned reg, lbits value) override;
  void pc_write_callback(sbits value) override;
  void trap_callback() override;
```

c_emulator/rvfi_dii_sail.cpp
```diff
@@ -0,0 +1,77 @@
...
+// Use anonymous namespaces to avoid naming conflicts
+namespace {
+rvfi_handler *rvfi;
+}
+
+void register_rvfi_dii_handler(rvfi_handler *handler)
+{
+  rvfi = handler;
+}
+
```
arichardson:
```suggestion
extern std::optional<rvfi_handler> rvfi;
```

If you declare the variable here you don't need the setter. All the other code below will work as before since you can check for presence of a std::optional with `if (rvfi)`.

c_emulator/rvfi_dii_sail.cpp
```diff
@@ -0,0 +1,77 @@
...
+  }
+  return UNIT;
+}
+
+uint32_t rvfi_get_insn(unit)
+{
+  if (rvfi) {
+    return rvfi->rvfi_get_insn();
+  } else {
+    exit(1);
```
arichardson: Print an error message before exiting?

c_emulator/sail_utils.h
```diff
@@ -0,0 +1,37 @@
...
+  }
+  fputs(post, stream);
+}
+
+template <typename T>
+void print_raw_as_hex(const_sail_string str, const T *data)
+{
+  fprint_raw_as_hex(str, data, "\n", stdout);
+  return;
+}
```
arichardson:
```suggestion
void fprint_raw_as_hex(const char *pre, const T &data, const char *post,
                       FILE *stream)
{
  fputs(pre, stream);
  const uint8_t *byte_ptr = reinterpret_cast<const uint8_t *>(&data);
  size_t size = sizeof(T);
  fputs("0x", stream);
  for (size_t i = size; i > 0; i--) {
    fprintf(stream, "%02x", byte_ptr[i - 1]);
  }
  fputs(post, stream);
}

template <typename T>
void print_raw_as_hex(const_sail_string str, const T &data)
{
  fprint_raw_as_hex(str, data, "\n", stdout);
  return;
}
```
I'd use a const reference here to show we are printing a single element.

arichardson (commented):
A few more comments but generally I think we are very close to being ready to merge this.

c_emulator/rvfi_dii_sail.cpp
trdthg: The reason I did this is that the rvfi variable is defined in c_emulator, and if I use extern, then riscv-model cannot be compiled separately.

I hope there is a way (and finally want) to completely separate rvfi from riscv-model. currently, I only thought of `class rvfi_api_if` (similar to callback_if) or adding a bunch of new callbacks can achieve this.

arichardson: Ah in that case maybe we just move the variable to the model and declare it as extern in the main.cpp?

trdthg:
Need some help dealing with lean build issue. @ineol 

```
Fatal error: exception Invalid_argument("List.combine")
```

It seems to come from `(List.combine import_files (List.combine import_names_camel import_refs_camel))`

I did my best to do some statistics

- len(import_names_camel) = 101
- len(import_refs_camel) = 102

ineol:
@trdthg thank you for the ping! we're trying to figure it out [there](https://github.com/rems-project/sail/issues/1423)

c_emulator/rvfi_dii_sail.cpp
Timmmm: My idea was that we add enough callbacks to `callback_if` that RVFI could be implemented with just that.

arichardson:
Is there anything missing from this PR? I believe we've merged all of this?

trdthg:
> Is there anything missing from this PR? 

Rebased and TestRIG passed with this and master branch. I have completed what I intended to do.

Also fixed the Lean build @ineol, it seems like just a minor issue
```sail
val rvfi_set_inst_data_insn = impure "rvfi_set_inst_data_insn" : bits(64) -> unit
function rvfi_set_inst_data_insn(_) = () // This line is not needed, I just wrote it wrong
```

model/postlude/step.sail
```diff
@@ -245,7 +245,7 @@ function try_step(step_no : nat, exit_wait : bool) -> bool = {
       // Record the next PC for RVFI.
       // TODO: Remove this and implement it in C.
       if get_config_rvfi() then {
-        rvfi_pc_data[rvfi_pc_wdata] = zero_extend(get_arch_pc())
+        rvfi_set_pc_data_wdata(zero_extend(get_arch_pc()))
```
arichardson, line 247: This Todo comment above can be deleted. Or can we completely remove this line since it should be handled by pc_write_callback?

arichardson (commented):
Mostly looking good. IMO we should merge this asap to avoid further conflicts and then improve it post-merge rather than stealing until it is perfect.

model/postlude/step.sail
trdthg: > it should be handled by pc_write_callback?
looks good, also hope to do this in further pr 😇.

c_emulator/CMakeLists.txt
```diff
@@ -32,6 +32,25 @@ add_custom_target(generated_config_schema DEPENDS config_schema.h)
...
+    rvfi_dii.cpp
+    rvfi_dii_sail.h
+    rvfi_dii_sail.cpp
+    sail_utils.h
+    sail_utils.cpp
+)
+
+target_include_directories(rvfi_dii
+    PUBLIC
+        # find `sail.h`.
```
Timmmm: Not sure this comment is correct?

c_emulator/sail_utils.cpp
```diff
@@ -0,0 +1,20 @@
...
+}
+
+void convert_lbits_to_u8s(lbits val, uint8_t *arr, size_t len)
+{
+  if (len * 8 < val.len) {
+    fprintf(stderr, "Not enough space to convert lbits to u8s!\n");
+    exit(EXIT_FAILURE);
+  }
+  mpz_export(arr, NULL, -1, 1, 0, 0, *val.bits);
+}
```
Timmmm, line 20: I think we can use `std::vector` here. Coincidentally I also needed these methods and wrote this, which does a bit more checking too.

## `lbits.h`
```
#pragma once

#include <sail.h>
#include <vector>

// Convert lbits a vector. Throws an exception if lb is not
// a multiple of 8 bits.
std::vector<uint8_t> lbits_to_vector(lbits lb);

// Convert a vector of bits to lbits.
lbits vector_to_lbits(const std::vector<uint8_t> &data);

// Convert to uint64_t. This does no checking on the sign or size.
// Bad things will probably happen if they aren't correct.
uint64_t lbits_to_int(lbits lb);
```

## `lbits.cpp`
```
#include "lbits.h"
#include <stdexcept>

// Convert lbits a vector. Throws an exception if lb is not
// a multiple of 8 bits.
std::vector<uint8_t> lbits_to_vector(lbits lb)
{
  if (lb.len % 8 != 0) {
    throw std::invalid_argument("Can't convert lbits to vector because its "
                                "length is not a multiple of 8");
  }

  // Number of bits in the mpz type. This must be less than
  // the number of bits Sail thought.
  std::size_t bits = mpz_sizeinbase(*lb.bits, 2);

  if (bits > lb.len) {
    throw std::invalid_argument(
        "lbits GMP content has more bits than Sail thinks");
  }

  std::vector<uint8_t> output(lb.len / 8, 0);

  // Export the lbits. Note this doesn't pad with zeros but
  // that's ok because data is prefilled with zeros.
  mpz_export(
      // Destination buffer.
      output.data(),
      // Where to write the output size (we ignore this).
      nullptr,
      // Endianness of words (-1 = little).
      -1,
      // Size of words in bytes.
      1,
      // Endianness *within* words (0 = native). Doesn't matter in this case.
      0,
      // Number of "nail" bits (upper bits to clear in each word)
      0,
      // The GMP value. Note the sign is ignored but it should always be
      // positive.
      *lb.bits);

  return output;
}

// Convert a vector of bits to lbits.
lbits vector_to_lbits(const std::vector<uint8_t> &data)
{
  lbits lb;
  CREATE(lbits)(&lb);

  mpz_import(
      // Output value.
      *lb.bits,
      // Size of input.
      data.size(),
      // Endianness of words (-1 = little).
      -1,
      // Size of words in bytes.
      1,
      // Endianness *within* words (0 = native). Doesn't matter in this case.
      0,
      // Number of "nail" bits (upper bits to clear in each word).
      0,
      // Input data.
      data.data());

  return lb;
}

uint64_t lbits_to_int(lbits lb)
{
  return CONVERT_OF(fbits, lbits)(lb, false);
}
```

c_emulator/riscv_sim.cpp
```diff
@@ -51,14 +51,12 @@ std::string trace_log_path;
 FILE *trace_log = NULL;
 std::string dtb_file;
 int rvfi_dii_port = 0;
-std::optional<rvfi_handler> rvfi;
+extern std::optional<rvfi_handler> rvfi;
```
Timmmm: This line should be in `rvfi_dii_sail.h`.

c_emulator/riscv_sim.cpp
```diff
@@ -613,29 +605,34 @@ int inner_main(int argc, char **argv)
...
   if (!dtb_file.empty()) {
     fprintf(stderr, "using %s as DTB file.\n", dtb_file.c_str());
     write_dtb_to_rom(read_file(dtb_file));
   }
 
-  const std::string &initial_elf_file = elfs[0];
-  uint64_t entry = rvfi ? rvfi->get_entry()
-                        : load_sail(initial_elf_file, /*main_file=*/true);
+  uint64_t entry;
+  if (rvfi) {
```
```suggestion
  if (rvfi != nullptr) {
```

c_emulator/riscv_sim.cpp
```diff
@@ -613,29 +605,34 @@ int inner_main(int argc, char **argv)
...
+    if (!rvfi->setup_socket(config_print_rvfi))
+      return 1;
```
```suggestion
      return EXIT_FAILURE;
```

Does this print an error message?

c_emulator/rvfi_dii.cpp
```diff
@@ -15,8 +15,140 @@
...
+// Struct Member Functions
+
+RVFI_DII_Instruction_Packet
+RVFI_DII_Instruction_Packet::from_u64(uint64_t value)
+{
+  return RVFI_DII_Instruction_Packet {
+      .rvfi_insn = (uint32_t)(value & 0xFFFFFFFF),
+      .rvfi_time = (uint16_t)((value >> 32) & 0xFFFF),
+      .rvfi_cmd = (uint8_t)((value >> 48) & 0xFF),
+      .padding = (uint8_t)((value >> 56) & 0xFF),
```
Timmmm: Should use `static_cast<>`s. The `& 0xF...`s are also unnecessary I think.

c_emulator/sail_utils.h
```diff
@@ -0,0 +1,37 @@
...
+lbits convert_u8s_to_lbits(const uint8_t *arr, unsigned int num_bits);
+
+template <typename T> inline sbits make_sbits(T val)
+{
+  static_assert(sizeof(T) <= sizeof(uint64_t),
+                "sbits only accepts types that fit in uint64_t.");
+  return {sizeof(T) * 8, static_cast<uint64_t>(val)};
+}
+
+template <typename T> inline lbits make_lbits(T val)
```
Timmmm: This should probably constrain `T` a bit, like this:

```
template <typename T, std::enable_if_t<std::is_integral_v<T>, bool> = true>
...
```

Also I don't think you need `inline`?

c_emulator/sail_utils.h
```diff
@@ -0,0 +1,37 @@
...
+{
+  fputs(pre, stream);
+  const uint8_t *byte_ptr = reinterpret_cast<const uint8_t *>(&data);
+  size_t size = sizeof(T);
+  fputs("0x", stream);
+  for (size_t i = size; i > 0; i--) {
+    fprintf(stream, "%02x", byte_ptr[i - 1]);
+  }
+  fputs(post, stream);
+}
```
Timmmm, line 17: I think rather than printing as hex it should convert it to `std::string`. Then we can switch to `std::format("{:x}", data);` in C++20. Until then you can use `stringstream` like [this](https://stackoverflow.com/a/5100745/265521).

c_emulator/rvfi_dii_sail.cpp
```diff
@@ -0,0 +1,72 @@
+#include <optional>
+#include "rvfi_dii.h"
+#include "rvfi_dii_sail.h"
+
+std::optional<rvfi_handler> rvfi;
+
+#ifdef __cplusplus
```
Timmmm: A `.cpp` file is always C++ - these `#ifdef`s are only ever needed in headers which might get `#include`d by C or C++. (You still need `extern "C"` though.)

c_emulator/riscv_sim.cpp
arichardson: Probably better to use .has_value()?

Timmmm: Oh yeah I forgot it wasn't a pointer... one of the reasons I don't like implicit bool conversion!

c_emulator/rvfi_dii.cpp
arichardson:
```suggestion
      .rvfi_insn = static_cast<uint32_t>(value),
      .rvfi_time = static_cast<uint16_t>(value >> 32),
      .rvfi_cmd = static_cast<uint8_t>(value >> 48),
      .padding = static_cast<uint8_t>(value >> 56),
```
As @Timmmm said no need for the masking.

c_emulator/sail_utils.cpp
trdthg: why not just upstream this? 👀
And if we use `std::vector`, there will be an extra copy from `uint8_t rvfi_mem_wdata[32]` to vector?
`std::span` looks like a good solution, but this also needs c++20

Timmmm: Yeah I'm happy for this to be used in this PR! I wouldn't worry about copying 32 bytes. But also I think C++20 is an acceptable minimum requirement at this point (we should probably discuss it though).

PR #1100, opened 2025-06-25
Change register bit-width of vstart from 16 to xlen

nadime15:
The spec says that `vstart` is XLEN bits wide so this changes the underlying value to `xlenbits`. This allows removing various `zero_extend()`s. It still only allows writing the lowest VLEN_pow bits.

### DIFF
--- a/model/riscv_insts_vext_vset.sail
+++ b/model/riscv_insts_vext_vset.sail
@@ -111,7 +111,7 @@ function clause execute VSETVLI(ma, ta, sew, lmul, rs1, rd) = {
 
   csr_write_callback("vtype", vtype.bits);
   csr_write_callback("vl", vl);
-  csr_write_callback("vstart", zero_extend(vstart));
+  csr_write_callback("vstart", vstart);
 
   RETIRE_SUCCESS
 }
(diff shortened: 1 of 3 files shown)


Timmmm:
Yeah there is a note about that in the comments:

```
// Note: The spec recommends trapping if vstart is out of bounds but the
// current implementation does not do that because write_CSR() doesn't
// allow traps.
```

I think you can also do a search and replace of `zero_extend(vstart)` and change it to just `vstart`.

nadime15:
> Yeah there is a note about that in the comments:

Oh I am blind, I did not see that

pmundkur (approved):
LGTM.  We should tune the PR description before adding to the merge queue.

Timmmm (approved):

jordancarlin:
> Yeah there is a note about that in the comments:
> 
> ```
> // Note: The spec recommends trapping if vstart is out of bounds but the
> // current implementation does not do that because write_CSR() doesn't
> // allow traps.
> ```
> 
> I think you can also do a search and replace of `zero_extend(vstart)` and change it to just `vstart`.

Might be worth opening an issue to track this.

jordancarlin (approved):

pmundkur:
> Might be worth opening an issue to track this.

Issue #1104

PR #1101, opened 2025-06-25
Move `effectivePrivilege` from `riscv_sys_regs` to `riscv_sys_control`.

pmundkur:
For Sdext, the effective privilege is also governed by the `dcsr` CSR.  However, Sdext itself needs access to `riscv_sys_regs`.  This move resolves the dependencies.


pmundkur:
This is part of pulling out pieces of #954 to reduce the size of that PR.

PR #1102, opened 2025-06-25
Add a cosimulator API for lockstep simulation with another simulator.

pmundkur:
This builds on the callbacks from the model to drive a simulator like Spike in lock-step.

Implement the cosimulator API for the Spike simulator.  This PR implements GPR and CSR cross-checks; floating-point and vector checks are left for later.

Add a USE_SPIKE build option for the Spike cosimulator; when disabled (default), use a null cosimulator.  When built to link against the Spike libraries, the Spike cosimulator is enabled with an `--enable-spike` cli option.

The Spike install is located using the $RISCV environment variable described in the Spike build instructions.  This needs to be in the environment for the USE_SPIKE build option.

TODOS: fix divergences in HTIF, PMP and floating point; handle vector.


pmundkur:
This is a draft, to get some feedback.  I'm wondering if there is some synergy with #1099.

Timmmm:
I think we shouldn't tie the simulation library interface up with co-simulation. Instead we should expose the simulator as a library that just lets you say "run one step & tell me what happened". Then that interface can be used by other consumers - cosimulators, structured logging, RVFI, etc.

The simulator library shouldn't be driving the cosimulation process. It should be the other way around - there's an external cosimulation process that drives/monitors the Sail simulation library. We might provide an example cosimulator that works with Spike but I don't think we want to require that because people have their own cosimulators (e.g. RVFI).

pmundkur:
> I think we shouldn't tie the simulation library interface up with co-simulation. Instead we should expose the simulator as a library that just lets you say "run one step & tell me what happened". Then that interface can be used by other consumers - cosimulators, structured logging, RVFI, etc.
> 
> The simulator library shouldn't be driving the cosimulation process. It should be the other way around - there's an external cosimulation process that drives/monitors the Sail simulation library. We might provide an example cosimulator that works with Spike but I don't think we want to require that because people have their own cosimulators (e.g. RVFI).

Agreed, that would be a better way to go.  The way it works in `riscv_sim.cpp` now felt quite hacky as I was doing it.  But I didn't want to first have to do a big refactor of `riscv_sim.cpp` just to get some useful lockstep debugging working.  This has already found a few subtle bugs (e.g. #1065, #1078 ).

The other change I would make is to try to use Spike's commitlog infrastructure instead of accessing its state directly.   I think the challenge with the more general (and better) approach you describe above is to have other simulators keep easily comparable commitlogs for efficient state comparison on every step.  This PR just punted and used the Sail model's commits to check Spike's state directly.

In the meantime, I'm happy to keep this PR as a draft and try to fill the remaining holes in the state comparison until we decide to do a big refactor.

PR #1108, opened 2025-06-29
Use newtype cfregidx for compressed floating point register indices

### LINKED ISSUE #1052: Add separate type for compressed floating point register indices
We have separate types for integer, float, vector and compressed registers, but compressed registers can also be float or integer, so maybe we want a compressed float register index type. See [here](https://github.com/riscv/sail-riscv/pull/1047#discussion_r2146327955).

KotorinMinami:
Add `newtype cfregidx` so that we don't need to use `cregidx` for compressed floating point register indices.

Fixes #1052

### DIFF
--- a/model/riscv_fdext_regs.sail
+++ b/model/riscv_fdext_regs.sail
@@ -54,12 +54,14 @@ function nan_unbox(m, x) = if 'n == 'm then x else (
 
 newtype fregidx = Fregidx : bits(5)
 newtype fregno = Fregno : range(0, 31)
+// identifiers in RVC floating-point instructions
+newtype cfregidx = Cfregidx : bits(3)
 
 function fregidx_bits(Fregidx(r) : fregidx) -> bits(5) = r
 
-// TODO: We should have a separate Cfregidx type for compressed float registers.
-function cregidx_to_fregidx (Cregidx(b) : cregidx) -> fregidx = Fregidx(0b01 @ b)
+function cfregidx_to_fregidx (Cfregidx(b) : cfregidx) -> fregidx = Fregidx(0b01 @ b)
 
+mapping encdec_cfreg : cfregidx <-> bits(3) = { Cfregidx(r) <-> r }
 /*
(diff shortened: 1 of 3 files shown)


pmundkur (approved):

model/riscv_fdext_regs.sail
```diff
@@ -54,12 +54,13 @@ function nan_unbox(m, x) = if 'n == 'm then x else (
 
 newtype fregidx = Fregidx : bits(5)
 newtype fregno = Fregno : range(0, 31)
+newtype cfregidx = Cfregidx : bits(3) /* identifiers in RVC floating-point instructions */
```
Timmmm: Very small nit but I generally prefer comments on their own line - it reduces formatting/blame churn (and I like the style better). Also I would use `//` - I don't really see any reason to use `/*`.

model/riscv_insts_zcd.sail
```diff
@@ -20,9 +20,8 @@ function clause execute (C_FLDSP(uimm, rd)) = {
 }
 
 mapping clause assembly = C_FLDSP(uimm, rd)
-      if (xlen == 32 | xlen == 64)
   <-> "c.fldsp" ^ spc() ^ freg_name(rd) ^ sep() ^ hex_bits_9(uimm @ 0b000) ^ "(" ^ sp_reg_name() ^ ")"
-      if (xlen == 32 | xlen == 64)
+  when (xlen == 32 | xlen == 64)
```
Timmmm: I think we could actually delete this now. RV128 doesn't exist. Could do it as a separate PR if you think that's too unrelated...

Timmmm (approved):

Timmmm:
LGTM, just needs a rebase (sorry!)

jordancarlin (approved):

Timmmm:
Sorry one more rebase needed for the `ast` -> `instruction` rename.

Timmmm:
Oops, mouse slipped.

PR #1110, opened 2025-06-30
Add Hypervisor CSR and virtual Supervisor CSR handling

trdthg:
Add Hypervisor CSRs with legalization functions and update Supervisor CSRs for virtual and non-virtual access.

This is part of the ongoing work for adding hypervisor support.

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -6,7 +6,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 // =======================================================================================
 
-// Machine-mode and supervisor-mode state definitions.
+// Machine-mode, hypervisor-mode, supervisor-mode state definitions.
 
 // privilege level
 
@@ -133,6 +133,10 @@ function clause write_CSR(0x301, value) = { misa = legalize_misa(misa, value); O
 // you are technically allowed to have writable misa[U/S].
 function clause currentlyEnabled(Ext_U) = hartSupports(Ext_U) & misa[U] == 0b1 & currentlyEnabled(Ext_Zicsr)
 function clause currentlyEnabled(Ext_S) = hartSupports(Ext_S) & misa[S] == 0b1 & currentlyEnabled(Ext_Zicsr)
+function clause currentlyEnabled(Ext_H) =
+  hartSupports(Ext_H) & misa[H] == 0b1 & misa[I] == 0b1 &
+  currentlyEnabled(Ext_S) & currentlyEnabled(Ext_Zicsr) &
+  (currentlyEnabled(Ext_Sv32) | currentlyEnabled(Ext_Sv39))
 // This will need adjusting for dynamic XLEN.
 function clause currentlyEnabled(Ext_Ssu64xl) = hartSupports(Ext_Ssu64xl) & currentlyEnabled(Ext_S)
 
@@ -145,6 +149,8 @@ function clause currentlyEnabled(Ext_Sv57) = hartSupports(Ext_Sv57) &  currently
 function virtual_memory_supported() -> bool = {
   currentlyEnabled(Ext_Sv32) | currentlyEnabled(Ext_Sv39) | currentlyEnabled(Ext_Sv48) | currentlyEnabled(Ext_Sv57)
 }
+
+function inVirtualPrivilege() -> bool = privLevel_is_virtual(cur_privilege)
 
 // This measure needs to be > currentlyEnabled_measure(Ext_Sv{32,39,48,57}),
 // currently all equal to 2, see currentlyEnabled_measure in extensions.sail
@@ -170,8 +176,8 @@ bitfield Mstatus : bits(64) = {
   //MDT  : 42,
   MPELP: 41,
 
-  //MPV  : 39,
-  //GVA  : 38,
+  MPV  : 39,
+  GVA  : 38,
 
   MBE  : 37,
   SBE  : 36,
@@ -211,8 +217,8 @@ private function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
   let o = [o with
     // MDT = v[MDT],
     MPELP = if hartSupports(Ext_Zicfilp) then v[MPELP] else 0b0,
-    // MPV = v[MPV],
-    // GVA = v[GVA],
+    MPV   = if currentlyEnabled(Ext_H) then v[MPV] else 0b0,
+    GVA   = if currentlyEnabled(Ext_H) then v[GVA] else 0b0,
     // We don't currently support changing MBE and SBE.
     // MBE = v[MBE],
     // SBE = v[SBE],
@@ -279,24 +285,49 @@ function clause write_CSR((0x300, value) if xlen == 64) = { mstatus = legalize_m
 function clause write_CSR((0x300, value) if xlen == 32) = { mstatus = legalize_mstatus(mstatus, mstatus.bits[63 .. 32] @ value); Ok(mstatus.bits[31 .. 0]) }
 function clause write_CSR((0x310, value) if xlen == 32) = { mstatus = legalize_mstatus(mstatus, value @ mstatus.bits[31 .. 0]); Ok(mstatus.bits[63 .. 32]) }
 
+bitfield Hstatus : bits(64) = {
+  HUPMM : 49 .. 48,
+  VSXL  : 33 .. 32,
+  VTSR  : 22,
+  VTW   : 21,
+  VTVM  : 20,
+  VGEIN : 17 .. 12,
+  HU    : 9,
+  SPVP  : 8,
+  SPV   : 7,
+  GVA   : 6,
+  VSBE  : 5,
+}
+
+function legalize_hstatus(h : Hstatus, v : bits(64)) -> Hstatus = {
+  let v = Mk_Hstatus(v);
+  [h with
+    // HUPMM = if currentlyEnabled(Ext_Ssnpm) & xlen == 64 then v[HUPMM] else 0b00,
+    // We don't currently support changing VSXL
+    // VSXL = v[VSXL]
+    VTSR  = v[VTSR],
+    VTW   = v[VTW],
+    VTVM  = v[VTVM],
+    VGEIN = v[VGEIN], // TODO: Add config option
+    HU    = v[HU],
+    SPVP  = v[SPVP],
+    SPV   = v[SPV],
+    GVA   = v[GVA],
+    // We don't currently support changing VSBE
+    // VSBE  = v[VSBE],
+  ]
+}
+
+register hstatus : Hstatus = [Mk_Hstatus(zeros()) with
+  VSXL = architecture_bits(if xlen == 32 then RV32 else RV64)
+]
+
+mapping clause csr_name_map = 0x600 <-> "hstatus"
+function clause is_CSR_accessible(0x600, _, _) = currentlyEnabled(Ext_H) //hstatus
+function clause read_CSR(0x600) = hstatus.bits[xlen - 1 .. 0]
+function clause write_CSR(0x600, value) = { hstatus = legalize_hstatus(hstatus, zero_extend(value)); Ok(hstatus.bits[xlen - 1 .. 0]) }
+
 // architecture and extension checks
-
-function architecture(priv : Privilege) -> Architecture = {
-  if   xlen == 32
-  then return RV32;
-
-  architecture_bits(match priv {
-    Machine           => misa[MXL],
-    Supervisor        => mstatus[SXL],
-    User              => mstatus[UXL],
-    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-  })
-}
-
-function in32BitMode() -> bool = {
-  architecture(cur_privilege) == RV32
-}
 
 bitfield Seccfg : bits(64) = {
   // Enable Zicfilp in M-mode
@@ -440,6 +471,54 @@ function clause write_CSR((0x30A, value) if xlen == 64) = { menvcfg = legalize_m
 function clause write_CSR((0x31A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, value @ menvcfg.bits[31 .. 0]); Ok(menvcfg.bits[63 .. 32]) }
 function clause write_CSR(0x10A, value) = { senvcfg = legalize_senvcfg(senvcfg, zero_extend(value)); Ok(senvcfg.bits[xlen - 1 .. 0]) }
 
+// Hypervisor Environment Configuration
+
+bitfield HEnvcfg : bits(64) = {
+  STCE  : 63,
+  PBMTE : 62,
+  ADUE  : 61,
+  DTE   : 59,
+  PMM   : 33 .. 32,
+  CBZE  : 7,
+  CBCFE : 6,
+  CBIE  : 5 .. 4,
+  SSE   : 3,
+  LPE   : 2,
+  FIOM  : 0,
+}
+
+register henvcfg : HEnvcfg
+
+function legalize_henvcfg(h : HEnvcfg, v : bits(64)) -> HEnvcfg = {
+  let v = Mk_HEnvcfg(v);
+  [h with
+    STCE  = if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then v[STCE]  else 0b0,
+    PBMTE = if currentlyEnabled(Ext_Svpbmt) then v[PBMTE] else 0b0,
+    // ADUE = if currentlyEnabled(Ext_Svadu) then v[ADUE] else 0b0,
+    // DTE  = if currentlyEnabled(Ext_Ssdbltrp) then v[DTE] else 0b0,
+    // PMM  = if currentlyEnabled(Ext_Ssnpm) then v[PMM] else 0b0,
+    CBZE  = if currentlyEnabled(Ext_Zicboz) then v[CBZE]  else 0b0,
+    CBCFE = if currentlyEnabled(Ext_Zicbom) then v[CBCFE] else 0b0,
+    CBIE  = if currentlyEnabled(Ext_Zicbom) then v[CBIE]  else 0b00,
+    // SSE  = if currentlyEnabled(Ext_Zicfiss) then v[SSE] else 0b0,
+    LPE   = if hartSupports(Ext_Zicfilp) then v[LPE]  else 0b0,
+    FIOM  = if sys_enable_writable_fiom  then v[FIOM] else 0b0,
+  ]
+}
+
+mapping clause csr_name_map = 0x60A <-> "henvcfg"
+mapping clause csr_name_map = 0x61A <-> "henvcfgh"
+
+function clause is_CSR_accessible(0x60A, _, _) = currentlyEnabled(Ext_H) // henvcfg
+function clause is_CSR_accessible(0x61A, _, _) = currentlyEnabled(Ext_H) & xlen == 32 // henvcfgh
+
+function clause read_CSR(0x60A) = henvcfg.bits[xlen - 1 .. 0]
+function clause read_CSR(0x61A if xlen == 32) = henvcfg.bits[63 .. 32]
+
+function clause write_CSR((0x60A, value) if xlen == 32) = { henvcfg = legalize_henvcfg(henvcfg, henvcfg.bits[63 .. 32] @ value); Ok(henvcfg.bits[xlen - 1 .. 0]) }
+function clause write_CSR((0x60A, value) if xlen == 64) = { henvcfg = legalize_henvcfg(henvcfg, value); Ok(henvcfg.bits[xlen - 1 .. 0]) }
+function clause write_CSR((0x61A, value) if xlen == 32) = { henvcfg = legalize_henvcfg(henvcfg, value @ henvcfg.bits[31 .. 0]); Ok(henvcfg.bits[63 .. 32]) }
+
 // Return whether or not FIOM is currently active, based on the current
 // privilege and the menvcfg/senvcfg settings. This means that I/O fences
 // imply memory fence.
@@ -557,6 +636,18 @@ function clause is_CSR_accessible(0x106, _, _) = currentlyEnabled(Ext_S) // scou
 function clause read_CSR(0x106) = zero_extend(scounteren.bits)
 function clause write_CSR(0x106, value) = { scounteren = legalize_scounteren(scounteren, value); Ok(zero_extend(scounteren.bits)) }
 
+// hcounteren
+private function legalize_hcounteren(_c : Counteren, v : xlenbits) -> Counteren = {
+  let supported_counters = sys_writable_hpm_counters[31 .. 3] @ 0b111;
+  Mk_Counteren(v[31 .. 0] & supported_counters)
+}
+
+register hcounteren : Counteren
+mapping clause csr_name_map = 0x606 <-> "hcounteren"
+function clause is_CSR_accessible(0x606, _, _) = currentlyEnabled(Ext_H)
+function clause read_CSR(0x606) = zero_extend(hcounteren.bits)
+function clause write_CSR(0x606, value) = { hcounteren = legalize_hcounteren(hcounteren, value); Ok(zero_extend(hcounteren.bits)) }
+
 // mcounteren
 function legalize_mcounteren(_c : Counteren, v : xlenbits) -> Counteren = {
   let supported_counters = sys_writable_hpm_counters[31 .. 3] @ 0b111;
@@ -568,7 +659,6 @@ mapping clause csr_name_map = 0x306  <-> "mcounteren"
 function clause is_CSR_accessible(0x306, _, _) = currentlyEnabled(Ext_U) // mcounteren
 function clause read_CSR(0x306) = zero_extend(mcounteren.bits)
 function clause write_CSR(0x306, value) = { mcounteren = legalize_mcounteren(mcounteren, value); Ok(zero_extend(mcounteren.bits)) }
-
 
 // mcountinhibit
 bitfield Counterin : bits(32) = {
@@ -633,6 +723,57 @@ function clause read_CSR(0xF13) = mimpid
 function clause read_CSR(0xF14) = mhartid
 function clause read_CSR(0xF15) = mconfigptr
 
+// Additional M-mode registers introduced by hypervisor extension
+
+// TODO: Add config options and legalization functions for mtinst and mtval2
+register mtinst : xlenbits
+register mtval2 : xlenbits
+
+mapping clause csr_name_map = 0x34A  <-> "mtinst"
+mapping clause csr_name_map = 0x34B  <-> "mtval2"
+
+function clause is_CSR_accessible(0x34A, _, _) = currentlyEnabled(Ext_H) // mtinst
+function clause is_CSR_accessible(0x34B, _, _) = currentlyEnabled(Ext_H) // mtval2
+
+function clause read_CSR(0x34A) = mtinst
+function clause read_CSR(0x34B) = mtval2
+
+function clause write_CSR(0x34A, value) = { mtinst = value; Ok(mtinst) }
+function clause write_CSR(0x34B, value) = { mtval2 = value; Ok(mtval2) }
+
+// HS-mode registers
+
+register htimedelta : bits(64)
+
+mapping clause csr_name_map = 0x605 <-> "htimedelta"
+mapping clause csr_name_map = 0x615 <-> "htimedeltah"
+
+// TODO: Should raise virtual instruction exception if accessed by VU/VS-Mode.
+// Needs refactoring of CSR access logic.
+function clause is_CSR_accessible(0x605, _, _) = currentlyEnabled(Ext_H) // htimedelta
+function clause is_CSR_accessible(0x615, _, _) = currentlyEnabled(Ext_H) & xlen == 32 // htimedeltah
+
+function clause read_CSR(0x605) = htimedelta[xlen - 1 .. 0]
+function clause read_CSR(0x615 if xlen == 32) = htimedelta[63 .. 32]
+
+function clause write_CSR(0x605, value) = { htimedelta[xlen - 1 .. 0] = value; Ok(htimedelta[xlen - 1 .. 0]) }
+function clause write_CSR((0x615, value) if xlen == 32) = { htimedelta[63 .. 32] = value; Ok(htimedelta[63 .. 32]) }
+
+register htval  : xlenbits
+register htinst : xlenbits
+
+mapping clause csr_name_map = 0x643 <-> "htval"
+mapping clause csr_name_map = 0x64A <-> "htinst"
+
+function clause is_CSR_accessible(0x643, _, _) = currentlyEnabled(Ext_H) // htval
+function clause is_CSR_accessible(0x64A, _, _) = currentlyEnabled(Ext_H) // htinst
+
+function clause read_CSR(0x643) = htval
+function clause read_CSR(0x64A) = htinst
+
+function clause write_CSR(0x643, value) = { htval = value; Ok(htval) }
+function clause write_CSR(0x64A, value) = { htinst = value; Ok(htinst) }
+
 // S-mode registers
 
 // sstatus reveals a subset of mstatus
@@ -696,10 +837,47 @@ private function legalize_sstatus(m : Mstatus, v : xlenbits) -> Mstatus = {
   legalize_mstatus(m, lift_sstatus(m, Mk_Sstatus(zero_extend(v))).bits)
 }
 
+// vsstatus
+register vsstatus : Mstatus = [Mk_Mstatus(zeros()) with
+  UXL = architecture_bits(if xlen == 32 then RV32 else RV64)
+]
+
+mapping clause csr_name_map = 0x200  <-> "vsstatus"
+function clause is_CSR_accessible(0x200, _, _) = currentlyEnabled(Ext_H) // vsstatus
+function clause read_CSR(0x200) = lower_mstatus(vsstatus).bits[xlen - 1 .. 0]
+function clause write_CSR(0x200, value) = { vsstatus = legalize_sstatus(vsstatus, value); Ok(lower_mstatus(vsstatus).bits[xlen - 1 .. 0]) }
+
 mapping clause csr_name_map = 0x100  <-> "sstatus"
 function clause is_CSR_accessible(0x100, _, _) = currentlyEnabled(Ext_S) // sstatus
-function clause read_CSR(0x100) = lower_mstatus(mstatus).bits[xlen - 1 .. 0]
-function clause write_CSR(0x100, value) = { mstatus = legalize_sstatus(mstatus, value); Ok(lower_mstatus(mstatus).bits[xlen - 1 .. 0]) }
+
+function clause read_CSR(0x100) = lower_mstatus(if inVirtualPrivilege() then vsstatus else mstatus).bits[xlen - 1 .. 0]
+function clause write_CSR(0x100, value) =
+  if inVirtualPrivilege()
+  then { vsstatus = legalize_sstatus(vsstatus, value); Ok(lower_mstatus(vsstatus).bits[xlen - 1 .. 0]) }
+  else { mstatus  = legalize_sstatus(mstatus, value); Ok(lower_mstatus(mstatus).bits[xlen - 1 .. 0]) }
+
+// Virtual Supervisor related state register
+register vstvec    : Mtvec
+register vsscratch : xlenbits
+register vsepc     : xlenbits
+register vscause   : Mcause
+register vstval    : xlenbits
+
+mapping clause csr_name_map = 0x240  <-> "vsscratch"
+mapping clause csr_name_map = 0x242  <-> "vscause"
+mapping clause csr_name_map = 0x243  <-> "vstval"
+
+function clause is_CSR_accessible(0x240, _, _) = currentlyEnabled(Ext_H) // vsscratch
+function clause is_CSR_accessible(0x242, _, _) = currentlyEnabled(Ext_H) // vscause
+function clause is_CSR_accessible(0x243, _, _) = currentlyEnabled(Ext_H) // vstval
+
+function clause read_CSR(0x240) = vsscratch
+function clause read_CSR(0x242) = vscause.bits
+function clause read_CSR(0x243) = vstval
+
+function clause write_CSR(0x240, value) = { vsscratch = value; Ok(vsscratch) }
+function clause write_CSR(0x242, value) = { vscause.bits = value; Ok(vscause.bits) }
+function clause write_CSR(0x243, value) = { vstval = value; Ok(vstval) }
 
 // other non-VM related supervisor state
 register stvec    : Mtvec
@@ -716,13 +894,75 @@ function clause is_CSR_accessible(0x140, _, _) = currentlyEnabled(Ext_S) // sscr
 function clause is_CSR_accessible(0x142, _, _) = currentlyEnabled(Ext_S) // scause
 function clause is_CSR_accessible(0x143, _, _) = currentlyEnabled(Ext_S) // stval
 
-function clause read_CSR(0x140) = sscratch
-function clause read_CSR(0x142) = scause.bits
-function clause read_CSR(0x143) = stval
+function clause read_CSR(0x140) = if inVirtualPrivilege() then vsscratch else sscratch
+function clause read_CSR(0x142) = if inVirtualPrivilege() then vscause.bits else scause.bits
+function clause read_CSR(0x143) = if inVirtualPrivilege() then vstval else stval
 
-function clause write_CSR(0x140, value) = { sscratch = value; Ok(sscratch) }
-function clause write_CSR(0x142, value) = { scause.bits = value; Ok(scause.bits) }
-function clause write_CSR(0x143, value) = { stval = value; Ok(stval) }
+function clause write_CSR(0x140, value) = if inVirtualPrivilege() then { vsscratch = value; Ok(vsscratch) } else { sscratch = value; Ok(sscratch) }
+function clause write_CSR(0x142, value) = if inVirtualPrivilege() then { vscause.bits = value; Ok(vscause.bits) } else { scause.bits = value; Ok(scause.bits) }
+function clause write_CSR(0x143, value) = if inVirtualPrivilege() then { vstval = value; Ok(vstval) } else { stval = value; Ok(stval) }
+
+//
+// HS-mode address translation and protection (hgatp) layout.
+// The actual hgatp register is defined in an architecture-specific file.
+bitfield Hgatp64 : bits(64) = {
+  Mode : 63 .. 60,
+  VMID : 57 .. 44,
+  PPN  : 43 .. 0
+}
+
+bitfield Hgatp32 : bits(32) = {
+  Mode : 31,
+  VMID : 28 .. 22,
+  PPN  : 21 .. 0
+}
+
+// TODO: Add these options to config file once Hypervisor supported is added
+let HBare_supported  : bool = true
+let Sv32x4_supported : bool = HBare_supported  & xlen == 32
+let Sv39x4_supported : bool = HBare_supported  & xlen == 64
+let Sv48x4_supported : bool = Sv39x4_supported & xlen == 64
+let Sv57x4_supported : bool = Sv48x4_supported & xlen == 64
+
+// TODO: Add VMIDLEN config option and mask VMID field to implemented width
+function legalize_hgatp(
+  arch : Architecture,
+  prev_value : xlenbits,
+  written_value : xlenbits,
+) -> xlenbits = {
+  // Unlike satp, a write to hgatp with an unsupported MODE value is not ignored.
+  // Instead, the fields are WARL in the normal way: an unsupported MODE falls
+  // back to HBare while preserving the other fields (VMID, PPN).
+  if xlen == 32 then {
+    let v = Mk_Hgatp32(written_value);
+    // Force 16 KiB alignment with PPN[1:0] = 0b00
+    let h = Mk_Hgatp32(v[Mode] @ 0b00 @ v[VMID] @ v[PPN][21 .. 2] @ 0b00);
+    match hgatpMode_of_bits(arch, 0b000 @ h[Mode]) {
+      None() => prev_value,
+      Some(mode) => match mode {
+        HBare  if HBare_supported  => h.bits,
+        Sv32x4 if Sv32x4_supported => h.bits,
+        _ => [h with Mode = zeros()].bits,
+      }
+    }
+  } else if xlen == 64 then {
+    let v = Mk_Hgatp64(written_value);
+    // Force 16 KiB alignment with PPN[1:0] = 0b00
+    let h = Mk_Hgatp64(v[Mode] @ 0b00 @ v[VMID] @ v[PPN][43 .. 2] @ 0b00);
+    match hgatpMode_of_bits(arch, h[Mode]) {
+      None() => prev_value,
+      Some(mode) => match mode {
+        HBare  if HBare_supported  => h.bits,
+        Sv39x4 if Sv39x4_supported => h.bits,
+        Sv48x4 if Sv48x4_supported => h.bits,
+        Sv57x4 if Sv57x4_supported => h.bits,
+        _ => [h with Mode = zeros()].bits,
+      }
+    }
+  } else {
+    internal_error(__FILE__, __LINE__, "Unsupported xlen " ^ dec_str(xlen))
+  }
+}
 
 //
 // S-mode address translation and protection (satp) layout.
@@ -801,14 +1041,32 @@ val get_16_random_bits = impure {
 
 // There are several features that are controlled by machine/supervisor enable
 // bits (m/senvcfg, m/scounteren, etc.). This abstracts that logic.
-function feature_enabled_for_priv(p : Privilege, machine_enable_bit : bits(1), supervisor_enable_bit : bits(1)) -> bool = match p {
+function feature_enabled_for_priv(p : Privilege, machine_enable_bit : bits(1), supervisor_enable_bit : bits(1), hypervisor_enable_bit : bits(1)) -> bool = match p {
   Machine => true,
   Supervisor => machine_enable_bit == 0b1,
   User => machine_enable_bit == 0b1 & (not(currentlyEnabled(Ext_S)) | supervisor_enable_bit == 0b1),
-  VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-  VirtualUser => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+  VirtualSupervisor => machine_enable_bit == 0b1 & hypervisor_enable_bit == 0b1,
+  VirtualUser => machine_enable_bit == 0b1 & hypervisor_enable_bit == 0b1 & supervisor_enable_bit == 0b1,
 }
 
 // Return true if the counter is enabled.
 function counter_enabled(index: range(0, 31), priv : Privilege) -> bool =
-  feature_enabled_for_priv(priv, mcounteren.bits[index], scounteren.bits[index])
+  feature_enabled_for_priv(priv, mcounteren.bits[index], scounteren.bits[index], hcounteren.bits[index])
+
+// Return the effective architecture for the given privilege mode
+function architecture(priv : Privilege) -> Architecture = {
+  if   xlen == 32
+  then return RV32;
+
+  architecture_bits(match priv {
+    Machine           => misa[MXL],
+    Supervisor        => mstatus[SXL],
+    User              => mstatus[UXL],
+    VirtualSupervisor => hstatus[VSXL],
+    VirtualUser       => vsstatus[UXL],
+  })
+}
+
+function in32BitMode() -> bool = {
+  architecture(cur_privilege) == RV32
+}
--- a/model/exceptions/sys_exceptions.sail
+++ b/model/exceptions/sys_exceptions.sail
@@ -23,8 +23,8 @@ function prepare_trap_vector(p : Privilege, cause : Mcause) -> xlenbits = {
     Machine           => mtvec,
     Supervisor        => stvec,
     User              => internal_error(__FILE__, __LINE__, "Invalid privilege level"),
-    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualUser       => internal_error(__FILE__, __LINE__, "Invalid privilege level"),
+    VirtualSupervisor => vstvec,
   };
   match tvec_addr(tvec, cause) {
     Some(epc) => epc,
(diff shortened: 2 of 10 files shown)


model/riscv_hext_control.sail
```diff
@@ -0,0 +1,79 @@
...
+mapping clause csr_name_map = 0x34A  <-> "mtinst"
+mapping clause csr_name_map = 0x34B  <-> "mtval2"
+
+/* hypervisor trap setup */
+mapping clause csr_name_map = 0x600  <-> "hstatus"
+mapping clause csr_name_map = 0x602  <-> "hedeleg"
+mapping clause csr_name_map = 0x603  <-> "hideleg"
+mapping clause csr_name_map = 0x604  <-> "hie"
+mapping clause csr_name_map = 0x606  <-> "hcounteren"
+mapping clause csr_name_map = 0x607  <-> "hgeie"
```
pmundkur: `hdelegh` is missing

model/riscv_hext_control.sail
```diff
@@ -0,0 +1,79 @@
...
+mapping clause csr_name_map = 0x280  <-> "vsatp"
+
+function clause is_CSR_defined(0x34A) = currentlyEnabled(Ext_H) // mtinst
+function clause is_CSR_defined(0x34B) = currentlyEnabled(Ext_H) // mtval2
+function clause is_CSR_defined(0x600) = currentlyEnabled(Ext_H) // hstatus
+function clause is_CSR_defined(0x602) = currentlyEnabled(Ext_H) // hedeleg
+function clause is_CSR_defined(0x603) = currentlyEnabled(Ext_H) // hideleg
+function clause is_CSR_defined(0x604) = currentlyEnabled(Ext_H) // hie
+function clause is_CSR_defined(0x606) = currentlyEnabled(Ext_H) // hcounteren
+function clause is_CSR_defined(0x607) = currentlyEnabled(Ext_H) // hgeie
```
pmundkur: `hdelegh` is missing at `0x612`, and also needs `xlen == 32`

model/riscv_hext_control.sail
```diff
@@ -0,0 +1,79 @@
...
+function clause is_CSR_defined(0x643) = currentlyEnabled(Ext_H) // htval
+function clause is_CSR_defined(0x644) = currentlyEnabled(Ext_H) // hip
+function clause is_CSR_defined(0x645) = currentlyEnabled(Ext_H) // hvip
+function clause is_CSR_defined(0x64A) = currentlyEnabled(Ext_H) // htinst
+function clause is_CSR_defined(0xE12) = currentlyEnabled(Ext_H) // hgeip
+function clause is_CSR_defined(0x60A) = currentlyEnabled(Ext_H) // henvcfg
+function clause is_CSR_defined(0x61A) = currentlyEnabled(Ext_H) // henvcfgh
```
pmundkur:
```suggestion
function clause is_CSR_defined(0x61A) = currentlyEnabled(Ext_H) & xlen == 32 // henvcfgh
```

model/riscv_hext_control.sail
```diff
@@ -0,0 +1,79 @@
...
+function clause is_CSR_defined(0x680) = currentlyEnabled(Ext_H) // hgatp
+function clause is_CSR_defined(0x605) = currentlyEnabled(Ext_H) // htimedelta
+function clause is_CSR_defined(0x615) = currentlyEnabled(Ext_H) // htimedeltah
```
pmundkur:
```suggestion
function clause is_CSR_defined(0x615) = currentlyEnabled(Ext_H) & xlen == 32 // htimedeltah
```

pmundkur (changes requested):
The model is inconsistent about where it puts `is_CSR_defined` and `csr_name_map`.  For system registers, these are in `_sys_regs` instead of `_sys_control`, while for vector they are in `_vext_control`.  Not in this PR, but we should be more consistent about this.

nadime15:
~@Mingzh~ @trdthg i I rebased this PR, please take a look as well and let us update it if anything is missing or incorrect. I added all Hypervisor CSRs together with their legalization methods. I also updated the corresponding Supervisor CSRs that are virtualized and added the additional Machine mode CSR's and bitfields.

**This PR/Draft is still work in progress**. I need to read the spec a few more times to make sure I have not missed anything, but first round feedback would be helpful. 

There a few design choices I am not so sure.

1. I first centralized the handling of the virtualized CSRs, but I realized quickly that this would require quite a bit of refactoring. So I stopped, since I was not sure if it was worth it and its a bit ugly, especially the write. I had something like this in mind.

```
function virtualized_S_CSR_read(csr : bits(12)) -> xlenbits = {
  let is_virtual = privLevel_is_virtual(cur_privilege);
  match csr {
    0x100 => if is_virtual then vsstatus.bits[xlen - 1 .. 0] else lower_mstatus(mstatus).bits[xlen - 1 .. 0],
    0x104 => if is_virtual then lower_mie_to_vsie(mie, hideleg).bits else lower_mie(mie, mideleg).bits,
    0x105 => if is_virtual then get_vstvec()   else get_stvec(),			
    0x140 => if is_virtual then vsscratch     else sscratch,				
    0x141 => if is_virtual then get_xepc(VirtualSupervisor) else get_xepc(Supervisor),
    0x142 => if is_virtual then vscause.bits  else scause.bits,
    0x143 => if is_virtual then vstval        else stval,	 

    0x144 => if is_virtual then lower_mip_to_vsip(mip, hideleg).bits     else lower_mip(mip, mideleg).bits,
    0x180 => if is_virtual then vsatp         else satp,
    _     => internal_error(__FILE__, __LINE__, "Unhandled virtualized CSR read"),
  }
}

function virtualized_S_CSR_write(csr : bits(12), value : xlenbits) -> xlenbits = {
  if privLevel_is_virtual(cur_privilege) then {
    match csr {
      0x100 => { vsstatus     = legalize_vsstatus(vsstatus, value); vsstatus.bits[xlen - 1 .. 0] },
      0x104 => { mie         = legalize_vsie(mie, hideleg, value); Ok(lower_mie_to_vsie(mie, hideleg).bits) },
      0x105 => { set_vstvec(value) },
      0x140 => { vsscratch    = value; vsscratch },
      0x141 => { set_xepc(VirtualSupervisor, value) },
      0x142 => { vscause.bits = value; vscause.bits },
      0x143 => { vstval       = value; vstval },

      0x144 => { mip = legalize_vsip(mip, hideleg, value); lower_mip_to_vsip(mip, hideleg).bits },
      0x180 => { vsatp        = legalize_satp(architecture(cur_privilege), vsatp, value); vsatp },
      _     => internal_error(__FILE__, __LINE__, "Unhandled virtualized CSR write"),
    }
  } else {
    match csr {
      0x100 => { mstatus     = legalize_sstatus(mstatus, value); lower_mstatus(mstatus).bits[xlen - 1 .. 0] },
      0x104 => { mie         = mie = legalize_sie(mie, mideleg, value); Ok(lower_mie(mie, mideleg).bits) },
      0x105 => { set_stvec(value) },

      0x140 => { sscratch    = value; sscratch },
      0x141 => { set_xepc(Supervisor, value) },
      0x142 => { scause.bits = value; scause.bits },
      0x143 => { stval       = value; stval },
      0x144 => { mip         = legalize_sip(mip, mideleg, value); Ok(lower_mip(mip, mideleg).bits) },

      0x180 => { satp        = legalize_satp(architecture(cur_privilege), satp, value); satp },

      _     => internal_error(__FILE__, __LINE__, "Unhandled S-mode CSR write"),
    }
  }
}
```

2. In general, if people are okay with the way I split the CSRs, I tried to group them as much as possible. However, some H and VS registers had to be placed in sys_regs.sail for now. 

Once this is merged, additional steps such as the trap handler, hypervisor instructions, and two stage translation can be built on top of this but it can already serve as a base.

trdthg:
> But I realized quickly that this would require quite a bit of refactoring.

Yes. I’ve also attempted several refactors before, but the results were always less than ideal or just can't move on.  (I recall running into circular dependency issues while dealing with hgapt, really bad experience. :D)

Here’s my current opinion: while the "H" extension chapter is technically independent, the H CSRs it provides are essentially system-level. From an engineering perspective, H CSRs are largely replicas of S CSRs, with significant coupling at the code level. We should treat H CSRs and S CSRs as the same level. Just directly put vsxx reg related definations and functions next to sxx reg.

Take `vsepc` as an example, it’s the VS-mode version of sepc. In #1419, the solution I eventually chosed was placing it right next to `sepc` in `model/exceptions/sys_exceptions.sail`, rather than in `h_regs.sail` or `sys_regs.sail`. This approach made the codebase immediately more organized.

Furthermore, once you adopt the approch above, more optimization opportunities will show up. For example, the read/write logic across mepc, sepc, and vsepc can be refactore from a different abstraction angle.

For now, I tend to this kind of simpler, more straightforward approach that is clear at a glance.

> BTW, you @ the wrong person, hope it didn't bother him.

nadime15:
@trdthg If you have the time, could you review this PR?

nadime15 (approved):

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,327 @@
...
+  [m with VSSI = v[VSSI]]
+}
+
+function lower_mip_to_hvip(m : Minterrupts) -> Minterrupts = {
+  [Mk_Minterrupts(zeros()) with
+    VSEI = m[VSEI],
+    VSTI = m[VSTI],
+    VSSI = m[VSSI],
+  ]
+}
```
trdthg:
```suggestion
function lower_mip_to_hvip(m : Minterrupts) -> Minterrupts =
  [Mk_Minterrupts(zeros()) with
    VSEI = m[VSEI],
    VSTI = m[VSTI],
    VSSI = m[VSSI],
  ]
```

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,327 @@
...
+register henvcfg : HEnvcfg
+
+function legalize_henvcfg(h : HEnvcfg, v : bits(64)) -> HEnvcfg = {
+  let v = Mk_HEnvcfg(v);
+  [h with
+    STCE  = if currentlyEnabled(Ext_Sstc)    then v[STCE]  else 0b0,
+    PBMTE = if currentlyEnabled(Ext_Svpbmt)  then v[PBMTE] else 0b0,
+    // ADUE  = if currentlyEnabled(Ext_Svadu) then v[ADUE] else 0b0,
+    // DTE   = if currentlyEnabled(Ext_Ssdbltrp) then v[DTE] else 0b0,
+    // PMM   = if currentlyEnabled(Ext_Ssnpm) then v[PMM] else 0b0,
```
trdthg: > The PMM field is read-only zero for RV32.

Although Ssnpm hasn't been implemented yet, it's still worth adding this to the comments.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,327 @@
...
+    (_,    0x0) => Some(HBare),
+    (RV32, 0x1) => Some(Sv32x4),
+    (RV64, 0x8) => Some(Sv39x4),
+    (RV64, 0x9) => Some(Sv48x4),
+    (RV64, 0xA) => Some(Sv57x4),
+    (_, _)      => None()
+  }
+
+bitfield Hgatp64 : bits(64) = {
+  Mode : 63 .. 60,
```
trdthg:
```suggestion
  MODE : 63 .. 60,
```

Image

If we are following the casing shown in the diagram, maybe we should change it to uppercase?

model/core/sys_regs.sail
```diff
@@ -771,6 +846,34 @@ private function lower_mstatus(m : Mstatus) -> Sstatus = {
...
 
+private function legalize_vsstatus(o : Sstatus, v : xlenbits) -> Sstatus = {
+  let s = Mk_Sstatus(zero_extend(v));
+
+  let dirty = extStatus_of_bits(s[FS]) == Dirty |
+              extStatus_of_bits(s[XS]) == Dirty |
+              extStatus_of_bits(s[VS]) == Dirty;
+  [o with
+    SD   = bool_to_bit(dirty),
+    // UXL
```
trdthg:
```suggestion
    // SDT = if currentlyEnabled(Ext_Ssdbltrp) then s[SDT] else 0b0,
```

model/core/hypervisor_regs.sail
nadime15: Yeah, I intended to do that, but since satp.Asid and satp.Mode still have only the first letter capitalized, I followed that style. I would prefer to keep it like that (in order to keep the diffs small) and once we consolidate these two we can change that.

model/core/hypervisor_regs.sail
nadime15: I mean is this not this implied by the currentlyEnabled clause that is only enabled if we are in RV64? I would rather keep it like that unless someone else things its worth mentioning as well

model/core/hypervisor_regs.sail
trdthg: Sounds nice

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,285 @@
...
+    VSTI = m[VSTI],
+    VSSI = m[VSSI],
+  ]
+}
+
+mapping clause csr_name_map = 0x645 <-> "hvip"
+mapping clause csr_name_map = 0x604 <-> "hie"
+mapping clause csr_name_map = 0x644 <-> "hip"
+
+function clause is_CSR_accessible(0x645, _, _) = currentlyEnabled(Ext_H) // hvip
```
radimkrcmar: All the H CSRs are not available in VS mode (virtual-instruction exception).
I think it would be good to implement a generic check for that early.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,285 @@
...
+  LPE   : 2,
+  FIOM  : 0,
+}
+
+register henvcfg : HEnvcfg
+
+function legalize_henvcfg(h : HEnvcfg, v : bits(64)) -> HEnvcfg = {
+  let v = Mk_HEnvcfg(v);
+  [h with
+    STCE  = if currentlyEnabled(Ext_Sstc)    then v[STCE]  else 0b0,
```
radimkrcmar: STCE bit is read-only zero when menvcfg.STCE is 0.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,285 @@
...
+function clause read_CSR(0x60A) = henvcfg.bits[xlen - 1 .. 0]
+function clause read_CSR(0x61A if xlen == 32) = henvcfg.bits[63 .. 32]
+
+function clause write_CSR((0x60A, value) if xlen == 32) = { henvcfg = legalize_henvcfg(henvcfg, henvcfg.bits[63 .. 32] @ value); Ok(henvcfg.bits[xlen - 1 .. 0]) }
+function clause write_CSR((0x60A, value) if xlen == 64) = { henvcfg = legalize_henvcfg(henvcfg, value); Ok(henvcfg.bits[xlen - 1 .. 0]) }
+function clause write_CSR((0x61A, value) if xlen == 32) = { henvcfg = legalize_henvcfg(henvcfg, value @ henvcfg.bits[31 .. 0]); Ok(henvcfg.bits[63 .. 32]) }
+
+// Hypervisor Counter Enable
+
+function legalize_hcounteren(_c : Counteren, v : xlenbits) -> Counteren = {
```
radimkrcmar: Why not use `legalize_mcounteren` (possibly after renaming to `legalize_counteren`), since the code is the same?

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,285 @@
...
+  written_value : xlenbits,
+) -> xlenbits = {
+  if xlen == 32 then {
+    let v = Mk_Hgatp32(written_value);
+    // Force 16 KiB alignment with PPN[1:0] = 0b00
+    let h = Mk_Hgatp32(v[Mode] @ 0b00 @ v[VMID] @ v[PPN][21 .. 2] @ 0b00);
+    match hgatpMode_of_bits(arch, 0b000 @ h[Mode]) {
+      None() => prev_value,
+      Some(mode) => match mode {
+        HBare  if currentlyEnabled(Ext_Svbare) => h.bits,
```
radimkrcmar: Is it specified that Svbare implies HBare?
And similar for the other SvABx4 modes... I don't recall a connection, so they could be completely new config options.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,285 @@
...
+        Sv32x4 if currentlyEnabled(Ext_Sv32)   => h.bits,
+        _ => prev_value,
```
radimkrcmar: Would sail catch that these two options are actually exhaustive?
(2 cases of a 1 bit variable)

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,285 @@
...
+      }
+    }
+  } else if xlen == 64 then {
+    let v = Mk_Hgatp64(written_value);
+    // Force 16 KiB alignment with PPN[1:0] = 0b00
+    let h = Mk_Hgatp64(v[Mode] @ 0b00 @ v[VMID] @ v[PPN][43 .. 2] @ 0b00);
```
radimkrcmar: There should also be a config option of VMID length.  (Maybe just slap a TODO here, so we don't forget.)

model/core/sys_regs.sail
```diff
@@ -555,6 +597,48 @@ function clause write_CSR((0x302, value) if xlen == 32) = { medeleg = legalize_m
...
+function clause is_CSR_accessible(0x603, _, _) = currentlyEnabled(Ext_H) // hideleg
+
+function clause read_CSR(0x602) = hedeleg.bits[xlen - 1 .. 0]
+function clause read_CSR(0x612 if xlen == 32) = hedeleg.bits[63 .. 32]
+function clause read_CSR(0x603) = hideleg.bits[xlen - 1 .. 0]
+
+function clause write_CSR((0x602, value) if xlen == 64) = { hedeleg = legalize_hedeleg(hedeleg, value); Ok(hedeleg.bits) }
+function clause write_CSR((0x602, value) if xlen == 32) = { hedeleg = legalize_hedeleg(hedeleg, hedeleg.bits[63 .. 32] @ value); Ok(hedeleg.bits[31 .. 0]) }
+function clause write_CSR((0x612, value) if xlen == 32) = { hedeleg = legalize_hedeleg(hedeleg, value @ hedeleg.bits[31 .. 0]); Ok(hedeleg.bits[63 .. 32]) }
+function clause write_CSR(0x603, value) = { hideleg = legalize_hideleg(hideleg, value); Ok(hideleg.bits) }
```
radimkrcmar:
```
For bits of mideleg that are zero, the corresponding bits in hideleg are read-only zeros.
```
I think it would be nice to implement this clause for documentation and sail extension purposes, even though it doesn't seem to make a difference now.

model/core/sys_regs.sail
```diff
@@ -771,6 +871,35 @@ private function lower_mstatus(m : Mstatus) -> Sstatus = {
...
+private function legalize_vsstatus(o : Sstatus, v : xlenbits) -> Sstatus = {
+  let s = Mk_Sstatus(zero_extend(v));
+
+  let dirty = extStatus_of_bits(s[FS]) == Dirty |
+              extStatus_of_bits(s[XS]) == Dirty |
+              extStatus_of_bits(s[VS]) == Dirty;
+  [o with
+    SD   = bool_to_bit(dirty),
+    // UXL
+    UXL  = s[UXL],
```
radimkrcmar: We don't support switching UXL, and it's won't be correct just to pass it through, so maybe just a TODO instead for now?

radimkrcmar (changes requested):
Look good in general.

I also noticed that the `vstimecmp` CSR is missing.

model/core/hypervisor_regs.sail
nadime15: Yeah I wanted to add this as part of the 2-Stage Address Translation PR, even though we dont even support configuration of ASID length.

model/core/hypervisor_regs.sail
nadime15: Yes, that is a good point. Unfortunately, we are still waiting for #1402, which will allow us to return a Virtual Instruction exception for CSR accesses under VS.

model/core/hypervisor_regs.sail
nadime15: Yeah Good question, I think I did this because we also have `legalize_scounteren`. I assumed this implied that each mode requires its own legalize_* method. However, in this case, we can use `legalize_mcounteren` instead.

model/core/hypervisor_regs.sail
nadime15: > Is it specified that Svbare implies HBare?

Yes and no (its implied I would say), but you are right I should remove this. 

I found in the spec: 

> The Hypervisor extension depends on ... and standard page-based address translation must be supported, either Sv32 for RV32, or a minimum of Sv39, for RV64.

and 

> and Implementations are not required to support all defined MODE settings when HSXLEN=64.

(I don't know if the statement above implies that Sv32x4 needs to be supported for HSXLEN=32/RV32).

I guess this implies that only the modes that are actually required are (H)Bare for RV32 and RV64.

So if I understand correctly, this means we would need to add config options for the other ones as well (Sv32x, Sv39x4, Sv48x and Sv48x4).

Off-topic but I just realized that we need to check that the number of physical bits is at least 34 in RV32 with Sv32. Otherwise, you could translate from a Guest Virtual Address to a Guest Physical Address and then access it in Bare mode (G-Stage) with a physical address that might not fit. Luckily, #1412 addresses this issue.

model/core/hypervisor_regs.sail
nadime15: Nope due to `mode` being a 4 bit variable unfortunately and the if statement.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,285 @@
...
+
+function legalize_hvip(m : Minterrupts, v : xlenbits) -> Minterrupts = {
+  let v = Mk_Minterrupts(v);
+  [m with
+    VSEI = v[VSEI],
+    VSTI = v[VSTI],
+    VSSI = v[VSSI],
+  ]
+}
+
```
pmundkur: Might want to leave a TODO for 
```
For each writable bit in sie, the corresponding bit shall be read-only zero in both hip and hie. Hence, the
nonzero bits in sie and hie are always mutually exclusive, and likewise for sip and hip.
```

model/core/sys_regs.sail
```diff
@@ -497,17 +523,33 @@ private function legalize_mie(o : Minterrupts, v : xlenbits) -> Minterrupts = {
...
 }
-
 // exception processing state
 
 bitfield Medeleg : bits(64) = {
+  SAMO_GPage_Fault  : 23,
+  Virtual_Instr     : 22,
+  Load_GPage_Fault  : 21,
+  Fetch_GPage_Fault : 20,
+  Software_Check    : 19,
```
pmundkur: `legalize_medeleg()` will have to zero out these new bits when there is no hypervisor.
```
For exceptions that cannot occur in less privileged modes, the corresponding medeleg bits should be read-only zero.
```

model/core/sys_regs.sail
pmundkur: Leaving a TODO comment for this is fine for now; see also the mutual exclusivity constraints coming from `sie`.

model/core/sys_regs.sail
```diff
@@ -732,6 +819,22 @@ function clause read_CSR(0xF13) = mimpid
...
+mapping clause csr_name_map = 0x34B  <-> "mtval2"
+
+function clause is_CSR_accessible(0x34A, _, _) = currentlyEnabled(Ext_H) // mtinst
+function clause is_CSR_accessible(0x34B, _, _) = currentlyEnabled(Ext_H) // mtval2
+
+function clause read_CSR(0x34A) = mtinst
+function clause read_CSR(0x34B) = mtval2
+
+function clause write_CSR(0x34A, value) = { mtinst = value; Ok(mtinst) }
+function clause write_CSR(0x34B, value) = { mtval2 = value; Ok(mtval2) }
```
pmundkur, line 742: Might want to leave config and legalization TODOs for these:

```
mtval2 is a WARL register that must be able to hold zero and may be capable of holding only an arbitrary
subset of other 2-bit-shifted guest physical addresses, if any.
mtinst is a WARL register that need only be able to hold the values that the implementation may
automatically write to it on a trap.
```

For e.g. I think mtval2 legalization would be affected by https://github.com/riscv/sail-riscv/pull/1412,  Similarly for `htval` and `htinst`.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,292 @@
...
+    STCE  = if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then v[STCE]  else 0b0,
+    PBMTE = if currentlyEnabled(Ext_Svpbmt)  then v[PBMTE] else 0b0,
+    // ADUE  = if currentlyEnabled(Ext_Svadu) then v[ADUE] else 0b0,
+    // DTE   = if currentlyEnabled(Ext_Ssdbltrp) then v[DTE] else 0b0,
+    // PMM   = if currentlyEnabled(Ext_Ssnpm) then v[PMM] else 0b0,
+    CBZE  = if currentlyEnabled(Ext_Zicboz)  then v[CBZE]  else 0b0,
+    CBCFE = if currentlyEnabled(Ext_Zicbom)  then v[CBCFE] else 0b0,
+    CBIE  = if currentlyEnabled(Ext_Zicbom)  then v[CBIE]  else 0b00,
+    // SSE = if currentlyEnabled(Ext_Zicfiss) then v[SSE] else 0b0,
+    LPE   = if currentlyEnabled(Ext_Zicfilp) then v[LPE]   else 0b0,
```
pmundkur:
```suggestion
    LPE   = if hartSupports(Ext_Zicfilp) then v[LPE]   else 0b0,
```
`currentlyEnabled(Zicfilp)` includes the `get_xLPE()` predicate that applies when decoding the `LPAD` instruction but not when writing the LPE bits (see `legalize_{ms}envcfg()`).

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,292 @@
...
+// Hypervisor Time Delta
+
+register htimedelta : bits(64)
+
+mapping clause csr_name_map = 0x605 <-> "htimedelta"
+mapping clause csr_name_map = 0x615 <-> "htimedeltah"
+
+// TODO: Should raise virtual instruction exception if accessed by VU/VS-Mode.
+// Needs refactoring of CSR access logic.
+// TODO: Time CSR needs to modified and shall return htimedelta + mtime
```
pmundkur:
```suggestion
// TODO: Time CSR needs to be modified and shall return htimedelta + mtime
```
Should this comment go into `extensions/Zicntr/zicntr_control.sail ` where `mtime` is actually read?

model/core/sys_regs.sail
```diff
@@ -832,22 +980,70 @@ private function lower_mie(m : Minterrupts, d : Minterrupts) -> Sinterrupts = {
...
   lift_sip(m, d, Mk_Sinterrupts(v))
 }
 
 mapping clause csr_name_map = 0x144  <-> "sip"
 function clause is_CSR_accessible(0x144, _, _) = currentlyEnabled(Ext_S) // sip
-function clause read_CSR(0x144) = lower_mip(mip, mideleg).bits
-function clause write_CSR(0x144, value) = { mip = legalize_sip(mip, mideleg, value); Ok(lower_mip(mip, mideleg).bits) }
-
+function clause read_CSR(0x144) =
+if privLevel_is_virtual(cur_privilege)
```
pmundkur: indentation
```suggestion
  if privLevel_is_virtual(cur_privilege)
```

model/core/sys_regs.sail
```diff
@@ -864,12 +1060,37 @@ private function legalize_sie(m : Minterrupts, d : Minterrupts, v : xlenbits) ->
...
+register vscause   : Mcause
+register vstval    : xlenbits
+
+mapping clause csr_name_map = 0x240  <-> "vsscratch"
+mapping clause csr_name_map = 0x242  <-> "vscause"
+mapping clause csr_name_map = 0x243  <-> "vstval"
 
+function clause is_CSR_accessible(0x240, _, _) = currentlyEnabled(Ext_H) // vsscratch
+function clause is_CSR_accessible(0x242, _, _) = currentlyEnabled(Ext_H) // vscause
+function clause is_CSR_accessible(0x243, _, _) = currentlyEnabled(Ext_H) // vstval
```
pmundkur, line 872: These registers should not be accessible from VS-mode.  The `check_CSR_priv()` function breaks here, see #1503.   There's no need to fix that in this PR, since the fix should be isolated to `check_CSR_priv()`.

pmundkur (changes requested):
I haven't finished reviewing this, but some comments so far...

model/core/sys_regs.sail
pmundkur: Actually, these and the other `vs*` registers should require `& cur_privilege != VirtualSupervisor` since both HS-mode and VS-mode use the same encoding in the CSR address.

model/core/hypervisor_regs.sail
radimkrcmar: > >  and Implementations are not required to support all defined MODE settings when HSXLEN=64.
>
> (I don't know if the statement above implies that Sv32x4 needs to be supported for HSXLEN=32/RV32).

I interpret it the same way, although it is inconsistent with RV32 satp, which doesn't have to support both modes.

> I guess this implies that only the modes that are actually required are (H)Bare for RV32 and RV64.

Hm, I don't think RV64 HBare is required, because it's just one of the `all defined MODE settings`.
(Bare in satp is not needed if you have other modes, so it would make sense for HBare to be separate too.)

> So if I understand correctly, this means we would need to add config options for the other ones as well (Sv32x, Sv39x4, Sv48x and Sv48x4).

I agree, and HBare.

> Off-topic but I just realized that we need to check that the number of physical bits is at least 34 in RV32 with Sv32.

I think it is fine to have less physical bits than what first stage translation produces (34 on RV32 and 56 on RV64). 
Anything above the physical width triggers an exception, and htval/mtval2 is right shifted by two bits to fit it.

model/core/sys_regs.sail
nadime15: Yeah, you are right. Thats something I planned to fix separately, but it might be better to do it now (I also need #1402 for this to be fully complet).

```
function check_CSR_priv(csr : csreg, p : Privilege) -> bool = {
  let csr_priv = csrPriv(csr);
  let priv_level_bits = privLevel_to_bits(p);
  // S-level or higher (HS, VS, or M)
  if csr_priv == 0b10 then priv_level_bits >=_u 0b01 else priv_level_bits >=_u csr_priv
}
```

and then later I check, for each `is_CSR_accessible`, whether the access was virtual. If so I return a Virtual Instruction Exception, and then something like:

```
function clause is_CSR_accessible(0x600, p, _) = {  // hstatus
  if not(currentlyEnabled(Ext_H)) then
    CSRAccess_Illegal
  else if privLevel_is_virtual(p) then
    CSRAccess_Virtual
  else
    CSRAccess_Success
}
```

I could check whether we access hypervisor CSRs in while in VU/VS in check_CSR_priv() and return CSRAccess_Virtual, but I think this is cleaner for now.

model/core/hypervisor_regs.sail
nadime15: > Hm, I don't think RV64 HBare is required, because it's just one of the all defined MODE settings.
> (Bare in satp is not needed if you have other modes, so it would make sense for HBare to be separate too.)

This took me a bit to understand, but what you are essentially saying is that satp can exist even without Bare for example, on RV32 it could be set by default to Sv32. In that case, the satp register would have to be initialized with SATP.MODE = Sv32 and be read-only. Does this make sense? ~~I am pretty sure we don’t support this yet~~. 

satp still seems to follow a partial hierarchy. Sv57 is only supported if Sv49 is supported, Sv49 only if Sv39, and so on. So I guess we would follow the same structure here for Hgatp.mode? 

But supporting no mode at all (not even Bare/HBare) effectively means that satp/hgatp is hardwired to zero, which implies no address translation == Bare mode.

model/core/sys_regs.sail
pmundkur: I created #1504 before seeing your reply :/  Yes, either fix would work; but it might be best as a separate PR.

nadime15: Sure no problem, thanks!

model/core/hypervisor_regs.sail
radimkrcmar: > Does this make sense?

Yes.

> So I guess we would follow the same structure here for Hgatp.mode?

I don't think it's actually specified...   Seems like all modes are independent.

(I think the hierarchy is going to be a standard in practice, though, since the implementation cost is small, and RVA23 profile requires both HBare, and x4 of all modes that satp supports.)

> But supporting no mode at all (not even Bare/HBare) effectively means that satp/hgatp is hardwired to zero, which implies no address translation == Bare mode.

It ought to be a configuration error if no modes are supported.
You can't express no mode at all, since there every value has a different meaning. :)

model/core/hypervisor_regs.sail
nadime15: Thats already enforced by the lowering functions  `lower_mie_to_hie`, `lower_mie` `lower_mip_to_hip` and `lower_mip`. For example, bits that are writable in sie are not writable in hie and vice versa.

model/core/hypervisor_regs.sail
nadime15: Ahh ok I see thanks! 

```
It ought to be a configuration error if no modes are supported.
You can't express no mode at all, since there every value has a different meaning. :)
```
Oh really? I think currently our model is able to run S-mode with no mode supported?! 

Edit: Ok seems like we auto-enable Svbare when Supervisor is enabled.

I hope its ok if I will leave a TODO for now make Sv*x4 require HBare to get started, and add the separate configs once Hypervisor support is implemented.

model/core/sys_regs.sail
```diff
@@ -797,9 +940,12 @@ private function legalize_sstatus(m : Mstatus, v : xlenbits) -> Mstatus = {
 
 mapping clause csr_name_map = 0x100  <-> "sstatus"
 function clause is_CSR_accessible(0x100, _, _) = currentlyEnabled(Ext_S) // sstatus
-function clause read_CSR(0x100) = lower_mstatus(mstatus).bits[xlen - 1 .. 0]
-function clause write_CSR(0x100, value) = { mstatus = legalize_sstatus(mstatus, value); Ok(lower_mstatus(mstatus).bits[xlen - 1 .. 0]) }
 
+function clause read_CSR(0x100) = if privLevel_is_virtual(cur_privilege) then vsstatus.bits[xlen - 1 .. 0] else lower_mstatus(mstatus).bits[xlen - 1 .. 0]
```
pmundkur: It might be nice to have a helper like
```
function inVirtualPrivilege() -> bool = privLevel_is_virtual(cur_privilege)
```
defined near `in32BitMode()` and use that everywhere instead.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,291 @@
...
+  ]
+}
+
+function legalize_hie(m : Minterrupts, v : xlenbits) -> Minterrupts = {
+  let v = Mk_Minterrupts(v);
+  [m with
+    SGEI = if mideleg[SGEI] == 0b1 then v[SGEI] else 0b0,
+    VSEI = if mideleg[SEI]  == 0b1 then v[VSEI] else 0b0,
+    VSTI = if mideleg[STI]  == 0b1 then v[VSTI] else 0b0,
+    VSSI = if mideleg[SSI]  == 0b1 then v[VSSI] else 0b0,
```
pmundkur: Shouldn't these be
```suggestion
    VSEI = if mideleg[VSEI]  == 0b1 then v[VSEI] else 0b0,
    VSTI = if mideleg[VSTI]  == 0b1 then v[VSTI] else 0b0,
    VSSI = if mideleg[VSSI]  == 0b1 then v[VSSI] else 0b0,
```
"For bits of mideleg that are zero, the _corresponding_ bits in hideleg, hip, and hie are read-only zeros." ?

model/core/sys_regs.sail
pmundkur: Shouldn't `legalize_mi{ep}`  also now handle the `SGEI/VSEI/VSTI/VSSI` bits?   Otherwise the new `lower_mi{ep}_to_{vsi{ep},` functions will always find `0b0` at those bits when lowering.  If this is planned for a different PR, perhaps put a TODO?

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,291 @@
...
+    VSSI = m[VSSI],
+  ]
+}
+
+function lower_mie_to_hie(m : Minterrupts) -> Minterrupts = {
+  [Mk_Minterrupts(zeros()) with
+    SGEI = if mideleg[SGEI] == 0b1 then m[SGEI] else 0b0,
+    VSEI = if mideleg[SEI]  == 0b1 then m[VSEI] else 0b0,
+    VSTI = if mideleg[STI]  == 0b1 then m[VSTI] else 0b0,
+    VSSI = if mideleg[SSI]  == 0b1 then m[VSSI] else 0b0,
```
pmundkur: Shouldn't this be
```suggestion
    VSEI = if mideleg[VSEI]  == 0b1 then m[VSEI] else 0b0,
    VSTI = if mideleg[VSTI]  == 0b1 then m[VSTI] else 0b0,
    VSSI = if mideleg[VSSI]  == 0b1 then m[VSSI] else 0b0,
```

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,291 @@
...
+  ]
+}
+
+function lower_mip_to_hip(m : Minterrupts) -> Minterrupts = {
+  [Mk_Minterrupts(zeros()) with
+    SGEI = if mideleg[SGEI] == 0b1 then bool_to_bit((hgeip & hgeie) != zeros()) else 0b0,
+    VSEI = if mideleg[SEI]  == 0b1 then m[VSEI] else 0b0,
+    VSTI = if mideleg[STI]  == 0b1 then m[VSTI] else 0b0,
+    VSSI = if mideleg[SSI]  == 0b1 then m[VSSI] else 0b0,
```
pmundkur: Similarly, here
```suggestion
    VSEI = if mideleg[VSEI]  == 0b1 then m[VSEI] else 0b0,
    VSTI = if mideleg[VSTI]  == 0b1 then m[VSTI] else 0b0,
    VSSI = if mideleg[VSSI]  == 0b1 then m[VSSI] else 0b0,
```

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,291 @@
...
+
+function lower_mip_to_hvip(m : Minterrupts) -> Minterrupts = {
+  [Mk_Minterrupts(zeros()) with
+    VSEI = m[VSEI],
+    VSTI = m[VSTI],
+    VSSI = m[VSSI],
+  ]
+}
+
+function lower_mie_to_hie(m : Minterrupts) -> Minterrupts = {
```
pmundkur: Most of the lowering functions in `sys_regs` provide the delegator register as an argument.  It would be good to follow that convention in this file too.
```suggestion
function lower_mie_to_hie(m : Minterrupts, d : Minterrupts) -> Minterrupts = {
```

model/core/hypervisor_regs.sail
pmundkur: Hmm, those `mideleg` bits are read-only 0b1 for `Ext_H`, so just unconditionally copy those bits?

model/core/hypervisor_regs.sail
radimkrcmar: Sounds good, I wouldn't complicate our implementation too much to support corner-cases.

(Technically, the supported vsatp modes can differ from supported satp modes, yet I don't expect any sensible implementation will make use of that freedom.)

model/core/hypervisor_regs.sail
radimkrcmar: I think it would be clearer to implement the whole logic as
```
Mk_Minterrupts(v & mideleg.bits)
```

nadime15: Yeah I just copy these bits unconditionally now. 

`Mk_Minterrupts(v & mideleg.bits)`

I think this would only work for legalize_hideleg right? Not the other ones (hie, hip)

pmundkur: I think `&` with `mideleg` is necessary for `hie` and `hip` but not sufficient.  It could be done as a last step, maybe.

model/core/hypervisor_regs.sail
nadime15: Hmm, yeah, because the delegator register is used in those cases. I could do this, but then I would have to append an underscore just to mark them as unused. Dont you think its better to keep it as it is?

model/core/sys_regs.sail
nadime15: I think it was fine because we call the legalize_v* functions first to set mip/mie, followed by the lowering function. But you are right, without adding these to the M-mode counterpart registers, M-mode couldnt update them by writing directly to mip/mie. I have made that change!

model/core/hypervisor_regs.sail
pmundkur: Yes, it's okay now.  The comment applied to your previous code where you checked `mideleg` when copying the bits.  Perhaps comment here that `mideleg` is not being checked since those bits of `mideleg` are always enabled for H.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,291 @@
...
+function clause write_CSR(0x607, value) = { hgeie = legalize_hgeie(value); Ok(hgeie) }
+
+// Hypervisor Interrupt Registers
+
+function legalize_hvip(m : Minterrupts, v : xlenbits) -> Minterrupts = {
+  let v = Mk_Minterrupts(v);
+  [m with
+    VSEI = v[VSEI],
+    VSTI = v[VSTI],
+    VSSI = v[VSSI],
```
radimkrcmar: I think only VSSI is an alias to mip state.  The other bits (VSTI/VSEI) actually seem stored in hvip, since the read-only hip (hence mip) bit is a logical `or` of hvip and vstimecmp/hgeip.

model/core/hypervisor_regs.sail
radimkrcmar: Good point, the sie/sip and hie/hip bits are mutually exclusive views into the mie/mip register.
hideleg also needs a bit of filtering, as the bits that have a defined virtual interrupt (SEI, STI, SSI) and SGEI are read-only 0, even when 1 in mideleg.  The ISA design does lend itself to defining bitmasks for the subsets.

model/core/hypervisor_regs.sail
trdthg: I just noticed that the spec said 
- "VSSIP in hip is an alias (writable) of the same bit in hvip."
- "Bits SGEIP, VSEIP, VSTIP, and VSSIP in mip are aliases for the same bits in hypervisor CSR hip"

so we get: mip[VSSIP] is an alias of hip[VSSIP], hip[VSSIP] is an alias of hvip[VSSIP]

Does this imply that `VSSI` is also stored in `hvip`?

nadime15: So my understanding now is that `hvip.VSSI = mip.VSSI = hip.VSSI`, so I think its enough to store VSSI only in mip and other registers are views on mip. 

I have pushed a new commit (https://github.com/riscv/sail-riscv/pull/1110/commits/f038fa04b1777cbd6d18bd341f7c4aa67823f52a) that I think addresses your feedback, @radimkrcmar. Let me know if it does not.

For reads, I make sure that hvip.VSSI stays in sync with mip.VSSI via `lower_mip_hvip()`. For writes, mip.VSSI is updated whenever hvip.VSSI is written.

nadime15:
@radimkrcmar @pmundkur Is anything missing? I think this is now in an acceptable state. It would be good to get this merged, since basically every other part of the hypervisor spec depends on these CSRs.

model/exceptions/sys_exceptions.sail
```diff
@@ -54,7 +54,7 @@ function set_xepc(p : Privilege, value : xlenbits) -> xlenbits = {
     Supervisor        => sepc = target,
     User              => internal_error(__FILE__, __LINE__, "Invalid privilege level"),
     VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
```
pmundkur: I think this could be 
```suggestion
    VirtualUser       => internal_error(__FILE__, __LINE__, "Invalid privilege level"),
```
and similarly for `get_xepc` above.  We could also change this later when updating trap handling.

model/core/sys_regs.sail
```diff
@@ -482,19 +536,34 @@ private function legalize_mip(o : Minterrupts, v : xlenbits) -> Minterrupts = {
...
+    // NOTE/TODO: "VSEIP is read-only in hip, and is the logical-OR of these interrupt sources.
+    // * bit VSEIP of hvip;"
+    // * the bit of hgeip selected by hstatus.VGEI; and (TODO)
+    // * any other platform-specific external interrupt signal directed to VS-Level (TODO)"
+    VSEI = if currentlyEnabled(Ext_H) then hvip[VSEI] else 0b0,
+    // NOTE/TODO: "VSTIP is read-only in hip, and is the logical-OR of
+    // * hvip.VSTIP; and
+    // * any other platform-specfic timer interrupt signal directed to VS-Level (TODO)"
+    VSTI = if currentlyEnabled(Ext_H) then hvip[VSTI] else 0b0,
+    VSSI = if currentlyEnabled(Ext_H) then v[VSSI] else 0b0,
```
pmundkur: Since `SGEI, VS{EI,TI}` in `mip` are aliases of the corresponding bits in `hip`, and those bits are read-only, shouldn't this be
```suggestion
    SGEI = o[SGEI],
    VSEI = o[VSEI],
    VSTI = o[VSTI],
    VSSI = if currentlyEnabled(Ext_H) then v[VSSI] else 0b0,
```
i.e. there should be no modification of those bits?

pmundkur (commented):
Looking good! I'm still going through it, some comments so far...

model/core/sys_regs.sail
nadime15: But these values must be computed and stored somewhere (I guess you are implying that the computation of these bitfields would be moved into legalize_hip?). As you said, hip (and hvip, only VSSI) are views of mip, and I think thats why mip is the natural place to compute and store these bits.

Its more about making sure we dont have any stale values. hgeip and hgeie are not views of mip. So if I move the computation out of mip into some other location, I need to ensure that every time mip or hip is read, the derived values (SGEI, VSEI, etc.) are up to date because they are actually set by other bitfiels (see the deleted comment). Imagine hgeip changes at runtime, this directly affects mip.SGEI. If code running in m-mode then reads mip.SGEI, it could potentially observe a stale value. 

Same goes for mip.VSTI, which depends on hvip.VSTI (which is a unique bit of hvip and [not a view](https://github.com/riscv/sail-riscv/pull/1110#discussion_r2754184791)), I would need to make sure mip.VSTI has no stale values (which could happen if you apply your change and then write to hvip without making sure mip.VSTI is actually set to the new value of hvip.VSTI)

radimkrcmar: Right, mip.VSTI and mip.VSEI do not need any physical storage -- they are a function of hvip and vstimecmp/hgeip.  (We only need a single xlen-bit status to store all the bits required to compute mip/sip/vsip/hip/hvip, yet that storage is neither of those CSRs.)

I agree it's possible to cache the computed values in mip, but I don't think the current code is doing good job in keeping it updated.  (Caching is hard without a proper framework.)
For example read_csr(hip/mip) seems buggy, since writing hvip doesn't re-legalize mip, and neither does reading mip nor hip, so a modification of hvip doesn't propagate to mip or hip.

I share @pmundkur's worry as well, since mip.STI is a function of stimecmp, yet we don't do anything about it in legalize_mip, which makes the handling of VSTI seemingly inconsistent.
I think it would actually be nice to implement mip.VSTI with vstimecmp to clarify how mip is supposed to be updated outside of explicit CSR accesses.

nadime15: > I agree it's possible to cache the computed values in mip, but I don't think the current code is doing good job in keeping it updated. (Caching is hard without a proper framework.)
> For example read_csr(hip/mip) seems buggy, since writing hvip doesn't re-legalize mip, and neither does reading mip nor hip, so a modification of hvip doesn't propagate to mip or hip.

Yeah, I agree, but I think this is a general problem we have in the model, see the discussion in #1550. 
The problem is we basically never legalize CSRs before we read them so we have stale values and thats why I said we  should change all CSR reads and legalize the corresponding register beforehand.
We also discussed this on Monday as part of the discussion about whether we can have implicit setters and getters that are called on every register access.

In general, I think we should avoid accessing registers directly (even internally), or at least legalize them beforehand.

So I am not sure whether it’s better to just legalize mip every time we read hip/hvip/mip/sip, etc. 

I think its harder to modify all these bits that are part of mip if one of the conditions changes. Because this means now, I have to legalize_mip everytime hgeip/hgeie/hvip/hstatus etc. has changed, which I think introduces even more changes (now I have to do the same computation X times instead of one time in mip) and just shifts the problem somewhere else? 

> I think it would actually be nice to implement mip.VSTI with vstimecmp to clarify how mip is supposed to be updated outside of explicit CSR accesses.

vstimecmp is part of the Sstc extension, which is why I did not include it, since this PR was meant to focus only on the Hypervisor chapter.

pmundkur: I think we had discussed legalizing on reads as being more flexible.  One way would be to make the affected CSR registers private and provide read accessors (as was done for `senvcfg` in #1408).  

The logical `OR` involved in some of these bits is a read-time phenomenon; see the discussion of the `OR` for mip.SEIP 
https://docs.riscv.org/reference/isa/priv/machine.html#3-1-9-machine-interrupt-mip-and-mie-registers
so it doesn't make sense to do it in the legalizer called when mip is being written.  Changing read-only bits in a register when that register is written to, for whatever reason, is confusing and most likely wrong.

One path forward could be to factor out a separate PR that makes `mip`, `mie` private and provides read accessors for them, and then build on top of that in this PR.  We could also split out the interrupt related CSRs, which is where a lot of the complexity lies, and put the other simpler CSRs in a different separate PR which might be easier to review and merge.

nadime15: > Changing read-only bits when a register is written to, for whatever reason, is confusing and most likely wrong.

These bits are read-only from software's perspective (you can't write them via CSR instructions), but internally they still need to be updated somehow, what you do internally is completely up to the implementation.

Besides that, the spec talks from the perspective of a hardware implementation. Obviously in hardware you would just compute this via OR on a read, but in our software model we are grouping legalization logic into legalize methods which are responsible for keeping bits within a given CSR up to date.

The `read_senvcfg` accessor approach scatters legalization logic across two functions, some logic lives in `legalize_senvcfg` and some in read_senvcfg (in fact its actually just a more complicated wrapper around `legalize_senvcfg()`). A cleaner alternative would be to call legalize_senvcfg on both reads and writes (read_CSR()/write_CSR() will just call legalize_CSR()), keeping all the logic centralized in one function. This way you can actually keep the legalization logic in one place, and dont have to differentiate between reads and writes.

I also think the alternative [proposal](https://github.com/riscv/sail-riscv/pull/1110#discussion_r2825107009) is actually more confusing: for some reason VSSI is gated by a hypervisor extension check while SGEI, VSEI and VSTI are simply o[SGEI] etc. which raises the question of how and where those values are ever actually updated. This doesn't solve the staleness problem, it just hides it by silently preserving old values without any explanation of how they got there in the first place.

pmundkur: There is a difference between read legalization (where you are constructing bits of a register value, or dispatching machine operations, by reading _only_ existing machine state), and write legalization (where you are reading an _arbitrary_ user-provided register value and updating bits in the machine state).  Conflating the two into one legalization function might not make matters simpler, and instead likely do the opposite.  I agree that when it is possible, it would be better to have one legalizer.  But I remain to be convinced that what is planned for `legalize_mip` currently (indicated by the embedded comments in it) is a reasonable approach.

The difference between VSSI and {SGEI, VSEI, VSTI} is that the former is an alias of a _writable_ bit, whereas the latter are aliases of _read-only_ bits.  The reason for gating of VSSI on hypervisor is that that bit is _inactive_ in the absense of hypervisor support. This could be spelled out in a comment to be less confusing.

Note that 
```
[o with
   ...
   SGEI = o[SGEI],
   VSEI = o[VSEI],
   VSTI = o[VSTI],
   VSSI = if currentlyEnabled(Ext_H) then v[VSSI] else 0b0,
]
```
simplifies to
```
[o with
   ...
   VSSI = if currentlyEnabled(Ext_H) then v[VSSI] else 0b0,
]
```

nadime15: I don't agree here:

> There is a difference between read legalization (where you are constructing bits of a register value, or dispatching machine operations, by reading only existing machine state), and write legalization (where you are reading an arbitrary user-provided register value and updating bits in the machine state). Conflating the two into one legalization function might not make matters simpler, and instead likely do the opposite.

There is no fundamental difference between read and write legalization, what matters is simply that when you access a CSR, you get a fresh value that reflects the current state of the machine. Whether you achieve that by updating mip on every write to hvip/hgeip/hgeie, or by computing the value fresh on every read, is an implementation detail. Both approaches can be correct (one is just harder to implement than the other).

nadime15: Additionally, I don't understand whats so hard about just calling legalize_CSR for each write and read (thats what I proposed to do)

Timmmm: > There is no fundamental difference between read and write legalization

From a "what can software rely on" point of view there is no difference. From a "what can software observe" point of view there is definitely a difference. So if the model is being used for lockstep verification it does matter.

(Sorry I haven't read this full discussion!)

nadime15: Your point is valid if the internal register state is directly observable by a lockstep verifier, (but that is generally not the case) mip can only be observed through CSR accesses, at which point both approaches produce the correct value. So there is no observable difference from a lockstep verification perspective. 

Additionally, if a lockstep verifier did have access to internal register state, it would fail against any implementation that doesn't operate in exactly the same like our model does, out-of-order cores, implementations with different optimizations, etc. That would make the verifier essentially useless for anything other than a single reference implementation, which is probably not the intent.

allenjbaum:
Lockstep, as implemented by HMC, used an RVVI interface, which is able to
observe internal (architected) state.

model/core/sys_regs.sail
Timmmm: The internal register state can be observed in some case, where a field's legal values are not constant. For example in #1539 - `mstatus[MPP]`'s legal values depend on `misa[S]`. If you:

1. Write `misa[S]` = 0
2. Write `mstatus[MPP]` = 0b01 (supervisor IIRC)
3. Write `misa[S]` = 1
4. Read `mstatus[MPP]`

You would read different values depending on whether it was legalised on read or write. (I'm making no judgement about which is better here! Just saying there is an observable difference.)

nadime15: Oh yeah I see, this got more complicated than I expected! You are right that this would make an observable difference, even though the spec says the case is unspecified.

To be honest, I am not sure what to think about this. Currently the model legalizes on write, but #1408 would introduce read time legalization, which means we end end up mixing both approaches, and I believe we have to introduce read time legalization anyway and would start mixing both. So I think in this regard and from a verification point of view the model would become inconsistent (unless we introduce some configs to determine the behavior).

My original assumption was that we could use the same legalization function for both reads and writes:
```
write_CSR(CSR, value) = legalize_CSR(CSR, value)
read_CSR(CSR) = legalize_CSR(CSR, CSR)
```

I assumed that for reads, **where no external state has changed**, legalize_CSR(CSR, CSR) == CSR would always hold (this is important to make sure that in the read time legalization case we don't alter any CSR bits by passing the CSR as the "value" variable to the function), since a previously legalized value passed through again should return the same result. The only counterexample I could think of was a field like FIELD = not(FIELD), but that doesn't exist (I believe). And as far as I can tell, all CSRs in this model are effectively idempotent in this regard, so legalize_CSR(CSR, CSR) == CSR should hold as long as no external state has changed, which would mean that legalize_CSR(CSR, CSR) != CSR can only happen due to a change of an external state and because of that some legalization was actually applied.

Timmmm: Yeah it's really awkward. I *think* your idea of legalising on read and write works... but I haven't thought about it too much.

Slightly off topic, but it would be really handy to have some kind of "RISC-V garden" where someone has hooked up 100 real RISC-V devices to one machine and you can execute an ELF on all of them simultaneously to see what the range of real-world behaviour is. Probably would be very complicated though especially for machine mode code.

jordancarlin:

PR #1113, opened 2025-06-30
Use integers for access sizes instead of an enum

Timmmm:
For instructions that can access power-of-two sizes (1, 2, 4, 8 bytes), switch to using integers for the sizes instead of an enum. This will allow more elegant support for 16-byte accesses in future, and is slightly nicer already.

There are more possible improvements:

1. Merge some of the repeated encdec clauses.
2. Use power of two instead of bytes, i.e. `range(0, 3)` instead of `{1, 2, 4, 8}`.

However this keeps the change fairly minimal.

### DIFF
--- a/model/riscv_insts_zaamo.sail
+++ b/model/riscv_insts_zaamo.sail
@@ -14,10 +14,10 @@ function clause currentlyEnabled(Ext_Zabha) = hartSupports(Ext_Zabha) & currentl
 // Zabha defines AMOs for byte and halfword
 function amo_width_valid(size : word_width) -> bool = {
   match size {
-    BYTE   => currentlyEnabled(Ext_Zabha),
-    HALF   => currentlyEnabled(Ext_Zabha),
-    WORD   => true,
-    DOUBLE => xlen >= 64,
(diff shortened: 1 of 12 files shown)


pmundkur (approved):
Nice improvement!

arichardson (approved):

model/riscv_insts_zaamo.sail
```diff
@@ -14,10 +14,10 @@ function clause currentlyEnabled(Ext_Zabha) = hartSupports(Ext_Zabha) & currentl
 // Zabha defines AMOs for byte and halfword
 function amo_width_valid(size : word_width) -> bool = {
   match size {
```
jordancarlin, line 16: Maybe `match bytes` is a better fit for this now?

jordancarlin (approved):

model/riscv_insts_zaamo.sail
Timmmm: Yeah the names are a bit all over the place now. I made #1116 so we don't forget.

PR #1114, opened 2025-06-30
Add config flags to control PMP address mode support

### LINKED ISSUE #1109: Missing configuration for PMP support in sail model
Hello, maintainers

As discussed in https://github.com/riscv-non-isa/riscv-arch-test/issues/665, some PMP implementation may only support NAPOT mode, without support for NA4 and TOR mode, but this is not addressed in sail model, could you add some configuration to config it, so riscv architecture test can use the configuration to ignore cases which dont have NA4 or TOR mode.

Thanks

Timmmm:
This adds config flags to allow marking PMP address matching modes as not supported. If software tries to write an unsupported mode it will get set to OFF instead.

For NA4 mode it is required that the grain G is 0. If it isn't then that mode is unsupported. This is also checked in `config_is_valid()` which is redundant... but I think that's ok for now. Another option would be to print a warning like `Warning: NA4 is marked as supported but G > 0 so it has been disabled`.

Fixes #1109

### DIFF
--- a/model/riscv_pmp_regs.sail
+++ b/model/riscv_pmp_regs.sail
@@ -19,24 +19,11 @@ let sys_pmp_grain : range(0, 63) = config memory.pmp.grain
 
 enum PmpAddrMatchType = {OFF, TOR, NA4, NAPOT}
 
-val pmpAddrMatchType_of_bits : bits(2) -> PmpAddrMatchType
-function pmpAddrMatchType_of_bits(bs) = {
-  match bs {
-    0b00 => OFF,
-    0b01 => TOR,
-    0b10 => NA4,
-    0b11 => NAPOT
-  }
-}
-
-val pmpAddrMatchType_to_bits : PmpAddrMatchType -> bits(2)
-function pmpAddrMatchType_to_bits(bs) = {
-  match bs {
-    OFF   => 0b00,
-    TOR   => 0b01,
-    NA4   => 0b10,
-    NAPOT => 0b11
-  }
+mapping pmpAddrMatchType_encdec : PmpAddrMatchType <-> bits(2) = {
+  OFF   <-> 0b00,
+  TOR   <-> 0b01,
+  NA4   <-> 0b10,
+  NAPOT <-> 0b11,
 }
 
 bitfield Pmpcfg_ent : bits(8) = {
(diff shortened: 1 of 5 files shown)


model/riscv_pmp_regs.sail
```diff
@@ -115,12 +102,21 @@ function pmpWriteCfg(n: range(0, 63), cfg: Pmpcfg_ent, v: bits(8)) -> Pmpcfg_ent
...
+    // is a WARL field.
+    let mode_supported : bool = match pmpAddrMatchType_encdec(cfg[A]) {
+      OFF   => true,
+      TOR   => config memory.pmp.tor_supported,
+      // "When G >= 1, the NA4 mode is not selectable."
+      NA4   => (config memory.pmp.na4_supported : bool) & sys_pmp_grain == 0,
+      NAPOT => config memory.pmp.napot_supported,
+    };
+
+    // In this implementation we set the mode to OFF if an unsupported mode is selected.
```
jordancarlin, line 115: Probably worth eventually adding an option to keep the current mode if an unsupported mode is selected. I think that plus the current version of defaulting to OFF should cover almost all use cases.

jordancarlin (approved):

model/riscv_pmp_regs.sail
Timmmm: Yeah, I think we can wait until someone asks for it. I think ignoring the write would be a kind of bad design so hopefully nobody will actually do that!

allenjbaum:
I would prefer if Sail defaulted to "keep unchanged" for all writes to WARL
fields* with illegal configurations.
That is the easiest to implement and to test, and always works (writing
zero may actually be illegal for some fields).

* (with the single exception of 2 bit fields that are implemented with a
single flipflop (so only 00 or 11 are legal).
In that case, the implementation has to choose to write the value that is
in the MSB or the LSB of the field.

Timmmm:
Since this legalisation isn't generic I don't think we need to worry about it not working for other WARL fields. Leaving PMP unchanged isn't very fail-safe, so I'd prefer not to do that in case people think it is somehow the recommended behaviour.

allenjbaum:
There is nothing that is failsafe. Code that write WARL fields either must
know apriori what values are legal (e.g. through discovery)
or they need to check that the write resulting in the desired value (which
is a form of discovery) , or don't check and take their chances.
And if  there is a bug, all zeroes may not be the safe option  - it might
be be the most unsafe option..

Timmmm:
>  or don't check and take their chances.

Yeah this is the case where it may make a difference - software that assumes a mode is supported and forgets to check that the write was successful. E.g. [I don't think OpenSBI reads the value back](https://github.com/riscv-software-src/opensbi/blob/a32a91069119e7a5aa31e6bc51d5e00860be3d80/lib/sbi/riscv_asm.c#L349), at least not automatically.

> And if there is a bug, all zeroes may not be the safe option  - it might be be the most unsafe option..

Yeah you could have a config where pmp0 denies access to some sensitive memory and pmp1 allows access to all memory. Then software tries to write pmp0 with some unsupported mode, disabling it and leaving all memory exposed. But making that fail-safe would be impossible anyway I think.

allenjbaum:
Well, leaving the configuration unchanged seems quite a bit less bad, so
(again) I would strongly urge that a default action should be simply to
ignore the write
 (there is at least one case where that's the required action - SATP mode
writes, I think), and it is also the easiest to implement.
 At some point SAIL will likely require overrides of default WARL behavior,
and this will minimize the need for that.

PR #1117, opened 2025-07-01
Rationalize the location of vector CSRs and their handling.

pmundkur:
Move vector CSRs like `vtype`, `vstart`, `vl` from `sys_regs` to `vext_regs`, where `vcsr` was already located.  This avoids the vector CSRs being split across two files.

Move `{is_CSR_defined,read_CSR,write_CSR}` from `vext_control` to `vext_regs` to be consistent with the current `sys_regs,sys_control` split.

Move other `vreg` accessor utils from `vext_regs` to `vext_control`.

This will also help documentation.


pmundkur:
I'm not sure `_control` is a good suffix here, but we can rename files later.

Timmmm (approved):

PR #1120, opened 2025-07-02
Merge float load/store encdec clauses

Timmmm:
Use `size_enc` to deduplicate these.

### DIFF
--- a/model/riscv_insts_fext.sail
+++ b/model/riscv_insts_fext.sail
@@ -264,6 +264,14 @@ function fle_S   (v1,       v2,        is_quiet) = {
 
 function haveSingleFPU() -> bool = currentlyEnabled(Ext_F) | currentlyEnabled(Ext_Zfinx)
 
+function float_load_store_width_supported(width : word_width) -> bool =
+  match width {
+    1 => false,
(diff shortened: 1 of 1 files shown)


model/riscv_insts_fext.sail
```diff
@@ -264,6 +264,14 @@ function fle_S   (v1,       v2,        is_quiet) = {
 
 function haveSingleFPU() -> bool = currentlyEnabled(Ext_F) | currentlyEnabled(Ext_Zfinx)
 
+function float_load_store_width_supported(width : word_width) -> bool =
```
Timmmm, line 267: If anyone can think of a better name for this function...!

jordancarlin: Maybe just `float_width_supported`, but that might be confusing because of the Zfinx type extensions.

jordancarlin (approved):
Nice simplification.

model/riscv_insts_fext.sail
Timmmm: Yeah... that's what I was thinking. Oh well, this will do for now.

PR #1122, opened 2025-07-03
Add support for local counter-overflow interrupts

trdthg:
If the Sscofpmf extension is not implemented, xip.LCOFIP and xie.LCOFIE are read-only zeros.

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -441,13 +441,15 @@ function is_fiom_active() -> bool = {
 /* interrupt processing state */
 
 bitfield Minterrupts : xlenbits = {
-  MEI : 11, /* external interrupts */
+  LCOFI : 13, // local counter overflow interrupts
+
+  MEI : 11, // external interrupts
   SEI : 9,
 
-  MTI : 7,  /* timers interrupts */
+  MTI : 7,  // timers interrupts
   STI : 5,
 
-  MSI : 3,  /* software interrupts */
+  MSI : 3,  // software interrupts
   SSI : 1,
 }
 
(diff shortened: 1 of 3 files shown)


model/riscv_sys_regs.sail
```diff
@@ -441,6 +441,8 @@ function is_fiom_active() -> bool = {
 /* interrupt processing state */
 
 bitfield Minterrupts : xlenbits = {
+  LCOFIP : 13, /* local count overflow interrupt */
```
jordancarlin: This should be `LCOFI` to match the style of the other interrupts since the same bitfield is used for both `mip` (where it is `LCOFIP`) and `mie` (where it is `LCOFIE`). Also might as well switch to `//` style comments while you're editing this.

jordancarlin (changes requested):
This also needs to be handled for S-mode:
> If the Sscofpmf extension is implemented, bits sip.LCOFIP and sie.LCOFIE are the interrupt-pending and interrupt-enable bits for local-counter-overflow interrupts. LCOFIP is read-write in sip and reflects the occurrence of a local counter-overflow overflow interrupt request resulting from any of the mhpmeventn.OF bits being set. If the Sscofpmf extension is not implemented, sip.LCOFIP and sie.LCOFIE are read-only zeros.

This interrupt also needs to be added to `findPendingInterrupt` so that it is actually handled. I guess that could be a separate PR, but I think it probably makes more sense to fully implement the interrupt in a single PR. Otherwise we get into a strange state where the bit is writable but it doesn't behave as expected.

trdthg:
Why lift_sip only check SSI here?

```
function lift_sip(o : Minterrupts, d : Minterrupts, s : Sinterrupts) -> Minterrupts = {
  let m : Minterrupts = o;
  let m = if d[SSI] == 0b1 then [m with SSI = s[SSI]] else m;
  m
}
```

jordancarlin:
Isn’t the code to actually trigger the interrupt on a counter overflow still missing?

trdthg:
According to this (from section 20.1)

> The OF bit is set when the corresponding hpmcounter overflows
> ...
> Count overflow never results from writes to the mhpmcountern or mhpmeventn registers, only from hardware increments of counter registers.

This seems to imply that `writeCSR` will never cause an overflow, checking overflow here is useless.

And the model doesn't actually increment any of the HPM counters yet (for example, when simulating branch prediction, and it failed), we have no place to check and trigger this interrupt.

Perhaps we need a `writeCSRByHardware` to distinguish it from the `writeCSR` called by software in order to achieve this.

(I've only recently taken a close look at the hpmcounter stuff. My initial intention was just to add the LCOFI bit. 😂)

PR #1123, opened 2025-07-03
Build libgmp as position independent code

Timmmm:
When building libgmp from source, enable position independent code. This isn't strictly necessary right now, but if you want to build the Sail model as a shared library then you do need this, and in general position independent code causes fewer annoying linking problems.


jordancarlin (approved):
We really need to get #745 sorted out, but LGTM.

pmundkur (approved):

PR #1124, opened 2025-07-03
Allow `write_CSR` to indicate an illegal instruction trap.

pmundkur:
This enables data-dependent illegal instruction traps on CSR writes, such as on out-of-bounds values for vstart.

Indicating other kinds of traps would need a bigger refactor, but this should suffice for now.


jordancarlin (approved):
I don’t love wrapping everything in `Ok`, but this does seem like the best approach for now.

PR #1128, opened 2025-07-07
Fix CSR trace output to print correct width based on XLEN

### LINKED ISSUE #1094: CSR logged as 64 bits when running RV32
When running in RV32, all CSRs seems to be logged with 64 bits.
`mem[X,0x080000C68] -> 0xF457`
`mem[X,0x080000C6A] -> 0xC1B0`
`[2607] [M]: 0x80000C68 (0xC1B0F457) vsetivli x8, 0x1, e64, m8, tu, mu`
`x8 <- 0x00000001`
`CSR mstatus (0x300) <- 0x0000000080000680`
`CSR mstatush (0x310) <- 0x0000000000000000`
`CSR vstart (0x008) <- 0x0000000000000000`
`CSR vtype (0xC21) <- 0x000000000000001B`
`CSR vl (0xC20) <- 0x0000000000000001`
`CSR vstart (0x008) <- 0x0000000000000000`

nadime15:
This fixes #1094, where CSR values were incorrectly logged as 64-bit when running in RV32 mode.

### DIFF
--- a/c_emulator/riscv_callbacks.cpp
+++ b/c_emulator/riscv_callbacks.cpp
@@ -96,8 +96,8 @@ unit csr_full_write_callback(const_sail_string csr_name, unsigned reg,
                              sbits value)
 {
   if (config_print_reg) {
-    fprintf(trace_log, "CSR %s (0x%03X) <- 0x%016" PRIX64 "\n", csr_name, reg,
-            value.bits);
+    fprintf(trace_log, "CSR %s (0x%03X) <- 0x%0*" PRIX64 "\n", csr_name, reg,
+            static_cast<int>(zxlen / 4), value.bits);
   }
   return UNIT;
 }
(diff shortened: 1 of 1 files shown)


nadime15:
As far as I can tell, we dont need to worry about CSRs that are always defined as 64 bits wide while xlen=32, because in RV32 we only ever read and write the lower 32 bits, as accessing the full 64 bits is not allowed by the spec.

> 
> 2.7. Explicit Accesses to CSRs Wider than XLEN
> If a standard CSR is wider than XLEN bits, then an explicit read of the CSR returns the register’s least-significant XLEN bits, and an explicit write to the CSR modifies only the register’s least-significant XLEN bits, leaving the upper bits unchanged.

The only downside of this change is that 32-bit registers are now printed as 64-bit values when XLEN is 64. Im not sure whether people prefer a mixed-width format, where accessing a 32-bit CSR prints 32 bits and accessing a 64-bit CSR prints 64 bits, even when XLEN is 64.

c_emulator/riscv_callbacks.cpp
```diff
@@ -96,8 +96,8 @@ unit csr_full_write_callback(const_sail_string csr_name, unsigned reg,
                              sbits value)
 {
   if (config_print_reg) {
-    fprintf(trace_log, "CSR %s (0x%03X) <- 0x%016" PRIX64 "\n", csr_name, reg,
-            value.bits);
+    fprintf(trace_log, "CSR %s (0x%03X) <- 0x%0*" PRIX64 "\n", csr_name, reg,
+            static_cast<int>(zxlen / 4), value.bits);
```
Timmmm: Can we use `value.length` or whatever it is instead and stop peeking at `zxlen`/`zflen`?

jordancarlin (approved):

PR #1129, opened 2025-07-07
Improve command-line argument parsing and help formatting

nadime15:
I tried to improve the help text a bit and make it clearer which options have required or optional arguments. Some of the changes include:

- Show error message rather than help text when argument is missing
- Replace atoi() with strtol() for signature-granularity and rvfi-dii port validation
- Improve help output formatting with argument indicators
- Add descripion for each option showing required/optional arguments

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -105,7 +105,7 @@ char *sailcov_file = NULL;
 static struct option options[] = {
     {"device-tree-blob",               required_argument, 0, 'b'                },
     {"terminal-log",                   required_argument, 0, 't'                },
-    {"show-times",                     required_argument, 0, 'p'                },
+    {"show-times",                     no_argument,       0, 'p'                },
     {"test-signature",                 required_argument, 0, 'T'                },
     {"signature-granularity",          required_argument, 0, 'g'                },
     {"rvfi-dii",                       required_argument, 0, 'r'                },
@@ -130,13 +130,58 @@ static struct option options[] = {
 static void print_usage(const char *argv0, int ec)
 {
   fprintf(stdout, "Usage: %s [options] <elf_file> [<elf_file> ...]\n", argv0);
-  fprintf(stdout, "       %s [options] -r <port>\n", argv0);
+  fprintf(stdout, "\nOptions:\n");
+
   struct option *opt = options;
   while (opt->name) {
-    if (isprint(opt->val))
-      fprintf(stdout, "\t -%c\t --%s\n", (char)opt->val, opt->name);
+    if (isprint(opt->val)) {
+      fprintf(stdout, "  -%c, --%-30s", (char)opt->val, opt->name);
+    } else {
+      fprintf(stdout, "      --%-30s", opt->name);
+    }
+
+    if (opt->val == 'b')
+      fprintf(stdout, " <file>     Device tree blob file");
+    else if (opt->val == 't')
+      fprintf(stdout, " <file>     Terminal log output file");
+    else if (opt->val == 'p')
+      fprintf(stdout, "            Show execution times");
+    else if (opt->val == 'T')
+      fprintf(stdout, " <file>     Test signature file");
+    else if (opt->val == 'g')
+      fprintf(stdout, " <size>     Signature granularity");
+    else if (opt->val == 'r')
+      fprintf(stdout, " <port>     RVFI DII port");
+    else if (opt->val == 'h')
+      fprintf(stdout, "            Show this help message");
+    else if (opt->val == 'c')
+      fprintf(stdout, " <file>     Configuration file");
+    else if (opt->val == 'v')
+      fprintf(stdout, " [file]     Enable tracing");
+    else if (opt->val == 'V')
+      fprintf(stdout, " [file]     Disable tracing");
+    else if (opt->val == 'l')
+      fprintf(stdout, " <count>    Instruction limit");
+    else if (strcmp(opt->name, "print-default-config") == 0)
+      fprintf(stdout, "            Print default configuration");
+    else if (strcmp(opt->name, "validate-config") == 0)
+      fprintf(stdout, "            Validate configuration");
+    else if (strcmp(opt->name, "trace-output") == 0)
+      fprintf(stdout, " <file>     Trace output file");
+    else if (strcmp(opt->name, "enable-experimental-extensions") == 0)
+      fprintf(stdout, "            Enable experimental extensions");
+    else if (strcmp(opt->name, "print-device-tree") == 0)
+      fprintf(stdout, "            Print device tree");
+    else if (strcmp(opt->name, "print-isa-string") == 0)
+      fprintf(stdout, "            Print ISA string");
+#ifdef SAILCOV
+    else if (strcmp(opt->name, "sailcov-file") == 0)
+      fprintf(stdout, " <file>     Sailcov output file");
+#endif
     else
-      fprintf(stdout, "\t   \t --%s\n", opt->name);
+      fprintf(stdout, "Unknown argument");
+
+    fprintf(stdout, "\n");
     opt++;
   }
   exit(ec);
@@ -251,11 +296,20 @@ static int process_args(int argc, char **argv)
       sig_file = strdup(optarg);
       fprintf(stderr, "using %s for test-signature output.\n", sig_file);
       break;
-    case 'g':
-      signature_granularity = atoi(optarg);
+    case 'g': {
+      char *p;
+      long val = strtol(optarg, &p, 10);
+      if (*p != '\0' || val <= 0 || val > INT_MAX) {
+        fprintf(stderr,
+                "Error: signature-granularity must be a positive number, got "
+                "'%s'\n",
+                optarg);
+        exit(EXIT_FAILURE);
+      }
+      signature_granularity = static_cast<int>(val);
       fprintf(stderr, "setting signature-granularity to %d bytes\n",
               signature_granularity);
-      break;
+    } break;
     case 'h':
       print_usage(argv[0], EXIT_SUCCESS);
       break;
@@ -283,8 +337,16 @@ static int process_args(int argc, char **argv)
       do_print_isa = true;
       break;
     case 'r': {
+      char *p;
+      long val = strtol(optarg, &p, 10);
+      if (*p != '\0' || val <= 0 || val > 65535) {
+        fprintf(stderr,
+                "Error: port number must be between 1 and 65535, got '%s'\n",
+                optarg);
+        exit(EXIT_FAILURE);
+      }
       config_enable_rvfi = true;
-      int rvfi_dii_port = atoi(optarg);
+      int rvfi_dii_port = static_cast<int>(val);
       rvfi = rvfi_handler(rvfi_dii_port);
       break;
     }
@@ -317,11 +379,54 @@ static int process_args(int argc, char **argv)
       break;
 #endif
     case OPT_TRACE_OUTPUT:
+      if (optarg == NULL || strlen(optarg) == 0) {
+        fprintf(stderr,
+                "Error: option '--trace-output' requires an argument <file>\n");
+        fprintf(stderr, "Try '%s --help' for more information.\n", argv[0]);
+        exit(EXIT_FAILURE);
+      }
       trace_log_path = optarg;
       fprintf(stderr, "using %s for trace output.\n", trace_log_path);
       break;
     case '?':
-      print_usage(argv[0], EXIT_FAILURE);
+      if (optopt == 'b') {
+        fprintf(stderr,
+                "Error: option '-%c/--device-tree-blob' requires an argument "
+                "<file>\n",
+                optopt);
+      } else if (optopt == 't') {
+        fprintf(
+            stderr,
+            "Error: option '-%c/--terminal-log' requires an argument <file>\n",
+            optopt);
+      } else if (optopt == 'T') {
+        fprintf(stderr,
+                "Error: option '-%c/--test-signature' requires an argument "
+                "<file>\n",
+                optopt);
+      } else if (optopt == 'g') {
+        fprintf(stderr,
+                "Error: option '-%c/--signature-granularity' requires an "
+                "argument <size>\n",
+                optopt);
+      } else if (optopt == 'r') {
+        fprintf(stderr,
+                "Error: option '-%c/--rvfi-dii' requires an argument <port>\n",
+                optopt);
+      } else if (optopt == 'c') {
+        fprintf(stderr,
+                "Error: option '-%c/--config' requires an argument <file>\n",
+                optopt);
+      } else if (optopt == 'l') {
+        fprintf(
+            stderr,
+            "Error: option '-%c/--inst-limit' requires an argument <count>\n",
+            optopt);
+      } else {
+        fprintf(stderr, "Error: unknown option '-%c'\n", optopt);
+      }
+      fprintf(stderr, "Try '%s --help' for more information.\n", argv[0]);
+      exit(EXIT_FAILURE);
       break;
     }
   }


nadime15:
Thats how the help text looked before:

```
Usage: ./build/c_emulator/sail_riscv_sim [options] <elf_file> [<elf_file> ...]
       ./build/c_emulator/sail_riscv_sim [options] -r <port>
         -b      --device-tree-blob
         -t      --terminal-log
         -p      --show-times
         -T      --test-signature
         -g      --signature-granularity
         -r      --rvfi-dii
         -h      --help
         -c      --config
                 --print-default-config
                 --validate-config
         -v      --trace
         -V      --no-trace
                 --trace-output
         -l      --inst-limit
                 --enable-experimental-extensions
                 --print-device-tree
                 --print-isa-string
```


after:
```

Usage: ./build/c_emulator/sail_riscv_sim [options] <elf_file> [<elf_file> ...]

Options:
  -b, --device-tree-blob               <file>     Device tree blob file
  -t, --terminal-log                   <file>     Terminal log output file
  -p, --show-times                                Show execution times
  -T, --test-signature                 <file>     Test signature file
  -g, --signature-granularity          <size>     Signature granularity
  -r, --rvfi-dii                       <port>     RVFI DII port
  -h, --help                                      Show this help message
  -c, --config                         <file>     Configuration file
      --print-default-config                      Print default configuration
      --validate-config                           Validate configuration
  -v, --trace                          [file]     Enable tracing
  -V, --no-trace                       [file]     Disable tracing
      --trace-output                   <file>     Trace output file
  -l, --inst-limit                     <count>    Instruction limit
      --enable-experimental-extensions            Enable experimental extensions
      --print-device-tree                         Print device tree
      --print-isa-string                          Print ISA string  
```



Error messages look like:

 ```
/build/c_emulator/sail_riscv_sim: option '--inst-limit' requires an argument
 Error: option '-l/--inst-limit' requires an argument <count>
 Try './build/c_emulator/sail_riscv_sim --help' for more information.
```

File: c_emulator/riscv_sim.cpp [hunk 1]
```diff
@@ -130,13 +130,58 @@ static struct option options[] = {
...
+    else if (opt->val == 'g')
+      fprintf(stdout, " <size>     Signature granularity");
+    else if (opt->val == 'r')
+      fprintf(stdout, " <port>     RVFI DII port");
+    else if (opt->val == 'h')
+      fprintf(stdout, "            Show this help message");
+    else if (opt->val == 'c')
+      fprintf(stdout, " <file>     Configuration file");
+    else if (opt->val == 'v')
+      fprintf(stdout, " [file]     Enable tracing");
```
radimkrcmar: The `[file]` is not a file, and has a more complex type that deserves explanation in the help message: `instr|reg|mem|rvfi|platform|step|all`

c_emulator/riscv_sim.cpp
```diff
@@ -130,13 +130,58 @@ static struct option options[] = {
...
+    if (opt->val == 'b')
+      fprintf(stdout, " <file>     Device tree blob file");
+    else if (opt->val == 't')
+      fprintf(stdout, " <file>     Terminal log output file");
+    else if (opt->val == 'p')
+      fprintf(stdout, "            Show execution times");
+    else if (opt->val == 'T')
+      fprintf(stdout, " <file>     Test signature file");
+    else if (opt->val == 'g')
+      fprintf(stdout, " <size>     Signature granularity");
```
radimkrcmar: The granularity could say that it's in bytes.

nadime15:
> Looks much better, but why don't we use one of the many libraries that handle command line arguments, and can automatically produce the help message from their definition?

My intention was just to add a quick fix and make it clearer which options need an argument and which dont. And we use already `getopt`, are there better alternatives?

> The argument is usually placed right to the parameter.

True, I find it less readable, but I can change it.

radimkrcmar:
> My intention was just to add a quick fix and make it clearer which options need an argument and which dont. And we use already getopt, are there better alternatives?

Yeah, getopt has the advantage that it's a POSIX standard, which also means that basically anything should be better.
[CLI11](https://github.com/CLIUtils/CLI11) seems to be the best one for C++ right now. It also has a single-header version for easy setup.

Beware, it's not first-hand experience, I luckily haven't done any command line parsing in C for years. :)

nadime15:
Ok yeah this looks helpful! I would suggest (if everyone is ok with it) to keep this PR small for now, and open a follow-up issue to refactor the command line parsing logic with a proper C++ lib like [CLI11.](https://github.com/CLIUtils/CLI11)

Timmmm:
I agree, switching to something that isn't from the 70s would be great! CLI11 looks nice.

Timmmm:
(But I don't think that needs to block merging this since it's already done.)

pmundkur:
I implemented the improvements in this PR in CLI11.  The output now looks like this:
```
$ ./build/c_emulator/sail_riscv_sim -h
Sail RISC-V Model 


./build/c_emulator/sail_riscv_sim [OPTIONS] [elfs...]


OPTIONS:
  -h, --help        Print this help message and exit 
  -p, --show-times  Show execution times 
      --version     Print model version 
      --build-info  Print build information 
      --print-default-config 
                    Print default configuration 
      --validate-config 
                    Validate configuration 
      --print-device-tree 
                    Print device tree 
      --print-isa-string 
                    Print ISA string 
      --enable-experimental-extensions 
                    Enable experimental extensions 
  -N, --use-abi-names 
                    Use ABI register names 
  -b, --device-tree-blob <file> 
                    Device tree blob file 
  -t, --terminal-log <file> 
                    Terminal log output file 
  -T, --test-signature <file> 
                    Test signature file 
  -c, --config <file> 
                    Configuration file 
      --trace-output <file> 
                    Trace output file 
  -g, --signature-granularity <uint> 
                    Signature granularity 
  -r, --rvfi-dii <int> (within [1 - 65535]) 
                    RVFI DII port 
  -l, --inst-limit <uint> 
                    Instruction limit 
  -v, --trace <tracer> 
                    Enable tracing option 
  -V, --no-trace <tracer> 
                    Disable tracing option 

 where <tracer> is one of the following in the left column: 
   all : All of the other tracers
   instr : Instruction execution
   mem : Memory accesses
   platform : Privilege changes, MMIO, interrupts
   reg : Register accesses
   rvfi : RVFI tracing
   step : Steps
```

The short and long options are closer together, but option descriptions are sometimes on the next line.

Widening the formatter's column width from 20 to 50 gives:
```
Sail RISC-V Model 


./build/c_emulator/sail_riscv_sim [OPTIONS] [elfs...]


OPTIONS:
  -h,           --help                            Print this help message and exit 
  -p,           --show-times                      Show execution times 
                --version                         Print model version 
                --build-info                      Print build information 
                --print-default-config            Print default configuration 
                --validate-config                 Validate configuration 
                --print-device-tree               Print device tree 
                --print-isa-string                Print ISA string 
                --enable-experimental-extensions  Enable experimental extensions 
  -N,           --use-abi-names                   Use ABI register names 
  -b,           --device-tree-blob <file>         Device tree blob file 
  -t,           --terminal-log <file>             Terminal log output file 
  -T,           --test-signature <file>           Test signature file 
  -c,           --config <file>                   Configuration file 
                --trace-output <file>             Trace output file 
  -g,           --signature-granularity <uint>    Signature granularity 
  -r,           --rvfi-dii <int> (within [1 - 65535]) 
                                                  RVFI DII port 
  -l,           --inst-limit <uint>               Instruction limit 
  -v,           --trace <tracer>                  Enable tracing option 
  -V,           --no-trace <tracer>               Disable tracing option 

 where <tracer> is one of the following in the left column: 
   all : All of the other tracers
   instr : Instruction execution
   mem : Memory accesses
   platform : Privilege changes, MMIO, interrupts
   reg : Register accesses
   rvfi : RVFI tracing
   step : Steps

```

See the edited PR description for backward incompatible changes.

The 'step' tracer doesn't really do much other than print newlines; I can remove it in different PR.

Error messages look like:
```
$./build/c_emulator/sail_riscv_sim -l  build/test/2025-06-22/riscv-tests/rv64ui-p-addw 
Could not convert: --inst-limit = build/test/2025-06-22/riscv-tests/rv64ui-p-addw
Run with --help for more information.

$ ./build/c_emulator/sail_riscv_sim --no-trace foo  build/test/2025-06-22/riscv-tests/rv64ui-p-addw 
--no-trace: foo not in {all,instr,mem,platform,reg,rvfi,step}
Run with --help for more information.

$ ./build/c_emulator/sail_riscv_sim -c foo  build/test/2025-06-22/riscv-tests/rv64ui-p-addw 
--config: File does not exist: foo
Run with --help for more information.
```

c_emulator/riscv_sim.cpp
```diff
@@ -221,147 +202,122 @@ static void read_dtb(const char *path)
...
+      ->check(CLI::Range(1, 65535))
+      ->option_text("<int> (within [1 - 65535])");
+  app.add_option("-l,--inst-limit", insn_limit, "Instruction limit")
+      ->option_text("<uint>");
+#ifdef SAILCOV
+  app.add_option("--sailcov-file", sailcov_file, "Sail coverage output file")
+      ->option_text("<file>");
+#endif
+
+  app.add_option_function("-v,--trace", enable_printer, "Enable tracing option")
```
jordancarlin: Since we're breaking the external API anyway, it's worth rethinking all of our command line flags. I think `-v` for controlling the trace is going to be confusing, especially now that we have a `--version` flag.

c_emulator/riscv_sim.cpp
```diff
@@ -221,147 +202,122 @@ static void read_dtb(const char *path)
...
+  app.add_flag("--print-default-config", do_print_default_config,
+               "Print default configuration");
+  app.add_flag("--validate-config", do_validate_config,
+               "Validate configuration");
+  app.add_flag("--print-device-tree", do_print_dts, "Print device tree");
+  app.add_flag("--print-isa-string", do_print_isa, "Print ISA string");
+  app.add_flag("--enable-experimental-extensions",
+               rv_enable_experimental_extensions,
+               "Enable experimental extensions");
+  app.add_flag("-N, --use-abi-names", config_use_abi_names,
```
jordancarlin: Your PR message says you used `-A` for this option. I think either works, but let's make sure to update the commit message accordingly before merging if we keep `-N`.

radimkrcmar:
Great work!

I'm a bit disappointed that CLI11 doesn't allow control of the short flag width.  Maybe we can apply this change?
```
diff --git a/c_emulator/riscv_sim.cpp b/c_emulator/riscv_sim.cpp
index 03a0aede6c40..f93aaac24338 100644
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -251,7 +251,7 @@ static void setup_options(CLI::App &app)
   // short and long options closer together.  This however pushes some
   // of the option descriptions into the next line, since the long
   // options now flow into the descriptions column.
-  app.get_formatter()->column_width(20);
+  app.get_formatter()->column_width(45);
 
   app.add_flag("-p,--show-times", do_show_times, "Show execution times");
   app.add_flag("--version", do_print_version, "Print model version");
diff --git a/dependencies/CLIUtils/CLI11.hpp b/dependencies/CLIUtils/CLI11.hpp
index 9fa9cc026604..731182613534 100644
--- a/dependencies/CLIUtils/CLI11.hpp
+++ b/dependencies/CLIUtils/CLI11.hpp
@@ -5016,6 +5016,9 @@ class FormatterBase {
     /// The width of the left column (options/flags/subcommands)
     std::size_t column_width_{30};
 
+    /// The alignment of long flags depends on the length of short flags
+    std::size_t short_flag_max_length_{2};
+
     /// The width of the right column (description of options/flags/subcommands)
     std::size_t right_column_width_{65};
 
@@ -11414,9 +11417,8 @@ CLI11_INLINE std::string Formatter::make_option(const Option *opt, bool is_posit
         std::string longNames = detail::join(vlongNames, ", ");
 
         // Calculate setw sizes
-        const auto shortNamesColumnWidth = static_cast<int>(column_width_ / 3);  // 33% left for short names
-        const auto longNamesColumnWidth = static_cast<int>(std::ceil(
-            static_cast<float>(column_width_) / 3.0f * 2.0f));  // 66% right for long names and options, ceil result
+        const auto shortNamesColumnWidth = static_cast<int>(short_flag_max_length_) + 2 + 2; // + Indent + Separator
+        const auto longNamesColumnWidth = static_cast<int>(column_width_) - shortNamesColumnWidth;  // XXX: can be negative
         int shortNamesOverSize = 0;
 
         // Print short names
```
to make the options look like:
```
  -h, --help                                 Print this help message and exit 
  -p, --show-times                           Show execution times 
      --version                              Print model version 
      --build-info                           Print build information 
      --print-default-config                 Print default configuration 
      --validate-config                      Validate configuration 
      --print-device-tree                    Print device tree 
      --print-isa-string                     Print ISA string 
      --enable-experimental-extensions       Enable experimental extensions 
  -N, --use-abi-names                        Use ABI register names 
  -b, --device-tree-blob <file>              Device tree blob file 
  -t, --terminal-log <file>                  Terminal log output file 
  -T, --test-signature <file>                Test signature file 
  -c, --config <file>                        Configuration file 
      --trace-output <file>                  Trace output file 
  -g, --signature-granularity <uint>         Signature granularity 
  -r, --rvfi-dii <int> (within [1 - 65535])  RVFI DII port 
  -l, --inst-limit <uint>                    Instruction limit 
  -v, --trace <tracer>                       Enable tracing option 
  -V, --no-trace <tracer>                    Disable tracing option 
```

radimkrcmar (approved):

.pre-commit-config.yaml
```diff
@@ -23,6 +23,7 @@ repos:
     hooks:
       - id: clang-format
         types_or: [c, c++] # Don't try running clang-format on .json files
+        exclude: "dependencies/.*" # We should really not reformat any dependencies.
```
Timmmm: Is this not handled by the global exclude?

c_emulator/riscv_sim.cpp
```diff
@@ -40,62 +41,83 @@ enum {
...
+    {"reg",      {&config_print_reg, "Register accesses"}           },
+    {"mem",      {&config_print_mem_access, "Memory accesses"}      },
+    {"rvfi",     {&config_print_rvfi, "RVFI tracing"}               },
+    {"platform",
+     {&config_print_platform, "Privilege changes, MMIO, interrupts"}},
+    {"step",     {&config_print_step, "Steps"}                      },
+    {"all",      {NULL, "All of the other tracers"}                 },
+};
+
+std::function<void(const std::string &)> enable_printer {
```
Timmmm: Could these not be normal functions? This seems weird.

c_emulator/riscv_sim.cpp
```diff
@@ -40,62 +41,83 @@ enum {
...
+      auto ent = printers.find(opt);
+      if (ent != printers.end()) {
+        if (opt == "all") {
+          config_print_instr = true;
+          config_print_mem_access = true;
+          config_print_reg = true;
+          config_print_platform = true;
+          config_print_rvfi = true;
+        } else {
+          *std::get<0>(ent->second) = true;
```
Timmmm: The old version is much easier to understand and not any longer.

c_emulator/riscv_sim.cpp
```diff
@@ -40,62 +41,83 @@ enum {
...
+    {"platform",
+     {&config_print_platform, "Privilege changes, MMIO, interrupts"}},
+    {"step",     {&config_print_step, "Steps"}                      },
+    {"all",      {NULL, "All of the other tracers"}                 },
+};
+
+std::function<void(const std::string &)> enable_printer {
+    [](const std::string &opt) {
+      auto ent = printers.find(opt);
+      if (ent != printers.end()) {
```
Timmmm: Will this silently ignore incorrect arguments?

c_emulator/riscv_sim.cpp
```diff
@@ -221,147 +202,122 @@ static void read_dtb(const char *path)
...
-      char *p;
-      unsigned long long val;
-      errno = 0;
-      val = strtoull(optarg, &p, 0);
-      if (*p != '\0' || val > UINT64_MAX
-          || (val == ULLONG_MAX && errno == ERANGE)) {
-        fprintf(stderr, "invalid instruction limit %s\n", optarg);
-        exit(EXIT_FAILURE);
+
+class HelpFormat : public CLI::Formatter {
```
```suggestion
class HelpFormatter : public CLI::Formatter {
```

c_emulator/riscv_sim.cpp
```suggestion
  app.add_flag("-N,--use-abi-names", config_use_abi_names,
```

(I guess it works, just for consistency with the other ones.)

c_emulator/riscv_sim.cpp
```diff
@@ -716,16 +731,20 @@ int main(int argc, char **argv)
...
+      ? rvfi->get_entry()
+      : load_sail(initial_elf_file.c_str(), /*main_file=*/true);
 
   /* Load any additional ELF files into memory */
-  for (int i = files_start + 1; i < argc; i++) {
-    fprintf(stdout, "Loading additional ELF file %s.\n", argv[i]);
-    (void)load_sail(argv[i], /*main_file=*/false);
+  std::vector<std::string>::const_iterator it = elfs.begin();
+  std::advance(it, 1);
+  for (; it != elfs.end(); it++) {
```
```suggestion
  for (auto it = elfs.cbegin() + 1; it != elfs.cend(); it++) {
```

c_emulator/riscv_sim.cpp
Timmmm: Yeah I strongly support changing `-v`!

To be honest I think we should probably remove a lot of the short options. This isn't the sort of program like `ls` or `git` that you're going to be running all day long, and short options are bad in that case.

.pre-commit-config.yaml
pmundkur: Hmm, it does seem to be handled when I check now.  Not sure now why I put that in there. Removed now.

c_emulator/riscv_sim.cpp
pmundkur: The API does require a `std::function`
https://github.com/nadime15/sail-riscv/blob/d5c0fe197e973e8061c1dae8190675041f742c3a/dependencies/CLIUtils/CLI11.hpp#L7229

c_emulator/riscv_sim.cpp
pmundkur: Yeah, but I think the way `config_use_abi_names` was a special case wasn't very clear.  The CLI11 API requires a single argument function unfortunately.

c_emulator/riscv_sim.cpp
pmundkur: No, the second example I gave showed this case:
```
$ ./build/c_emulator/sail_riscv_sim --no-trace foo  build/test/2025-06-22/riscv-tests/rv64ui-p-addw 
--no-trace: foo not in {all,instr,mem,platform,reg,rvfi,step}
Run with --help for more information.
```
The `->check(CLI::IsMember(printers))` takes care of it.

c_emulator/riscv_sim.cpp
pmundkur: Done.

c_emulator/riscv_sim.cpp
pmundkur: Makes sense; I removed the short options.

pmundkur:
> Great work!
> 
> I'm a bit disappointed that CLI11 doesn't allow control of the short flag width. Maybe we can apply this change?
> 
> diff --git a/dependencies/CLIUtils/CLI11.hpp b/dependencies/CLIUtils/CLI11.hpp
> index 9fa9cc026604..731182613534 100644

I'm a bit reluctant to modify the dependency file directly, since it would create porting issues if we upgrade to a newer version.

dependencies/CLIUtils/CLI11.hpp
```diff
@@ -0,0 +1,11527 @@
...
+       (*(result.end() - 2) == ']')) {
+        // this is an escape clause for odd strings
+        std::string nstrs{'['};
+        bool duplicated{true};
+        for(std::size_t ii = 2; ii < result.size() - 2; ii += 2) {
+            if(result[ii] == result[ii + 1]) {
+                nstrs.push_back(result[ii]);
+            } else {
+                duplicated = false;
+                break;
```
radimkrcmar, line 11419: [Moving the discussion here, because the main thread does not support threading.]
>>    I'm a bit disappointed that CLI11 doesn't allow control of the short flag width. Maybe we can apply this change?
>>
>>   diff --git a/dependencies/CLIUtils/CLI11.hpp b/dependencies/CLIUtils/CLI11.hpp
>>    index 9fa9cc026604..731182613534 100644
>
> I'm a bit reluctant to modify the dependency file directly, since it would create porting issues if we upgrade to a newer version.

I agree, it's definitely a potential burden (I was expecting we'd just never update).  I could try to push a change like this upstream, because it doesn't seem to make much sense to subclass the `Formatter` instead, but I might be the only one having issues with reading the column_width(20) help...

Is there even an interest in changing the output?

c_emulator/riscv_sim.cpp
Timmmm: Yeah but I think `std::function` can be implicitly constructed from a normal function pointer - try it!

Timmmm: I would recommend this:

```
void set_config_print(const std::string &var, bool val)
{
  if (var == "all") {
    config_print_instr = val;
    config_print_mem_access = val;
    config_print_reg = val;
    config_print_platform = val;
    config_print_rvfi = val;
  } else if (var == "instr") {
    config_print_instr = val;
  } else if (var == "reg") {
    config_print_reg = val;
  } else if (var == "mem") {
    config_print_mem_access = val;
  } else if (var == "rvfi") {
    config_print_rvfi = val;
  } else if (var == "platform") {
    config_print_platform = val;
  } else if (var == "step") {
    config_print_step = val;
  } else {
    // Should be detected by `->check(CLI::IsMember(printers))`.
    assert(false);
  }
}

...

  app.add_option_function("--trace", [](const std::string &var) { set_config_print(var, true); }, "Enable tracing option")
      ->check(CLI::IsMember(printers))
      ->option_text("<tracer>");
  app.add_option_function("--no-trace", [](const std::string &var) { set_config_print(var, false); },
                          "Disable tracing option")
      ->check(CLI::IsMember(printers))
      ->option_text("<tracer>");
```

dependencies/CLIUtils/CLI11.hpp
pmundkur: Pushing this upstream would be great!  And yes, the output you showed is much more readable.  Perhaps having this local tweak in the dependency is okay, I'm just worried we'd forget about it.

c_emulator/riscv_sim.cpp
pmundkur: This gives the following error:
```
/home/mundkur/src/hw/sail-ol/sail-riscv/c_emulator/riscv_sim.cpp: In function ‘void setup_options(CLI::App&)’:
/home/mundkur/src/hw/sail-ol/sail-riscv/c_emulator/riscv_sim.cpp:322:26: error: no matching function for call to ‘CLI::App::add_option_function(const char [8], setup_options(CLI::App&)::<lambda(const string&)>, const char [22])’
  322 |   app.add_option_function("--trace", [](const std::string &var) { set_config_print(var, true); }, "Enable tracing option")
      |   ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/mundkur/src/hw/sail-ol/sail-riscv/c_emulator/riscv_sim.cpp:18:
/home/mundkur/src/hw/sail-ol/sail-riscv/dependencies/CLIUtils/CLI11.hpp:7229:13: note: candidate: ‘template<class ArgType> CLI::Option* CLI::App::add_option_function(std::string, const std::function<void(const ArgType&)>&, std::string)’
 7229 |     Option *add_option_function(std::string option_name,
      |             ^~~~~~~~~~~~~~~~~~~
/home/mundkur/src/hw/sail-ol/sail-riscv/dependencies/CLIUtils/CLI11.hpp:7229:13: note:   template argument deduction/substitution failed:
/home/mundkur/src/hw/sail-ol/sail-riscv/c_emulator/riscv_sim.cpp:322:26: note:   ‘setup_options(CLI::App&)::<lambda(const string&)>’ is not derived from ‘const std::function<void(const ArgType&)>’
  322 |   app.add_option_function("--trace", [](const std::string &var) { set_config_print(var, true); }, "Enable tracing option")
      |   ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```
This is with `c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0` (g++)

Timmmm: Yeah I guess this is slightly beyond C++'s template deduction rules. I pushed a commit that does it with an explicit `<std::string>` - hope that's ok!

Also I pushed a commit that allows `--trace` and `--no-trace` to be used more than once, which the old interface allowed.

Gotta say I'm not super impressed with CLI11. Definitely better than argparse but things like reflowing the footer... seems pretty amateur. Maybe I've just been spoiled by Rust's `clap` and Python's `Typer` which are both really really good.

Anyway LGTM with these changes.

pmundkur: > Yeah I guess this is slightly beyond C++'s template deduction rules. I pushed a commit that does it with an explicit `<std::string>` - hope that's ok!
> 
> Also I pushed a commit that allows `--trace` and `--no-trace` to be used more than once, which the old interface allowed.

Great, thanks!

dependencies/CLIUtils/CLI11.hpp
pmundkur: @radimkrcmar could you make your suggested change in a followon PR after this one is merged?  That way, we'll have the customization in a separate commit in the history.  Since this is the PR that brings in CLI11, having it here will mean the customization won't be easy to see.

radimkrcmar: Will do, and I'll also send a patch upstream.  (Although any upstream solution can't be as nice as I'd like it to be because of backward compatibility.)

nadime15:
@pmundkur Could you show the output again so we can check the updated version?

pmundkur:
```
$ ./build/c_emulator/sail_riscv_sim -h
Sail RISC-V Model 


./build/c_emulator/sail_riscv_sim [OPTIONS] [elfs...]


OPTIONS:
  -h, --help        Print this help message and exit 
      --show-times  Show execution times 
      --version     Print model version 
      --build-info  Print build information 
      --print-default-config 
                    Print default configuration 
      --validate-config 
                    Validate configuration 
      --print-device-tree 
                    Print device tree 
      --print-isa-string 
                    Print ISA string 
      --enable-experimental-extensions 
                    Enable experimental extensions 
      --use-abi-names 
                    Use ABI register names in trace log 
      --device-tree-blob <file> 
                    Device tree blob file 
      --terminal-log <file> 
                    Terminal log output file 
      --test-signature <file> 
                    Test signature file 
      --config <file> 
                    Configuration file 
      --trace-output <file> 
                    Trace output file 
      --signature-granularity <uint> 
                    Signature granularity 
      --rvfi-dii <int> (within [1 - 65535]) 
                    RVFI DII port 
      --inst-limit <uint> 
                    Instruction limit 
      --trace <tracer> 
                    Enable tracing option 
      --no-trace <tracer> 
                    Disable tracing option 

 where <tracer> is one of the following in the left column: 
   all : All of the other tracers
   instr : Instruction execution
   mem : Memory accesses
   platform : Privilege changes, MMIO, interrupts
   reg : Register accesses
   rvfi : RVFI tracing
   step : Steps
```
It will look better with the tweak from @radimkrcmar

c_emulator/riscv_sim.cpp
pmundkur: The short options have been removed and the PR description updated, so this is not an issue now.

pmundkur (approved):
LGTM, lets merge this soon.

nadime15:
Looks good to me, thanks @pmundkur for finishing this!

Timmmm:
I added a commit that changes the `--trace` flag to `--trace-mem` etc. It's *much* simpler. Help looks like this:

```
      --trace-instr Enable trace output for instruction execution 
      --trace-reg   Enable trace output for register access 
      --trace-mem   Enable trace output for memory accesses 
      --trace-rvfi  Enable trace output for RVFI 
      --trace-platform 
                    Enable trace output for privilege changes, MMIO, interrupts 
      --trace-step  Add a blank line between steps in the trace output 
      --trace-all   Enable all trace output 
```

One slight difference is that `--trace-all` will now enabled `--trace-step` whereas before (in this PR) it wouldn't. I'm not sure if that was intentional but I think it's more readable with the blank lines.

When we have a proper machine readable output we should probably put some effort into making the trace output nicer in general (e.g. colours, adding ELF symbols, etc.) and I think these options will change anyway at that point.

Timmmm:
Oh wait I just remembered the custom formatter was also needed to remove the `POSITIONALS`... But rather than adding it back I think it's better to use it to actually document the input, so I've changed it to this:

```
Sail RISC-V Model 


c_emulator/sail_riscv_sim [OPTIONS] [elfs...]


POSITIONALS:
  elfs TEXT ...     List of ELF files to load. They will be loaded in order, possibly 
                    overwriting each other. PC will be set to the entry point of the 
                    first file. This is optional with some arguments, e.g. 
                    --print-isa-string. 

OPTIONS:
  -h, --help        Print this help message and exit 
...
```

arichardson:
Not reviewed the code changes but the new help looks good to me. One thing I'd suggest: maybe keep `--no-trace` to turn of all tracing so you can append it to a command line to override all previous assignments. Would also ensure we don't break existing scripts.

c_emulator/riscv_sim.cpp
```diff
@@ -220,148 +159,83 @@ static void read_dtb(const char *path)
...
+      "--trace-all",
+      [] {
+        config_print_instr = true;
+        config_print_reg = true;
+        config_print_mem_access = true;
+        config_print_rvfi = true;
+        config_print_platform = true;
+        config_print_step = true;
+      },
+      "Enable all trace output");
```
arichardson, line 229:
```suggestion
  app.add_flag_callback(
      "--trace-all",
      [] {
        config_print_instr = true;
        config_print_reg = true;
        config_print_mem_access = true;
        config_print_rvfi = true;
        config_print_platform = true;
        config_print_step = true;
      },
      "Enable all trace output");
    app.add_flag_callback(
      "--no-trace",
      [] {
        config_print_instr = false;
        config_print_reg = false;
        config_print_mem_access = false;
        config_print_rvfi = false;
        config_print_platform = false;
        config_print_step = false;
      },
      "Disable all trace output");
```
Keep this option?

Timmmm: Seems reasonable to me. If we do add it I think the help message should say `Disable all trace output (overrides previous flags to enable trace)` or something otherwise its existence is kind of misleading.

pmundkur: The callbacks are not called in the order the options appear on the command line, but in the order the options are defined
https://github.com/CLIUtils/CLI11/blob/bb9bd85e3b8129571b084911affc6f0e9ae6be25/include/CLI/impl/App_inl.hpp#L1253

This unfortunately means that `--no-trace --trace-all` and `--trace-all --no-trace` would have the same result, and the result would depend on which of these options was defined last instead of which appeared last on the command line.  So appending to a command line doesn't quite work for these options with callbacks.

arichardson: Oh that's odd, I guess we don't include it then...

Timmmm: Ha yeah I get the feeling that this isn't the best designed CLI library... Ah well. We can improve it when we rewrite the whole thing in Rust :-D

Issue #1126, opened 2025-07-03
rv_16_random_bits is not deterministic and uses /dev/urandom


Timmmm:
This function is used to provide random data to the `seed` CSR:

```
// Provides entropy for the scalar cryptography extension.
uint64_t rv_16_random_bits(void)
{
  // This function can be changed to support deterministic sequences of
  // pseudo-random bytes. This is useful for testing.
  const char *name = "/dev/urandom";
  FILE *f = fopen(name, "rb");
  uint16_t val;
  if (fread(&val, 2, 1, f) != 1) {
    fprintf(stderr, "Unable to read 2 bytes from %s\n", name);
  }
  fclose(f);
  return (uint64_t)val;
}
```

It seems like the author was aware that just using `/dev/urandom` was a bad idea. Since we are using C++ now a better implementation is trivial:

```
// Provides entropy for the scalar cryptography extension.
uint64_t rv_16_random_bits()
{
  static std::mt19937_64 rng(0);
  return static_cast<uint16_t>(rng());
}
```

This is deterministic and should give identical results across different compilers/C++ libraries (weirdly only some C++ PRNGs are guarnateed to do that). The global `static` is unfortunate but we can solve that later.

PR #1131, opened 2025-07-08
Replace rv_16_random_bits implemention with cpp std mt19937

KotorinMinami:
For #1126. Since Sail is currently running on a single core, I think using global `static` is still acceptable.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -82,6 +82,8 @@ bool config_enable_rvfi = false;
 struct timeval init_start, init_end, run_end;
 uint64_t total_insns = 0;
 uint64_t insn_limit = 0;
+std::optional<uint64_t> rng_seed;
+
 #ifdef SAILCOV
 char *sailcov_file = nullptr;
 #endif
@@ -163,6 +165,7 @@ static void setup_options(CLI::App &app)
       ->option_text("<int> (within [1 - 65535])");
   app.add_option("--inst-limit", insn_limit, "Instruction limit")
       ->option_text("<uint>");
+  app.add_option("--rng-seed", rng_seed, "RNG seed")->option_text("<uint>");
 #ifdef SAILCOV
   app.add_option("--sailcov-file", sailcov_file, "Sail coverage output file")
       ->option_text("<file>");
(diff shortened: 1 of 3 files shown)


jrtc27:
Having an option to make the RNG be deterministic seems useful, but I strongly believe that the RNG should be non-deterministic by default; not doing so sounds like it's just asking for trouble.

Timmmm (approved):
LGTM. We can solve the multicore issue later by putting the model in a class and putting the functions like these in a base class of the model class.

jrtc27 (changes requested):
See prior comment

Timmmm:
Sorry didn't see your comment. Why would it be asking for trouble? I would have thought *not* making it deterministic would be asking for trouble.

I believe this is the only non-deterministic thing in the model so if you run your ELF twice you're guaranteed to get the same result. That's a hard requirement for verification which is probably the main use of the model at the moment.

I think there should be an option to specify the seed, including the option to get a random seed from the OS, but having that the default seems like a footgun in this instance.

arichardson:
How about updating this PR to add a rng-seed flag? Then we could keep the non-deterministic behaviour but with an easy way to make it deterministic?

arichardson (changes requested):
Would be good to add a config option to select the seed/enable deterministic mode

nadime15:
@KotorinMinami  I rebased this PR, hope thats ok. I also added an option to set the RNG deterministically.

c_emulator/riscv_sim.cpp
```diff
@@ -82,6 +82,8 @@ bool config_enable_rvfi = false;
 struct timeval init_start, init_end, run_end;
 uint64_t total_insns = 0;
 uint64_t insn_limit = 0;
+std::optional<uint64_t> rng_seed;
```
arichardson, line 85: Could just declare this as extern in one of the two files and then there is no need for the setter and having two duplicate variables?

Timmmm: This PR is kind of subsumed by #1274 btw...

nadime15: Oh I see, just the rng-seed flag is missing but we can do this in a subsequent PR. Unless I am missing anything else, we can probably close this one.

Timmmm:
Closing since this was done as part of another change.

PR #1132, opened 2025-07-09
Fix encoding for `ZBA_RTYPEUW``

jordancarlin:
PR #1119 broke `sh#add.uw` instruction decoding. This splits the `encdec` clauses for `add.uw` and `sh#add.uw` to fix the encoding.


Timmmm:
Oops! I'm going to add running the riscv-arch-tests back to our roadmap...

Timmmm:
I wonder if this should really be a different `instruction` given it doesn't share the `func7` or whatever it is. I think I assumed that it did because it was the same `ZBA_RTYPEUW` and it's kind of elegant to encode it as a shift-add with zero shift.

This if fine for now anyway.

pmundkur (approved):

Issue #1116, opened 2025-07-01
Sort out `width` / `size` names


Timmmm:
We randomly switch between `width` and `size` for the value that represents the number of bytes an instruction is operating on (the `word_width` type). We should pick one and use it everywhere. I think I prefer `width`. I briefly checked the ISA manual and it isn't really consistent, but the Vector extension uses `width` quite a bit (SEW, EEW, "mixed-width", etc.).

PR #1138, opened 2025-07-10
replace "size" names with "width" in instruction operate

KotorinMinami:
For #1116

Sort out the `size` name, mainly include:
- paddr width 
- Load and Store width
- encode mapping name
- pte, vpn, ppn width

The `width` in sail code for the above entries was consistently named `size`, hence the modification.

### DIFF
--- a/model/riscv_vmem.sail
+++ b/model/riscv_vmem.sail
@@ -47,17 +47,17 @@ type PTW_Result('v : Int), is_sv_mode('v) = result((PTW_Output('v), ext_ptw), (P
 // Write a Page Table Entry.
 function write_pte forall 'n, 'n in {4, 8} . (
   paddr    : physaddr,
-  pte_size : int('n),
+  pte_width : int('n),
   pte      : bits('n * 8),
 ) -> MemoryOpResult(bool) =
-  mem_write_value_priv(paddr, pte_size, pte, Supervisor, false, false, false)
+  mem_write_value_priv(paddr, pte_width, pte, Supervisor, false, false, false)
 
 // Read a Page Table Entry.
 function read_pte forall 'n, 'n in {4, 8} . (
   paddr    : physaddr,
-  pte_size : int('n),
+  pte_width : int('n),
 ) -> MemoryOpResult(bits(8 * 'n)) =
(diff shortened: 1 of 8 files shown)


jordancarlin:
I'd also be inclined to leave pte, vpn, and ppn as they are (using size). The rest of the changes look good though.

KotorinMinami:
All right. I would turn them back. By the way, the `tlb_get_pte` function in `model/riscv_vmem_tlb.sail` like:
```
function tlb_get_pte forall 'n, 'n in {4, 8}. (
  pte_width : int('n),
  ent       : TLB_Entry,
) -> bits('n * 8) = ent.pte['n * 8 - 1 .. 0]
```
Would it be better if I change it to `size` ?

Timmmm:
Yeah I think so.

Timmmm (approved):
Nice, thanks!

model/riscv_vmem_tlb.sail
```diff
@@ -35,9 +35,9 @@ struct TLB_Entry = {
 // PTE access functions, for reading flags, and for modifying the dirty and
 // accessed bits.
 function tlb_get_pte forall 'n, 'n in {4, 8}. (
-  pte_width : int('n),
+  pte_size : int('n),
```
pmundkur:
```suggestion
  pte_size  : int('n),
```

model/riscv_vmem.sail
```diff
@@ -227,8 +227,8 @@ function translate_TLB_hit forall 'v, is_sv_mode('v) . (
   ent       : TLB_Entry,
 ) -> TR_Result(ppn_bits('v), PTW_Error) = {
 
-  let pte_width = if sv_width == 32 then 4 else 8;
-  let pte       = tlb_get_pte(pte_width, ent);  // Step 2 of VATP.
+  let pte_size = if sv_width == 32 then 4 else 8;
```
pmundkur:
```suggestion
  let pte_size  = if sv_width == 32 then 4 else 8;
```

pmundkur (approved):
LGTM.  Minor nit: if existing code has horizontal alignment, lets try to preserve it.

PR #1140, opened 2025-07-11
Move privilege-based CSR access into scattered is_CSR_accessible function

Timmmm:
Instead of having special extra functions to check accessibility of CSRs that depend on the privilege mode, and `is_write`, just add those parameters to `is_CSR_defined` (and rename it to `is_CSR_accessible`).

This also fixes a bug (discovered by inspection when making this change) - the `check_Counteren` was not checking the `...h` registers for RV32.

### DIFF
--- a/model/riscv_vmem.sail
+++ b/model/riscv_vmem.sail
@@ -162,7 +162,7 @@ function pt_walk(
 // PUBLIC: see also riscv_insts_zicsr.sail and other CSR-related files
 register satp : xlenbits
 mapping clause csr_name_map = 0x180  <-> "satp"
-function clause is_CSR_defined(0x180) = currentlyEnabled(Ext_S)
+function clause is_CSR_accessible(0x180, priv, _) = currentlyEnabled(Ext_S) & not(priv == Supervisor & mstatus[TVM] == 0b1)
 function clause read_CSR(0x180) = satp
 function clause write_CSR(0x180, value) = { satp = legalize_satp(cur_architecture(), satp, value); Ok(satp) }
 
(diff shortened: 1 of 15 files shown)


model/riscv_vmem.sail
```diff
@@ -162,7 +162,7 @@ function pt_walk(
 // PUBLIC: see also riscv_insts_zicsr.sail and other CSR-related files
 register satp : xlenbits
 mapping clause csr_name_map = 0x180  <-> "satp"
-function clause is_CSR_defined(0x180) = currentlyEnabled(Ext_S)
+function clause is_CSR_accessible(0x180, priv, _) = currentlyEnabled(Ext_S) & priv == Supervisor & mstatus[TVM] == 0b1
```
jordancarlin:
```suggestion
function clause is_CSR_accessible(0x180, priv, _) = currentlyEnabled(Ext_S) & not(priv == Supervisor & mstatus[TVM] == 0b1)
```

Timmmm:
> I don’t love the extra verbosity, but seems like it’s worth it in this case.

Yeah I was wondering if it was worth having two mappings and checking them both...

```
function clause is_CSR_accessible(0x123) =

function clause is_CSR_accessible_2(0x456, priv, _) =
```

Not sure what to call it though... and it seems somewhat more complicated just to avoid some `_`s.

I guess we could avoid half the `_`s by putting priv and is_write in a tuple, but again that seems not worth it.

pmundkur:
I prefer the way you have it in this PR as a single function, with the same signature as `check_CSR`; I think two functions may be confusing.

model/riscv_vmem.sail
Timmmm: Good spot! This feels like the kind of thing AI PR review would be good for...

jordancarlin (approved):

pmundkur (approved):

Issue #1135, opened 2025-07-10
vector widening reduction instructions not allowing LMUL=8


georgiatai:
When testing whether LMUL=8 is legal for vector widening reduction instructions (`vwredsum.vs`, `vwredsumu.vs`). Spike did not trap but sail did (trapping message included below). According to the spec, widening reduction instructions should not be dependent on LMUL, and thus should be allowed to run with an LMUL of 8:

> The scalar input and output operands are held in element 0 of a single vector register, not a vector register group, so any vector register can be the scalar source or destination of a vector reduction regardless of LMUL setting.

This is the assembly that was being run:
```# Testcase cp_custom_vreductionw_vd_vs1_emul_16
li x5, 1                                          # Load desired vl value
vsetvli x0, x5, e16, m8, tu, mu
csrr x4, vtype                                    # save vtype register for after load
csrr x9, vl                                       # save vl register for after load
vsetvli x0, x9, e32, m1, ta, ma                   # set lmul to 1 for load
la x24, vd_random_base_016                        # Load address of desired value
vle32.v v2, (x24)                                 # Load desired value from memory into v2
vsetvl x0, x9, x4                                 # restore vl and vtype setting
la x8, vs2_random_base_016                        # Load address of desired value
vle16.v v16, (x8)                                 # Load desired value from memory into v16
csrr x26, vtype                                   # save vtype register for after load
csrr x21, vl                                      # save vl register for after load
vsetvli x0, x21, e32, m1, ta, ma                  # set lmul to 1 for load
la x10, vs1_random_base_016                       # Load address of desired value
vle32.v v26, (x10)                                # Load desired value from memory into v26
vsetvl x0, x21, x26                               # restore vl and vtype setting
vwredsum.vs v2, v16, v26
vsetvli x6, x0, e32, m1, ta, ma                   # change lmul to 1 and set vl to vlmax to store whole register (offgroup)
RVTEST_SIGUPD_V(x16, x6, 1, 32,  v2)              # stores v2 (sew = 32, AVL = 1) in signature with base (x16) and helper (x6) register```

Sail trapping message:
```mem[X,0x080000894] -> 0x0157
mem[X,0x080000896] -> 0xC70D
[555] [M]: 0x80000894 (0xC70D0157) vwredsum.vs v2, v16, v26
trapping from M to M to handle illegal-instruction
handling exc#0x02 at priv M with tval 0xC70D0157
CSR mstatus (0x300) <- 0x0000000080001E00
[...]

PR #1141, opened 2025-07-11
Allow LMUL=8 for widening reduction instruction

nadime15:
This fixes issue #1135 by allowing LMUL=8 for widening reduction instructions.

I also fixed `vwsll.[vv,vx,vi]`, which were using the widening reduction validation function and replaced them with `illegal_variable_width()`.


nadime15:
Btw. this case was not caught by [riscv-vector-tests](https://github.com/chipsalliance/riscv-vector-tests). I will open a PR to include this case. Once this PR is accepted I will update our vector test repo.

Timmmm (approved):

pmundkur (approved):

PR #1142, opened 2025-07-14
Add a release notes file to docs.

pmundkur:
This describes the various changes in the upcoming release in one place.  It would be good to update this as part of the release procedure.


pmundkur:
This depends on #572, #887 and #1129.

Timmmm:
I feel like this can happily live in the actual Github release notes. Does it need to be in the repo?

Actual notes look great though!

pmundkur:
I'd prefer it to be in the repo so changes could go through PR review.  Perhaps I should copy the notes from previous releases into this file?

pmundkur:
A changelog is also required by Debian packaging: https://lintian.debian.org/tags/no-changelog
(A package is required to provide `/usr/share/doc/pkg` due to https://lintian.debian.org/tags/no-copyright-file)

doc/ReleaseNotes.md
```diff
@@ -0,0 +1,107 @@
...
+- A single executable binary for the various ISA versions (RV32/RV64)
+  and floating-point extensions (F/D).
+
+- The introduction of a JSON-based configuration system.
+
+- The use of the Sail module system to provide a modular structure
+  to the RISC-V model.
+
+- Many new extensions, including complete coverage of the RVA23 vector
+  and vector crypto extensions.
```
jordancarlin: While we have full coverage of the vector crypto extensions, we don't technically have full coverage of the RVA23 vector extensions yet since it optionally includes the vector bfloat16 extensions.

doc/ReleaseNotes.md
```diff
@@ -0,0 +1,107 @@
...
+  scripts for more recent Linux boot images.
+
+## Single executable binary
+
+Releases of previous versions of the model had two executable binaries
+(`riscv_sim_rv32d` and `riscv_sim_rv64d`) to support the RV32 and RV64
+ISAs respectively, that each integrated support for the F and D
+extensions. This release provides a single binary (`sail_riscv_sim`)
+that can be configured to implement either RV32 or RV64, and either
+the F or D extensions, or both.
```
jordancarlin: Nit: You can't have F or D, it's either F or F and D (D implies F).

jordancarlin (commented):
One benefit of keeping a ChangeLog in the repo (side note, might be worth calling it that instead of ReleaseNotes) is that we can update it as major new features are added so that it is easy to generate release notes when it comes time to do a release without needing to look through all of the previous commits.

doc/ReleaseNotes.md
nadime15: What is actually needed to Support bfloat16 extensions? Do we need a different softfloat library?

Timmmm: Yes, but ideally we would implement this in pure Sail. Eventually the rest of the float code would move to pure Sail too.

@Incarnation-p-lee has been doing some great work adding Sail code for float to the Sail compiler: https://github.com/rems-project/sail/tree/sail2/lib/float

Might be worth checking how complete that is at some point - there are some `Not implemented yet.`s so I guess it's not 100%.

One awkward complication is that it uses the float size (16, 32, 64, 128) to distinguish between the types, but f16 and bfloat16 are both 16 bits so I'm not sure what the best option is there. (Worst case: pretend bfloat16 is 17 bits   :-D).

pmundkur: The softfloat on github at https://github.com/ucb-bar/berkeley-softfloat-3 looks like it supports bfloat16.  Until we get native Sail float support, it might be worth updating our softfloat implementation.

nadime15: Ahh ok, thats cool! I have just started familiarizing myself with the topic, but is it possible to add support for bfloat16 to our existing softfloat library while work on native float support in Sail is ongoing?

I just had a quick look at TensorFlows code ([bfloat16.cc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/bfloat16.cc) and [bfloat16.h](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/bfloat16.h)) and it seems to me that every time an actual bfloat operation is performed, it is converted from bfloat to float and then back from float to bfloat.

As the comments in the source code says the transformation from float to bfloat is lossy but I guess this is somehow expected and "deterministic" (in the sense that the expected information lose is always the same) 

I just wonder if a native bfloat16 implementation either in software or hardware would return the same results when compared against the approach above (convert to from bfloat16 to float and back after the computation is done)

I am probably simplifying things and missing a few details, but could we do the same as TensorFlow and add some conversation features and bfloat16 support?

nadime15: > The softfloat on github at https://github.com/ucb-bar/berkeley-softfloat-3 looks like it supports bfloat16. Until we get native Sail float support, it might be worth updating our softfloat implementation.

Oh nice! I missed that! Yeah that would be really cool to get this in.

Incarnation-p-lee: > Yes, but ideally we would implement this in pure Sail. Eventually the rest of the float code would move to pure Sail too.
> 
> @Incarnation-p-lee has been doing some great work adding Sail code for float to the Sail compiler: https://github.com/rems-project/sail/tree/sail2/lib/float
> 
> Might be worth checking how complete that is at some point - there are some `Not implemented yet.`s so I guess it's not 100%.
> 
> One awkward complication is that it uses the float size (16, 32, 64, 128) to distinguish between the types, but f16 and bfloat16 are both 16 bits so I'm not sure what the best option is there. (Worst case: pretend bfloat16 is 17 bits :-D).

Thanks @Timmmm. Yes, it is still under development, and the ALU part is more complicated than others. At most 35% progress for floating point(BF not counted) I think.
I got really busy durning the last 2 months, and will get back in next month (Aug).

jordancarlin: I think the version of softfloat in Spike (which has unfortunately diverged from upstream) has more complete bf16 support, so it might make sense to use that instead (https://github.com/riscv-software-src/riscv-isa-sim/tree/master/softfloat).

Great to hear about the continued progress on native Sail floating point support. Since it seems like it might still be a little while before it’s ready, I agree that updating softfloat would be a good short term solution.

doc/ChangeLog.md
```diff
@@ -0,0 +1,107 @@
...
+- A single executable binary for the various ISA versions (RV32/RV64)
+  and floating-point extensions (F/D).
+
+- The introduction of a JSON-based configuration system.
+
+- The use of the Sail module system to provide a modular structure
+  to the RISC-V model.
+
+- Many new extensions, including almost complete coverage of the RVA23
+  vector crypto extensions.
```
jordancarlin, line 16: This still isn’t quite right. “Almost” was needed when it said “vector and vector crypto”, but now that you changed it to just “vector crypto”, we can claim complete control coverage.

jordancarlin (approved):

doc/ChangeLog.md
```diff
@@ -0,0 +1,117 @@
...
+
+A `--build-info` option prints more build information for the model.
+This should be used when reporting issues in the model.
+
+## Other changes
+
+- The model now requires the Sail 0.19.1 compiler version.
+
+- The Berkeley SoftFloat dependency has been updated to the latest
+  upstream version.
```
nadime15, line 101: Nice! Once we merge #1129 and #1157, we can merge this.

nadime15 (approved):

PR #1143, opened 2025-07-15
Fix number of defined `pmpcfg` CSRs

CharlyCst:
According to the spec, an implementation can implement 0, 16 or 64 PMP entries, which is controlled by `sys_pmp_count`. One entry consists in one `pmpaddr` register and 8 bits of configuration.

Therefore, there should be 0, 4 or 16 `pmpcfg` registers, as each hold 4 configuration entries. However, if `sys_pmp_count` is 16 the current Sail model defines 16 `pmpaddr` and 16 `pmpcfg` registers, i.e. there are more configurations than there are `pmpaddr`.

0 PMP entries -> 0 `pmpaddr`, 0 `pmpcfg`
16 PMP entries -> 16 `pmpaddr`, 16 `pmpcfg` <-- the problem is here
64 PMP entries -> 64 `pmpaddr`, 16 `pmpcfg`

This fixes the model to have 4 `pmpcfg` registers in that case.

### DIFF
--- a/model/riscv_pmp_regs.sail
+++ b/model/riscv_pmp_regs.sail
@@ -236,7 +236,7 @@ mapping clause csr_name_map = 0x3EE  <-> "pmpaddr62"
 mapping clause csr_name_map = 0x3EF  <-> "pmpaddr63"
 
 // pmpcfgN
-function clause is_CSR_defined(0x3A) @ idx : bits(4) = sys_pmp_count > unsigned(idx) & (idx[0] == bitzero | xlen == 32)
+function clause is_CSR_defined(0x3A) @ idx : bits(4) = sys_pmp_count > 4 * unsigned(idx) & (idx[0] == bitzero | xlen == 32)
 function clause read_CSR(0x3A @ idx : bits(4) if idx[0] == bitzero | xlen == 32) = pmpReadCfgReg(unsigned(idx))
 function clause write_CSR((0x3A @ idx : bits(4), value) if idx[0] == bitzero | xlen == 32) = {
   let idx = unsigned(idx);


Timmmm:
Oo I think you're right. Good spot!

Can you change it to *4 on the other side rather than /4 (doesn't really make any difference but Sail deals with multiplication better than division in general and I find it to be more reliable).

Also how did you find this just out of interest?

CharlyCst:
Pushed the `*4` version :)

I found it while working on formal verification for [Miralis](https://github.com/CharlyCst/miralis). Miralis is an M-mode software that virtualizes M-mode, that allows to run traditional M-mode firmware in a sandboxed environment. The virtualization is done by running the firmware in U-mode, and trapping and emulating privileged instructions.
I wrote a Sail to Rust compiler to translate the RISC-V model to Rust, I am currently working on making this [usable as a library](https://github.com/CharlyCst/softcore-rs) by other Rust projects. I use the Rust version of the model with [Kani](https://github.com/model-checking/kani), a Rust model checker, to check that Miralis emulates the RISC-V instructions properly.

In short, I verify that Miralis and the Sail model agree on how to execute M-mode instructions, usually Miralis is wrong, but in this case it seems it was right and the model had a bug.

Timmmm:
Damn that's fancy! And amazing work on the Sail Rust backend. I've been wondering about trying that for a while. What do you use for arbitrary precision arithmetic? Any plans to get it merged into Sail?

CharlyCst:
Thanks!
Indeed unbounded arithmetic is one of the issues. I would like the Rust code to be easy to analyze by verification tools and I don't think there is any simple solution for unbounded arithmetic in Rust that would play nicely with SAT or STM solvers without dedicated support, although I didn't try so this is only speculation.
For now I am using `i128` for all non-bitvector arithmetic, and I rely on Rust to panic in case of overflows. That's not ideal, but on the other hand I expect most ISAs don't actually need unbounded integers. On the RISC-V model I only found overflows for 64 bits integers in the PMP implementation, and all computations fit within 128 bits. I am wondering how much it would change to remove unbounded integers from Sail, probably not much (ignoring backward compatibility)? Or maybe it could be possible to write some tools to compute the upper bound needed by a given model, which would be super helpful for non-theorem-provers targets.

As for upsteaming, I didn't get to talk to the Sail team yet, but if there is interests for sure!
There are too many rough edges for now, however. So the compiler needs a bit more work.

Timmmm:
> That's not ideal, but on the other hand I expect most ISAs don't actually need unbounded integers.

Yeah I think 99% of the bit vectors in RISC-V Sail are not more than 128 bits, but there are couple of notable exceptions:

* vector registers
* CHERI capabilities, I think most of the time we use 128 bits + 1 bit, but there's probably places where we have 129 bit vectors

> Or maybe it could be possible to write some tools to compute the upper bound needed by a given model, which would be super helpful for non-theorem-provers targets.

Yeah that sounds useful. There are a few flags for other backends that do this too:

```
  --Ofixed-int                  assume fixed size integers rather than GMP arbitrary precision integers
  --Ofixed-bits                 assume fixed size bitvectors rather than arbitrary precision bitvectors
  --sv-int-size <n>                set the maximum width for unknown integers
  --sv-bits-size <n>               set the maximum width for bitvectors with unknown width
```

I presume it doesn't actually verify them though. I reckon it's probably possible... but I don't know how you'd deal with the vector registers. It would be really awkward to not be able to use `bits(vlen)` for those.

@Alasdair btw

Timmmm:
Actually I guess the vector registers have *known* size, so a flag to limit the size of only *unknown* bit vectors to 128 bits would be useful. For known size bit vectors you can just use a byte array I guess. I don't think we ever have integers of unknown size more than 128 bits.

CharlyCst:
Ho cool I didn't know about those flags, thanks!
I didn't run into unbounded bitvectors yet, but I did find bounded bitvectors of unknown sizes. I don't handle those properly yet, but the plan is to create a new type with a byte array that tracks the bitvector size dynamically and convert it back to fixed-size bitvector as soon as possible.

Can the flags limiting the max size affect the semantic or is there some static analysis to ensure that is not the case?

Maybe we can move the discussion somewhere else if it's better, by the way.

CharlyCst:
And yes CHERI is an interesting use case, I discussed with some [Capstone](https://capstone.kisp-lab.org/specs/) people who are interested in experimenting with the Rust model, and they have capabilities that might overflow 128 bits too.
I think I can easily substitute the `i128` for a custom type, so a byte byte array or custom `i256` could work for those ISA.

pmundkur (approved):

PR #1144, opened 2025-07-15
Add config and doc files to CPack install.

pmundkur:
This updates the installed files for CPack: in addition to the simulator binary, these now include the RV32 and RV64 configs, the config schema and copyright files (for the model and the softfloat dependency).

It also adds initial support for the generation of .deb and .rpm packages (very experimental).


pmundkur:
It might be good to change the CMake project name from `sail_riscv` to `sail-riscv`.  This PR retains it but changes it for `CPACK_PACKAGE_NAME` for packaging purposes, which is hackish.

The current file hierarchy is:
```
└── usr
    ├── bin
    │   └── sail_riscv_sim
    └── share
        ├── doc
        │   └── sail-riscv
        │       └── copyright
        └── sail-riscv
            ├── config
            │   ├── rv32d.json
            │   └── rv64d.json
            ├── sail_riscv_config_schema.json
            └── sail_riscv_model.json


```
but it could probably be improved.

This PR also depends on #1142 to fix a lintian error.  Both rpmlint and lintian recommend a man page.

cmake/packaging.cmake
```diff
@@ -0,0 +1,51 @@
...
+# Settings for DEB.
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Prashanth Mundkur <${CPACK_PACKAGE_CONTACT}>")
+set(CPACK_DEBIAN_FILE_NAME DEB-DEFAULT)
+set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
+
+# Settings for RPM.
+set(CPACK_RPM_FILE_NAME RPM-DEFAULT)
+set(CPACK_RPM_PACKAGE_DESCRIPTION ${CPACK_PACKAGE_DESCRIPTION})
+set(CPACK_RPM_PACKAGE_LICENSE "BSD-2-Clause")
+set(CPACK_RPM_PACKAGE_AUTOREQ ON)
```
Timmmm: Minor but I try to just use `TRUE`/`FALSE` for consistency - we have `YES` and `ON` here which are unfortunately equivalent!

Timmmm (approved):

model/CMakeLists.txt
```diff
@@ -334,6 +337,9 @@ foreach (variant IN ITEMS "" "smt" "rocq" "rmem" "lean" "lem" "isabelle")
         )
 
         add_custom_target(generated_sail_riscv_docs DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/${model_doc}")
+        install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${model_doc}"
```
jordancarlin: I’m not convinced that we should be installing the documentation bundle by default. Most uses for this will be in another project that will want a specific version, so a system install seems unlikely to be used. It might also confuse people who are only aware of the simulator and mistakenly use it as a config file.

pmundkur: I included it since it was part of the Github release bundle, though not in the `tar.gz`.  Would it be less confusing if it was under `share/sail-riscv/gen-docs/` ?
It would be nice to eventually include all generated outputs in these released versioned packages, at least for archival purposes.

Timmmm: Yeah I think this could go either way - it doesn't really do any harm... except it is a couple of MB.

For Github releases we want it as a separate file so it can easily be downloaded on its own. I'd probably go with Jordan and just omit it for now.

Alasdair:
It might already be done, but one thing to be careful of is we need to include all LICENSE files for any code we are using in release tarballs, e.g. the softfloat license, and possibly the spike license if we use their code for the debug integration.

pmundkur:
The current hierarchy is now:
```
└── usr
    ├── bin
    │   └── sail_riscv_sim
    └── share
        ├── doc
        │   └── sail-riscv
        │       ├── Berkeley-SoftFloat-LICENSE.txt
        │       └── copyright
        └── sail-riscv
            ├── config
            │   ├── rv32d.json
            │   └── rv64d.json
            └── sail_riscv_config_schema.json

```

PR #1146, opened 2025-07-16
Decouple callbacks impl from model by introducing callback interface

trdthg:
Currently the callbacks implementation depends directly on the RVFI implementation. This commit avoids that by introducing a callback registration mechanism, allowing multiple targets to receive the callbacks. Currently the targets are text logs, and RVFI. In future we will add a binary log format, and probably a C lockstep verification API.


c_emulator/riscv_callbacks_if.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "sail.h"
+
+// This class is not declared in "riscv_callbacks.h" as that file is
+// included in the C model generated by Sail.
+
+class callbacks_if {
+public:
+  virtual ~callbacks_if() { };
```
```suggestion
  virtual ~callbacks_if() = default;
```

Timmmm (approved):
Looks like a great improvement to me!

c_emulator/riscv_callbacks.cpp
```diff
@@ -1,139 +1,99 @@
...
-#include "riscv_config.h"
-#include "riscv_platform_impl.h"
-#include "riscv_sail.h"
-#include <stdlib.h>
-#include <vector>
-#include <inttypes.h>
+#include <set>
 
-void print_lbits_hex(lbits val, int length = 0)
+std::set<callbacks_if *> callbacks;
```
arichardson:
```suggestion
std::vector<callbacks_if *> callbacks;
```
I don't think there is any need for this to be a std::set which will make iteration over the elements slower.

c_emulator/riscv_callbacks.cpp
```diff
@@ -1,139 +1,99 @@
...
-  if (length == 0) {
-    length = val.len;
-  }
-  std::vector<uint8_t> data(length);
-  mpz_export(data.data(), nullptr, -1, 1, 0, 0, *val.bits);
-  for (int i = length - 1; i >= 0; --i) {
-    fprintf(trace_log, "%02" PRIX8, data[i]);
-  }
-  fprintf(trace_log, "\n");
+  callbacks.insert(cbi);
```
arichardson:
```suggestion
  callbacks.push_back(cbi);
```

c_emulator/riscv_callbacks.cpp
```diff
@@ -1,139 +1,99 @@
...
+}
+
+void remove_callback_if(callbacks_if *cbi)
+{
+  callbacks.erase(cbi);
```
arichardson:
```suggestion
  std::erase(callbacks, cbi);
```
This is C++20, otherwise
```suggestion
callbacks.erase(std::remove(callbacks.begin(), callbacks.end(), cbi), vec.end());
```

c_emulator/riscv_callbacks_if.h
```diff
@@ -0,0 +1,36 @@
...
+  virtual void csr_full_read_callback(const_sail_string csr_name, unsigned reg,
+                                      sbits value)
+      = 0;
+  virtual void vreg_write_callback(unsigned reg, lbits value) = 0;
+  virtual void pc_write_callback(sbits value) = 0;
+  virtual void trap_callback() = 0;
+};
+
+void register_callback_if(callbacks_if *cbi);
+void remove_callback_if(callbacks_if *cbi);
```
arichardson:
```suggestion
void register_callback(callbacks_if *cbi);
void remove_callback(callbacks_if *cbi);
```

I'm not sure we need the _if on the type either, it's quite obviously an interface.

c_emulator/riscv_sim.cpp
```diff
@@ -306,6 +309,7 @@ static int process_args(int argc, char **argv)
       config_enable_rvfi = true;
       int rvfi_dii_port = atoi(optarg);
       rvfi = rvfi_handler(rvfi_dii_port);
+      register_callback_if(callbacks_impl);
```
arichardson: Now we no longer print things unless rvfi is enabled, we should also add it in the enable tracing case or split those into two separate callback types.

c_emulator/riscv_callbacks_impl.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+#include "sail.h"
+#include "riscv_callbacks_if.h"
+
+class callbacks_if_impl : public callbacks_if {
```
arichardson:
```suggestion
class default_callbacks : public callbacks {
```
How about this?

c_emulator/riscv_sim.cpp
```diff
@@ -53,6 +55,7 @@ char *dtb_file = NULL;
 unsigned char *dtb = NULL;
 size_t dtb_len = 0;
 std::optional<rvfi_handler> rvfi;
+callbacks_if *callbacks_impl = new callbacks_if_impl();
```
```suggestion
default_callbacks callbacks;
```
Can allocate this by value and just pass it with &callbacks further down.

c_emulator/riscv_callbacks.cpp
arichardson: Ah, we should probably check if it is already registered and then skip adding it.

```suggestion
  if (std::find(callbacks.begin(), callbacks.end(), cbi) == callbacks.end())
      callbacks.insert(cbi);
```

Something like this (although there is probably a shorter way of writing this with C++20. I know using a std::set avoids this complexity but for something called multiple times for each instruction we should make iteration as fast as possible.

arichardson (approved):
This is a great improvement. I'd suggest splitting the rvfi and tracing callbacks into two classes, but that can always be done as a follow-up.

LGTM with comments addressed.

c_emulator/riscv_callbacks_if.h
```diff
@@ -0,0 +1,36 @@
...
+// This class is not declared in "riscv_callbacks.h" as that file is
+// included in the C model generated by Sail.
+
+class callbacks_if {
+public:
+  virtual ~callbacks_if() = default;
+
+  virtual void mem_write_callback(const char *type, sbits paddr, uint64_t width,
+                                  lbits value)
+      = 0;
```
arichardson, line 15: clang-format is being really weird here, I wonder if we can fix this. But no need for this PR.

c_emulator/riscv_callbacks_log.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+#include "sail.h"
+#include "riscv_callbacks_if.h"
+
+class log_callbacks_impl : public callbacks_if {
```
arichardson:
```suggestion
class log_callbacks : public callbacks_if {
```

No need for the _impl, this is obviously an implementation.

c_emulator/riscv_callbacks_rvfi.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+#include "sail.h"
+#include "riscv_callbacks_if.h"
+
+class rvfi_callbacks_impl : public callbacks_if {
```
arichardson:
```suggestion
class rvfi_callbacks : public callbacks_if {
```

Same here

c_emulator/riscv_sim.cpp
```diff
@@ -703,8 +708,10 @@ int main(int argc, char **argv)
     print_isa();
   }
 
-  char *initial_elf_file = argv[files_start];
   init_logs();
+  log_callbacks_impl log_callbacks(config_print_reg, config_print_mem_access,
+                                   config_use_abi_names, trace_log);
+  register_callback(&log_callbacks);
```
```suggestion
  log_callbacks log_callbacks(config_print_reg, config_print_mem_access,
                              config_use_abi_names, trace_log);
  register_callback(&log_callbacks);
```

c_emulator/riscv_sim.cpp
```diff
@@ -53,6 +57,7 @@ char *dtb_file = NULL;
 unsigned char *dtb = NULL;
 size_t dtb_len = 0;
 std::optional<rvfi_handler> rvfi;
+rvfi_callbacks_impl rvfi_callbacks;
```
```suggestion
rvfi_callbacks rvfi_callbacks;
```

arichardson: Although I do wonder if we could just have rvfi_handler subclass `callbacks_if`.
Can be done as a follow-up cleanup.

c_emulator/riscv_callbacks_if.h
trdthg: I haven't found the relevant clang-format config yet. 
`= 0` might be treated as `{}` ? I don't know. 🤔 Leave a TODO here.

c_emulator/CMakeLists.txt
```diff
@@ -59,10 +60,23 @@ add_dependencies(riscv_model generated_sail_riscv_model)
...
+set(riscv_callbacks_impl
+    riscv_callbacks_log.cpp
+    riscv_callbacks_log.h
+    riscv_callbacks_rvfi.cpp
+    riscv_callbacks_rvfi.h
+)
+set_source_files_properties(${riscv_callbacks_impl}
+    PROPERTIES COMPILE_OPTIONS
+    # Implementing all callbacks and using all args are not required
+    -Wno-unused
```
arichardson: I'd just omit the parameter names to fix this warning.

c_emulator/riscv_callbacks_log.cpp
```diff
@@ -0,0 +1,112 @@
...
+  if (trace_log && config_print_reg) {
+    fprintf(trace_log, "v%d <- 0x", reg);
+    // TODO: the width of `value` is currently `vlenmax` bits which can be much
+    // greater than VLEN. In future we will remove `vlenmax`, then we can remove
+    // the `zVLEN / 8` argument here.
+    print_lbits_hex(trace_log, value, zVLEN / 8);
+  }
+}
+
+void log_callbacks::pc_write_callback(sbits value) { }
```
arichardson:
```suggestion
void log_callbacks::pc_write_callback(sbits) { }
```
This should fix the -Wunused warning.

arichardson (approved):
Thanks!

Timmmm (approved):
Very good. I think this will make it quite easy to add binary logging output now.

Timmmm:
Rebased & added a couple of minor style fixes - missing braces and explicit `nullptr` checks.

PR #1147, opened 2025-07-16
Add README.md for berkeley-softfloat-3 library

nadime15:
GitHub did not render README.txt or README.html, so this adds a README.md.

### DIFF
--- a/dependencies/softfloat/berkeley-softfloat-3/README.md
+++ b/dependencies/softfloat/berkeley-softfloat-3/README.md
@@ -1,0 +1,23 @@
+
+Package Overview for Berkeley SoftFloat Release 3e
+==================================================
+
+John R. Hauser<br>
+2018 January 20
+
+
+Berkeley SoftFloat is a software implementation of binary floating-point
+that conforms to the IEEE Standard for Floating-Point Arithmetic.  SoftFloat
+is distributed in the form of C source code.  Building the SoftFloat sources
+generates a library file (typically `softfloat.a` or `libsoftfloat.a`)
+containing the floating-point subroutines.
+
+
+The SoftFloat package is documented in the following files in the `doc`
+subdirectory:
+
+* [SoftFloat.html](http://www.jhauser.us/arithmetic/SoftFloat-3/doc/SoftFloat.html) Documentation for using the SoftFloat functions.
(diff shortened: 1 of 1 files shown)


pmundkur (approved):

Timmmm:
I dunno, this seems a little unnecessary and it also means this directory isn't just a straight copy of the upstream code. I guess it doesn't matter too much.

nadime15:
The README.md is actually part of the [berkeley-softfloat-3](https://github.com/ucb-bar/berkeley-softfloat-3) repository. 

I recommend that you have a look at https://github.com/riscv/sail-riscv/tree/master/dependencies/softfloat/berkeley-softfloat-3 and see how its rendered now.

nadime15:
Btw. the code that we have right now is not a straight copy. 99% of the code is the same but some files are different including (at least) one folder name that was changed from (`Linux-RISCV64-GCC` to `Linux-RISCV-GCC`) including the files in the folder.

Timmmm:
Ah ok fair enough. I thought I updated it so that it was a whole ago but I must have not actually made that PR.

nadime15:
@jordancarlin Yes, I am working on it

pmundkur:
A Spike install creates a `softfloat` include directory and `libsoftfloat.so` library; we could use that at the expense of adding a build dependency.  That's an alternative to bringing in the whole directory and then tracking upstream fixes.  Either option sounds good to me.  Spike's softfloat has more updates than the github `ucb-bar` version, so the former (Spike's) might be preferable (to `ucb-bar`).

nadime15:
To be honest, I would prefer not to merge the softfloat version from Spike and stick with [berkeley-softfloat-3](https://github.com/ucb-bar/berkeley-softfloat-3). Its harder to maintain the library otherwise, and we would have to keep track of changes within Spikes codebase. As far as I can tell the berkeley repository has everything we need for now.

Its true that Spike has broader support (they even support direct bfloat16 arithmetic for some reason, which the upstream Berkeley version currently does not), but there is no reason Spike couldnt push those changes upstream. In fact, some of the people maintaining Spike also maintain the Berkeley library.

nadime15:
See https://github.com/ucb-bar/berkeley-softfloat-3/issues/24#issuecomment-1891039292 and https://github.com/ucb-bar/berkeley-softfloat-3/pull/26

Issue #1016, opened 2025-06-07
Predicate `currentlyEnabled()` on `mstatus.{UXL,SXL,MXL}` instead of `xlen` where relevant


pmundkur:
This is more accurate and helps eventually support writable UXL/SXL in the future.  For most extensions, the XL for `cur_privilege` would apply, but for others it won't.
See  https://github.com/riscv/sail-riscv/pull/969#discussion_r2132506357 and https://github.com/riscv/sail-riscv/pull/969#discussion_r2133209557

radimkrcmar:
Sounds good.  `mstatus.MXL` was probably meant to be `xlen`.

Would you accept patches that rename `xlen` to `mxlen`?
I think it would be very confusing if we named something else as XLEN and used `xlen` as MXLEN.

pmundkur:
> Would you accept patches that rename `xlen` to `mxlen`? I think it would be very confusing if we named something else as XLEN and used `xlen` as MXLEN.

That makes sense.  `XLEN` is now inherently a dynamic entity (whereas MXLEN is defined as a constant in recent specs), so it makes sense to avoid using it in our definitions and configuration.

PR #1148, opened 2025-07-16
Make architecture checks in `currentlyEnabled` dynamic by checking `cur_architecture()` instead of `xlen`.

pmundkur:
`cur_architecture()` examines `mstatus.{UXL,SXL,MXL}`, which helps eventually support writable `UXL/SXL` in the future.

This affects checks for Zcf and the virtual memory modes.

This addresses #1016.

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -223,11 +223,11 @@ function in64BitMode() -> bool = {
   cur_architecture() == RV64
 }
 
-function in32BitSupervisorMode() -> bool = {
+function supervisorModeIs32Bit() -> bool = {
   architecture(get_mstatus_SXL(mstatus)) == RV32
 }
 
-function in64BitSupervisorMode() -> bool = {
+function supervisorModeIs64Bit() -> bool = {
   architecture(get_mstatus_SXL(mstatus)) == RV64
 }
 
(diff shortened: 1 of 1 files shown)


model/riscv_sys_regs.sail
```diff
@@ -208,6 +203,42 @@ function get_mstatus_UXL(m : Mstatus) -> arch_xlen = {
...
+// unlikely and not currently supported by the model,
+// you are technically allowed to have writable misa[U/S].
+function clause currentlyEnabled(Ext_U) = hartSupports(Ext_U) & misa[U] == 0b1 & currentlyEnabled(Ext_Zicsr)
+function clause currentlyEnabled(Ext_S) = hartSupports(Ext_S) & misa[S] == 0b1 & currentlyEnabled(Ext_Zicsr)
+
+function clause currentlyEnabled(Ext_Svbare) = currentlyEnabled(Ext_S)
+function clause currentlyEnabled(Ext_Sv32) = hartSupports(Ext_Sv32) & currentlyEnabled(Ext_S) & in32BitMode()
+function clause currentlyEnabled(Ext_Sv39) = hartSupports(Ext_Sv39) & currentlyEnabled(Ext_S) & in64BitMode()
+function clause currentlyEnabled(Ext_Sv48) = hartSupports(Ext_Sv48) & currentlyEnabled(Ext_S) & in64BitMode()
+function clause currentlyEnabled(Ext_Sv57) = hartSupports(Ext_Sv57) & currentlyEnabled(Ext_S) & in64BitMode()
```
nadime15: I just had a really quick look at that, but dont you have to check whether **SXLEN** is 32 or 64? It looks like you are checking the current privilege mode's len mode (MXL, SXL, or UXL) instead. So what happens if you in U-mode and do a vmem access? You still should check SXLEN

model/riscv_insts_zcf.sail
```diff
@@ -6,7 +6,7 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-function clause currentlyEnabled(Ext_Zcf) = hartSupports(Ext_Zcf) & currentlyEnabled(Ext_F) & currentlyEnabled(Ext_Zca) & (currentlyEnabled(Ext_C) | not(hartSupports(Ext_C)))
+function clause currentlyEnabled(Ext_Zcf) = hartSupports(Ext_Zcf) & in32BitMode() & currentlyEnabled(Ext_F) & currentlyEnabled(Ext_Zca) & (currentlyEnabled(Ext_C) | not(hartSupports(Ext_C)))
```
nadime15, line 9: I am not sure if `in32BitMode()` is correct here. I am a bit confused about the definition of RV32 and RV64:

> Zcf is only relevant to RV32, it cannot be specified for RV64.

Whats the exact definition of a RV32 system? I think this is determined by MXLEN (?) which sets XLEN
So what happens if MXLEN (=XLEN) = 64 (RV64) and you set UXLEN to 32. This would enable Zcf. 

I dont know if this is the intended behaviour. Or does it depend on the current `LEN` of the current mode like you currently have?

pmundkur: My reading was that depends on the XLEN of the current mode, to allow for running RV32 VMs (with Zcf) on an RV64 hypervisor.

nadime15: Hmm yeah I thought in this case it would check the "virtualized misa"  and check for VSXL. But honestly I am not sure. @radimkrcmar You probably have more insight :D

radimkrcmar: I would like to exercise my right to remain silent... the manual is effectively unspecified about this, and I think it makes more sense to interpret it as @pmundkur does, but both Spike and QEMU make the extension depend on `mxlen == 32`.

Anyway, adding the check here is always correct, because the extension definitely can't be enabled when `xlen > 32`, but we might also/instead want to make sure that `hartSupport(Ext_Zcf)` cannot be true if `mxlen > 32`.

allenjbaum:
These are the kinds of assertions that UDB wants to be able to extract from
the spec in order to generate tests.
IF you can point out the section/paragraph in the spec that makes it clear,
please document it.

PR #1149, opened 2025-07-17
Rename xlen to mxlen in the config.

pmundkur:
MXLEN is the architectural constant, so use it in user-exposed name in the config.

### DIFF
--- a/config/rv32d.json
+++ b/config/rv32d.json
@@ -1,6 +1,6 @@
 {
   "base": {
-    "xlen": 32,
+    "mxlen": 32,
     "writable_misa": true,
     "writable_fiom": true,
     "writable_hpm_counters": {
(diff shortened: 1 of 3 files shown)


config/rv32d.json
```diff
@@ -1,6 +1,6 @@
 {
   "base": {
-    "xlen": 32,
+    "mxlen": 32,
```
jordancarlin, line 3: If we're going to call this `mxlen` instead of `xlen`, it's probably worth a comment indicating that `xlen` in all priv modes is currently the same. Otherwise this might be misleading for some users.

radimkrcmar: Maybe we can already add options for the valid values of sxlen/uxlen to the config, and deny instances where they're not mxlen.

Timmmm: I feel like that should probably wait until we actually implement it, otherwise it might be quite annoying to temporarily imply that the option is available.

PR #1155, opened 2025-07-21
Update reset vector

Timmmm:
This is maybe slightly more complicated but it means the same code runs on RV32 and RV64.


pmundkur (changes requested):
If we want to change this, it would be nice to make it configurable, and retain the old version as an option.  This version would break lock-step checking with Spike.

Timmmm:
Ah yeah interesting... Does Spike *always* use that reset vector? I think in our version of the emulator I made it not use the reset vector at all by default because we don't use one in our DUT testing.

I definitely think it should be optional anyway (and maybe off by default).

pmundkur:
> Ah yeah interesting... Does Spike _always_ use that reset vector?

It uses the reset vector if there is a device-tree blob, which it loads by default unless explicitly disabled with `--disable-dtb`.  The reset vector is used in all its CI tests.

Timmmm:
I think #1220 is a better approach actually, so I'll close this. Although now that I re-read your comment I guess it will also break lock-step testing with Spike.

In any case I don't think we should make it a goal to have exactly the same boot code as Spike forever. #1220 allows using your own boot code fairly easily, and alternatively any lock-step checking code can just skip the first N instructions from Spike.

PR #1156, opened 2025-07-21
Add --always flag to git describe

Timmmm:
This provides more useful information and avoids an ugly error if you don't have any tags at all.

### DIFF
--- a/cmake/project_version.cmake
+++ b/cmake/project_version.cmake
@@ -5,10 +5,12 @@ set(sail_riscv_release_version "0.7")
 find_package(Git)
 
 if (Git_FOUND)
-  # We could remove --tags if we started annotating the release tags.
-  # --broken allows building in a corrupt git repo.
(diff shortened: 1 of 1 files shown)


cmake/project_version.cmake
```diff
@@ -5,10 +5,11 @@ set(sail_riscv_release_version "0.7")
 find_package(Git)
 
 if (Git_FOUND)
-  # We could remove --tags if we started annotating the release tags.
+  # --tags means it searches for lightweight tags as well as annotated ones.
   # --broken allows building in a corrupt git repo.
+  # --always means it uses the commit hash if it can't find any tags.
```
jordancarlin: Might as well add a description of the dirty flag while you’re modifying this.

Timmmm: Done, I clarified the descriptions a bit too (hopefully).

pmundkur (approved):

PR #1157, opened 2025-07-21
Update softfloat library to support bfloat16 conversions

nadime15:
Pulled in latest upstream changes from berkeley-softfloat-3 to enable bfloat16 conversion support.
    
https://github.com/ucb-bar/berkeley-softfloat-3
    
Based on upstream commit a0c6494cdc11865811dec815d5c0049fba9d82a8.


nadime15:
The first commit just pulls in the code, I didnt touch anything except updating `.pre-commit-config.yaml`, removing `.clang-format` and adding new library files.

My actual changes are in the second commit (https://github.com/riscv/sail-riscv/pull/1157/commits/f3ca3c93338675a2dd25c03d5654bc3c9ce5ec99), which reverts to the original files.

dependencies/softfloat/berkeley-softfloat-3/build/Linux-RISCV-GCC/Makefile
```diff
@@ -396,5 +394,4 @@ softfloat$(LIB): $(OBJS_ALL)
 
 .PHONY: clean
 clean:
-	$(DELETE) $(OBJS_ALL) softfloat$(LIB)
-
+	$(DELETE) $(OBJS_ALL) softfloat$(LIB)
```
Timmmm: This bit can probably be left as it was?

dependencies/softfloat/berkeley-softfloat-3/build/Linux-RISCV-GCC/Makefile
```diff
@@ -36,18 +36,16 @@
 
 SOURCE_DIR ?= ../../source
 SPECIALIZE_TYPE ?= RISCV
-MARCH ?= rv64gcv_zfh_zfhmin
-MABI ?= lp64d
-CC ?= riscv64-unknown-linux-gnu-gcc
+
```
Timmmm: I think you can leave these as they were. Just delete them from the `$(CC)` line.

dependencies/softfloat/berkeley-softfloat-3/build/Linux-RISCV-GCC/platform.h
```diff
@@ -35,20 +35,13 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-/*----------------------------------------------------------------------------
-*----------------------------------------------------------------------------*/
-#ifdef __GNUC_STDC_INLINE__
-#define INLINE inline
-#else
-#define INLINE extern inline
-#endif
+#define INLINE_LEVEL 5
+#define SOFTFLOAT_FAST_INT64
+#define SOFTFLOAT_FAST_DIV64TO32
```
Timmmm: Seems like this file could be left as it was and these `#defines` could be defined in the `Makefile` like they were?

nadime15: We actually dont use the Makefile, it just sits there :D It seems we only make use of `platform.h` 

Everything gets pulled in via `CMakeLists.txt`.

Interestingly, we do this:

`target_compile_definitions(softfloat PRIVATE SOFTFLOAT_ROUND_ODD)`

The other defines are in `platform.h`

If you prefer (I do, since I dont see a good reason to split the defines across two files?),

I could pull them from `platform.h` and add them via `target_compile_definitions`.

Timmmm: Oh yeah I forgot about that. 

> I could pull them from platform.h and add them via target_compile_definitions.

Yeah that makes sense!

dependencies/softfloat/berkeley-softfloat-3/build/Linux-RISCV-GCC/Makefile
nadime15: I pushed the original content back to the Makefile since its not used anyway. Maybe we should just delete it?

dependencies/softfloat/berkeley-softfloat-3/build/Linux-RISCV-GCC/platform.h
nadime15: I moved all defines from platform.h to CMake, except `#define INLINE static inline`, the compiler started complaining. Maybe thats fine, otherwise platform.h would be completely empty. I would like to avoid touching every file that includes it just to remove the include.

Another option would be to skip setting the defines in CMake and keep everything in platform.h.

nadime15: > except #define INLINE static inline, the compiler started complaining.

Hmm actually its possible 

```
target_compile_definitions(softfloat PRIVATE
    LITTLEENDIAN=1
    INLINE_LEVEL=5
    SOFTFLOAT_FAST_INT64
    SOFTFLOAT_FAST_DIV64TO32
    INLINE=static\ inline
    SOFTFLOAT_ROUND_ODD
)
```

previously I had

```
target_compile_definitions(softfloat PRIVATE
    LITTLEENDIAN=1
    INLINE_LEVEL=5
    SOFTFLOAT_FAST_INT64
    SOFTFLOAT_FAST_DIV64TO32
    INLINE="static inline"
    SOFTFLOAT_ROUND_ODD
)
```

Timmmm: Probably `"INLINE=static inline"` might work too... but yeah passing `#define`s containing spaces from CMake to a compiler involves the confluence of 3 different string handlers all of which are terrible at handling spaces!

If not `INLINE=static\ inline` is good!

nadime15: Done! Should I go ahead and remove platform.h and the Makefile entirely? Or just restore the original content? 

I think the latter maybe makes more sense, I also could rename the folder again from Linux-RISCV-GCC to Linux-RISCV64-GCC (the original name) because we dont make use of it anyway. So the whole library matches upstream.

Timmmm: Yeah I think restore it and the directory name so it matches upstream.

dependencies/softfloat/berkeley-softfloat-3/build/Linux-RISCV-GCC/Makefile
pmundkur: Since the Makefile is in upstream, I'd just keep it even if we don't use it, to match the Makefiles in the other build platform directories.

dependencies/softfloat/berkeley-softfloat-3/build/Linux-RISCV-GCC/platform.h
nadime15: I restored all files and now the local softfloat lib matches upstream. I had to add a stub `platform.h` to satisfy the `#include "platform.h"` in the upstream code and to avoid using the default `#define'`s.

I wanted to avoid touching the source files and removing the include from each one.

Let me know what you think.

nadime15:
I can't approve my own PR's but looks good to me. Thanks @pmundkur !

dependencies/softfloat/berkeley-softfloat-3/.clang-format
```diff
@@ -1,3 +0,0 @@
-# This is third-party code and should not be reformatted
-DisableFormat: true
-SortIncludes: Never
```
arichardson: Can we move this file to the top-level dependencies/ folder? That should make clang-format work as expected even when not run from a pre-commit hook.

The only reason I didn't do this before is that there wasn't a top-level dependencies folder.

jordancarlin (approved):

Timmmm (approved):

Timmmm:
Hmm I think this broke my build somehow. I keep getting undefined references to things like `softfloat_countLeadingZeros32`... Can't quite figure out why.

Timmmm:
Fix: #1176

PR #1158, opened 2025-07-21
Add support for Zfbfmin extension

nadime15:
Add following instructions to support bfloat16 conversions:

```
fcvt.bf16.s
fcvt.s.bf16
```


nadime15:
The first two commits can be ignored, the support for Zfbfmin is added in https://github.com/riscv/sail-riscv/pull/1158/commits/c8c91f6912b340709ca68df3dd894abdf8639add. The other two commits just update the softfloat lib and are based on #1157.

model/riscv_fdext_regs.sail
```diff
@@ -325,9 +358,10 @@ function wF_or_X_D(i, data) = {
   }
 }
 
-overload F_or_X_H = { rF_or_X_H, wF_or_X_H }
-overload F_or_X_S = { rF_or_X_S, wF_or_X_S }
-overload F_or_X_D = { rF_or_X_D, wF_or_X_D }
+overload F_or_X_BF16 = { rF_or_X_BF16, wF_or_X_BF16 }
+overload F_or_X_H    = { rF_or_X_H, wF_or_X_H }
+overload F_or_X_S    = { rF_or_X_S, wF_or_X_S }
+overload F_or_X_D    = { rF_or_X_D, wF_or_X_D }
```
Timmmm: IMO we shouldn't endlessly align these things. I makes `git blame` way more annoying to use.

Timmmm:
Wasn't the whole point of bf16 that conversion to/from f32 is completely trivial truncation. Do we really need to use softfloat here?

See https://en.wikipedia.org/wiki/Bfloat16_floating-point_format#Rounding_and_conversion

Seems like conversion from f32 to bf16 might need some rounding support but I think lets just implement that now. We'll only have to do it later anyway. There is some rounding code in [this file](https://github.com/rems-project/sail/blob/f6c7540c066eaecde982cb71e977722a32f37a3d/lib/float/arith_internal.sail), I dunno if you can reuse that somehow.

Definitely don't need softfloat for bf16 to f32 anyway.

nadime15:
Im not a floating-point expert, so I am a bit cautious doing things on my own here. I dont really see an disadvantage to just use the bf16_to_f32 function from the lib. 

Yes, as far as I can tell, the conversion should be trivial, but I preferred to rely on a proven library rather than doing it by myself. Especially after looking at the actual [code](https://github.com/ucb-bar/berkeley-softfloat-3/blob/master/source/bf16_to_f32.c), it does not look as straightforward as just shifting a few bits.

> Seems like conversion from f32 to bf16 might need some rounding support but I think lets just implement that now. We'll only have to do it later anyway. There is some rounding code in [this file](https://github.com/rems-project/sail/blob/f6c7540c066eaecde982cb71e977722a32f37a3d/lib/float/arith_internal.sail), I dunno if you can reuse that somehow.


Is this not the code that is still in development? I actually though the softfloat lib takes care of the rounding via


```
#define SOFTFLOAT_PRELUDE(rm)                                                  \
  softfloat_exceptionFlags = 0;                                                \
  softfloat_roundingMode = (uint_fast8_t)rm

#define SOFTFLOAT_POSTLUDE(res)                                                \
  zfloat_result = res.v;                                                       \
  zfloat_fflags = (mach_bits)softfloat_exceptionFlags
```

And
 ```
bfloat16_t  softfloat_roundPackToBF16( bool sign, int_fast16_t exp, uint_fast16_t sig )
 uint_fast8_t roundingMode;
    bool roundNearEven;
    uint_fast8_t roundIncrement, roundBits;
    bool isTiny;
    uint_fast16_t uiZ;
    union ui16_bf16 uZ;

    /*------------------------------------------------------------------------
    *------------------------------------------------------------------------*/
    roundingMode = softfloat_roundingMode;
    roundNearEven = (roundingMode == softfloat_round_near_even);
    roundIncrement = 0x40;
    if ( ! roundNearEven && (roundingMode != softfloat_round_near_maxMag) ) {

...
```

I think there is no need to modify/add anything.

**Comment by Timmmm** [comment]
> I dont really see an disadvantage to just use the bf16_to_f32 function from the lib.

There are several disadvantages:

* They only work with the C backend. They can't be used by the SV or Lean backends, or the SMT property checks. This is one of the main reasons the float code is being rewritten in Sail.
* They make it harder to understand the spec because you have to jump into C rather than staying in pure Sail.
* It kind of adds extra work because we'll have to undo this work later to change it to be pure Sail.

> Yes, as far as I can tell, the conversion should be trivial, but I preferred to rely on a proven library rather than doing it by myself.

I would write a short test to check against softfloat exhaustively (there are only 2^32 * 6 inputs which you can probably go through in a few hours. Also are there tests for this extension somewhere?

> Is this not the code that is still in development?

Yes but that bit of the code seems to be fairly complete as far as I can see?

> I think there is no need to modify/add anything.

Right, I mean you need some rounding logic if you aren't using softfloat.

**Comment by nadime15** [comment]
> There are several disadvantages:
> 
>     They only work with the C backend. They can't be used by the SV or Lean backends, or the SMT property checks. This is one of the main reasons the float code is being rewritten in Sail.
>     They make it harder to understand the spec because you have to jump into C rather than staying in pure Sail.
>     It kind of adds extra work because we'll have to undo this work later to change it to be pure Sail.
> 

I mean in this case you couldn use the extension anyway for different backends except C. You would have one instruction float32_to_bf16 thats written in C and bf16_to_float32 written in Sail.

So if you really wanna go into this direction I think it makes more sense to write the whole extension in Sail (which will probably be done at some point when Sail has native support for float) instead of just one instruction and in this case I think it just makes sense to wait until Sail has native float support.

> Also are there tests for this extension somewhere?

No not really... I wrote a few tests and compared against Spike that worked (I will edit this message later on and add a link). PyTorch has bfloat16  support and could be used to generate a bunch of tests but they have their own softfloat lib I think. 

> Right, I mean you need some rounding logic if you aren't using softfloat.

Yes when the extension/instruction is written in pure Sail.

Edit: 

> I would write a short test to check against softfloat exhaustively (there are only 2^32 * 6 inputs which you can probably go through in a few hours.

Thats a good idea, I will do that!

**Comment by nadime15** [comment]
@Timmmm I gave it a try and wrote `fcvt.s.bf16` in Sail, I am not done yet, I need to organize the code a bit better, get rid of magic numbers, remove the bf16_to_f32 API. etc. Let me know what you think of this early version!

**Comment by nadime15** [comment]
I have fully verified `fcvt.s.bf16`. I wrote a script that exhaustively converts all `2^16` possible bit patterns to float32 and checks the results, including exception flags. So far every single test passes.

I have not tested the full float32 range for `fcvt.bf16.s`, but I did add bfloat16 support to the vector test suite ([PR #77](https://github.com/chipsalliance/riscv-vector-tests/pull/77)), which allowed me to test the vector versions of these instructions (`Zvfbfmin`). So far, all vector tests pass.

Also, since `Zvfbfwma` is basically based on `fcvt.s.bf16`  and `vfmacc.v[vf]` for the widening multiply-accumulate, we can say `Zvfbfwma` is verified as well, unless I have overlooked something. `Zvfbfwma` was also tested with the vector test suite.

File: model/riscv_fdext_regs.sail [hunk 2]
```diff
@@ -210,6 +210,20 @@ function wF_bits(Fregidx(i) : fregidx, data: flenbits) -> unit = {
 
 overload F = {rF_bits, wF_bits, rF, wF}
 
+val rF_BF16 : fregidx -> bits(16)
+function rF_BF16(i) = {
+  assert(flen >= 16);
```
Timmmm: Isn't the minimum `flen` always 32? I don't think you can support half float without F:

> The BF16 extensions defined in this specification (i.e., Zfbfmin, Zvfbfmin, and Zvfbfwma) depend on the single-precision floating-point extension F.

Oh wait I see this is copied from the other ones. I wonder if we should remove them all except `assert(flen >= 64)`.

model/riscv_insts_zfbfmin_utils.sail
```diff
@@ -0,0 +1,27 @@
...
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function split_BF16 (v : bits(16)) -> (bits(1), bits(8), bits(7)) = (v[15..15], v[14..7], v[6..0])
+
+function bf16_to_f32 (v : bits(16)) -> bits(32) = {
+  let (sign, exp, mant) = split_BF16(v);
+  let is_nan = exp == 0xFF & mant != 0b0000000;
+  let is_inf = exp == 0xFF & mant == 0b0000000;
```
Timmmm:
```suggestion
  let is_nan = exp == ones() & mant != zeros();
  let is_inf = exp == ones() & mant == zeros();
```

model/riscv_insts_zfbfmin_utils.sail
```diff
@@ -0,0 +1,27 @@
...
+
+  if is_nan then {
+    if mant[6] == bitzero then accrue_fflags(nvFlag());
```
Timmmm:
```suggestion
    if signed(mant) >= 0 then accrue_fflags(nvFlag());
```

Might not be as efficient at runtime but I think it probably more closely matches the intent and avoids the slightly magic/risky `6`.

model/riscv_insts_zfbfmin_utils.sail
```diff
@@ -0,0 +1,27 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+function split_BF16 (v : bits(16)) -> (bits(1), bits(8), bits(7)) = (v[15..15], v[14..7], v[6..0])
```
Timmmm: Should be `fsplit_BF16` for consistency with the other functions.

model/riscv_insts_zfbfmin_utils.sail
```diff
@@ -0,0 +1,27 @@
...
+  let is_nan = exp == 0xFF & mant != 0b0000000;
+  let is_inf = exp == 0xFF & mant == 0b0000000;
+
+  if is_nan then {
+    if mant[6] == bitzero then accrue_fflags(nvFlag());
+    // 32-Bit Floating-Point NaN
+    canonical_NaN_S()
+  } else if is_inf then {
+    // Infinity
+    sign @ 0xFF @ zeros(23)
```
Timmmm: I would add `fmake_BF16` for consistency and then

```
  fmake_BF16(sign, ones(), zeros())
} else {
  fmake_BF16(sign, exp, mant @ zeros(16))
}
```

This kind of style will work better when we use the generic versions from the Sail standard library. (Though tbf I'm not sure if it is possible to do that with BF16 since it is the same size as f16.)

Timmmm (commented):
LGTM, couple of minor nits.

model/riscv_fdext_regs.sail
nadime15: Ah yeah good point. I can do this in a follow up PR or in this one, just let me know what you prefer. I have just changed it now to flen >= 32 (if we get rid of the asserts anyway it shouldn matter this much)

model/riscv_insts_zfbfmin_utils.sail
nadime15: We return float values here, calling `fmake_S` should do the trick. But I still can add `fmake_B16`,  we might use it at some point.

nadime15: Done

model/riscv_validate_config.sail
```diff
@@ -127,10 +127,19 @@ function check_pmp() -> bool = {
   valid
 }
 
+function check_bfloat16() -> bool = {
+  var valid : bool = true;
+  if hartSupports(Ext_Zfbfmin) & not(hartSupports(Ext_F)) then {
+    valid = false;
```
Timmmm, line 138: This needs to print a message.

nadime15: Done

model/riscv_insts_zfbfmin_utils.sail
```diff
@@ -0,0 +1,29 @@
...
+
+function fmake_BF16 (sign : bits(1), exp : bits(8), mant : bits(7)) -> (bits(16)) = (sign @ exp @ mant)
+
+function bf16_to_f32 (v : bits(16)) -> bits(32) = {
+  let (sign, exp, mant) = fsplit_BF16(v);
+  let is_nan = exp == ones() & mant != zeros();
+  let is_inf = exp == ones() & mant == zeros();
+
+  if is_nan then {
+    if signed(mant) >= 0 then accrue_fflags(nvFlag());
```
Timmmm: Sorry to go backwards, but I wonder if it would make more sense to return the new fflags, like all the other float functions do. I.e.

```
function bf16_to_f32 (v : bits(16)) -> (bits(32), fflags) = {
  let (sign, exp, mant) = fsplit_BF16(v);
  let is_nan = exp == ones() & mant != zeros();
  let is_inf = exp == ones() & mant == zeros();

  let flags = if is_nan & signed(mant) >= 0 then nvFlag() else zeros();
  let value =
    if is_nan then canonical_NaN_S()
    else if is_inf then fmake_S(sign, ones(), zeros())
    else fmake_S(sign, exp, mant @ zeros(16));

  (value, flags)
}
```

And put the `accrue_fflags` in the instruction.

That also makes this function pure, which is a lot nicer.

nadime15: Good point! Done

model/riscv_fdext_regs.sail
```diff
@@ -252,10 +266,20 @@ function wF_D(i, data) = {
...
 overload F_D = { rF_D, wF_D }
 
+val rF_or_X_BF16 : fregidx -> bits(16)
+function rF_or_X_BF16(i) = {
+  assert(flen >= 32);
+  assert(hartSupports(Ext_F) != hartSupports(Ext_Zfinx));
+  if   hartSupports(Ext_F)
+  then F_BF16(i)
+  else X(fregidx_to_regidx(i))[15..0]
+}
```
jordancarlin: Zfbfmin requires F, so there should never be a case where it is reading from the X registers. I think we can drop this (and the corresponding `wF_or_X_BF16`) and directly use `F_BF16` instead.

nadime15: Ah, thanks for pointing that out! I have removed the asserts for `rF_BF16/wF_16` since BF16 instructions should not execute if the F extension is not enabled, which we already check via currentlyEnabled. 

This could provide a slight performance boost for heavy BF16 workloads. Tell me if you think I should them add back in.

model/riscv_insts_zfbfmin.sail
```diff
@@ -0,0 +1,53 @@
...
+/*=======================================================================================*/
+
+// Zfbfmin also defines FLH, FSH, FMV.X.H, and FMV.H.X instructions as defined in the Zfh extension
+
+function clause currentlyEnabled(Ext_Zfbfmin) = hartSupports(Ext_Zfbfmin) & currentlyEnabled(Ext_F)
+
+union clause instruction = FCVT_BF16_S : (fregidx, rounding_mode, fregidx)
+
+mapping clause encdec = FCVT_BF16_S(rs1, rm, rd)
+ <-> 0b01000 @ 0b10 @ 0b01000 @ encdec_freg(rs1) @ encdec_rounding_mode(rm) @ encdec_freg(rd) @ 0b1010011 when currentlyEnabled(Ext_Zfbfmin)
```
pmundkur: Indentation by 2, and the `when` on the next line
```suggestion
  <-> 0b01000 @ 0b10 @ 0b01000 @ encdec_freg(rs1) @ encdec_rounding_mode(rm) @ encdec_freg(rd) @ 0b1010011
  when currentlyEnabled(Ext_Zfbfmin)
```

model/riscv_insts_zfbfmin.sail
```diff
@@ -0,0 +1,53 @@
...
+  <-> "fcvt.bf16.s" ^ spc() ^ freg_name(rd) ^ sep() ^ freg_name(rs1) ^ sep() ^ frm_mnemonic(rm)
+
+union clause instruction = FCVT_S_BF16 : (fregidx, rounding_mode, fregidx)
+
+mapping clause encdec = FCVT_S_BF16(rs1, rm, rd)
+ <-> 0b01000 @ 0b00 @ 0b00110 @ encdec_freg(rs1) @ encdec_rounding_mode(rm) @ encdec_freg(rd) @ 0b1010011 when currentlyEnabled(Ext_Zfbfmin)
+
+function clause execute (FCVT_S_BF16(rs1, rm, rd)) = {
+  if select_instr_or_fcsr_rm(rm) == None()
+    then return Illegal_Instruction();
```
pmundkur:
```suggestion
  if select_instr_or_fcsr_rm(rm) == None()
  then return Illegal_Instruction();
```

nadime15: Done

PR #1160, opened 2025-07-22
Extract jumping to a PC into a function

Timmmm:
Create a function that JAL, JALR and branch instructions can all use for jumping to a target address.

### DIFF
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -43,7 +43,8 @@ mapping clause assembly = UTYPE(imm, rd, op)
 /* ****************************************************************** */
 
 // Jump execution to a specified target address. This can fail
-// due to the target being misaligned, or due to extension checks.
+// due to the target not being 4-byte aligned, or due to extension checks.
+// Callers must ensure that the target address is 2-byte aligned.
 function jump_to(target : xlenbits) -> ExecutionResult = {
   // Extensions get the first checks on the prospective target address.
   match ext_control_check_pc(target) {
@@ -51,6 +52,11 @@ function jump_to(target : xlenbits) -> ExecutionResult = {
     Ext_ControlAddr_OK(target) => {
       // Perform standard alignment check.
       let target_bits = bits_of(target);
+      // Check target is at least 2-byte aligned (callers must ensure
+      // this so it can be an assertion).
+      assert(target_bits[0] == bitone);
+      // If it is not 4-byte aligned an compressed instructions are
+      // not enabled then raise an alignment exception.
       if bit_to_bool(target_bits[1]) & not(currentlyEnabled(Ext_Zca)) then {
         Memory_Exception(target, E_Fetch_Addr_Align())
       } else {
--- a/model/riscv_insts_base.sail
+++ b/model/riscv_insts_base.sail
@@ -41,39 +41,37 @@ mapping clause assembly = UTYPE(imm, rd, op)
   <-> utype_mnemonic(op) ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_signed_20(imm)
 
 /* ****************************************************************** */
+
+// Jump execution to a specified target address. This can fail
+// due to the target being misaligned, or due to extension checks.
+function jump_to(target : xlenbits) -> ExecutionResult = {
+  // Extensions get the first checks on the prospective target address.
+  match ext_control_check_pc(target) {
+    Ext_ControlAddr_Error(e) => Ext_ControlAddr_Check_Failure(e),
+    Ext_ControlAddr_OK(target) => {
+      // Perform standard alignment check.
+      let target_bits = bits_of(target);
+      if bit_to_bool(target_bits[1]) & not(currentlyEnabled(Ext_Zca)) then {
+        Memory_Exception(target, E_Fetch_Addr_Align())
+      } else {
+        set_next_pc(target_bits);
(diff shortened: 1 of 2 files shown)


arichardson (commented):
I really link this to avoid duplicated code but it looks like there is some subtle change in behaviour here (considering the test failure)?

model/riscv_jalr_seq.sail
```diff
@@ -9,26 +9,14 @@
...
- * some manner, but for now, we just keep a reordered definition to improve simulator
- * performance.
- */
-  let t : xlenbits = X(rs1) + sign_extend(imm);
-  /* Extensions get the first checks on the prospective target address. */
-  match ext_control_check_addr(t) {
-    Ext_ControlAddr_Error(e) =>
-      Ext_ControlAddr_Check_Failure(e),
-    Ext_ControlAddr_OK(addr) => {
-      let target = [bits_of(addr) with 0 = bitzero];  /* clear addr[0] */
```
jrtc27, line 24: This weirdo special behaviour for JALR has been lost

arichardson: Ah yeah I guess it could be kept in the new function unconditionally since it will always be zero for the other two callers.

arichardson: Or probably better to just zero it in the JALR execute

model/riscv_insts_base.sail
```diff
@@ -41,42 +41,40 @@ mapping clause assembly = UTYPE(imm, rd, op)
...
-but this is difficult
-*/
-
-function clause execute (JAL(imm, rd)) = {
-  let target = PC + sign_extend(imm);
-  /* Extensions get the first checks on the prospective target address. */
+// Jump execution to a specified target address. This can fail
+// due to the target being misaligned, or due to extension checks.
+function jump_to(target : xlenbits) -> ExecutionResult = {
+  // Extensions get the first checks on the prospective target address.
```
arichardson, line 49:
```suggestion
  assert(target[0] == 0);
  // Extensions get the first checks on the prospective target address.
```

jrtc27: Well, ext_control_check_pc is currently expecting to get the possibly-LSB-set target, not the masked one. I don't think that's really necessary though, even CHERI ISAv9 bounds checks CJALR based on the masked target, not the unmasked one. I guess it was (accidentally?) checking the unmasked target against PCC for JALR though? CHERI QEMU checks the masked target, as do CHERI Flute and CHERI Toooba.

arichardson: true should not be needed

Alasdair:
Looking at this PR reminds me that the syntax suggested in this comment
```
/*
ideally we want some syntax like
mapping clause encdec = JAL(imm @ 0b0, rd) <-> imm[19] @ imm[9..0] @ imm[10] @ imm[18..11] @ encdec_reg(rd) @ 0b1101111
```
actually exists now. I should make a separate PR for that.

model/riscv_jalr_seq.sail
Timmmm: I fixed JALR and added an assertion in `jump_to` to check the lowest bit is 0.

Arguably we should have `newtype pc_addr = PcAddr : bits(xlen - 1)` which omits that bit entirely... but that seemed like it might just be confusing and overkill. (But I would be open to being persuaded that we should; types are good.)

model/riscv_insts_base.sail
```diff
@@ -41,42 +41,46 @@ mapping clause assembly = UTYPE(imm, rd, op)
...
   match ext_control_check_pc(target) {
     Ext_ControlAddr_Error(e) => Ext_ControlAddr_Check_Failure(e),
     Ext_ControlAddr_OK(target) => {
-      /* Perform standard alignment check */
+      // Perform standard alignment check.
       let target_bits = bits_of(target);
+      // Check target is at least 2-byte aligned (callers must ensure
+      // this so it can be an assertion).
+      assert(target_bits[0] == bitone);
```
jordancarlin:
```suggestion
      assert(target_bits[0] == bitzero);
```

model/riscv_insts_base.sail
```diff
@@ -41,42 +41,46 @@ mapping clause assembly = UTYPE(imm, rd, op)
...
+      // If it is not 4-byte aligned an compressed instructions are
```
jordancarlin:
```suggestion
      // If it is not 4-byte aligned and compressed instructions are
```

model/riscv_insts_base.sail
Timmmm: lol how :-D

model/riscv_insts_base.sail
```diff
@@ -41,42 +41,46 @@ mapping clause assembly = UTYPE(imm, rd, op)
...
+
+function clause execute (JAL(imm, rd)) = {
+  let link_address = get_next_pc();
+  match jump_to(PC + sign_extend(imm)) {
+    Retire_Success() => { X(rd) = link_address; Retire_Success() },
+    failure => failure,
+  }
+}
+
 /* TODO: handle 2-byte-alignment in mappings */
```
Timmmm, line 84: I wonder if this TODO was referring to the idea of not appending the `@ 0b0` in the `instruction`?

jordancarlin (approved):

pmundkur (approved):

Issue #1139, opened 2025-07-11
Validate JSON config file with schema before passing to Sail model


jordancarlin:
The Sail model gives very cryptic error messages when anything is missing from the config file (often referring to pattern match failures). We should validate the config against the JSON schema that Sail produces before passing it to the generated portion of the model so that we can give a clear error message indicating that the config is wrong and hopefully what specific keys cause the problem.

Arielfoever:
https://github.com/rems-project/sail/pull/1388 I put the message from CJson to the output but it is limited due to CJson it a simple lab. The change of lib is needed if you want more info.

jordancarlin:
There are a list of C++ json schema validation libraries to choose from here: https://json-schema.org/tools?query=&sortBy=name&sortOrder=ascending&groupBy=toolingTypes&licenses=&languages=C%2B%2B&drafts=&toolingTypes=&environments=&showObsolete=false&supportsBowtie=false#validator.

Arielfoever:
> There are a list of C++ json schema validation libraries to choose from here: https://json-schema.org/tools?query=&sortBy=name&sortOrder=ascending&groupBy=toolingTypes&licenses=&languages=C%2B%2B&drafts=&toolingTypes=&environments=&showObsolete=false&supportsBowtie=false#validator.

I reviewed your five options.

- Blaze f5-json-schema jsoncons have an incompatible license was used while we obviously couldn't get a commercial license.
- JSON schema validation for JSON for Modern C++ needs a lot of changes of cmake

I'd love to use https://github.com/tristanpenman/valijson

Alasdair:
You don't necessarily need a library, you can just a validator that's a separate program so it doesn't need to be C or C++.

Timmmm:
> You don't necessarily need a library, you can just a validator that's a separate program so it doesn't need to be C or C++.

That feels quite fragile though, and would probably make distribution a bit harder. My preference would still be for Sail to generate C code to do the validation, but that is extra work.

Timmmm:
Especially because JSON Schema can't express all of the type requirements that Sail can (e.g. `elen_exp <= vlen_exp`).

edolnx:
[...]

PR #1161, opened 2025-07-23
Validate JSON config file with schema

Arielfoever:
fix https://github.com/riscv/sail-riscv/issues/1139


.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "dependencies/valijson"]
+	path = dependencies/valijson
+	url = https://github.com/tristanpenman/valijson
+[submodule "dependencies/nlohmann-json"]
+	path = dependencies/nlohmann-json
+	url = https://github.com/nlohmann/json.git
```
Timmmm: Since these are both single file header only libraries, let's just copy them into `dependencies` like for CLI11.

c_emulator/riscv_sim.cpp
```diff
@@ -347,8 +353,69 @@ static int process_args(int argc, char **argv)
...
+  // Validate JSON
+
+  using namespace valijson;
+  using namespace valijson::adapters;
+
+  nlohmann::json schemaDocument;
+  if (!valijson::utils::loadDocument(JSON_SCHEMA, schemaDocument)) {
+    fprintf(stderr, "Failed to load config schema document '%s'.\n",
+            JSON_SCHEMA);
+    // cerr << "Failed to load schema document." << endl;
```
Timmmm: Yeah lets use `std::cerr`

c_emulator/riscv_sim.cpp
```diff
@@ -347,8 +353,69 @@ static int process_args(int argc, char **argv)
   }
 
   if (!have_config) {
-    sail_config_set_string(DEFAULT_JSON);
+    config_file = strdup(DEFAULT_JSON);
```
Timmmm: Probably better to change `config_file` to `std::string` and then you can just `config_file = DEFAULT_JSON;`.

config/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 set(default_config_filename "rv64d.json")
+set(config_schema "schema.json")
 file(READ ${default_config_filename} DEFAULT_JSON)
+file(READ ${config_schema} JSON_SCHEMA)
```
Timmmm: The schema is generated by the build process so we don't need to check it in. Also that makes it slightly awkward because you can't use `configure_file` to turn it into a C header.  Really we want [this](https://en.cppreference.com/w/c/preprocessor/embed) but it's not available until C23.

Probably the way to do this is something like

```
add_custom_command(
  DEPENDS "default_config.h.in" "generate_default_config_h.cmake" "rv64d.json" "${schema_file}"
  COMMAND cmake -P generate_default_config_h.cmake
  OUTPUTS "default_config.h"
  VERBATIM
)
```

and in `generate_default_config_h.cmake` put

```
set(default_config_filename "rv64d.json")
set(config_schema "schema.json")
file(READ ${default_config_filename} DEFAULT_JSON)
file(READ ${config_schema} JSON_SCHEMA)
configure_file("default_config.h.in" "default_config.h" @ONLY)
```

Arielfoever:
```
[ariel@archlinux sail-riscv]$ build/c_emulator/sail_riscv_sim tests/rv64si-p-wfi -c config/rv64d-mod.json
Stage1
nlohmann::json failed to parse the config file
Parse error:[json.exception.parse_error.101] parse error at line 268, column 1: syntax error while parsing object - unexpected end of input; expected '}'
[ariel@archlinux sail-riscv]$ nano config/rv64d-mod.json 
[ariel@archlinux sail-riscv]$ build/c_emulator/sail_riscv_sim tests/rv64si-p-wfi -c config/rv64d-mod.json
Stage1
Stage2
Stage3
ailed to validate config.
Error #1
  context: <root>[extensions][Sv57][supported]
  desc:    Value type not permitted by 'type' constraint.
Error #2
  context: <root>[extensions][Sv57]
  desc:    Failed to validate against schema associated with property name 'supported'.
Error #3
  context: <root>[extensions]
  desc:    Failed to validate against schema associated with property name 'Sv57'.
Error #4
  context: <root>
  desc:    Failed to validate against schema associated with property name 'extensions'.
```

Arielfoever:
Everything looks good except the CI.

Alasdair:
Note that currently Sail assumes the latest version of JSON schema 2020-12. Some libraries are older and don't support that. See https://json-schema.org/tools

.gitmodules
Arielfoever: They are not single file header except you bundle them manually. I don't like it since it's hard to upgrade.

c_emulator/riscv_sim.cpp
Arielfoever: Done

c_emulator/riscv_sim.cpp
Arielfoever: Reject due to `sail_config_set_string` need a `const char*` even though I don't like it.

Timmmm: You can use `.c_str()` to get a `const chat*` from a `std::string`.

.gitmodules
Timmmm: Ah yeah only nlohmann is (check the GitHub releases).

Still I think we should avoid submodules at almost all costs. We could use `git subtree` or CMake's `external_project_add` instead.

config/CMakeLists.txt
Arielfoever: Not understand. What is the purpose?

.gitmodules
arichardson: I agree, let's use external_project_add() unless it's a single header that can be imported easily

config/CMakeLists.txt
Arielfoever: done

CMakeLists.txt
```diff
@@ -152,6 +152,12 @@ add_subdirectory("sail_runtime")
 # Sail configuration
 add_subdirectory("config")
 
+#JSON validation
+set(valijson_BUILD_TESTS OFF CACHE BOOL "don't build valijson tests")
```
arichardson: Since both of these have CMake support, how about using [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html)?

nadime15: Is there a good reason why some of our external libs are fetched dynamically and some are just part of the repo? We have softfloat and are about to add CLI11 as well now as part of the repo.

.gitmodules
Arielfoever: I don't like subtree but I did it.

CMakeLists.txt
Arielfoever: > Since both of these have CMake support, how about using [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html)?

done

config/CMakeLists.txt
Timmmm: I don't think you've pushed any changes?

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "dependencies/valijson"]
+    path = dependencies/valijson
+    url = https://github.com/tristanpenman/valijson
+[submodule "dependencies/nlohmann-json"]
+    path = dependencies/nlohmann-json
+    url = https://github.com/nlohmann/json.git
```
Timmmm: Need to remove this change.

CMakeLists.txt
```diff
@@ -146,6 +146,34 @@ include(GNUInstallDirs)
...
+        GIT_REPOSITORY https://github.com/tristanpenman/valijson.git
+        GIT_TAG master
+    )
+    message(STATUS "Downloading valijson from remote...")
+    FetchContent_MakeAvailable(valijson)
+endif()
+find_package(nlohmann_json QUIET)
+if(NOT nlohmann_json_FOUND)
+    include(FetchContent)
+    # 配置要下载的库信息
```
Timmmm: I wish I could read Chinese, but... can you change this to English?

config/CMakeLists.txt
Arielfoever: forgot to push :(

.gitmodules
Arielfoever: removed

jordancarlin (commented):
While `FetchContent` is definitely better than a submodule, I don't see a reason not include the header files directly in the repo in this case (similar to what we are doing for CLI11).

Nlohmann json is offered as a single header file, so direct integration of that one should be easy. 

Valijson has a built in script for creating a bundled header with support for a single parser, so should be straight forward to integrate and update as needed as well. https://github.com/tristanpenman/valijson?tab=readme-ov-file#bundled-headers:
```
./bundle.sh nlohmann_json > valijson_nlohmann_bundled.hpp
```

Arielfoever:
> While `FetchContent` is definitely better than a submodule, I don't see a reason not include the header files directly in the repo in this case (similar to what we are doing for CLI11).
> 
> Nlohmann json is offered as a single header file, so direct integration of that one should be easy.
> 
> Valijson has a built in script for creating a bundled header with support for a single parser, so should be straight forward to integrate and update as needed as well. https://github.com/tristanpenman/valijson?tab=readme-ov-file#bundled-headers:
> 
> ```
> ./bundle.sh nlohmann_json > valijson_nlohmann_bundled.hpp
> ```

No I don't like it. It will cause difficulty of bumping version.

Timmmm:
You just have to run the script again surely? I agree with Jordan. It's not ideal but this is C++ and the best you can hope for is "not ideal".

I don't imagine we will be updating these very often.

Arielfoever:
> You just have to run the script again surely? I agree with Jordan. It's not ideal but this is C++ and the best you can hope for is "not ideal".
> 
> I don't imagine we will be updating these very often.

yeah but I'd keep the possibility of debugging and adjustment is retained.

Arielfoever:
> You just have to run the script again surely? I agree with Jordan. It's not ideal but this is C++ and the best you can hope for is "not ideal".
> 
> I don't imagine we will be updating these very often.

Reject due to the bundled `valijson_nlohmann_bundled.hpp` is based on the installed system lib `nlohmann/json.hpp`. I'd suggest to put it on agenda.

Timmmm:
Are you sure? The script doesn't look like it includes nlohmann to me.

Arielfoever:
> Are you sure? The script doesn't look like it includes nlohmann to me.

So you need to install nlohmann/json or you will get no such file.

Timmmm:
I tried it. It doesn't require nlohmann or bundle it.

Arielfoever:
> I tried it. It doesn't require nlohmann or bundle it.

The bundled `valijson_nlohmann_bundled.hpp` line 10208:

```cpp
#include <nlohmann/json.hpp>
```

arichardson:
> Note that currently Sail assumes the latest version of JSON schema 2020-12. Some libraries are older and don't support that. See https://json-schema.org/tools

Hmm looking at that table it seems like the only option is https://github.com/danielaparker/jsoncons. I guess one advantage of pulling that in would be that it also supports binary formats like cbor which might be useful for trace output.

Timmmm:
Interesting. I don't know if you were in the meeting yesterday but we were thinking we would:

1. Temporarily bundle [Boon](https://github.com/santhosh-tekuri/boon) in the binary download and try to call that to do the validation.
2. Long term, change the Sail compiler to generate a C validation function. This feels nicer since it you can't accidentally avoid it, don't have to add a JSON validation tool/library, and also because it can validate Sail constraints that can't be expressed in JSON Schema (e.g. `elen <= vlen`). It requires changes to the Sail compiler though.

Maybe jsoncons is a better option than Boon for step 1?

pmundkur:
jsoncons seems like a fairly large dependency; it seems simpler to just call boon.  if we need to have binary trace outputs, we could revisit this.

Timmmm:
Yeah fair enough. I think for binary trace we will want to use something more efficient than CBOR anyway.

Arielfoever:
Convert to draft due to CLI11 has not be implemented.

Arielfoever:
> Interesting. I don't know if you were in the meeting yesterday but we were thinking we would:
> 
>     1. Temporarily bundle [Boon](https://github.com/santhosh-tekuri/boon) in the binary download and try to call that to do the validation.
> 
>     2. Long term, change the Sail compiler to generate a C validation function. This feels nicer since it you can't accidentally avoid it, don't have to add a JSON validation tool/library, and also because it can validate Sail constraints that can't be expressed in JSON Schema (e.g. `elen <= vlen`). It requires changes to the Sail compiler though.
> 
> 
> Maybe jsoncons is a better option than Boon for step 1?

Boon seems has conflict on license. Could you please ask a lawyer?

Timmmm:
Boon is MIT or Apache licensed. That basically means you can do anything. It should be fine.

Arielfoever:
> Boon is MIT or Apache licensed. That basically means you can do anything. It should be fine.

good idea but I like jsoncons.

Arielfoever:
ready to review thanks to jsoncons. see https://github.com/Arielfoever/sail-riscv/tree/backup/jsonval/valijson branch if you want to see valijson.

TODO: add default value to schema and then patch config.

edolnx:
> jsoncons seems like a fairly large dependency; it seems simpler to just call boon. if we need to have binary trace outputs, we could revisit this.

I understand that this is a large dependency. As we've been going down the path of an external tool that seems to also be a large dependency as well. I am at this point inclined to agree with @Arielfoever that having a large dep that is in the language we already use is better than a lot of the problems we are having bringing in an external tool.

Timmmm:
Yeah I think I agree. This does seem less awkward than having to deal with calling an external program.

It's not a big change in terms of code so hopefully we can easily delete it one day when Sail generates a proper config validation function. I might have a go at that at some point.

Arielfoever:
Rebase done.

Arielfoever:
What's wrong with CI?

config/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ foreach (CONFIG__BASE__XLEN IN ITEMS 32 64)
 endforeach()
 
 file(READ "${CMAKE_CURRENT_BINARY_DIR}/rv64d_v256_e64.json" DEFAULT_JSON)
+file(READ "${CMAKE_BINARY_DIR}/sail_riscv_config_schema.json" JSON_SCHEMA)
```
arichardson, line 24:
```
CMake Error at config/CMakeLists.txt:24 (file):
  file failed to open for reading (No such file or directory):

    /home/runner/work/sail-riscv/sail-riscv/build/sail_riscv_config_schema.json

```
You'll need to make sure this file exist before reading it

pmundkur: The schema file is generated during the build and is not present during cmake's generation phase.

c_emulator/riscv_sim.cpp
```diff
@@ -567,13 +623,11 @@ int inner_main(int argc, char **argv)
...
-  } else {
-    sail_config_set_string(DEFAULT_JSON);
-  }
-
-  init_sail_configured_types();
+  sail_config_set_string(validate_json(config_file).c_str());
+  sail_set_abstract_xlen();
+  sail_set_abstract_vlen_exp();
+  sail_set_abstract_ext_d_supported();
+  sail_set_abstract_elen_exp();
```
pmundkur, line 630: This seems like a bad rebase.

CMakeLists.txt
```diff
@@ -153,18 +153,46 @@ option(COVERAGE "Compile with Sail coverage collection enabled.")
...
 
-# Sail configuration
-add_subdirectory("config")
-
 # Sail model generated C code.
 add_subdirectory("model")
 
+# Sail configuration
+add_subdirectory("config")
+
```
pmundkur, line 195: Is there a reason for this reordering?

jordancarlin:
Superseded by #1291.

PR #1165, opened 2025-07-24
Improve vector nfields handling

Timmmm:
Instead of storing the raw bits in the decoded `instruction` store the actual integer value of `nfields`. This also allows us to handle the fact that non-power of two integers are reserved for vector whole register load/stores (`VLRETYPE`/`VSRETYPE`). Previously that was handled in the `execute()` clause, which is unfortunate because then we can decode those reserved instructions but not disassemble them. (And we probably shouldn't be decoding instructions that are *always* reserved anyway.)

### DIFF
--- a/model/riscv_insts_vext_mem.sail
+++ b/model/riscv_insts_vext_mem.sail
@@ -11,36 +11,6 @@
 /* Chapter 7: Vector Loads and Stores                                              */
 /* ******************************************************************************* */
 
-
-mapping nfields_int : bits(3) <-> nfields = {
-  0b000     <-> 1,
-  0b001     <-> 2,
-  0b010     <-> 3,
-  0b011     <-> 4,
-  0b100     <-> 5,
-  0b101     <-> 6,
-  0b110     <-> 7,
-  0b111     <-> 8
-}
-
-mapping nfields_string : bits(3) <-> string = {
-  0b000     <-> "",
-  0b001     <-> "seg2",
-  0b010     <-> "seg3",
-  0b011     <-> "seg4",
-  0b100     <-> "seg5",
-  0b101     <-> "seg6",
-  0b110     <-> "seg7",
-  0b111     <-> "seg8"
-}
-
-mapping nfields_int_string : bits(3) <-> string = {
-  0b000 <-> "1",
-  0b001 <-> "2",
-  0b011 <-> "4",
-  0b111 <-> "8",
-}
-
 mapping vlewidth_bitsnumberstr : vlewidth <-> string = {
   VLE8      <-> "8",
   VLE16     <-> "16",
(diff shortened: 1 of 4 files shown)


model/riscv_insts_vext_mem.sail
```diff
@@ -11,34 +11,40 @@
...
-  0b000 <-> "1",
-  0b001 <-> "2",
-  0b011 <-> "4",
-  0b111 <-> "8",
+mapping nfields_pow2_string : nfields_pow2 <-> string = {
+  1 <-> "1",
+  2 <-> "2",
+  4 <-> "4",
+  8 <-> "8",
 }
```
pmundkur: LGTM, but these helpers should probably also move to `riscv_vext_types`

Timmmm: Yeah I was on the fence about that but I think you're right. I'll move them.

Timmmm: Done.

model/riscv_vext_types.sail
```diff
@@ -0,0 +1,53 @@
...
+  0b011 <-> 4,
+  0b111 <-> 8,
+}
+
+mapping nfields_pow2_string : nfields_pow2 <-> string = {
+  1 <-> "1",
+  2 <-> "2",
+  4 <-> "4",
+  8 <-> "8",
+}
```
nadime15, line 53: I wonder if it makes sense to add this to the prelude and generalize it for other use cases, maybe in a follow-up PR.

Timmmm: Is it possible to generalise it? I guess you can do something like `pow2_string_4`, `pow2_string_8`, etc?

jordancarlin (approved):

model/riscv_vext_types.sail
nadime15: Yeah I was thinking of something like:

```
mapping pow2_string : pow2 <-> string = {
  1 <-> "1",
  2 <-> "2",
  4 <-> "4",
  8 <-> "8",
  16 <-> "16",
   ...
}
```
then you define:
```
type range_pow2('q) = 'q in { 1, 2, 4, 8, 16, ... }
type pow2 = { 'q, range_pow2('q). int('q) }
```

And then define `nfields_range_pow2` which is a subtype of `range_pow2('q) `just with additional constraints, or does make use of it somehow like this:

```
type nfields_range_pow2('q) = range_pow2('q) & 'q <= 8
type nfields_pow2 = { 'q, nfields_range_pow2('q). int('q) }
```

Timmmm: I'm not sure that would work because when you have `SOME_INSTRUCTION : (nfields_pow2)` and you try to do

```
mapping clause assembly = SOME_INSTRUCTION(nf) <-> "some.instruction" ^ spc() ^ pow2_string(nf)
```

Sail will complain that 16 is unmatched - i.e. it can't map `some.instruction 16` to a `SOME_INSTRUCTION`.

Arguably we could change `assembly` to a `scattered function` that only works as a disassembler. It's not like the assembler really works (I assume). Then I think it would be ok:

```
function clause disassembler (SOME_INSTRUCTION(nf)) = "some.instruction" ^ spc() ^ pow2_string(nf)
```

nadime15: > Sail will complain that 16 is unmatched - i.e. it can't map some.instruction 16 to a SOME_INSTRUCTION.

Yeah you are right, I couldnt get it to work. I thought maybe using `nfields_pow2_string` instead of `pow2_string` could be a workaround, but that probably just makes things worse...

PR #1168, opened 2025-07-25
Simplify model/CMakeLists.txt by removing `variant`

Timmmm:
Through various previous simplifications this had devolved into the for-switch antipattern. We can remove the for loop and simply execute things in sequence.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -19,483 +19,481 @@ set(sail_common
 
 set(project_file "riscv.sail_project")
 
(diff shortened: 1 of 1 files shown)


Timmmm:
Dunno why Github's diff is so bad here... I recommend you review with the hide whitespace option since 90% of this is just changing indentation:

image

jordancarlin (approved):

pmundkur (approved):
Much nicer now!

PR #1170, opened 2025-07-26
Add gating for vector instructions.

wwwwwwOwO:
Improve the constraint checks when decoding vector instructions. 

- For Load/Store, check segment element widths against the supported EEW.
- For others, check supported SEW. In particular, the SEWs supported by the Widening/Narrowing instructions are different from other instructions.

To meet the SEW requirement, I separated out the floating-point reduction instructions.

### DIFF
--- a/model/extensions/V/vext_fp_utils_insts.sail
+++ b/model/extensions/V/vext_fp_utils_insts.sail
@@ -15,8 +15,7 @@
 // 2. Valid floating-point rounding mode
 function valid_fp_op(SEW : sew_bitsize, rm_3b : bits(3)) -> bool = {
   // SEW=8 is valid for conversions between INT8 and FP16(vfncvt[.rtz].x[u].f.w, etc.)
-  // 128-bit floating-point values will be supported in future extensions.
-  let valid_sew = (SEW >= 8 & SEW <= 128);
+  let valid_sew = (SEW >= 8 & SEW <= 64);
   let valid_rm = not(rm_3b == 0b101 | rm_3b == 0b110 | rm_3b == 0b111);
   valid_sew & valid_rm
 }


jordancarlin:
Now that #1163 is merged, this seems fairly high priority. Right now the vector configuration does not actually match with which instructions/element widths are allowed.

pmundkur:
@wwwwwwOwO It looks like the floating point vector tests are failing for 64-bit ELEN.

wwwwwwOwO:
Hi all, I got into a problem:
For the 4 instructions `vmulh.vv/vx, vmulhu.vv/vx, vmulhsu.vv/vx, and vsmul.vv/vx`, which are integer operations, only supported by full V extension when SEW=64, but `currentlyEnabled(Ext_V)`  would check the `mstatus[FS]` which are only matter for floating-point operations, then it fail tests for these instructions.

From implementation perspective, it seems only the vector FP instructions should depend on F/D and mstatus[FS], while integer vector ops should remain available regardless of FP state. The V extension exceeds Zve64d only by the above 4 integer instructions (the function`currentlyEnabled(Ext_V)` is only used by them). 

Would it be possible to change the dependency of `currentlyEnabled(Ext_V)` from `currentlyEnabled(Ext_Zve64d)` to `currentlyEnabled(Ext_Zve64x)`? This change should be fine, since the dependency between the Zve and V has been validated during the configuration stage.

nadime15:
I just had a quick look and did not think this through, but one idea I had is that instead of modifying the dependencies directly, we could adjust the encdec clause and do something like this (it wont compile, but just to show the idea):

```
mapping clause encdec = MVXTYPE(funct6, vm, vs2, rs1, vd)
  <-> encdec_mvxfunct6(funct6) @ vm @ encdec_vreg(vs2) @ encdec_reg(rs1) @ 0b110 @ encdec_vreg(vd) @ 0b1010111
  when match funct6 {
    MVX_VMULH | MVX_VMULHU | MVX_VMULHSU => 
      currentlyEnabled(Ext_Zve64x) & (get_sew() != 64 | (hartSupports(Ext_V) & misa[V] == 0b1 & currentlyEnabled(Ext_Zvl128b) & mstatus[VS] != 0b00)),
    _ => 
      currentlyEnabled(Ext_Zve64x)
  }
```

Like that we would skip the check for Zve64d and still make sure V is enabled.

pmundkur:
This looks pretty close.  I'll push a slight cleanup to make it easier to review.  Perhaps we could pull this in for the release if there are no issues in review.

model/extensions/V/vext_fp_utils_insts.sail
```diff
@@ -14,8 +14,9 @@
 // 1. Valid element width of floating-point numbers
 // 2. Valid floating-point rounding mode
 function valid_fp_op(SEW : sew_bitsize, rm_3b : bits(3)) -> bool = {
-  // 128-bit floating-point values will be supported in future extensions
-  let valid_sew = (SEW >= 16 & SEW <= 128);
+  /* SEW=8 is valid for conversions between INT8 and FP16(vfncvt[.rtz].x[u].f.w, etc.)
+     128-bit floating-point values will be supported in future extensions */
```
nadime15:
```suggestion
  // SEW=8 is valid for conversions between INT8 and FP16(vfncvt[.rtz].x[u].f.w, etc.)
  // 128-bit floating-point values will be supported in future extensions
```

model/extensions/V/vext_fp_utils_insts.sail
```diff
@@ -14,8 +14,9 @@
 // 1. Valid element width of floating-point numbers
 // 2. Valid floating-point rounding mode
 function valid_fp_op(SEW : sew_bitsize, rm_3b : bits(3)) -> bool = {
-  // 128-bit floating-point values will be supported in future extensions
-  let valid_sew = (SEW >= 16 & SEW <= 128);
+  // SEW=8 is valid for conversions between INT8 and FP16(vfncvt[.rtz].x[u].f.w, etc.)
+  // 128-bit floating-point values will be supported in future extensions.
+  let valid_sew = (SEW >= 8 & SEW <= 128);
```
jordancarlin: Interest seems to have moved away from 128-bit floating point. What about changing this to have a max SEW of 64 and we can add 128 bit back later if a new extension is ever ratified?

jordancarlin (approved):

wwwwwwOwO:
So glad this got approved! Thanks everyone😊

PR #1171, opened 2025-07-26
Add DYN (Position-Independent Executable file) support

Arielfoever:
add a `-d` parameter for DYN files. Then give a offset, mostly `0x80000000` like openSBI.


os-boot/linux/Makefile
```diff
@@ -72,7 +72,7 @@ LIMIT_INSTRUCTIONS := 20000000
 # Run with the Sail emulator from this repo.
 .PHONY: sail
 sail: build/sail.dtb $(elf) $(SAIL_SIM)
-	$(SAIL_SIM) --no-trace -p -l $(LIMIT_INSTRUCTIONS) --device-tree-blob build/sail.dtb $(elf)
+	$(SAIL_SIM) --no-trace -p -l $(LIMIT_INSTRUCTIONS) --device-tree-blob build/sail.dtb $(elf) -d 0x80000000
```
arichardson, line 75: It would be nice to make this the default behaviour when passed a PIE executable

jrtc27: Yeah if you don't specify it it should use a sane default like the start of DRAM

c_emulator/riscv_sim.cpp
```diff
@@ -128,7 +128,8 @@ static struct option options[] = {
 #endif
     {"print-device-tree",              no_argument,       0, OPT_PRINT_DTS      },
     {"print-isa-string",               no_argument,       0, OPT_PRINT_ISA      },
-    {0,                                0,                 0, 0                  }
+    {0,                                0,                 0, 0                  },
+    {"dynamic",                        required_argument, 0, 'd'                }
```
jrtc27: Uh how does this option work if it's after the terminator? Presumably the short option works because that's separate, but the long option doesn't (the 'd' here is to make the long option report the same result as for the short one parsed by the alphabet soup string).

c_emulator/riscv_sim.cpp
```diff
@@ -385,7 +404,12 @@ uint64_t load_sail(char *f, bool main_file)
   bool is32bit;
   uint64_t entry;
   uint64_t begin_sig, end_sig;
-  load_elf(f, &is32bit, &entry);
+  if (load_offset) {
+    // dynamic load
```
jrtc27, line 408: Uh ET_EXEC should ignore it (or give an error, though that's annoying if loading multiple ELF files, e.g. an ET_DYN OpenSBI and an ET_EXEC kernel)

os-boot/linux/Makefile
Arielfoever: > It would be nice to make this the default behaviour when passed a PIE executable

No, you need to specify the offset or the loader will reject to load DYN. for openSBI it can be `0x80000000` but different value may be work too. 

sail didn't provide the API for detecting whether it is a DYN or EXEC.

c_emulator/riscv_sim.cpp
Arielfoever: so how could I fix it(?

c_emulator/riscv_sim.cpp
Arielfoever: do you mean we load an `fw_jump.elf` then with a kernel?

c_emulator/riscv_sim.cpp
jrtc27: Well you could start by testing this option and observing that it doesn't work. Then fix it by swapping lines 131 and 132.

c_emulator/riscv_sim.cpp
jrtc27: fw_jump vs fw_dynamic is an orthogonal concern (referring to whether OpenSBI hard-codes the location at which the payload is loaded) to whether the kernel being loaded requests a fixed address (that may or may not match OpenSBI's expectation).

os-boot/linux/Makefile
jrtc27: Ok so Sail's load_elf_at_offset is really load_elf_with_dyn(/pie)_offset; that is it will ignore the offset if the provided binary is not ET_DYN. This means you should just use load_elf_at_offset unconditionally with a sensible offset, and if it's an ET_EXEC it will ignore the offset and load at the object's fixed address.

c_emulator/riscv_sim.cpp
```diff
@@ -397,14 +421,17 @@ uint64_t load_sail(char *f, bool main_file)
     fprintf(stderr, "Unable to locate tohost symbol; disabling HTIF.\n");
     rv_enable_htif = false;
   } else {
+    rv_htif_tohost += load_offset;
```
jrtc27, line 424: These need a way to know whether load_elf_at_offset used load_offset or not; at first glance I don't see a way to tell. I suspect Sail needs a lookup_sym_with_(dyn/pie_)offset that is like load_elf_at_offset in conditionally including the PIE (dynamic) offset.

c_emulator/riscv_sim.cpp
Arielfoever: everything looks good.

c_emulator/riscv_sim.cpp
Arielfoever: currently the sail gives no api on detect whether it is DYN or EXEC.

os-boot/linux/Makefile
Timmmm: Slightly unfortunate naming. We should definitely add a comment to clarify that behaviour.

Timmmm: The other option is that we just use our own ELF loading code. That's what I ended up doing (using ELFIO).

nadime15:
I prefer and suggest we keep the `-d` option reserved for debug mode, and use a different one for this PR.

Arielfoever:
> I prefer and suggest we keep the `-d` option reserved for debug mode, and use a different one for this PR.

which char do you like?

Timmmm:
In the other PR to update the CLI we've removed short options so when you rebase on that I'd just remove it.

Also I don't think having a single flag actually works because we support loading multiple elfs. We might need some syntax like `foo.elf@0x100 bar.elf@0x400`.

Arielfoever:
Convert to draft since it needs https://github.com/riscv/sail-riscv/pull/1177 merging.

Arielfoever:
Do we still need this pr? @Timmmm

Timmmm:
Yes, if we want to support loading executables at arbitrary offsets. I don't know how much demand there is for that though. What was the original motivation for this?

pmundkur:
I think the original motivation was loading the default OpenSBI payload.

allenjbaum:
If I'm understanding this correctly (not a sure thing), the dynamic part lets the Sail model load ELFs at addresses that are in .json files, as opposed to some fixed address. The .S files should themselves be position independent (because they need to be run.
We want that capability so that we can configure the Sail model to match what a DUT is actually executing
(so exceptions report exactly the same EPC and TVAL, etc)

Arielfoever:
> I think the original motivation was loading the default OpenSBI payload.

That's correct.

PR #1172, opened 2025-07-26
riscv_sim.cpp: add SIGINT processor.

Arielfoever:


Timmmm:
What's the motivation here? Adding explicit signal handlers is usually something that should be avoided if at all possible.

Arielfoever:
> What's the motivation here? Adding explicit signal handlers is usually something that should be avoided if at all possible.

to generate the show execution times on ctrl-c.

```
[ariel@archlinux sail-riscv]$ build/c_emulator/sail_riscv_sim os-boot/linux/fw_payload.elf --no-trace -p --device-tree-blob os-boot/linux/build/sail.dtb -l 1000000000
will show execution times on completion.
using os-boot/linux/build/sail.dtb as DTB file.
Read 1655 bytes of DTB from os-boot/linux/build/sail.dtb.
Running file os-boot/linux/fw_payload.elf.
ELF Entry @ 0x80000000
HTIF located at 0x80042a78
^CInitialization:   144 msecs
Execution:        1909 msecs
Instructions:     116907
Perf:             61.240 Kips
```

c_emulator/riscv_sim.cpp
```diff
@@ -684,7 +686,12 @@ void init_logs()
   }
 #endif
 }
-
+void close_sig_handler(int sig)
+{
+  finish(1);
+  fprintf(stderr, "SIGINT %d detected, closd.", sig);
```
arichardson, line 692: Calling stdio functions from signal handlers is not safe.

Timmmm:
Yeah we should avoid touching signal handlers at all costs. They're *extremely* error prone (see Alex's comment for example) and this doesn't seem like a big enough benefit.

jordancarlin:
Closing this due to agreement that this is not something we want to do and lack of author response.

PR #1176, opened 2025-07-29
Fix softfloat debug builds

Timmmm:
I'm not sure exactly why softfloat is set up like this, but with `INLINE` only defined as `inline`, then it was possible in debug builds to get undefined references to some of the inlined functions (e.g. `softfloat_countLeadingZeros32`). Apparently in C99 `inline` functions *might* not actually get emitted; you need to use `static inline`. I don't fully understand why but ChatGPT called this "The C99 Inline Mess" when I asked about it so probably not worth digging!

The old version of `platform.h` also used `static inline`.

### DIFF
--- a/dependencies/softfloat/berkeley-softfloat-3/build/Linux-x86_64-GCC/platform.h
+++ b/dependencies/softfloat/berkeley-softfloat-3/build/Linux-x86_64-GCC/platform.h
@@ -41,7 +41,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 /*----------------------------------------------------------------------------
 *----------------------------------------------------------------------------*/
 #ifdef __GNUC_STDC_INLINE__
-#define INLINE inline
+#define INLINE static inline
 #else
 #define INLINE extern inline
 #endif
@@ -51,4 +51,3 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #define SOFTFLOAT_BUILTIN_CLZ 1
 #define SOFTFLOAT_INTRINSIC_INT128 1
 #include "opts-GCC.h"
-


nadime15:
Is there a way to reproduce your error? Because on my side everything seems to work so far

dependencies/softfloat/berkeley-softfloat-3/build/Linux-x86_64-GCC/platform.h
```diff
@@ -41,7 +41,7 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 /*----------------------------------------------------------------------------
 *----------------------------------------------------------------------------*/
 #ifdef __GNUC_STDC_INLINE__
-#define INLINE inline
+#define INLINE static inline
```
nadime15, line 44: I am not sure (I checked online and got mixed answers) but could we just overwrite `#define INLINE `  through CMake via: 

```
target_compile_definitions(softfloat PRIVATE
    INLINE="static inline"  <-------------------------
    SOFTFLOAT_ROUND_ODD
    INLINE_LEVEL=5
    SOFTFLOAT_FAST_DIV32TO16
    SOFTFLOAT_FAST_DIV64TO32
    SOFTFLOAT_FAST_INT64
)
```
This might take precedence over header definitions (?)

That way, we wouldnt need to modify platform.h.

Timmmm: Unfortunately then you get a ton of warnings about it being defined twice.

Timmmm:
It's probably highly compiler dependent, but I'm using `c++ (GCC) 11.5.0 20240719 (Red Hat 11.5.0-5)` and building with `-DCMAKE_BUILD_TYPE=Debug`.

pmundkur (approved):

nadime15:
Ok I ran into the same problem after switching from `RelWithDebInfo` to `Debug`.

PR #1177, opened 2025-07-29
Load ELFs via ELFIO and add symbols to the trace output

Timmmm:
Add ELFIO dependency and use it to load ELFs into memory instead of using Sail's built-in ELF support. This gives us greater control, e.g. it will allow supporting position independent ELFs in future. It also allows printing symbols in the log, which is implemented here.

Example output:

```
$ c_emulator/sail_riscv_sim test/2025-07-16/riscv-tests/rv64mi-p-breakpoint --trace-instr
ELF Entry @ 0x80000000
HTIF located at 0x80001000
begin_signature: 0x80002000
end_signature: 0x80002010
[0] [M]: 0x0000000000001000 (0x00000297) auipc x5, 0x0                                                    
[1] [M]: 0x0000000000001004 (0x02028593) addi x11, x5, 0x20                                               
[2] [M]: 0x0000000000001008 (0xF1402573) csrrs x10, mhartid, x0                                           
[3] [M]: 0x000000000000100C (0x0182B283) ld x5, 0x18(x5)                                                  
[4] [M]: 0x0000000000001010 (0x00028067) jalr x0, 0x0(x5)                                                 
[5] [M]: 0x0000000080000000 (0x0540006F) jal x0, 0x54                                                     _start+0
[6] [M]: 0x0000000080000054 (0x00000093) addi x1, x0, 0x0                                                 reset_vector+0
[7] [M]: 0x0000000080000058 (0x00000113) addi x2, x0, 0x0                                                 reset_vector+4
[8] [M]: 0x000000008000005C (0x00000193) addi x3, x0, 0x0                                                 reset_vector+8
[9] [M]: 0x0000000080000060 (0x00000213) addi x4, x0, 0x0                                                 reset_vector+12
[10] [M]: 0x0000000080000064 (0x00000293) addi x5, x0, 0x0                                                 reset_vector+16
[11] [M]: 0x0000000080000068 (0x00000313) addi x6, x0, 0x0                                                 reset_vector+20
[12] [M]: 0x000000008000006C (0x00000393) addi x7, x0, 0x0                                                 reset_vector+24
[13] [M]: 0x0000000080000070 (0x00000413) addi x8, x0, 0x0                                                 reset_vector+28
[14] [M]: 0x0000000080000074 (0x00000493) addi x9, x0, 0x0                                                 reset_vector+32
[15] [M]: 0x0000000080000078 (0x00000513) addi x10, x0, 0x0                                                reset_vector+36
[16] [M]: 0x000000008000007C (0x00000593) addi x11, x0, 0x0                                                reset_vector+40
[17] [M]: 0x0000000080000080 (0x00000613) addi x12, x0, 0x0                                                reset_vector+44
[18] [M]: 0x0000000080000084 (0x00000693) addi x13, x0, 0x0                                                reset_vector+48
[19] [M]: 0x0000000080000088 (0x00000713) addi x14, x0, 0x0                                                reset_vector+52
[20] [M]: 0x000000008000008C (0x00000793) addi x15, x0, 0x0                                                reset_vector+56
[21] [M]: 0x0000000080000090 (0x00000813) addi x16, x0, 0x0                                                reset_vector+60
[22] [M]: 0x0000000080000094 (0x00000893) addi x17, x0, 0x0                                                reset_vector+64
[23] [M]: 0x0000000080000098 (0x00000913) addi x18, x0, 0x0                                                reset_vector+68
[24] [M]: 0x000000008000009C (0x00000993) addi x19, x0, 0x0                                                reset_vector+72
[25] [M]: 0x00000000800000A0 (0x00000A13) addi x20, x0, 0x0                                                reset_vector+76
[26] [M]: 0x00000000800000A4 (0x00000A93) addi x21, x0, 0x0                                                reset_vector+80
[27] [M]: 0x00000000800000A8 (0x00000B13) addi x22, x0, 0x0                                                reset_vector+84
[28] [M]: 0x00000000800000AC (0x00000B93) addi x23, x0, 0x0                                                reset_vector+88
[29] [M]: 0x00000000800000B0 (0x00000C13) addi x24, x0, 0x0                                                reset_vector+92
[30] [M]: 0x00000000800000B4 (0x00000C93) addi x25, x0, 0x0                                                reset_vector+96
[31] [M]: 0x00000000800000B8 (0x00000D13) addi x26, x0, 0x0                                                reset_vector+100
[32] [M]: 0x00000000800000BC (0x00000D93) addi x27, x0, 0x0                                                reset_vector+104
[33] [M]: 0x00000000800000C0 (0x00000E13) addi x28, x0, 0x0                                                reset_vector+108
[34] [M]: 0x00000000800000C4 (0x00000E93) addi x29, x0, 0x0                                                reset_vector+112
[35] [M]: 0x00000000800000C8 (0x00000F13) addi x30, x0, 0x0                                                reset_vector+116
[36] [M]: 0x00000000800000CC (0x00000F93) addi x31, x0, 0x0                                                reset_vector+120
[37] [M]: 0x00000000800000D0 (0xF1402573) csrrs x10, mhartid, x0                                           reset_vector+124
[38] [M]: 0x00000000800000D4 (0x00051063) bne x10, x0, 0x0                                                 reset_vector+128
[39] [M]: 0x00000000800000D8 (0x00000297) auipc x5, 0x0                                                    reset_vector+132
[40] [M]: 0x00000000800000DC (0x01028293) addi x5, x5, 0x10                                                reset_vector+136
[41] [M]: 0x00000000800000E0 (0x30529073) csrrw x0, mtvec, x5                                              reset_vector+140
[42] [M]: 0x00000000800000E4 (0x74445073) csrrwi x0, 0x744, 0x8                                            reset_vector+144
[43] [M]: 0x00000000800000E8 (0x00000297) auipc x5, 0x0                                                    reset_vector+148
[44] [M]: 0x00000000800000EC (0x01028293) addi x5, x5, 0x10                                                reset_vector+152
[45] [M]: 0x00000000800000F0 (0x30529073) csrrw x0, mtvec, x5                                              reset_vector+156
[46] [M]: 0x00000000800000F4 (0x18005073) csrrwi x0, satp, 0x0                                             reset_vector+160
[47] [M]: 0x00000000800000F8 (0x00000297) auipc x5, 0x0                                                    reset_vector+164
[48] [M]: 0x00000000800000FC (0x02428293) addi x5, x5, 0x24                                                reset_vector+168
[49] [M]: 0x0000000080000100 (0x30529073) csrrw x0, mtvec, x5                                              reset_vector+172
[50] [M]: 0x0000000080000104 (0x0010029B) addiw x5, x0, 0x1                                                reset_vector+176
[51] [M]: 0x0000000080000108 (0x03529293) slli x5, x5, 0x35                                                reset_vector+180
[52] [M]: 0x000000008000010C (0xFFF28293) addi x5, x5, -0x1                                                reset_vector+184
[53] [M]: 0x0000000080000110 (0x3B029073) csrrw x0, pmpaddr0, x5                                           reset_vector+188
[54] [M]: 0x0000000080000114 (0x01F00293) addi x5, x0, 0x1f                                                reset_vector+192
[55] [M]: 0x0000000080000118 (0x3A029073) csrrw x0, pmpcfg0, x5                                            reset_vector+196
[56] [M]: 0x000000008000011C (0x30405073) csrrwi x0, mie, 0x0                                              reset_vector+200
[57] [M]: 0x0000000080000120 (0x00000297) auipc x5, 0x0                                                    reset_vector+204
[58] [M]: 0x0000000080000124 (0x01428293) addi x5, x5, 0x14                                                reset_vector+208
[59] [M]: 0x0000000080000128 (0x30529073) csrrw x0, mtvec, x5                                              reset_vector+212
[60] [M]: 0x000000008000012C (0x30205073) csrrwi x0, medeleg, 0x0                                          reset_vector+216
[61] [M]: 0x0000000080000130 (0x30305073) csrrwi x0, mideleg, 0x0                                          reset_vector+220
[62] [M]: 0x0000000080000134 (0x00000193) addi x3, x0, 0x0                                                 reset_vector+224
[63] [M]: 0x0000000080000138 (0x00000297) auipc x5, 0x0                                                    reset_vector+228
[64] [M]: 0x000000008000013C (0xECC28293) addi x5, x5, -0x134                                              reset_vector+232
[65] [M]: 0x0000000080000140 (0x30529073) csrrw x0, mtvec, x5                                              reset_vector+236
[66] [M]: 0x0000000080000144 (0x00100513) addi x10, x0, 0x1                                                reset_vector+240
[67] [M]: 0x0000000080000148 (0x01F51513) slli x10, x10, 0x1f                                              reset_vector+244
[68] [M]: 0x000000008000014C (0x00055C63) bge x10, x0, 0x18                                                reset_vector+248
[69] [M]: 0x0000000080000164 (0x00000293) addi x5, x0, 0x0                                                 reset_vector+272
[70] [M]: 0x0000000080000168 (0x00028A63) beq x5, x0, 0x14                                                 reset_vector+276
[71] [M]: 0x000000008000017C (0x30005073) csrrwi x0, mstatus, 0x0                                          reset_vector+296
[72] [M]: 0x0000000080000180 (0x00002537) lui x10, 0x2                                                     reset_vector+300
[73] [M]: 0x0000000080000184 (0x8005051B) addiw x10, x10, -0x800                                           reset_vector+304
[74] [M]: 0x0000000080000188 (0x30052073) csrrs x0, mstatus, x10                                           reset_vector+308
[75] [M]: 0x000000008000018C (0x00000297) auipc x5, 0x0                                                    reset_vector+312
[76] [M]: 0x0000000080000190 (0x01428293) addi x5, x5, 0x14                                                reset_vector+316
[77] [M]: 0x0000000080000194 (0x34129073) csrrw x0, mepc, x5                                               reset_vector+320
[78] [M]: 0x0000000080000198 (0xF1402573) csrrs x10, mhartid, x0                                           reset_vector+324
[79] [M]: 0x000000008000019C (0x30200073) mret                                                             reset_vector+328
[80] [M]: 0x00000000800001A0 (0x00200193) addi x3, x0, 0x2                                                 reset_vector+332
[81] [M]: 0x00000000800001A4 (0x00000517) auipc x10, 0x0                                                   reset_vector+336
[82] [M]: 0x00000000800001A8 (0x01450513) addi x10, x10, 0x14                                              reset_vector+340
[83] [M]: 0x00000000800001AC (0x30551573) csrrw x10, mtvec, x10                                            reset_vector+344
[84] [M]: 0x00000000800001B0 (0x00800593) addi x11, x0, 0x8                                                reset_vector+348
[85] [M]: 0x00000000800001B4 (0x7A55A073) csrrs x0, 0x7a5, x11                                             reset_vector+352
[86] [M]: 0x00000000800001B8 (0x30551073) csrrw x0, mtvec, x10                                             reset_vector+356
[87] [M]: 0x00000000800001BC (0x30046073) csrrsi x0, mstatus, 0x8                                          reset_vector+360
[88] [M]: 0x00000000800001C0 (0x7A001073) csrrw x0, tselect, x0                                            reset_vector+364
[89] [M]: 0x00000000800001C4 (0x7A0025F3) csrrs x11, tselect, x0                                           reset_vector+368
[90] [M]: 0x00000000800001C8 (0x10B01663) bne x0, x11, 0x10c                                               reset_vector+372
[91] [M]: 0x00000000800002D4 (0x0FF0000F) fence iorw, iorw                                                 pass+0
[92] [M]: 0x00000000800002D8 (0x00100193) addi x3, x0, 0x1                                                 pass+4
[93] [M]: 0x00000000800002DC (0x05D00893) addi x17, x0, 0x5d                                               pass+8
[94] [M]: 0x00000000800002E0 (0x00000513) addi x10, x0, 0x0                                                pass+12
[95] [M]: 0x00000000800002E4 (0x00000073) ecall                                                            pass+16
[96] [M]: 0x0000000080000004 (0x34202F73) csrrs x30, mcause, x0                                            trap_vector+0
[97] [M]: 0x0000000080000008 (0x00800F93) addi x31, x0, 0x8                                                trap_vector+4
[98] [M]: 0x000000008000000C (0x03FF0A63) beq x30, x31, 0x34                                               trap_vector+8
[99] [M]: 0x0000000080000010 (0x00900F93) addi x31, x0, 0x9                                                trap_vector+12
[100] [M]: 0x0000000080000014 (0x03FF0663) beq x30, x31, 0x2c                                               trap_vector+16
[101] [M]: 0x0000000080000018 (0x00B00F93) addi x31, x0, 0xb                                                trap_vector+20
[102] [M]: 0x000000008000001C (0x03FF0263) beq x30, x31, 0x24                                               trap_vector+24
[103] [M]: 0x0000000080000040 (0x00001F17) auipc x30, 0x1                                                   write_tohost+0
[104] [M]: 0x0000000080000044 (0xFC3F2023) sw x3, -0x40(x30)                                                write_tohost+4
[105] [M]: 0x0000000080000048 (0x00001F17) auipc x30, 0x1                                                   write_tohost+8
[106] [M]: 0x000000008000004C (0xFA0F2E23) sw x0, -0x44(x30)                                                write_tohost+12
SUCCESS
```


model/riscv_step.sail
```diff
@@ -86,6 +86,19 @@ function run_hart_waiting(step_no : int, wr: WaitReason, instbits : instbits, ex
...
+
+    var padding = "    ";
+    // TODO: This is horrifically inefficient.
+    foreach (i from string_length(instr) to 60) {
+      padding = padding ^ " ";
+    };
+
+    print_instr("[" ^ dec_str(step_no) ^ "] [" ^ to_str(cur_privilege) ^ "]: " ^ bits_str(PC) ^ " (" ^ bits_str(opcode) ^ ") " ^ instr ^ padding ^ symbolize(PC));
+  }
+
```
pmundkur: Wouldn't it be better to move this print to C++?  The other uses of `print_instr` are now misleading since the `print_instr`s in the Waiting state are not really printing instructions.  Some other prints now happen in callbacks, perhaps these should too?    This would also remove the need for the odd `symbolize` function in `riscv_platform`.

dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory("elfio")
+add_subdirectory("softfloat")
```
pmundkur, line 3: This could include `add_subdirectory("CLIUtils")`

CMakeLists.txt
```diff
@@ -143,8 +143,11 @@ option(COVERAGE "Compile with Sail coverage collection enabled.")
 
 include(GNUInstallDirs)
 
-# Softfloat support.
-add_subdirectory("dependencies/softfloat")
+# Third party dependencies.
+add_subdirectory("dependencies")
+
+# CLI11
+add_subdirectory("dependencies/CLIUtils")
```
pmundkur: This could be moved to `dependencies/CMakeLists.txt`.

model/riscv_step.sail
Timmmm: Yes I totally agree.

Arielfoever:
I like new logs.

Arielfoever:
https://github.com/Arielfoever/sail-riscv/tree/test/elfio try this. Tested on rv64 elfs. There should be more test.

c_emulator/elf_loader.cpp
```diff
@@ -0,0 +1,110 @@
...
+    Elf64_Addr value = 0;
+    Elf_Xword size = 0;
+    unsigned char bind = '\x00';
+    unsigned char type = '\x00';
+    Elf_Half section_index = 0;
+    unsigned char other = '\x00';
+
+    while (accessor.get_symbol(index, name, value, size, bind, type,
+                               section_index, other)) {
+      if ((type == STT_NOTYPE || type == STT_FUNC)
```
jrtc27: Why NOTYPE but not OBJECT (or the now rare COMMON)?

c_emulator/elf_loader.h
```diff
@@ -0,0 +1,47 @@
...
+
+  // Return the architecture (RV32/64).
+  Architecture architecture() const;
+
+  // Return the entry point (where execution should begin).
+  uint64_t entry() const;
+
+  // Return true if this is a Position-Independent Executable (the `type` is
+  // ET_DYN).
+  bool pie() const;
```
jrtc27: This appears unused

Arielfoever: it will be used in loading dyn in https://github.com/riscv/sail-riscv/pull/1177

jrtc27: So add it there?

Timmmm: Yeah fair enough. I'll remove it.

c_emulator/elf_loader.cpp
Timmmm: Good question. I couldn't find any *real* documentation for how these actually work (beyond `STT_OBJECT: the symbol is an object` notumentation). As I recall this was just based on what happened to be in the riscv-tests ELFs. Now that I think about it I guess they are NOTYPE because it's hand-written assembly. I'll add OBJECT and COMMON.

dependencies/CMakeLists.txt
Timmmm: Done

CMakeLists.txt
Timmmm: Done

c_emulator/elf_loader.cpp
Timmmm: Done

c_emulator/elf_loader.h
Timmmm: Done

c_emulator/elf_loader.cpp
jrtc27: Yeah NOTYPE is ubiquitous in hand-written assembly because it generally doesn't matter, but it does on CHERI, and it has on other architectures in the past. OBJECT is for data, FUNC is for code, COMMON is for ye olde -fcommon .bss symbols with all the weird linkage that comes of that. SECTION, FILE, TLS and GNU_IFUNC are pretty self-explanatory.

model/riscv_step.sail
Timmmm: I moved it to C. I had to add `print_log_instr` back.... which kind of sucks. I think it would be better to just add a method to `callback_if` for instruction decode, and do the logging in `riscv_callbacks_log.cpp`. What do you think?

Arguably we could change all of the `print_log()` calls in Sail to actually be `callback_if` methods. I won't do them all in this PR though.

jordancarlin: Agreed. Moving more of our logging to callbacks seems like it would be cleaner and also gives us more flexibility. For example, for a machine readable trace, it might be nice to print what instructions are executing but not include the disassembly.

arichardson:
Should we drop the elfio examples from the import? That's most of the new files and they don't really serve any purpose.

c_emulator/elf_loader.cpp
```diff
@@ -0,0 +1,105 @@
...
+                                 + std::to_string(seg->get_memory_size())
+                                 + ")");
+      }
+      // Write the data to memory.
+      writer(seg->get_physical_address(),
+             reinterpret_cast<const uint8_t *>(seg->get_data()),
+             seg->get_file_size());
+      // If the memory size is greater than the file size the remaining
+      // bytes should be zeroed.
+      if (seg->get_memory_size() > seg->get_file_size()) {
```
Arielfoever, line 67: This generate a lot of zeros and cause 

```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```

Need investigation. Sounds it only happened in Release mode rather than Debug.

I'd believe their is something wrong with `return ELF(std::move(reader));`.

c_emulator/elf_loader.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+ELF ELF::open(const std::string &filename)
+{
+  ELFIO::elfio reader;
+
+  if (!reader.load(filename)) {
+    throw std::runtime_error("File '" + filename
+                             + "' is not found or it is not an ELF file");
+  }
```
Arielfoever: Suggest remove this `open` and use `ELF(const std::string &filename);`

This will fix the `bad:alloc` bug.

```suggestion
ELF::ELF(const std::string &filename)
{
  if (!m_reader.load(filename)) {
    throw std::runtime_error("File '" + filename
                             + "' is not found or it is not an ELF file");
  }

  if (m_reader.get_machine() != ELFIO::EM_RISCV) {
    throw std::runtime_error("File '" + filename
                             + "' is not a RISC-V ELF file");
  }
}
```

c_emulator/elf_loader.h
```diff
@@ -0,0 +1,43 @@
...
+           writer) const;
+
+  // Load and return the symbol table. It isn't cached - every time you call
+  // this the entire symbol table is loaded from disk. Note only STT_FUNC and
+  // STT_NOTYPE symbols that are not SHN_UNDEF are returned.
+  std::map<std::string, uint64_t> symbols() const;
+
+private:
+  explicit ELF(ELFIO::elfio reader);
+
```
Arielfoever:
```suggestion
ELF(const std::string &filename);
private:

```

c_emulator/CMakeLists.txt
```diff
@@ -38,6 +40,8 @@ add_library(riscv_model
     riscv_softfloat.h
     config_utils.cpp
     config_utils.h
+    symbol_table.cpp
+    symbol_table.h
```
Arielfoever, line 44: How about move our elf loader to a separate place?

c_emulator/elf_loader.cpp
```diff
@@ -0,0 +1,105 @@
...
+  using namespace ELFIO;
+
+  std::map<std::string, uint64_t> symbolMap;
+
+  const section *symtab = m_reader.sections[".symtab"];
+  if (symtab != nullptr) {
+    const_symbol_section_accessor accessor(m_reader, symtab);
+    unsigned index = 0;
+    std::string name;
+    Elf64_Addr value = 0;
```
Arielfoever, line 88: Do you consider `Elf32_Addr` for RV32?

Arielfoever: And emmm do we have `Elf128_Addr`?

c_emulator/elf_loader.h
```diff
@@ -0,0 +1,43 @@
...
+  // can do anything else with them).
+  void
+  load(std::function<void(uint64_t /* address */, const uint8_t * /* data */,
+                          uint64_t /* length */)>
+           writer) const;
+
+  // Load and return the symbol table. It isn't cached - every time you call
+  // this the entire symbol table is loaded from disk. Note only STT_FUNC and
+  // STT_NOTYPE symbols that are not SHN_UNDEF are returned.
+  std::map<std::string, uint64_t> symbols() const;
```
Arielfoever, line 39: How about `unordered_map` for increased speed?

c_emulator/elf_loader.cpp
Arielfoever: ~~I'd suggest use `return ELF(reader)`;.~~

`return ELF(reader)` is unavailable since you can't copy it.

c_emulator/elf_loader.cpp
Timmmm: `get_symbol` always returns 64-bits even for 32-bit ELFs. There's no such things as a 128-bit ELF so Elf128_Addr isn't needed.

c_emulator/CMakeLists.txt
Timmmm: Could do but probably not necessary at this point I think.

c_emulator/elf_loader.cpp
Timmmm: I definitely prefer using a static function rather than a constructor. Using constructors for fallible operations is a bad idea for lots of reasons. It is a shame that `std::move()` is so hard to use correctly though - I definitely might have made a mistake.

How did you get `std::bad_alloc`? I can't seem to reproduce it. What command line, compiler, etc?

Timmmm: Ah never mind, I found the problem with ubsan. There are some raw pointers in ELFIO and it doesn't have move constructors. I'll just put it all in a unique_ptr.

Arielfoever: > I definitely prefer using a static function rather than a constructor. Using constructors for fallible operations is a bad idea for lots of reasons. It is a shame that `std::move()` is so hard to use correctly though - I definitely might have made a mistake.我绝对更喜欢使用静态函数而不是构造函数。出于多种原因，使用构造函数进行易出错的作是一个坏主意。遗憾的是 `std::move()` 很难正确使用 - 我肯定可能犯了一个错误。
> 
> How did you get `std::bad_alloc`? I can't seem to reproduce it. What command line, compiler, etc?你是怎么得到 `std::bad_alloc` 的？我似乎无法复制它。什么命令行、编译器等？

Archlinux + ocaml 5.3.0 + sail master + gcc 15.1.1

```
[ariel@archlinux sail-riscv]$ build/c_emulator/sail_riscv_sim build/test/2025-07-16/riscv-tests/rv64ui-v-lh
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
已中止                  （核心已转储）build/c_emulator/sail_riscv_sim build/test/2025-07-16/riscv-tests/rv64ui-v-lh
```

Arielfoever: According to @wychlw 's investigation, It seems that `std::move()` cause page fault in memory only in release mode.

I think it might be caused in elfio since it's using `std::forward()` to deal with memory address.

Arielfoever: Why not using `ELF()` to construct and then use `ELF.open(const string filename)` to open it. no `std::move()`

Arielfoever: > Ah never mind, I found the problem with ubsan. There are some raw pointers in ELFIO and it doesn't have move constructors. I'll just put it all in a unique_ptr.啊没关系，我发现了 ubsan 的问题。ELFIO 中有一些原始指针，它没有移动构造函数。我会把它们全部放在一个unique_ptr里。

Why not using `ELF()` to construct and then use `ELF.open(const string filename)` to open it. no `std::move()` because `elfio` don't support copy and it's safer.

Timmmm: > I think it might be caused in elfio since it's using std::forward() to deal with memory address.

No, it's because `std::vector<std::unique_ptr<section>>  sections_;` in `class elfio` stores a raw pointer to its sibling `endianess_convertor                    convertor;` and that doesn't get fixed up by a move constructor - see `elfio_section.hpp:357`.

> Why not using ELF() to construct and then use ELF.open(const string filename) to open it.

That would work, but it means an `ELF` no longer represents an open file. You can have an `ELF` that isn't open. The types are much cleaner if that isn't possible.

Timmmm: I opened a bug for this with ELFIO: https://github.com/serge1/ELFIO/issues/157

c_emulator/elf_loader.h
Timmmm: I didn't do this because when building the reverse map you may have multiple symbols for the same location, and if you use `std::unordered_map` then the one that "wins" is non-deterministic. With the current code it is the last one alphabetically that wins, which is totally arbitrary but it's at least deterministic. This also isn't performance sensitive because we use the reverse map for symbolization.

Timmmm:
> Should we drop the elfio examples from the import? 

Removed. I also removed the doc PDF since it's 300 kB and we don't really need a copy of it.

jordancarlin:
Looks like the handwritten support files need to be updated for `print_log_instr`:
```
error: LeanRV64D/RiscvStep.lean:288:19: Unknown identifier `print_log_instr`
```

model/prelude.sail
```diff
@@ -55,7 +55,12 @@ overload max = {max_int}
 
 val print_string = pure "print_string" : (string, string) -> unit
 
+// Print to the --trace log file.
 val print_log = pure {interpreter: "print_endline", c: "print_log", lem: "print_dbg", _: "print_endline"} : string -> unit
+// Print an instruction to the --trace log file. The second argument is PC which can be used for
+// smbolization. This is only supported by the C backend so other backends just do a normal print.
```
jordancarlin:
```suggestion
// symbolization. This is only supported by the C backend so other backends just do a normal print.
```

dependencies/elfio/LICENSE.txt
jordancarlin, line 1: This file needs to be installed with the other licenses.

c_emulator/riscv_sim.cpp
```diff
@@ -231,53 +232,70 @@ static void setup_options(CLI::App &app)
...
+    for (uint64_t i = 0; i < length; ++i) {
+      write_mem(address + i, data[i]);
+    }
+  });
+
   if (!main_file) {
-    /* Don't scan for test-signature/htif symbols for additional ELF files. */
-    return entry;
+    // Don't scan for test-signature/htif symbols for additional ELF files.
+    return elf.entry();
```
pmundkur: I'm not sure how often loading multiple ELFs is used, but returning here would miss scanning those for symbols.  We could just skip the test-signature/htif symbols, but add the rest?

dependencies/elfio/CMakeLists.txt
```diff
@@ -0,0 +1,149 @@
...
+    option(ELFIO_BUILD_TESTS "Build ELFIO tests" OFF)
+
+    # Generate output of compile commands during generation
+    set(CMAKE_EXPORT_COMPILE_COMMANDS, ON)
+endif()
+
+# Read version from header file
+set(version_header "elfio/elfio_version.hpp")
+file(READ ${version_header} ver)
+string(REGEX MATCH "#define ELFIO_VERSION \"([0-9\.]+)\"" _ ${ver})
```
pmundkur, line 23: This version should get added to `c_emulator/sail_riscv_version.{h,cpp}` so that it gets reported in `--build-info`.

c_emulator/riscv_sim.cpp
```diff
@@ -231,53 +233,72 @@ static void setup_options(CLI::App &app)
...
+  // Load into memory.
+  elf.load([](uint64_t address, const uint8_t *data, uint64_t length) {
+    // TODO: We could definitely improve on rts.c's memory implementation
+    // (which is O(N^2)) and writing one byte at a time here.
+    for (uint64_t i = 0; i < length; ++i) {
+      write_mem(address + i, data[i]);
+    }
+  });
+
+  fprintf(stdout, "ELF Entry @ 0x%" PRIx64 "\n", elf.entry());
```
pmundkur: This will now get printed out for each file, but I guess that's okay.

Timmmm: Good point; I moved it out of the function and slightly reworded it (since it will print the RVFI entry point which isn't from an ELF).

PR #1178, opened 2025-07-30
Simplify vector read_single_element/write_single_element

Timmmm:
The old code for this was quite convoluted and inefficient.


Timmmm:
Most of the code in each function is identical so we could extract it into a function. It's probably fine for now though.

Timmmm:
Ooo Lean works again apparently...

PR #1179, opened 2025-07-31
Fix `read_vreg` for negative LMUL in widening vector instructions.

pmundkur:
Call the single register read only when all the (potentially wider) elements fit into `vlen` bits.

This fixes a bug with widening instructions with negative LMUL, where a negative LMUL was assumed to imply that a single register read sufficed.  In this situation, using `num_elems` in conjunction with the widened SEW could cause out-of-bounds read-accesses into the register's bits.

### DIFF
--- a/model/riscv_vext_control.sail
+++ b/model/riscv_vext_control.sail
@@ -22,12 +22,15 @@ function get_num_elem(LMUL_pow, SEW) = {
 /* Reads a single vreg into multiple elements */
 val read_single_vreg : forall 'n 'm, 'n >= 0 & is_sew_bitsize('m) . (int('n), int('m), vregidx) -> vector('n, bits('m))
 function read_single_vreg(num_elem, SEW, vrid) = {
+  // TODO: This should be part of the type signature for this function.
+  assert(num_elem * SEW <= vlen);
+
   let bv = V(vrid);
   var result : vector('n, bits('m)) = vector_init(zeros());
 
   foreach (i from 0 to (num_elem - 1)) {
     let start_index = i * SEW;
-    result[i] = slice(bv, start_index, SEW);
+    result[i] = bv[start_index + SEW - 1 .. start_index];
   };
 
   result


nadime15:
I assume thats a fix for #894 ?

pmundkur:
> I assume thats a fix for #894 ?

Yes.

Timmmm:
Ah nice. Probably worth copying the explanation from the other PR too.

Timmmm:
I added a fix for the `slice()` in `read_single_reg` so we can be sure this actually fixed the issue (it should now fail CI if not because it's not relying on a disabled C `assert()` to do the bounds check). Hope that's ok!

model/riscv_vext_control.sail
```diff
@@ -63,7 +66,8 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
     /* vrid must be a multiple of emul */
     assert(false, "invalid register group: vrid is not a multiple of EMUL")
   } else {
-    if LMUL_pow < 0 then {
+    /* Check if a single register read suffices. */
+    if num_elem * SEW < vlen then {
```
nadime15, line 70: A short and elegant fix! I like it! 

I feel like we could even get rid of the whole if/else block by doing:

```
let 'num_elem_single = vlen / SEW;
assert('num_elem_single >= 0);

let nregs = if num_elem * SEW < vlen then 1 else (2 ^ LMUL_pow_reg);

foreach (i_lmul from 0 to (nregs - 1)) {
  let r_start_i : int = i_lmul * 'num_elem_single;
  let r_end_i   : int = r_start_i + 'num_elem_single - 1;
  let vrid_lmul     : vregidx = vregidx_offset(vrid, to_bits_unsafe(5, i_lmul));
  let single_result : vector('num_elem_single, bits('m)) = read_single_vreg('num_elem_single, SEW, vrid_lmul);
  foreach (r_i from r_start_i to r_end_i) {
    let s_i : int = r_i - r_start_i;
    assert(0 <= r_i & r_i < num_elem);
    assert(0 <= s_i & s_i < 'num_elem_single);
    result[r_i] = single_result[s_i];
  }
}
```
I am not sure if this version is clearer or better.

Timmmm: Yeah I have some changes that get rid of a mountain of this code. I don't know why it was written in such a complicated way.

https://github.com/riscv/sail-riscv/commit/0bfe8c4f1f0539e5a3f05544004681a36ee0237f#diff-54589354e5b8327ba1ca21a054b982b948e4bcdeb21a5b5add591d48e1cea44a

nadime15: Ah ok wow, this looks much better!

jordancarlin (approved):

PR #1184, opened 2025-07-31
Add support for Zvfbfmin extension

nadime15:
Add the following instructions to support vector bfloat16 conversions:

```
vfncvtbf16.f.f.w
vfwcvtbf16.f.f.v
```

### DIFF
--- a/model/termination/termination.sail
+++ b/model/termination/termination.sail
@@ -89,6 +89,7 @@ function currentlyEnabled_measure(ext : extension) -> int =
     Ext_Zvl512b => 0,
     Ext_Zvl1024b => 0,
     Ext_Zve32x => 1,
+    Ext_Zvfbfmin => 3,
     Ext_Zve64f => 3,
     Ext_Zve64d => 4,
     Ext_Zvfh => 4,


nibrunieAtSi5 (commented):
Wouldn't it make sense to initially open the review against a different base branch than `master`: the branch https://github.com/nadime15/sail-riscv/tree/add_ext_zfbfmin master (to avoid having to review the Zfbfmin related changes here and in https://github.com/riscv/sail-riscv/pull/1158)

nadime15:
Good point, I could have done it that way as well and it would have been cleaner!

The correct order to merge these extensions is #1158, then #1184, and finally #1185. I assume reviewers focus on the newest commit. Usually, I rebase the follow-up PRs immediately after the preceding PRs are merged to minimize the number of impacted files.

model/riscv_insts_zvfbfmin.sail
```diff
@@ -0,0 +1,110 @@
+//=======================================================================================
+//  This Sail RISC-V architecture model, comprising all files and
+//  directories except where otherwise noted is subject the BSD
+//  two-clause license in the LICENSE file.
+//
+//  SPDX-License-Identifier: BSD-2-Clause
+//=======================================================================================
+
+// TODO: Zvfbfmin depends on Zve32f. Once Zve32f is supported we can simplify the clause below
+function clause currentlyEnabled(Ext_Zvfbfmin) = hartSupports(Ext_Zvfbfmin) & currentlyEnabled(Ext_V) & vlen >= 32
```
Timmmm: Can you have vlen < 32?

Timmmm (commented):
LGTM but I only skimmed it. Does the vector test suite have any tests for this?

model/riscv_insts_zvfbfmin.sail
nadime15: In theory, V fixes `vlen` at 128 (though we dont currently enforce this), but `vlen` can be as low as 8 as long as `vlen ≥ elen`.

Speaking of tests, I added bfloat16 support, including tests for Zvfbfmin and Zvfbfwma, to the vector test suite a while ago (https://github.com/chipsalliance/riscv-vector-tests/pull/77). I verified the implementation, assuming I did not make any mistakes.

Timmmm: Oh yeah V requires a *minimum* vlen of 128, and elen=64. That is interesting!

I think this check should be `elen >= 32` though right?

jordancarlin: From the spec:
> This extension depends upon Zve32f vector extension.

Zve32f in turn requires `vlen >= 32`, `elen >=32`, and support for single precision floating point vector operations. Seems like we should just wait on #1163 and then make it dependent on `currentlyEnabled(Ext_Zve32f)`.

nadime15: Sorry yeah, V requires a minimum vlen of 128 not fixes. 

> Zve32f in turn requires vlen >= 32, elen >=32, and support for single precision floating point vector operations. Seems like we should just wait on https://github.com/riscv/sail-riscv/pull/1163 and then make it dependent on currentlyEnabled(Ext_Zve32f).

Thats fine by me, we can wait until #1163 is merged.

jordancarlin: I don't think it needs to wait. For now it is probably fine to just gate it on the V extension and we can update it in #1163. That is how several of the vector crypto extensions are handled for now as well. I was just saying that is what the long term check should be for this extension.

nadime15: Okay, sounds good, I just added the additional check for elen.

model/extensions/bfloat16/zvfbfmin_insts.sail
```diff
@@ -0,0 +1,110 @@
+//=======================================================================================
+//  This Sail RISC-V architecture model, comprising all files and
+//  directories except where otherwise noted is subject the BSD
+//  two-clause license in the LICENSE file.
+//
+//  SPDX-License-Identifier: BSD-2-Clause
+//=======================================================================================
+
+// TODO: Zvfbfmin depends on Zve32f. Once Zve32f is supported we can simplify the clause below
+function clause currentlyEnabled(Ext_Zvfbfmin) = hartSupports(Ext_Zvfbfmin) & currentlyEnabled(Ext_V) & vlen >= 32 & elen >= 32
```
Timmmm: I think you only really need to check `elen` since it's always true that `vlen >= elen`. And that's why this check is here - you can't convert to/from f32 if you don't even support f32. vlen is irrelevant.

nadime15: Yes, in theory vlen is irrelevant here since vlen ≥ elen and elen ≥ 32, so that’s just a lower bound. But that was intentional to model Zve32f and stay as close to the spec as possible. 

The whole clause is inaccurate anyway and should be replaced once Zve32f is available. And if I wanted to be really precise, I would not even need to check elen or vlen, because strictly speaking the V extension already fixes elen = 64.

Timmmm: Hmm good point; we've used `V` incorrectly. I'll make an issue for that.

pmundkur: Once #1163 gets in, this should probably be
```suggestion
function clause currentlyEnabled(Ext_Zvfbfmin) = hartSupports(Ext_Zvfbfmin) & (currentlyEnabled(Ext_V) | currentlyEnabled(Zve32f)) 
```

nadime15: Done

nadime15:
Yeah I am going to that once this PR and #1185 is merged.

jordancarlin:
Looks like this was merged but the PR wasn’t marked as merged. Not sure what went wrong here.

PR #1186, opened 2025-07-31
Enable C/C++ assertions in all builds.

### LINKED ISSUE #1181: Assertions not enabled in CI
Currently in CI we build the emulator in release mode, and CMake adds `-DNDEBUG` in that case which disables C `assert()`. These are [sometimes important](https://github.com/riscv/sail-riscv/pull/894#issuecomment-3139038238) so we probably want CI to either:

1. Build in debug mode
2. Not disable assertions

I'd say probably the second option is the best because the vector tests take a long time to run and we don't want them to be even slower.

Unfortunately it's not super easy but here's now LLVM does it: https://github.com/llvm/llvm-project/blob/1f7885cf9c6801d11491c8c194c999f7223dd141/llvm/cmake/modules/HandleLLVMOptions.cmake#L118

pmundkur:
This undefines NDEBUG in the top-level CMake file to ensure that assertions in library code are enabled.  Tested with for Release and RelWithDebInfo builds.

Fixes #1181.


pmundkur:
This would need to be merged after #1179 otherwise the vector tests will fail.

jordancarlin (approved):

pmundkur:
How did Github allow this to get merged even though the tests failed during merge?
https://github.com/riscv/sail-riscv/actions/runs/16679543401/job/47214565257

nadime15:
I see this run here:

> Enable C/C++ assertions in all builds. (#1186)
> CI #3846: Commit [abe94fd](https://github.com/riscv/sail-riscv/commit/abe94fd17d6525425892baa52fe6b04a7e6399f9) pushed by [github-merge-queue](https://github.com/apps/github-merge-queue)

https://github.com/riscv/sail-riscv/actions/runs/16679795271

PR #1189, opened 2025-08-01
Implement JSON schema validation for the configuration.

pmundkur:
This uses an externally installed [boon](https://github.com/santhosh-tekuri/boon).

For now, it needs to be explicitly enabled with `--validate-config-schema`. With this flag, _only_ schema conformance is checked; that is, `--validate-config` is still needed to validate other aspects of the configuration.

The schema is embedded into the executable, removing the need
to locate the schema file at run time.

### DIFF
--- a/c_emulator/CMakeLists.txt
+++ b/c_emulator/CMakeLists.txt
@@ -20,6 +20,16 @@ message(STATUS "CXX compiler version: ${cxx_compiler_version}")
 # make version available to emulator
 configure_file(sail_riscv_version.cpp.in sail_riscv_version.cpp @ONLY)
 
+# generate the schema populator script
+set(gen_default_schema_input "${CMAKE_CURRENT_SOURCE_DIR}/default_config_schema.h.in")
+configure_file(generate_schema_include.cmake.in generate_schema_include.cmake @ONLY)
+add_custom_command(
+    OUTPUT default_config_schema.h
+    COMMAND cmake -P generate_schema_include.cmake
+    DEPENDS ${schema_file}
+)
+add_custom_target(generated_default_config_schema DEPENDS default_config_schema.h)
+
 ## A library that contains the C model and support code.
 
 add_library(riscv_model
@@ -38,12 +48,13 @@ add_library(riscv_model
     riscv_sail.h
     riscv_softfloat.c
     riscv_softfloat.h
+    "${CMAKE_CURRENT_BINARY_DIR}/default_config_schema.h"
     config_utils.cpp
     config_utils.h
     symbol_table.cpp
     symbol_table.h
-    sail_riscv_version.h
     "${CMAKE_CURRENT_BINARY_DIR}/sail_riscv_version.cpp"
+    sail_riscv_version.h
     "${CMAKE_BINARY_DIR}/sail_riscv_model.c"
     "${CMAKE_BINARY_DIR}/sail_riscv_model.h"
 )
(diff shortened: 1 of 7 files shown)


pmundkur:
When the config is not specified, the default config is written to a temp file, giving a slightly unsatisfactory output from `boon`:
```
$ ./build/c_emulator/sail_riscv_sim --validate-config-schema
schema /home/mundkur/src/hw/sail-ol/sail-riscv/build/sail_riscv_config_schema.json: ok

instance /tmp/default_config_fFneVS: ok
```

Otherwise it gives:
```
$ ./build/c_emulator/sail_riscv_sim --validate-config-schema --config config/rv64d.json 
schema /home/mundkur/src/hw/sail-ol/sail-riscv/build/sail_riscv_config_schema.json: ok

instance config/rv64d.json: ok
```

Auto-locating the schema file has not yet been implemented for non-Linux platforms.

pmundkur:
If `boon` is not installed:
```
$ ./build/c_emulator/sail_riscv_sim --validate-config-schema --config config/rv64d.json 
Error validating schema file with 'boon'. Please install boon with 'cargo install boon-cli' and ensure it is in PATH.
```

Timmmm:
> When the config is not specified, the default config is written to a temp file

Is there much point validating the built-in config? The user can't change it and hopefully we haven't screwed it up! We could add CTest test to check all of the configs we generate to verify that.

pmundkur:
> > When the config is not specified, the default config is written to a temp file
> 
> Is there much point validating the built-in config? The user can't change it and hopefully we haven't screwed it up! We could add CTest test to check all of the configs we generate to verify that.

There's not much point, but an outcome of how the cli should work: without an argument, the default built-in config is chosen.  It's something a new user would try, and it's already how the `--validate-config` option works.

Timmmm:
I think I would just have it print a message like

> Please provide a config with the --config argument. To see the default config use --print-default-config.

pmundkur:
> I think I would just have it print a message like
> 
> > Please provide a config with the --config argument. To see the default config use --print-default-config.

I think that won't be very helpful for the user, since `boon` requires its input in a file.  The user would need to save the output of `--print-default-config` to a file, and then use that file with `--config`.

It would also be different from how `--validate-config` already works.

Timmmm:
> The user would need to save the output of --print-default-config to a file, and then use that file with --config.

Sure, but that's probably what they want to do anyway right. I don't see much point in just saying "yep, the built in config is valid"? If you want you could just hard-code that message I guess, but that seems less useful information than explaining how to validate a config that might actually be invalid.

pmundkur:
Perhaps we could add to that to the footer of the help message, something like
```
To validate the schema of a configuration, invoke with
  --config <cfg> --validate-config-schema
If no config file is specified, the default configuration is validated
```

pmundkur:
If the package is to bundle `boon`, I'm wondering where it should be installed?  The current package directories under the install prefix are
```
bin
share/doc/sail-riscv
share/sail-riscv/
share/sail-riscv/config
```
Should it go under `lib` as
```
lib/sail-riscv/boon
```
?

Timmmm:
> If no config file is specified, the default configuration is validated

Why though? It should never be invalid so I don't know why you'd ever want to do that.

> If the package is to bundle boon, I'm wondering where it should be installed?

Good question. I think in `bin` so that they don't have to do extra steps to add it to `PATH`. That's what we did with `z3` in Sail. There's an argument it might conflict with an existing `boon` install but I think that's pretty unlikely really, and not too difficult to fix if it happens.

pmundkur:
Closing this since going with a library option instead of an external binary seems a nicer option.

PR #1191, opened 2025-08-05
add option to control tval update by ebreak

chihminchao:
1. add 'excp_breakpoint_updates_tval' option
2. apply the option to exception handling

reference
  based on privileged spec 3.1.16
  " The  hardware  platform  will  specify  which  exceptions  must  set mtval
  informatively, which may unconditionally set it to zero"

  "If  mtval  is written with a nonzero value when a breakpoint,
  address-misaligned, access-fault, or page-fault
  exception occurs on an instruction fetch, load, or store, then  mtval
  will contain the faulting virtual address."

  It is implementation-defined that ebreak could update tval with zero
  of faulting virtual address.

### DIFF
--- a/config/rv32d.json
+++ b/config/rv32d.json
@@ -7,7 +7,8 @@
       "len": 32,
       "value": "0xFFFF_FFFF"
     },
-    "mtval_has_illegal_instruction_bits": false
+    "mtval_has_illegal_instruction_bits": false,
+    "excp_breakpoint_updates_tval": true
   },
   "memory": {
     "pmp": {
--- a/model/riscv_sys_control.sail
+++ b/model/riscv_sys_control.sail
@@ -8,6 +8,9 @@
 
 /* Machine-mode and supervisor-mode functionality. */
 
+/* whether breakpoint exception stores pc or zero into tval */
+let plat_excp_breakpoint_updates_tval : bool = config base.excp_breakpoint_updates_tval
+
 function effectivePrivilege(t : AccessType(ext_access_type), m : Mstatus, priv : Privilege) -> Privilege =
   if   t != InstructionFetch() & m[MPRV] == 0b1
   then privLevel_of_bits(m[MPP])
@@ -168,7 +171,7 @@ function trap_handler(del_priv : Privilege, intr : bool, c : exc_code, pc : xlen
        mstatus[MPIE] = mstatus[MIE];
        mstatus[MIE]  = 0b0;
        mstatus[MPP]  = privLevel_to_bits(cur_privilege);
-       mtval         = tval(info);
+       mtval         = if not(plat_excp_breakpoint_updates_tval) & c == 0x03 then zeros() else tval(info);
        mepc          = pc;
 
        cur_privilege = del_priv;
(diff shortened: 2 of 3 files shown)


config/rv32d.json
```diff
@@ -7,7 +7,8 @@
       "len": 32,
       "value": "0xFFFF_FFFF"
     },
-    "mtval_has_illegal_instruction_bits": false
+    "mtval_has_illegal_instruction_bits": false,
+    "excp_breakpoint_updates_tval": true
```
pmundkur, line 11: According to the spec, `tval` would need a flag per relevant exception.  Something like
```
"base": {
...
  "mtval_exception_setters": {
    "illegal_instruction": false,
    "breakpoint": true,
    ...
  }
}
```
This doesn't cover the "either behavior" case, or the "depending on the event that caused the exception".  We could add those later if needed.  Thoughts?

model/riscv_sys_control.sail
```diff
@@ -168,7 +171,7 @@ function trap_handler(del_priv : Privilege, intr : bool, c : exc_code, pc : xlen
        mstatus[MPIE] = mstatus[MIE];
        mstatus[MIE]  = 0b0;
        mstatus[MPP]  = privLevel_to_bits(cur_privilege);
-       mtval         = tval(info);
+       mtval         = if not(plat_excp_breakpoint_updates_tval) & c == 0x03 then zeros() else tval(info);
```
nadime15, line 174: Does it make sense to get rid of the magic number and to do the following?

```suggestion
       mtval         = if not(plat_excp_breakpoint_updates_tval) & c == exceptionType_to_bits(E_Breakpoint())) then zeros() else tval(info);
```

arichardson: Or maybe push it into the caller?

config/rv32d.json
jordancarlin: UDB has a fairly lengthy list of parameters for what gets written to `mtval`: https://github.com/riscv-software-src/riscv-unified-db/blob/4d88f2942cdf6f9019ec51538804a2790f506db0/spec/std/isa/ext/Sm.yaml#L321-L400. There are comparable lists for the other *tval registers as well.

model/riscv_sys_control.sail
chihminchao: The 'c' variable passed into this function is declared an int type.

nadime15: Yeah lets do what @arichardson has said

```suggestion
       mtval         = tval(info);
```

and in 

 
```
exception_handler (..)
   ...
   let trap_value = if e.trap == E_Breakpoint() then Some(zeros()) else e.excinfo;
   trap_handler(del_priv, false, exceptionType_to_bits(e.trap), pc, trap_value, e.ext)
   ...
``` 

(a function named tval already exists, so we can use tval as a variable name again)

You could even go a step further and define a function that returns `trap_value`, since there are multiple possible values for it.

jordancarlin:
Superseded by #1225.

chihminchao:
It is good to have better and elegant solution.

PR #1197, opened 2025-08-11
Remove duplicate config_print checks

### LINKED ISSUE #1174: config_print_instr is checked twice
This flow is weird...

```
        if   get_config_print_instr()
        then {
          print_instr("[" ^ dec_str(step_no) ^ "] [" ^ to_str(cur_privilege) ^ "]: " ^ BitStr(PC) ^ " (" ^ BitStr(w) ^ ") " ^ to_str(instruction));
        };

...

unit print_instr(sail_string s)
{
  if (config_print_instr)
    fprintf(trace_log, "%s\n", s);
  return UNIT;
}
```

We check `config_print_instr` twice. I would say since we are checking it in Sail we don't need to check it again in C. And it makes sense to check it in Sail because we avoid the overhead of string formatting if it's disabled. So we should probably remove the C checks.

Probably the same issue for other `config_print..`s but I haven't checked.

trdthg:
The config flags `config_print_instr` and `config_print_platform` are checked twice in the Sail and C code. This commit removes the C functions `print_instr()` and `print_platform()`, and replaces them with a single function `print_trace()`, which doesn't do a redundant check.

Fixes #1174

### DIFF
--- a/model/riscv_platform.sail
+++ b/model/riscv_platform.sail
@@ -90,11 +90,13 @@ function within_phys_mem forall 'n, 'n <= max_mem_access. (Physaddr(addr) : phys
            & (addr_int + sizeof('n)) <= (rom_base_int + rom_size_int))
   then    true
   else {
-    print_platform("within_phys_mem: " ^ bits_str(addr) ^ " not within phys-mem:");
-    print_platform("  plat_rom_base: " ^ bits_str(plat_rom_base));
-    print_platform("  plat_rom_size: " ^ bits_str(plat_rom_size));
-    print_platform("  plat_ram_base: " ^ bits_str(plat_ram_base));
-    print_platform("  plat_ram_size: " ^ bits_str(plat_ram_size));
+    if get_config_print_platform() then {
+      print_log("within_phys_mem: " ^ bits_str(addr) ^ " not within phys-mem:");
+      print_log("  plat_rom_base: " ^ bits_str(plat_rom_base));
+      print_log("  plat_rom_size: " ^ bits_str(plat_rom_size));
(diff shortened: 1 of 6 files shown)


Timmmm (approved):
Much better!

model/riscv_platform.sail
```diff
@@ -90,11 +90,11 @@ function within_phys_mem forall 'n, 'n <= max_mem_access. (Physaddr(addr) : phys
            & (addr_int + sizeof('n)) <= (rom_base_int + rom_size_int))
   then    true
   else {
-    print_platform("within_phys_mem: " ^ bits_str(addr) ^ " not within phys-mem:");
-    print_platform("  plat_rom_base: " ^ bits_str(plat_rom_base));
-    print_platform("  plat_rom_size: " ^ bits_str(plat_rom_size));
-    print_platform("  plat_ram_base: " ^ bits_str(plat_ram_base));
-    print_platform("  plat_ram_size: " ^ bits_str(plat_ram_size));
+    trace("within_phys_mem: " ^ bits_str(addr) ^ " not within phys-mem:");
```
arichardson: We no longer check config_print_platform here?

arichardson (approved):
I believe one check is missing otherwise LGTM.

trdthg:
```
sail-riscv/c_emulator/riscv_prelude.h:12:6: warning: declaration of ‘unit log(sail_string)’ conflicts with built-in declaration ‘double log(double)’ [-Wbuiltin-declaration-mismatch]
   12 | unit log(sail_string s);
      |      ^~~
```

log conflicts with log in math.h, so change to print_log

PR #1200, opened 2025-08-12
Replace slice() with [] syntax

### LINKED ISSUE #1182: Remove all uses of `slice()`
The vector code has a lot of calls to `slice()`. This is a built in Sail function to do bit vector slicing. Normally you would use `foo[7 .. 0]` but the vector code does a lot of `slice(foo, 0, 8)` instead. I presume this is because Sail doesn't check for out-of-bounds slicing with `slice()` but it does with `[]`, and the vector code is really badly typed so it was probably a lazy way to ignore type checking errors.

We should remove all of the `slice()`s, add `assert()`s where necessary, and then eventually try to improve the typing so the `assert()`s aren't necessary.

trdthg:
Resolve #1182


model/riscv_insts_vext_utils.sail
```diff
@@ -179,8 +179,11 @@ function get_velem_quad(v, i) = v[4 * i + 3] @ v[4 * i + 2] @ v[4 * i + 1] @ v[4
 /* Divide the input bitvector into 4 equal slices and store them in vd starting at position 4*i */
 val write_velem_quad : forall 'n, 'n > 0 . (vregidx, sew_bitsize, bits('n), nat) -> unit
 function write_velem_quad(vd, SEW, input, i) = {
-  foreach(j from 0 to 3)
-    write_single_element(SEW, 4 * i + j, vd, slice(input, j * SEW, SEW));
+  foreach(j from 0 to 3) {
+    // write_single_element(SEW, 4 * i + j, vd, slice(input, j * SEW, SEW));
```
Timmmm: Probably remove this comment, and maybe add a `// TODO: Change the type signature to enforce this assertion.`

model/riscv_insts_vext_utils.sail
```diff
@@ -448,13 +451,17 @@ function get_fixed_rounding_incr(vec_elem, shift_amount) = {
   if shift_amount == 0 then 0b0
   else {
     let rounding_mode = vcsr[vxrm];
+    assert(shift_amount <= 'm);
```
Timmmm: Definitely add a `// TODO: Change the type signature to enforce this assertion.`

We shouldn't try to do it here though - it will likely make the PR much bigger.

model/riscv_insts_vext_utils.sail
```diff
@@ -448,13 +451,18 @@ function get_fixed_rounding_incr(vec_elem, shift_amount) = {
   if shift_amount == 0 then 0b0
   else {
     let rounding_mode = vcsr[vxrm];
+    // TODO: Change the type signature to enforce this assertion.
+    assert(shift_amount <= 'm);
```
pmundkur:
```suggestion
    assert(shift_amount < 'm);
```
You need this to access `vec_elem[shift_amount]`.

pmundkur (changes requested):

pmundkur:
Zero-length slices occurred in some tests, for which the indexing tripped asserts.  I pushed a fix, hope that's okay.

model/riscv_insts_vext_utils.sail
```diff
@@ -448,13 +451,18 @@ function get_fixed_rounding_incr(vec_elem, shift_amount) = {
...
+    assert(shift_amount < 'm);
     match rounding_mode {
-      0b00 => slice(vec_elem, shift_amount - 1, 1),
-      0b01 => bool_to_bits(
-        (slice(vec_elem, shift_amount - 1, 1) == 0b1) & (slice(vec_elem, 0, shift_amount - 1) != zeros() | slice(vec_elem, shift_amount, 1) == 0b1)),
+      0b00 => bit_to_bits(vec_elem[shift_amount - 1]),
+      0b01 => {
+        bool_to_bits(
+          vec_elem[shift_amount - 1] == bitone &
+          ((if shift_amount == 1 then false else vec_elem[shift_amount - 2 .. 0] != zeros()) | vec_elem[shift_amount] == bitone))
```
Timmmm, line 461:
```suggestion
          ((shift_amount > 1 & vec_elem[shift_amount - 2 .. 0] != zeros()) | vec_elem[shift_amount] == bitone))
```

model/riscv_insts_vext_utils.sail
```diff
@@ -448,13 +451,18 @@ function get_fixed_rounding_incr(vec_elem, shift_amount) = {
...
+      },
       0b10 => 0b0,
       0b11 => bool_to_bits(
-        not(slice(vec_elem, shift_amount, 1) == 0b1) & (slice(vec_elem, 0, shift_amount) != zeros()))
+        not(vec_elem[shift_amount] == bitone) & (vec_elem[shift_amount - 1 .. 0] != zeros()))
```
Timmmm:
```suggestion
        vec_elem[shift_amount] != bitone & vec_elem[shift_amount - 1 .. 0] != zeros()
      )
```

model/riscv_insts_vext_utils.sail
pmundkur: The `if` allowed the removal of the `assert(shift_amount - 1 >= 0);` due to how the type constraints handle `if`s.  Doing it the suggested way will re-introduce that assert.  I lean towards keeping the `if` and removing the assert.

Timmmm: I'm pretty sure I tried it and it works. The type checker is smart enough to handle boolean and/or as flow control.

pmundkur: Without the assert, one gets:
```
Type error:
./riscv_insts_vext_utils.sail:460.10-461.110:
460 |          vec_elem[shift_amount - 1] == bitone &
    |          ^-------------------------------------
461 |          ((shift_amount > 1 & vec_elem[shift_amount - 2 .. 0] != zeros()) | vec_elem[shift_amount] == bitone))
    |-------------------------------------------------------------------------------------------------------------^
    | No overloading for (operator !=), tried:
    | * neq_int
    |   ./riscv_insts_vext_utils.sail:461.31-62:
    |   461 |          ((shift_amount > 1 & vec_elem[shift_amount - 2 .. 0] != zeros()) | vec_elem[shift_amount] == bitone))
    |       |                               ^-----------------------------^ checking function argument has type int('fv7579#n)
    |       | Could not resolve quantifiers for subrange_bits
    |       | * (0 <= 0 & (0 <= ('n - 2) & ('n - 2) < 'm))
    | * neq_bool
    |   ./riscv_insts_vext_utils.sail:461.31-62:
    |   461 |          ((shift_amount > 1 & vec_elem[shift_amount - 2 .. 0] != zeros()) | vec_elem[shift_amount] == bitone))
    |       |                               ^-----------------------------^ checking function argument has type bool('ex5#)
    |       | Could not resolve quantifiers for subrange_bits
    |       | * (0 <= 0 & (0 <= ('n - 2) & ('n - 2) < 'm))
    | * neq_bits
    |   ./riscv_insts_vext_utils.sail:461.31-62:
    |   461 |          ((shift_amount > 1 & vec_elem[shift_amount - 2 .. 0] != zeros()) | vec_elem[shift_amount] == bitone))
    |       |                               ^-----------------------------^ checking function argument has type bitvector('fv7624#n)
    |       | Could not resolve quantifiers for subrange_bits
    |       | * (0 <= 0 & (0 <= ('n - 2) & ('n - 2) < 'm))
    | * neq_anything
    |   ./riscv_insts_vext_utils.sail:461.31-62:
    |   461 |          ((shift_amount > 1 & vec_elem[shift_amount - 2 .. 0] != zeros()) | vec_elem[shift_amount] == bitone))
    |       |                               ^-----------------------------^ checking function argument has type 'a
    |       | Could not resolve quantifiers for subrange_bits
    |       | * (0 <= 0 & (0 <= ('n - 2) & ('n - 2) < 'm))

```

Timmmm: Oh yeah turns out I forgot to save the edit before compiling. :-D

@alasdair - not very critical but this feels like something that should work.

PR #1201, opened 2025-08-12
Update the ReadingGuide in the documentation.

pmundkur:
This also moves it to the AsciiDoc format.


pmundkur:
I can remove the .md file after review.

I'm not sure how useful the automatically generated depgraph is.  The project file is simpler in comparison.  Perhaps it can be removed?

Is there an asciidoc link-checker that has good support for local links?

Timmmm:
Not as far as I know. I think Asciidoctor deliberately doesn't check links because technically they might be valid later even if they aren't at compilation time. You basically have to use an HTML link checker on the output.

doc/figs/sail_riscv_project.dot.svg
Timmmm, line 1: I wonder if this would be more comprehensible with the orthogonal line option?

pmundkur: It starts looking more like a chip layout with "splines=ortho" :-)   I'm inclined to just not have a diagram, since the project file seems easier to understand than this.
[sail_riscv_project_edited.dot.pdf](https://github.com/user-attachments/files/21742476/sail_riscv_project_edited.dot.pdf)

pmundkur:
> I dunno about using Asciidoc. Asciidoctor is kind of a right pain to deal with because it's based around Ruby (the `riscv-isa-manual` guys use Docker because of this). I think it would make sense if we're going to use `asciidoctor-sail` - are you planning that? Otherwise I can't see any benefit over Markdown?

We already have  `doc/notes_Virtual_Memory.adoc`, so I thought it would make sense to switch these docs too for consistency.

Our README-style docs don't need any tooling, so we shouldn't need Docker etc.  I hadn't thought of using `asciidoctor-sail`, but switching now would enable its use in the future if we wanted to.

doc/figs/sail_riscv_project.dot.svg
Timmmm: Ha yeah that didn't really help. I agree, if anything the diagram makes it look worse than it is.

pmundkur:
It looks like Markdown has a decent linkchecker that supports local links
https://github.com/tcort/markdown-link-check
that can be integrated into pre-commit checks and CI.  I tried it and it seems to work.

This will be important to prevent broken links in the docs as files get moved around and renamed (see https://github.com/riscv/sail-riscv/pull/1210#discussion_r2280662875 and https://github.com/riscv/sail-riscv/pull/1210#issuecomment-3194392338).

Unless there are objections, I'll redo this PR in Markdown.

jrtc27:
asciidoctor itself is fine, the problem is all the plugins that RISC-V docs pull in. In particular, wavedrom uses node.js. I've kept riscv-elf-psabi-doc only using `asciidoctor asciidoctor-pdf` so you just need to `gem install` those, and no Docker image shenanigans is needed.

pmundkur:
@jrtc27 is there an invocation you use to check for broken links?  I was able to break some links in the riscv-elf-psabi-doc repo and still build the pdf and html.  My asciidoctor versions are pretty old though: 
```
$ asciidoctor -V
Asciidoctor 2.0.16 [https://asciidoctor.org]
Runtime Environment (ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux-gnu]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)
$ asciidoctor-pdf -V
Asciidoctor PDF 1.6.2 using Asciidoctor 2.0.16 [https://asciidoctor.org]
Runtime Environment (ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux-gnu]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)

```

jrtc27:
No, there isn't anything there to check that. Presumably tools exist, but I don't know what.

pmundkur:
Well, that was my motivation for moving (back) to Markdown, since the updated ReadingGuide in this PR now has links to the source files that would break without a checker.

pmundkur:
It looks like link checking external URLs is a bit flaky.  Perhaps it should be turned off?  The point was to mainly check internal doc links anyway.

nadime15 (approved):
Looks good!

Timmmm:
Oh is that because GMP ban github from accessing it? I think we should just whitelist that URL.

doc/figs/riscvcsimdeps.tex
```diff
@@ -39,24 +56,31 @@
                                                          \texttt{riscv\_platform\_impl}};
 
   \coordinate (clib) at ($(cmodel)!1.5!(cmodel |- platimpl)$);
-
   \node (csail)      [csail]  at (clib)              {\textbf{C Sail library}\\
                                                          \texttt{sail,rts,elf}\\
                                                          Sail values\\
                                                          raw memory\\
                                                          ELF loading};
```
jordancarlin: ELF loading is no longer done by the Sail library.

doc/ReadingGuide.md
```diff
@@ -1,155 +1,278 @@
...
+various Sail backends.
+[prelude_mem_addrtype.sail](../model/prelude_mem_addrtype.sail) and
+[prelude_mem_metadata.sail](../model/prelude_mem_metadata.sail)
+contain other low-level definitions related to memory.
+
+[riscv_extensions.sail](../model/riscv_extensions.sail) sets up the
+basic infrastructure for the definition of modules implementing RISC-V
+extensions. The `hartSupports` function determines whether an
+extension is supported by the model configuration, while the
+`currentlyEnabled` determines whether the extension is usable given
```
jordancarlin:
```suggestion
`currentlyEnabled` function determines whether the extension is usable given
```

doc/ReadingGuide.md
```diff
@@ -1,155 +1,278 @@
...
+
+### Extensions
+
+The `extensions` module contains a sequence of submodules, each
+typically implementing an ISA extension. In some cases, submodules
+implementing related extensions (e.g. `Zaamo` and `Zalrsc`) may be
+grouped together and nested within another submodule (e.g. `A`) under
+the `extensions` module. This nested structure helps to organize the
+files implementing large related extensions such as those in the
+Vector (`V`) and cryptography (`K`) extensions.
```
jordancarlin: There is no `K` extension.
```suggestion
Vector (`V`) and cryptography (`Zk*`) extensions.
```

README.md
```diff
@@ -6,7 +6,7 @@ This repository contains a formal specification of the RISC-V architecture, writ
 The model specifies assembly language formats of the instructions, the corresponding
 encoders and decoders, and the instruction semantics. A [reading guide](doc/ReadingGuide.md)
 to the model is provided in the [doc/](doc/) subdirectory, along with a guide on [how to
-extend](doc/ExtendingGuide.md) the model.
+add a new extension](doc/AddingExtensions.md) to the model.
```
jordancarlin, line 9: What is the plan for `ExtendingGuide.md`? Do you seen `AddingExtensions.md` replacing it? Or will they live side by side (in which case `ExtendingGuide.md` still needs to be updated in a later PR).

doc/figs/riscvcsimdeps.tex
```diff
@@ -39,24 +56,30 @@
                                                          \texttt{riscv\_platform\_impl}};
 
   \coordinate (clib) at ($(cmodel)!1.5!(cmodel |- platimpl)$);
-
   \node (csail)      [csail]  at (clib)              {\textbf{C Sail library}\\
                                                          \texttt{sail,rts,elf}\\
```
jordancarlin:
```suggestion
                                                         \texttt{sail,rts}\\
```

README.md
pmundkur: As discussed in today's meeting, I propose we delete the `ExtendingGuide.md` in a subsequent PR.  That PR could also address updating `notes_Virtual_Memory.adoc` and converting it to Markdown as well, so the link-checker could check the file references in it.

jordancarlin: Discussed during the meeting. `ExtendingGuide.md` will be removed in a future PR.

jordancarlin (approved):

PR #1203, opened 2025-08-12
Simplify riscv_insts_vext_mem's function signatures

### LINKED ISSUE #1194: Simplify type signatures in `riscv_insts_vext_mem.sail`
All of the functions in `riscv_insts_vext_mem.sail` have overly-complicated generic type signatures like this:

```
val process_vlxseg : forall 'f 'ib 'db 'ip 'dp 'n, nfields_range('f) & is_mem_width('ib) & is_mem_width('db) & ('n > 0). (int('f), bits(1), vregidx, int('ib), int('db), int('ip), int('dp), regidx, vregidx, int('n), int) -> ExecutionResult
```

You usually only need generic parameters like that when the parameters or return value are linked in some way. That isn't the case here so this could be much simpler. Something like:

```
val process_vlxseg : (nfields, bits(1), vregidx, word_width, word_width, int, int, regidx, vregidx, nat1, int) -> ExecutionResult
```

(Ideally with less `int` but that can come later.)

KotorinMinami:
Since none of the argument or return types are linked they can all be expressed without `forall`, which is a lot simpler.

Fixes #1194

### DIFF
--- a/model/riscv_insts_vext_mem.sail
+++ b/model/riscv_insts_vext_mem.sail
@@ -46,13 +46,14 @@ mapping clause encdec = VLSEGTYPE(nf, vm, rs1, width, vd)
   <-> encdec_nfields(nf) @ 0b0 @ 0b00 @ vm @ 0b00000 @ encdec_reg(rs1) @ encdec_vlewidth(width) @ encdec_vreg(vd) @ 0b0000111
   when currentlyEnabled(Ext_V)
 
-val process_vlseg : forall 'f 'b 'n 'p, nfields_range('f) & is_mem_width('b) & ('n > 0). (int('f), bits(1), vregidx, int('b), regidx, int('p), int('n)) -> ExecutionResult
+val process_vlseg : (nfields, bits(1), vregidx, word_width, regidx, int, nat1) -> ExecutionResult
 function process_vlseg (nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem) = {
-  let EMUL_reg : int = if EMUL_pow <= 0 then 1 else 2 ^ EMUL_pow;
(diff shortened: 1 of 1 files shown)


model/riscv_insts_vext_mem.sail
```diff
@@ -240,14 +243,15 @@ mapping clause encdec = VLSSEGTYPE(nf, vm, rs2, rs1, width, vd)
...
+val process_vlsseg : (nfields, bits(1), vregidx, word_width, regidx, regidx, int, nat1) -> ExecutionResult
 function process_vlsseg (nf, vm, vd, load_width_bytes, rs1, rs2, EMUL_pow, num_elem) = {
-  let EMUL_reg : int = if EMUL_pow <= 0 then 1 else 2 ^ EMUL_pow;
-  let vm_val  : bits('n) = read_vmask(num_elem, vm, zvreg);
-  let vd_seg  : vector('n, bits('f * 'b * 8)) = read_vreg_seg(num_elem, load_width_bytes * 8, EMUL_pow, nf, vd);
-  let rs2_val : int = unsigned(get_scalar(rs2, xlen));
+  let EMUL_reg = if EMUL_pow <= 0 then 1 else 2 ^ EMUL_pow;
+  let vm_val  = read_vmask(num_elem, vm, zvreg);
+  let vd_seg  = read_vreg_seg(num_elem, load_width_bytes * 8, EMUL_pow, nf, vd);
+  let rs2_val = signed(get_scalar(rs2, xlen));
```
Timmmm: I think this was probably unintentional?

KotorinMinami: Oh, yes! Sorry!

Issue #1205, opened 2025-08-13
Revisiting the `addr_checks` API


pmundkur:
The functions to check control addresses in `riscv_addr_checks.sail` were originally introduced to support the CHERI Sail model.  They have the following signatures:
```
ext_fetch_check_pc(start_pc : xlenbits, pc : xlenbits) -> Ext_FetchAddr_Check(ext_fetch_addr_error)
ext_control_check_pc(pc : xlenbits) -> Ext_ControlAddr_Check(ext_control_addr_error)
```
where
```
union Ext_FetchAddr_Check ('a : Type) = {
  Ext_FetchAddr_OK  : virtaddr,  /* PC value to use for the actual fetch */
  Ext_FetchAddr_Error : 'a
}
union Ext_ControlAddr_Check ('a : Type) = {
  Ext_ControlAddr_OK : virtaddr, /* PC value to use for the target of the control operation */
  Ext_ControlAddr_Error : 'a
}
```

The API implies that the address in the `_OK()` case could differ from the value that was passed in, so callers need to unwrap the address from `_OK()` before proceeding (e.g. in `jump_to`):
```
match ext_control_check_pc(target) {
    Ext_ControlAddr_Error(e) => Ext_ControlAddr_Check_Failure(e),
    Ext_ControlAddr_OK(target) => {
      // Perform standard alignment check.
      let target_bits = bits_of(target);
      ...
```

This makes it difficult to sequence exception priorities, for e.g. in the debug extension:
https://github.com/riscv/sail-riscv/pull/954#discussion_r2192104135

However, in actual practice, even in the CHERI model, the returned address is essentially the same as the `pc` argument:
https://github.com/CHERI-Alliance/sail-cheri-riscv/blob/96c96126f48fd7e73bb3aedffd35a70547dfaf49/src/cheri_addr_checks.sail#L47
https://github.com/CHERI-Alliance/sail-cheri-riscv/blob/96c96126f48fd7e73bb3aedffd35a70547dfaf49/src/cheri_addr_checks.sail#L100

Can we change this API to remove the address from the return type, for e.g. to return an `option(ext_control_addr_error)`?
```
ext_fetch_check_pc(start_pc : xlenbits, pc : xlenbits) -> option(ext_fetch_addr_error)
ext_control_check_pc(pc : xlenbits) -> option(ext_control_addr_error)
```
This removes the need to unwrap the (unchanged) address in the `_OK()` case before proceeding.

[...]

PR #1206, opened 2025-08-13
Remove unused  `ext_control_check_addr` hook function.

pmundkur:
This accidentally became unused when `jump_to` was added, however there is currently no difference between `ext_control_check_addr` and `ext_control_check_pc` in the latest implemented CHERI spec (0.9.3), and in the latest draft spec this check has been removed entirely.

### DIFF
--- a/model/riscv_addr_checks.sail
+++ b/model/riscv_addr_checks.sail
@@ -36,10 +36,6 @@ type ext_control_addr_error = unit
  * value, which will have bit[0] cleared by the caller if needed.
  */
 
-/* the control address is derived from a non-PC register, e.g. in JALR */
-function ext_control_check_addr(pc : xlenbits) -> Ext_ControlAddr_Check(ext_control_addr_error) =
-  Ext_ControlAddr_OK(Virtaddr(pc))
-
 /* the control address is derived from the PC register, e.g. in JAL */
 function ext_control_check_pc(pc : xlenbits) -> Ext_ControlAddr_Check(ext_control_addr_error) =
   Ext_ControlAddr_OK(Virtaddr(pc))


jrtc27:
The CTSRD-CHERI model still uses this to enforce bounds checks on branches prior to the jump. RVY doesn't currently do this, but that was a recent, contentions change that may still be reintroduced by a future extension; you will note that the CHERI-Alliance model still has checks there.

jrtc27:
Oh, this is because JAL and JALR were unified and now both use ext_control_check_pc? That in and of itself was a bit of a dodgy change because they're not the same thing (pre-ISAv9 they differed in whether they were the offset (_pc, since that's what PC used to be) or address (_addr, and what PC is now) to use for the new PCC).

jrtc27:
So yes, for ISAv9 this is ok. Though I do see now that the CTSRD-CHERI model still has an option to use the old offsetting interpretation for PCC and DDC (c.f. #1205), and that's even the default still.

pmundkur:
Hmm, perhaps we should defer this removal then?  That won't be a problem as long as we fix the API as per #1205.

Timmmm:
> Oh, this is because JAL and JALR were unified and now both use ext_control_check_pc? That in and of itself was a bit of a dodgy change because they're not the same thing

Oops that was an oversight, but yeah they are [basically identical in the current CHERI](https://github.com/CHERI-Alliance/sail-riscv/blob/4f1dae4d0d38348a83dd8f3d670264330c767a22/model/cheri/cheri_addr_checks.sail#L89-L106). The only difference is `ext_control_check_addr()` zeros the lowest bit, but that has already been done in `jalr` itself so it shouldn't make any difference.

Also it seems like this check has been entirely removed in the latest CHERI spec anyway:

> A future extension may raise an exception on the JALR instruction itself if the target [pcc](https://riscv.github.io/riscv-cheri/#pcc) will raise a CHERI exception at the target.

PR #1207, opened 2025-08-13
Remove unused `ext_check_phys_mem_{read,write}` hooks.

pmundkur:
These seem to be unused in both the CTSRD and Alliance CHERI models.


Timmmm (approved):
Makes sense. We also never used these. I think anything you'd want to do with these is probably better off as a PMA check (when I eventually get #866 working).

PR #1208, opened 2025-08-15
Add the Zihintntl and Zihintpause extensions.

pmundkur:
These extensions are just hints and the Sail implementation does not do anything special when they are executed. However enabling the extensions means that a) the hint's assembly will be printed in logs, and b) the extensions will appear in the `-march` ISA string.


doc/ChangeLog.md
```diff
@@ -5,6 +5,9 @@
   and `config/rv64d.json` respectively. Dynamic switching between
   the 'E' and 'I' base ISAs is not supported.
 
+- The following extensions have been added:
+  Zihintntl, Zihintpause
```
jordancarlin: It will be easier to maintain this as a bulleted list so we don't get as many merge conflicts.

model/riscv.sail_project
```diff
@@ -300,6 +300,29 @@ extensions {
     requires riscv
     files riscv_insts_zifencei.sail
   }
+
+  Zihintntl {
```
jordancarlin, line 304: In the past, we had discussed making this a separate top-level key from `extensions` (called `hints`) and putting `before extensions` for the entire `hints` module. Seems like that might still be a simpler approach. Anyone else have thoughts on this?

model/riscv.sail_project
```diff
@@ -300,6 +300,29 @@ extensions {
...
+    Zihintntl_insts_compressed {
+      requires riscv, Zihintntl_types, Zihintntl_insts
+      before Zca // C.NTL instructions override C.ADD
+      files riscv_insts_zihintntlc.sail
+    }
+  }
+
+  Zihintpause {
+    requires riscv
+    before Zifenci // PAUSE overrides a FENCE
```
jordancarlin: `FENCE` instructions are part of the base (I) instruction set. Only `fence.i` is part of Zifencei.

model/riscv_insts_zihintntlc.sail
jordancarlin, line 1: Since the compressed Zihintntl instructions are not technically a separate extension, I would just keep them in `riscv_insts_zihintntl.sail` instead of creating a new file. This is similar to the Zcb instructions that rely on other extensions (like M or Zbb).

config/rv64d.json
```diff
@@ -97,6 +97,12 @@
     "Zifencei": {
       "supported": true
     },
+    "Zihintntl": {
+      "supported": true
+    },
+    "Zihintpause": {
+      "supported": true
```
Timmmm, line 104: Kind of feels weird to have a flag to "support" these when it doesn't affect execution, but I guess it enabled/disables the assembly, and in future we might have some kind of platform callback for them...

model/riscv.sail_project
Timmmm: I agree that sounds sensible. Then hopefully you can just say all `hints` are before all `extensions`. I think we said we'd do something similar for maybe operations (Zimop).

config/rv64d.json
jordancarlin: Maybe we add a comment to the keys for now indicating that they only impact assembly?

model/riscv.sail_project
pmundkur: This would simplify the project file but might break if an extension added both hints and instructions.

jordancarlin: Other than the base ISA, I don’t think that’s happened yet. Seems like something we can address if that ever comes up (maybe by splitting the hints into a separate file).

jordancarlin: Decided to keep this as it is for now during today's meeting.

config/rv64d.json
pmundkur: Comments in JSON are non-standard, so I hesitate to do this.  Even so, I'm not sure of the value of the comment?

model/riscv.sail_project
pmundkur: Interestingly, there are proposals floating around for simultaneously adding hints and new instructions:
https://lists.riscv.org/g/sig-vector/topic/risc_v_vector_register_gather/114260761
It was discussed at the Vector SIG today.

config/rv64d.json
jordancarlin: I'm fine with it as is. Was just an option if we were concerned about using the term supported here.

doc/ChangeLog.md
```diff
@@ -5,6 +5,10 @@
   and `config/rv64d.json` respectively. Dynamic switching between
   the 'E' and 'I' base ISAs is not supported.
 
+- The following extensions have been added:
+  . Zihintntl
+  . Zihintpause
```
jordancarlin: Doesn't seem like this is rendering correctly. The extensions are on a single line separated by periods.

Image

model/riscv_insts_zihintntl.sail
```diff
@@ -0,0 +1,37 @@
...
+mapping clause assembly = NTL(op) <-> "ntl." ^ ntl_name(op)
+
+/* ****************************************************************** */
+
+union clause instruction = C_NTL : ntl_type
+
+// C.NTL is encoded as an C.ADD
+mapping clause encdec_compressed = C_NTL(op)
+  <-> 0b100 @ 0b1 @ 0b00000 @ encdec_ntl(op) @ 0b10
+  when currentlyEnabled(Ext_Zihintntl) & currentlyEnabled(Ext_C)
```
jordancarlin: Seems like this should actually be Zca, not C. See https://github.com/riscv/riscv-isa-manual/pull/2215 and https://github.com/riscv/riscv-isa-manual/issues/2212.

jordancarlin (approved):

PR #1210, opened 2025-08-15
Move extension-specific enums from `riscv_types` into their own files. 

pmundkur:
Enums used for the instruction type needed to appear before its definition, causing enums used for instructions from extensions to collect in `riscv_types`.  This moves them into extension-specific `_types` files, making each extension more self-contained.

### DIFF
--- a/model/riscv_ibase_types.sail
+++ b/model/riscv_ibase_types.sail
@@ -1,0 +1,17 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
+
+enum uop = {LUI, AUIPC}                          /* upper immediate ops */
+enum bop = {BEQ, BNE, BLT, BGE, BLTU, BGEU}      /* branch ops */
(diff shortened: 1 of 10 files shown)


pmundkur:
It might be easier to review each commit individually.

There's another similar cleanup for the M extension which I can submit as a followup PR once this is reviewed.

model/riscv_ibase_types.sail
jordancarlin, line 1: Similar to `riscv_insts_base.sail`, I'd just call this `riscv_base_types.sail` since it applies equally to both `I` and `E`.

model/riscv_zaamo_types.sail
jordancarlin, line 1: Should this be named `riscv_aext_types.sail` since it applies to both Zaamo and Zalrsc?

jordancarlin (commented):
A few minor nits, but otherwise seems like a good idea to start moving things out of the generic riscv files.

Timmmm:
I only skimmed the code but this looks like a good idea. I wonder if eventually we should make subdirectories for each extension? And maybe drop the redundant `riscv_` prefix on the files?

model/riscv_ibase_types.sail
jordancarlin: I think something got mixed up. It looks like `riscv_types` was renamed `riscv_base_types` and we still have a separate `riscv_ibase_types` file. I would keep `riscv_types` as it is and change the new `riscv_ibase_types` to just `riscv_base_types` since it corresponds to the base instruction set.

pmundkur: oops.

jordancarlin (approved):

Timmmm (approved):

PR #1211, opened 2025-08-17
Run full testsuite on PRs with `run-full-tests` label

jordancarlin:
Adds a PR label check to the run-full-tests workflow so that it can be run on PRs by adding the `run-full-tests` label.

This is primarily intended for PRs that make significant changes to vector so that the full vector suite can be run before merging.

### DIFF
--- a/.github/workflows/run-full-tests.yml
+++ b/.github/workflows/run-full-tests.yml
@@ -1,9 +1,10 @@
-name: Run risc-v tests
+name: Run full tests
 on:
   workflow_dispatch:
   schedule:
     - cron: "0 0 * * 0"
   pull_request:
+    # Explicit listing of types is necessary to run the workflow when a label is added
     types:
       - opened
       - synchronize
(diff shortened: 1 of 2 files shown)


.github/workflows/run-riscv-tests.yml
```diff
@@ -3,11 +3,20 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "0 0 * * 0"
+  pull_request:
+    types:
```
Timmmm, line 8: Worth a comment to say why this is needed.

.github/workflows/run-riscv-tests.yml
```diff
@@ -3,11 +3,20 @@ on:
...
+      - reopened
+      - labeled
+      - unlabeled
 env:
   SAIL_VERSION: "0.19.1"
   CMAKE_VERSION: "3.20"
 jobs:
   run-riscv-tests:
+    # If triggered by a PR event, only run if the 'run-riscv-tests' label is present
+    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-riscv-tests') }}
```
Timmmm: Kind of confusing label because all of the `riscv-tests` are already run for all PRs. I think we should rename this file and the label. Maybe something like `run-full-tests`?

Timmmm (commented):
Good idea!

.github/workflows/run-riscv-tests.yml
jordancarlin: Agreed. Updated the label and workflow name.

.github/workflows/run-full-tests.yml
```diff
@@ -1,13 +1,23 @@
...
   schedule:
     - cron: "0 0 * * 0"
+  pull_request:
+    # Explicit listing of types is necessary to run the workflow when a label is added
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
```
nadime15, line 13:
```suggestion
    types:
      - opened
      - synchronize
      - reopened
      - labeled
```

I think we can remove `unlabeled` here, because it cant trigger a run anyway and will never satisfy the if condition

jordancarlin: It is necessary so that the job is removed from the PR is the label is removed.

nadime15: Ahh ok I see!

Issue #1196, opened 2025-08-08
Vector segment load/store EEW calculation is wrong.


Timmmm:
E.g. for `vlsegNeM.v` we have:

```
function clause execute(VLSEGTYPE(nf, vm, rs1, width, vd)) = {
  let load_width_bytes = vlewidth_bytesnumber(width);
  let EEW = load_width_bytes * 8;
  let EEW_pow = vlewidth_pow(width);
  let SEW_pow = get_sew_pow();
  let LMUL_pow = get_lmul_pow();
  let EMUL_pow = EEW_pow - SEW_pow + LMUL_pow;
  let num_elem = get_num_elem(EMUL_pow, EEW); /* # of element of each register group */

  assert(num_elem > 0);

  if illegal_load(vd, vm, nf, EEW, EMUL_pow) then return Illegal_Instruction();

  process_vlseg(nf, vm, vd, load_width_bytes, rs1, EMUL_pow, num_elem)
}

mapping clause assembly = VLSEGTYPE(nf, vm, rs1, width, vd)
  <-> "vl" ^ nfields_string(nf) ^ "e" ^ vlewidth_bitsnumberstr(width) ^ ".v" ^ spc() ^ vreg_name(vd) ^ sep() ^ "(" ^ reg_name(rs1) ^ ")" ^ maybe_vmask(vm)
```

`get_sew_pow()` returns the value from `vtype[sew]`. But as far as I can see from the spec that value should be ignored except for indexed loads/stores. The spec isn't super explicit about this - the closest I could find is:

> Vector unit-stride and constant-stride use the EEW/EMUL encoded in the instruction for the data values, while vector indexed loads and stores use the EEW/EMUL encoded in the instruction for the index values and the SEW/LMUL encoded in vtype for the data values.

And also motivation, which only makes sense if `vtype[sew]` is entirely ignored for those instructions:

> Vector memory unit-stride and constant-stride operations directly encode EEW of the data to be transferred statically in the instruction to reduce the number of vtype changes when accessing memory in a mixed-width routine. 

@nadime15 presumably this means the case where `vtype[sew]` is not the same as the EEW from the instruction isn't actually tested at all in the vector test suite, which seems like a fairly big miss!

nadime15:
Hmm yeah we should definitely test these cases. I will give it a try and add these cases to the vector test suite.

pmundkur:
[...]

PR #1213, opened 2025-08-18
Greatly simplify the code for reading & writing vector registers/masks

Timmmm:
The original code was very over-complicated.

Where `assert(-3 <= EMUL_pow & EMUL_pow <= 3);` is added, the calculation of EMUL_pow is clearly wrong - see #1196. Since this was an existing issue I have just added a TODO rather than fixing it.

This also introduces `static_assert()` which can be used for documenting properties that are already known by the type checker to be true (and it will give nice error messages if you edit the code so that it is no longer true).

### DIFF
--- a/model/extensions/V/vext_control.sail
+++ b/model/extensions/V/vext_control.sail
@@ -25,91 +25,25 @@ function clause currentlyEnabled(Ext_V) = hartSupports(Ext_V) & (misa[V] == 0b1)
 function clause currentlyEnabled(Ext_Zvfh)    = hartSupports(Ext_Zvfh)     & currentlyEnabled(Ext_Zve32f) & currentlyEnabled(Ext_Zfhmin)
 function clause currentlyEnabled(Ext_Zvfhmin) = (hartSupports(Ext_Zvfhmin) & currentlyEnabled(Ext_Zve32f)) | currentlyEnabled(Ext_Zvfh)
 
-// num_elem means max(VLMAX,VLEN/SEW)) according to Section 5.4 of RVV spec
-function get_num_elem(LMUL_pow : int, SEW : sew_bitsize) -> nat1 = {
-  let LMUL_pow_reg = if LMUL_pow < 0 then 0 else LMUL_pow;
-  // Ignore lmul < 1 so that the entire vreg is read, allowing all masking to
-  // be handled in init_masked_result
-  let num_elem = (2 ^ LMUL_pow_reg) * vlen / SEW;
-  assert(num_elem > 0);
-  num_elem
-}
-
-// Reads a single vreg into multiple elements
-val read_single_vreg : forall 'n 'm, 'n >= 0 & is_sew_bitsize('m) . (int('n), int('m), vregidx) -> vector('n, bits('m))
-function read_single_vreg(num_elem, SEW, vrid) = {
-  // TODO: This should be part of the type signature for this function.
-  assert(num_elem * SEW <= vlen);
-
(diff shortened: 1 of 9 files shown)


pmundkur (commented):
This simplification has introduced regressions for negative LMUL.

Timmmm:
Yeah I must have made a mistake somewhere. I'll have a look.

Timmmm:
Rebased & fixed. I think there are some further improvements that could be made. E.g. a colleague pointed out we could make an even tighter type for `VLMAX`: `range(vlen/elen, vlen)` but one step at a time!

model/extensions/V/vext_control.sail
pmundkur, line 49: I wonder if we should assert that `vlen % EEW` is zero, to accompany the other asserts in this function.  And similarly in `write_single_element`.

model/extensions/V/vext_control.sail
```diff
@@ -171,58 +99,63 @@ function write_single_element(EEW, index, vrid, value) = {
...
+// vrid:     Index of the first register in the group.
+// data:     Data of each element.
+//
+val write_vreg : forall 'n 'sew, 'n >= 0 & is_sew_bitsize('sew) . (int('n), int('sew), LMUL_pow, vregidx, vector('n, bits('sew))) -> unit
+function write_vreg(num_elem, SEW, LMUL_pow, vrid, vec) = {
+  // Number of registers in the group (treating fractional lmul as 0).
+  let group_size = 2 ^ max(LMUL_pow, 0);
+
+  // Number of SEW-bit elements per vector register.
+  let elem_per_reg = vlen / SEW;
```
pmundkur, line 124: Perhaps another assert here that `vlen % SEW == 0`?

model/extensions/V/vext_control.sail
```diff
@@ -171,58 +99,63 @@ function write_single_element(EEW, index, vrid, value) = {
...
+  // Normally num_elem == group_size * elem_per_reg, however
+  // in the specific case of fractional lmul with widening instructions
+  // get_num_elem() returns vlen / SEW, but SEW is doubled before
+  // feeding into this instruction, so elem_per_reg is half the
+  // pre-widening value. Since `group_size` will be unchanged
+  // (because max(LMUL_pow, 0) is still 0), we will end up
+  // with num_elem == group_size * (elem_per_reg * 2).
+  //
+  // In that case the second half of the `vec` array will be ignored.
+  assert(num_elem >= group_size * elem_per_reg);
```
pmundkur: Why not 
```suggestion
  assert(num_elem == group_size * elem_per_reg | num_elem == 2 * group_size * elem_per_reg);
```

model/extensions/V/vext_control.sail
Timmmm: I think that's statically known... so an `assert()` would technically be redundant. But also I've thought before that we could do something like `static_assert(...);` I wonder if you can write it like this:

```
function static_assert forall ('b : Bool), 'b . (_ : bool('b)) -> unit = ()
```

I'll try  it!

Timmmm: Sweet that actually worked first time! Might be worth bringing this up at the monday meeting to explain it.

model/extensions/V/vext_control.sail
Timmmm: Done

pmundkur (approved):
This is a great improvement!

It would be good to get an explanation of the error messages from `static_assert`.

PR #1215, opened 2025-08-18
Handle empty CLI input by showing help and exiting unsuccessfully

nadime15:
MRunning the program without any arguments previously caused it to exit with a failure due to missing ELF files. Now, if no arguments are provided, the program prints the help text and exits unsuccessfully, without requiring an ELF file.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -564,7 +564,7 @@ int main(int argc, char **argv)
 
   if (argc == 1) {
     fprintf(stdout, "%s\n", app.help().c_str());
-    exit(EXIT_SUCCESS);
+    exit(EXIT_FAILURE);
   }
 
   // long_options_offset() is a local addition, so when updating CLI11,


jrtc27:
You should still print out the error that no files were provided. It's infuriating when you run a command with an incorrect set of arguments and it just dumps out its help to tell you that you did _something_ wrong without saying _what_. Yes, in this case maybe it's obvious, but still. Best practice is to print the error and _then_ the usage.

nadime15:
@jrtc27 I get your point but I was inspired by our current help message which says:

`Usage: [OPTIONS] [elfs...]`

So you could run the tool without any ELFs or Options. However you could just run it and print the current version with no additional files.

The case I added just catches when 0 arguments are provided, I think, as you have said this is kind of obvious.

pmundkur (approved):

c_emulator/riscv_sim.cpp
```diff
@@ -562,6 +562,11 @@ int main(int argc, char **argv)
   argv = app.ensure_utf8(argv);
   setup_options(app);
 
+  if (argc == 1) {
+    fprintf(stdout, "%s\n", app.help().c_str());
+    exit(EXIT_SUCCESS);
```
arichardson: I'd prefer if it exits with a failure return code but no strong opinion here.

Timmmm: I agree. Any situation where you care about the exit code is probably non-interactive, and printing help means something has gone wrong.

Alasdair:
This made me think about the behaviour of `sail` when ran with no arguments, and it also does nothing and exits successfully. Maybe I should change that?

c_emulator/riscv_sim.cpp
nadime15: Shouldn we then modify the usage hint:

`Usage: [OPTIONS] [elfs...]` 

Correct me if I am wrong but this tells you that you can call the tool without any arguments.

I tried a bunch of tools and yes most return non zero in this case but some return 0 for example like rustc returns 0 even though the usage hint is: 

`Usage: rustc [OPTIONS] INPUT`

Timmmm: Well, you can run it without options OR without elfs... I guess that's true with most programs (you can do `--version` at least). But I think it still makes sense to have `[ELF]` because it's not obvious that some of the arguments don't need an elf.

Really we should have subcommands like `sail_riscv_sim run foo.elf` and `sail_riscv_sim print_isa_string` but... does CLI11 support that? I don't think it matters too much anyway.

Does rustc return 0 *and* print a help message with no inputs?

Timmmm:
> Maybe I should change that?

Yeah feels like printing help and exiting with an error would be more useful.

Timmmm (approved):

c_emulator/riscv_sim.cpp
nadime15: `Does rustc return 0 and print a help message with no inputs?`

Yes, but to be honest, I tried a number of tools and it took a while to find one that returns 0 when called with no inputs :D

jordancarlin (approved):

c_emulator/riscv_sim.cpp
Timmmm: I guess it's probably a bug but almost nobody actually runs `rustc` directly so probably nobody noticed.

arichardson (approved):

Issue #1112, opened 2025-06-30
Adding configuration options from UDB


pmundkur:
It would be worth looking at the "parameters" in the UDB project as a source of configuration parameters.   

One way to get these is to look at the PDF artifact
https://riscv-software-src.github.io/riscv-unified-db/pdfs/RVB23ProfileRelease.pdf
and search for "parameters".

Otherwise, from a built UDB, one can do
```
./bin/udb list parameters
```

james-ball-qualcomm:
The vast majority of the parameters will have mappings to normative rules (mostly 1:1 but could be many:1, 1:many, or 1:none). That's because we want parameters to have a link to normative text in the ISA manual that defines the parameters wherever appropriate. So, I'd suggest another starting point is to look at all these parameter normative rules. I've attached the latest set of them checked into the main branch of the ISA manual as an HTML file. Just scroll the navigation bar on the left all the way down like in this screenshot:

<img width="3363" height="1969" alt="Image" src="https://github.com/user-attachments/assets/3492e996-c63c-4274-853e-1b080a1e96c9" />

[norm-rules.html](https://github.com/user-attachments/files/25156648/norm-rules.html)

PR #1216, opened 2025-08-19
Make {m,s}status[VS] read-only zero when Ext_V not supported

trdthg:
This is discovered during testing with spike. When the V extension is disabled in the configuration, running the `csrrs reg sstatus VS_MASK` instruction, spike does not set the VS field, but sail does

### DIFF
--- a/model/riscv_sys_regs.sail
+++ b/model/riscv_sys_regs.sail
@@ -247,7 +247,12 @@ function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
     FS = if hartSupports(Ext_Zfinx) then extStatus_to_bits(Off) else v[FS],
     MPP = if have_privLevel(v[MPP]) then v[MPP] else privLevel_to_bits(lowest_supported_privLevel()),
     SPP = if currentlyEnabled(Ext_S) then v[SPP] else 0b0,
-    VS = v[VS],
+    /* TODO: make this configurable
+     * If neither the v registers nor S-mode is implemented, then VS is
+     * read-only zero. If S-mode is implemented but the v registers are not,
+     * VS may optionally be read-only zero.
+     */
+    VS = if hartSupports(Ext_V) then v[VS] else 0b00,
     MPIE = v[MPIE],
     SPIE = if currentlyEnabled(Ext_S) then v[SPIE] else 0b0,
     MIE = v[MIE],


model/riscv_sys_regs.sail
```diff
@@ -247,7 +247,7 @@ function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
     FS = if hartSupports(Ext_Zfinx) then extStatus_to_bits(Off) else v[FS],
     MPP = if have_privLevel(v[MPP]) then v[MPP] else privLevel_to_bits(lowest_supported_privLevel()),
     SPP = if currentlyEnabled(Ext_S) then v[SPP] else 0b0,
-    VS = v[VS],
+    VS = if hartSupports(Ext_V) then v[VS] else 0b00,
```
nadime15, line 255: I just read the following part from the spec:

> If neither the v registers nor S-mode is implemented, then VS is read-only zero. If S-mode is
> implemented but the v registers are not, VS may optionally be read-only zero.

In my opinion, this means: when S-mode exists but no vector hardware is present, the specification allows the VS field to behave as if vector support existed.

We should leave a `TODO` comment above saying that this should eventually become a configurable option.

trdthg: It's really hard to name this kind of configuration
"xstatus_vs_is_read_only_zero_when_not_support_ext_v_and_support_ext_s"
Any suggestions?

nadime15: I dont think we need to support this configuration yet but it might make sense to add it once someone requests it.

> It's really hard to name this kind of configuration

Yeah I agree :D Maybe something like `vs_smode_only_warl` and later we would access it through `mstatus.vs_smode_only_warl` (as part of the json config file)

jordancarlin: UDB uses the following parameter to control this:

```
  MSTATUS_VS_WRITABLE:
    description: |
      When `S` is enabled but `V` is not, mstatus.VS is optionally writable.
    schema:
      type: boolean
    extra_validation: |
      assert MSTATUS_VS_WRITABLE == true  if ext?(:V)
      assert MSTATUS_VS_WRITABLE == false if (!ext?(:S) && !ext?(:V))
```

So just a generic `MSTATUS_VS_WRITABLE` and then check that it meets the requirements of the enabled extensions. Not sure if this approach is better, but it would be good for sail and UDB to start converging on parameter names and behaviors (see also #1112).

Timmmm: I would just go with `xstatus_vs_read_only` and then we can use the validation check to say it must be true if <condition>.

But yeah it can just be a to-do for now.

Timmmm (approved):

Issue #1136, opened 2025-07-10
Template JSON config instead of using `jq`


Timmmm:
Currently we have the JSON config copy/pasted twice for RV32 and RV64 - I think with two copies the niceness of having the actual JSON there outweighs the repetition.

However since #1003 we also need loads more copies for different VLENs and ELENs. That PR uses `jq` to make the new configs but adding a dependency on `jq` sucks.

We should probably just template the JSON and use CMake's `configure_file()` to generate the files.

PR #1217, opened 2025-08-19
Simplify virtual memory config validation

Timmmm:
The `hartSupports()` clauses already check `xlen` so there's no need to worry about it here.

### DIFF
--- a/model/riscv_validate_config.sail
+++ b/model/riscv_validate_config.sail
@@ -18,39 +18,20 @@ function check_privs() -> bool = {
 // This logic will need to change once MXL/SXL/UXL can differ.
 function check_mmu_config() -> bool = {
   var valid : bool = true;
-  if xlen == 64 then {
-    if not(hartSupports(Ext_S)) & (hartSupports(Ext_Sv57) | hartSupports(Ext_Sv48) | hartSupports(Ext_Sv39))
-    then {
-      valid = false;
-      print_endline("Supervisor mode (S) disabled but one of (Sv57, Sv48, Sv39) is enabled: cannot support address translation without supervisor mode.");
-    };
-    if hartSupports(Ext_Sv57) & not(hartSupports(Ext_Sv48))
-    then {
-      valid = false;
-      print_endline("Sv57 is enabled but Sv48 is disabled: supporting Sv57 requires supporting Sv48.");
-    };
-    if hartSupports(Ext_Sv48) & not(hartSupports(Ext_Sv39))
-    then {
-      valid = false;
-      print_endline("Sv48 is enabled but Sv39 is disabled: supporting Sv48 requires supporting Sv39.");
-    };
-    if hartSupports(Ext_Sv32)
-    then {
-      valid = false;
-      print_endline("Sv32 is enabled: Sv32 is not supported on RV64.");
-    };
-  } else {
-    assert(xlen == 32);
-    if not(hartSupports(Ext_S)) & hartSupports(Ext_Sv32)
-    then {
-      valid = false;
-      print_endline("Supervisor mode (S) is disabled but Sv32 is enabled: cannot support address translation without supervisor mode.");
-    };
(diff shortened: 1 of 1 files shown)


model/riscv_validate_config.sail
```diff
@@ -18,39 +18,20 @@ function check_privs() -> bool = {
...
+  };
+  if hartSupports(Ext_Sv57) & not(hartSupports(Ext_Sv48))
+  then {
+    valid = false;
+    print_endline("Sv57 is enabled but Sv48 is disabled: supporting Sv57 requires supporting Sv48.");
+  };
+  if hartSupports(Ext_Sv48) & not(hartSupports(Ext_Sv39))
+  then {
+    valid = false;
+    print_endline("Sv48 is enabled but Sv39 is disabled: supporting Sv48 requires supporting Sv39.");
```
nadime15, line 34: In #1163 I proposed to use something like:
```
function require_extension(str : string, ext: extension, dep: list(extension)) -> bool = {
  if hartSupports(ext) then {
    foreach (i from 0 to (length(dep) - 1)) {
      if not(hartSupports(dep[i])) then {
        print_endline(str);
        return false;
      }
    }
  };
  true
}
```

with 
```valid = require_extension("Supervisor mode (S) disabled but one of (Sv32, Sv39, Sv48, Sv57) is enabled: cannot support address translation without supervisor mode.", Ext_S, [Ext_Sv32, Ext_Sv39, Ext_Sv48, Ext_Sv57]);```

Do you think it makes sense to use this here, or even at all? I think this pattern will reappear over and over again

Timmmm: Yeah maybe... it does add a fair bit of complexity though. I think we should probably wait until this file is longer so we can see what the common patterns are.

jordancarlin:
I wonder if we're better off removing it from the `hartSupports` clauses so that this validation will actually throw errors if an illegal combination is set. We could also directly using the config values in the validation functions so they are checked before `hartSupports` suppresses the error.

While including the check in `hartSupports` makes sense, I'm concerned that silently disabling features that are enabled in the config file without even printing a warning message will be confusing to some users.

pmundkur:
Agreed,  the configuration validation should make sense on its own, without relying on the checks that are done in `hartsSupports`.

Timmmm:
So the fundamental question is what happens if you mark Sv39 as supported on RV32? Both before and after this change the model will run fine and the validation will succeed. Do we want it to fail?

That will make #1136 more complex. It also means you can't blindly set everything to `true` to get the maximum support. I do think it could make debugging harder - e.g. you set something to `true` and then have to spend some time figuring out why it doesn't actually work, but that's not really too different from e.g. Sv39 being disabled due to `misa[S]` being 0.

But also I can see why you'd want an error if you set Sv57 to `true` and Sv39 to `false`.

pmundkur:
> So the fundamental question is what happens if you mark Sv39 as supported on RV32? Both before and after this change the model will run fine and the validation will succeed. Do we want it to fail?

I say we should fail.

nadime15:
> That will make https://github.com/riscv/sail-riscv/issues/1136 more complex. It also means you can't blindly set everything to true to get the maximum support. I do think it could make debugging harder - e.g. you set something to true and then have to spend some time figuring out why it doesn't actually work, but that's not really too different from e.g. Sv39 being disabled due to misa[S] being 0.

After #1163 this wont be possible anymore anyway. I just recently ran into an [issue](https://github.com/chipsalliance/riscv-vector-tests/issues/78) while debugging the vector test suite. Spike automatically sets VLEN=128 if `v` is in the ISA string, even though `zvl64b` and `zve64d` were also present (`V` overrides `Zvl64b` and sets VLEN=128).

That caused a bunch of failures, and I had no idea where they came from. I think we should catch cases like this. We should let the user set the config however they like, but still double-check in the background whether its valid (even if its means more work) and fail immediately with a clear error message if not.

PR #1219, opened 2025-08-20
Add an option to choose a smaller set of modules for Lean build

Alasdair:
Add a `SAIL_MODULES` variable to the CMakeLists file that allows specifying a smaller set of modules, useful to constraint the Lean build to be something like RV64IM or similar, when that's all we need for some use cases. For example:
```
 cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="I;M;riscv_termination;riscv_main"
```
Not sure if this is the best way to do this with CMake, or if we want to make it work more broadly other than just for Lean.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -6,6 +6,8 @@ function(string_upper_initial s result)
     set("${result}" "${new}" PARENT_SCOPE)
 endfunction()
 
+set(SAIL_MODULES "--all-modules" CACHE STRING "List of modules to compile into the model (defaults to all)")
+
 # Common Sail flags for all commands.
 set(sail_common
     # Don't allow implicit var declaration (like Python). This is
@@ -117,7 +119,7 @@ add_custom_command(
         --c-preserve rvfi_wX
         --c-preserve rvfi_trap
         # Input files.
-        --all-modules
+        ${SAIL_MODULES}
         ${project_file}
 )
 
@@ -163,7 +165,7 @@ add_custom_command(
         # The name of the JSON file.
         --doc-bundle ${model_doc}
         # Input files.
-        --all-modules
+        ${SAIL_MODULES}
         ${project_file}
 )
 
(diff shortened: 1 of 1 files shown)


Timmmm:
I think either this should apply everywhere we use `--all-modules`, or It should be called something like `LEAN_SAIL_MODULES`? Otherwise LGTM.

jordancarlin:
It would be good to add a test build of the model with a subset of the modules to CI. If that doesn’t happen in this PR, let’s open an issue so it doesn’t get forgotten.

Alasdair:
Just updated this PR so it uses the `SAIL_MODULES` variable everywhere, except when determining the `sail_srcs` variable. A few people I know are quite interested in seeing this available so they can work with a reduced model.

I'll make a follow up PR to fix the termination file pulling in too much once this is in.

jordancarlin (approved):
LGTM. Would be good to get a CI job that builds a smaller subset of the modules running once the termination bug is fixed to avoid bitrot.

tobiasgrosser:
I will work on that CI job. Can we merge this PR meanwhile? That will help us unblock some of the other issues?

model/CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ function(string_upper_initial s result)
     set("${result}" "${new}" PARENT_SCOPE)
 endfunction()
 
+if (NOT DEFINED SAIL_MODULES)
+  set(SAIL_MODULES "--all-modules")
+endif()
```
Timmmm: The idiomatic way to have a non-bool option is with `set`. Don't ask me why there's a special case for bools....

```suggestion
set(SAIL_MODULES "--all-modules" CACHE STRING "List of modules to compile into the model (defaults to all)")
```

There's a newer slightly nicer syntax but it requires CMake 4.2.

Timmmm (changes requested):

model/CMakeLists.txt
Alasdair: Done

PR #1220, opened 2025-08-20
Don't use generated reset vector to initialise registers

Timmmm:
Instead of loading and executing actual instructions into the model to set a0=mhartid, a1=&dtb, we can just do that directly in the model. The DTB is always loaded at the start of the ROM so we don't even need to separately tell the model where it is.

This has the advantage that no "extra" instructions are executed before the ELF's which is less confusing and simpifies testing against other simulators.

Finally I simplified how the DTB is actually loaded, and improved the printing of exceptions.


arichardson (approved):
This looks a lot simpler

model/riscv_model.sail
```diff
@@ -24,4 +24,9 @@ function init_model(config_filename : string) -> unit = {
   hart_state = HART_ACTIVE();
   init_platform();
   reset();
+
+  // For compatibility with the de facto boot interface, set a0=mhartid, a1=&dtb.
+  // The C++ code always places the DTB at the start of the ROM.
+  X(Regno(10)) = mhartid;
+  X(Regno(11)) = to_bits_checked(config platform.rom.base : int);
```
jordancarlin, line 35: I think it would be worth having an option (either a config option or a CLI flag) to disable this initialization. This is useful for things like lockstep execution if a DUT does not initialize these registers.

Timmmm: The state of these registers is undefined after reset so you have to set them to a known value anyway.

Let's revisit this when we have a proper lockstep interface. If your lockstep system checks all state (RVVI-style) rather than just state changes then we might need to be able to force all initial values of all registers into the model.

jordancarlin: Works for me.

Timmmm:
Tested with Linux boot; still works.

model/riscv_model.sail
pmundkur: I'd prefer if this was hoisted out into a separate Sail function.  Having it here undoes the effect of `reset` in the previous line.  It would be good to have `init_model` leave the hart in a chip reset state.
Having it in a separate function would also make it easier to have this configurable through a cli option later.

Timmmm: Done

c_emulator/riscv_sim.cpp
```diff
@@ -299,75 +271,29 @@ uint64_t load_sail(const std::string &filename, bool main_file)
...
-    fprintf(stderr,
-            "Configuration value platform.rom.base does not match %" PRIu64
-            ".\n",
-            rom_base);
-  }
-  if (rom_size != get_config_uint64({"platform", "rom", "size"})) {
-    fprintf(stderr,
-            "Configuration value platform.rom.size does not match %" PRIu64
-            ".\n",
-            rom_size);
```
pmundkur, line 347: I missed this earlier, but we might want to retain the zero-filling of the rest of the ROM.  Checking whether it is at a page boundary could probably move to `riscv_validate_config:check_mem_layout()` where there is already a comment about it.

Timmmm: Yeah I wasn't sure about that. Why would it be needed? In any case I don't think we need to bother for now at least because memory is initialised to 0 anyway. In future we should support pseudorandom initialisation...

But I don't see why we would need to fill the page with zeros anyway?

pmundkur: It may have been needed to boot Linux at one time, but it may not be needed anymore given that Linux boots without it.

pmundkur (approved):

PR #1226, opened 2025-08-20
Refactor the `ext_fetch_check_pc` API.

pmundkur:
This makes it return just an option instead of also a possibly modified target address.

The more general API was designed for an earlier CHERI model, but now the generality is not used.

Addresses #1205.

### DIFF
--- a/model/riscv_addr_checks.sail
+++ b/model/riscv_addr_checks.sail
@@ -14,8 +14,7 @@ type ext_fetch_addr_error = unit
  * start_pc: the PC at the start of the current fetch sequence
  * pc:       the PC for the current granule
  *
- * returns:  the *virtual* memory address to use for the fetch.
- *           any address translation errors are reported for pc, not the returned value.
+ * returns:  an optional fetch error.
  */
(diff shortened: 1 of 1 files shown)


pmundkur:
The similar refactor for `ext_control_check_pc` conflicts with #1206, so I'll submit it once that is merged.

model/riscv_addr_checks_common.sail
```diff
@@ -14,10 +14,7 @@
  * extensions would need to define their own functions to override them.
  */
 
-union Ext_FetchAddr_Check ('a : Type) = {
-  Ext_FetchAddr_OK  : virtaddr,  /* PC value to use for the actual fetch */
-  Ext_FetchAddr_Error : 'a
-}
+type Ext_FetchAddr_Check ('a : Type) = option('a)
```
Timmmm: Could probably ditch this and just use `option()` directly.

model/riscv_addr_checks.sail
```diff
@@ -18,7 +18,7 @@ type ext_fetch_addr_error = unit
  *           any address translation errors are reported for pc, not the returned value.
```
Timmmm: This comment needs updating.

Timmmm (approved):

PR #1227, opened 2025-08-21
Fix build failure due to circular dependency introduced by merging #1210 and #1208.

pmundkur:
The dependency on `I` in #1208 needed to be refined to `I_types` after `I` was split into `I_types` and `I_insts` in #1210.


pmundkur:
I've seen flakiness here before: https://github.com/riscv/sail-riscv/pull/1186#issuecomment-3145136560

Issue #1195, opened 2025-08-07
Delete vlewidth_bytesnumber


Timmmm:
In `riscv_insts_vext_mem.sail`, there's a mapping `vlewidth_bytesnumber`, but almost everywhere it is used, it's something like this:

```
  let load_width_bytes = vlewidth_bytesnumber(width);
  let EEW = load_width_bytes * 8;
  let EEW_pow = vlewidth_pow(width);
```

When it could just be

```
  let EEW_pow = vlewidth_pow(width);
  let EEW = 2 ^ EEW_pow;
```

Ideally `vlewidth_pow()` would actually be in bytes and then we just multiply by 8 or add 3 to get bits.

nadime15:
@Timmmm We can close this issue?

jordancarlin:
Resolved in #1229.

PR #1229, opened 2025-08-21
Delete function vlenwidth_bytesnumber

KotorinMinami:
For #1195


Timmmm (approved):
LGTM. It's a shame all of these things were done in bits instead of bytes. I think we should probably change that in future.

pmundkur (approved):

PR #1230, opened 2025-08-21
Fix location of generated configs for install.

pmundkur:
The package build broke since it was looking in the source directory for the generated config files.

### DIFF
--- a/config/CMakeLists.txt
+++ b/config/CMakeLists.txt
@@ -11,7 +11,7 @@ foreach (CONFIG__BASE__XLEN IN ITEMS 32 64)
             set(CONFIG_XLEN_IS_64 "false")
             set(CONFIG_XLEN_IS_${CONFIG__BASE__XLEN} "true")
 
-            configure_file(config.json.in ${config_filename})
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):

config/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ foreach (CONFIG__BASE__XLEN IN ITEMS 32 64)
 
             configure_file(config.json.in ${config_filename})
```
arichardson: I know this is implicitly relative to ${CMAKE_CURRENT_BINARY_DIR}, but might make sense to be explicit here?
```suggestion
            configure_file(config.json.in ${CMAKE_CURRENT_BINARY_DIR}/${config_filename})
```

jordancarlin: ~~Should be `${CMAKE_CURRENT_SOURCE_DIR}`, not `${CMAKE_CURRENT_BINARY_DIR}`.~~
Change showed strangely via email. Disregard.

PR #1231, opened 2025-08-22
feat: migration of old changes and fixes for recent upstreams

dtumad:
Initial changes to support the sp1 infrastructure based on existing fork. Also fixing issues from recent upstream fixes we submitted.


jrtc27:
What on earth is any of this? This is a mishmash of weird, wholly-undocumented changes.

dtumad:
@jrtc27  Sorry, this was meant to be pushed to our fork, should have checked that

PR #1234, opened 2025-08-22
Refactor the control_check_pc API.

pmundkur:
This makes it return just an optional exception instead of also a possibly modified target address.

The more general API was once used for an earlier CHERI model, but now the generality is not used.

Addresses #1205.


nadime15 (approved):

model/riscv_addr_checks_common.sail
```diff
@@ -14,11 +14,6 @@
  * extensions would need to define their own functions to override them.
  */
 
-union Ext_ControlAddr_Check ('a : Type) = {
-  Ext_ControlAddr_OK : virtaddr, /* PC value to use for the target of the control operation */
-  Ext_ControlAddr_Error : 'a
-}
-
 union Ext_DataAddr_Check ('a : Type) = {
```
Timmmm, line 17: I think we could do the same for this (in another PR if you want). @jrtc27 am I right in thinking the latest CHERI never changes the address that's used for memory accesses?

pmundkur: I'll do that in another PR.

jordancarlin (approved):

PR #1241, opened 2025-08-29
Convert most block comments to line comments

### LINKED ISSUE #1240: Convert block comments to line comments
There's a bit of a mix of block and line comments in the Sail code at the moment. IMO there's really never a need to use block comments in modern code. I presume they were originally invented so that commenting large bits of code was less tedious when you were using `ed` in the 70s, but all modern code editors support Ctrl-/ to (un)comment multiple lines. They do also enable comments like this:

```
let a = x /* blah blah */ + y /* blah blah */ + z;
```

But I think that's pretty bad style in general.

Zig doesn't support them at all - see https://github.com/ziglang/zig/issues/161 - which doesn't seem to have been too popular but I think was a sound decision. We probably don't need to go so extreme; I think just changing all the block comments (except maybe the license headers?) to line comments will make it pretty obvious what the intended style is.

Timmmm:
Automatic conversion of block comments to line comments. There are some minor places where it got it slightly wrong which I fixed manually.

Fixes #1240


model/riscv_ext_regs.sail
```diff
@@ -1,25 +1,22 @@
...
-/* This file contains register handling functions that can be
- * overridden by extensions.
- */
+// This file contains register handling functions that can be
+// overridden by extensions.
 
-/*!
-THIS(csrno, priv, isWrite) allows an extension to block access to csrno,
-at Privilege level priv. It should return true if the access is allowed.
-*/
```
arichardson: Not sure what the equivalent line comment version of doc comments is in sail?

In LLVM function doc comments generally use a triple slash `///` (which some IDEs understand and highlight them differently and parse the markup inside the doc comment.

The other style I've seen is `//!` which is apparently also valid for doxygen https://www.doxygen.nl/manual/docblocks.html.

I don't think we are currently using any of the doc comments here, but in the past we were using the function annotations in the sail model to generate nicely formatted LaTeX for the CHERI specification.

I think the /*! comments should use whatever line-doc-comment style sail understands (@Alasdair?).

arichardson (dismissed):
I'd be in favour of this. Just need to figure out how to correctly replace the `/*!` comments (`///` or `//!` -- if possible I'd prefer the former)

model/riscv_ext_regs.sail
Timmmm: Hmm I looked at the source, but unfortunately it only supports block doc comments. So I think our options are:

1. Just turn them into normal comments.
2. Use `///` or `//!` and hope that Sail one day supports that syntax.

I'd be in favour of 1. We don't use doc comments for anything currently and if we ever do then we'll probably want to go through the whole codebase and improve the comments anyway.

jordancarlin: Looks like there are only 3 instances of doc comments in the model anyway, so just sticking with normal comments for now seems reasonable.

arichardson: That's fair I think they were mostly used in the sail-cheri-riscv fork. I don't think we ever pulled any function doc comments from the 'base' model into the generated LaTeX.

I have a slight preference for 2, but don't really feel strongly about this since there are only 3 uses.

If we are removing the doc comments, I'd also vote for removing the `THIS(csrno, priv, isWrite)` markup that doesn't make sense otherwise.

Alasdair:
I can add Rust-style `///` documentation comments. I could also change `/*!` into `/**` to match that if people would accept the breaking change.

model/riscv_ext_regs.sail
Timmmm: Ah is that something Sail understands? I thought it was just the author being weird :-D

jrtc27:
What's the point of this change? Which style is more prevalent? If C-style comments are more common then we should convert the C++-style comments to that. If C++-style comments are more common then fine we can do this. But the problem is mixing, not one over the other. That's just personal preference.

jrtc27:
> I can add Rust-style `///` documentation comments. I could also change `/*!` into `/**` to match that if people would accept the breaking change.

Please don't break sail-cheri-riscv's use of saildoc. Adding new syntax is fine, but please don't remove syntax we're using.

Alasdair:
It would make `///` and `/*!` inconsistent with Rust, as `//!` and `/*!` in Rust are a different type of documentation comment that attach to the thing they are inside rather than the next syntactic item.

model/riscv_ext_regs.sail
Alasdair: Yes it just substitutes the name of the identifier you are documenting so you don't have to re-write it every time you want to reference it.

Timmmm: Ah neat feature. Haven't seen that before.

jrtc27:
So can we add the new syntax for what we currently have, wait a bit for sail-cheri-riscv to migrate, and only then remove, or change the meaning of, the existing syntax?

Timmmm:
Since we only have 3 doc comments... I think I'll just leave them as-is for this PR. We can discuss them separately if anyone strongly cares.

> What's the point of this change?

1. To get consistent style.
2. To make editing comments easier.

> Which style is more prevalent?

It's a mix but I think probably block comments are a little more common (maybe 60:40). But I think that's probably partly because the official VSCode extension doesn't highlight line comments correctly (or it didn't last time I checked) so people probably assumed they aren't supported.

IMO line comments are better because:

1. Better IDE support: Ctrl+/ doesn't work with block comments.
2. Easier to edit - especially for single-line comments, you don't need any end tag.
3. Easier to format. No question about starting the lines with `*` or should the `/*` or `*/` go on their own lines or how should they be indented.

jrtc27:
> Since we only have 3 doc comments... I think I'll just leave them as-is for this PR. We can discuss them separately if anyone strongly cares.
> 
> > What's the point of this change?
> 
> 1. To get consistent style.
> 2. To make editing comments easier.
> 
> > Which style is more prevalent?
> 
> It's a mix but I think probably block comments are a little more common (maybe 60:40). But I think that's probably partly because the official VSCode extension doesn't highlight line comments correctly (or it didn't last time I checked) so people probably assumed they aren't supported.

I don't think people were using VSCode to write the Sail model until recently.

> IMO line comments are better because:
> 
> 1. Better IDE support: Ctrl+/ doesn't work with block comments.

Ctrl-/ is for quick-and-dirty commenting out of lines of code (i.e. not things that should ever be committed, and thus don't need to conform to any style). Are people actually using it to write comments from scratch rather than just typing the comment characters? The main selling point of Ctrl-/ is you can use it from anywhere in the line to comment it out, but if you're writing a comment your cursor already is at the place the comment should start.

> 2. Easier to edit - especially for single-line comments, you don't need any end tag.

I have never in my life found myself hindered by the need to type `*/`, and I would imagine that in the modern shiny text editors IDEs people are using that the `*/` is automatically inserted for you (if not you should find the option to turn it on, it will make your life easier when you have to deal with any project that chooses to use C-style comments).

> 3. Easier to format. No question about starting the lines with `*` or should the `/*` or `*/` go on their own lines or how should they be indented.

`//` have their own questions: Is there a space after? Do block comments get initial `//` lines with no other content around them to delineate them? There's always going to be subjectivity no matter what you pick. But also most projects using C-style comments these days have settled on the multi-line format being:

```
code
/*
 * Big long comment
 * goes here.
 */
code
```

I will say my personal preference is for that style. But moreso because I find:

```
code
// Big long comment
// goes here.
code
```

to not have enough visual separation between the comment and the surrounding code, and that:

```
code
//
// Big long comment
// goes here
//
code
```

is more readable, but looks a bit odd compared with the C-style comments where the `/*` and `*/` provide a clear bracketing yet also a more natural separation.

But that's all personal preference, which I think is what's really going on on your side. Both styles have their pros and cons, and no one style is objectively better. You perceive C++-style as better, which is ok, but your message comes across to me as slightly clouding your ability to appreciate what's objective and subjective, and what the benefits of other styles are.

Timmmm:
I guess you could say its my personal preference for easier-to-edit comment styles. :-D

Anyway I have updated the diff to leave the doc comments unchanged, and also I manually fixed a few very minor issues (some extra lines with only `//`, and one of the block comments was indented incorrectly).

I also tediously skimmed the entire diff to check everything looks ok, which it does.

model/riscv_device_tree.sail
```diff
@@ -32,46 +32,45 @@ function generate_isa_base() -> string = {
...
+// multi-letter extensions by an underscore. If multiple
+// supervisor-level extensions are listed, they should be ordered
+// alphabetically.
+//
+// . Standard machine-level extensions (prefixed with the letters "Sm")
+// should be listed after standard lesser-privileged extensions, and
+// like other multi-letter extensions, must be separated from other
+// multi-letter extensions by an underscore. If multiple
+// machine-level extensions are listed, they should be ordered
+// alphabetically.
```
pmundkur: These paragraphs lost their indentation.

model/riscv_freg_type.sail
```diff
@@ -27,11 +27,10 @@ val fregval_into_freg : flenbits -> fregtype
...
 
-/* Rounding Mode
-    Rounding modes occur as a 3-bit field in F,D instructions,
-    and also as a 3-bit 'frm' field in the 'fcsr' CSR.
-    RISC-V uses the following IEEE-defined rounding modes.
-*/
+// Rounding Mode
+// Rounding modes occur as a 3-bit field in F,D instructions,
+// and also as a 3-bit 'frm' field in the 'fcsr' CSR.
+// RISC-V uses the following IEEE-defined rounding modes.
```
pmundkur: It would be nice to preserve the original indentation.

model/riscv_insts_vext_utils.sail
```diff
@@ -1,62 +1,58 @@
...
   sep() ^ "v0.t"  <-> 0b0
 }
 
-/* Check for valid EEW and EMUL values in:
- *  1. vector widening/narrowing instructions
- *  2. vector load/store instructions
- */
+// Check for valid EEW and EMUL values in:
+// 1. vector widening/narrowing instructions
+// 2. vector load/store instructions
```
pmundkur: It would be good to preserve the original indentation.  There are several such places in this file.

model/riscv_insts_zfa.sail
```diff
@@ -554,18 +554,18 @@ function clause execute (FMVP_D_X(rs2, rs1, rd)) = {
...
-  /* Concatenate the two values using '@' operator */
-  /* e.g. */
-  /*   rs1 = 0x01234567 */
-  /*   rs2 = 0x89abcdef */
-  /*   rd  = rs1 @ rs2 => 0x89abcdef01234567 */
+  // Concatenate the two values using '@' operator
+  // e.g.
+  // rs1 = 0x01234567
+  // rs2 = 0x89abcdef
+  // rd  = rs1 @ rs2 => 0x89abcdef01234567
```
pmundkur: Lost indentation.

model/riscv_mem.sail
```diff
@@ -1,43 +1,42 @@
...
+// mem_write_value_priv - uses a default value for metadata
+// mem_write_value      - both of the above partial applications
+//
+// The internal implementation first performs a PMP check (if PMP is
+// enabled), and then dispatches to MMIO regions or physical memory as
+// per the platform memory map.
+//
+// A helper to check address alignment is also provided in the external API.
+//
+// is_aligned_addr - checks for alignment of a physical or virtual address
```
pmundkur: Lost indentation.

model/riscv_vmem_utils.sail
```diff
@@ -1,70 +1,65 @@
...
+// For write accessors, there are functions for each kind of source
+// data register type (integer, floating point, vector).
+//
+// The external API of this module is:
+//
+// check_misaligned - checks for the alignment of a virtual address given platform settings
+//
+// vmem_read        - reads data of the specified width
+// vmem_write       - writes a specified data payload
+
```
pmundkur: More lost indentation.

model/riscv_device_tree.sail
Timmmm: Fixed.

model/riscv_freg_type.sail
Timmmm: This comment was also in the wrong place. I've moved it.

model/riscv_insts_vext_utils.sail
Timmmm: Fixed

model/riscv_insts_zfa.sail
Timmmm: Fixed

model/riscv_mem.sail
Timmmm: Fixed

model/riscv_vmem_utils.sail
Timmmm: Fixed

model/riscv_fdext_regs.sail
```diff
@@ -414,11 +413,11 @@ mapping freg_or_reg_name : fregidx <-> string = {
...
-/* **************************************************************** */
-/* Floating Point CSR                                               */
-/*     fflags    address 0x001    same as fcrs [4..0]               */
-/*     frm       address 0x002    same as fcrs [7..5]               */
-/*     fcsr      address 0x003                                      */
+// ***************************************************************
+// Floating Point CSR
+// fflags    address 0x001    same as fcrs [4..0]
+// frm       address 0x002    same as fcrs [7..5]
+// fcsr      address 0x003
```
pmundkur, line 420: It still lost indentation here, but this seems less important.

model/riscv_sys_regs.sail
```diff
@@ -939,16 +931,15 @@ mapping clause csr_name_map = 0x7a2  <-> "tdata2"
...
- *       appears _after_ this one in the compile order meaning the valspec
- *       for this function is unavailable when it's first encountered in
- *       read_seed_csr. Hence it appears here.
- */
+//
+// Entropy Source - Platform access to random bits.
+// NOTE: This would be better placed in riscv_platform.sail, but that file
+// appears _after_ this one in the compile order meaning the valspec
+// for this function is unavailable when it's first encountered in
+// read_seed_csr. Hence it appears here.
```
pmundkur, line 942: Another one here.

Timmmm:
I fixed those too. I'll update the style guide in a separate PR so it's a bit easier to review.

Issue #1074, opened 2025-06-19
process_rfvv_single and process_rfvv_widen don't set vstart=0 when vl=0


Timmmm:
There's an early return in these functions for vl=0 but it doesn't set vstart=0.

Timmmm:
Ditto `process_vlre`, `process_vsre`, `execute(RIVVTYPE`, `execute(RMVVTYPE`, `execute(VVMTYPE`, `execute(VVMCTYPE`, `execute(VVCMPTYPE`, `execute(VXMTYPE`, `execute(VXMCTYPE`, `execute(VXCMPTYPE`, `execute(VIMTYPE`, `execute(VIMCTYPE`, `execute(VICMPTYPE`

PR #1243, opened 2025-08-31
make sure vstart is zero when instruction early return

KotorinMinami:
for #1074

### DIFF
--- a/model/extensions/V/vext_red_insts.sail
+++ b/model/extensions/V/vext_red_insts.sail
@@ -64,7 +64,7 @@ function clause execute(RIVVTYPE(funct6, vm, vs2, vs1, vd)) = {
   write_single_element(SEW_widen, 0, vd, sum);
   // other elements in vd are treated as tail elements, currently remain unchanged
   // TODO: configuration support for agnostic behavior
-  set_vstart(zeros());
+  // illegal_widening_reduction checks vstart == 0, and raises illegal-instruction otherwise.
   RETIRE_SUCCESS
 }
 
@@ -135,7 +135,7 @@ function clause execute(RMVVTYPE(funct6, vm, vs2, vs1, vd)) = {
   write_single_element(SEW, 0, vd, sum);
   // other elements in vd are treated as tail elements, currently remain unchanged
   // TODO: configuration support for agnostic behavior
-  set_vstart(zeros());
+  // illegal_reduction checks vstart == 0, and raises illegal-instruction otherwise.
   RETIRE_SUCCESS
 }
 
--- a/model/extensions/V/vext_fp_red_insts.sail
+++ b/model/extensions/V/vext_fp_red_insts.sail
@@ -66,7 +66,7 @@ function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_po
   write_single_element(SEW, 0, vd, sum);
   // other elements in vd are treated as tail elements, currently remain unchanged
   // TODO: configuration support for agnostic behavior
-  set_vstart(zeros());
+  // illegal_fp_reduction checks vstart == 0, and raises illegal-instruction otherwise.
   RETIRE_SUCCESS
 }
 
@@ -106,7 +106,7 @@ function process_rfvv_widening_reduction(funct6, vm, vs2, vs1, vd, num_elem_vs, 
   write_single_element(SEW_widen, 0, vd, sum);
   // other elements in vd are treated as tail elements, currently remain unchanged
   // TODO: configuration support for agnostic behavior
-  set_vstart(zeros());
+  // illegal_fp_widening_reduction checks vstart == 0, and raises illegal-instruction otherwise.
   RETIRE_SUCCESS
 }
 
--- a/model/extensions/V/vext_vm_insts.sail
+++ b/model/extensions/V/vext_vm_insts.sail
@@ -43,7 +43,7 @@ function clause execute(VVMTYPE(funct6, vs2, vs1, vd)) = {
 
   let (initial_result, mask) : (bits('n), bits('n)) = match init_masked_result_carry(num_elem, SEW, LMUL_pow, vd_val) {
     Ok(v)   => v,
-    Err(()) => return Illegal_Instruction()
+    Err(()) => return Illegal_Instruction(),
   };
   var result = initial_result;
 


model/riscv_insts_vext_fp_red.sail
```diff
@@ -32,10 +32,16 @@ function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_po
   let rm_3b = fcsr[FRM];
   let num_elem_vd = get_num_elem(0, SEW); /* vd regardless of LMUL setting */
 
-  if illegal_fp_reduction(SEW, rm_3b) then return Illegal_Instruction();
+  if illegal_fp_reduction(SEW, rm_3b) then {
+    set_vstart(zeros());
```
Timmmm: Wouldn't it be simpler just to `set_vstart(zeros())` at the start of each instruction unconditionally?

KotorinMinami: This could potentially affect certain instructions that use the vstart register, theoretically, this is the case, and furthermore, it also fails to comply with the specification.

Timmmm: I don't follow. It is spec compliant to only support vstart=0 which is what the model currently does everywhere as far as I remember. Arguably we should support other vstarts, especially for loads/stores, but we can cross that bridge later.

To be clear, when I say "at the start of each instruction" I don't mean globally for all instructions. Just at the start of instructions that are required to set vstart.

radimkrcmar: I don't think the spec allows us to zero `vstart` on `Illegal_Instruction`.  30.3.7. Vector Start Index (vstart) Register:
```
vstart is not modified by vector instructions that raise illegal-instruction exceptions.
```
(The patch should be changed to only zero on RETIRE_SUCCESS.)

I think we already set `vstart` to non-zero value for memory violations during vector load/store, and we also allow software to set `vstart` through the `vstart` CSR, so zeroing at the beginning would be provably incorrect.
We could raise illegal-instruction exception if the `vstart` value at the beginning isn't something that the implementation can produce.

Timmmm: Yeah that makes sense. So the original code is correct then.

radimkrcmar: It is correct, we're just inconsistent with zeroing `vstart`.

`illegal_fp_reduction` checks `vstart == 0`, and raises illegal-instruction otherwise.  It's weird that the early `return RETIRE_SUCCESS` doesn't zero `vstart` and the final one does, when both code paths assume that `vstart` was zero.

I think that removing the zeroing before the final return will cause less confusion to readers than adding the zeroing before the early return.  A comment that we're not zeroing `vstart` because we've checked `vstart == 0` might help as well.

Timmmm: Hmm good point. Yeah I agree replacing the final zeroing with a comment might make sense.

model/riscv_insts_vext_fp_red.sail
```diff
@@ -32,10 +32,14 @@ function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_po
...
-  if illegal_fp_reduction(SEW, rm_3b) then return Illegal_Instruction();
+  if illegal_fp_reduction(SEW, rm_3b) then {
+    return Illegal_Instruction();
+  };
   assert(SEW != 8);
 
-  if unsigned(vl) == 0 then return RETIRE_SUCCESS; /* if vl=0, no operation is performed */
+  if unsigned(vl) == 0 then {
+    return RETIRE_SUCCESS; /* if vl=0, no operation is performed */
+  };
```
Timmmm: I don't think we need to add braces to these really...

model/riscv_insts_vext_fp_red.sail
```diff
@@ -75,7 +75,9 @@ function process_rfvv_widening_reduction(funct6, vm, vs2, vs1, vd, num_elem_vs,
   let rm_3b          = fcsr[FRM];
   let SEW_widen      = SEW * 2;
 
-  if illegal_fp_widening_reduction(SEW, rm_3b, SEW_widen) then return Illegal_Instruction();
+  if illegal_fp_widening_reduction(SEW, rm_3b, SEW_widen) then {
+    return Illegal_Instruction();
+  };
```
Timmmm: And here. Otherwise lgtm

model/extensions/V/vext_red_insts.sail
```diff
@@ -133,9 +133,9 @@ function clause execute(RMVVTYPE(funct6, vm, vs2, vs1, vd)) = {
   };
 
   write_single_element(SEW, 0, vd, sum);
-  // other elements in vd are treated as tail elements, currently remain unchanged
-  // TODO: configuration support for agnostic behavior
-  set_vstart(zeros());
+  /* other elements in vd are treated as tail elements, currently remain unchanged */
+  /* TODO: configuration support for agnostic behavior */
```
Timmmm: Minor rebase issue with these comments...

Timmmm:
Hmm thinking about this more I'm not sure about it. Maybe what would actually be least confusing is if do something like this:

```
val process_rfvv_single : (rfvvfunct6, bits(1), vregidx, vregidx, vregidx, nat1, sew_bitsize, range(-3, 3)) -> ExecutionResult
function process_rfvv_single(funct6, vm, vs2, vs1, vd, num_elem_vs, SEW, LMUL_pow) = {
  let rm_3b = fcsr[FRM];
  let num_elem_vd = get_num_elem(0, SEW); // vd regardless of LMUL setting

  if illegal_fp_reduction(SEW, rm_3b) then return Illegal_Instruction();
  assert(SEW != 8);

  if unsigned(vl) != 0 then {
    let 'n = num_elem_vs;
    let 'd = num_elem_vd;
    let 'm = SEW;

    let vm_val  : bits('n)     = read_vmask(num_elem_vs, vm, zvreg);
    let vd_val  : vector('d, bits('m)) = read_vreg(num_elem_vd, SEW, 0, vd);
    let vs2_val : vector('n, bits('m)) = read_vreg(num_elem_vs, SEW, LMUL_pow, vs2);
    let mask    : bits('n)     = match init_masked_source(num_elem_vs, LMUL_pow, vm_val) {
      Ok(v)   => v,
      Err(()) => return Illegal_Instruction()
    };

    var sum : bits('m) = read_single_element(SEW, 0, vs1); // vs1 regardless of LMUL setting
    foreach (i from 0 to (num_elem_vs - 1)) {
      if mask[i] == bitone then {
        sum = match funct6 {
          // currently ordered/unordered sum reductions do the same operations
          FVV_VFREDOSUM   => fp_add(rm_3b, sum, vs2_val[i]),
          FVV_VFREDUSUM   => fp_add(rm_3b, sum, vs2_val[i]),
          FVV_VFREDMAX    => fp_max(sum, vs2_val[i]),
          FVV_VFREDMIN    => fp_min(sum, vs2_val[i]),
          _               => internal_error(__FILE__, __LINE__, "Widening op unexpected")
        }
      }
    };

    write_single_element(SEW, 0, vd, sum);
    // other elements in vd are treated as tail elements, currently remain unchanged
    // TODO: configuration support for agnostic behavior
  };
  
  set_vstart(zeros());
  RETIRE_SUCCESS
}
```

Also `Err(()) => return Illegal_Instruction()` is only coincidentally correct because vstart must be zero. If we supported non-zero vstart then vstart=10, vl=0 presumably should raise an exception.

Maybe we should discuss this in the meeting.

pmundkur:
I prefer what you proposed, since I think it would be actually less confusing to have `vstart` zeroed at the end.  When we support non-zero `vstart`, the execution clause will have to zero it at the end anyway.

> Also `Err(()) => return Illegal_Instruction()` is only coincidentally correct because vstart must be zero. If we supported non-zero vstart then vstart=10, vl=0 presumably should raise an exception.

In this case (vstart=10,vl=0), won't the `illegal_fp_reduction()` check raise the exception?

PR #1245, opened 2025-08-31
Add Ziccif and Ziccid extensions

ved-rivos:
If the Ziccif extension is implemented, main memory regions with both the cacheability and coherence PMAs must support instruction fetch, and any instruction fetches of naturally aligned power-of-2 sizes of at most min(ILEN,XLEN) bits are atomic. This fetch state machine is updated to be compliant to this extension. 

The Ziccid extension mandates more stringent requirements for consistency between instruction fetches and other memory accesses than those imposed by the base ISA. The primary intent of the Ziccid extension is to accelerate JIT compilation in multiprocessor systems.In straightforward implementations, maintaining coherence between instruction caches and the data-memory system suffices to satisfy this extension’s strictures. The Ziccid extension can be viewed as a means to codify the concept of instruction-cache coherence. This is primarily added as an ISA string to the model.

Update has been tested with the units tests.

Specifcation: https://github.com/aswaterman/riscv-misc/blob/main/isa/ziccid.adoc

### DIFF
--- a/model/riscv_termination.sail
+++ b/model/riscv_termination.sail
@@ -81,6 +81,7 @@ function currentlyEnabled_measure(ext : extension) -> int =
     Ext_Zcmop => 3,
     Ext_Zfhmin => 3,
     Ext_Zhinx => 3,
+    Ext_Ziccid => 3,
     Ext_Zvkb => 3,
     Ext_Sscofpmf => 3,
     Ext_Svrsw60t59b => 3,


aswaterman (approved):
LGTM, but I'm not one of the SAIL model experts.

ved-rivos:
@pmundkur - could you please help me with cause for the lean backend failure

model/riscv_extensions.sail
```diff
@@ -112,6 +112,14 @@ function clause hartSupports(Ext_Zihpm) = config extensions.Zihpm.supported
...
 mapping clause extensionName = Ext_Zimop <-> "zimop"
 function clause hartSupports(Ext_Zimop) = config extensions.Zimop.supported
+// Instruction-Fetch Atomicity
+enum clause extension = Ext_Ziccif
+mapping clause extensionName = Ext_Ziccif <-> "ziccif"
+function clause hartSupports(Ext_Ziccif) = config extensions.Ziccif.supported
+// Instruction/Data Coherence and Consistency
+enum clause extension = Ext_Ziccid
+mapping clause extensionName = Ext_Ziccid <-> "ziccid"
+function clause hartSupports(Ext_Ziccid) = config extensions.Ziccid.supported
```
jordancarlin: Guard this extension behind the `--enable-experimental-extensions` flag until it is ratified:
```suggestion
function clause hartSupports(Ext_Ziccid) = config extensions.Ziccid.supported : bool & sys_enable_experimental_extensions()
```

model/riscv_extensions.sail
```diff
@@ -402,6 +410,8 @@ let extensions_ordered_for_isa_string = [
   // Zi
   Ext_Zicbom,
   Ext_Zicboz,
+  Ext_Ziccif,
+  Ext_Ziccid,
```
jordancarlin: These are reversed for canonical ordering.

model/riscv_fetch.sail
```diff
@@ -12,12 +12,37 @@
 
 function isRVC(h : half) -> bool = not(h[1 .. 0] == 0b11)
 
-function fetch() -> FetchResult = {
-  /* fetch PC check for extensions: extensions return a transformed PC to fetch,
-   * but any exceptions use the untransformed PC.
-   */
-  // TODO: Add a parameter to try_step() to allow forcing the fetch result and use that instead.
+function read2_at(vaddr : xlenbits) -> result(bitvector(16), FetchResult) = {
```
jordancarlin:
```suggestion
function read2_at(vaddr : xlenbits) -> result(bits(16), FetchResult) = {
```

model/riscv_fetch.sail
```diff
@@ -12,12 +12,37 @@
...
+        Ok(paddr,  _) =>
+          match mem_read(InstructionFetch(), paddr, 2, false, false, false) {
+            Err(e)  => Err(F_Error(e, vaddr)),
+            Ok(ilo) => Ok(ilo)
+          }
+      }
+  }
+}
+
+function read4_at(vaddr : xlenbits) -> result(bitvector(32), FetchResult) = {
```
jordancarlin:
```suggestion
function read4_at(vaddr : xlenbits) -> result(bits(32), FetchResult) = {
```

model/riscv_fetch.sail
```diff
@@ -12,12 +12,37 @@
...
+      match translateAddr(Virtaddr(vaddr), InstructionFetch()) {
+        Err(e, _)     => Err(F_Error(e, vaddr)),
+        Ok(paddr,  _) =>
+          match mem_read(InstructionFetch(), paddr, 4, false, false, false) {
+            Err(e)  => Err(F_Error(e, vaddr)),
+            Ok(w32) => Ok(w32)
+          }
+      }
+  }
+}
```
jordancarlin: Seems like these could be combined into a single function that takes a parameter indicating how many bytes to read. It can be constrained to only accept 2 or 4 as legal values.

README.md
```diff
@@ -125,6 +125,8 @@ For booting operating system images, see the information under the
 - Sv32, Sv39, Sv48 and Sv57 page-based virtual-memory systems
 - Svbare extension for Bare mode virtual-memory translation
 - Physical Memory Protection (PMP)
+- Ziccif Extension for Instruction-Fetch Atomicity
+- Ziccid Extension for Instruction/Data Coherence and Consistency
```
jordancarlin: Please include version numbers as is done for the other extensions in this list. It is also (mostly) in canonical order, so move Ziccif up to be with the other Zi* extensions. Ziccid should be in the list below for unratified extensions.

model/riscv_extensions.sail
```diff
@@ -112,6 +112,14 @@ function clause hartSupports(Ext_Zihpm) = config extensions.Zihpm.supported
 enum clause extension = Ext_Zimop
 mapping clause extensionName = Ext_Zimop <-> "zimop"
 function clause hartSupports(Ext_Zimop) = config extensions.Zimop.supported
+// Instruction-Fetch Atomicity
+enum clause extension = Ext_Ziccif
+mapping clause extensionName = Ext_Ziccif <-> "ziccif"
+function clause hartSupports(Ext_Ziccif) = config extensions.Ziccif.supported
```
jordancarlin, line 126: Ziccif and Ziccid need `currentlyEnabled` clauses. I believe this is also the cause of the Lean build failure.

The following should work:

```
function clause currentlyEnabled(Ext_Ziccif) = hartSupports(Ext_Ziccif)
```
 
```
function clause currentlyEnabled(Ext_Ziccid) = hartSupports(Ext_Ziccid) & currentlyEnabled(Ext_Ziccif)
```

model/riscv_extensions.sail
ved-rivos: Updated.

model/riscv_extensions.sail
ved-rivos: updated

model/riscv_fetch.sail
ved-rivos: updated

model/riscv_fetch.sail
ved-rivos: updated

model/riscv_fetch.sail
ved-rivos: merged

README.md
ved-rivos: done

model/riscv_extensions.sail
ved-rivos: done

ved-rivos:
@jordancarlin Could you suggest how to resolve this failure?
```
/home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D/RiscvExtensions.setup.json --json
error: LeanRV64D/RiscvExtensions.lean:544:37: Unknown identifier `sys_enable_experimental_extensions`
error: build failed
```

model/riscv_fetch.sail
```diff
@@ -29,33 +40,44 @@ function fetch() -> FetchResult = {
...
-       */
-      match mem_read(InstructionFetch(), ppclo, 2, false, false, false) {
-        Err(e)  => F_Error(e, PC),
-        Ok(ilo) =>
-          if   isRVC(ilo)
-          then F_RVC(ilo)
-          else {
-            /* fetch PC check for the next instruction granule */
-            let PC_hi = PC + 2;
-            match ext_fetch_check_pc(PC, PC_hi) {
```
jrtc27, line 44: You can't just change how these hooks are called. CHERI needs to be able to bounds check the program counter, and it needs to be precise.

model/riscv_fetch.sail
```diff
@@ -29,33 +40,44 @@ function fetch() -> FetchResult = {
...
-            };
-
-            match translateAddr(Virtaddr(PC_hi), InstructionFetch()) {
-              Err(e, _)    => F_Error(e, PC_hi),
-              Ok(ppchi, _) =>
-                match mem_read(InstructionFetch(), ppchi, 2, false, false, false) {
-                  Err(e)  => F_Error(e, PC_hi),
-                  Ok(ihi) => F_Base(append(ihi, ilo))
-                }
+  // An implementation with the Ziccif extension fetches instructions in a manner
```
jrtc27, line 42: So how do I turn this off and get the old fetch-16-bits-at-a-time approach that is also a valid RISC-V implementation?

model/riscv_fetch.sail
```diff
@@ -29,33 +40,44 @@ function fetch() -> FetchResult = {
...
+        if isRVC(w32[15..0]) then
+          F_RVC(w32[15..0])
+        else
+          F_Base(w32)
+      }
+    }
+  } else {
+    // PC % 4 == 2:
+    // Step 1: atomic 2B at PC.
+    // If not compressed, Step 3: atomic 4B at PC+2; use only its low halfword.
```
jrtc27: Why does this need to be an atomic 4B fetch that throws away 2B? Isn't that indistinguishable, aside from if you're snooping the bus or fetching from side-effecting memory, from fetching 2B?

jrtc27: To me it seems like the sensible approach would be "if Ziccif is enabled and PC is aligned and ext_fetch_check_pc grants access to the entire 4 bytes then fetch 4 bytes atomically, otherwise use the old pattern (fetch 2 bytes and conditionally fetch a further 2 bytes, making sure to include the ext_fetch_check_pc checks)".

Alasdair:
> @jordancarlin Could you suggest how to resolve this failure?
> 
> ```
> /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D/RiscvExtensions.setup.json --json
> error: LeanRV64D/RiscvExtensions.lean:544:37: Unknown identifier `sys_enable_experimental_extensions`
> error: Lean exited with code 1
> error: build failed
> ```

Don't worry too much about Lean backend failures, those of use who are working on Sail to Lean can fix issues, and we can merge PRs that fail this particular job.

model/riscv_fetch.sail
ved-rivos: Please see the Ziccif specification - step 3.

model/riscv_fetch.sail
ved-rivos: fetching byte by byte is also a valid implementation. While there are many valid implementations, there is now a specified implementation. Among the many valid but adhoc implementations we can retain the one specified implementation.

model/riscv_fetch.sail
ved-rivos: The hook still exists. Please see read_insn_bytes function.

ved-rivos:
> 
> Don't worry too much about Lean backend failures, those of use who are working on Sail to Lean can fix issues, and we can merge PRs that fail this particular job.

Thanks @Alasdair. I wanted to make sure that it wasnt my code that caused a breakage.

model/riscv_fetch.sail
jrtc27: Yes but you're now calling it in different situations with different expectations, so it doesn't work any more.

model/riscv_fetch.sail
jrtc27: Huh. That's what it says. But *why*? What possible benefit is there for that?

jrtc27: I get that in the presence of >32-bit instructions this matters, but in a core that doesn't have such instructions, why would you ever fetch instruction bytes you know you're going to throw away?

ved-rivos: The code put in place here was to be extensible to the forthcoming 48b and 64b instructions.

ved-rivos: But I can update it to fetch 2B instead and it will be equivalent. When 48b and 64b instructions are added, this can be extended.

jrtc27: Yeah, there's more work to be done to generalise this to >32-bit instructions in the presence of CHERI, so deferring that (with a comment explaining why it's written this way) would be helpful.

model/riscv_fetch.sail
ved-rivos: I see I have an error. The read_insn_bytes should invoke the hook with params (vaddr, vaddr+num_bytes). Fixed that.

model/riscv_fetch.sail
ved-rivos: I have predicated the atomic fetch of 4B with Ziccif being enabled. With Ziccif disabled it would degenerate to the fetch 16-bits-at-a-time approach.

model/riscv_fetch.sail
jrtc27: In fact you're going to have issues generalising this to >32-bit instructions even without CHERI. Currently you can ignore the problem that a PMP imposes because the PMP's granularity is 4 byte size and alignment, and therefore given ILEN is 32-bit i.e. 4 bytes you cannot currently ever have a case where fetching these extra bytes (whether because you were already aligned and fetched a whole 4 bytes that contained a 2-byte instruction at the start, or because you were unaligned and fetched 2 + 4 bytes and threw away the final 2 bytes for a resulting 4-byte instruction) would introduce new PMP faulting behaviour. But once you have 48-bit instructions, or more, you can't just fetch things in bigger power-of-two blocks as a generalisation of this, because that might start to straddle PMP boundaries and introduce faults on those wider accesses that would not be there in the 2-byte-at-a-time fetch approach, and you don't want to be able to take instruction fetch faults for instruction bytes you aren't executing.

CHERI is just a case of the same problem due to supporting more fine-grained protection than 4 byte size and alignment, and therefore shows up even when ILEN is 32-bit. And at least for CHERI it's protecting on the virtual address using a capability you have to hand, so you can tell in the frontend when to cut off the fetching (or poison some of the bytes), but for the PMP you can't know until you've translated the address.

I have to say I'm a little concerned that there's no mention of the interaction between Ziccif and the PMP for ILEN > 32 in the Ziccif spec linked to in this PR. Has this been considered at all?

aswaterman: > I have to say I'm a little concerned that there's no mention of the interaction between Ziccif and the PMP for ILEN > 32 in the Ziccif spec linked to in this PR. Has this been considered at all?

Yes, it has been.  Ziccif is only concerned with atomicity and ordering, not protection.  The instruction-fetch granule and PMP granule remain conceptually separate, and protections do not apply to discarded bytes.  (But in practice it is likely that implementations will choose a PMP granule no smaller than the instruction-fetch granule.)

I'll add a note to the spec to this effect.

ved-rivos: Thanks.

jordancarlin:
@ved-rivos I believe I fixed the issue in the Lean backend you were running into in #1272. This should pass CI if you rebase.

ved-rivos:
> @ved-rivos I believe I fixed the issue in the Lean backend you were running into in #1272. This should pass CI if you rebase.

@jordancarlin Thanks. I rebased the PR.

ved-rivos:
@jordancarlin unfortunately CI still fails

```
✖ [62/101] Building LeanRV64D.RiscvSysRegs (5.7s)
trace: .> LEAN_PATH=/home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean /home/runner/.elan/toolchains/leanprover--lean4-nightly---nightly-2025-08-02/bin/lean --tstack=400000 /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/LeanRV64D/RiscvSysRegs.lean -o /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D/RiscvSysRegs.olean -i /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D/RiscvSysRegs.ilean -c /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D/RiscvSysRegs.c --setup /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D/RiscvSysRegs.setup.json --json
error: LeanRV64D/RiscvSysRegs.lean:696:58: failed to prove termination, possible solutions:
  - Use `have`-expressions to prove the remaining goals
  - Use `termination_by` to specify a different well-founded relation
  - Use `decreasing_by` to specify your own tactic for discharging this kind of goal
⊢ max (currentlyEnabled_measure Ext_Ziccif) 0 < currentlyEnabled_measure Ext_Ziccid
error: build failed
Some required targets logged failures:
- LeanRV64D.RiscvSysRegs
```

Should this be true?
`def sys_enable_experimental_extensions () = false
`

jordancarlin:
Lean is unable to track dependencies between extensions, so they need to be explicitly spelled out in the `currentlyEnabled_measure` function in `riscv_termination.sail`. I just pushed an update to this branch to add that for Ziccid.

ved-rivos:
> Lean is unable to track dependencies between extensions, so they need to be explicitly spelled out in the `currentlyEnabled_measure` function in `riscv_termination.sail`. I just pushed an update to this branch to add that for Ziccid.

Thank you. BTW, what is Lean?

jordancarlin:
Lean is one of the more actively used theorem prover backends that Sail is able to generate code for. See https://lean-lang.org. @ineol or @tobiasgrosser could probably tell you more.

README.md
```diff
@@ -126,10 +126,11 @@ For booting operating system images, see the information under the
 - Svinval extension for fine-grained address-translation cache invalidation, v1.0
 - Svrsw60t59b extension for PTE reserved-for-software bits 60-59, v1.0
 - Physical Memory Protection (PMP)
+- Ziccif Extension for Instruction-Fetch Atomicity, v1.0
```
jordancarlin: Move this up to be with the other Zi* extensions.

model/riscv_extensions.sail
```diff
@@ -116,6 +116,16 @@ function clause hartSupports(Ext_Zihpm) = config extensions.Zihpm.supported
...
+// Instruction/Data Coherence and Consistency
+enum clause extension = Ext_Ziccid
+mapping clause extensionName = Ext_Ziccid <-> "ziccid"
+function clause hartSupports(Ext_Ziccid) = config extensions.Ziccid.supported : bool & sys_enable_experimental_extensions()
+// Instruction-Fetch Atomicity
+enum clause extension = Ext_Ziccif
+mapping clause extensionName = Ext_Ziccif <-> "ziccif"
+function clause hartSupports(Ext_Ziccif) = config extensions.Ziccif.supported
+function clause currentlyEnabled(Ext_Ziccif) = hartSupports(Ext_Ziccif)
+function clause currentlyEnabled(Ext_Ziccid) = hartSupports(Ext_Ziccid) & currentlyEnabled(Ext_Ziccif)
```
jordancarlin, line 128: We should probably add a check in `riscv_validate_config.sail` to make sure someone doesn't try enabling Ziccid while Ziccif is disabled.

jordancarlin (changes requested):
A couple of minor nits, but otherwise LGTM.

README.md
ved-rivos: done

model/riscv_extensions.sail
ved-rivos: Added a function to check for instruction dependencies. I have included a couple as example and pulled in the newly added check for BF16 but this function needs to be filled out fully.

model/riscv_fetch.sail
```diff
@@ -11,11 +11,23 @@
...
-  // TODO: Add a parameter to try_step() to allow forcing the fetch result and use that instead.
+val read_insn_bytes : forall 'n, 'n in {2, 4} .  (xlenbits, int('n)) -> result(bits('n * 8), FetchResult)
+function read_insn_bytes(vaddr, num_bytes) = {
+  match ext_fetch_check_pc(vaddr, vaddr + num_bytes) {
+    Some(e) => Err(F_Ext_Error(e)),
+    None()  =>
+      match translateAddr(Virtaddr(vaddr), InstructionFetch()) {
+        Err(e, _)     => Err(F_Error(e, vaddr)),
+        Ok(paddr,  _) =>
+          match mem_read(InstructionFetch(), paddr, num_bytes, false, false, false) {
```
pmundkur, line 22: I think the right memory accessor to use here is `phys_mem_read()` instead of `mem_read()` since the latter does PMP checks.

model/riscv_fetch.sail
```diff
@@ -27,32 +39,40 @@ function fetch() -> FetchResult = {
...
+  // 3. Otherwise, atomically fetch M bytes from memory at address pc + T,
+  //    increment T by M, and go back to step 2.
+  if PC[1] == bitzero & currentlyEnabled(Ext_Ziccif) then {
+    match read_insn_bytes(PC, 4) {
+      Err(fe) => fe,
+      Ok(w32) => {
+        if isRVC(w32[15..0]) then
+          F_RVC(w32[15..0])
+        else
+          F_Base(w32)
```
pmundkur, line 59: According to the spec, "Memory protection is applied only to bytes that are not discarded by this algorithm." Shouldn't there be a phys_access_check() here on the relevant address range to implement the PMP checks before returning the FetchResult?

model/riscv_fetch.sail
ved-rivos: @pmundkur mem_read invokes mem_read_priv. mem_read_priv invokes mem_read_priv_meta. mem_read_priv_meta invokes checked_mem_read. checked_mem_read does phys_access_check, which invokes pmpCheck, and if successful invokes phys_mem_read. So I think mem_read() is the right function to invoke here.

model/riscv_fetch.sail
ved-rivos: @pmundkur In general the PMP grain is 2^(G+2). When G is 0, the minimal granularity is 4 bytes. The bits 1:0 of PMP address is always 0. This code is fine as a 4 byte aligned fetch will either be allowed or disallowed. When a future ISA extension adds ILEN > 32b then when implementing that we will need to deal with more cases.

pmundkur:
Rebased and slightly tweaked as #1630 for Ziccif.

PR #1247, opened 2025-09-01
Reset PC on reset, and improve tohost address handling

Timmmm:
1. Add a PC reset address that PC is reset to every time the chip is reset. Configure this to be the ELF entry point.
2. Switch the ELF entry point and HTIF tohost setting to be setters rather than callbacks. These are static values and setters are much easier to deal with.
3. Remove the unused `elf_tohost` and `elf_entry` functions.

### DIFF
--- a/c_emulator/riscv_platform.cpp
+++ b/c_emulator/riscv_platform.cpp
@@ -69,16 +69,6 @@ unit plat_term_write(mach_bits s)
   return UNIT;
 }
 
-bool plat_enable_htif(unit)
-{
-  return rv_enable_htif;
-}
-
-mach_bits plat_htif_tohost(unit)
-{
-  return rv_htif_tohost;
-}
-
 bool sys_enable_experimental_extensions(unit)
 {
   return rv_enable_experimental_extensions;
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -53,6 +53,9 @@ int rvfi_dii_port = 0;
 std::optional<rvfi_handler> rvfi;
 std::vector<std::string> elfs;
 
+// The address of the HTIF tohost port, if it is enabled.
+std::optional<uint64_t> htif_tohost_address;
+
 rvfi_callbacks rvfi_cbs;
 
 std::string sig_file;
@@ -250,10 +253,10 @@ uint64_t load_sail(const std::string &filename, bool main_file)
     const auto &tohost = symbols.find("tohost");
     if (tohost == symbols.end()) {
       fprintf(stderr, "Unable to locate tohost symbol; disabling HTIF.\n");
-      rv_enable_htif = false;
+      htif_tohost_address = std::nullopt;
     } else {
-      rv_htif_tohost = tohost->second;
-      fprintf(stdout, "HTIF located at 0x%0" PRIx64 "\n", rv_htif_tohost);
+      htif_tohost_address = tohost->second;
+      fprintf(stdout, "HTIF located at 0x%0" PRIx64 "\n", *htif_tohost_address);
     }
     // Locate test-signature locations if any.
     const auto &begin_sig = symbols.find("begin_signature");
@@ -291,8 +294,13 @@ void write_dtb_to_rom(const std::vector<uint8_t> &dtb)
 
 void init_sail(uint64_t elf_entry, const char *config_file)
 {
+  // zset_pc_reset_address must be called before zinit_model
+  // because reset happens inside init_model().
+  zset_pc_reset_address(elf_entry);
+  if (htif_tohost_address.has_value()) {
+    zenable_htif(*htif_tohost_address);
+  }
   zinit_model(config_file != nullptr ? config_file : "");
-  zforce_pc(elf_entry);
   zinit_boot_requirements(UNIT);
 }
 
--- a/handwritten_support/RiscvExtrasExecutable.lean
+++ b/handwritten_support/RiscvExtrasExecutable.lean
@@ -22,11 +22,6 @@ section defs
 variable [Arch]
 
 -- Platform definitions
-def elf_tohost (_:Unit) : Int := panic "TODO"
-def elf_entry (_:Unit) : Int := panic "TODO"
-def plat_enable_htif (_ : Unit) := false
-def plat_htif_tohost (_:Unit) : BitVec n := 0x80001000
-
 section Effectful
 
 variable {Register : Type} {RegisterType : Register → Type} [DecidableEq Register] [Hashable Register]
--- a/model/main.sail
+++ b/model/main.sail
@@ -9,26 +9,10 @@
 // When the symbolic execution is running a litmus test, it sets a
 // different entry point for each thread in the compiled litmus test.
 
-val get_entry_point : unit -> xlenbits
-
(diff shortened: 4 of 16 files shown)


model/riscv_sys_control.sail
```diff
@@ -295,6 +295,12 @@ function reset_misa() -> unit = {
...
 
+// Address to reset PC to on reset.
+register pc_reset_address : xlenbits = zeros()
+// This is called externally to set the PC reset address to the ELF entry point.
+// bits(64) is used because it's easier to deal with in C.
+function set_pc_reset_address(addr : bits(64)) -> unit = pc_reset_address = trunc(addr)
+
 // This function is called on reset, so it should only perform the reset actions
 // described in the "Reset" section of the privileged architecture specification.
 function reset_sys() -> unit = {
```
nadime15, line 306: Thats not part of this, PR but I have noticed that we dont reset the `hart_state` when `reset()` is called. Its set to `HART_ACTIVE` in `init_model()` (./riscv_model.sail:24) but as the comment above the function says, its not called for every chip reset.

Would it make sense to move `hart_state = HART_ACTIVE();` to reset()?

c_emulator/riscv_sim.cpp
```diff
@@ -53,6 +53,9 @@ int rvfi_dii_port = 0;
 std::optional<rvfi_handler> rvfi;
 std::vector<std::string> elfs;
 
+// The address if the HTIF tohost port, if it is enabled.
```
```suggestion
// The address of the HTIF tohost port, if it is enabled.
```

c_emulator/riscv_sim.cpp
```diff
@@ -291,8 +294,13 @@ void write_dtb_to_rom(const std::vector<uint8_t> &dtb)
 
 void init_sail(uint64_t elf_entry, const char *config_file)
 {
+  // zset_pc_reset_address must be called before zinit_model before
+  // zinit_model because reset happens inside init_model().
```
```suggestion
  // zset_pc_reset_address must be called before zinit_model
  // because reset happens inside init_model().
```

model/riscv_platform.sail
```diff
@@ -330,46 +320,58 @@ function reset_htif () -> unit = {
...
 
-val htif_store: forall 'n, 0 < 'n <= 8. (physaddr, int('n), bits(8 * 'n)) -> MemoryOpResult(bool)
+val htif_store : forall 'n, 0 < 'n <= max_mem_access . (physaddr, int('n), bits(8 * 'n)) -> MemoryOpResult(bool)
 function htif_store(Physaddr(paddr), width, data) = {
   if   get_config_print_platform()
   then print_log("htif[" ^ hex_bits_str(paddr) ^ "] <- " ^ bits_str(data));
-  /* Store the written value so that we can ack it later. */
-  if      width == 8
+
+  let base : physaddrbits = match htif_tohost_base {
```
arichardson, line 352: I guess this match could be moved to a helper function?

arichardson (approved):

model/riscv_platform.sail
```diff
@@ -14,17 +14,6 @@
...
-  interpreter: "Elf_loader.elf_tohost",
-  c: "elf_tohost"
-} :  unit -> int
-
-val elf_entry = pure {
-  interpreter: "Elf_loader.elf_entry",
-  c: "elf_entry"
-} : unit -> int
-
-
```
jordancarlin, line 27: These should be removed from all of the handwritten support files as well.

model/riscv_sys_control.sail
Timmmm: Yeah that makes sense. I don't think it matters too much at the moment because we don't actually have a way of calling `reset()` mid-test, but we will eventually when there's a proper library API.

Timmmm: Done.

model/riscv_platform.sail
Timmmm: I dunno, I only use it twice. Doesn't seem worth it?

model/riscv_platform.sail
Timmmm: I did I think? Did I miss some?

jordancarlin: > I did I think? Did I miss some?

Nope. Apparently I just missed some of the changed files.

jordancarlin (approved):

model/riscv_sys_control.sail
nadime15: Ok cool, thanks! Yeah we still work on the Debug Extension which allows you to reset the Chip and I was planning to use the `reset()` function

Timmmm: Ah nice, yeah that makes sense!

model/riscv_platform.sail
arichardson: Yeah I guess no need to change it.

model/riscv_model.sail
```diff
@@ -8,6 +8,9 @@
 
 // Chip reset. This only does the minimum resets required by the RISC-V spec.
 function reset() -> unit = {
+  // Come out of WFI/WRS.
+  hart_state = HART_ACTIVE();
+
```
pmundkur, line 13: The `hart_state` initialization can be removed from `init_model` below since it's now redundant.

c_emulator/riscv_platform.cpp
```diff
@@ -69,16 +69,6 @@ unit plat_term_write(mach_bits s)
...
-bool plat_enable_htif(unit)
-{
-  return rv_enable_htif;
-}
-
-mach_bits plat_htif_tohost(unit)
-{
-  return rv_htif_tohost;
-}
-
```
pmundkur, line 81: I think these also need to be removed from `riscv_platform.h`, `riscv_extras{_sequential}.lem`  and `riscv_extras.v`.

model/main.sail
```diff
@@ -9,24 +9,7 @@
...
-$else
-
-function get_entry_point() = zero_extend(0x1000)
-
-$endif
-
 function main() : unit -> unit = {
-  PC = get_entry_point();
-  print_bits("PC = ", PC);
-
```
pmundkur, line 29: It's best to keep the print of the PC for backwards compatibility; perhaps put it after the `init_model()`?

pmundkur (changes requested):

model/riscv_model.sail
Timmmm: Removed

c_emulator/riscv_platform.cpp
Timmmm: Oops, removed.

model/main.sail
Timmmm: Added back after init_model().

handwritten_support/RiscvExtrasExecutable.lean
```diff
@@ -22,11 +22,6 @@ section defs
 variable [Arch]
 
 -- Platform definitions
```
pmundkur, line 24: It looks like this comment can be removed.

Timmmm: I think it was referring to `plat_term_write`/`plat_term_read` too?

pmundkur: Ah, possible.

PR #1251, opened 2025-09-02
Add basic syscall support

trdthg:
Allow the emulator to handle system calls from guest applications Changes include:
- mem.cpp, mem.h: add read/write/clear functions for memory access, enabling the emulator to read from and write to guste memory space
- syscall.cpp, syscall.h: implemente the syscall_t class, which manages a table of syscall handler functions, and dispatchs syscall to host.
- riscv_syscall.h: interface for sail code


trdthg:
impl semihosting instead

PR #1254, opened 2025-09-03
Add missing sail_set_abstract_base_E_enabled()

Timmmm:
This was added in one place, but not here.


pmundkur (commented):
Now that there are several `set_abstract_` calls, it might be worth consolidating them into a single function, perhaps including `model_init()`.  This would help avoid issues like this in the future.

Timmmm:
I think instead I will just update the Sail compiler so these calls aren't necessary at all.

Timmmm:
Sail compiler PR: https://github.com/rems-project/sail/pull/1436

PR #1256, opened 2025-09-03
Fix macOS first-party test linking

### LINKED ISSUE #1255: Mac CI builds are broken
```
ld: unknown options: -Bstatic --no-relax -T 
```

I guess they updated things, and switched to Apple's new linker. Does anyone have a Mac and could figure out how to fix this?

jordancarlin:
Fixes #1255 and gets CI passing again.

### DIFF
--- a/test/first_party/CMakeLists.txt
+++ b/test/first_party/CMakeLists.txt
@@ -68,6 +68,12 @@ if (NOT CROSS_COMPILER_COMMAND_RV32 OR NOT CROSS_COMPILER_COMMAND_RV64)
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


test/first_party/CMakeLists.txt
```diff
@@ -68,6 +68,12 @@ if (NOT CROSS_COMPILER_COMMAND_RV32 OR NOT CROSS_COMPILER_COMMAND_RV64)
     endif()
 endif()
 
+# On macOS, use lld to avoid ld: unknown options: -Bstatic --no-relax -T
```
arichardson:
```suggestion
# On macOS, use lld instead of the system linker which does not support ELF
```
Maybe this?

PR #1228, opened 2025-08-21
Fix a missing `set_abstract` for `config base.E`.

pmundkur:
The symptom was `base.E` could be deleted from the config without a failure.

The uninitialized `abstract_zbase_E_enabled` in the generated C was `false` and hence this was not noticed earlier.

PR #1260, opened 2025-09-05
Consolidate the setting of abstract types in a single function.

pmundkur:
It is too easy to miss adding each one in multiple places. This should go away once the compiler can generate them as part of `model_init()`.


pmundkur:
For e.g., #1254 and #1228 both missed the unit test wrapper.

Timmmm (approved):

PR #1261, opened 2025-09-06
Use boon for build-time schema validation of generated configs.

pmundkur:
boon is built from source since boon binary releases are only available for Linux and MacOS, and the Linux binaries do not work on older Ubuntu.  Building from source however does introduce a build dependency on cargo and a Rust toolchain.

### DIFF
--- a/dependencies/boon/CMakeLists.txt
+++ b/dependencies/boon/CMakeLists.txt
@@ -1,28 +1,32 @@
-find_package(Cargo)
-if(NOT Cargo_FOUND)
-    set(try_rustup "installing rustup from https://rustup.rs")
-    if (APPLE)
-        message(FATAL_ERROR "Cargo not found. Try 'brew install rust' or ${try_rustup}.")
-    elseif (UNIX)
-        message(FATAL_ERROR "Cargo not found. Try 'sudo apt install cargo' or 'sudo dnf install cargo' or ${try_rustup}.")
-    else()
-        message(FATAL_ERROR "Cargo not found. Try ${try_rustup}.")
-    endif()
+option(ENABLE_BUILD_TIME_SCHEMA_VALIDATION "Enable schema validation of generated configurations using Boon." OFF)
+
+if (ENABLE_BUILD_TIME_SCHEMA_VALIDATION)
+  find_package(Cargo)
+  if(NOT Cargo_FOUND)
+      set(try_rustup "installing rustup from https://rustup.rs")
+      if (APPLE)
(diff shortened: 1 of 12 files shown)


pmundkur:
This should also help the boon dependency of #1189.

Timmmm:
If we're going to build from source I wonder if we should just go all the way and build it as a library and link with it? Then we don't need to worry about finding the binary or writing out temporary files.

Timmmm:
(As an interim solution still - I still think the best long term solutions is for Sail to generate C code to do the validation.)

pmundkur:
I think linking to boon should be out-of-scope for this PR, which is just doing build-time validation, but should be considered for #1189.

Given that the Lean executable simulator is getting more real, it would be good to harmonize the schema validation approach across both simulators if possible.   If it's possible to link boon into both simulators, then it might make sense to do it for the C++ simulator in #1189.

dependencies/boon/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+find_package(Cargo)
+if(NOT Cargo_FOUND)
```
arichardson: I am a bit concerned about making this cargo dependency mandatory for something that is only used at build time. Users that don't have rust installed should still be able to build the model and run it. Can we make this either 1) completely optional or 2) require a -DSCHEMA_VALIDATION=OFF flag?

pmundkur: #1189 will require boon to be installed for run-time validation.  I'd prefer run-time schema validation to not be optional.

pmundkur: I made this completely optional.

Using a library approach would require the emulator to be built and then executed to validate the generated configurations (including the default config). In case that turns out to be overly complicated, I updated this PR and made the boon dependency optional, and enabled it in CI.

Timmmm: You can do that in a couple of ways:

1. Add it as a CTest test. Definitely the easiest way, but it does require actually running the tests.
2. Add a custom target that depends on the emulator and runs it. It has to output a file (could just be stdout from the emulator). Something like this:

```
add_custom_command(
  DEPENDS sail_riscv_sim
  OUTPUTS validation_report.txt
  COMMAND sail_riscv_sim --config ... --validate-schema >validation_report.txt
)
add_custom_target(validate_report ALL DEPENDS validation_report.txt)
```

Timmmm: Implemented in #1297

pmundkur:
Closing in favor of #1297

PR #1262, opened 2025-09-07
Simplify read/write kinds

Timmmm:
Remove the `option()` return type on `read_kind_of_flags` and remove some `read_kind`/`write_kind` values because they are all unreachable.

`res`/`con` are only true for AMOs and lr/sc. In those cases we always condition `rl` on `aq` or vice versa. This means half the entries are unreachable.

### DIFF
--- a/model/prelude_mem.sail
+++ b/model/prelude_mem.sail
@@ -18,8 +18,6 @@ $include <concurrency_interface.sail>
 
 enum write_kind = {
   Write_plain,
-  Write_RISCV_release,
-  Write_RISCV_strong_release,
   Write_RISCV_conditional,
   Write_RISCV_conditional_release,
   Write_RISCV_conditional_strong_release,
@@ -28,8 +26,6 @@ enum write_kind = {
 enum read_kind = {
   Read_plain,
   Read_ifetch,
-  Read_RISCV_acquire,
-  Read_RISCV_strong_acquire,
   Read_RISCV_reserved,
   Read_RISCV_reserved_acquire,
   Read_RISCV_reserved_strong_acquire,
@@ -84,8 +80,6 @@ function write_ram(wk, Physaddr(addr), width, data, meta) = {
   let request : Mem_write_request('n, 64, physaddrbits, unit, RISCV_strong_access) = struct {
     access_kind = match wk {
       Write_plain => AK_explicit(struct { variety = AV_plain, strength = AS_normal }),
-      Write_RISCV_release => AK_explicit(struct { variety = AV_plain, strength = AS_rel_or_acq }),
-      Write_RISCV_strong_release => AK_arch(struct { variety = AV_plain }),
       Write_RISCV_conditional => AK_explicit(struct { variety = AV_exclusive, strength = AS_normal }),
       Write_RISCV_conditional_release => AK_explicit(struct { variety = AV_exclusive, strength = AS_rel_or_acq }),
       Write_RISCV_conditional_strong_release => AK_arch(struct { variety = AV_exclusive }),
@@ -126,8 +120,6 @@ function read_ram(rk, Physaddr(addr), width, read_meta) = {
     access_kind = match rk {
       Read_plain => AK_explicit(struct { variety = AV_plain, strength = AS_normal }),
       Read_ifetch => AK_ifetch(),
-      Read_RISCV_acquire => AK_explicit(struct { variety = AV_plain, strength = AS_rel_or_acq }),
-      Read_RISCV_strong_acquire => AK_arch(struct { variety = AV_plain }),
       Read_RISCV_reserved => AK_explicit(struct { variety = AV_exclusive, strength = AS_normal }),
       Read_RISCV_reserved_acquire => AK_explicit(struct { variety = AV_exclusive, strength = AS_rel_or_acq }),
       Read_RISCV_reserved_strong_acquire => AK_arch(struct { variety = AV_exclusive }),
--- a/model/riscv_mem.sail
+++ b/model/riscv_mem.sail
@@ -55,44 +55,31 @@ function is_aligned_bits(vaddr : xlenbits, width : word_width) -> bool =
 
 overload is_aligned_addr = {is_aligned_paddr, is_aligned_vaddr, is_aligned_bits}
 
-function read_kind_of_flags (aq : bool, rl : bool, res : bool) -> option(read_kind) =
+function read_kind_of_flags (aq : bool, rl : bool, res : bool) -> read_kind =
   match (aq, rl, res) {
-    (false, false, false) => Some(Read_plain),
-    (true, false, false)  => Some(Read_RISCV_acquire),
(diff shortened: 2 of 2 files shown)


Timmmm:
In #792 I am going to move these flags into `AccessType` and change them into an enum so it can be simplified even further.

model/core/phys_mem_interface.sail
```diff
@@ -28,8 +26,6 @@ enum write_kind = {
 enum read_kind = {
   Read_plain,
   Read_ifetch,
-  Read_RISCV_acquire,
```
arichardson, line 30: Aren't these entries useful for https://github.com/riscv/riscv-zalasr?

Timmmm: Ah yeah true. Maybe we should just add them back when that is implemented though.

Also I'm not a memory ordering expert but now I think about it, aren't the `_conditional` and `_reserved` enums mis-named? They are also used for AMOs and those aren't really conditional/reserved (unless you consider an AMO to be like an LR/SC that have been merged so they always succeed... ?).

pmundkur:
@Alasdair could you review this?

Timmmm:
I'll merge this in a couple of days if nobody objects. I don't think there's really anything controversial here.

model/core/phys_mem_interface.sail
```diff
@@ -17,8 +17,6 @@ $include <concurrency_interface.sail>
 
 enum write_kind = {
   Write_plain,
-  Write_RISCV_release,
-  Write_RISCV_strong_release,
```
pmundkur, line 21: This might break the memory model in ISLA.
https://github.com/rems-project/isla/blob/dcece5f1862b961a8b0f205db9d6a0b461bc4335/configs/riscv32.toml#L57
https://github.com/rems-project/isla/blob/dcece5f1862b961a8b0f205db9d6a0b461bc4335/configs/riscv64_ubuntu.toml#L57

Timmmm: Ok I'll just leave them in for now with a note.

Timmmm:
@pmundkur are you ok with this merging as it is now with the enums added back in?

model/sys/mem.sail
```diff
@@ -46,45 +46,32 @@ function is_aligned_vaddr(Virtaddr(addr) : virtaddr, width : nat1) -> bool =
...
-    (true, false, false)  => Some(Read_RISCV_acquire),
-    (true, true, false)   => Some(Read_RISCV_strong_acquire),
-    (false, false, true)  => Some(Read_RISCV_reserved),
-    (true, false, true)   => Some(Read_RISCV_reserved_acquire),
-    (true, true, true)    => Some(Read_RISCV_reserved_strong_acquire),
-    (false, true, false)  => None(), // should these be instead throwing error_not_implemented as below?
-    (false, true, true)   => None()
+    (false, false, false) => Read_plain,
+    (true, false, false)  => internal_error(__FILE__, __LINE__, "Load with acquire semantics should be unreachable"),
+    (true, true, false)   => internal_error(__FILE__, __LINE__, "Load with acquire-release semantics should be unreachable"),
```
pmundkur: Perhaps specify unreserved, since the reserved cases are reachable?
```suggestion
    (true, false, false)  => internal_error(__FILE__, __LINE__, "Unreserved load with acquire semantics should be unreachable"),
    (true, true, false)   => internal_error(__FILE__, __LINE__, "Unreserved load with acquire-release semantics should be unreachable"),
```

model/sys/mem.sail
```diff
@@ -46,45 +46,32 @@ function is_aligned_vaddr(Virtaddr(addr) : virtaddr, width : nat1) -> bool =
...
+    // This is never called with rl=true and aq=false.
+    (false, true, false)  => internal_error(__FILE__, __LINE__, "Load with release semantics should be unreachable"),
+    (false, true, true)   => internal_error(__FILE__, __LINE__, "Load-reserved with release semantics should be unreachable"),
   }
 
 function write_kind_of_flags (aq : bool, rl : bool, con : bool) -> write_kind =
   match (aq, rl, con) {
     (false, false, false) => Write_plain,
-    (false, true,  false) => Write_RISCV_release,
+    (false, true,  false) => internal_error(__FILE__, __LINE__, "Store with release semantics should be unreachable"),
```
pmundkur: Perhaps specify unconditional?
```suggestion
    (false, true,  false) => internal_error(__FILE__, __LINE__, "Unconditional store with release semantics should be unreachable"),
```

model/sys/mem.sail
```diff
@@ -46,45 +46,32 @@ function is_aligned_vaddr(Virtaddr(addr) : virtaddr, width : nat1) -> bool =
...
     (false, false, true)  => Write_RISCV_conditional,
     (false, true , true)  => Write_RISCV_conditional_release,
-    (true,  true,  false) => Write_RISCV_strong_release,
+    (true,  true,  false) => internal_error(__FILE__, __LINE__, "Store with aquire-release semantics should be unreachable"),
```
pmundkur: Fix typo, and specify unconditional, since the conditional case is reachable?
```suggestion
    (true,  true,  false) => internal_error(__FILE__, __LINE__, "Unconditional store with acquire-release semantics should be unreachable"),
```

jordancarlin (approved):

jordancarlin:
Seems like this broke the Lean build in a different way from how it’s broken on master:

> trace: .> LEAN_PATH=/home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean /home/runner/.elan/toolchains/leanprover--lean4-nightly---nightly-2025-08-02/bin/lean --tstack=400000 /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/LeanRV64D/PhysMemInterface.lean -o /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D/PhysMemInterface.olean -i /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/lib/lean/LeanRV64D/PhysMemInterface.ilean -c /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D/PhysMemInterface.c --setup /home/runner/work/sail-riscv/sail-riscv/build/model/Lean_RV64D/.lake/build/ir/LeanRV64D/PhysMemInterface.setup.json --json
  error: LeanRV64D/PhysMemInterface.lean:284:15: cannot lift `(<- ...)` over a binder, this error usually happens when you are trying to lift a method nested in a `fun`, `let`, or `match`-alternative, and it can often be fixed by adding a missing `do`
  error: LeanRV64D/PhysMemInterface.lean:331:15: cannot lift `(<- ...)` over a binder, this error usually happens when you are trying to lift a method nested in a `fun`, `let`, or `match`-alternative, and it can often be fixed by adding a missing `do`


If you can find an easy fix for it that’d be great, but I think we can merge either way since Lean is broken on master anyway.

PR #1263, opened 2025-09-08
Move print_log_instr to new fetch_callback

trdthg:
This add a new callback type fetch_callback, which will be called after fetch done. I set a default empty implementation for fetch_callback, so we don't have to add an empty implementation for rvfi_callback as well.
This also move the function print_log_instr to fetch_callback.

But this is just a propose now, because there are still many ugly places in the code.

1. The c implemation of fetch_callback is much complex then sail version. The interaction between sail and C is not very convenient.
2. It decoded twice, both in sail(before execute) and c(fetch_callback). 
	One solution maybe is passing the decoded `zinstruction` value to c, but this will cause issue about file dependency
    `fetch_callback(riscv_core) -> zinstruction(riscv_postlude) -> (riscv_core)`

### DIFF
--- a/c_emulator/riscv_callbacks_log.cpp
+++ b/c_emulator/riscv_callbacks_log.cpp
@@ -107,3 +107,24 @@ void log_callbacks::vreg_write_callback(unsigned reg, lbits value)
 void log_callbacks::pc_write_callback(sbits) { }
 
 void log_callbacks::trap_callback() { }
+
+#define CREATEX(T, val)                                                        \
+  T val;                                                                       \
+  CREATE(T)(&val);
+
+void log_callbacks::fetch_callback(sail_int step_no, uint64_t code, bool rvc)
+{
+  CREATEX(sail_string, cur_privilege_str);
+  zprivLevel_to_str(&cur_privilege_str, zcur_privilege);
+  CREATEX(sail_string, insn_str);
+  char code_str[11] = {};
+  if (rvc) {
+    zassembly_half(&insn_str, (uint16_t)code);
+    sprintf(code_str, "0x%" PRIX16, (uint16_t)code);
+  } else {
+    zassembly_word(&insn_str, (uint32_t)code);
+    sprintf(code_str, "0x%" PRIX32, (uint32_t)code);
+  }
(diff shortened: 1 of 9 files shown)


model/riscv_step.sail
```diff
@@ -116,10 +113,7 @@ function run_hart_active(step_no: nat) -> Step = {
         sail_instr_announce(w);
         let instbits : instbits = zero_extend(w);
         let instruction = ext_decode(w);
-        if   get_config_print_instr()
-        then {
-          print_log_instr("[" ^ dec_str(step_no) ^ "] [" ^ to_str(cur_privilege) ^ "]: " ^ bits_str(PC) ^ " (" ^ bits_str(w) ^ ") " ^ to_str(instruction), zero_extend(PC));
-        };
+        fetch_callback(step_no, zero_extend(w), false);
```
nadime15, line 116: By removing the `get_config_print_instr()` check, it now prints regardless of whether the command-line argument was provided?.

c_emulator/riscv_callbacks_log.cpp
```diff
@@ -107,3 +107,24 @@ void log_callbacks::vreg_write_callback(unsigned reg, lbits value)
 void log_callbacks::pc_write_callback(sbits) { }
 
 void log_callbacks::trap_callback() { }
+
+#define CREATEX(T, val)                                                        \
```
Timmmm, line 111: This macro doesn't seem worth it for 2 uses. Macros are massive foot-guns so we should avoid them if possible.

model/riscv_callbacks.sail
```diff
@@ -7,6 +7,9 @@
 /*=======================================================================================*/
 
 // Callbacks for state-changing events
+val fetch_callback = pure {c: "fetch_callback"} : (nat, bits(32), bool) -> unit
```
Timmmm, line 10: I think if we use `forall 'n, 'n in {16, 32}. ... bits('n)` then it will translate to `sbits` in C and we won't need the `bool`.

Timmmm (changes requested):
Doesn't this lose the symbolization, and leave `print_log_instr` unused?

PR #1264, opened 2025-09-08
Update the virtual memory documentation.

pmundkur:
Convert the vmem notes to Markdown like the other docs, and bring the diagram up-to-date.  Also update some comments in the vmem code.


model/riscv_vmem_tlb.sail
```diff
@@ -79,10 +79,10 @@ function tlb_hash forall 'v, is_sv_mode('v) . (
 ) -> tlb_index_range =
   unsigned(vpn[5 .. 0])
 
-// PUBLIC: invoked in init_vmem() [riscv_vmem.sail]
+// PUBLIC: invoked in reset_vmem() [riscv_vmem.sail]
```
Timmmm, line 82: That reminds me, Sail supports a `Private` annotation now. We should probably start organising the files into directories that match the module structure, and use `Private`.

Issue #1266, opened 2025-09-09
Reorganise file structure into modules


Timmmm:
Currently we have a massive list of `.sail` files, most of which redundantly start with `riscv_`. Now that we use `riscv.sail_project` to make a module structure we should probably organise the actual files into directories that match the module structure, and also remove the `riscv_` prefixes while we're at it. A lot of the module names also have `riscv_` that I think we could drop, e.g. `riscv_exceptions`.

We probably only need to go one level deep, i.e. `model/extensions/I/base_types.sail` rather than `model/extensions/I/I_types/base_types.sail`.

PR #1267, opened 2025-09-10
Reorganise file structure into modules

KotorinMinami:
for #1266


pmundkur (changes requested):
Thanks for taking a crack at this!  The `AddingExtensions.md::"Naming the new files"` needs to be updated, as well as ` doc/figs/notes_Virtual_Memory_Fig1.svg`.

model/riscv.sail_project
```diff
@@ -2,68 +2,68 @@ variable TERMINATION_FILE = false
...
+    core/sys_regs.sail,
+    core/ext_regs.sail,
+    core/addr_checks_common.sail,
+    core/addr_checks.sail,
+    core/misa_ext.sail,
+    core/softfloat_interface.sail,
 }
 
 riscv_exceptions {
   requires riscv_core
```
jordancarlin: How about also renaming these modules to drop the `riscv_` prefix. Applies to `riscv_exceptions`, `riscv_core`, `riscv_termination`, and `riscv_postlude`.

arichardson:
For a lot of extensions, the file name duplicates the folder name. Would it make sense to just use something like `Zicond/instructions.sail` and `Zicond/types.sail` instead of `model/extensions/Zicond/{insts_zicond,zicond_types}.sail`?

Timmmm:
On the other hand it might cause more confusion if we have a Sail module structure that differs from the filesystem structure.

I think I would say we should stick to filesystem=modules where possible, and perhaps in future make the modules more fine grained so we don't have so many things confusingly lumped together in `core` and `sys`.

pmundkur (commented):
It would be good to merge this soon to avoid having to continually fix conflicts. In today's meeting, the consensus was to have the filesystem structure match the module structure as much as possible, with other fine-tuning (e.g. splitting the prelude out) left for later.   But we should at least also rename `doc/figs/notes_Virtual_Memory_Fig1.svg` as suggested.

pmundkur (approved):

pmundkur:
This has conflicts again.  Let's merge this as soon as this is rebased.

model/riscv.sail_project
```diff
@@ -180,203 +180,203 @@ extensions {
...
   }
 
   Zihintpause {
     requires riscv
     before I_insts // PAUSE overrides a FENCE
-    files riscv_insts_zihintpause.sail
+    files extensions/Zihintpause/zihintpause_insts.sail
   }
 
   bfloat16 {
```
nadime15, line 382: This needs to be updated now that Zfbfmin has been merged.

nadime15: Ups looks like the line references did not work, I meant this part:

```
bfloat16 {
 bfloat16_core {
      requires riscv, FD
      files riscv_insts_zfbfmin_utils.sail
    }
    Zfbfmin {
      requires riscv, FD, bfloat16_core
      files riscv_insts_zfbfmin.sail
    }
}
```

pmundkur:
Rebased.

nadime15 (approved):

PR #1271, opened 2025-09-11
Add html doc bundle to build and release.

### LINKED ISSUE #1265: Export HTML as part of release
Currently for the release flow we generate the JSON doc bundle for use in the ISA manual. Sail can also generate [fancy HTML](https://alasdair.github.io/sail-riscv/riscv_insts_base.html) using the `-html` backend, and then the Sail snippets auto-link to it. We should add a CMake target to do that and to zip it all up, and add it into the release. Maybe we should put the JSON in the same zip.

pmundkur:
This bundles the Sail unit tests as well, not sure it's worth excluding those.

The generated html and json are separate for now; it might be more convenient that way.

Fixes #1265.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -68,7 +68,7 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      - name: Build simulators and JSON
+      - name: Build simulators, JSON and HTML
         run: |
           git config --global --add safe.directory '*'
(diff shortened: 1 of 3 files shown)


Timmmm:
Nice! LGTM once the mkdir issue is fixed somehow.

jordancarlin (approved):

PR #1274, opened 2025-09-12
Use C++ Sail output

Timmmm:
This switches from using Sail's C output to using C++. The output code gets wrapped in a `class`, which means we can create more than one instance of it (e.g. for multicore).

The `Model` class is currently *not* thread safe due to the use of temporary globals in `sail.c`. So although you can simulate multicore systems, you have to execute them one at a time.

The handling of platform callbacks is a little convoluted but I think this is a quite flexible solution. Sail is instructed to derive the `Model` class from `PlatformImpl` which has virtual methods for all of the platform callbacks with default nop implementations.

This means the default `Model` can be constructed and has a "nop" platform. Then we add our implementation of the platform with `ModelImpl` which overrides those methods.

`ModelImpl` also allows registering callback receivers.

This is not 100% perfect yet. We still have a global `g_model`, which is not ideal.

### DIFF
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -200,8 +200,7 @@ bool ModelImpl::valid_reservation(unit)
 
 unit ModelImpl::plat_term_write(mach_bits s)
 {
-  char c = s & 0xff;
-  plat_term_write_impl(c);
+  plat_term_write_impl(static_cast<char>(s));
   return UNIT;
 }
 


model/core/prelude.sail
```diff
@@ -34,17 +34,17 @@ overload operator | = {or_vec}
 
 overload operator ^ = {xor_vec, concat_str}
 
-val sub_vec = pure {c: "sub_bits", lean: "_lean_sub", _: "sub_vec"} : forall 'n. (bits('n), bits('n)) -> bits('n)
+val sub_vec = pure {cpp: "sub_bits", lean: "_lean_sub", _: "sub_vec"} : forall 'n. (bits('n), bits('n)) -> bits('n)
```
arichardson: I wonder if people still want to generate C code for the model for downstream uses? Should we keep both, or have sail default the c one to the cpp if not defined? Or the other way around?

Timmmm: Oh yeah I think we may as well keep both when they are both available - at least all the built-in Sail functions. Not sure about the other ones...

arichardson: Agreed, probably only needed for the builtins.

arichardson: Also is there any reason this is not part of sail itself?

Timmmm: Yeah I dunno, I feel like there's quite a bit in risc-v's prelude that should probably be in Sail's prelude.

Alasdair: It's currently set up so the first definition will take precedence, and subsequent definitions are a warning so things can safely be moved into the Sail prelude without breaking anything.

nadime15:
This is probably unrelated for now, but enabling multi-hart simulation would affect the Debug Module we are working on. 
At the moment, we assume only a single hart is available, but in theory the Debug Module can manage a large number of harts.

So far, all of our code has been written in Sail (since it makes handling the many registers and bitfields easier), but I think parts of the Debug Module would need to be moved back to C++ to properly support 1+N harts.

Timmmm:
That's a good point. I was sort of vaguely aware of that. I guess in theory you could still write it in Sail as long as it was a completely separate Sail model, and then you'd need to connect them somehow.

nadime15:
> I guess in theory you could still write it in Sail as long as it was a completely separate Sail model, and then you'd need to connect them somehow.

Yeah, that would be perfect if this was possible! We can revisit this topic once this PR is merged, along with the debug stuff.

c_emulator/riscv_callbacks_log.cpp
```diff
@@ -31,79 +31,98 @@ log_callbacks::log_callbacks(bool config_print_reg,
...
+              static_cast<int>(value.len / 4), value.bits);
     }
   }
 }
 
-void log_callbacks::freg_write_callback(unsigned reg, sbits value)
+void log_callbacks::freg_write_callback(model::Model &model, unsigned reg,
+                                        sbits value)
 {
+  (void)model;
```
arichardson:
```suggestion
void log_callbacks::freg_write_callback(model::Model &, unsigned reg,
                                        sbits value)
{
```

Could do that instead? Or add [[maybe_unused]] to the argument?

Timmmm: Done

model/core/prelude.sail
Timmmm: I added `c:` for the built-in functions.

c_emulator/riscv_sim.cpp
```diff
@@ -82,18 +80,23 @@ uint64_t insn_limit = 0;
 char *sailcov_file = nullptr;
 #endif
 
+// Single global model instance.
+// TODO: This shouldn't be global, but it is due facilitate gradual transition
```
jordancarlin:
```suggestion
// TODO: This shouldn't be global, but it is to facilitate gradual transition
```

model/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ separate_arguments(sail_srcs UNIX_COMMAND "${sail_list_files}")
 
 # Generate C code from Sail model
```
jordancarlin:
```suggestion
# Generate C++ code from Sail model
```

c_emulator/riscv_sail.h
jordancarlin, line 1: Any reason not to drop this file now and directly include `sail_riscv_model.h` where this file used to be included?

model/CMakeLists.txt
jordancarlin:
```suggestion
############# C++ #############
```

jordancarlin (changes requested):
Some initial comments. Will review in more detail later.

I also think it might be worth renaming the `c_emulator` directory to `cpp_emulator`.

arichardson:
> Some initial comments. Will review in more detail later.
> 
> I also think it might be worth renaming the `c_emulator` directory to `cpp_emulator`.

Or just `emulator` since the ocaml one no longe exists?

c_emulator/riscv_sail.h
Timmmm: Yeah it was sort of a selfish reason though - Codasip is using a modified version of the Sail model for microarchitectural lockstep verification (or "macro-ops" instead of instructions), and having every `#include "sail_riscv_model.h"` go through `riscv_sail.h` made it really easy to switch headers with a single `#ifdef` in `riscv_sail.h`, i.e. I modified `riscv_sail.h` to be like

```
#ifdef MOP_MODEL
#include "sail_mop_model.h"
#else
#include "sail_riscv_model.h"
#endif
```

Buuutt... I don't work there any more, and I guess it's not too hard to just modify all the places that `#incldue "sail_riscv_model.h"` to have that ifdef instead.

Timmmm:
> Or just emulator since the ocaml one no longe exists?

I think I originally did that when the OCaml emulator was removed but people wanted to keep the `c_`. I can't remember why exactly. Maybe in case we add the OCaml one back one day? Or a Rust one? :-D

Timmmm:
Rebased. That was fairly painful so I'm afraid I had to remove the reservation printing option. I don't think it's too bad since it didn't really exist until very recently. We can add it back later.

It would be good to get this merged fairly soon to avoid more painful rebases!

c_emulator/riscv_sail.h
Timmmm: Done.

c_emulator/riscv_callbacks_if.h
```diff
@@ -2,64 +2,80 @@
...
+}
 
 class callbacks_if {
 public:
   virtual ~callbacks_if() = default;
-  // TODO: finding ways to improve the format
-  virtual void fetch_callback([[maybe_unused]] sbits opcode) { }
-  virtual void mem_write_callback([[maybe_unused]] const char *type,
+
+  virtual void fetch_callback([[maybe_unused]] model::Model &model,
```
arichardson: Is the namespace and class hardcoded in sail? I'd find riscv::Model slightly more readable. But no need to change this just an observation.

c_emulator/riscv_model_impl.h
```diff
@@ -0,0 +1,93 @@
...
+
+#include "sail.h"
+#include "sail_riscv_model.h"
+
+extern FILE *trace_log;
+extern int term_fd;
+void plat_term_write_impl(char c);
+
+// Model wrapped with an implementation of its platform callbacks.
+class ModelImpl : public model::Model {
```
arichardson: Might be worth marking this class as final so that the compiler can turn all the virtual function calls into a direct call. We can probably also change the way the sail emits C++ output, but using final should hopefully be equivalent when building with LTO.

c_emulator/riscv_callbacks_if.h
Timmmm: No there is a CLI option to change it. I wonder if it should actually be `hart::Model` because in future we might also end up with `plic::Model`, `imsic::Model`, `debug_module::Model` etc. if we decide to go that route...

Timmmm: Done

c_emulator/riscv_model_impl.h
Timmmm: Done.

test/unit_tests/main_unit_tests.cpp
```diff
@@ -6,7 +6,7 @@
 // Generated in the model C code. This is a simple test runner that just
```
jordancarlin, line 6:
```suggestion
// Generated in the model C++ code. This is a simple test runner that just
```

doc/ReadingGuide.md
```diff
@@ -278,16 +278,7 @@ platform model with any ISA implementation choices if specified,
...
-by the Sail compiler and contains calls to the platform interface
-[riscv_platform.h](../c_emulator/riscv_platform.h) for
-platform-specific information and to the callback interface
-[riscv_callbacks.h](../c_emulator/riscv_callbacks.h) for state-change
-events and logging. These provide the definitions declared as
-externally specified in the Sail files
-[platform.sail](../model/sys/platform.sail),
-[sys_reservation.sail](../model/sys/sys_reservation.sail) and
-[callbacks.sail](../model/core/callbacks.sail).
+TODO: Update description.
```
jordancarlin: A basic version of the new structure should be included before merging.

c_emulator/riscv_platform_if.h
```diff
@@ -0,0 +1,191 @@
...
+class callbacks_if;
+
+namespace model {
+class Model;
+}
+
+// The Model class derives from this one so when Sail calls C callback
+// functions it actually calls methods of this class. However they are
+// virtual functions so they actually call the Platform implementations
+// (see riscv_platform_impl.h). It's done this way because:
```
jordancarlin: `riscv_model_impl.h`? There is no `riscv_platform_impl.h`. And if so, should we rename this file to `riscv_model_if.h` (or vice versa)?

jordancarlin:
Oops. Seems like I hit the wrong button and triggered a review from Copilot. Feel free to disregard.

c_emulator/riscv_platform_if.h
```diff
@@ -0,0 +1,191 @@
...
+  virtual unit trap_callback([[maybe_unused]] bool is_interrupt,
+                             [[maybe_unused]] fbits cause)
+  {
+    return UNIT;
+  }
+
+  // Provides entropy for the scalar cryptography extension.
+  virtual mach_bits plat_get_16_random_bits(unit)
+  {
+    return UNIT;
```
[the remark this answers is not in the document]
Timmmm: Ooo technically doesn't matter but this is correct. Fixed.

c_emulator/riscv_platform_if.h
```diff
@@ -0,0 +1,191 @@
+#pragma once
+
+#include "sail.h"
+
+class callbacks_if;
+
+namespace model {
```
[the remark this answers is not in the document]
Timmmm: Oops this declaration was unused. Removed.

Timmmm:
It caught a couple of very minor issues at least...

c_emulator/riscv_platform_if.h
Timmmm: Oops yeah should have been `riscv_model_impl.h`. I think `platform` is still an ok name though because this is an interface to the platform for the model... if that makes sense?

doc/ReadingGuide.md
Timmmm: Done

c_emulator/riscv_model_impl.cpp
```diff
@@ -0,0 +1,279 @@
...
+{
+  return m_reservation_valid;
+}
+
+unit ModelImpl::plat_term_write(mach_bits s)
+{
+  char c = s & 0xff;
+  plat_term_write_impl(c);
+  return UNIT;
+}
```
nadime15, line 205: Does it make sense to add a comment here and to describe what is happening? I had to pipe this into chatgpt to understand it, especially the bit manipulation was a not obvious with the magic number.
```suggestion
unit ModelImpl::plat_term_write(mach_bits s)
{
  plat_term_write_impl(static_cast<char>(s));
  return UNIT;
}
```

Timmmm: Yeah I think `static_cast<>` is better and more obvious. Probably don't need a comment with that.

c_emulator/riscv_platform_if.h
jordancarlin: Fine either way, but I think it would be easier to follow the logic if the interface and implementation names match.

c_emulator/riscv_platform_if.h
```diff
@@ -0,0 +1,187 @@
...
+  virtual unit fetch_callback([[maybe_unused]] sbits opcode)
+  {
+    return UNIT;
+  }
+  virtual unit mem_write_callback([[maybe_unused]] const char *type,
+                                  [[maybe_unused]] sbits paddr,
+                                  [[maybe_unused]] uint64_t width,
+                                  [[maybe_unused]] lbits value)
+  {
+    return UNIT;
```
arichardson, line 30: I wonder if we should just move the callback registration and "loop over all registered callbacks" logic directly into this class? Then we don't need the it to be virtual and we can remove the [[maybe_unused]].

Alternatively, make them pure virtual, then we can also drop the [[maybe_unused]]?

Timmmm: I didn't do it like that because at this level you don't have access to `hart::Model`, so you can't do things like checking what xlen is in your callback by reading `hart::Model::zxlen` (we probably shouldn't do that, but you get the idea).

Timmmm:
Ok I'll give it a good few days, seeing as this is a pretty huge change, and then merge it.

c_emulator/riscv_platform_if.h
arichardson: Thanks that makes sense. Should the callback functions be pure virtual though?

Timmmm: That doesn't quite work because the generated unit testing function instantiates `hart::Model` to run the unit tests.

I guess maybe they should throw exceptions or something so you don't get confused why e.g. reservations don't work in unit tests...

(And I guess ideally we'd pass the model instance into the unit test function, but it doesn't work like that currently.)

Timmmm: I made https://github.com/rems-project/sail/issues/1556 so we don't forget... hopefully.

Arielfoever:
> Oops. Seems like I hit the wrong button and triggered a review from Copilot. Feel free to disregard.

little bit off topic.

Just wondering where is the button? Want to try in #1370

jordancarlin:
> > Oops. Seems like I hit the wrong button and triggered a review from Copilot. Feel free to disregard.
> 
> little bit off topic.
> 
> Just wondering where is the button? Want to try in #1370

In the top right where you can request reviews from anyone else.

PR #1275, opened 2025-09-12
Return softfloat flags and result as a tuple

Timmmm:
Since we now use a proper header for the generated model code, we can #include it in `riscv_softfloat.h` and use the `(fflags, bits(..))` types to directly return the tuples that the rest of the Sail code wants instead of using the super hacky method of poking some registers inside the model.

This is much much cleaner, makes the softfloat functions pure, and removes their dependency on the model *implementation*. Now it only depends on the model types (and that is really only because C doesn't have native tuple types; with fully native C++ codegen we could theoretically have them return `std::tuple<...>`).

I also switched `riscv_softfloat.c` to C++ which has slightly nicer syntax for returning values (and is more consistent with everything else being C++).


Timmmm:
Note I'm pretty sure the `= (0, 0)` in the Lem code is wrong but I have no idea what the Lem syntax is to make a zero bitvector.

Also this PR is a pre-requisite for #1274.

model/riscv_softfloat_interface.sail
```diff
@@ -37,488 +37,89 @@ type bits_WU     = bits(32)   /* Unsigned integer */
...
+val riscv_f64ToUi64 = pure {c: "softfloat_f64toui64", lem: "softfloat_f64_to_ui64"} : (bits_rm, bits_D) -> (bits_fflags, bits_LU)
+val riscv_i64ToF64 = pure {c: "softfloat_i64tof64", lem: "softfloat_i64_to_f64"} : (bits_rm, bits_L) -> (bits_fflags, bits_D)
+val riscv_ui64ToF64 = pure {c: "softfloat_ui64tof64", lem: "softfloat_ui64_to_f64"} : (bits_rm, bits_LU) -> (bits_fflags, bits_D)
+val riscv_f16ToF32 = pure {c: "softfloat_f16tof32", lem: "softfloat_f16_to_f32"} : (bits_rm, bits_H) -> (bits_fflags, bits_S)
+val riscv_f16ToF64 = pure {c: "softfloat_f16tof64", lem: "softfloat_f16_to_f64"} : (bits_rm, bits_H) -> (bits_fflags, bits_D)
+val riscv_f32ToF64 = pure {c: "softfloat_f32tof64", lem: "softfloat_f32_to_f64"} : (bits_rm, bits_S) -> (bits_fflags, bits_D)
+val riscv_f32ToF16 = pure {c: "softfloat_f32tof16", lem: "softfloat_f32_to_f16"} : (bits_rm, bits_S) -> (bits_fflags, bits_H)
+val riscv_f64ToF16 = pure {c: "softfloat_f64tof16", lem: "softfloat_f64_to_f16"} : (bits_rm, bits_D) -> (bits_fflags, bits_H)
+val riscv_f64ToF32 = pure {c: "softfloat_f64tof32", lem: "softfloat_f64_to_f32"} : (bits_rm, bits_D) -> (bits_fflags, bits_S)
```
nadime15: Thats a nice change! 

```suggestion
val riscv_f64ToF32 = pure {cpp: "softfloat_f64tof32", lem: "softfloat_f64_to_f32"} : (bits_rm, bits_D) -> (bits_fflags, bits_S)
```
Is it necessary to switch from c to cpp here?

model/riscv_softfloat_interface.sail
```diff
@@ -37,488 +37,89 @@ type bits_WU     = bits(32)   /* Unsigned integer */
...
-  (float_fflags[4 .. 0], float_result[31 .. 0])
-}
+val riscv_f16ToI32 = pure {c: "softfloat_f16toi32", lem: "softfloat_f16_to_i32"} : (bits_rm, bits_H) -> (bits_fflags, bits_W)
+val riscv_f16ToUi32 = pure {c: "softfloat_f16toui32", lem: "softfloat_f16_to_ui32"} : (bits_rm, bits_H) -> (bits_fflags, bits_WU)
+val riscv_i32ToF16 = pure {c: "softfloat_i32tof16", lem: "softfloat_i32_to_f16"} : (bits_rm, bits_W) -> (bits_fflags, bits_H)
+val riscv_ui32ToF16 = pure {c: "softfloat_ui32tof16", lem: "softfloat_ui32_to_f16"} : (bits_rm, bits_WU) -> (bits_fflags, bits_H)
+val riscv_f16ToI64 = pure {c: "softfloat_f16toi64", lem: "softfloat_f16_to_i64"} : (bits_rm, bits_H) -> (bits_fflags, bits_L)
+val riscv_f16ToUi64 = pure {c: "softfloat_f16toui64", lem: "softfloat_f16_to_ui64"} : (bits_rm, bits_H) -> (bits_fflags, bits_LU)
+val riscv_i64ToF16 = pure {c: "softfloat_i64tof16", lem: "softfloat_i64_to_f16"} : (bits_rm, bits_L) -> (bits_fflags, bits_H)
+val riscv_ui64ToF16 = pure {c: "softfloat_ui64tof16", lem: "softfloat_ui64_to_f16"} : (bits_rm, bits_L) -> (bits_fflags, bits_H)
```
jordancarlin: Should be unsigned.
```suggestion
val riscv_ui64ToF16 = pure {c: "softfloat_ui64tof16", lem: "softfloat_ui64_to_f16"} : (bits_rm, bits_LU) -> (bits_fflags, bits_H)
```

model/riscv_softfloat_interface.sail
```diff
@@ -37,488 +37,89 @@ type bits_WU     = bits(32)   /* Unsigned integer */
...
+
+val riscv_f32ToI32 = pure {c: "softfloat_f32toi32", lem: "softfloat_f32_to_i32"} : (bits_rm, bits_S) -> (bits_fflags, bits_W)
+val riscv_f32ToUi32 = pure {c: "softfloat_f32toui32", lem: "softfloat_f32_to_ui32"} : (bits_rm, bits_S) -> (bits_fflags, bits_WU)
+val riscv_i32ToF32 = pure {c: "softfloat_i32tof32", lem: "softfloat_i32_to_f32"} : (bits_rm, bits_W) -> (bits_fflags, bits_S)
+val riscv_ui32ToF32 = pure {c: "softfloat_ui32tof32", lem: "softfloat_ui32_to_f32"} : (bits_rm, bits_WU) -> (bits_fflags, bits_S)
+val riscv_f32ToI64 = pure {c: "softfloat_f32toi64", lem: "softfloat_f32_to_i64"} : (bits_rm, bits_S) -> (bits_fflags, bits_L)
+val riscv_f32ToUi64 = pure {c: "softfloat_f32toui64", lem: "softfloat_f32_to_ui64"} : (bits_rm, bits_S) -> (bits_fflags, bits_LU)
+val riscv_i64ToF32 = pure {c: "softfloat_i64tof32", lem: "softfloat_i64_to_f32"} : (bits_rm, bits_L) -> (bits_fflags, bits_S)
+val riscv_ui64ToF32 = pure {c: "softfloat_ui64tof32", lem: "softfloat_ui64_to_f32"} : (bits_rm, bits_L) -> (bits_fflags, bits_S)
```
jordancarlin: Should be unsigned.
```suggestion
val riscv_ui64ToF32 = pure {c: "softfloat_ui64tof32", lem: "softfloat_ui64_to_f32"} : (bits_rm, bits_LU) -> (bits_fflags, bits_S)
```

model/riscv_softfloat_interface.sail
Timmmm: No, the "c" here is referring to the Sail C backend which we are still using.

nadime15: Ahh ok I see, I assume after #1274 we would switch to cpp here?

Timmmm: Yep

model/riscv_softfloat_interface.sail
Timmmm: Ooo good spot. The mistake was in the original code but I guess it didn't matter because they're both `bits(64)`. How did you notice this?

arichardson (approved):
Not reviewed in detail but this looks much nicer!

model/riscv_softfloat_interface.sail
jordancarlin: The `riscv_*` version in the original code was correct, but the `extern_*` version was wrong. Spotted while reviewing.

Timmmm:
I just want to find out how to do the Lem code correctly before merging...

Alasdair:
For lem you can just use `fail` or `failwith <string>`. See https://github.com/rems-project/lem/blob/master/library/assert_extra.lem.

Timmmm:
Great thanks! I updated it to `fail`.

Issue #1250, opened 2025-09-02
Add syscall support


trdthg:
I've been working on this recently，the target is to making sail-riscv able to run user programs like qemu-user/spike/[NEMU](https://github.com/OpenXiangShan/NEMU)

Currently I'm testing riscv-pk with spike (thanks to #1102)

related issues

- #252
- #346 

I have some code locally and will push incrementally.

jrtc27:
I don't think this is a good idea. The model is for specifying and emulating the architecture, and being usable for formal proofs, not for being an all-singing all-dancing emulator that includes OS-specific syscall emulation.

trdthg:
I think this doesn't go against the model.
Implementing syscall will only insert one dispatch_syscall function on the sail side, and the rest are C stuff, which are separated from the model. Also this won't require too much code.

jrtc27:
But what syscalls? Are you implementing some toy OS set of syscalls? Are you reimplementing some existing OS's syscalls? Or are you translating to the native OS's syscalls (and if so, which OSes do you support)? None of those are at all a part of the architecture. They are all part of a software stack that sits on top of the architecture that you are choosing to bake into the model, and that applies even if written in C. This is not an implement-an-OS repository, this is an implement-an-architecture repository.

trdthg:
> But what syscalls?

Translating to the native OS's syscalls. I currently only plan to support a partial set of file-related system calls, specifically read/write/open/..., and exit.

jrtc27:
Experience with qemu-user says that this is a much more complicated task than anyone ever thinks, and I have no desire to see sail-riscv end up rediscovering all of the hell that qemu-user has gone through to get something that mostly works.

If your goal is to support reading/writing files, better to define some kind of device plugin abstraction so that people who want it can load some kind of VirtIO device and use that, rather than getting into the horrible world of syscall emulation (and ultimately ending up favouring a specific OS's ABI, shunning all others, which is not something that an OS-agnostic project like sail-riscv should be wading into).

Timmmm:
Yeah I'm not sure about this either. HTIF isn't exactly well specified. Also I don't really understand why you would want to run another emulator *inside* the Sail emulator?

I do think we want proper terminal IO but I'd much rather that was done by implementing a *real* UART device - probably AXI UART Lite. That is very simple but it also requires PLIC support... which is also fairly simple but a bit more work so I haven't got around to it yet.

What syscalls would you want apart from terminal IO? Do you want host filesystem access?

[...]

PR #1276, opened 2025-09-14
WIP: Add Semihosting Support

trdthg:
link to #1250 

Currently I have only implemented and tested a subset of the semihosting operations, based on the gem5 implementation, and there are still a lot of work to be done, including integrating the tests into CMake.
So for now, this is just a show of how to integrate semihosting into the sail-model to ensure we‘re heading in the right direction.

### DIFF
--- a/c_emulator/mem.cpp
+++ b/c_emulator/mem.cpp
@@ -2,25 +2,39 @@
 #include <cstddef>
 #include "mem.h"
 
+bool mem::target_is_little_endian = true;
+
 void mem::read(uint64_t addr, void *dst, size_t len)
 {
   uint8_t *dest_bytes = static_cast<uint8_t *>(dst);
-  for (size_t i = 0; i < len; i += 1) {
-    dest_bytes[i] = read_mem(addr + i);
+  if (same_endian()) {
+    for (size_t i = 0; i < len; i++) {
+      dest_bytes[i] = read_mem(addr + i);
+    }
+  } else {
+    for (size_t i = 0; i < len; i++) {
+      dest_bytes[i] = read_mem(addr + (len - 1 - i));
+    }
   }
 }
 
 void mem::write(uint64_t addr, const void *src, size_t len)
 {
   const uint8_t *src_bytes = static_cast<const uint8_t *>(src);
-  for (size_t i = 0; i < len; i += 1) {
-    write_mem(addr + i, src_bytes[i]);
+  if (same_endian()) {
+    for (size_t i = 0; i < len; i++) {
+      write_mem(addr + i, src_bytes[i]);
+    }
+  } else {
+    for (size_t i = 0; i < len; i++) {
+      write_mem(addr + (len - 1 - i), src_bytes[i]);
+    }
   }
 }
 
 void mem::clear(uint64_t addr, size_t len)
 {
-  for (size_t i = 0; i < len; i += 1) {
+  for (size_t i = 0; i < len; i++) {
     write_mem(addr + i, 0);
   }
 }
--- a/c_emulator/mem.h
+++ b/c_emulator/mem.h
@@ -7,6 +7,15 @@
 
 class mem {
 public:
+  // Set target endian, default is little endian
+  static void set_target_endian(bool is_little)
+  {
+    target_is_little_endian = is_little;
+  }
+
+  // Read from sail mem, start from addr, len = len
+  // This function will read bytes in reverse order if host endian and target
+  // endian are different
   static void read(uint64_t addr, void *dst, size_t len);
 
   // start from addr, len = sizeof(T)
@@ -51,4 +60,18 @@ public:
 
   static std::string read_string(const char *ptr, size_t len);
 
+private:
+  static bool target_is_little_endian;
+  static bool host_is_little_endian()
+  {
+    union {
+      uint16_t u16;
+      uint8_t u8[2];
+    } test = {.u16 = 0x0001};
+    return test.u8[0] == 0x01;
+  }
+  static bool same_endian()
+  {
+    return host_is_little_endian() == target_is_little_endian;
+  }
 };


.gitignore
```diff
@@ -12,3 +12,4 @@ z3_problems
 /build
 # CMake build directories used by CLion.
 /cmake-build-*
+test/semihosting
```
arichardson: This probably shouldn't be here? If you want to add it locally you cna use .git/info/exclude

c_emulator/semihosting.cpp
```diff
@@ -0,0 +1,368 @@
...
+};
+
+}
+
+// Prefix = 0x01f01013, // slli x0, x0, 0x1f Entry NOP
+// EBreak = 0x00100073, // ebreak            Break to debugger
+// Suffix = 0x40705013, // srai x0, x0, 7    NOP encoding semihosting
+bool is_semihosting()
+{
+  return (mem::read<uint32_t>(zPC.bits - 4) == 0x01f01013)
```
arichardson, line 525: This here is slightly fragile, if someone builds the sail emulator on a big-endian platform they will not detect this. Either ensure you convert from little endian or just compare the individual bytes?
E.g. `char magic[12]; mem::read(zPC.bits - 4, magic, 12); memcmp(magic, ...) == 0)`

c_emulator/semihosting.cpp
```diff
@@ -0,0 +1,368 @@
+// See LICENSE for license details.
```
arichardson, line 1: Unrelated but I think we should just add a `SPDX-License-Identifier:` line to all files?

Some of this looks copied from gem5, I think this is fine due to the license, but probably need to add some attribution?

.gitignore
trdthg: I've added a few test files and integrated them with both CMake and CI. Although I meet some issues:

- The riscv-gnu-toolchain seems have issue when compiling Picolibc. https://github.com/picolibc/picolibc/issues/1086
- Picolibc doesn't seem to work well with Clang, so for now, the semihosting tests are only supported compiling with GCC. https://github.com/RIOT-OS/RIOT/issues/19642
- I don't know if the riscv64-unknown-elf-gcc package is available on macOS via Brew, so I only enabled Ubuntu in the CI.
- The picolibc package from apt on Ubuntu 22.04 is missing crt0-semihost.so, so I have to build picolibc from source from the official release. (ubuntu2404 is ok, at least in my machine)

c_emulator/semihosting.cpp
trdthg: This will be addressed next. And I guess we should also push forward #751?

trdthg:
Features are basically complete now, also added a few tests and integrated them with both CMake and CI

Add two new command-line arguments :
- `--enable-semihosting`: to enable or disable semihosting.
- `--elf-args`: to pass arguments to the ELF for the `GetCmdline` interface.

Also added two new CMake parameters:
- `DOWNLOAD_PICOLIBC`: whether to download picolibc source from github release and build it locally.
- `SEMIHOSTING_TESTS`: whether to enable the semihosting tests. (macos is not able to run this yet)

The C++ template syntax here is quite complex, but it's also the way to keep the SYS_XXX implementations concise. I haven't found a better solution. Although I believe this is already much simpler than the gem5 implementation.

c_emulator/semihosting.cpp
arichardson: I think this is unrelated since I am concerned about the host platform being big endian, not the simulation target. Instructions are always the same endianess so this  pr should be independent of that

c_emulator/mem.cpp
```diff
@@ -0,0 +1,70 @@
...
+  return ret;
+}
+
+std::string mem::read_string(const char *ptr, size_t len)
+{
+  std::string ret;
+  uint64_t addr = (uint64_t)ptr;
+  for (size_t i = 0; i < len; i += 1) {
+    char c = read_mem((uint64_t)addr);
+    if (c) {
```
arichardson, line 64: Why are we skipping nul bytes here? Wouldn't it make sense to also just stop and treat Len as a maximum length?

c_emulator/mem.cpp
```diff
@@ -0,0 +1,70 @@
+#include <inttypes.h>
+#include <cstddef>
+#include "mem.h"
+
+bool mem::target_is_little_endian = true;
+
+void mem::read(uint64_t addr, void *dst, size_t len)
```
arichardson, line 7: I don't think these functions should be doing the byteswapping, that should be done by wrapper functions that read an integer type

c_emulator/mem.h
```diff
@@ -0,0 +1,77 @@
...
+
+  // Read from sail mem, start from addr, len = len
+  // This function will read bytes in reverse order if host endian and target
+  // endian are different
+  static void read(uint64_t addr, void *dst, size_t len);
+
+  // start from addr, len = sizeof(T)
+  template <typename T> static T read(uint64_t addr)
+  {
+    uint8_t buffer[sizeof(T)] = {0};
```
arichardson, line 24:
```suggestion
    alignas(T) uint8_t buffer[sizeof(T)] = {0};
```
Otherwise the reinterpret_cast is unsafe. Or you could use a union.

c_emulator/semihosting.cpp
```diff
@@ -0,0 +1,534 @@
...
+    } else {
+      printf("Unknown file mode for the ':tt' special file");
+      return -EINVAL;
+    }
+    return 0;
+  } else {
+    std::string real_mode(mode);
+    if (real_mode[0] == 'w')
+      real_mode[0] = 'a';
+    file = fopen(name.c_str(), real_mode.c_str());
```
arichardson, line 52: Should we support access to the host file system by default? Gem5 requires an extra flag to do that and I feel we should do here too.

c_emulator/semihosting.cpp
```diff
@@ -0,0 +1,534 @@
...
+RiscvSemihosting::File::File(const std::string &name, const char *mode)
+    : name(name)
+    , mode(mode)
+    , file(nullptr)
+{
+}
+
+RiscvSemihosting::File::~File()
+{
+  if (file) {
```
arichardson, line 24: We should not be closing stdin/stdout/stderr here

c_emulator/semihosting.cpp
```diff
@@ -0,0 +1,534 @@
...
+}
+
+RiscvSemihosting::FileFeatures::FileFeatures(const std::string &name,
+                                             const char *mode)
+    : name(name)
+    , mode(mode)
+{
+}
+
+int64_t RiscvSemihosting::FileFeatures::read(uint8_t *buffer, uint64_t size)
```
arichardson, line 120: I wonder if we can replace this whole class with a call to fmemopen() in read-only mode. That should hopefully be equivalent

PR #1284, opened 2025-09-18
Improve exception code types

Timmmm:
Switch `exc_code` to 6 bits, [0, 64). For exceptions, higher values are reserved and can't be implemented without significant changes because there would be no space in `medeleg[h]`.

For interrupts, they are actually even further limited to [0, xlen), unless you have AIA which adds `mieh` and so on.

I also changed the functions to mappings which is a bit nicer and removes the need for `num_of_ExceptionType()`.

It's unfortunate that we have to use binary literals but it's not too bad.


nadime15 (approved):

Timmmm:
@pmundkur can you make @nadime15 a member of the repo (or whatever is required to make his reviews worthy according to Github)?

pmundkur (approved):

PR #1285, opened 2025-09-18
Add basic NMI support

Timmmm:
If you call `zset_pending_nmi(0)` before `ztry_step()` then this will now trigger an NMI and jump to a configurable address.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -55,6 +55,7 @@
       "base": 33554432,
       "size": 786432
     },
+    "nmi_handler_address": 0,
     "clock_frequency": 1000000000,
     "instructions_per_tick": 2,
     "wfi_is_nop": true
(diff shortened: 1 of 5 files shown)


Timmmm:
This code isn't actually called from anywhere. We did something like this on a previous project and it worked will though. Some things to think about:

1. Is adding the extra function the best way, or should we simply add a parameter to `try_step()` that forces an NMI. After having written the code I'm kind of thinking maybe the parameter is a better approach. It feels a bit less hacky.
2. How do we test this?

Timmmm:
Discussed in the meeting:

1. We'll go with a separate function for now.
2. Won't bother testing it for now. We don't test normal interrupts anyway. Eventually we can use a "trickbox" (MMIO device) to trigger NMIs from tests.

So I think this is ready. No idea what the Lean failure is.

PR #1286, opened 2025-09-19
Reduce duplication in vsetvl-type instructions

Timmmm:
Reduce duplication by adding a function that can execute all three `vsetvl`-type instructions and delegating to those.

This also makes it a bit easier to see how the instructions differ and how they are the same.

### DIFF
--- a/model/extensions/V/vext_vset_insts.sail
+++ b/model/extensions/V/vext_vset_insts.sail
@@ -103,7 +103,7 @@ function execute_vsetvl_type(
   //
   // Also some cases require that VLMAX doesn't change (i.e. LMUL/SEW is constant).
   if SEW_pow_new > (LMUL_pow_new + elen_exp) |
-     (requires_fixed_vlmax & (lmul_sew_ratio != lmul_sew_ratio_new | not(valid_vtype())) then {
+     (requires_fixed_vlmax & (lmul_sew_ratio != lmul_sew_ratio_new | not(valid_vtype()))) then {
     handle_illegal_vtype();
     return RETIRE_SUCCESS;
   };


Timmmm:
This is on top of #1287.

model/extensions/V/vext_vset_insts.sail
```diff
@@ -156,43 +150,7 @@ function clause execute VSETVL(rs2, rs1, rd) = {
...
-
-  // reset vstart to 0
-  set_vstart(zeros());
-
-  csr_write_callback("vtype", vtype.bits);
-  csr_write_callback("vl", vl);
-  csr_write_callback("vstart", vstart);
-
-  RETIRE_SUCCESS
+  execute(VSETVLI(ma, ta, sew, lmul, rs1, rd))
```
jordancarlin: I'm not convinced that we want the `execute` clauses to be recursive (we should revisit #905 which is related to this as well). What about pulling the logic out into a separate function that both `vsetvli` and `vsetvl` call?

Timmmm: Yeah good point, I will do that.

Timmmm: I split it into a separate function, `execute_vsetvli`.

model/extensions/V/vext_vset_insts.sail
```diff
@@ -140,58 +149,12 @@ mapping clause encdec = VSETVL(rs2, rs1, rd)
...
-
-  // reset vstart to 0
-  set_vstart(zeros());
-
-  csr_write_callback("vtype", vtype.bits);
-  csr_write_callback("vl", vl);
-  csr_write_callback("vstart", vstart);
-
-  RETIRE_SUCCESS
+  execute_vsetvli(vt[vma], vt[vta], vt[vsew], vt[vlmul], rs1, rd)
```
nadime15: It might make more sense to merge vsetvli, vsetivli, and vsetvl into a single function, passing an enum to indicate which instruction to execute and handling the corresponding case accordingly. I think it's a bit easier to maintain.

Timmmm: How about this version? I also refactored it a bit so the shared code only deals with AVL and the `rs1 != zreg` type stuff is done in the instruction (since it doesn't apply to `vsetivli`).

model/extensions/V/vext_vset_insts.sail
```diff
@@ -77,47 +76,50 @@ function calculate_new_vl(AVL : xlenbits, VLMAX : range(1, vlen)) -> range(0, vl
...
-    // Keep existing vl if it is compatible and vtype.ill is valid.
-    let ratio_pow_new = SEW_pow_new - LMUL_pow_new;
-    let is_legal_vset = ratio_pow_new == ratio_pow_ori & valid_vtype();
-    if not(is_legal_vset) then { handle_illegal_vtype(); return RETIRE_SUCCESS }
+  // "For a given supported fractional LMUL setting, implementations must
+  // support SEW settings between SEWMIN and LMUL * ELEN, inclusive."
+  //
+  // Also some cases require that VLMAX doesn't change (i.e. LMUL/SEW is constant).
+  if SEW_pow_new > (LMUL_pow_new + elen_exp) |
+     (requires_fixed_vlmax & (lmul_sew_ratio != lmul_sew_ratio_new)) then {
```
pmundkur: The `requires_fixed_vlmax` case should also ensure: "Use of the instructions
is also reserved if `vill` was 1 beforehand."
```suggestion
     (requires_fixed_vlmax & (lmul_sew_ratio != lmul_sew_ratio_new | not(valid_vtype())) then {
```

pmundkur (changes requested):
Nice cleanup.

model/extensions/V/vext_vset_insts.sail
nadime15: Yeah looks good!

nadime15 (approved):

model/extensions/V/vext_vset_insts.sail
Timmmm: Hmm this is handled on line 165, but it looks like it was actually missing for `vsetvli` before, even though the comment says `and vtype.ill is valid` (unless there's a check I'm not seeing).

It probably makes sense to handle it here anyway.

Timmmm: > unless there's a check I'm not seeing

Wait, I am blind. Also the check on 165 is actually for the *new* `vtype`.

Good catch!

model/extensions/V/vext_vset_insts.sail
```diff
@@ -77,47 +76,50 @@ function calculate_new_vl(AVL : xlenbits, VLMAX : range(1, vlen)) -> range(0, vl
...
-    // Keep existing vl if it is compatible and vtype.ill is valid.
-    let ratio_pow_new = SEW_pow_new - LMUL_pow_new;
-    let is_legal_vset = ratio_pow_new == ratio_pow_ori & valid_vtype();
-    if not(is_legal_vset) then { handle_illegal_vtype(); return RETIRE_SUCCESS }
+  // "For a given supported fractional LMUL setting, implementations must
+  // support SEW settings between SEWMIN and LMUL * ELEN, inclusive."
+  //
+  // Also some cases require that VLMAX doesn't change (i.e. LMUL/SEW is constant).
+  if SEW_pow_new > (LMUL_pow_new + elen_exp) |
+     (requires_fixed_vlmax & (lmul_sew_ratio != lmul_sew_ratio_new | not(valid_vtype())) then {
```
pmundkur:
```suggestion
     (requires_fixed_vlmax & (lmul_sew_ratio != lmul_sew_ratio_new | not(valid_vtype()))) then {
```
should fix the build I think

pmundkur (changes requested):

pmundkur:
Careful when rebasing this; the web editor is not picking up the  `when currentlyEnabled(Ext_Zve32x)` for `VSETVLI`.

Timmmm:
Thanks, totally would have missed that!

PR #1288, opened 2025-09-19
Improve float classify code

Timmmm:
Use the Sail library's new float classify function. This reduces code duplication and the presence of magic numbers.

This is the first use of Sail's built-in float library, which we should try to migrate to.

The SMT properties should ideally live in the Sail repo but it doesn't have SMT testing set up yet.

### DIFF
--- a/model/core/float_classify.sail
+++ b/model/core/float_classify.sail
@@ -20,9 +20,6 @@ mapping float_class_onehot_bits : float_class <-> bits(10) = {
   float_class_qnan               <-> 0b1000000000,
 }
 
-// Helper function for proofs.
-function bool_int(b : bool) -> {0, 1} = if b then 1 else 0
-
 // Prove that the float is in one broad class.
 $[property]
 function float_in_one_broad_class(f : bits(16)) -> bool =
--- a/model/core/prelude.sail
+++ b/model/core/prelude.sail
@@ -102,6 +102,12 @@ mapping bool_bits : bool <-> bits(1) = {
   false  <-> 0b0,
 }
 
+// Useful for some proofs where you want to assert that exactly one bool is true.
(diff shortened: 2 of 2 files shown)


Timmmm:
I restored the SMT properties since they work with the latest Sail compiler. I think they probably won't work with the current one due to a bug.

Also I'm wondering if maybe I should just add the classify function to Sail's float library.

model/core/float_classify.sail
```diff
@@ -0,0 +1,54 @@
...
+  float_class_positive_zero      <-> 0b0000010000,
+  float_class_positive_subnormal <-> 0b0000100000,
+  float_class_positive_normal    <-> 0b0001000000,
+  float_class_positive_inf       <-> 0b0010000000,
+  float_class_snan               <-> 0b0100000000,
+  float_class_qnan               <-> 0b1000000000,
+}
+
+// Helper function for proofs.
+function bool_int(b : bool) -> {0, 1} = if b then 1 else 0
```
jordancarlin: Probably makes more sense for this to be in the prelude.

jordancarlin (approved):

pmundkur (approved):
Ideally, all of `float_classify.sail` should go into Sail's `classify.sail`, and the SMT properties should be checked in Sail's CI.

Timmmm:
Yeah I agree. Unfortunately the Sail repo doesn't have SMT testing set up for its lib files, and I didn't want to just throw these properties away. Hopefully we can move them there later!

Are you guys ok with merging this even though it breaks the Lean build? We have another PR ready that also breaks the Lean build in a different way so I vote we accept some breakage temporarily.

pmundkur:
Agreed, we should merge this.

PR #1289, opened 2025-09-19
CLI11: update to the current git

radimkrcmar:
Upstream implemented a different solution for the width configuration, which means we have to change our use a bit.

There should be no changes in the command line interface.

I think it might make sense to wait for an official release of CLI11, or even never update.  In any case, I'm posting this to show the required changes case we do want to update.


radimkrcmar:
Superseded by https://github.com/riscv/sail-riscv/pull/1544.

PR #1290, opened 2025-09-21
Add H_ext trap_handler support

KotorinMinami:
The second step in adding the H extension involves adding the hstatus register, as well as privilege level switching for traps in the presence of the H extension and support for related registers.


model/core/sys_regs.sail
```diff
@@ -446,14 +446,18 @@ function is_fiom_active() -> bool = {
...
-  MEI : 11, // external interrupts
-  SEI : 9,
+  SGEI : 12, /* guest external interrupts */
+  MEI   : 11, /* external interrupts */
+  VSEI : 10,
+  SEI   : 9,
 
-  MTI : 7,  // timers interrupts
-  STI : 5,
+  MTI   : 7,  /* timers interrupts */
```
arichardson: Why did you change the comment style here?

model/extensions/H/hext_regs.sail
```diff
@@ -0,0 +1,265 @@
+/*=======================================================================================*/
+/*  This Sail RISC-V architecture model, comprising all files and                        */
+/*  directories except where otherwise noted is subject the BSD                          */
+/*  two-clause license in the LICENSE file.                                              */
+/*                                                                                       */
+/*  SPDX-License-Identifier: BSD-2-Clause                                                */
+/*=======================================================================================*/
```
nadime15: We have dropped C-style comments, only use `//` everywhere now

model/core/sys_regs.sail
```diff
@@ -435,14 +435,18 @@ function is_fiom_active() -> bool = {
...
 
-  MTI : 7,  // timers interrupts
-  STI : 5,
+  MTI   : 7,  // timers interrupts
+  VSTI : 6,
+  STI   : 5,
 
-  MSI : 3,  // software interrupts
-  SSI : 1,
+  MSI   : 3,  // software interrupts
```
Timmmm, line 447: Nit but can you put the comments on the line above? End-of-line comments are normally more awkward in my experience.

model/core/types.sail
```diff
@@ -65,21 +65,23 @@ mapping architecture_bits : Architecture <-> arch_xlen = {
...
+  (0b00, 0b1) <-> VirtualUser,
+  (0b01, 0b0) <-> Supervisor,
+  (0b01, 0b1) <-> VirtualSupervisor,
+  (0b11, 0b0) <-> Machine,
   forwards _       => internal_error(__FILE__, __LINE__, "Invalid privilege level or virtual mode"),
 }
 
 function privLevel_to_bits(p : Privilege) -> nom_priv_bits = { let (p, _) = privLevel_bits(p); p }
 
+function privLevel_is_virtual(p : Privilege) -> bool = { let (_, v) = privLevel_bits(p); v == 0b1 }
```
Timmmm, line 83: Might be clearer just like this:

```
function privLevel_is_virtual(p : Privilege) -> bool =
  match p {
    User => false,
    VirtualUser => true,
...
  }
```

model/extensions/H/hext_control.sail
```diff
@@ -0,0 +1,23 @@
+// =======================================================================================
+// This Sail RISC-V architecture model, comprising all files and
+// directories except where otherwise noted is subject the BSD
+// two-clause license in the LICENSE file.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+// =======================================================================================
+
+function is_trap_write_GVA(c: exc_code) -> bool = match c {
```
Timmmm, line 9: Hmm the spec isn't clear to me here:

> For any trap (breakpoint, address misaligned, access fault, page fault, or guest-page fault) that writes a guest virtual address to mtval, GVA is set to 1. For any other trap into M-mode, GVA is set to 0.

What if you get an access fault (or whatever) but the implementation writes 0 to `mtval`? Or is that not allowed with hypervisor? It doesn't seem to say that anywhere. Closest is

> CSR mtval must not be read-only zero

but you could still write 0 to it.

Maybe a spec bug?

Anyway I think we can improve this with some minor refactoring to `trap_handler`: instead of passing `intr : bool, c : exc_code` we should pass `c : TrapCause` where `TrapCause` is

```
union TrapCause = {
  Interrupt(InterruptType),
  Exception(ExceptionType),
}
```

Then you can use symbolic names here. I'll make a separate PR to do this.

model/extensions/H/hext_regs.sail
```diff
@@ -0,0 +1,265 @@
...
+// directories except where otherwise noted is subject the BSD
+// two-clause license in the LICENSE file.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+// =======================================================================================
+
+// TODO: Ext_H depends on RV32I or RV64I, not RV32E or RV64E.
+function clause currentlyEnabled(Ext_H) = hartSupports(Ext_H) & misa[H] == 0b1 & virtual_memory_supported() & currentlyEnabled(Ext_U)
+
+// extra machine registers
```
Timmmm, line 12: Probably delete all this commented out code for now - you can add a `// TODO: Add hypervisor CSRs here.`

model/extensions/H/hext_regs.sail
```diff
@@ -0,0 +1,265 @@
...
+  VGEIN : 17 .. 12,
+
+  HU    : 9,
+  SPVP  : 8,
+  SPV   : 7,
+  GVA   : 6,
+  VSBE  : 5
+}
+
+function get_hstatus_VSXL(h : Hstatus) -> arch_xlen = {
```
Timmmm, line 161: We changed how this worked for `get_mstatus_UXL` etc. See #1292

model/extensions/H/hext_regs.sail
```diff
@@ -0,0 +1,265 @@
...
+}
+
+function legalize_hstatus(o : Hstatus, v : bits(64)) -> Hstatus = {
+  let v = Mk_Hstatus(v);
+
+  [o with
+    VTSR  = v[VTSR],
+    VTW   = v[VTW],
+    VTVM  = v[VTVM],
+    VGEIN = v[VGEIN], // Note: 0 <= hstatus.VGEIN <= GEILEN
```
Timmmm, line 174: Can you expand this note? Is this a thing we need to implement in future?

model/extensions/H/hext_regs.sail
```diff
@@ -0,0 +1,265 @@
...
+function legalize_hedeleg(o: Medeleg, v: bits(64)) -> Medeleg = {
+  let h = Mk_Medeleg(v);
+  // Mask read-only zero bits
+  let h = [h with SAMO_GPage_Fault  = 0b0];
+  let h = [h with Virtual_Instr     = 0b0];
+  let h = [h with Load_GPage_Fault  = 0b0];
+  let h = [h with Fetch_GPage_Fault = 0b0];
+  let h = [h with MEnvCall          = 0b0];
+  let h = [h with VSEnvCall         = 0b0];
+  let h = [h with SEnvCall          = 0b0];
```
Timmmm, line 206: You can do all this in one expression:

```
function legalize_hedeleg(o: Medeleg, v: bits(64)) -> Medeleg =
  [Mk_Medeleg(v) with
    SAMO_GPage_Fault = 0b0,
    Virtual_Instr = 0b0,
    ...
    SEnvCall = 0b0,
  ]
```

Timmmm (commented):
Had a skim & looks pretty good to me. I'm not very familiar with the hypervisor spec though so it's quite superficial.

jordancarlin:

PR #1294, opened 2025-09-26
Use composite action to remove workflow redundancy

jordancarlin:


.github/workflows/compile.yml
```diff
@@ -24,67 +24,25 @@ jobs:
...
-        id: cache-opam-save
-        uses: actions/cache/save@v4
-        with:
-          path: ~/.opam
-          key: ${{ steps.cache-opam-restore.outputs.cache-primary-key }}
-
-      - name: Load OPAM environment (macOS)
-        if: startsWith(matrix.os, 'macos-' )
-        run: echo "$HOME/.opam/default/bin" >> $GITHUB_PATH
+          sail-version: "${{ env.SAIL_VERSION }}"
```
arichardson: Could add this as the default to action.yml to avoid needing to update it in two places?

jordancarlin: Updated

arichardson (approved):
Great to see less code duplication

.github/actions/sail-setup/action.yml
```diff
@@ -0,0 +1,96 @@
...
+    default: 0.19.1
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install packages (Linux)
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        sudo apt install -y --no-install-recommends pkg-config libgmp-dev curl ninja-build
```
pmundkur, line 20: Might be worth adding `z3` to the default installed packages here for Linux and below for macOS since it is used by the `smt_properties` test.  Its installation then won't be needed for the from-source Sail builds.
The `smt_properties` test is still passing without this, so clearly z3 is present without this, but I'm not sure why.

jordancarlin: z3 is bundled with the Sail binary release, so we only need to install it when we are building from source on Linux. I'd rather use the bundled z3 for most of our tests to make sure it continues working with that version.

Not sure where the current z3 installation is coming from on macOS, but I can add that one globally since we aren't using a Sail binary there.

pmundkur: Ok, in that case, installing z3 only when it's needed like you have it here makes sense.

Issue #1235, opened 2025-08-22
`--print-isa-string` does not check for a valid configuration


pmundkur:

Arielfoever:
It should based on https://github.com/riscv/sail-riscv/pull/1161

Arielfoever:
After https://github.com/riscv/sail-riscv/pull/1291#issuecomment-3345468408 merged, sounds like everything checked.

```
[ariel@archlinux sail-riscv]$ build/c_emulator/sail_riscv_sim  --print-isa-string --config config/test.json 
rv64imacbv_zicbom_zicbop_zicboz_zicfilp_zicntr_zicond_zicsr_zifencei_zihintntl_zihintpause_zihpm_zimop_zabha_zacas_zawrs_zfa_zfbfmin_zfh_zca_zcb_zcd_zcmop_zbc_zbkb_zbkc_zbkx_zknd_zkne_zknh_zkr_zksed_zksh_zkt_zvbb_zvbc_zvkg_zvkned_zvknha_zvknhb_zvksed_zvksh_zvkt_sscofpmf_sstc_svinval_svrsw60t59b_smcntrpmf
```

Timmmm:
It still doesn't call `validate_config(config_file)` though. Tbh I think we could just do that unconditionally.

PR #1296, opened 2025-09-29
c_emulator: always check config except default

Arielfoever:
Fix https://github.com/riscv/sail-riscv/issues/1235#issuecomment-3345999052


c_emulator/riscv_sim.cpp
```diff
@@ -574,22 +574,20 @@ int inner_main(int argc, char **argv)
...
     sail_config_set_file(config_file.c_str());
   } else {
+    // Do not check default config
     sail_config_set_string(get_default_config());
   }
 
   init_sail_configured_types();
   model_init();
 
-  if (do_validate_config) {
```
Timmmm, line 580: This should print a success message and then `return EXIT_SUCCESS`.

c_emulator/riscv_sim.cpp
```diff
@@ -574,22 +574,20 @@ int inner_main(int argc, char **argv)
...
-  validate_config_schema(config_file);
-
   // Initialize the model.
   if (!config_file.empty()) {
+    // Always validate the schema conformance of the config.
+    validate_config_schema(config_file);
+    validate_config(config_file);
     sail_config_set_file(config_file.c_str());
   } else {
+    // Do not check default config
```
Timmmm: IMO this is the right thing to do and we'll just check the generated configs at build time. I think @pmundkur may disagree though so we should check with him.

Timmmm (changes requested):
We also need to:

1. Remove the *success* message from `validate_config()` (so it doesn't get printed unless you explicitly ask for `--validate-config`. It should still print an error if the config is invalid.
2. Probably change the help for `--validate-config` like this:

```
  app.add_flag("--validate-config", do_validate_config,
               "Exit after config validation (it is always validated)");
```

Arielfoever:
Will come back and deal with the comments after few hours.

c_emulator/riscv_sim.cpp
```diff
@@ -574,22 +575,34 @@ int inner_main(int argc, char **argv)
...
+    if (do_validate_config) {
+      fprintf(stderr, "Default Config had passed validations when building.\n");
+    }
   }
 
   init_sail_configured_types();
   model_init();
 
+  bool result = validate_config(config_file);
+  if (!result) {
```
Timmmm: May as well delete `validate_config` and just do 
```suggestion
  if (!zconfig_is_valid(UNIT)) {
```

Otherwise this LGTM.

c_emulator/riscv_sim.cpp
pmundkur: Yes, I do think that just always validating the config is simpler instead of littering the code with special cases like this.

Timmmm: Ok I'm fine with that too.

Arielfoever:
Rebase and deal with comment.

Timmmm:
I think you might have pushed the wrong commit - the comments are still outstanding. I can fix them if you like.

c_emulator/riscv_sim.cpp
Arielfoever: So you think it's always check config?

c_emulator/riscv_sim.cpp
Arielfoever: There is always message in Line 599 and exit in Line 602

c_emulator/riscv_sim.cpp
Timmmm: Yes, always check the config.

c_emulator/riscv_sim.cpp
Timmmm: I think we probably don't want to print a message in normal use though. Like this:

```
  if (!validate_config(config_file)) {
    fprintf(stderr, "Fatal Error: validate_config failed.\n");
    exit(EXIT_FAILURE);
  }
  if (do_validate_config) {
    fprintf(stderr, "Config passed validation.\n");
    exit(EXIT_SUCCESS);
  }
```

c_emulator/riscv_sim.cpp
```diff
@@ -563,22 +552,30 @@ int inner_main(int argc, char **argv)
...
+
   } else {
+    // Do not check default config schema
     sail_config_set_string(get_default_config());
   }
 
   init_sail_configured_types();
   model_init();
 
+  if (!validate_config(config_file)) {
```
```suggestion
  if (!zconfig_is_valid(UNIT)) {
```

c_emulator/riscv_sim.cpp
```diff
@@ -563,22 +552,30 @@ int inner_main(int argc, char **argv)
     fprintf(stderr, "using %s for trace output.\n", trace_log_path.c_str());
   }
 
-  // Always validate the schema conformance of the config.
-  validate_config_schema(config_file);
-
```
pmundkur, line 569: Please don't move this to be conditional.  As the comment said,  *Always validate the schema ...*

c_emulator/riscv_sim.cpp
```diff
@@ -563,22 +552,30 @@ int inner_main(int argc, char **argv)
     fprintf(stderr, "using %s for trace output.\n", trace_log_path.c_str());
   }
 
-  // Always validate the schema conformance of the config.
-  validate_config_schema(config_file);
```
Timmmm, line 568: I think @pmundkur wanted to always do this too.

c_emulator/riscv_sim.cpp
```diff
@@ -563,22 +552,30 @@ int inner_main(int argc, char **argv)
...
 
   init_sail_configured_types();
   model_init();
 
+  if (!zconfig_is_valid(UNIT)) {
+    fprintf(stderr, "Fatal Error: validate_config failed.\n");
+    exit(EXIT_FAILURE);
+  } else {
+    fprintf(stderr, "Config passed validations.\n");
+  }
```
pmundkur: The previous code helpfully specified the configuration file that failed the validity check.  It would be good to keep that feature (it's useful when running a testing script that runs over multiple files and configurations).

c_emulator/riscv_sim.cpp
Arielfoever: Limit to `void validate_config_schema(const std::string &conf_file)` - you can only input file to it - shall I edit the input?

c_emulator/riscv_sim.cpp
Arielfoever: Limit to `void validate_config_schema(const std::string &conf_file)` - you can only input file to it - shall I edit the input?

Timmmm: It treats the empty string as the default config so it was ok as it was.

c_emulator/riscv_sim.cpp
Arielfoever: done

c_emulator/riscv_sim.cpp
```diff
@@ -563,7 +564,6 @@ int inner_main(int argc, char **argv)
     fprintf(stderr, "using %s for trace output.\n", trace_log_path.c_str());
   }
 
-  // Always validate the schema conformance of the config.
   validate_config_schema(config_file);
```
pmundkur: I'm not sure why this comment was removed?

c_emulator/riscv_sim.cpp
```diff
@@ -82,15 +82,16 @@ uint64_t insn_limit = 0;
...
-  }
-  exit(EXIT_SUCCESS);
+  return zconfig_is_valid(UNIT);
+  // const char *s = zconfig_is_valid(UNIT) ? "valid" : "invalid";
+  // if (!conf_file.empty()) {
+  //   fprintf(stdout, "Configuration in %s is %s.\n", conf_file.c_str(), s);
+  // } else {
+  //   fprintf(stdout, "Default configuration is %s.\n", s);
+  // }
+  // exit(EXIT_SUCCESS);
```
Timmmm: This function appears to have returned! Would it be easier if I just pushed the final tweaks?

Arielfoever: Not sure for your idea? Do you want to keep it as no return?

Timmmm: Delete this function. Just call `zconfig_is_valid(UNIT)` directly.

pmundkur:
@Arielfoever I've made some minor fixes, hope that's ok.

c_emulator/riscv_sim.cpp
```diff
@@ -577,14 +564,30 @@ int inner_main(int argc, char **argv)
   init_sail_configured_types();
   model_init();
 
-  if (do_validate_config) {
-    validate_config(config_file);
+  // Validate the configuration; exit if that's all we were asked to do
+  // or if the validation failed.
+  {
+    bool config_is_valid = zconfig_is_valid(UNIT);
+    const char *s = config_is_valid ? "valid" : "invalid";
```
jordancarlin, line 571: Maybe use a more descriptive string name. Something like `validation_string`?

jordancarlin (approved):

c_emulator/riscv_sim.cpp
pmundkur: Not sure it's worth it since it's just a temporary.

pmundkur (approved):

PR #1297, opened 2025-09-29
Validate generated configs at build time

Timmmm:
Once the emulator is built this will run it with `--validate-config` on all of the generated configs (this includes the default one).

This is done by default with `make` (including `./build_simulators.sh`).

I verified that it works by breaking the config and then fixing it again.

### DIFF
--- a/config/CMakeLists.txt
+++ b/config/CMakeLists.txt
@@ -1,4 +1,4 @@
-set(config_validation_files)
+set(generated_config_filenames)
 
 # Create a variety of configuration files for different XLEN/ELEN/VLENs.
 foreach (CONFIG__BASE__XLEN IN ITEMS 32 64)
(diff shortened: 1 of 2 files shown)


config/CMakeLists.txt
```diff
@@ -16,10 +18,30 @@ foreach (CONFIG__BASE__XLEN IN ITEMS 32 64)
             install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${config_filename}
                 DESTINATION ${CMAKE_INSTALL_DATADIR}/${CMAKE_PROJECT_NAME}/config
             )
+
+            # After the emulator is built, run it to verify this config is valid.
+            add_custom_command(
```
Arielfoever: Why not set it with `POST_BUILD` 

```cmake
add_custom_command(sail_riscv_sim POST_BUILD
 COMMAND something
)
```

Timmmm: Because I didn't know about it! :-D Seems like a good idea. This makes it a bit trickier:

> The <target> must be defined in the current directory; targets defined in other directories may not be specified.

but I'll try it.

Timmmm: Yeah that seems a lot simpler, thanks for the suggestion!

PR #1302, opened 2025-09-29
Fix line-style doc comments

Timmmm:
The very latest Sail compiler (used in the Lean builds) supports `///` style block comments, however doc comments are only allowed for top level definitions.

We had some for bitfield fields in RVFI. This commit changes those to normal comments, and also one other doc comment that wasn't attached to a definition.

Technically the compiler change is a breaking change but I think it's pretty minor so I think us just fixing it is fine.

### DIFF
--- a/model/core/rvfi_dii.sail
+++ b/model/core/rvfi_dii.sail
@@ -45,47 +45,47 @@ function print_instr_packet(bs) = {
 
 
 bitfield RVFI_DII_Execution_Packet_InstMetaData : bits(192) = {
-  /// The rvfi_order field must be set to the instruction index. No indices
-  /// must be used twice and there must be no gaps. Instructions may be
-  /// retired in a reordered fashion, as long as causality is preserved
-  /// (register and memory write operations must be retired before the read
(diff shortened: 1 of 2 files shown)


pmundkur (approved):

Alasdair:
I can add doc comment support for struct and bitfield members I think. They are already allowed for union members:
```
union foo = {
  /// Allowed here
  Ctor : T,
}
```
so it would make sense in structs too.

Alasdair:
See https://github.com/rems-project/sail/pull/1483. Need to do a bit more plumbing to actually attach them into the syntax tree.

In general if there are places where it's A) likely users will want to write documentation comments and B) we can't actually support them for whatever reason, the best thing might be to parse them and then print a warning that they are not being applied. Otherwise a documentation comment in the wrong place will just emit a fairly plain 'syntax error'.

Timmmm:
That would be good. I'll just merge this for now since I don't think we care about doc comments in RVFI (and there's a PR to move it to C anyway).

PR #1301, opened 2025-09-29
Refactor vmem_read() and vmem_write() functions

nadime15:
Add vmem_read_addr() function that accept direct memory addresses and pull check_misligned() into vmem_write_addr. This is needed for debug module abstract commands (Access Memory) which don't use register-based addressing.

PR #1303, opened 2025-09-29
Enable 128-bit memory access

nadime15:
Enable 128-bit memory access for potential use by the debug module via abstract commands (Access Memory). Remove unused function is_aligned_bits().

### DIFF
--- a/model/sys/vmem_utils.sail
+++ b/model/sys/vmem_utils.sail
@@ -160,7 +160,7 @@ function vmem_write_addr(vaddr, width, data, acc, aq, rl, res) = {
 
 // External API
 
-function check_misaligned(vaddr : virtaddr, width : word_width) -> bool = {
+function check_misaligned(vaddr : virtaddr, width : word_width_wide) -> bool = {
   if plat_enable_misaligned_access then {
     false
   } else {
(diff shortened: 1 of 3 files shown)


model/sys/vmem_utils.sail
```diff
@@ -160,7 +160,7 @@ function vmem_write_addr(vaddr, width, data, acc, aq, rl, res) = {
 
 // External API
 
-function check_misaligned(vaddr : virtaddr, width : word_width) -> bool = {
+function check_misaligned(vaddr : virtaddr, width : word_width_wide) -> bool = {
```
Timmmm, line 163: This doesn't feel like quite the right type. Technically it's the same as `is_mem_width()` but it was kind of meant to be specific to instruction-encoding of access widths.

You could just use `nat1` for now since the underlying functions use that anyway (in future we should constrain them to something non-infinite for performance reasons, but it's fine for now).

```suggestion
function check_misaligned(vaddr : virtaddr, width : nat1) -> bool = {
```

nadime15:
I [asked](https://github.com/riscv/riscv-debug-spec/issues/1148) for clarification on whether the debug module in 32-bit or 64-bit systems can perform 128-bit memory accesses using abstract commands (Access Memory). The answer is no, so I will close this because we dont support RV128I yet (which is not frozen yet). #1301 still applies. It would be nice if you could have a quick look, @Timmmm.

Timmmm:
We could remove `is_aligned_bits` still I guess.

PR #1307, opened 2025-09-30
compile lean executable with limited modules

jn80842:
Compiling the lean executable with the full set of sail_riscv models consumes too much memory. I'm trying to use the module system to compile in all we need, but I'm not sure of the dependencies between modules and I can't find a combination of modules that compiles successfully. For example, it looks like termination.sail uses a lot of constructors from modules that I'm trying to exclude.

`Type error:
termination/termination.sail:45.4-21:
45 |    C_FLDSP(uimm, rd) => 1,
   |    ^---------------^
   | C_FLDSP is not a union constructor or mapping in pattern C_FLDSP(uimm, rd)`

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -20,6 +20,7 @@ set(sail_common
 )
 
 set(project_file "riscv.sail_project")
+set(executable_project_file "riscv_executable.sail_project")
 
 # Reconfigure if the project file changes.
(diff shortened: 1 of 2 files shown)


pmundkur:
This might require the `compressed_measure` function to be scattered like the instructions into module-specific termination files.

jordancarlin:
Rather than adding a new project file, we should finish the approach that @Alasdair started in #1219.

jordancarlin:
Closing in favor of #1219 now that it is merged. A separate PR will need to follow to resolve the termination file issues.

PR #1309, opened 2025-10-01
Add linux boot to CI

jordancarlin:
Runs the first 20,000,000 instructions of Linux boot as part of CI. This takes ~25 minutes, which is the same length as the existing CI jobs. The current checking is a couple of basic grep strings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,12 +57,57 @@ jobs:
           name: event.json
           path: ${{ github.event_path }}
 
+  linux:
+    name: "Linux boot"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install packages
+        run: sudo apt-get install -y --no-install-recommends device-tree-compiler
+
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
+      - name: Common setup
+        uses: ./.github/actions/sail-setup
+        with:
+          cmake-version: latest
+
+      - name: Build simulator
+        run: |
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo
+          ninja -C build
+
+      - name: Restore cache (linux image)
+        uses: actions/cache/restore@v4
+        id: linux-cache-restore
+        with:
+          path: os-boot/linux/build/fw_payload.elf
+          key: linux-image-${{ hashFiles('os-boot/linux/*.url', 'os-boot/linux/Makefile') }}
+
+      - name: Build Linux image
+        if: steps.linux-cache-restore.outputs.cache-hit != 'true'
+        run: make -C os-boot/linux --jobs=$(nproc)
+
+      - name: Cache Linux image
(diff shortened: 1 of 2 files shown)


pmundkur:
Might be worth mentioning in the changelog.

jordancarlin:
> Might be worth mentioning in the changelog.

I think the changelog should primarily be for user facing changes, not developer/CI specific things.

pmundkur:
I do think users would benefit from knowing the testing done for model.  The changelog already mentions the switching to upstream `riscv-tests` and `riscv-vector-tests`.

.github/workflows/ci.yml
```diff
@@ -57,12 +57,57 @@ jobs:
...
+      - name: Cache Linux image
+        if: steps.linux-cache-restore.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: os-boot/linux/build/fw_payload.elf
+          key: ${{ steps.linux-cache-restore.outputs.cache-primary-key }}
+
+      - name: Linux boot
+        run: |
+          make -C os-boot/linux build/sail.dtb
```
Timmmm, line 100: Couldn't you do `make -C os-boot/linux sail 2>&1 | tee linux-boot.log`?

jordancarlin: The dependency chain in the Makefile prevents caching the compiled Linux elf.

Timmmm: Ah I see, that's unfortunate. Can you add a comment explaining that? Otherwise LGTM.

Timmmm (approved):

.github/workflows/ci.yml
```diff
@@ -57,12 +57,57 @@ jobs:
...
+          ./build/c_emulator/sail_riscv_sim --show-times --inst-limit 20000000 --device-tree-blob os-boot/linux/build/sail.dtb os-boot/linux/build/fw_payload.elf 2>&1 | tee linux-boot.log
```
arichardson, line 102: Suggestion for a follow-up improvement: We could use a python pexpect (or tcl expect) script to drive the interaction with the linux boot to have some control over timeouts (e.g. only wait N seconds until we see the first print and report a failure if not)

.github/workflows/ci.yml
jordancarlin: Updated

.github/workflows/ci.yml
jordancarlin: Definitely lots that we can do to improve this, especially with regards to benchmarking. I figured it was worth at least getting Linux booting in CI and we can add things from there.

Timmmm:
One other improvement we could do is disabling PMP for this since it has some really bad performance issues at the moment. But we can do that later. Nice work @jordancarlin !

.github/workflows/ci.yml
arichardson: Agreed, there is absolutely no need to do this now!

PR #1310, opened 2025-10-01
Use private keyword in virtual memory code

Alasdair:
The virtual memory code had `// PUBLIC:` and `// PRIVATE:` comments describing the intended usage of various identifiers. This can now be enforced using the Sail `private` keyword.

I left some `// PUBLIC:` comments which described how and where certain things were intended to be used.

Sail currently doesn't like `private bitfield ...`. I think due to how bitfields generate various overloaded auxilliary definitions, so I omit that for now.

`private` declarations have also been added to other modules, and the CODE_STYLE has been updated with a suggestion to encourage its use.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,9 @@ automated verification that is presented in a way that the reviewers
 of the pull request can run. It is desirable that the pull request
 explains how it relates to the existing RISC-V architectural tests.
 
+When contributing a bug fix, it is often helpful to point to or quote
+the pertinent portion of the spec in the PR or the code when relevant.
+
 We recommend installing pre-commit hooks that ensure certain basic coding
 style issues can be detected and fixed before submitting the pull request.
 To set up these hooks, install [pre-commit](https://pre-commit.com)
--- a/model/extensions/V/vext_vset_insts.sail
+++ b/model/extensions/V/vext_vset_insts.sail
@@ -49,17 +49,23 @@ mapping vtype_assembly : string <-> (bits(1), bits(1), bits(3), bits(3)) = {
   hex_bits_8(ma @ ta @ sew @ lmul)  <-> (ma, ta, sew, lmul),
 }
 
-private function handle_illegal_vtype() -> unit = {
+private function handle_illegal_vtype(rd : regidx) -> unit = {
   // Note: Implementations can set vill or trap with an illegal
   // instruction if the vtype setting is not supported.
   // TODO: configuration support for both solutions.
-  // NOTE: vstart should NOT be zeroed out if trapping.
+  // NOTE: vstart should NOT be zeroed out and rd
+  //       should NOT be written if trapping.
 
   vtype.bits = 0b1 @ zeros(xlen - 1); // set vtype.vill
   vl = zeros();
   csr_write_callback("vtype", vtype.bits);
   csr_write_callback("vl", vl);
   set_vstart(zeros());
+
+  // "The vset{i}vl{i} instructions set the vtype and vl CSRs based on their arguments,
+  //  and write the new value of vl into rd."
+  // So `rd` should be written even if the arguments were illegal or reserved.
+  X(rd) = vl;
 }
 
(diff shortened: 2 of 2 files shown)


Timmmm:
Nice. So this basically means only files in the `model/sys` directory/module can use these? Probably worth documenting in the Sail manual how this works.

Also how come in the grammar in the Sail manual it's listed as `Private` but here it's `private`?

model/sys/vmem_ptw.sail
```diff
@@ -37,15 +35,13 @@ function ptw_error_to_str(e : PTW_Error) -> string = {
   }
 }
 
-// PUBLIC
 overload to_str = {ptw_error_to_str}
```
arichardson, line 38: This one seems like it was originally wrong? `union PTW_Error` is private and `ptw_error_to_str` is also private so

model/sys/vmem.sail
```diff
@@ -259,7 +248,7 @@ function translate_TLB_hit forall 'v, is_sv_mode('v) . (
   }
 }
 
-// PRIVATE: translate on TLB miss (do a page-table walk)
+// Translate on TLB miss (do a page-table walk)
 function translate_TLB_miss forall 'v, is_sv_mode('v) . (
```
jordancarlin: Looks like private was dropped here.

pmundkur:
The Lean build broke.  It looks like this needs a compiler fix to allow private functions in termination measures?

model/sys/vmem_ptw.sail
Alasdair: If you add a private definition to an overload, the overload will only resolve to the private function inside the module. I think that is the intent here?

Alasdair:
Not sure they should be. The actual fix might be to move the offending termination measures into the correct module so they can access what they need.

Alasdair:
> Also how come in the grammar in the Sail manual it's listed as Private but here it's private?

There's a sed script that turns the internal token names (which are OCaml constructors, so must be capitalised) into the strings that generate them. That script is probably missing the private token.

jordancarlin:
@Alasdair Now that #1376 is merged, it seems like this should be possible without breaking the theorem provers.

pmundkur:
I rebased this and extended the usage of `private` to some other modules.  I used it only for file-scope privacy for now, leaving module-scope privacy for later.

CODE_STYLE.md
```diff
@@ -112,3 +112,6 @@ corresponding elements vertically aligned.
 - Do not use the `ext*` types and hooks for standard extensions unless
   providing a stub implementation; these are reserved for use by out-of-tree
   extensions that provide their own non-stub implementations
+
+- Helper functions that are local to a file or module should be marked
+  with the `private` keyword.
```
jordancarlin: Enums/unions/etc. too.

jordancarlin (approved):

nadime15:
Did we get an answer to that?

> Nice. So this basically means only files in the model/sys directory/module can use these? Probably worth documenting in the Sail manual how this works.

I would like to know how this works.

pmundkur:
> Did we get an answer to that?
> 
> > Nice. So this basically means only files in the model/sys directory/module can use these? Probably worth documenting in the Sail manual how this works.
> 
> I would like to know how this works.

It makes a type or function private to the module in which the definition appears.  For a function with a `val` declaration, `private` should be put in the `val`. It doesn't seem to work for bitfields.

If the function or type is used from another module, you get an error like:
```
Type error:
extensions/vector_crypto/zvbb_insts.sail:69.27-37:
69 |  let rs1_val : bits('m) = get_scalar(rs1, SEW);
   |                           ^--------^
   | Cannot use private definition
   | 
   | extensions/V/vext_utils_insts.sail:202.12-22:
   | 202 |private val get_scalar : forall 'm, 'm >= 8. (regidx, int('m)) -> bits('m)
   |     |            ^--------^ private definition here in V_instructions

```
where `get_scalar` in module `V_instructions` was declared `private` but used in module `Zvbb`.

PR #1312, opened 2025-10-02
Update docs and diagram to mention the configuration schema.

pmundkur:


doc/figs/riscvcsimdeps.svg
jordancarlin, line 1: Should the `model configuration` have an arrow going into `C translation` instead of coming out of it?

doc/ReadingGuide.md
```diff
@@ -276,3 +278,12 @@ The Sail run-time system provides a C library for use with its C backend, which
 provides the low-level details of the implementation of raw memory and
 bitvectors (typically optimized to use the native machine word
 representation).
+
+The model is configured by a file in JSON format that specifies
+various ISA options (e.g. RV32 or RV64, number of PMP entries), the
+memory map and whether various extensions are enabled. Users of the
+model can provide their own configuration file based on the [sample
+template](../config/config.json.in). The Sail compiler generates a
```
jordancarlin: Might be worth pointing people at the generated configs as the thing to base their configs off of instead of the CMake template.

pmundkur: I added a mention of those, but kept the CMake template as well since we can actually point to it from the docs.

doc/figs/riscvcsimdeps.svg
pmundkur: The C translation depends on the model config (there are calls to `sail_config_get()` in the generated C).  This direction keeps it consistent with the line in the doc that says "The black arrows indicate dependency relationships", and also since the config does not depend on the generated C model.

PR #1315, opened 2025-10-03
Simplify the exit of the `run_sail()` function.

pmundkur:
This removes a legacy unused `diverged` flag and `goto` from `run_sail()`.

Also simplify the `finish()` function called from `run_sail()` on termination.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -359,12 +359,13 @@ void close_logs(void)
 
 void finish()
 {
-  // `model_fini()` exits with failure if there was a Sail exception.
-  model_fini();
-
-  if (!sig_file.empty()) {
+  // Don't write a signature if there was an internal Sail exception.
+  if (!have_exception && !sig_file.empty()) {
     write_signature(sig_file.c_str());
   }
+
+  // `model_fini()` exits with failure if there was a Sail exception.
+  model_fini();
 
   if (do_show_times) {
     if (gettimeofday(&run_end, NULL) < 0) {


c_emulator/riscv_sim.cpp
```diff
@@ -357,18 +357,20 @@ void close_logs(void)
...
 
-void finish(int ec)
+void finish()
 {
+  // `model_fini()` exits with failure if there was a Sail exception.
+  model_fini();
+
   if (!sig_file.empty()) {
     write_signature(sig_file.c_str());
   }
```
nadime15: Does it make a difference if model_fini() is called before write_signature(), given that the call order was changed?

pmundkur: Yes.  It seems to be pointless (and likely misleading) to write a test signature when the Sail model had an exception.  Test signatures should be written only for runs that don't have any Sail exceptions.

Timmmm: I think @nadime15 meant that you now call `model_fini()` before `write_signature()` even on success, which maybe isn't guaranteed to work. `model_fini()` frees model variables that might be used by `write_signature()`.

I think it *should* be ok though. `write_signature()` only uses `read_mem` and I'm pretty sure `model_fini()` doesn't touch memory.

c_emulator/riscv_sim.cpp
```diff
@@ -357,18 +357,20 @@ void close_logs(void)
...
 }
 
-void finish(int ec)
+void finish()
 {
+  // `model_fini()` exits with failure if there was a Sail exception.
```
Timmmm: That is weird and it probably shouldn't!

c_emulator/riscv_sim.cpp
pmundkur: Good point, I'll change it to not rely on `model_fini()` not freeing memory.

jordancarlin (approved):

PR #1321, opened 2025-10-08
Some fixes for documentation inclusion of scalar and vector crypto functions.

pmundkur:
- Attach documentation comments to functions used in the Sail supplements of the crypto specs.  Use anchors for other documentation comments.

- Move `val` declarations into function signatures to make documentation inclusion of functions more standalone and comprehensible.

- Fold some long lines.

The doc comments will need the next Sail compiler release to show up in the doc bundle, but this doesn't break compilation with the current version.

### DIFF
--- a/model/extensions/K/types_kext.sail
+++ b/model/extensions/K/types_kext.sail
@@ -16,39 +16,33 @@
 // Cryptography extension shared / utility functions
 // ----------------------------------------------------------------------
 
-// Auxiliary function for performing GF multiplicaiton
-val xt2 : bits(8) -> bits(8)
-function xt2(x) = {
+/// Auxiliary function for performing GF multiplication.
+function xt2(x : bits(8)) -> bits(8) = {
   (x << 1) ^ (if bit_to_bool(x[7]) then 0x1b else 0x00)
 }
 
-val xt3 : bits(8) -> bits(8)
-function xt3(x) = x ^ xt2(x)
+function xt3(x : bits(8)) -> bits(8) = x ^ xt2(x)
 
-// Multiply 8-bit field element by 4-bit value for AES MixCols step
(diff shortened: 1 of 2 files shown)


model/extensions/K/types_kext.sail
```diff
@@ -168,53 +161,46 @@ let aes_sbox_inv_table : vector(256, bits(8)) = [
...
-// Lookup function - takes an index and a table, and retrieves the
-// x'th element of that table. Note that the Sail vector literals
-// start at index 255, and go down to 0.
-val sbox_lookup : (bits(8), vector(256, bits(8))) -> bits(8)
-function sbox_lookup(x, table) = {
+/// Lookup function - takes an index and a table, and retrieves the
+/// x'th element of that table. Note that the Sail vector literals
+/// start at index 255, and go down to 0.
+function sbox_lookup(x : bits(8), table : vector(256, bits(8))) -> bits(8) = {
   table[255 - unsigned(x)]
```
Timmmm, line 168: Btw @Alasdair I wonder if there's a way to make this less confusing. This is only here so the author didn't have to write the sbox table backwards (big endian)... because it's a little confusing that array literals are bit endian.

I wonder if array literals should be little endian if they use the `[]` syntax irrespective of the global order?

Timmmm (approved):
Much nicer. What does the anchor do exactly?

pmundkur:
> Much nicer. What does the anchor do exactly?

It allows you to pull in unattached comments, like this [one](https://github.com/riscv/sail-riscv/blob/0e96ba6ea043a167d91c17add49337d9d4651d5d/model/extensions/K/types_kext.sail#L242)

That comment is also in the Sail supplement of the crypto specs.

jordancarlin (approved):

PR #1323, opened 2025-10-09
Remove monomorphic `val` declarations by moving types into the function signature.

pmundkur:
This simplifies including these functions into documentation since the `val` does not have to be included as well for completeness.

The code is often more readable as well.


Timmmm:
Nice, I've been meaning to do this for ages.

PR #1324, opened 2025-10-09
Enable Dependabot to automatically bump GitHub Actions 3rd party action versions

jordancarlin:


Timmmm:
What does this do exactly?

jordancarlin:
Dependabot is a GitHub feature to automatically create PRs to bump dependencies. It works with lots of different ecosystems, including GitHub Actions. See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates and https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot for more details.

Many other riscv repos use it to keep actions and submodules up to date.

PR #1325, opened 2025-10-09
Don't save caches when running in the merge queue

jordancarlin:
We have a lot of cache thrashing happening right now (https://github.com/riscv/sail-riscv/actions/caches) because each branch saves its own cache. There is no resaon to save caches from merge queue branches that will only ever run CI once.


jordancarlin:
> Probably worth adding a comment to explain this expression because it's quite complex now!

Added.

PR #1330, opened 2025-10-10
Simplify compressed changelog generation

Timmmm:
Create the compressed archive at configure time, which means you don't have to remember to run `make compressed_changelog` before running `cpack`. This also removes a dependency on the `gzip` executable which probably won't be available on Windows.

`gzip` was previously used to get a 0 timestamp embedded in the gzip file, which Debian requires. However we can also do that with `export SOURCE_DATE_EPOCH=0`.

I added `SOURCE_DATE_EPOCH` to the Github release workflow. We don't actually generate a Debian package there yet but it's probably a good idea anyway (some other tools respect it).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,6 +38,10 @@ jobs:
 
     env:
       SAIL_VERSION: "0.19.1"
+      # This helps with reproducible builds by not embedding variable timestamps in files.
+      # This is needed in particular for the Debian package's compressed changelog.
+      # See the comments in `cmake/packaging.cmake`.
+      SOURCE_DATE_EPOCH: "0"
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
jordancarlin, line 1: Does adding `set(ENV{SOURCE_DATE_EPOCH} 0)` to the CMake file work to accomplish the same thing?

Timmmm: Presumably but I wasn't sure if we wanted to do that always.

jordancarlin: The old approach always did this and didn't seem to cause issues, so I'd probably be in favor of doing it this way for consistency. I'm ok with either approach though.

PR #1331, opened 2025-10-13
Simplify sew/lmul_pow_val

Timmmm:
Remove the `sew_pow_val` and `lmul_pow_val` mappings and just use `unsigned()`/`signed()` instead. This does require explicit assertions for the reserved values rather than implicit pattern match failures but I think it's a bit nicer anyway.

### DIFF
--- a/model/extensions/V/vext_regs.sail
+++ b/model/extensions/V/vext_regs.sail
@@ -244,25 +244,7 @@ bitfield Vtype  : xlenbits = {
 register vtype : Vtype
 
 type SEW_pow = range(3, 6)
-
-mapping sew_pow_val : bits(3) <-> SEW_pow = {
-  0b000 <-> 3,
-  0b001 <-> 4,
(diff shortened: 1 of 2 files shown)


Timmmm:
But the mapping is just normal unsigned/signed number encoding. I don't think we need that.

jordancarlin:
> I'm not sure about this. I think it might be better from a documentation point of view to retain the mappings.

I think because the mapping is just a normal signed/unsigned conversion, it actually makes it less clear. Having a mapping makes it seem like there is something special/unique, but in this case there is not.

pmundkur:
> > I'm not sure about this. I think it might be better from a documentation point of view to retain the mappings.
> 
> I think because the mapping is just a normal signed/unsigned conversion, it actually makes it less clear. Having a mapping makes it seem like there is something special/unique, but in this case there is not.

Ok, makes sense.

PR #1332, opened 2025-10-13
Improve rsqrt7 code

Timmmm:
1. Rename `sub` (which could be interpreted as "subtract") to `is_subnormal`.
2. Rename `table` which is very vague to `sig_out_lut`.
3. Use Sail's `float_decompose` to make the code generic over float size.
4. Fix use of `to_bits_unsafe`
5. Improve types and remove unnecessary assertion.


Timmmm:
This doesn't work yet due to [this issue](https://github.com/rems-project/sail/issues/748).

Timmmm:
This should work now. The `recip7` code needs similar changes but it was a little more involved so I will do it later.

Timmmm:
Ok maybe not!

PR #1333, opened 2025-10-13
Simplify trace_log global variable

Timmmm:
Default to `stdout` and avoid doing an assignment inside an if-expression.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -48,7 +48,7 @@ bool do_print_isa = false;
 std::string config_file;
 std::string term_log;
 std::string trace_log_path;
-FILE *trace_log = nullptr;
+FILE *trace_log = stdout;
 std::string dtb_file;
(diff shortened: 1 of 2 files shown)


c_emulator/riscv_sim.cpp
```diff
@@ -485,12 +485,13 @@ void init_logs()
...
 
-  if (trace_log_path.empty()) {
-    trace_log = stdout;
-  } else if ((trace_log = fopen(trace_log_path.c_str(), "w+")) == NULL) {
-    fprintf(stderr, "Cannot create trace log '%s': %s\n",
-            trace_log_path.c_str(), strerror(errno));
-    exit(EXIT_FAILURE);
+  if (!trace_log_path.empty()) {
+    trace_log = fopen(trace_log_path.c_str(), "w+");
+    if (trace_log == NULL) {
```
nadime15: In regard to #1334
```suggestion
    if (trace_log == nullptr) {
```

PR #1337, opened 2025-10-14
Add wavedrom annotations for some instruction encodings.

pmundkur:
These have been added only to instructions that have their own dedicated diagram in the manual.  Instructions that are grouped into a single diagram will need more tooling support from the asciidoc backend.

In some cases, bitvector literals in encodings have been split or joined to better match the spec diagram.

### DIFF
--- a/model/extensions/I/base_insts.sail
+++ b/model/extensions/I/base_insts.sail
@@ -69,8 +69,9 @@ function jump_to(target : xlenbits) -> ExecutionResult = {
 // *****************************************************************
 union clause instruction = JAL : (bits(21), regidx)
 
-mapping clause encdec = JAL(imm_19 @ imm_7_0 @ imm_8 @ imm_18_13 @ imm_12_9 @ 0b0, rd)
-  <-> imm_19 : bits(1) @ imm_18_13 : bits(6) @ imm_12_9 : bits(4) @ imm_8 : bits(1) @ imm_7_0 : bits(8) @ encdec_reg(rd) @ 0b1101111
+$[wavedrom "_ offset[20:1] _ _ dest JAL"]
+mapping clause encdec = JAL(imm_19 @ imm_7_0 @ imm_8 @ imm_18_9 @ 0b0, rd)
+  <-> imm_19 : bits(1) @ imm_18_9 : bits(10) @ imm_8 : bits(1) @ imm_7_0 : bits(8) @ encdec_reg(rd) @ 0b1101111
 
 function clause execute JAL(imm, rd) = {
   let link_address = get_next_pc();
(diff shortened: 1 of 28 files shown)


model/extensions/I/base_insts.sail
```diff
@@ -69,8 +69,9 @@ function jump_to(target : xlenbits) -> ExecutionResult = {
 // *****************************************************************
 union clause instruction = JAL : (bits(21), regidx)
 
-mapping clause encdec = JAL(imm_19 @ imm_7_0 @ imm_8 @ imm_18_13 @ imm_12_9 @ 0b0, rd)
-  <-> imm_19 : bits(1) @ imm_18_13 : bits(6) @ imm_12_9 : bits(4) @ imm_8 : bits(1) @ imm_7_0 : bits(8) @ encdec_reg(rd) @ 0b1101111
+$[wavedrom "_ offset[20:1] _ _ dest JAL"]
+mapping clause encdec = JAL(imm_19 @ imm_7_0 @ imm_8 @ imm_18_9 @ 0b0, rd)
+  <-> imm_19 : bits(1) @ imm_18_9 : bits(10) @ imm_8 : bits(1) @ imm_7_0 : bits(8) @ encdec_reg(rd) @ 0b1101111
```
Alasdair, line 74: For JAL you can get really close to the ISA manual with something like:
```sail
mapping clause encdec = JAL(imm @ 0b0, rd)
  <-> imm[19] @ imm[9..0] @ imm[10] @ imm[18..11] @ encdec_reg(rd) @ 0b1101111
```
Unfortunately the manual considers `imm` to include the extra zero in addition to the 20 bits in the encoding so it goes from `imm[20]` to `imm[1]`, and Sail bitvectors are always 0-indexed so it has to specified from 19 to 0. I considered something like:
```sail
mapping clause encdec = JAL((_ @ 0b0) as imm, rd)
  <-> imm[20] @ imm[10..1] @ imm[11] @ imm[19..12] @ encdec_reg(rd) @ 0b1101111
```
which would let you generate the exact diagram in the manual from Sail, but it seemed over-complicated for basically just JAL at the time.

Timmmm: Yeah I guess there's also stuff like [compressed load/stores](https://riscv-isa-manual.timhutt.co.uk/riscv-unprivileged.html#_register_based_loads_and_stores) and [c.addi4spn](https://riscv-isa-manual.timhutt.co.uk/riscv-unprivileged.html#_integer_register_immediate_operations) that have implicit zeros in their immediates.

I think it's probably not worth complicating the Sail for this though.

```
mapping clause encdec = JAL(imm @ 0b0, rd)
  <-> imm[19] @ imm[9..0] @ imm[10] @ imm[18..11] @ encdec_reg(rd) @ 0b1101111
```

We should probably update the encdec clauses to this at least - it's easier to understand, and we have quite a few `encdec` clauses that use `imm_N_M` instead. I made #1339 to track this.

PR #1342, opened 2025-10-14
add base case to currentlyEnabled

jn80842:
currentlyEnabled should return false for any extensions that are not included in the current build.

### DIFF
--- a/model/postlude/insts_end.sail
+++ b/model/postlude/insts_end.sail
@@ -26,6 +26,8 @@ mapping clause assembly = C_ILLEGAL(s) <-> "c.illegal" ^ spc() ^ hex_bits_16(s)
 
 // *****************************************************************
 
+function clause currentlyEnabled _ = false
+
 // End definitions
 end extension
 end extensionName


pmundkur (commented):
I'm assuming this is driven by a pruned model.  Wouldn't there be a similar case for `hartSupports`?

Timmmm:
Has anyone actually managed to compile a pruned model yet? I thought there were some issues with the termination measure or something? Maybe we should hold off on this until we know how it's going to work.

jn80842:
If by a pruned model you mean a model compiled with only a subset of modules, then yes, we're working with one right now. Running `cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="postlude;Zicbom_insts;Zicboz;Zicsr_insts;Zkr" && cmake --build build --target generated_lean_executable_rv64d` we get a model that compiles successfully (we're still debugging the executable, but I believe the bugs we're seeing now aren't related to the module system).

Timmmm:
Ah nice! We should probably add one or two examples to CI.

PR #1345, opened 2025-10-15
Add Linux arm build to CI and release

jordancarlin:
Now that there is a Linux Arm build of the Sail compiler, we should release one of the model too. This helps when running a Linux docker image on an Arm (Apple Silicon) Mac.

### DIFF
--- a/.github/actions/sail-setup/action.yml
+++ b/.github/actions/sail-setup/action.yml
@@ -29,7 +29,12 @@ runs:
     - name: Install CMake
       shell: bash
       run: |
-        curl --location https://github.com/Kitware/CMake/releases/download/v${{ inputs.cmake-version }}/cmake-${{ inputs.cmake-version }}-${{ runner.os }}-$(arch).tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
+        if [ "${{ runner.os }}" = "Linux" ]; then
+          cmake_suffix=$(arch)
+        elif [ "${{ runner.os }}" = "macOS" ]; then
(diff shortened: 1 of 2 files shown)


pmundkur:
The cache in `sail-setup` would now need to be indexed on `${{ runner.arch }}` along with `${{ runner.os }}`

jordancarlin:
> The cache in `sail-setup` would now need to be indexed on `${{ runner.arch }}` along with `${{ runner.os }}`

Good catch. More problematically, it seems like the setup-cmake action is not smart enough to download an arm binary of CMake:
```
/home/runner/work/_temp/6863df02-698f-493d-8b1e-1a34a2fce50a.sh: line 3: /opt/hostedtoolcache/cmake/4.1.2/arm64/cmake-4.1.2-linux-x86_64/bin/cmake: cannot execute binary file: Exec format error
```

We'll need to switch to manually installing the desired version of CMake instead.

Timmmm:
Could we fix that action? Might not be too hard.

jordancarlin:
> Could we fix that action? Might not be too hard.

There has been nothing merged to that repo in over a year and there are 2 existing attempts to add support that both ran into issues (https://github.com/jwlawson/actions-setup-cmake/pull/74 and https://github.com/jwlawson/actions-setup-cmake/pull/78).

I went ahead and switched to a simple curl download instead. The only downside is that we have to manually maintain what `latest` means now.

Timmmm:
Ok that seems like a better option.

PR #1346, opened 2025-10-15
Attest release artifacts.

pmundkur:


.github/workflows/release.yml
```diff
@@ -85,6 +89,22 @@ jobs:
...
+            build/sail-riscv-*
+            build/model/sail_riscv_model.json
+            build/model/sail_riscv_html.tgz
+
+      - name: "Generate attestations (Linux-aarch64)"
+        if: matrix.name == 'Linux-aarch64'
+        uses: actions/attest-build-provenance@v3
+        with:
+          subject-path: |
+            build/sail-riscv-*
```
jordancarlin, line 112: We generate all of the artifacts for all architectures right now, so we could simplify this by unconditionally attesting all of the artifacts. We then only download the desired subset in the next phase.

Alternatively we should update the workflow to only compile doc bundles for x86.

pmundkur: Let me try if the verification would work in that case.

pmundkur: It works, but it makes sense to only build what will be used for the release so the docs are now only built on x86_64 in the release workflow.

jordancarlin: Great. Meant to do that in #1345.

jordancarlin (approved):

PR #1347, opened 2025-10-15
Increment version and update changelog.

pmundkur:
Add a mention of the ARM binary and release attestation.


pmundkur:
Depends on #1346

jordancarlin (approved):
You could do this is the same PR that bumps the version number, but either way is fine.

PR #1348, opened 2025-10-16
Simplify vector_support_level

### LINKED ISSUE #1344: Simplify vector_support_level
Currently we have this:

```
enum vector_support = {
  Disabled,
  Integer,
  Float_single,
  Float_double,
  Full,
}

mapping vector_support_level : vector_support <-> nat = {
  Disabled     <-> 0,
  Integer      <-> 1,
  Float_single <-> 2,
  Float_double <-> 3,
  Full         <-> 4,
}

let vector_support_config = config extensions.V.support_level : vector_support
let vector_support_config_level = vector_support_level(vector_support_config)
```

And it's used like this:

```
if vector_support_config_level >= vector_support_level(Integer)
```

But instead you can just do this:

```
// These must be in increasing order!
enum vector_support = {
  Disabled,
  Integer,
  Float_single,
  Float_double,
  Full,
}

let vector_support_level : vector_support = config extensions.V.support_level

function vector_support_lt(x : vector_support, y : vector_support) -> bool = num_of_vector_support(x) < num_of_vector_support(y)
function vector_support_gt(x : vector_support, y : vector_support) -> bool = num_of_vector_support(x) > num_of_vector_support(y)
function vector_support_le(x : vector_support, y : vector_support) -> bool = num_of_vector_support(x) <= num_of_vector_support(y)
function vector_support_ge(x : vector_support, y : vector_support) -> bool = num_of_vector_support(x) >= num_of_vector_support(y)

overload operator < = {vector_support_lt}
overload operator > = {vector_support_gt}
overload operator <= = {vector_support_le}
overload operator >= = {vector_support_ge}
```

And then use it like this:

```
if vector_support_level >= Integer
```

which is a lot nicer. We could leave the explicit mapping if necessary but I think using `num_of_` and adding a comment that the order matters is probably fine.

KotorinMinami:
Instead of mapping to integers, use `num_of_` and define comparison operators for the enum directly.

Fixes #1344

### DIFF
--- a/model/core/vlen.sail
+++ b/model/core/vlen.sail
@@ -6,6 +6,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 // =======================================================================================
 
+// The order of vector_support is significant - each level implies all of the previous levels.
 enum vector_support = {
   Disabled,
   Integer,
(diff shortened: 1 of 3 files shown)


model/core/vlen.sail
```diff
@@ -14,16 +14,17 @@ enum vector_support = {
...
 }
 
-mapping vector_support_level : vector_support <-> nat = {
-  Disabled     <-> 0,
-  Integer      <-> 1,
-  Float_single <-> 2,
-  Float_double <-> 3,
-  Full         <-> 4,
-}
+let vector_support_config_level = config extensions.V.support_level : vector_support
```
Timmmm: I think it's clearer if we put the type annotation on the variable, also we could shorten the name.

```
let vector_support_level : vector_support = config extensions.V.support_level
```

Otherwise this looks great!

Timmmm: And rename it to `vector_support_level`? Might be worth adding a comment to the definition of `vector_support` to say that the order is significant - each level implies all of the previous levels.

Timmmm (approved):
Awesome, thanks!

PR #1349, opened 2025-10-16
Use let instead of register for plat_clint config variables

### LINKED ISSUE #1270: Replace `register` with `let` for global `to_bits_checked()`
```
// Use `register` instead of `let` since `to_bits_checked` is not a pure initializer
// due to its use of `assert`.
register plat_rom_base : physaddrbits = to_bits_checked(config platform.rom.base : int)
register plat_rom_size : physaddrbits = to_bits_checked(config platform.rom.size : int)
```

I believe this was fixed in the Sail compiler so we can change them to `let` (although some will be removed anyway by #866 ).

jordancarlin:
Fixes #1270 now that the Sail compiler version has been updated.

### DIFF
--- a/model/sys/platform.sail
+++ b/model/sys/platform.sail
@@ -26,10 +26,8 @@ let plat_enable_dirty_update : bool = config memory.translation.dirty_update
 let plat_enable_misaligned_access : bool = config memory.misaligned.supported
 
 // Location of clock-interface, which should match with the spec in the DTB
-// Use `register` instead of `let` since `to_bits_checked` is not a pure initializer
-// due to its use of `assert`.
-register plat_clint_base : physaddrbits = to_bits_checked(config platform.clint.base : int)
-register plat_clint_size : physaddrbits = to_bits_checked(config platform.clint.size : int)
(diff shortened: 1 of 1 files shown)


jordancarlin:
Looks like it works for the C backend now, but is still broken for Lean:
```
error: LeanRV64D/Platform.lean:188:39: invalid use of `(<- ...)`, must be nested inside a 'do' expression
  error: LeanRV64D/Platform.lean:190:39: invalid use of `(<- ...)`, must be nested inside a 'do' expression
  error: LeanRV64D/Platform.lean:695:14: Type mismatch
    within_clint addr width
  has type
    Bool
  but is expected to have type
    SailM ?m.124590
  error: LeanRV64D/Platform.lean:702:14: Type mismatch
    within_clint addr width
  has type
    Bool
  but is expected to have type
    SailM ?m.125110
  error: LeanRV64D/Platform.lean:706:9: Type mismatch
    within_clint paddr width
  has type
    Bool
  but is expected to have type
    SailM ?m.125544
  error: LeanRV64D/Platform.lean:720:9: Type mismatch
    within_clint paddr width
  has type
    Bool
  but is expected to have type
    SailM ?m.126322
```

Issue #545, opened 2024-09-09
Towards a defined log/trace output from the Sail model


rsnikhil:
This first comment just describes the issue(s).  Follow-up comments
discuss possible solutions.

Currently the Sail model writes out a log file in ASCII format.  Issues:
* Logs can become very large (a binary format would be much smaller)

* Downstream tools (verification, trace-driven simulation, ...) have
    to parse ASCII text to recreate the log data

* We haven't been precise about defining a log format, and we update
    it as needed in an ad hoc fashion. As a result downstream tools
    break when we make such changes.

rsnikhil:
I have looked at 4 public RISC-V specifications for "trace/log", and
none of them seem suitable for a full trace output from the Sail
model.

I recommend the Sail model use a 5th alternative, "Bluespec trace",
which I will discuss in a follow-up comment.

1. RVI E-Trace ("Efficient Trace")
    https://github.com/riscv-non-isa/riscv-trace-spec/releases/download/v2.0.3/riscv-trace-spec-asciidoc.pdf
    From RVI; ratified

2. RVI N-Trace ("Nexus Trace") https://github.com/riscv/tg-nexus-trace
    From RVI; ratified (nor nearing ratification)

3. RVFI ("RISC-V Formal Verification Interface")
    https://github.com/SymbioticEDA/riscv-formal From SymbioticEDA

4. RVVI ("RISC-V Verification Interface")
    https://github.com/riscv-verification/RVVI From Imperas originally
    (now part of Synopsys), but now public.  Generated by the
    Imperas/Synopsys simulator.  Used by OpenHWGroup.

E-Trace and N-Trace:

* focus on compressed control-flow trace (recording only
    "unpredictable" branch/jump/trap/xRET)

* rely on access to the original ELF file in order to reconstruct the
    full control trace

* are not suitable for self-modifying code/JIT where there is no ELF file

* do not capture arch-state updates (GPRs, FPRs, CSRs).  E-Trace has a
     section for capturing memory updates.

RVFI and RVVI are similar (RVVI's README calls itself a kind of
superset of RVFI).  They specify (nearly) all architectural state
updates that we'd need in a log, but they are not very suitable for
recording in a file, because it is *VERY* wide (e.g., all 32 GPR values,
all 32 FPR values, all defined CSR values, ...).

rsnikhil:
Bluespec, Inc. has defined a trace model (PDF attached) which could be used by the Sail model.

[...]

PR #1354, opened 2025-10-21
Add more state change callbacks

### LINKED ISSUE #1352: Add trap details to trap callback
The callback interface has `trap_callback()` but it doesn't say:

1. Was it an interrupt or exception.
2. What was the exception code.
3. Which privilege mode is it trapping to.

That would be really useful.

Timmmm:
This adds a few state change callbacks that we needed for an internal binary logging format. They aren't used by the RVFI or log callback handlers but they should also be useful for the open source binary logging format that we eventually add (#545).

* Add a `fetch_callback` that says which instruction was executed (if any).
* Add `redirect_callback` that is called for taken and unconditional branches.
* Add `is_interrupt` and `cause` to `trap_callback`.

I also renamed `value` to `new_pc` to be a bit more explicit.

Fixes #1352

### DIFF
--- a/model/core/callbacks.sail
+++ b/model/core/callbacks.sail
@@ -7,6 +7,12 @@
 // =======================================================================================
 
 // Callbacks for state-changing events
+
+/// Called upon successful instruction fetch with the opcode.
+/// If there is a fetch error then this is not called (but trap_callback() will be).
+val fetch_callback = pure {c: "fetch_callback"} : forall 'n, 'n in {16, 32}. (bits('n)) -> unit
+function fetch_callback(_) = ()
+
 val mem_write_callback = pure {c: "mem_write_callback"} : forall 'n, 0 < 'n <= max_mem_access . (/* access type */ string, /* addr */ physaddrbits, /* width */ int('n), /* value */ bits(8 * 'n)) -> unit
 function mem_write_callback(_) = ()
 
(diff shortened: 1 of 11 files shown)


model/sys/sys_control.sail
```diff
@@ -154,7 +154,10 @@ function track_trap(p : Privilege) -> unit = {
 // handle exceptional ctl flow by updating nextPC and operating privilege
 function trap_handler(del_priv : Privilege, c : TrapCause, pc : xlenbits, info : option(xlenbits), ext : option(ext_exception))
                      -> xlenbits = {
-  trap_callback();
+  let is_interrupt = trapCause_is_interrupt(c);
+  let cause        = trapCause_bits(c);
+
+  trap_callback(is_interrupt, trapCause_bits(c));
```
jordancarlin:
```suggestion
  let cause        = trapCause_bits(c);

  trap_callback(is_interrupt, cause);
```

model/core/callbacks.sail
```diff
@@ -7,6 +7,12 @@
 // =======================================================================================
 
 // Callbacks for state-changing events
+
+// Called upon successful instruction fetch with the opcode.
+// If there is a fetch error then this is not called (but trap_callback() will be).
```
pmundkur: Perhaps we should start using the doc-comment '///' prefix when adding these kinds of comments in case we want to extract them to documentation later (here and below in this file)?   We could fix existing uncommented functions or comments without that prefix in a different PR later when we actually need to extract them.  Just a thought, feel free to ignore.

Timmmm: Yeah I'm not sure about this because it's not clear to me what we would use these kind of doc comments for... but I guess it doesn't hurt so I've updated them to be doc comments.

PR #1357, opened 2025-10-22
Change is_CSR_accessible return type to CSRAccessResult enum

trdthg:


model/core/sys_regs.sail
```diff
@@ -856,9 +856,9 @@ mapping clause csr_name_map = 0x140  <-> "sscratch"
 mapping clause csr_name_map = 0x142  <-> "scause"
 mapping clause csr_name_map = 0x143  <-> "stval"
 
-function clause is_CSR_accessible(0x140, _, _) = currentlyEnabled(Ext_S) // sscratch
-function clause is_CSR_accessible(0x142, _, _) = currentlyEnabled(Ext_S) // scause
-function clause is_CSR_accessible(0x143, _, _) = currentlyEnabled(Ext_S) // stval
+function clause is_CSR_accessible(0x140, _, _) = if currentlyEnabled(Ext_S) then CSRAccessSuccess else CSRNotExist // sscratch
+function clause is_CSR_accessible(0x142, _, _) = if currentlyEnabled(Ext_S) then CSRAccessSuccess else CSRNotExist // scause
```
arichardson:
```suggestion
function clause is_CSR_accessible(0x142, _, _) = csrDefinedBy(Ext_S) // scause
```

How about using a helper function (not sure what the best name is though) for the simple case where it just depends on an extension being enabled?

Timmmm: `existsIfEnabled(Ext_S)` or `accessibleIfEnabled(Ext_S)`?

model/extensions/Zicsr/zicsr_insts.sail
```diff
@@ -29,9 +29,12 @@ mapping clause encdec = CSRImm(csr, imm, rd, op)
   when currentlyEnabled(Ext_Zicsr)
 
 function doCSR(csr : csreg, rs1_val : xlenbits, rd : regidx, op : csrop, is_CSR_Write: bool) -> ExecutionResult = {
-  if not(check_CSR(csr, cur_privilege, is_CSR_Write))
-  then Illegal_Instruction()
-  else if not(ext_check_CSR(csr, cur_privilege, is_CSR_Write))
+  match check_CSR(csr, cur_privilege, is_CSR_Write) {
+    CSRAccessSuccess => (),
+    CSRAccessIllegal => return Illegal_Instruction(),
+    CSRNotExist => return Illegal_Instruction(),
```
Timmmm: So currently these have exactly the same effect? What will the difference be in future? I think we need a comment for that.

model/core/sys_regs.sail
```diff
@@ -409,9 +409,9 @@ mapping clause csr_name_map = 0x30A  <-> "menvcfg"
 mapping clause csr_name_map = 0x31A  <-> "menvcfgh"
 mapping clause csr_name_map = 0x10A  <-> "senvcfg"
 
-function clause is_CSR_accessible(0x30A, _, _) = currentlyEnabled(Ext_U) // menvcfg
-function clause is_CSR_accessible(0x31A, _, _) = currentlyEnabled(Ext_U) & (xlen == 32) // menvcfgh
-function clause is_CSR_accessible(0x10A, _, _) = currentlyEnabled(Ext_S) // senvcfg
+function clause is_CSR_accessible(0x30A, _, _) = if currentlyEnabled(Ext_U) then CSRAccessSuccess else CSRNotExist // menvcfg
+function clause is_CSR_accessible(0x31A, _, _) = if currentlyEnabled(Ext_U) & (xlen == 32) then CSRAccessSuccess else CSRNotExist // menvcfgh
+function clause is_CSR_accessible(0x10A, _, _) = if currentlyEnabled(Ext_S) then CSRAccessSuccess else CSRNotExist // senvcfg
```
Timmmm: `CSRNotExist` at least doesn't feel like the right name here, even if it is the right behaviour. The CSR can still *exist*, but just be temporarily inaccessible if supervisor mode is disabled for example.

I guess you could say it effectively doesn't exist at that time... maybe we can think of a better name. Probably the names should refer to the actual effect that returning them has, but as it is the same at the moment I don't know what to suggest.

model/core/sys_regs.sail
arichardson: Either of these sound good to me. I think it might also make sense to have a variant for the *h CSRs to avoid complicated if then else

model/core/sys_regs.sail
trdthg: then we use `CSRNotEffective` and `CSREffectiveIfEnabled(Ext_S)`? And only use CSRNotExist in fallback.

trdthg:
I just split CSRNotExist into these two

- CSRNotDefined: CSR is not defined in spec or not implemented by sail-model yet
    - maybe we can also have a CSRNotImplemented, but this cannot be distinguished.
- CSRNotEnabled: CSR is disabled by config or misa

I have also categorized things like xlen == 32 to CSRNotEnabled, because according to the manual, you can modify the width of a CSR by changing SXLEN at runtime.

---

I also considered whether CSRNotEnabled needs to be further divided into

- CSRNotSupported(like hartSupported): CSR is implemented but disabled by config which can't be modified at runtime
- CSRNotEnabled: CSR is disabled by things like misa that can be modified at runtine

But you have to do a lot of work, for example

```sail
function clause currentlyEnabled(Ext_U) = hartSupports(Ext_U) & misa[U] == 0b1 & currentlyEnabled(Ext_Zicsr)
function clause is_CSR_accessible(0x30A, _, _) = currentlyEnabled(Ext_U) // menvcfg

currentlyEnabled(Ext_U)
    |
    v
if hartSupports(Ext_U) & hartSupports(Ext_Zicsr) then 
	// Here you can just check `misa[U] == 0b1`, but you must carefully check
    // the intersection of each `currentlyEnabled` and `hartSupports`.
	if currentlyEnabled(Ext_U) & currentlyEnabled(Ext_Zicsr)
	then CSRNotEnabled
	else CSRAccessSuccess
else CSRNotSupported
```

Considering we don't need such a detailed division at the moment, I don't really plan to do this.

model/core/csr_begin.sail
```diff
@@ -15,10 +15,50 @@ scattered mapping csr_name_map
...
+  //
+  // 1. The CSR is explicitly disabled by the json configuration, it can't be
+  //    enabled at runtime
+  // 2. The CSR is disabled by misa bit, it could be modified at runtime
+  // 3. The CSR does not exist in current xlen. According to spec, you can
+  //    change the width of a CSR by changing SXLEN or UXLEN
+  //    "When MXLEN=32, the SXL and UXL fields do not exist, and SXLEN=32 and UXLEN=32.
+  //     When MXLEN=64, if S-mode is not supported, then SXL is read-only zero. Otherwise,
+  //     it is a WARL field that encodes the current value of SXLEN."
+  CSRNotEnabled,
```
arichardson: How is this different from CSRAccessIllegal? I think those two cases could be merged?

nadime15: I am just thinking out loud here but `CSRAccessIllegal` could mean in the current context that U-mode is accessing an M-mode CSR or that the CSR does not exist, or writing into a read-only CSR. `CSRNotEnabled` would apply if FP or vector CSRs are currently disabled, so the CSR exists but access is not permitted at the moment. But my thoughts do not match the comments in the code above.

arichardson: I guess both of those are just a special case of the "illegal" case. Do we need to distinguish this for anything other than improved logging? Looking at the current diff, handling those specializations adds significant extra logic that could be simpler otherwise:

```
function clause is_CSR_accessible(0xC02, priv, _) = if currentlyEnabled(Ext_Zicntr) then
                                                        if currentlyEnabled(Ext_Zicntr) & counter_enabled(2, priv) then CSRAccessSuccess else CSRAccessIllegal
                                                    else CSRNotEnabled // instret
```
Could just be
```
function clause is_CSR_accessible(0xC02, priv, _) = CSREnabledIf(Ext_Zicntr, counter_enabled(2, priv)) // instret
```

arichardson: I do think it's valuable to differentiate CSRAccessIllegal and CSRNotDefined, since the trace can then show "access to unknown CSR" and for CSRAccessIllegal "access constraints not met" or similar.

jordancarlin: Another reason for the distinction is that there are a few CSRs that are explicitly defined to throw illegal instruction exceptions, but the general case of accessing an undefined CSR is implementation dependent. Right now the model throws an illegal instruction exception, but it is equally legal to not take a trap and just continue execution.

Timmmm: I agree this is confusing and I'm not sure it makes sense to distinguish these cases until we actually *do* do something different with them (e.g. log them differently or abort on reserved behaviour).

Alternatively perhaps the behavioural and non-behavioural information could be split by turning it into an enum where the `Illegal` variant has a `reason` enum, or something like that.

nadime15: I initially assumed that an external debugger would differentiate between these cases (see https://github.com/riscv/riscv-debug-spec/issues/1140), but that does not seem to be the case (even though we could propagate this kind of information to an external debugger that certain CSR's are currently disabled if openocd plays along). Currently, we do not distinguish at all between not enabled and not implemented. I believe it would be beneficial to differentiate them for potential future use cases, logging as you described (or cases like the one @jordancarlin described).

Timmmm: I'm just a bit worried that because RISC-V itself doesn't make any distinction between them, and if we also don't make use of the distinction, then it's going to be really hard to tell what it actually means.

arichardson: Adding a reason string would be a great solution. I've debugged issues in. Qemu in the past where a csr was incorrectly marked as inaccessible and if I'd had a trace telling me why I'm getting the illegal instruction fault that would have been so much easier to diagnose.

nadime15: > I'm just a bit worried that because RISC-V itself doesn't make any distinction between them, and if we also don't make use of the distinction, then it's going to be really hard to tell what it actually means.

Yeah I agree it could get messy pretty quickly. 

I like your idea of passing the reason as an enum we could use internally/logging, without tying it to the spec itself.

trdthg:
Using a separate enum to represent the reason is indeed a good idea, now the code looks much better

```sail
enum CSRAccessIllegalReason = {
  // CSR is not defined in spec or not implemented by sail-model yet
  CSRNotDefined,

  // other
  CSRNotAccessible,
}

union CSRAccessResult = {
  CSRAccessSuccess : unit,
  CSRAccessIllegal : CSRAccessIllegalReason,
}
```

Now `CSRNotAccessible` is all cases that can raise `Illegal_Instruction` except for `CSRNotDefined`

In addition, I just wrote a more aggressive version that distinguishes various unaccessible reasons in detail. see #1402

model/core/csr_begin.sail
```diff
@@ -15,10 +15,49 @@ scattered mapping csr_name_map
...
+}
+
 // returns whether a CSR exists
-val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> bool
+val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> CSRAccessResult
 scattered function is_CSR_accessible
 
+function CSRDefinedIf(exist : bool) -> CSRAccessResult = if exist then CSRAccessSuccess() else CSRAccessIllegal(CSRNotDefined)
+
+function CSRAccessibleIfBool(accessible : bool) -> CSRAccessResult = if accessible then CSRAccessSuccess() else CSRAccessIllegal(CSRNotAccessible)
```
arichardson:
```suggestion
function CSRAccessibleIfBool(accessible : bool) -> CSRAccessResult = 
  if accessible then CSRAccessSuccess() 
  else CSRAccessIllegal(CSRNotAccessible)
```
For consistency with the other functions?

arichardson (approved):
looks good to me

model/sys/sys_control.sail
```diff
@@ -20,16 +20,16 @@ function csrPriv(csr : csreg) -> nom_priv_bits = csr[9..8]
 
 // Check that the CSR access is made with sufficient privilege.
 function check_CSR_priv(csr : csreg, p : Privilege) -> bool =
-  privLevel_to_bits(p) >=_u csrPriv(csr)
+  privLevel_to_bits(p) >=_u csrPriv(csr) & not(csrAccess(csr) == 0b11)
 
 // Check that the CSR access isn't a write to a read-only CSR.
 function check_CSR_access(csr : csreg, access_type : CSRAccessType) -> bool =
-  not((access_type == CSRWrite | access_type == CSRReadWrite) & (csrAccess(csr) == 0b11))
+  (access_type == CSRWrite | access_type == CSRReadWrite)
```
nadime15, line 27: Maybe it's just me but I dont think we should check whether we access a read-only CSR (while performing a write, etc.) in check_CSR_priv(). This functions (according to the name) should only check whether we have the sufficient privileges. It might make sense to define a new function to check that?

model/core/csr_begin.sail
```diff
@@ -15,10 +15,53 @@ scattered mapping csr_name_map
...
+val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> CSRAccessResult
 scattered function is_CSR_accessible
 
+function CSRDefinedIf(exist : bool) -> CSRAccessResult =
+  if exist then CSRAccessSuccess()
+  else CSRAccessIllegal(CSRNotDefined)
+
+function CSRAccessibleIfBool(accessible : bool) -> CSRAccessResult =
+  if accessible then CSRAccessSuccess()
+  else CSRAccessIllegal(CSRNotAccessible)
```
nadime15, line 53: I am curious what others think, but I feel it might be a bit unfortunate that we have two functions with the same signature, one overloaded and the other not. I understand the reasoning behind it, but I am not sure it would be immediately obvious to someone new contributing to the project.

They might have to dive deep into the code to notice the difference.

But to be honest, I dont have an alternative for now :D

PR #1358, opened 2025-10-22
Update is_CSR_accessible signature to accept CSRAccessType instead of isWrite

trdthg:
based on #1357

Instruction | rd is x0 | rs1 is x0 | Reads CSR | Writes CSR
-- | -- | -- | -- | --
CSRRW | Yes | - | No | Yes
CSRRW | No | - | Yes | Yes
CSRRS/CSRRC | - | Yes | Yes | No
CSRRS/CSRRC | - | No | Yes | Yes

### DIFF
--- a/model/core/ext_regs.sail
+++ b/model/core/ext_regs.sail
@@ -11,10 +11,10 @@
 // overridden by extensions.
 
 /*!
-THIS(csrno, priv, isWrite) allows an extension to block access to csrno,
+THIS(csrno, priv, CSRAccessType) allows an extension to block access to csrno,
 at Privilege level priv. It should return true if the access is allowed.
 */
-function ext_check_CSR(csrno : csreg, p : Privilege, isWrite : bool) -> bool = true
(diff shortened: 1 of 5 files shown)


model/core/ext_regs.sail
```diff
@@ -14,7 +14,7 @@
 THIS(csrno, priv, isWrite) allows an extension to block access to csrno,
 at Privilege level priv. It should return true if the access is allowed.
 */
-function ext_check_CSR(csrno : csreg, p : Privilege, isWrite : bool) -> bool = true
+function ext_check_CSR(csrno : csreg, p : Privilege, ac : CSRAccessType) -> bool = true
```
trdthg, line 17: I guess this will break cheri?

pmundkur:
This seems to have broken `csrrs`, and the tests enter an infinite trap loop.   It seems to be a simpler refactor to first implement the CSRAccessType before fixing the return type.  See #1369.

trdthg:
Finished in #1369

PR #1359, opened 2025-10-24
Rocq updates

bacam:
The Sail 0.20 release can build the RISC-V model again without using a development version.  Some minor updates are required: an out-of-date termination measure and some build changes.  I've also removed some obsolete functions in the handwritten Rocq code which might conflict with future Sail releases.

### DIFF
--- a/.github/workflows/rocq.yml
+++ b/.github/workflows/rocq.yml
@@ -1,14 +1,15 @@
 name: Rocq
 
-on: [workflow_dispatch]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
     runs-on: ubuntu-latest
     env:
       OPAMROOT: /home/rocq/.opam
+      SAIL_VERSION: "0.20"
     container:
-      image: rocq/rocq-prover:9.0.0
+      image: rocq/rocq-prover:9.0.1
       # github requires root to access some resources from outside the container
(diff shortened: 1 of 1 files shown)


coq-sail-riscv.opam
```diff
@@ -17,6 +17,17 @@ install: [
...
+  "sail_coq_backend" {>= "0.20"}
+  "sail_doc_backend" {>= "0.20"}
+  "sail_latex_backend" {>= "0.20"}
+  "sail_lean_backend" {>= "0.20"}
+  "sail_lem_backend" {>= "0.20"}
+  "sail_smt_backend" {>= "0.20"}
+  "sail_sv_backend" {>= "0.20"}
+  "sail_ocaml_backend" {>= "0.20"}
+  "sail_output" {>= "0.20"}
+  "sail_manifest" {>= "0.20"}
```
pmundkur, line 30: I'm curious about the reason for these added dependencies.

bacam: The Sail package has a "post" dependency on the backends, which ensures that they get installed but doesn't limit the order that packages are built in.  Unfortunately, this means that if another package depends on `sail` alone then opam might attempt to build it before the backends.  Hopefully we'll get rid of some or all of this in a future Sail release.

Issue #1339, opened 2025-10-14
Use better syntax for immediates in encdec clauses


Timmmm:
Instead of this

```
mapping clause encdec = JAL(imm_19 @ imm_7_0 @ imm_8 @ imm_18_9 @ 0b0, rd)
  <-> imm_19 : bits(1) @ imm_18_9 : bits(10) @ imm_8 : bits(1) @ imm_7_0 : bits(8) @ encdec_reg(rd) @ 0b1101111
```

We can apparently do this which is much nicer

```
mapping clause encdec = JAL(imm @ 0b0, rd)
  <-> imm[19] @ imm[9..0] @ imm[10] @ imm[18..11] @ encdec_reg(rd) @ 0b1101111
```

If you search for the regex `imm_?\d` you'll find a load of other similar instructions:

* C_ADDIW
* C_LI
* C_LUI
* C_LBU
* C_LHU
* C_SB
* C_SH
* STORE_FP
* BTYPE
* STORE
* VROR_VI

Maybe others.

PR #1364, opened 2025-10-28
Refactor immediates in encoding clauses for readability.

pmundkur:
Use indices into immediates instead of building the immediate from concatenated parts. 
Not all cases could be handled due to a Sail compiler bug: rems-project/sail#1540.

Addresses #1339.

### DIFF
--- a/model/extensions/C/zca_insts.sail
+++ b/model/extensions/C/zca_insts.sail
@@ -16,7 +16,7 @@
 union clause instruction = C_NOP : (bits(6))
 
 mapping clause encdec_compressed = C_NOP(imm)
-  <-> 0b000 @ imm[5] : bits(1) @ 0b00000 @ imm[4..0] : bits(5) @ 0b01
+  <-> 0b000 @ imm[5] @ 0b00000 @ imm[4..0] @ 0b01
   when currentlyEnabled(Ext_Zca)
 
 function clause execute C_NOP(_) = RETIRE_SUCCESS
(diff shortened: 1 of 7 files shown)


model/extensions/C/zca_insts.sail
```diff
@@ -15,8 +15,8 @@
 // *****************************************************************
 union clause instruction = C_NOP : (bits(6))
 
-mapping clause encdec_compressed = C_NOP(imm5 @ imm40)
-  <-> 0b000 @ imm5 : bits(1) @ 0b00000 @ imm40 : bits(5) @ 0b01
+mapping clause encdec_compressed = C_NOP(imm)
+  <-> 0b000 @ imm[5] : bits(1) @ 0b00000 @ imm[4..0] : bits(5) @ 0b01
```
jordancarlin: Is `bits(1)` still necessary now that we have the indices? Might be nice to eliminate that down the road (presumably would need some more tweaks in the Sail compiler).

jordancarlin (approved):

model/extensions/C/zca_insts.sail
pmundkur: Thanks for the reminder.  I was planning to remove the types but forgot..

jordancarlin: Even better! This looks great!

jordancarlin (approved):

Timmmm (approved):
Haven't tediously checked all of the numbers but this looks great to me! I think our testing is probably good enough to catch any mistakes.

PR #1366, opened 2025-10-30
Split config_print_platform into finer grained options

### LINKED ISSUE #1340: Consider splitting --trace-platform into finer-grained trace options
Currently `--trace-platform` tracks a lot of things: CLINT mtime, exceptions, interrupts, HTIF store writes, mret/sret, etc. The main problem is CLINT mtime, when a hart is halted in debug mode, it keeps printing the CLINT time, even though nothing is happening. This makes the logs very messy and debugging harder.  Its useful when someone just want to track mode changes or other specific events, without all the extra output.

It would be nice to break `--trace-platform` into separate trace options so we can enable only the things we want to see.

We can keep `--trace-platform` to group all these suboptions together, but it would be nice if users had more control over what they want to see.

KotorinMinami:
Split the config_print_platform to five options:

- config_print_clint
- config_print_exception
- config_print_interrupt
- config_print_htif
- config_print_pma

The main problem with a single option was CLINT mtime: when a hart is halted in debug mode, it keeps printing the CLINT time, even though nothing is happening.

Fixes #1340


c_emulator/riscv_sim.cpp
```diff
@@ -185,8 +189,26 @@ static void setup_options(CLI::App &app)
...
+  app.add_flag_callback(
+      "--trace-platform",
+      [] {
+        config_print_clint_mem_access = true;
+        config_print_exception = true;
+        config_print_interrupt = true;
+        config_print_htif = true;
+        config_print_pma = true;
+      },
+      "Enable trace output for privilege changes, MMIO, interrupts");
```
nadime15: Should this comment be more specific about which events are being traced?
```suggestion
      "Enable trace output for platform-level events (MMIO, interrupts, exceptions, CLINT, HTIF, PMA)"
```

nadime15 (commented):
@KotorinMinami Thanks for working on that!

model/sys/mem.sail
```diff
@@ -131,7 +131,7 @@ function pmaCheck forall 'n, 0 < 'n <= max_mem_access .
             //                                   "Invalid CacheAccessType; Inval, Clean and Flush check Read/Write permissions."),
           };
 
-          if get_config_print_platform() & not(canAccess)
+          if get_config_print_pma() & not(canAccess)
           then print_endline("PMA check failed for " ^ hex_bits_str(bits_of(paddr)) ^ " PMA " ^ to_str(attributes));
```
arichardson: All others seem to use `print_log()`, I assume this one should too?

nadime15 (approved):

c_emulator/riscv_config.h
```diff
@@ -5,6 +5,10 @@ extern bool config_print_instr;
 extern bool config_print_step;
 extern bool config_print_reg;
 extern bool config_print_mem_access;
-extern bool config_print_platform;
+extern bool config_print_clint_mem_access;
```
Timmmm: I think `...clint_mem_access` should be renamed to just `...clint` because it also controls the `clint_dispatch()` message which isn't a memory access.

c_emulator/riscv_sim.cpp
```diff
@@ -172,8 +176,27 @@ static void setup_options(CLI::App &app)
                "Enable trace output for memory accesses");
   app.add_flag("--trace-rvfi", config_print_rvfi,
                "Enable trace output for RVFI");
-  app.add_flag("--trace-platform", config_print_platform,
-               "Enable trace output for privilege changes, MMIO, interrupts");
+  app.add_flag("--trace-clint", config_print_clint_mem_access,
+               "Enable trace output for CLINT memory accesses");
```
```suggestion
               "Enable trace output for CLINT memory accesses and status");
```

Timmmm (changes requested):

c_emulator/riscv_prelude.cpp
```diff
@@ -40,9 +40,29 @@ bool get_config_print_instr(unit)
...
+}
+
+bool get_config_print_exception(unit)
+{
+  return config_print_exception;
+}
+
+bool get_config_print_interrupt(unit)
+{
+  return config_print_exception;
```
pmundkur:
```suggestion
  return config_print_interrupt;
```

pmundkur (changes requested):
Needs a typo fix, but otherwise LGTM.

pmundkur (approved):

Timmmm (approved):

Issue #1350, opened 2025-10-17
How about having is_CSR_accessible return enum?


trdthg:
Currently we have

```
if not(check_CSR(csr, cur_privilege, is_CSR_Write))
  then Illegal_Instruction()
else ...
```

And H extension describe

```
Specifically, a virtual-instruction exception is raised for the following cases:

in VS-mode, attempts to access a non-high-half counter CSR when the corresponding bit in hcounteren is 0 and the same bit in mcounteren is 1;

in VS-mode, if XLEN=32, attempts to access a high-half counter CSR when the corresponding bit in hcounteren is 0 and the same bit in mcounteren is 1;

in VU-mode, attempts to access a non-high-half counter CSR when the corresponding bit in either hcounteren or scounteren is 0 and the same bit in mcounteren is 1;

in VU-mode, if XLEN=32, attempts to access a high-half counter CSR when the corresponding bit in either hcounteren or scounteren is 0 and the same bit in mcounteren is 1;

in VS-mode or VU-mode, attempts to execute a hypervisor instruction (HLV, HLVX, HSV, or HFENCE);

in VS-mode or VU-mode, attempts to access an implemented non-high-half hypervisor CSR or VS CSR when the same access (read/write) would be allowed in HS-mode, assuming mstatus.TVM=0;

in VS-mode or VU-mode, if XLEN=32, attempts to access an implemented high-half hypervisor CSR or high-half VS CSR when the same access (read/write) to the CSR"s low-half partner would be allowed in HS-mode, assuming mstatus.TVM=0;

in VU-mode, attempts to execute WFI when mstatus.TW=0, or to execute a supervisor instruction (SRET or SFENCE);

in VU-mode, attempts to access an implemented non-high-half supervisor CSR when the same access (read/write) would be allowed in HS-mode, assuming mstatus.TVM=0;

in VU-mode, if XLEN=32, attempts to access an implemented high-half supervisor CSR when the same access to the CSR’s low-half partner would be allowed in HS-mode, assuming mstatus.TVM=0;

in VS-mode, attempts to execute WFI when hstatus.VTW=1 and mstatus.TW=0, unless the instruction completes within an implementation-specific, bounded time;

in VS-mode, attempts to execute SRET when hstatus.VTSR=1; and

in VS-mode, attempts to execute an SFENCE.VMA or SINVAL.VMA instruction or to access satp, when hstatus.VTVM=1.

[...]

PR #1368, opened 2025-10-31
Rename the AccessType for memory to be more explicit.

pmundkur:
CSR accesses will also need to be more explicit in the future to support the hypervisor, and will have similarly named read/write/readwrite accesses; this renaming helps disambiguate the two access targets.

### DIFF
--- a/model/core/types.sail
+++ b/model/core/types.sail
@@ -94,19 +94,19 @@ overload to_str = {privLevel_to_str}
 // memory access types
 
 union MemoryAccessType ('a : Type) = {
-  MemoryRead             : 'a,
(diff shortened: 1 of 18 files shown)


pmundkur:
As discussed in https://github.com/riscv/sail-riscv/issues/1350#issuecomment-3413591993  
cc @trdthg

jordancarlin (approved):
LGTM

Timmmm:
Instead or `MemoryRead`/`MemoryWrite` could we use `Load`/`Store`, then it hopefully wont clash with CSR reads/writes and it isn't quite so verbose?

Timmmm (approved):

PR #1370, opened 2025-11-01
Add callbacks and logging for page table walks

Arielfoever:
Add callbacks for each step of the page table walk process, and an option to log them.

This required some quite awkward changes to resolve a dependency loop. We can fix this better with [a change to the Sail compiler](https://github.com/rems-project/sail/issues/1560#issuecomment-3628341263).

Fixes #1356

PR #1372, opened 2025-11-02
Provide default callback implementations

Arielfoever:
Remove callbacks which are empty. Add default implementation in base class.

### DIFF
--- a/c_emulator/riscv_callbacks_if.h
+++ b/c_emulator/riscv_callbacks_if.h
@@ -13,34 +13,52 @@ public:
   virtual void mem_write_callback([[maybe_unused]] const char *type,
                                   [[maybe_unused]] sbits paddr,
                                   [[maybe_unused]] uint64_t width,
-                                  [[maybe_unused]] lbits value) { }
+                                  [[maybe_unused]] lbits value)
+  {
+  }
   virtual void mem_read_callback([[maybe_unused]] const char *type,
                                  [[maybe_unused]] sbits paddr,
                                  [[maybe_unused]] uint64_t width,
-                                 [[maybe_unused]] lbits value) { }
+                                 [[maybe_unused]] lbits value)
+  {
+  }
   virtual void
   mem_exception_callback([[maybe_unused]] sbits paddr,
-                         [[maybe_unused]] uint64_t num_of_exception) { }
+                         [[maybe_unused]] uint64_t num_of_exception)
+  {
+  }
   virtual void
   xreg_full_write_callback([[maybe_unused]] const_sail_string abi_name,
                            [[maybe_unused]] sbits reg,
-                           [[maybe_unused]] sbits value) { }
+                           [[maybe_unused]] sbits value)
+  {
+  }
   virtual void freg_write_callback([[maybe_unused]] unsigned reg,
-                                   [[maybe_unused]] sbits value) { }
+                                   [[maybe_unused]] sbits value)
+  {
+  }
   virtual void
   csr_full_write_callback([[maybe_unused]] const_sail_string csr_name,
                           [[maybe_unused]] unsigned reg,
-                          [[maybe_unused]] sbits value) { }
+                          [[maybe_unused]] sbits value)
+  {
+  }
   virtual void
   csr_full_read_callback([[maybe_unused]] const_sail_string csr_name,
                          [[maybe_unused]] unsigned reg,
-                         [[maybe_unused]] sbits value) { }
+                         [[maybe_unused]] sbits value)
+  {
+  }
   virtual void vreg_write_callback([[maybe_unused]] unsigned reg,
-                                   [[maybe_unused]] lbits value) { }
+                                   [[maybe_unused]] lbits value)
+  {
+  }
   virtual void pc_write_callback([[maybe_unused]] sbits new_pc) { }
   virtual void redirect_callback([[maybe_unused]] sbits new_pc) { }
   virtual void trap_callback([[maybe_unused]] bool is_interrupt,
-                             [[maybe_unused]] fbits cause) { }
+                             [[maybe_unused]] fbits cause)
+  {
+  }
 };
 
 void register_callback(callbacks_if *cb);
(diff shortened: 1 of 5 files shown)


Arielfoever:
Shall I remove these codes or just comment it?

Timmmm:
I think it's better to require implementing all of them explicitly, that way you can't miss any. For example when we add a new callback you have to consider how it should be handled in all the callback handlers.

Arielfoever:
People discussed this https://github.com/riscv/sail-riscv/pull/1370#discussion_r2486454251 https://github.com/riscv/sail-riscv/pull/1370#discussion_r2486460163

arichardson:
> I think it's better to require implementing all of them explicitly, that way you can't miss any. For example when we add a new callback you have to consider how it should be handled in all the callback handlers.

I agree that it's good to have the compiler force us to make sure we did the right thing, but this does add quite a bit of boilerplate in cases where the callback is not actually needed, so I'd argue it's slightly better to have a default empty implementation.

c_emulator/riscv_callbacks_log.cpp
```diff
@@ -109,18 +109,18 @@ void log_callbacks::vreg_write_callback(unsigned reg, lbits value)
...
+// void log_callbacks::redirect_callback(sbits new_pc)
+// {
+//   (void)new_pc;
+// }
 
-void log_callbacks::trap_callback(bool is_interrupt, fbits cause)
-{
-  (void)is_interrupt;
-  (void)cause;
-}
```
arichardson, line 126: I'm surprised this one is empty, I would have expected us to log traps.

nadime15: We do but not via callbacks (--trace-platform). 

See sys_control.sail

```
 then print_log("handling " ^ to_str(c)
                     ^ " at priv " ^ to_str(del_priv)
                     ^ " with tval " ^ bits_str(tval(info)));
```

jordancarlin: That should probably be turned into a callback eventually.

Arielfoever:
> > I think it's better to require implementing all of them explicitly, that way you can't miss any. For example when we add a new callback you have to consider how it should be handled in all the callback handlers.
> 
> I agree that it's good to have the compiler force us to make sure we did the right thing, but this does add quite a bit of boilerplate in cases where the callback is not actually needed, so I'd argue it's slightly better to have a default empty implementation.

I am sure the compiler is checking us to do the right thing. The just check if the callback has a implementation wherever they are. So copy paste empty implementation and have default empty one are equal.

Anyway, this pr is somehow increasing speed. Our CI told me it's 42 second quicker than master one.

arichardson:
> > > I think it's better to require implementing all of them explicitly, that way you can't miss any. For example when we add a new callback you have to consider how it should be handled in all the callback handlers.
> > 
> > 
> > I agree that it's good to have the compiler force us to make sure we did the right thing, but this does add quite a bit of boilerplate in cases where the callback is not actually needed, so I'd argue it's slightly better to have a default empty implementation.
> 
> I am sure the compiler is checking us to do the right thing. The just check if the callback has a implementation wherever they are. So copy paste empty implementation and have default empty one are equal.
> 
> Anyway, this pr is somehow increasing speed. Our CI told me it's 42 second quicker than master one.

The compiler might now be able to identify that these functions are always empty and remove all calls to them whereas before it wasn't able to look through the vtable loads for the two different empty implementations.

Arielfoever:
@Timmmm I think it should be a part of tgmm agenda.

@trdthg will give a briefing on my behalf if I need to see the doctor next Monday.

Timmmm:
Sorry I missed the last meeting (new job), but it sounds like I'm the only one who wanted the methods to be abstract so I'm happy to go with the consensus.

c_emulator/riscv_callbacks_if.h
```diff
@@ -8,31 +8,39 @@
...
+  virtual void
+  csr_full_read_callback([[maybe_unused]] const_sail_string csr_name,
+                         [[maybe_unused]] unsigned reg,
+                         [[maybe_unused]] sbits value) { };
+  virtual void vreg_write_callback([[maybe_unused]] unsigned reg,
+                                   [[maybe_unused]] lbits value) { };
+  virtual void pc_write_callback([[maybe_unused]] sbits new_pc) { };
+  virtual void redirect_callback([[maybe_unused]] sbits new_pc) { };
+  virtual void trap_callback([[maybe_unused]] bool is_interrupt,
+                             [[maybe_unused]] [[maybe_unused]] fbits cause) { };
```
```suggestion
                             [[maybe_unused]] fbits cause) { };
```

c_emulator/riscv_callbacks_if.h
```diff
@@ -8,31 +8,39 @@
...
-                                       sbits value)
-      = 0;
-  virtual void csr_full_read_callback(const_sail_string csr_name, unsigned reg,
-                                      sbits value)
-      = 0;
-  virtual void vreg_write_callback(unsigned reg, lbits value) = 0;
-  virtual void pc_write_callback(sbits new_pc) = 0;
-  virtual void redirect_callback(sbits new_pc) = 0;
-  virtual void trap_callback(bool is_interrupt, fbits cause) = 0;
+  virtual void fetch_callback([[maybe_unused]] sbits opcode) { };
```
Timmmm: Lots of extra semicolons.

```suggestion
  virtual void fetch_callback([[maybe_unused]] sbits opcode) { }
```

c_emulator/riscv_callbacks_log.cpp
```diff
@@ -31,10 +31,10 @@ log_callbacks::log_callbacks(bool config_print_reg,
...
 // The model assumes that these functions do not change the state of the model.
 
-void log_callbacks::fetch_callback(sbits opcode)
-{
-  (void)opcode;
-}
+// void log_callbacks::fetch_callback(sbits opcode)
+// {
+//   (void)opcode;
+// }
```
Timmmm: Delete all these rather than commenting them out.

pmundkur (approved):

PR #1373, opened 2025-11-02
Add new sail_smt_cache to git ignore list

Arielfoever:
`z3_problems` was renamed to the slightly more descriptive `sail_smt_cache` in the latest versions of the Sail compiler.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ z3_problems
 /build
 # CMake build directories used by CLion.
 /cmake-build-*
+# z3 cache
+sail_smt_cache


.gitignore
```diff
@@ -12,3 +12,4 @@ z3_problems
 /build
 # CMake build directories used by CLion.
 /cmake-build-*
+model/sail_smt_cache
```
Timmmm:
```suggestion
/model/sail_smt_cache
```

jordancarlin: Let’s just ignore any file with this name like we’ve been doing for `z3_problems`.

```suggestion
sail_smt_cache
```

Arielfoever: Rejected due to https://github.com/riscv/sail-riscv/pull/1373#discussion_r2484871547

Timmmm (approved):

PR #1376, opened 2025-11-04
Scatter termination.sail

Ptival:
Scatter the termination module in such a way that one could require some termination measures without having to include every single instruction in the output. This makes building a subset of the model easier.

The termination measures for `vmem_read_addr` and `vmem_write_addr` no longer need to be protected only for the Rocq output, so they get taken out of the `$iftarget coq`.


jordancarlin:
Would #905 eliminate the need for `compressed_measure`? If so, maybe we should get that PR updated and merged first.

pmundkur:
> Would #905 eliminate the need for `compressed_measure`? If so, maybe we should get that PR updated and merged first.

That's blocked on https://github.com/rems-project/sail/issues/1357, so I'm okay with merging this first and removing `compressed_measure` later.

pmundkur:
@Ptival looks like this needs newlines at end-of-file fixes to pass pre-commit.

Ptival:
Looks like I might have broken something on the Lean side too, I'll investigate:

https://github.com/riscv/sail-riscv/actions/runs/19080408022/job/54510837967

model/core/extensions.sail
```diff
@@ -19,14 +19,28 @@ scattered mapping extensionName
...
+    Ext_Zvks => 1,
+    Ext_C => 2,
+    Ext_Zvknc => 2,
+    Ext_Zvkng => 2,
+    Ext_Zvksc => 2,
+    Ext_Zvksg => 2,
+    _ => 0,
+  }
+
+termination_measure hartSupports(ext) = hartSupports_measure(ext)
```
Timmmm, line 39: Does inlining this work?

```
termination_measure hartSupports(ext) = match ext {
...
}
```

pmundkur:
FYI, I was able to rebase #905 (as #1382), which removed the measure for `execute` but also broke Lean in a new way (I think).

model/core/extensions.sail
Ptival: It does work, but you don't get to put the return type annotation. You get to put the argument type annotation though.

I'm happy to make the change if requested, but I don't know whether this is a desirable change or not.

Ptival:
> FYI, I was able to rebase #905 (as #1382), which removed the measure for `execute` but also broke Lean in a new way (I think).

Ah interesting, I'm going to have a quick look at why your error happens.

In the meantime, I have updated the `currentlyEnabled_measure` with working values. I added comment with the dependency chain I witnessed for future reference.

Leaving this as a draft PR until we figure out whether we can remove the whole `compressed_measure`.

Ptival:
I have followed the removal of `compressed_measure`, added some comments over the termination measures, and it seems to work well for my purposes.

model/core/sys_regs.sail
```diff
@@ -146,6 +146,8 @@ function virtual_memory_supported() -> bool = {
   currentlyEnabled(Ext_Sv32) | currentlyEnabled(Ext_Sv39) | currentlyEnabled(Ext_Sv48) | currentlyEnabled(Ext_Sv57)
 }
 
+termination_measure virtual_memory_supported(_) = 3
```
pmundkur, line 151: A comment here about the dependency chain, similar to the ones for `currentlyEnabled`, would be helpful

model/extensions/cfi/zicfilp_regs.sail
```diff
@@ -32,6 +32,8 @@ function get_xLPE(p : Privilege) -> bool =
     VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
   }
 
+termination_measure get_xLPE(_) = 2
```
pmundkur, line 37: A comment here about the dependency chain, similar to the ones for currentlyEnabled, would be helpful

model/sys/vmem_utils.sail
```diff
@@ -172,6 +172,8 @@ function vmem_read_addr(vaddr, offset, width, acc, aq, rl, res) = {
   Ok(data)
 }
 
+termination_measure vmem_read_addr repeat n
```
pmundkur, line 176: It would be helpful to explain the `repeat` construct here, and refer to it for the `vmem_write_addr` case below.

pmundkur (approved):
LGTM.   Since the measures are now appearing in the main code, it would be helpful to have some explanatory comments where previously there weren't any.

model/core/sys_regs.sail
Ptival: Fixed!

model/extensions/cfi/zicfilp_regs.sail
Ptival: Fixed!

model/sys/vmem_utils.sail
Ptival: I honestly don't know what this does precisely.

It's not very documented... according to the syntax documentation in the SAIL repo this shouldn't even be valid syntax...
image

@Alasdair I think maybe you added this, could you explain to me what this variant of `termination_measure` is?

Alasdair: Not sure if I added it, but it's a termination measure for the loop in the function body. In the syntax `repeat n` is a valid `loop_measure`.

I think this can be placed inline and attached to the loop itself, which might be more clear.

Ptival: Ah, I think I saw the syntax for putting it next to the loop, I will give it a try, as I agree it would be more clear.

Ptival: Alright, I attempted something. It needs `magic_hash` for some reason. I'll wait for CI to check because I think this is specifically exercised by the Rocq backend.

Alasdair: We shouldn't use that option in the model, it's for debugging generated code that uses internally generated identifiers that can't clash with user identifiers (by allowing it to be passed back into the frontend) . If it doesn't support it we should just keep the termination measure separate for now. I also don't guarantee that debug flags will persist between Sail versions, and that flag has mostly been superseeded by the `$sail_internal` directive which is file-local and more descriptive, so it is actual risk of being removed.

I'm really not sure why using an inline termination measure would require that option anyway. I'll look into changing that.

Ptival: Alright I reverted it!

bacam: I added the loop measure syntax originally.  At the time I was the only person using it, so it was important to keep all of the measures separate from the rest of the models (and that's still useful for models that are translated from other languages).  Now it would be better to keep the loop measures in the code, so I'll add proper syntax for that to Sail shortly.  I don't think it's worth holding up this PR for, though; we can tidy this up later.

Ptival:
@pmundkur 

I have rebased over most recent changes, this is ready for review if CI passes.

Timmmm:
Ooo the Lean build is magically fixed...

PR #1377, opened 2025-11-04
Add Sstvecd to supported extensions.

pmundkur:
This just captures already supported functionality in `stvec`. For now, this extension is just gated on support for the Supervisor mode. When we add granular configuration options for modes in `stvec`, this extension can be exposed as a configuration option.

Sstvecd is a profile-defined extension coming from RVA20S64.  It's not clear what version this extension is, so I'm just using v1.0.

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -572,7 +572,7 @@ function tvec_addr(m : Mtvec, c : Mcause) -> option(xlenbits) = {
   }
 }
 
-function clause currentlyEnabled(Ext_Sstvecd) = hartSupports(Ext_Sstvecd)
+function clause currentlyEnabled(Ext_Sstvecd) = hartSupports(Ext_Sstvecd) & currentlyEnabled(Ext_S)
 
 // Exception PC
 


model/core/sys_regs.sail
```diff
@@ -572,6 +572,8 @@ function tvec_addr(m : Mtvec, c : Mcause) -> option(xlenbits) = {
   }
 }
 
+function clause currentlyEnabled(Ext_Sstvecd) = hartSupports(Ext_Sstvecd)
```
jordancarlin: Should probably be this since presumably if S is disabled, Sstvecd would be too.

```suggestion
function clause currentlyEnabled(Ext_Sstvecd) = hartSupports(Ext_Sstvecd) & currentlyEnabled(Ext_S)
```

nadime15:
Somehow this confuses me:

> Sstvecd stvec.MODE must be capable of holding the value 0 (Direct). When stvec.MODE=Direct,
> stvec.BASE must be capable of holding any valid four-byte-aligned address.

What is the point of adding an extension that does not explicitly add anything (this PR)? I think its better to add the config options now and ensure that if Sstvecd is enabled, stvec.MODE is capable of holding the value 0, etc.

I think this approach would also be easier to maintain. I am worried that once we refactor, we might forget that Sstvecd was added.

pmundkur:
The point of this PR is just to indicate that the functionality required by the Sstvecd is already supported by the existing code.   It's one of the extensions required by RVA23.

I think the issue is what should the required behavior be if this extension is explicitly disabled.  Adding an option(s) for that requires a refactor of the trap handling code, which is a bit of work, since that code is too tightly coupled right now.  

I agree that once it is refactored, it would be easier to maintain, but I don't think we would forget that Sstvecd was added if we refactor.  (Note that nothing really got added for Sstvecd, we're just noting explicitly the already existing behavior.)

nadime15:
Ok, fair enough

nadime15 (approved):

model/core/extensions.sail
```diff
@@ -418,6 +418,11 @@ function clause hartSupports(Ext_Sscofpmf) = config extensions.Sscofpmf.supporte
 enum clause extension = Ext_Sstc
 mapping clause extensionName = Ext_Sstc <-> "sstc"
 function clause hartSupports(Ext_Sstc) = config extensions.Sstc.supported
+// stvec.MODE supports Direct, and in that mode stvec.BASE must be
+// capable of holding any valid four-byte-aligned address.
+enum clause extension = Ext_Sstvecd
+mapping clause extensionName = Ext_Sstvecd <-> "sstvecd"
+function clause hartSupports(Ext_Sstvecd) = hartSupports(Ext_S)
```
Timmmm, line 425: Should it be this?

```suggestion
function clause hartSupports(Ext_Sstvecd) = hardSupport(Ext_Svbare)
```

And then add a check that `Ext_Svbare` requires `Ext_S` to `check_mmu_config()` (we probably should have had that already I guess).

Technically it's slightly broken because the reset value of `satp` will still be 0 even if you don't support `Ext_Svbare`, but it's probably minor because what madman is actually going to make a chip that doesn't support Svbare?

Timmmm (changes requested):

model/core/extensions.sail
Timmmm: Wait ignore that - I was totally mixing up `satp` and `stvec`! This LGTM.

Timmmm (approved):

allenjbaum:
I don't see sstvecd listed in the priv spec. What is it? And where is it
defined?

jordancarlin:
> I don't see sstvecd listed in the priv spec. What is it? And where is it defined?

@allenjbaum It is one of the profile defined extensions. See https://github.com/riscv/riscv-profiles/blob/main/src/rva23-profile.adoc?plain=1#L256:

> *Sstvecd* `stvec.MODE` must be capable of holding the value 0
  (Direct).  When `stvec.MODE=Direct`, `stvec.BASE` must be capable of
  holding any valid four-byte-aligned address.

allenjbaum:
Ah, yup, a profile defined thing, not in the priv spec....
I argued that these isn't really an extension as such, but a restriction on
the value of parameters and should be given different names
(and if it were an extension, should be in the spec).

PR #1380, opened 2025-11-06
Fix htif memory range check

trdthg:
Previously an access that *just* touched the bottom of HTIF would have incorrectly been counted as within HTIF. E.g. if HTIF is at 0x8 and you do a 4-byte access to 0x4.

### DIFF
--- a/model/sys/platform.sail
+++ b/model/sys/platform.sail
@@ -50,7 +50,7 @@ function within_clint forall 'n, 0 < 'n <= max_mem_access . (Physaddr(addr) : ph
 function within_htif_writable forall 'n, 0 < 'n <= max_mem_access . (Physaddr(addr) : physaddr, width : int('n)) -> bool =
   match htif_tohost_base {
     None() => false,
-    Some(base) => (addr <_u base + htif_tohost_size) & (addr + width >=_u base)
+    Some(base) => (addr <_u base + htif_tohost_size) & (addr + width >_u base)
   }
 
 function within_htif_readable forall 'n, 0 < 'n <= max_mem_access . (addr : physaddr, width : int('n)) -> bool =


Timmmm:
I don't think the old condition was wrong was it? These are just two different options for what to do when there's a misaligned access that overlaps the edge of the HTIF device. The current option will say it is in HTIF memory and then give you an access fault. This version will say it is outside HTIF and then do a normal write to invisible memory "behind" HTIF.

I think the current behaviour is the better option.

trdthg:
For example

addr = 8002004, width = 4
base = 8002008, htif_tohost_size = 8

This is legal for the current behavior

```
  (addr <_u base + htif_tohost_size) & (addr + width >=_u base)

8002004 < 8002008 + 8                & 8002004 + 4 = 8002008

```

But addr should >= base, isn't it?

trdthg:
Oh, I got your idea, this will be checked inside htif_load/store

trdthg:
I found this during hypervisor testing. There is a test case doing a lot of sw operations to set the entire BSS region to 0. This process triggered HTIF. I'm not entirely sure what happened. I need to check it again.

trdthg:
@Timmmm I think we should change it to

```diff
- (addr <_u base + htif_tohost_size) & (addr + width >=_u base)
+ (addr <_u base + htif_tohost_size) & (addr + width >_u base)
                                                     ^ remove = here
     |htif|
|    |       this is not within htif 
```
---
found it
image

Timmmm:
Ah yes that looks correct.

model/sys/platform.sail
```diff
@@ -50,7 +50,7 @@ function within_clint forall 'n, 0 < 'n <= max_mem_access . (Physaddr(addr) : ph
 function within_htif_writable forall 'n, 0 < 'n <= max_mem_access . (Physaddr(addr) : physaddr, width : int('n)) -> bool =
   match htif_tohost_base {
     None() => false,
-    Some(base) => (addr <_u base + htif_tohost_size) & (addr + width >=_u base)
+    Some(base) => (addr <_u base + htif_tohost_size) & (addr + width >_u base)
```
pmundkur, line 53: Doesn't this still leave open the possibility that `addr <_u base`?   The checks in `within_clint()` look right to me.  Perhaps there should be a version of `core/range_util:range_subset` for such cases that could be used for these `within_*`?

pmundkur (commented):
Not sure this is correct.

model/sys/platform.sail
Timmmm: I'm pretty sure it is correct - can you give an example where it would fail?

But I do agree it's super confusing and should probably be wrapped in a generic function. in `range_util`.

pmundkur: addr = 8002004, width = 8
base = 8002008, htif_tohost_size = 8

Timmmm: Then `(addr <_u base + htif_tohost_size)` is true and `(addr + width >_u base)` is also true, so the whole expression is true which is what we want. See [my earlier comment](https://github.com/riscv/sail-riscv/pull/1380#issuecomment-3496104622) - the name `within` is pretty misleading; we should probably change it to something like:

```
function intersects_htif_writable forall 'n, 0 < 'n <= max_mem_access . (Physaddr(addr) : physaddr, width : int('n)) -> bool =
...
Some(base) => ranges_intersect(addr, width, base, htif_tohost_size),
```

pmundkur: Yeah, the function name is indeed misleading!  `intersects_*` is better.

pmundkur:
> I don't think the old condition was wrong was it? These are just two different options for what to do when there's a misaligned access that overlaps the edge of the HTIF device. The current option will say it is in HTIF memory and then give you an access fault. This version will say it is outside HTIF and then do a normal write to invisible memory "behind" HTIF.
> 

Shouldn't this behavior fall under an access that overlaps two PMAs?  I might have missed it, but I could not find information in the spec on how that should be handled.  The two possibilities you mention indicate the need for a config option.  We could discuss this at the meeting.

Timmmm:
I think normally you would have one PMA for your entire MMIO region that wouldn't support misaligned accesses at all.

Timmmm:
I added #1389 to track making this all make sense, but this fix is at least correct - are you happy to merge it and defer the rest of the work @pmundkur ?

PR #1382, opened 2025-11-06
Add ExecuteAs result for compressed instructions.

pmundkur:
The rationale for this is something we encountered when doing CHERIoT verification using the generated SystemVerilog from Sail. By having the compressed instructions directly invoke execute for the decompressed variant, we end up with (for example) all the logic for load and store instructions duplicated in the output.

It's also a bit awkward because it makes the main `execute` function recursive, which inhibits some useful Sail optimisations.

This resolves this by adding a return type to the `execute` function
```
ExecuteAs : instruction
```
which allows the top-level step function to handle executing the decompressed variant.

Removing the recursion in `execute` also allows the removal of the associated termination measures.

For the Lean build, `BEq` derivations have been suppressed for `ExecutionResult` and `Step` since they now embed an `instruction`, and the derivation for the latter was suppressed for performance reasons.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -404,6 +404,8 @@ foreach (xlen IN ITEMS 32 64)
         --lean-output-dir ${CMAKE_CURRENT_BINARY_DIR}
         --lean-force-output
         --lean-non-beq-type instruction
+        --lean-non-beq-type ExecutionResult
+        --lean-non-beq-type Step
     )
     set(lean_sail_default
         --lean-noncomputable


pmundkur:
This is a rebase of #905.  The Lean build breaks, but Rocq works (afaict).

model/postlude/step.sail
```diff
@@ -202,6 +206,8 @@ function try_step(step_no : nat, exit_wait : bool) -> bool = {
     Step_Fetch_Failure(vaddr, e) => handle_exception(bits_of(vaddr), e),
     Step_Waiting(_) => assert(hart_is_waiting(hart_state), "cannot be Waiting in a non-Wait state"),
     Step_Execute(Retire_Success(), _) => assert(hart_is_active(hart_state)),
+    // This case was handled above when processing the fetch result.
+    Step_Execute(ExecuteAs(_), _) => assert(false),
```
Timmmm, line 212: Slightly unfortunate but it might be type overload if we do this "properly".

Timmmm (approved):
LGTM, one minor suggestion.

model/postlude/step.sail
```diff
@@ -113,8 +113,10 @@ function run_hart_active(step_no: nat) -> Step = {
         // check for RVC once here instead of every RVC execute clause.
         else if currentlyEnabled(Ext_Zca) then {
           nextPC = PC + 2;
-          let r = execute(instruction);
-          Step_Execute(r, instbits)
+          match execute(instruction) {
+            ExecuteAs(inst) => Step_Execute(execute(inst), instbits),
+            r => Step_Execute(r, instbits),
+          }
```
Timmmm: Possibly clearer?

```
let result = match execute(instruction) {
  ExecuteAs(other_instuction) => execute(other_instruction),
  result => result,
};
Step_Execute(result, instbits)
```

Ptival:
Having a quick look at the Lean breakage, it's breaking because the inductive definition of `instruction` in `Defs.lean` is missing a `BEq` derivation clause, which is now necessary since `ExecuteAs` embeds an `instruction` within an `ExecutionResult`, which is trying to derive `BEq`.

On the SAIL side, it seems like every pretty-printed inductive ought to receive a `BEq` derivation clause, unless it is explicitly added to "non-BEq types".
So the culprit is [this line](https://github.com/riscv/sail-riscv/blob/28988dbe5c1cb7c2766734b7b1e31f0afef905f2/model/CMakeLists.txt#L406). I have tried removing the line, and I was able to build the Lean output, so it might be admissible? I don't know what made it unwanted in the past (maybe it's slower? didn't feel too slow on my end).

It seems to fix the Lean build for me.

jordancarlin (approved):

bacam:
The option to suppress a BEq instance in the Lean output was added in `e33b71704` for performance, so it should be safe to remove.  However, an alternative would be to add the same option for the `ExecutionResult` and `Step` types, which depend on the instruction type, which should make it build again.

PR #1383, opened 2025-11-06
Add support for Ssstrict.

pmundkur:
This assumes that any experimental extension is non-conforming, which may exclude conforming custom extensions.  We could fix this in the future by distinguishing between custom and non-conforming extensions.

### DIFF
--- a/handwritten_support/riscv_extras.v
+++ b/handwritten_support/riscv_extras.v
@@ -12,6 +12,8 @@ Require Import Lia.
 Import List.ListNotations.
 Open Scope Z.
 
+Axiom sys_enable_experimental_extensions : unit -> bool.
+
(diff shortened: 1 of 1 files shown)


pmundkur:
It looks like this should be the last extension added for RVA23, and requires hypervisor.
https://github.com/riscv/riscv-profiles/blob/35f286266fa69cb023b652203406cdea76960831/src/rva23-profile.adoc?plain=1#L372
> Ssstrict definition applies to the execution environment claiming to be RVA23-compatible, which must have the hypervisor extension.

Converting to a draft for now.

jordancarlin:
This should be a simple one to revisit soon and is one of the last pieces for RVA23.

PR #1384, opened 2025-11-07
Put Sail smt cache in /build directory

### LINKED ISSUE #1374: Put Sail smt cache in /build directory
We should probably change the Sail commands so that they put the `sail_smt_cache` file `${CMAKE_CURRENT_BINARY_DIR}`. Also I haven't double checked but potentially if you build in parallel different Sail calls will overwrite each other's caches so we might want to give each one a unique name.

KotorinMinami:
Change the Sail commands so that they put the `sail_smt_cache` file in `${CMAKE_CURRENT_BINARY_DIR}`.
Fixes #1374.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -20,6 +20,7 @@ set(sail_common
     # Minimum required Sail compiler version.
     --require-version ${SAIL_REQUIRED_VER}
     # Put Sail smt cache in /build directory since by default Sail will
+    # place it in its working directory which is ${CMAKE_CURRENT_SOURCE_DIR}.
     --memo-z3-path "${CMAKE_CURRENT_BINARY_DIR}/sail_smt_cache"
 )
 


Timmmm:
I'm not totally sure if we want to instead have this as part of each Sail command and use a different name for each one... But this is an improvement at least so lets go with it for now.

> it's currently unclear how Sail independently names caches

Can you explain what this means? I don't quite follow the question.

model/CMakeLists.txt
```diff
@@ -19,6 +19,8 @@ set(sail_common
     --strict-exponentials
     # Minimum required Sail compiler version.
     --require-version ${SAIL_REQUIRED_VER}
+    # Put Sail smt cache in /build directory since --memo-z3 is default enable
```
Timmmm:
```suggestion
    # Put Sail smt cache in /build directory since by default Sail will
    # place it in its working directory which is ${CMAKE_CURRENT_SOURCE_DIR}.
```

KotorinMinami:
It's mainly because of this in #1374 
> Also I haven't double checked but potentially if you build in parallel different Sail calls will overwrite each other's caches so we might want to give each one a unique name.

I can't find a good naming convention that both avoids re-creating caches and ensures different Sail compilers (different versions or different compile commits?) can be correctly distinguished when compiling parallelly.

Timmmm:
I was thinking we just have a different name for each `add_custom_command`, like `sail_smt_cache_c`, `sail_smt_cache_lean`, `sail_smt_cache_html`, `sail_smt_cache_json` etc. Some of those can run in parallel.

But actually now that I think about it I'm not sure there's any point. Sail reads the entire file at the beginning and writes it all at the end so as long as you don't get two compilers writing it at the same time it's probably fine. I'm not sure what happens if that does happen - I guess the first compiler would get a write error when the second one truncates its output file... but I've never actually seen that.

So anyway I think this is good as it is.

model/CMakeLists.txt
```diff
@@ -19,6 +19,8 @@ set(sail_common
     --strict-exponentials
     # Minimum required Sail compiler version.
     --require-version ${SAIL_REQUIRED_VER}
+    # Put Sail smt cache in /build directory since by default Sail will
```
arichardson, line 22:
```suggestion
    # Put Sail smt cache in /build directory since by default Sail will
    # place it in its working directory which is ${CMAKE_CURRENT_SOURCE_DIR}.
```

jordancarlin (approved):

PR #1385, opened 2025-11-07
Add option not to download jsoncons

bacam:
Replaces a previous hack to avoid a download in the Rocq opam package for compatibility with recent versions of cmake.

An alternative to this would be to use `ExternalProject` in the same way that the GMP support works, but that would be a little more involved and change the timing of the download from configure-time to build-time.

### DIFF
--- a/dependencies/jsoncons/CMakeLists.txt
+++ b/dependencies/jsoncons/CMakeLists.txt
@@ -1,5 +1,8 @@
 include(FetchContent)
 
+# Allow downloads to be turned off (in particular, for targets that don't need it).
+option(DOWNLOAD_JSONCONS "Download jsoncons" ON)
+
 # The jsoncons `unit_tests` target conflicts with our own.
 set(JSONCONS_BUILD_TESTS OFF)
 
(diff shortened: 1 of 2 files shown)


dependencies/jsoncons/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 include(FetchContent)
 
+# Allow downloads to be turned off (in particular, for targets that don't need it).
+option(DOWNLOAD_JSONCONS "Download jsoncons" ON)
```
jordancarlin, line 4: What about naming this something like `DOWNLOAD_C_LIBS`? Then we can gate any future libraries on it as well and it might be more obvious to users when they would want to disable it (if they’re not generating the C model).

bacam: It would be a bit awkward because the default for GMP is to use system version rather than downloading.

pmundkur: We could look at renaming if we have more libraries like jsoncons, but for now it seems clearer to just have it as `DOWNLOAD_JSONCONS`.  If there are more, we could rename it to `DOWNLOAD_THIRD_PARTY_C_LIBS` or something similar.  I'm assuming that the system GMP would be first or second-party, whatever they mean.

Timmmm:
What's the issue that downloading it causes?

bacam:
> What's the issue that downloading it causes?

The package build is sandboxed, without networking.

Timmmm:
Ah makes sense. I think it would be worth explaining that in the comment.

bacam:
I've added a comment to the package file (in addition to the one in the jsoncons cmake file).

PR #1388, opened 2025-11-07
Add configurations for page-fault traps that could set mtval.

pmundkur:
These were missed in #1225, and are needed for Sstvala.

### DIFF
--- a/model/sys/sys_control.sail
+++ b/model/sys/sys_control.sail
@@ -289,7 +289,10 @@ function xtval_exception_value(e : ExceptionType, excinfo : xlenbits) -> option(
     E_M_EnvCall() => false,
     E_Extension(_) => true,
     // the rest are all reserved exceptions
-    _ => config base.xtval_nonzero.reserved_faults,
+    E_Reserved_10() => config base.xtval_nonzero.reserved_exceptions,
+    E_Reserved_14() => config base.xtval_nonzero.reserved_exceptions,
+    E_Reserved_16() => config base.xtval_nonzero.reserved_exceptions,
+    E_Reserved_17() => config base.xtval_nonzero.reserved_exceptions,
   } then Some(excinfo) else None()
 }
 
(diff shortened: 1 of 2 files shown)


Timmmm (commented):
What other exceptions are there? Couldn't we just list them all?

pmundkur:
I added the other exceptions.  There is an unfortunate leak of the `E_Extension` hook for external extensions, perhaps we should suppress that from the config file and have a hard-coded default?

Timmmm:
Yeah I think just hard-code it for now. Technically the correct solution is to add yet another `ext_` function like `ext_xtval_exception_value(...)` but let's not worry about that until we try to integrate CHERI properly.

model/sys/sys_control.sail
```diff
@@ -272,15 +272,24 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
...
-    E_Illegal_Instr() => config base.xtval_nonzero.illegal_instruction,
-    _ => true
+    E_Fetch_Page_Fault() => config base.xtval_nonzero.fetch_page_fault,
+    E_Software_Check() => config base.xtval_nonzero.software_check,
+    E_U_EnvCall() => false,
+    E_S_EnvCall() => false,
+    E_M_EnvCall() => false,
+    E_Extension(_) => true,
+    // the rest are all reserved exceptions
+    _ => config base.xtval_nonzero.reserved_faults,
```
Timmmm: Might be worth listing them explicitly too so we don't forget this when adding extra exception types? Also technically I think only some exceptions are "faults".

```suggestion
    E_Reserved_10 => config base.xtval_nonzero.reserved,
    E_Reserved_14 => config base.xtval_nonzero.reserved,
    E_Reserved_16 => config base.xtval_nonzero.reserved,
    E_Reserved_17 => config base.xtval_nonzero.reserved,
```

(I think you'll have to do `let reserved = config base.xtval_nonzero.reserved;` at the top actually otherwise it will complain about reusing the `config`.)

Sorry if I'm nitpicking!

pmundkur: I'll be hoisting them into `let`s in `core/platform_config` to support Sstvala.  Iirc, `config` can't be reused for abstract types; I think that's where I ran into that complaint.

PR #1391, opened 2025-11-11
Flush the TLB when the address translation scheme (`satp.MODE`) is modified.

pmundkur:


pmundkur:
This should really also be done when the PMP configuration is changed, but that needs a refactor of the PMP code to sort out the dependencies.

Timmmm:
I don't think the spec requires this:

> Note that writing satp does not imply any ordering constraints between page-table updates and subsequent address translations, nor does it imply any invalidation of address-translation caches.

I don't see why it would be needed for PMP either unless I'm missing something. PMP checks happen after address translation.

pmundkur:
It's possible for a hardware implementation to be stricter than the spec requires.  For e.g. both Spike and Whisper flush their TLB after a `satp` write, and presumably they are modelling their respective chips.   So maybe this should be gated by a config option.

Timmmm:
Yeah I guess we could have it off-by-default and add it if someone asks?

You don't need it for lockstep verification because we're never going to exactly model a DUT's real TLB anyway, so I'm not sure what the use case would be.

jordancarlin:
> It's possible for a hardware implementation to be stricter than the spec requires. For e.g. both Spike and Whisper flush their TLB after a `satp` write, and presumably they are modelling their respective chips. So maybe this should be gated by a config option.

While it _can_ be stricter than required, it seems like that might hide bugs in tests/other code that is run on Sail that incorrectly relies on that behavior. Since the spec is explicit about needing an `sfence.vma` here, I don't see a need for this.

pmundkur:
Agreed that there is no immediate need.  I think if the Sail model is to be configurable to match other implementations (e.g. for compliance), we might need this parameter.

Shouldn't a test for `sfence.vma` be able to distinguish between stricter and looser implementations?

PR #1393, opened 2025-11-13
Slight refactor of address translation and PTE checks to ease implementing Zicfiss.

pmundkur:
- This explicitly separates Steps 6 and 8 of the address translation algorithm, so that the Zicfiss support in Step 7 can be cleanly inserted.

- The PTE checking API was adjusted.  Failures in steps 6 and 8 both result in page fault exceptions, and this failure cause was implicitly assumed.  However, failure in Step 7/Zicfiss results in an access fault, necessitating a need to be more explicit.  Adjusting for this explicitness has a minor impact on the `ext_` API used for the CHERI PTE and PTW hooks.  The documentation for these hooks has been updated.

Comments have been left to indicate the changes for Zicfiss.

### DIFF
--- a/model/sys/vmem_ptw.sail
+++ b/model/sys/vmem_ptw.sail
@@ -45,8 +45,8 @@ overload to_str = {ptw_error_to_str}
 // with any customization to check_PTE_permission().
 
 // PRIVATE
-function ext_get_ptw_error(eptef : pte_check_failure) -> PTW_Error =
-  match eptef {
+function ext_get_ptw_error(failure : pte_check_failure) -> PTW_Error =
+  match failure {
     PTE_No_Permission() => PTW_No_Permission(),
     PTE_Ext_Failure(_)  => PTW_No_Permission(),
     // TODO: Zicfiss will map PTE_No_Access to PTW_Access


nadime15 (approved):
I am not familiar with this part of the spec but the code looks good to me.

model/sys/vmem_pte.sail
```diff
@@ -134,8 +132,18 @@ function check_PTE_permission(ac        : MemoryAccessType(ext_access_type),
...
 
-  if access_ok & priv_ok then PTE_Check_Success(()) else PTE_Check_Failure((), ())
+  // Step 8 of VATP (see vmem.sail).
+  let access_ok : bool = match ac {
+    Load(_)             => pte_R | (pte_X & mxr),
+    Store(_)            => pte_W,
+    LoadStore(_, _)     => pte_W & (pte_R | (pte_X & mxr)),
+    InstructionFetch(_) => pte_X,
+  };
+  if not(access_ok) then PTE_Check_Failure((), PTE_No_Permission()) else PTE_Check_Success(());
```
Timmmm:
```suggestion
  if not(access_ok) then return PTE_Check_Failure((), PTE_No_Permission());
  
  PTE_Check_Success(())
```

Maybe a bit more symmetric?

model/sys/vmem_ptw.sail
```diff
@@ -45,8 +45,12 @@ overload to_str = {ptw_error_to_str}
 // with any customization to check_PTE_permission().
 
 // PRIVATE
-function ext_get_ptw_error(eptwf : ext_ptw_fail) -> PTW_Error =
-  PTW_No_Permission()
+function ext_get_ptw_error(eptef : pte_check_failure) -> PTW_Error =
```
Timmmm: `eptef`? Maybe `failure_cause : pte_check_failure` or something would be more readable?

PR #1394, opened 2025-11-14
Rename some page table walk error results for clarity.

pmundkur:
- `PTW_Access` -> `PTW_No_Access`
- `PTW_PTE_Update` -> `PTW_PTE_Needs_Update`


model/sys/vmem_ptw.sail
jordancarlin, line 23: While you're here, should we rename this one `PTW_PTE_Needs_Update`?

Issue #1353, opened 2025-10-17
Add pre-step and post-step callbacks


Timmmm:
We should add pre-step and post-step callbacks to the callback interface. Then we can move `rvfi->pre_step` and `rvfi->send_trace` into them.

Both callbacks should probably receive `is_waiting` as an argument.

Actually I think it might be a little awkward to move `rvfi->pre_step` in because its return value controls the loop. But we should provide the callback anyway. We might want to allow all callbacks to return something like `Continue | ExitSuccess | ExitFailure | Restart`.  Maybe.

nadime15:
Closing this since it was merged in #1395.

PR #1395, opened 2025-11-14
Expose per-step pre_step/post_step callbacks

ibvqeibob:
Add pre/post-step callbacks. These are currently unused but will be used in future by log/trace outputs and maybe RVFI.

- Extend `callbacks_if` with `{pre,post}_step_callback(bool is_waiting)` default no-op virtual methods.
- Add `call_{pre,post}_step_callbacks` helpers in `riscv_callbacks.{h,cpp}` to fan out the new callbacks to all registered handlers.
- Invoke `{pre,post}_step_callback` around each `ztry_step` in `run_sail()`, passing the current `is_waiting` state.

Related to: #1353

### DIFF
--- a/c_emulator/riscv_callbacks_log.cpp
+++ b/c_emulator/riscv_callbacks_log.cpp
@@ -101,15 +101,3 @@ void log_callbacks::vreg_write_callback(unsigned reg, lbits value)
     print_lbits_hex(trace_log, value);
   }
 }
-
-void log_callbacks::pre_step_callback(bool is_waiting)
-{
-  // TODO: per-step logging or statistics
-  (void)is_waiting;
-}
-
-void log_callbacks::post_step_callback(bool is_waiting)
-{
-  // TODO: post-step logging or statistics
-  (void)is_waiting;
-}
--- a/c_emulator/riscv_callbacks_log.h
+++ b/c_emulator/riscv_callbacks_log.h
@@ -10,8 +10,6 @@ public:
                 bool config_use_abi_names = false, FILE *trace_log = nullptr);
 
   // callbacks_if
-  void pre_step_callback(bool is_waiting) override;
-  void post_step_callback(bool is_waiting) override;
   void mem_write_callback(const char *type, sbits paddr, uint64_t width,
                           lbits value) override;
   void mem_read_callback(const char *type, sbits paddr, uint64_t width,


c_emulator/riscv_callbacks.cpp
```diff
@@ -18,6 +18,22 @@ void remove_callback(callbacks_if *cb)
                   callbacks.end());
 }
 
+//step 之前调用所有回调对象的 pre_step
```
nadime15: Please translate the comments into English

ibvqeibob:
> I'm curious what use case you had in mind for these callbacks. Is it worth generalizing the `pre_step` callback to include `rvfi->pre_step`?

The main use cases I had in mind for these per-step callbacks are:
1.lightweight per-step logging (e.g. PC / instruction traces),
2.collecting simple statistics (instruction counts, non-waiting steps, etc.),
3.and providing a single hook point where we can later plug in things like RVFI,debugging or breakpoint logic without touching `run_sail()` again.
Right now this PR only introduces the generic `pre_step`/`post_step` interface and wires it into the main loop, so that existing components like `log_callbacks` can start using it.

c_emulator/riscv_callbacks.cpp
```diff
@@ -18,6 +18,22 @@ void remove_callback(callbacks_if *cb)
                   callbacks.end());
 }
 
+// Invoke pre_step on all registered callbacks before each step
+void callbacks_pre_step(bool is_waiting)
```
Timmmm:
```suggestion
void pre_step_callback(bool is_waiting)
```

c_emulator/riscv_callbacks.cpp
```diff
@@ -18,6 +18,22 @@ void remove_callback(callbacks_if *cb)
...
+{
+  for (auto c : callbacks) {
+    c->pre_step(is_waiting);
+  }
+}
+
+// Invoke post_step on all registered callbacks after each step
+void callbacks_post_step(bool is_waiting)
```
Timmmm:
```suggestion
void post_step_callback(bool is_waiting)
```

c_emulator/riscv_callbacks_if.h
```diff
@@ -8,6 +8,13 @@
 class callbacks_if {
 public:
   virtual ~callbacks_if() = default;
+
+  // Callback invoked before each step
+  virtual void pre_step([[maybe_unused]] bool is_waiting) { }
```
```suggestion
  virtual void pre_step_callback([[maybe_unused]] bool is_waiting) { }
```

c_emulator/riscv_callbacks_if.h
```diff
@@ -8,6 +8,13 @@
...
+
+  // Callback invoked after each step
+  virtual void post_step([[maybe_unused]] bool is_waiting) { }
```
```suggestion
  virtual void post_step_callback([[maybe_unused]] bool is_waiting) { }
```

c_emulator/riscv_callbacks_if.h
```diff
@@ -63,3 +70,5 @@ class callbacks_if {
 
 void register_callback(callbacks_if *cb);
 void remove_callback(callbacks_if *cb);
+void callbacks_pre_step(bool is_waiting);
+void callbacks_post_step(bool is_waiting);
```
```suggestion
void call_pre_step_callbacks(bool is_waiting);
void call_post_step_callbacks(bool is_waiting);
```

Timmmm (changes requested):
Sorry for the slow review - this looks great. Just a few minor renamings to match the existing name style.

Timmmm:
> Is it worth generalizing the pre_step callback to include rvfi->pre_step?

I thought that too, but RVFI has special flow control stuff that's a bit awkward. I think we should probably just special-case it for now and think of something better in a future PR.

As for use cases, I need this exact thing for binary output support. You can collect information for a step in the callbacks then commit it in `post_step_callback()`.

@ibvqeibob let me know if you would be offended if I make the renaming changes - I suddenly have an immediate need for JSON output and it will be easier with this PR.

ibvqeibob:
> > Is it worth generalizing the pre_step callback to include rvfi->pre_step?
> 
> I thought that too, but RVFI has special flow control stuff that's a bit awkward. I think we should probably just special-case it for now and think of something better in a future PR.
> 
> As for use cases, I need this exact thing for binary output support. You can collect information for a step in the callbacks then commit it in `post_step_callback()`.
> 
> @ibvqeibob let me know if you would be offended if I make the renaming changes - I suddenly have an immediate need for JSON output and it will be easier with this PR.

Thanks a lot for the review and for pushing the renaming changes – they look great to me.
Keeping RVFI’s pre_step special-cased for now also makes sense. I’m happy that this per-step callback can help with your JSON/binary output work as well.
Please let me know if there’s anything else you’d like me to adjust in this PR.

Timmmm:
Thanks, there were a few more renamings to do and some unused variable warnings which I've fixed. LGTM now.

c_emulator/riscv_sim.cpp
Timmmm, line 457: After this is merged I think we can move almost all of this stuff into the log callback handler.

c_emulator/riscv_callbacks_log.cpp
```diff
@@ -101,3 +101,15 @@ void log_callbacks::vreg_write_callback(unsigned reg, lbits value)
...
+{
+  // TODO: per-step logging or statistics
+  (void)is_waiting;
+}
+
+void log_callbacks::post_step_callback(bool is_waiting)
+{
+  // TODO: post-step logging or statistics
+  (void)is_waiting;
+}
```
jordancarlin: No need to include the empty implementations here since we added the `[[maybe_unused]]` annotations.

jordancarlin (approved):
One minor nit, but LGTM.

c_emulator/riscv_callbacks_log.cpp
Timmmm: Oh yeah I forgot we have default implementations now. I'll delete these.

Timmmm: Done

PR #1121, opened 2025-07-02
Add Remote Bitbang protocol support for GDB/OpenOCD integration

nadime15:
This PR adds the Remote Bitbang protocol (mainly code from Spike), which is one of three components required to enable remote debugging (Remote Bitbang, the Debug Module and the Sdext Extension).

With this protocol in place, it's now possible to connect OpenOCD (compiled with Remote Bitbang support) to the simulator. OpenOCD will successfully initiate communication but will not complete the initialization process yet, as the Debug Module is still missing. That module will provide internal registers and support for accessing general-purpose registers and CSRs, both of which are not yet implemented.

To enable remote debugging, this PR also adds a new (internal for now) command-line option to Sail:

`./build/c_emulator/sail_riscv_sim -d 9824 ./test/riscv-tests/rv64ui-p-jal.elf`

The `-d <PORT>` option sets the port for OpenOCD to connect to. Sail will wait for an incoming connection, but will continue the simulation slowly.

PR #1396, opened 2025-11-14
Improved clang-format style

Timmmm:
This fixes various annoyances in our clang-format style that have been bothering me for a while:

* Binpacking arguments/parameters and aligning them to the opening `(` is *really* bad for diffs/conflict resolution. This changes it to try to match the "prettier" algorithm (also used by Sail's autoformatter), which has much simpler and better behaviour - if the arguments fit on one line do that, otherwise put each argument on its own line.
* Don't put functions on one line. This is also a little diff-unfriendly and feels unnecessarily inconsistent to me.
* The line length limit of 80 characters is very constrictive so I increased it to 120 characters.
* For some reason it didn't put `{` on a new line... except for function bodies, which is weirdly inconsistent. Now `{` never starts on a new line.
* Sort `#include`s. This might have been needed in the past but I verified it builds with sorted includes now.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,13 +1,37 @@
 Language: Cpp
-BasedOnStyle: WebKit
-AlignArrayOfStructures: Left
-AlignAfterOpenBracket: Align
-AllowShortFunctionsOnASingleLine: Empty
-ColumnLimit: 80
+
+# LLVM is the best tested style because it is used by the Clang-format developers.
+# Other styles tend to be a bit buggy.
+BasedOnStyle: LLVM
+
+# Binpacking arguments/parameters (i.e. wrapping them as if they were prose)
+# is really bad for diffing, conflicts, etc.
+BinPackArguments: false
+BinPackParameters: false
+AlignAfterOpenBracket: BlockIndent
+PackConstructorInitializers: CurrentLine
+AllowAllArgumentsOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
+
+# For consistency don't put functions or enums on one line.
+AllowShortEnumsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+
+# Slightly less claustrophobic line lengths.
+ColumnLimit: 120
+
+# This just removes some spaces in braced initialiser lists.
 Cpp11BracedListStyle: true
+
+# Indentation is 2 spaces. This matches the style in the Sail code.
 IndentWidth: 2
-PointerAlignment: Right
-SortIncludes: Never
-AlignTrailingComments: true
-AccessModifierOffset: -2
+ContinuationIndentWidth: 2
+
+# Don't allow single-statement unbraced blocks; a notorious footgun.
 InsertBraces: true
+
+# Don't split between the return type and function name, for consistency.
+PenaltyReturnTypeOnItsOwnLine: 10000
+
+# Nicer constructor lists.
+BreakConstructorInitializers: AfterColon
--- a/c_emulator/elf_loader.h
+++ b/c_emulator/elf_loader.h
@@ -25,10 +25,11 @@ public:
   // Return the entry point (where execution should begin).
   uint64_t entry() const;
 
-  using ElfWriteFn = std::function<void(uint64_t,        // address
-                                        const uint8_t *, // data
-                                        uint64_t         // length
-                                        )>;
+  using ElfWriteFn = std::function<void(
+    uint64_t,        // address
+    const uint8_t *, // data
+    uint64_t         // length
+  )>;
 
   // Call `writer()` to load all of the loadable sections into memory (or you
   // can do anything else with them).
--- a/c_emulator/riscv_callbacks_if.h
+++ b/c_emulator/riscv_callbacks_if.h
@@ -10,7 +10,7 @@ struct zMemoryAccessTypezIuzK;
 struct ztuple_z8z5enumz0zzPrivilegezCz0z5unitz9;
 struct zPTW_Error;
 
-}
+} // namespace hart
 
 class callbacks_if {
 public:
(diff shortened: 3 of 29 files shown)


Timmmm:
We should probably not merge this before the C++ support is merged and maybe a couple of the other C++ changes to avoid horrible merge conflicts.

Also it isn't perfect, e.g. it still formats `elf_loader.h` weirdly (couldn't figure out how to avoid that) but overall I think it's a lot nicer.

.clang-format
```diff
@@ -1,13 +1,40 @@
...
+
+# LLVM is the best tested style because it is used by the Clang-format developers.
+# Other styles tend to be a bit buggy.
+BasedOnStyle: LLVM
+
+# Binpacking arguments/parameters (i.e. wrapping them as if they were prose)
+# is really bad for diffing, conflicts, etc.
+BinPackArguments: false
+BinPackParameters: false
+AlignAfterOpenBracket: BlockIndent
```
arichardson, line 11: Looks like this was changed upstream and now has a different way of setting it: https://github.com/llvm/llvm-project/issues/80049.

How bad does it look if we keep these three parameters as the LLVM style default?

.clang-format
```diff
@@ -1,13 +1,40 @@
...
+PackConstructorInitializers: CurrentLine
+AllowAllArgumentsOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
+
+# For consistency don't put functions on one line.
```
arichardson:
```suggestion
# For consistency don't put functions or enums on one line.
```

.clang-format
```diff
@@ -1,13 +1,40 @@
...
-SortIncludes: Never
-AlignTrailingComments: true
-AccessModifierOffset: -2
+ContinuationIndentWidth: 2
+
+# Don't allow single-statement unbraced blocks; a notorious footgun.
 InsertBraces: true
+
+# Pretty arbitrary & down to taste.
+PointerAlignment: Right
```
arichardson: I think this is already the default.

.clang-format
Timmmm: It lays out the arguments like prose which I really don't like. Which bit of that issue did you mean? I can't see anything that has actually changed...

But now that I think about it, it's probably worth updating clang-format before we make this change. I'll make a separate PR for that.

arichardson: I imagine there is backwards compatibility so let's just keep it the way it is now

c_emulator/elf_loader.h
```diff
@@ -27,10 +27,8 @@ class ELF {
 
   // Call `writer()` to load all of the loadable sections into memory (or you
   // can do anything else with them).
-  void
-  load(std::function<void(uint64_t /* address */, const uint8_t * /* data */,
-                          uint64_t /* length */)>
-           writer) const;
+  void load(std::function<void(uint64_t /* address */, const uint8_t * /* data */, uint64_t /* length */)> writer
```
arichardson: Maybe this weird formatting can be avoided by having 

`using ElfWriteFn = std::function<void(uint64_t /* address */, const uint8_t * /* data */, uint64_t /* length */);`

somewhere further up?

Timmmm: Ooo good idea, I'll try that.

Timmmm: That works, great idea!

Timmmm:
Ok if you look at the individual commits there's one that changes the line length to 120 and then another that applies the rest of the changes.

c_emulator/riscv_callbacks_if.h
```diff
@@ -9,55 +9,51 @@ class callbacks_if {
...
   virtual ~callbacks_if() = default;
   // TODO: finding ways to improve the format
-  virtual void fetch_callback([[maybe_unused]] sbits opcode) { }
-  virtual void mem_write_callback([[maybe_unused]] const char *type,
-                                  [[maybe_unused]] sbits paddr,
-                                  [[maybe_unused]] uint64_t width,
-                                  [[maybe_unused]] lbits value)
-  {
+  virtual void fetch_callback([[maybe_unused]] sbits opcode) {
   }
```
pmundkur: Is there an override to keep the `{  }` for empty functions on the same line?

Timmmm: Yeah but I like it more like this. Makes it easier to add code if you want and it's one fewer special case.

c_emulator/riscv_callbacks_log.cpp
```diff
@@ -18,38 +17,48 @@ void print_lbits_hex(FILE *trace_log, lbits val, int length = 0)
...
 
-log_callbacks::log_callbacks(bool config_print_reg, bool config_print_mem_access, bool config_use_abi_names,
-                             FILE *trace_log)
-    : config_print_reg(config_print_reg)
-    , config_print_mem_access(config_print_mem_access)
-    , config_use_abi_names(config_use_abi_names)
-    , trace_log(trace_log)
-{
+log_callbacks::log_callbacks(
+  bool config_print_reg,
```
arichardson, line 21: I'm not sure about the one argument per line change: it's good for looking at diffs but does add a lot of extra newlines. I don't really mind though.

pmundkur:
@Timmmm now that the C++ backend is in, perhaps we could rebase this and merge?

Timmmm:
Yeah probably makes sense. I think the only big outstanding C++ changes is #1121 but it's mostly new files. Also there's a sequence of Git commands you can do to avoid having to manually resolve merge conflicts for formatting changes.

Timmmm:
I separated out #1448 since that is the only non-formatting change here.

Timmmm:
Ok this is ready now. If anyone encounters this and it gives them horrible merge/rebase conflicts here's what you can do:

1. Rebase/merge to just before this PR is merged.
2. Copy the new `.clang-format` into your working tree, and run `pre-commit run -a` to format all your code with the new style. Commit that.
3. Rebase/merge to this PR. You'll get a load of conflicts but here you can always accept in your favour. There's a way to do that automatically but I'd have to look it up, something like `git merge <this PR> -S ours`.
4. Probably squash your history because it will be a mess.

PR #1399, opened 2025-11-18
Fix satp legalization to use Supervisor instead of current privilege

nadime15:
When M-mode writes to satp, using `architecture(cur_privilege)` would incorrectly use misa.MXL instead of mstatus.SXL.

Since MXLEN==SXLEN currently, this has no functional effect, but the current code is misleading, satp depends on mstatus.SXL, not misa.MXL regardless of current mode.

### DIFF
--- a/model/sys/vmem.sail
+++ b/model/sys/vmem.sail
@@ -164,7 +164,7 @@ register satp : xlenbits
 mapping clause csr_name_map = 0x180  <-> "satp"
 function clause is_CSR_accessible(0x180, priv, _) = currentlyEnabled(Ext_S) & not(priv == Supervisor & mstatus[TVM] == 0b1)
 function clause read_CSR(0x180) = satp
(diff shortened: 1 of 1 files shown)


Timmmm:
The spec is not at all specific about this but I think this is correct.

Timmmm:
Btw how did you discover this?

jordancarlin (approved):

nadime15:
> The spec is not at all specific about this but I think this is correct.

I would argue it is, because Satp is defined in terms of SXLEN, which in turn is defined by mstatus.SXL. If you have MXLEN=64 but SXLEN=32 (once supported), the previous code would call satpMode_of_bits(RV64, ...) (if in M-Mode) even though satp is only 32 bits wide. I think the proper way would be to switch to M-mode, set SXLEN to 64, and then proceed.

> Btw how did you discover this?

By chance 😄 I noticed that while working on the debug stuff. It looked a bit weird, so I looked it up.

PR #1402, opened 2025-11-20
Enhance CSR access result granularity and report illegal access reasons

trdthg:
More aggressive version of #1357 

In this version, I overloaded `&` and `|` for `CSRAccessResult` to allow for bitwise operations, which allow me preserve most of the origin check logic.

```sail
currentlyEnabled(Ext_U) & (xlen == 32)
                    |
                    V
CSRRequireExt(Ext_U) & CSRRequireXlen(32)
```

One drawback is that expressions like `CSRRequire(A) | CSRRequire(B)` can't be fully returned unless we allow recursion

```sail
union CSRAccessResult {
	...
	CSRAccessResultOr : (CSRAccessResult, CSRAccessResult)
}
```

### DIFF
--- a/model/core/csr_begin.sail
+++ b/model/core/csr_begin.sail
@@ -15,9 +15,76 @@ scattered mapping csr_name_map
 function csr_name(csr : csreg) -> string = csr_name_map(csr)
 overload to_str = {csr_name}
 
+union CSRAccessIllegalReason = {
+  // Attempt to perform an unsupported operation (e.g., write to a read-only CSR)
+  CSR_Illegal_AccessType : CSRAccessType,
+
+  // Current privilege level is lower than the CSR's minimum required privilege
+  CSR_Illegal_Privilege : Privilege,
+
+  // Extension is not enabled
+  CSR_Illegal_ExtDisabled : extension,
+
+  // CSR is not accessible in the current XLEN mode (e.g., RV32-only CSR in RV64)
+  CSR_Illegal_XlenMismatch : unit,
+
+  // Counter CSR bit not enabled
+  CSR_Illegal_Counter : (range(0, 31), Privilege),
+
+  // Attempted read without a write raises an illegal-instruction exception
+  // regardless of mode and access control bits
+  CSR_Illegal_SeedReadWithoutWrite : unit,
+  // Attempted access seed csr, but mseccfg[XSEED] is not enabled
+  CSR_Illegal_XSEEDNotEnabled : Privilege,
+
+  // Used for one or more specific CSRS that are inaccessible due to complex conditions
+  // (e.g., OR, or privilege mixed bit checks).
+  CSR_Illegal_FCSR    : unit,
+  CSR_Illegal_Mseccfg : unit,
+  CSR_Illegal_Satp    : unit,
+  CSR_Illegal_Stateen : unit,
+  CSR_Illegal_Sstc    : unit,
+
+  // CSR is not defined in spec or not implemented by sail-model yet
+  CSR_Illegal_NotDefined : unit,
+}
+
+union CSRAccessResult = {
+  CSRAccess_Success : unit,
+
+  // Accessing the CSR is illegal currently and should raise Illegal_Instruction Exception
+  CSRAccess_Illegal : CSRAccessIllegalReason,
+
+  // Like CSRAccess_Illegal, but raise Virtual_Instruction Exception
+  CSRAccess_Virtual : unit,
+}
+
 // returns whether a CSR exists
-val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> bool
+val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> CSRAccessResult
 scattered function is_CSR_accessible
+
+function csr_defined_if(exist : bool) -> CSRAccessResult =
+  if   exist
+  then CSRAccess_Success()
+  else CSRAccess_Illegal(CSR_Illegal_NotDefined())
+
+// CSR require ext is enabled.
+function csr_requires_ext(ext : extension) -> CSRAccessResult =
+  if   currentlyEnabled(ext)
+  then CSRAccess_Success()
+  else CSRAccess_Illegal(CSR_Illegal_ExtDisabled(ext))
+
+function csr_requires_xlen_32() -> CSRAccessResult =
+  if   xlen == 32
+  then CSRAccess_Success()
+  else CSRAccess_Illegal(CSR_Illegal_XlenMismatch())
+
+function csr_access_result_and(r1 : CSRAccessResult, r2 : CSRAccessResult) -> CSRAccessResult =
+  if      r1 != CSRAccess_Success() then r1
+  else if r2 != CSRAccess_Success() then r2
+  else    CSRAccess_Success()
+
+overload operator & = { csr_access_result_and }
 
 // returns the value of the CSR if it is defined
 val read_CSR : csreg -> xlenbits
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -124,7 +124,7 @@ private function legalize_misa(m : Misa, v : xlenbits) -> Misa = {
 }
 
 mapping clause csr_name_map = 0x301  <-> "misa"
-function clause is_CSR_accessible(0x301, _, _) = true // misa
+function clause is_CSR_accessible(0x301, _, _) = CSRAccess_Success() // misa
 function clause read_CSR(0x301) = misa.bits
 function clause write_CSR(0x301, value) = { misa = legalize_misa(misa, value); Ok(misa.bits) }
 
@@ -269,8 +269,8 @@ register mstatus : Mstatus = {
 mapping clause csr_name_map = 0x300  <-> "mstatus"
 mapping clause csr_name_map = 0x310  <-> "mstatush"
 
-function clause is_CSR_accessible(0x300, _, _) = true // mstatus
-function clause is_CSR_accessible(0x310, _, _) = xlen == 32 // mstatush
+function clause is_CSR_accessible(0x300, _, _) = CSRAccess_Success() // mstatus
+function clause is_CSR_accessible(0x310, _, _) = csr_requires_xlen_32() // mstatush
 
 function clause read_CSR(0x300) = mstatus.bits[xlen - 1 .. 0]
 function clause read_CSR(0x310 if xlen == 32) = mstatus.bits[63 .. 32]
@@ -331,8 +331,13 @@ mapping clause csr_name_map = 0x747  <-> "mseccfg"
 mapping clause csr_name_map = 0x757  <-> "mseccfgh"
 
 // "mseccfg exists if Zkr is implemented, or if it is required by other processor features."
-function clause is_CSR_accessible(0x747, _, _) = currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp) // mseccfg
-function clause is_CSR_accessible(0x757, _, _) = (currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp)) & xlen == 32 // mseccfgh
+function mseccfg_csr_accessible() -> CSRAccessResult =
+  if (currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp))
+  then CSRAccess_Success()
+  else CSRAccess_Illegal(CSR_Illegal_Mseccfg())
+
+function clause is_CSR_accessible(0x747, _, _) = mseccfg_csr_accessible() // mseccfg
+function clause is_CSR_accessible(0x757, _, _) = mseccfg_csr_accessible() & csr_requires_xlen_32() // mseccfgh
 
 function clause read_CSR(0x747) = mseccfg.bits[xlen - 1 .. 0]
 function clause read_CSR(0x757 if xlen == 32) = mseccfg.bits[63 .. 32]
@@ -424,12 +429,12 @@ mapping clause csr_name_map = 0x30A  <-> "menvcfg"
 mapping clause csr_name_map = 0x31A  <-> "menvcfgh"
 mapping clause csr_name_map = 0x10A  <-> "senvcfg"
 
-function clause is_CSR_accessible(0x30A, _, _) = currentlyEnabled(Ext_U) // menvcfg
-function clause is_CSR_accessible(0x31A, _, _) = currentlyEnabled(Ext_U) & (xlen == 32) // menvcfgh
+function clause is_CSR_accessible(0x30A, _, _) = csr_requires_ext(Ext_U) // menvcfg
+function clause is_CSR_accessible(0x31A, _, _) = csr_requires_ext(Ext_U) & csr_requires_xlen_32() // menvcfgh
 
 // This should ideally also check `xstateen.ENVCFG`, but due to module
 // interdependency issues, that is handled by `stateen_allows_CSR_access`.
-function clause is_CSR_accessible(0x10A, _, _) = currentlyEnabled(Ext_S) // senvcfg
+function clause is_CSR_accessible(0x10A, _, _) = csr_requires_ext(Ext_S) // senvcfg
 
 function clause read_CSR(0x30A) = menvcfg.bits[xlen - 1 .. 0]
 function clause read_CSR(0x31A if xlen == 32) = menvcfg.bits[63 .. 32]
@@ -549,11 +554,11 @@ mapping clause csr_name_map = 0x302  <-> "medeleg"
 mapping clause csr_name_map = 0x312  <-> "medelegh"
 mapping clause csr_name_map = 0x303  <-> "mideleg"
 
-function clause is_CSR_accessible(0x304, _, _) = true // mie
-function clause is_CSR_accessible(0x344, _, _) = true // mip
-function clause is_CSR_accessible(0x302, _, _) = currentlyEnabled(Ext_S) // medeleg
-function clause is_CSR_accessible(0x312, _, _) = currentlyEnabled(Ext_S) & xlen == 32 // medelegh
-function clause is_CSR_accessible(0x303, _, _) = currentlyEnabled(Ext_S) // mideleg
+function clause is_CSR_accessible(0x304, _, _) = CSRAccess_Success() // mie
+function clause is_CSR_accessible(0x344, _, _) = CSRAccess_Success() // mip
+function clause is_CSR_accessible(0x302, _, _) = csr_requires_ext(Ext_S) // medeleg
+function clause is_CSR_accessible(0x312, _, _) = csr_requires_ext(Ext_S) & csr_requires_xlen_32() // medelegh
+function clause is_CSR_accessible(0x303, _, _) = csr_requires_ext(Ext_S) // mideleg
 
 function clause read_CSR(0x304) = mie.bits
 function clause read_CSR(0x344) = mip.bits
@@ -592,7 +597,7 @@ bitfield Mcause : xlenbits = {
 }
 register mcause : Mcause
 mapping clause csr_name_map = 0x342  <-> "mcause"
-function clause is_CSR_accessible(0x342, _, _) = true // mcause
+function clause is_CSR_accessible(0x342, _, _) = CSRAccess_Success() // mcause
 function clause read_CSR(0x342) = mcause.bits
 function clause write_CSR(0x342, value) = { mcause.bits = value; Ok(mcause.bits) }
 
@@ -639,8 +644,8 @@ register mscratch : xlenbits
 mapping clause csr_name_map = 0x343  <-> "mtval"
 mapping clause csr_name_map = 0x340  <-> "mscratch"
 
-function clause is_CSR_accessible(0x343, _, _) = true // mtval
-function clause is_CSR_accessible(0x340, _, _) = true // mscratch
+function clause is_CSR_accessible(0x343, _, _) = CSRAccess_Success() // mtval
+function clause is_CSR_accessible(0x340, _, _) = CSRAccess_Success() // mscratch
 
 function clause read_CSR(0x343) = mtval
 function clause read_CSR(0x340) = mscratch
@@ -665,7 +670,7 @@ private function legalize_scounteren(_c : Counteren, v : xlenbits) -> Counteren 
 
 register scounteren : Counteren
 mapping clause csr_name_map = 0x106  <-> "scounteren"
-function clause is_CSR_accessible(0x106, _, _) = currentlyEnabled(Ext_S) // scounteren
+function clause is_CSR_accessible(0x106, _, _) = csr_requires_ext(Ext_S) // scounteren
 function clause read_CSR(0x106) = zero_extend(scounteren.bits)
 function clause write_CSR(0x106, value) = { scounteren = legalize_scounteren(scounteren, value); Ok(zero_extend(scounteren.bits)) }
 
@@ -677,7 +682,7 @@ function legalize_mcounteren(_c : Counteren, v : xlenbits) -> Counteren = {
 
 register mcounteren : Counteren
 mapping clause csr_name_map = 0x306  <-> "mcounteren"
-function clause is_CSR_accessible(0x306, _, _) = currentlyEnabled(Ext_U) // mcounteren
+function clause is_CSR_accessible(0x306, _, _) = csr_requires_ext(Ext_U) // mcounteren
 function clause read_CSR(0x306) = zero_extend(mcounteren.bits)
 function clause write_CSR(0x306, value) = { mcounteren = legalize_mcounteren(mcounteren, value); Ok(zero_extend(mcounteren.bits)) }
 
@@ -697,7 +702,7 @@ private function legalize_mcountinhibit(_c : Counterin, v : xlenbits) -> Counter
 
 register mcountinhibit : Counterin
 mapping clause csr_name_map = 0x320  <-> "mcountinhibit"
-function clause is_CSR_accessible(0x320, _, _) = true // mcountinhibit
+function clause is_CSR_accessible(0x320, _, _) = CSRAccess_Success() // mcountinhibit
 function clause read_CSR(0x320) = zero_extend(mcountinhibit.bits)
 function clause write_CSR(0x320, value) = { mcountinhibit = legalize_mcountinhibit(mcountinhibit, value); Ok(zero_extend(mcountinhibit.bits)) }
 
@@ -733,11 +738,11 @@ mapping clause csr_name_map = 0xF13  <-> "mimpid"
 mapping clause csr_name_map = 0xF14  <-> "mhartid"
 mapping clause csr_name_map = 0xF15  <-> "mconfigptr"
 
-function clause is_CSR_accessible(0xf11, _, _) = true // mvendorid
-function clause is_CSR_accessible(0xf12, _, _) = true // marchdid
-function clause is_CSR_accessible(0xf13, _, _) = true // mimpid
-function clause is_CSR_accessible(0xf14, _, _) = true // mhartid
-function clause is_CSR_accessible(0xf15, _, _) = true // mconfigptr
+function clause is_CSR_accessible(0xf11, _, _) = CSRAccess_Success() // mvendorid
+function clause is_CSR_accessible(0xf12, _, _) = CSRAccess_Success() // marchdid
+function clause is_CSR_accessible(0xf13, _, _) = CSRAccess_Success() // mimpid
+function clause is_CSR_accessible(0xf14, _, _) = CSRAccess_Success() // mhartid
+function clause is_CSR_accessible(0xf15, _, _) = CSRAccess_Success() // mconfigptr
 
 function clause read_CSR(0xF11) = zero_extend(mvendorid)
 function clause read_CSR(0xF12) = marchid
@@ -809,7 +814,7 @@ private function legalize_sstatus(m : Mstatus, v : xlenbits) -> Mstatus = {
 }
 
 mapping clause csr_name_map = 0x100  <-> "sstatus"
-function clause is_CSR_accessible(0x100, _, _) = currentlyEnabled(Ext_S) // sstatus
+function clause is_CSR_accessible(0x100, _, _) = csr_requires_ext(Ext_S) // sstatus
 function clause read_CSR(0x100) = lower_mstatus(mstatus).bits[xlen - 1 .. 0]
 function clause write_CSR(0x100, value) = { mstatus = legalize_sstatus(mstatus, value); Ok(lower_mstatus(mstatus).bits[xlen - 1 .. 0]) }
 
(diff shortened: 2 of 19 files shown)


pmundkur (commented):
I wonder if CSRs should be represented by `struct`s that contain their required extension and xlen constraints; that might simplify the access checking logic.  It would perhaps also make it easier to extract CSR constraint information for the JSON documentation bundle instead of having it spread out across all these scattered access checking functions.

trdthg:
Considering the complexity of Or, I removed the overload `overload operator & = { csr_access_result_or }`, now they are changed to `csr_require_cond(cond : bool, reason : string)` which we should manually attach a reason.
I also removed the `CSRRequireHartSupported` reason, because the only place that uses it also requires Or.

model/core/sys_regs.sail
```diff
@@ -331,8 +331,10 @@ mapping clause csr_name_map = 0x747  <-> "mseccfg"
 mapping clause csr_name_map = 0x757  <-> "mseccfgh"
 
 // "mseccfg exists if Zkr is implemented, or if it is required by other processor features."
-function clause is_CSR_accessible(0x747, _, _) = currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp) // mseccfg
-function clause is_CSR_accessible(0x757, _, _) = (currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp)) & xlen == 32 // mseccfgh
+function clause is_CSR_accessible(0x747, _, _) = csr_require_cond(currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp),
+  "Zkr is not implemented or Zicfilp is supported") // mseccfg
```
```suggestion
  "Zkr is not implemented or Zicfilp is not supported") // mseccfg
```

model/core/sys_regs.sail
```diff
@@ -331,8 +331,10 @@ mapping clause csr_name_map = 0x747  <-> "mseccfg"
...
+function clause is_CSR_accessible(0x757, _, _) = csr_require_cond(currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp),
+  "Zkr is not implemented or Zicfilp is supported.") & csr_require_xlen(32) // mseccfgh
```
```suggestion
  "Zkr is not implemented or Zicfilp is not supported.") & csr_require_xlen(32) // mseccfgh
```

model/extensions/K/zkr_control.sail
```diff
@@ -42,15 +42,23 @@ function write_seed_csr () -> xlenbits = zeros()
...
+    Supervisor => csr_require_cond(mseccfg[SSEED] == 0b1, "Access seed in S mode require (mseccfg[SSEED] == 0b1)"),
+    User => csr_require_cond(mseccfg[USEED] == 0b1, "Access seed in U mode require (mseccfg[USEED] == 0b1)"),
+    // TODO: wait for Hypervisor extension
     VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
     VirtualUser => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    // VirtualSupervisor => CSRAccess_Virtual(),
+    // VirtualUser => match access_type {
+    //   CSRReadWrite => CSRAccess_Virtual(),
+    //   _ => CSRAccess_Illegal()
+    // },
```
pmundkur: That's a nice solution for handling this corner case!

model/core/csr_begin.sail
```diff
@@ -15,10 +15,54 @@ scattered mapping csr_name_map
...
+}
+
 // returns whether a CSR exists
-val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> bool
+val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> CSRAccessResult
 scattered function is_CSR_accessible
 
+function csr_defined_if(exist : bool) -> CSRAccessResult = if exist then CSRAccess_Success() else CSRAccess_Illegal(CSR_Illegal_NotDefined())
+
+function csr_require_ext(ext : extension) -> CSRAccessResult = if currentlyEnabled(ext) then CSRAccess_Success() else CSRAccess_Illegal(CSR_Illegal_ExtDisabled())
```
pmundkur: Did you think we may need something similar for `hartSupports`?  It's not clear that `require` corresponds to `currentlyEnabled`, but otherwise the approach seems useful.

model/extensions/Sstc/sstc.sail
```diff
@@ -10,16 +10,18 @@
 mapping clause csr_name_map = 0x14D  <-> "stimecmp"
 mapping clause csr_name_map = 0x15D  <-> "stimecmph"
 
-function sstc_CSRs_accessible(priv : Privilege) -> bool =
-    priv == Machine | (priv == Supervisor & mcounteren[TM] == 0b1 & menvcfg[STCE] == 0b1)
+function sstc_CSRs_accessible(priv : Privilege) -> CSRAccessResult =
+    csr_require_cond((priv == Machine) | (priv == Supervisor & mcounteren[TM] == 0b1 & menvcfg[STCE] == 0b1),
+        "Access sstc CSRs require (mcounteren[TM] == 0b1 & menvcfg[STCE] == 0b1) in S mode")
```
pmundkur:
```suggestion
        "Accessing sstc CSRs requires (mcounteren[TM] == 0b1 & menvcfg[STCE] == 0b1) in S mode")
```

model/sys/vmem.sail
```diff
@@ -182,7 +182,7 @@ termination_measure pt_walk(_,_,_,_,_,_,_,level,_, _) = level
 // PUBLIC: see also zicsr_insts.sail and other CSR-related files
 register satp : xlenbits
 mapping clause csr_name_map = 0x180  <-> "satp"
-function clause is_CSR_accessible(0x180, priv, _) = currentlyEnabled(Ext_S) & not(priv == Supervisor & mstatus[TVM] == 0b1)
+function clause is_CSR_accessible(0x180, priv, _) = csr_require_ext(Ext_S) & csr_require_cond(not(priv == Supervisor & mstatus[TVM] == 0b1), "Access satp require (priv == Supervisor & mstatus[TVM] == 0b1)")
```
pmundkur:
```suggestion
function clause is_CSR_accessible(0x180, priv, _) = csr_require_ext(Ext_S) & csr_require_cond(not(priv == Supervisor & mstatus[TVM] == 0b1), "Accessing satp requires (priv == Supervisor & mstatus[TVM] == 0b1)")
```

model/extensions/Zicsr/zicsr_insts.sail
```diff
@@ -40,33 +40,35 @@ function csr_access_type(op : csrop, rd_is_x0 : bool, rs1_imm_is_zero : bool) ->
...
-      }
-    } else {
-      csr_read_callback(csr, csr_val);
-      X(rd) = csr_val;
-      RETIRE_SUCCESS
+function doCSR(csr : csreg, rs1_val : xlenbits, rd : regidx, op : csrop, access_type: CSRAccessType) -> ExecutionResult = {
+  match check_CSR(csr, cur_privilege, access_type) {
+    CSRAccess_Success() => (),
+    CSRAccess_Illegal(_) => return Illegal_Instruction(),
+    // CSRAccess_Virtual() => return Virtual_Instruction(),
```
pmundkur: Yes, that's something like what was needed!

model/core/csr_begin.sail
trdthg: `hartSupports` is only used with by `mseccfg/mseccfgh` now.

```sail
// "mseccfg exists if Zkr is implemented, or if it is required by other processor features."
function clause is_CSR_accessible(0x747, _, _) = 
	csr_require_cond(currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp),
  		"Zkr is not implemented or Zicfilp is not supported") // mseccfg
```

> Did you think we may need something similar for hartSupports?

Not needed for now. I’m not sure how we would structure the `|` condition, once we figure this out, we can revisit whether it’s needed. Since I have already removed `overload operator & = { csr_access_result_or }`, `csr_require_ext_hart_supported` is no longer needed, and `csr_require_ext_currently_enabled` is a bit too lang for me. :P

nadime15:
@trdthg Any updates on this? Is anything missing, or does this PR just need to be reviewed? It would be good to move this forward.

model/extensions/K/zkr_control.sail
```diff
@@ -42,13 +42,16 @@ function write_seed_csr () -> xlenbits = zeros()
...
+  csr_require_ext(Ext_Zkr)
+  & (match priv {
+    Machine => match access_type {
+      // Read-only access is not allowed.
+      CSRRead => CSRAccess_Illegal(CSR_Illegal_AccessType()),
+      _ => CSRAccess_Success(),
+    },
+    Supervisor => csr_require_cond(mseccfg[SSEED] == 0b1, "Access seed in S mode require (mseccfg[SSEED] == 0b1)"),
+    User => csr_require_cond(mseccfg[USEED] == 0b1, "Access seed in U mode require (mseccfg[USEED] == 0b1)"),
+    // TODO: wait for Hypervisor extension
```
nadime15, line 51: I am not familiar with Zkr, but I noticed that access_type is now only checked when the hart is currently in M-mode. Was this intended?

trdthg: Yes, This is intended, because in this table https://github.com/riscv/sail-riscv/issues/1350#issuecomment-3413591993
Only in M mode is described `Attempted read without a write raises an illegal-instruction exception`
I'm not quite sure why this check was applied to every mode before. (Perhaps there are other descriptions in the manual that I missed. :D)

nadime15: It's a bit confusing because the table says for M-Mode

> The seed CSR is always available in machine mode as normal (with a CSR read-write instruction.) Attempted read without a write raises an illegal-instruction exception **regardless of mode** (!) and access control bits.

trdthg: I've reverted to the original logic so we can move forward. I guess it’s just a specification issue. I will open an Clarification issue later. https://github.com/riscv/riscv-isa-manual/issues/2588

model/exceptions/sys_exceptions.sail
```diff
@@ -85,10 +85,10 @@ mapping clause csr_name_map = 0x141  <-> "sepc"
...
 mapping clause csr_name_map = 0x341  <-> "mepc"
 
-function clause is_CSR_accessible(0x105, _, _) = currentlyEnabled(Ext_S) // stvec
-function clause is_CSR_accessible(0x141, _, _) = currentlyEnabled(Ext_S) // sepc
-function clause is_CSR_accessible(0x305, _, _) = true // mtvec
-function clause is_CSR_accessible(0x341, _, _) = true // mepc
+function clause is_CSR_accessible(0x105, _, _) = csr_require_ext(Ext_S) // stvec
+function clause is_CSR_accessible(0x141, _, _) = csr_require_ext(Ext_S) // sepc
+function clause is_CSR_accessible(0x305, _, _) = CSRAccess_Success()// mtvec
+function clause is_CSR_accessible(0x341, _, _) = CSRAccess_Success()// mepc
```
arichardson:
```suggestion
function clause is_CSR_accessible(0x305, _, _) = CSRAccess_Success() // mtvec
function clause is_CSR_accessible(0x341, _, _) = CSRAccess_Success() // mepc
```

arichardson (approved):

model/extensions/FD/fdext_control.sail
```diff
@@ -26,9 +26,9 @@ mapping clause csr_name_map = 0x003  <-> "fcsr"
 
 // These should ideally also check `xstateen.FCSR`, but due to module
 // interdependency issues, that is handled by `stateen_allows_CSR_access`.
-function clause is_CSR_accessible (0x001, _, _) = currentlyEnabled(Ext_F) | currentlyEnabled(Ext_Zfinx)
-function clause is_CSR_accessible (0x002, _, _) = currentlyEnabled(Ext_F) | currentlyEnabled(Ext_Zfinx)
-function clause is_CSR_accessible (0x003, _, _) = currentlyEnabled(Ext_F) | currentlyEnabled(Ext_Zfinx)
+function clause is_CSR_accessible (0x001, _, _) = csr_require_cond(currentlyEnabled(Ext_F) | currentlyEnabled(Ext_Zfinx), "Neither F nor Zfinx is enabled")
+function clause is_CSR_accessible (0x002, _, _) = csr_require_cond(currentlyEnabled(Ext_F) | currentlyEnabled(Ext_Zfinx), "Neither F nor Zfinx is enabled")
+function clause is_CSR_accessible (0x003, _, _) = csr_require_cond(currentlyEnabled(Ext_F) | currentlyEnabled(Ext_Zfinx), "Neither F nor Zfinx is enabled")
```
nadime15: Does it make sense to overload OR too and instead of bundling them in `csr_require_cond` we use 2 times `csr_require_ext()`? I think this would describe the case better.

The only potential drawback I can see (and I am not even sure this is actually an issue) is that it is unclear what happens if multiple illegal conditions apply simultaneously. Which illegal reason is returned, and how is are they prioritized?

model/core/csr_begin.sail
```diff
@@ -15,10 +15,68 @@ scattered mapping csr_name_map
...
+// often requires more complex logic (like or), for these condition, the
+// recommended way is to use 'csr_require_cond' with a custom description string.
+function csr_require_ext(ext : extension) -> CSRAccessResult =
+  if   currentlyEnabled(ext)
+  then CSRAccess_Success()
+  else CSRAccess_Illegal(CSR_Illegal_ExtDisabled())
+function csr_require_xlen(len : int) -> CSRAccessResult =
+  if   xlen == len
+  then CSRAccess_Success()
+  else CSRAccess_Illegal(CSR_Illegal_XlenMismatch())
```
nadime15: > 
```suggestion
function csr_require_ext(ext : extension) -> CSRAccessResult =
  if   currentlyEnabled(ext)
  then CSRAccess_Success()
  else CSRAccess_Illegal(CSR_Illegal_ExtDisabled())
 
function csr_require_xlen(len : int) -> CSRAccessResult =
  if   xlen == len
  then CSRAccess_Success()
  else CSRAccess_Illegal(CSR_Illegal_XlenMismatch())
```

model/extensions/Stateen/stateen_regs.sail
```diff
@@ -9,9 +9,9 @@
 function clause currentlyEnabled(Ext_Smstateen) = hartSupports(Ext_Smstateen) & currentlyEnabled(Ext_Zicsr)
 function clause currentlyEnabled(Ext_Ssstateen) = hartSupports(Ext_Ssstateen) & currentlyEnabled(Ext_Zicsr)
 
-function is_mstateen_accessible() -> bool = hartSupports(Ext_Smstateen)
-function is_hstateen_accessible() -> bool = currentlyEnabled(Ext_H) & (currentlyEnabled(Ext_Smstateen) | currentlyEnabled(Ext_Ssstateen))
-function is_sstateen_accessible() -> bool = currentlyEnabled(Ext_S) & (currentlyEnabled(Ext_Smstateen) | currentlyEnabled(Ext_Ssstateen))
+function is_mstateen_accessible() -> CSRAccessResult = csr_require_cond(hartSupports(Ext_Smstateen), "Smstateen extension is not supported")
+function is_hstateen_accessible() -> CSRAccessResult = csr_require_cond(currentlyEnabled(Ext_H) & (currentlyEnabled(Ext_Smstateen) | currentlyEnabled(Ext_Ssstateen)), "Hypervisor or Smstateen/Ssstateen extension is not enabled")
+function is_sstateen_accessible() -> CSRAccessResult = csr_require_cond(currentlyEnabled(Ext_S) & (currentlyEnabled(Ext_Smstateen) | currentlyEnabled(Ext_Ssstateen)), "S-mode or Smstateen/Ssstateen extension is not enabled")
```
nadime15: It's just me but I am not a big fan of modifying anything else than the is_CSR_accessible clauses directly. Would it be better to just to do something like 

`... & (if is_hstateen_accessible() then CSRAccess_Success() else CSRAccess_Illegal(CSR_Illegal_ExtDisabled())`

?

nadime15 (approved):

nadime15:
Ups, I accidentally approved the PR.

model/extensions/FD/fdext_control.sail
trdthg: > Which illegal reason is returned, and how is are they prioritized?

It only returns the top result, based the relative positions of the '&' operator.

```sail
function csr_access_result_and(r1 : CSRAccessResult, r2 : CSRAccessResult) -> CSRAccessResult =
  if      r1 != CSRAccess_Success() then r1
  else if r2 != CSRAccess_Success() then r2
  else    CSRAccess_Success()
```

> The only potential drawback I can see (and I am not even sure this is actually an issue) 

I think it's not a big issue for 'and' conditions, as you can eventually catch all the issues through multiple rounds of debugging. But it would gets much trickier with 'or' conditions, so I would suggest to prvice a reason string for complex conditions.

model/extensions/Stateen/stateen_regs.sail
trdthg: Then you have to copy and paste a lot for them

```sail
function clause is_CSR_accessible(0x30C, _, _) = is_mstateen_accessible()
function clause is_CSR_accessible(0x30D, _, _) = is_mstateen_accessible()
function clause is_CSR_accessible(0x30E, _, _) = is_mstateen_accessible()
function clause is_CSR_accessible(0x30F, _, _) = is_mstateen_accessible()

function clause is_CSR_accessible(0x60C, _, _) = is_hstateen_accessible()
function clause is_CSR_accessible(0x60D, _, _) = is_hstateen_accessible()
function clause is_CSR_accessible(0x60E, _, _) = is_hstateen_accessible()
function clause is_CSR_accessible(0x60F, _, _) = is_hstateen_accessible()
...
```

And I think making another wrapper for `(if is_hstateen_accessible() then CSRAccess_Success() else CSRAccess_Illegal(CSR_Illegal_ExtDisabled())` is really not necessary.

model/core/csr_begin.sail
```diff
@@ -15,10 +15,69 @@ scattered mapping csr_name_map
...
+  CSR_Illegal_Privilege : unit,
+
+  // Extension is not enabled
+  CSR_Illegal_ExtDisabled : unit,
+
+  // CSR is not accessible in the current XLEN mode (e.g., RV32-only CSR in RV64)
+  CSR_Illegal_XlenMismatch : unit,
+
+  // Other complex constraints not covered above, describe the specific reason using a string
+  CSR_Illegal_Condition : string,
```
Timmmm: I think we should just explicitly list them all. We've tried to avoid strings wherever possible in the model for various reasons and I can't see why we couldn't just have

```
CSR_Illegal_Stateen : (Privilege, range(0, 31)),
CSR_Illegal_SeedReadWithoutWrite : unit,
...
```

model/core/csr_begin.sail
```diff
@@ -15,10 +15,69 @@ scattered mapping csr_name_map
...
+// CSR require ext is enabled.
+// We don't provide a csr_require_hart_supported now, since 'hartSupported'
+// often requires more complex logic (like or), for these condition, the
+// recommended way is to use 'csr_require_cond' with a custom description string.
+function csr_require_ext(ext : extension) -> CSRAccessResult =
+  if   currentlyEnabled(ext)
+  then CSRAccess_Success()
+  else CSRAccess_Illegal(CSR_Illegal_ExtDisabled())
+
+function csr_require_xlen(len : int) -> CSRAccessResult =
```
Timmmm: Definitely shouldn't be using `int` here. As far as I can see it's only ever called with 32, so I'd either do:

```
function csr_require_xlen(len : int(32)) -> CSRAccessResult =
```

Or probably better:

```
function csr_requires_xlen_32() -> CSRAccessResult =
```

model/extensions/FD/fdext_control.sail
Timmmm: You could do it by adding the extension to `CSR_Illegal_ExtDisabled` (maybe a good idea anyway) and adding an `Or(a, b)` to the failure reason. Didn't I see that in earlier code or did I imagine that?

I dunno if it is worth the complexity though.

model/core/csr_begin.sail
```diff
@@ -15,10 +15,69 @@ scattered mapping csr_name_map
...
+function csr_defined_if(exist : bool) -> CSRAccessResult = if exist then CSRAccess_Success() else CSRAccess_Illegal(CSR_Illegal_NotDefined())
+
+// CSR require ext is enabled.
+// We don't provide a csr_require_hart_supported now, since 'hartSupported'
+// often requires more complex logic (like or), for these condition, the
+// recommended way is to use 'csr_require_cond' with a custom description string.
+function csr_require_ext(ext : extension) -> CSRAccessResult =
+  if   currentlyEnabled(ext)
+  then CSRAccess_Success()
+  else CSRAccess_Illegal(CSR_Illegal_ExtDisabled())
```
Timmmm: Might be nice to put the extension in here.

```suggestion
  else CSRAccess_Illegal(CSR_Illegal_ExtDisabled(ext))
```

Timmmm (changes requested):
Looks pretty good to me. My main concern is the use of strings. I think we can (and should) avoid them.

model/core/csr_begin.sail
trdthg: > Didn't I see that in earlier code or did I imagine that?

Yes, but I deleted it at some point due to complexity

> adding an Or(a, b) to the failure reason.

The biggest issue is that Sail doesn't support recursive type. You can only use this workaround, which even only supports one layer of recursion.

```sail
union CSRAccessResultWrapper = {
  CSRAccess_Result   : CSRAccessResult,
  CSRAccess_ResultOr : (CSRAccessResult, CSRAccessResult),
}
```

If we encounter a case like `A(false) | B(false) | C(false)`, the workaround will fail again, because sail doesn't allow us write
`CSRAccess_ResultOr(Illegal_A, CSRAccess_ResultOr(Illegal_B, Illegal_C))`

model/core/csr_begin.sail
trdthg: Only `|` condition and things like `mseccfg[SSEED] == 0b1` still use CSR_Illegal_Condition now.
We could have `CSR_Illegal_CSRBitDisabled(0x747, SSEED)`, but does sail provide a syntax to get the bit index of `SSEED`.

model/extensions/Stateen/stateen_regs.sail
```diff
@@ -9,9 +9,9 @@
 function clause currentlyEnabled(Ext_Smstateen) = hartSupports(Ext_Smstateen) & currentlyEnabled(Ext_Zicsr)
 function clause currentlyEnabled(Ext_Ssstateen) = hartSupports(Ext_Ssstateen) & currentlyEnabled(Ext_Zicsr)
 
-function is_mstateen_accessible() -> bool = hartSupports(Ext_Smstateen)
-function is_hstateen_accessible() -> bool = currentlyEnabled(Ext_H) & (currentlyEnabled(Ext_Smstateen) | currentlyEnabled(Ext_Ssstateen))
-function is_sstateen_accessible() -> bool = currentlyEnabled(Ext_S) & (currentlyEnabled(Ext_Smstateen) | currentlyEnabled(Ext_Ssstateen))
+function is_mstateen_accessible() -> CSRAccessResult = csr_requires_cond(hartSupports(Ext_Smstateen), "Smstateen extension is not supported")
```
nadime15: why not just 

`./extensions/Stateen/stateen_csrs.sail:18:function clause is_CSR_accessible(0x30C, _, _) = csr_required_cond(is_mstateen_accessible()), ... )`

Thats basically the same that you did with the floating points regs and you wouldn have to modify get_mstateen() etc.

model/core/csr_begin.sail
trdthg: > // TODO: refactor the Core module to prevent such cyclic dependencies;

lets do `CSR_Illegal_Stateen` later.

model/core/csr_begin.sail
Timmmm: Ah I didn't know that. Well in any case I think it's simpler just to hard-code the exceptions that don't fit into the existing options - `CSRAccess_FloatNotSupported` and so on.

PS: I think you replied to the wrong thread.

model/core/csr_begin.sail
Timmmm: > but does sail provide a syntax to get the bit index of SSEED

Hmm not as far as I know. Good question! But I don't think you need to be as generic as `CSRBitDisabled`. Something like `CSR_Illegal_SseedDisabled` would be fine.

nadime15: You wanna really go down this path and add X numbers of reason and potentially even more in the future? Dont you think its just easier to keep it generic and say unmet condition or something like that.

Timmmm: Looks like we only have about 5 special snowflake failure reasons at the moment which seems totally manageable to me?

model/core/csr_begin.sail
```diff
@@ -15,10 +15,76 @@ scattered mapping csr_name_map
 function csr_name(csr : csreg) -> string = csr_name_map(csr)
 overload to_str = {csr_name}
 
+union CSRAccessIllegalReason = {
+  // Attempt to perform an unsupported operation (e.g., write to a read-only CSR)
+  CSR_Illegal_AccessType : unit,
+
+  // Current privilege level is lower than the CSR's minimum required privilege
+  CSR_Illegal_Privilege : unit,
```
Timmmm: Btw we could add the required privilege level here too if it's not too difficult.

model/core/csr_begin.sail
nadime15: Arguably, we also have the satp register, unless you prefer to keep it generic and simply return `CSR_Illegal_Condition` (If we keep it generic here, why not do the same for the other cases? Where should we draw the line?). 

What happens once we introduce hypervisor CSRs that can trigger virtual instruction exceptions under different conditions? The vsatp register is similar to satp, and access to it might temporarily be denied even if the current mode is correct, do we return CSRAccess_Virtual(CSR_Illegal_VSATP_Condition())? 

Do we differentiate here too and have a new CSRAccessResult entry like: 

```
union CSRAccessResult = {
  CSRAccess_Success : unit,
  CSRAccess_Illegal : CSRAccessIllegalReason,
  CSRAccess_Virtual : CSRAccessIllegalReason,
}
```
And look at this table here from the [Hypervisor Spec](https://github.com/riscv/sail-riscv/issues/1350#issue-3524095722), do we add a reason for all of these cases? 

I am not fully convinced this is better, it might prefer to just return unmet cond. and optionally a string with an error message

Timmmm: Maybe we should punt on this for now then and just have like `CSR_Illegal_Uncategorized` with no string or something like that?

Presumably we'll have a follow-up PR to actually add logging/callbacks for this and when we have that it might be clearer what the best option is?

nadime15: Yeah sounds good.

model/core/csr_begin.sail
trdthg: > PS: I think you replied to the wrong thread.

Yes, but I didn't see the comment window for that one. Strange behavior.
image

trdthg: Rocq seems not happy with `CSR_Illegal_ExtDisabled : extension,`. Any experts?

```sail
File "./rv64d_types.v", line 261, characters 0-543:
Error:
The following term contains unresolved implicit arguments:
  (λ x : CSRAccessIllegalReason,
     match x with
     | CSR_Illegal_AccessType x' => encode (0, encode x')
     | CSR_Illegal_Privilege x' => encode (1, encode x')
     | CSR_Illegal_ExtDisabled x' => encode (2, encode x')
     | CSR_Illegal_XlenMismatch x' => encode (3, encode x')
     | CSR_Illegal_Counter x' => encode (4, encode x')
     | CSR_Illegal_SeedReadWithoutWrite x' => encode (5, encode x')
     | CSR_Illegal_Condition x' => encode (6, encode x')
     | CSR_Illegal_NotDefined x' => encode (7, encode x')
     end)
More precisely: 
- ?Countable: Cannot infer the implicit parameter Countable of encode whose
  type is "Countable extension" (no type class instance found) in
  environment:
  x : CSRAccessIllegalReason
  x' : extension
```

bacam: It's a combination of
* the Rocq backend attempting to provide an integer encoding of every type,
* the encoding for enumeration types is provided by the Sail front-end, but
* this doesn't include scattered enumerations like `encoding`.

I think I'll add a fallback to the Rocq backend.  We could temporarily work-around this by definition the encoding explicitly in Sail, but it would add quite a bit of noise to the model.

bacam: The development version of Sail can now cope with this.

model/core/csr_begin.sail
```diff
@@ -15,10 +15,76 @@ scattered mapping csr_name_map
...
+  // Like CSRAccess_Illegal, but raise Virtual_Instruction Exception
+  // CSRAccess_Virtual : ...,
+}
+
 // returns whether a CSR exists
-val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> bool
+val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> CSRAccessResult
 scattered function is_CSR_accessible
 
+function csr_defined_if(exist : bool) -> CSRAccessResult = if exist then CSRAccess_Success() else CSRAccess_Illegal(CSR_Illegal_NotDefined())
```
Timmmm:
```suggestion
function csr_defined_if(exist : bool) -> CSRAccessResult =
  if   exist
  then CSRAccess_Success()
  else CSRAccess_Illegal(CSR_Illegal_NotDefined())
```

Just for consistency with the other ones.

Timmmm (commented):
LGTM once the `CSR_Illegal_Condition` string is removed.

trdthg:
`CSR_Illegal_Condition` is removed now.

model/core/csr_begin.sail
```diff
@@ -15,10 +15,81 @@ scattered mapping csr_name_map
...
+
+union CSRAccessResult = {
+  CSRAccess_Success : unit,
+
+  // Accessing the CSR is illegal currently and should raise Illegal_Instruction Exception
+  CSRAccess_Illegal : CSRAccessIllegalReason,
+
+  // TODO: wait for hypervisor extension
+  // Like CSRAccess_Illegal, but raise Virtual_Instruction Exception
+  // CSRAccess_Virtual : ...,
```
nadime15: Can you uncomment this? I think its okay to merge this as well, so we can make use of it and don't have to wait for another PR.

pmundkur (commented):
Having gone through this again, I'm wondering if there could be a simpler version.  Afaiu, a CSR access failure can only result in an illegal-instruction exception, or a virtual-instruction exception.  It seems overkill to go into a finer granularity.  Secondly, the spec uses a notion of `HS-qualified` to capture when virtual-instruction exceptions can be thrown.  It would be good to capture and use that notion in the model.  It _might_ help avoid the huge `CSRAccessIllegalReason` union.

Timmmm:
I thought the intent was to help debugging? I.e. it's not in this PR but a future PR could actually print out the reason in the log. That can be super helpful when debugging test failures. Another thing I added locally at Codasip (but didn't upstream because it was a bit hacky) was logging of the PTW process when it fails. Saved me a mountain of time compared to the "something in this 10-step process went wrong" you get for page faults normally.

That sort of thing does complicate the model though so I would understand if you wanted to not have it.

Timmmm (approved):

arichardson (approved):
I have not reviewed the latest version to check that all conditions are exactly the same but I think this approach is great for readability and tracking the reason that an access is illegal.

Issue #925, opened 2025-05-07
Rework AccessType


Timmmm:
In addition to #787, I think there are some more things we should change about `AccessType`.

1. There are some `aq`, `rel`, `res`, `con` flags that are really part of the access type, but they're in separate bool flags for some reason.
2. `ReadWrite` is only used for AMOs so it should be called `AMO`.
3. `AMO` should include the AMO type. This is needed for PMA checks.
4. There should be `LoadReserved` and `StoreConditional` types.

I think in the end we'll have something like

```
union AccessType ('a : Type) = {
  Read             : 'a,
  Write            : 'a,
  // operation, Read, Write
  AMO              : (amoop, 'a, 'a),
  // acquire, release
  LoadReserved     : (bool, bool, 'a),
  StoreConditional : (bool, bool, 'a),
  InstructionFetch : unit,
  Cache            : CacheAccessType,
  // For cm.jt
  JumpTableFetch   : unit,
}
```

The purpose of the `'a` by the way is to allow CHERI to mark accesses as storing tagged capabilities or loading possibly-tagged capabilities. We might be able to remove it once CHERI is fully integrated (haven't really thought about that too much yet though).

jordancarlin:
We might want something more generic for `JumpTableFetch` that just indicates it is a non instruction fetch with execute access so that it can be used for other similar accesses down the road, but otherwise this seems like a good list.

trdthg:
what about amocas? it also may have `aq|rl`

trdthg:
```
// operation, Read, Write
  AMO              : (amoop, 'a, 'a),
```

Why AMO doesn't have acquire and release?

Timmmm:
Ah you're right; it should do.

pmundkur:
For Zicfiss, #377 added additional `SSRead/SSWrite/SSReadWrite` access types:
https://github.com/ved-rivos/sail-riscv-cfi/blob/161506364fe6d548ef7c9af8caa66e19a7eab890/model/riscv_types.sail#L136

Instead of replicating the `Load/Store/LoadStore` again for shadow stack access, we could use the `'a` parameter,  and have `'a` instantiated with `Data` or `ShadowStack`.

I'm not sure how that interacts with CHERI.  That spec (if I'm reading the right place) currently says Zicfiss is incompatible (so far):
[...]

PR #1405, opened 2025-11-24
Refactor the memory access type.

pmundkur:
The type now encodes the type of AMO operation for AMO accesses, separates out LR/SC, and explicitly encodes a CacheAccess variant.

This enables more accurate PMA and PMP checks for atomic and cache accesses, and simpler handling of cache accesses when checking permissions during address translation.

### DIFF
--- a/model/sys/vmem_pte.sail
+++ b/model/sys/vmem_pte.sail
@@ -128,7 +128,6 @@ private function check_PTE_permission(
 
   // Since we assume a valid PTE: Writable pages must be readable. Write-only
   // pages are reserved.
-
   assert(pte_W ==> pte_R);
 
   // Step 6 of VATP (see vmem.sail).


pmundkur:
Partially addresses #925.  I didn't change anything RVWMO related until we get litmus tests in CI.

Timmmm:
Ah damn sorry I had also mostly done this but not pushed the code... :-S I'll compare it with mine.

model/sys/mem.sail
```diff
@@ -176,7 +185,7 @@ function phys_access_check forall 'n, 0 < 'n <= max_mem_access . (
 
 // dispatches to MMIO regions or physical memory regions depending on physical memory map
 function checked_mem_read forall 'n, 0 < 'n <= max_mem_access . (
-  t : MemoryAccessType(ext_access_type),
+  t : MemoryAccessType(mem_payload),
```
nadime15: We use tons of different names for the same MemoryAccessType variable t, typ, accTy, ac, a etc. across the code. Probably makes sense to grep for all of them and agree on one common name unless the spec uses a specific name in that context?

pmundkur: This adds 95 lines to the diff.  I could push it as part of this PR if you think it won't complicate the review, otherwise perhaps it could be in a different PR?

nadime15: As you like, either way works for me

pmundkur: I pushed the change; happy to revert it and put it into a separate PR if needed.

model/extensions/Zicbop/zicbop_insts.sail
```diff
@@ -30,7 +30,37 @@ mapping prefetch_mnemonic : cbop_zicbop <-> string = {
...
   <-> prefetch_mnemonic(cbop) ^ spc() ^ hex_bits_12(offset) ^ "(" ^ opt_spc() ^ reg_name(rs1) ^ opt_spc() ^ ")"
 
-function clause execute ZICBOP(cbop, rs1, offset) = RETIRE_SUCCESS
+function clause execute ZICBOP(cbop, rs1, offset) = {
+  let rs1_val = X(rs1);
+  let cache_block_size = 2 ^ plat_cache_block_size_exp;
+
+  let addr = rs1_val + sign_extend(offset);
+  let cache_block_addr = addr & ~(zero_extend(ones(plat_cache_block_size_exp)));
+  let offset = cache_block_addr - rs1_val;
```
nadime15, line 41: ~~In the context of the pointer masking extension I looked at this instruction and I think none of this extra handling is needed. Even though similar instructions do it, it should be enough to pass rs1 and offset to get_transformed_data_addr (now ext_data_get_addr) if I am not mistaken.~~

Like here:

```
let cache_block_size = 2 ^ plat_cache_block_size_exp;
  match get_transformed_data_addr(rs1, sign_extend(offset), Load(Data), cache_block_size) {
```
Edit: Seems to be right!

Timmmm:
Sorry I haven't had a chance to review this yet, but here is the code I wrote:

https://github.com/Timmmm/sail-riscv/tree/user/timh/access_types

I didn't quite finish it though. When I get a chance I'll compare our approaches and hopefully we can pick the best bits from each (or feel free to also do that if you have time!).

pmundkur:
@Timmmm I've left the acquire/release flags alone for now.  I'd prefer to fold those in in a separate PR, to reduce the size of this one.  I'd also prefer to refactor those once we have litmus tests running, which I'm going to look at next.

pmundkur:
I'll update #1408 for Zicfiss once this has stabilized after review.

model/core/types.sail
```diff
@@ -93,26 +93,43 @@ overload to_str = {privLevel_to_str}
 
 // memory access types
 
+union cacheop = {
+  CB_manage   : unit,
+  CB_zero     : unit,
+  CB_prefetch : cbop_zicbop,
```
Timmmm, line 99: Can you separater `CB_manage` out into `CleanFlush` and `Inval`? CHERI needs to distinguish between those - see https://github.com/CHERI-Alliance/sail-riscv/blob/4f1dae4d0d38348a83dd8f3d670264330c767a22/model/cheri/cheri_addr_checks.sail#L141

model/sys/mem.sail
```diff
@@ -92,16 +92,29 @@ private function pmaCheck forall 'n, 0 < 'n <= max_mem_access .
...
-            //                                   "Invalid CacheAccessType; Inval, Clean and Flush check Read/Write permissions."),
+            Atomic(_, _, _)     => attributes.readable & attributes.writable,
+
+            // TODO: The PMAs shall be the same for all physical
+            // addresses in the cache block, and if write permission
+            // is granted by the supported access type PMAs, read
+            // permission shall also be granted.
+            CacheAccess(CB_zero())      => attributes.writable & attributes.supports_cbo_zero,
+            CacheAccess(CB_manage())    => attributes.readable | attributes.writable,
+            CacheAccess(CB_prefetch(p)) => match p {
```
Timmmm, line 137: Do we actually need a prefetch access type? Wouldn't it just be a load/store? From the perspective of the memory system I think they're indistinguishable.

model/sys/vmem_pte.sail
```diff
@@ -140,10 +140,29 @@ private function check_PTE_permission(
...
   let access_ok : bool = match access {
-    Load(_)             => pte_R | (pte_X & mxr),
-    Store(_)            => pte_W,
-    LoadStore(_, _)     => pte_W & (pte_R | (pte_X & mxr)),
-    InstructionFetch(_) => pte_X,
+    Load(_)                     => pte_R | (pte_X & mxr),
+    LoadReserved(_)             => pte_R | (pte_X & mxr),
+    Store(_)                    => pte_W,
+    StoreConditional(_)         => pte_W,
+    Atomic(_, _, _)             => pte_W & (pte_R | (pte_X & mxr)),
```
Timmmm: AMO doesn't require read access (and it's irrelevant anyway because you can't have write-only pages). I had this code:

```
    // AMO does not require read access, but write-only permissions are reserved.
    AMOAccess(_)        => pte_W,
```

We could maybe simplify and clarify this step a bit like:

```
// Writable pages must be readable. Write-only pages are reserved.
assert(pte_W ==> pte_R);

// Make eXecutable Readable.
let pte_R = pte_R | (pte_X & mxr);

let access_ok : bool = match access {
    Load(_)                     => pte_R,
    LoadReserved(_)             => pte_R,
    Store(_)                    => pte_W,
    StoreConditional(_)         => pte_W,
    Atomic(_, _, _)             => pte_W & pte_R,
...
```

model/sys/vmem_pte.sail
```diff
@@ -153,20 +172,36 @@ private function check_PTE_permission(
...
+                             // checks the accessed and dirty bits and
+                             // may either raise an exception or set the
+                             // bits as required."
+                             CacheAccess(CB_zero())      => true,
+                             CacheAccess(CB_prefetch(_)) => false,
                            });
   // Update 'accessed'-bit?
-  let update_a = (pte_flags[A] == 0b0);
+  // Cache-block management and zero instructions update the accessed bit,
+  // but cache-block prefetch instructions do not.
```
Timmmm, line 228: Ah is this why you added Prefetch? I wonder if it would make more sense as a flag on the other access types. Then it can be easily ignored everywhere except here.

Timmmm (commented):
Hey sorry I *finally* got around to reviewing this! Looks mostly great. I think the biggest thing I would change is the way the enums are arranged. What do you think about this?

```
enum cacheop = {
  // CHERI needs to distinguish Clean/Flush from Inval because
  // Inval requires more permissions.
  CleanFlush,
  Inval,
  Zero,
}

// I'm sure we can think of better names for these!
enum IsPrefetch = {
  Prefetch,
  NotPrefetch,
}

  Load             : (IsPrefetch, 'a),
  Store            : (IsPrefetch, 'a),
  LoadReserved     : 'a,
  StoreConditional : 'a,
  Atomic           : (amoop, 'a, 'a),
  InstructionFetch : IsPrefetch,
  CacheAccess      : cacheop,
}
```

The advantages:

1. Works with CHERI.
2. `cacheop` is simpler - enum instead of union.
3. No need to explicitly deal with prefetch accesses in 99% of the code.
4. I think this avoids needing to change the dependencies in `sail.riscv_project`?
5. It's a bit simpler I think.

Oh finally, it's maybe worth splitting out the `ext_access_type` rename into another PR? Kind of an awkward thing to name; `ext_access_type` is technically correct because it doesn't have to be "mem payload" as far as this model knows. In fact in the latest CHERI code I had to make it have quite a confusing meaning: https://github.com/CHERI-Alliance/sail-riscv/blob/2e206c622773099f3716016e17ab445df55055fb/model/cheri/cheri_vmem_types.sail#L32-L41

model/sys/mem.sail
pmundkur: Having a prefetch access type helps capture its treatment by memory protection as described in the spec.

model/sys/vmem_pte.sail
pmundkur: Done.

model/sys/vmem_pte.sail
pmundkur: My concern with adding `prefetch` as a flag to the other access types is that
every Load/Store in every extension would need to specify it (e.g. all the Load/Store accesses in the vector extensions).  It would be odd to have to specify with `NotPrefetch` in every extension that did a load/store memory operation.  So I left it as a separate access type for now.

Timmmm: Hmm fair point. Can we make them more symmetric with Load/Store/InstructionFetch at least? I.e.

```
enum cacheop = {
  // CHERI needs to distinguish Clean/Flush from Inval because
  // Inval requires more permissions.
  CleanFlush,
  Inval,
  Zero,
}

  Load             : 'a,
  LoadPrefetch : 'a,
  Store            : 'a,
  StorePrefetch : 'a,
  LoadReserved     : 'a,
  StoreConditional : 'a,
  Atomic           : (amoop, 'a, 'a),
  InstructionFetch : unit,
  InstructionPrefetch : unit, // Or InstructionFetchPrefetch? :-/
  CacheAccess      : cacheop,
}
```

Timmmm: @jordancarlin what do you think?

pmundkur: My concern with this, fwiw, is that this form will encourage the use of top-level `_` wildcard pattern matches.  I find it preferable to use nested-wildcards to ignore prefetches with `CacheAccess(CB_prefetch(_))` and ignore cache accesses with `CacheAccess(_)`.  
I'd also prefer to avoid duplicating definitions, like `PREFETCH_X` in `Zicbop` and then again `X_Prefetch` in `core`.

jordancarlin: I think I agree with @pmundkur on this one. I don't see the benefit of making all three variants of prefetch top-level keys. We could make prefetch itself a top-level key (with the three union values) instead of part of `cacheop` if you think that is simpler @Timmmm.

Timmmm: Yeah I would prefer that I think. The current structure seems like it treats prefetch as more special than it actually is, if you see what I mean. And it definitely seems weird to group it with the CBO ops because they're quite different.

pmundkur: Well, all these extensions manage caches (and they even defined together in the same section of the Cache Management chapter), so they are related in that sense, so to me it made sense to have them all under `CacheAccess`.

Timmmm: Ok fair enough I'll stop nitpicking 😄

model/core/mem_type_utils.sail
```diff
@@ -8,16 +8,38 @@
...
+    Atomic(_)                   => E_SAMO_Access_Fault(),
+    // Cache-block management (CB_manage) and zero (CB_zero)
+    // instructions raise store faults.
+    CacheAccess(CB_manage(_))   => E_SAMO_Access_Fault(),
+    CacheAccess(CB_zero())      => E_SAMO_Access_Fault(),
+    // Though prefetches don't raise exceptions, we return a nominal
+    // exception here to propagate the failed address translation to
+    // the execute of the calling instruction so that it could decide
+    // whether to actually proceed with a prefetch.
+    CacheAccess(CB_prefetch(_)) => E_Load_Access_Fault(),
```
Timmmm: I think if we're going to return something here we should probably return the right kind, i.e. fetch/samo access fault?

model/core/mem_type_utils.sail
```diff
@@ -8,16 +8,38 @@
...
+    Atomic(_)                   => E_SAMO_Addr_Align(),
+    // Cache-block management (CB_manage) and zero (CB_zero)
+    // instructions raise store faults.
+    CacheAccess(CB_manage(_))   => E_SAMO_Addr_Align(),
+    CacheAccess(CB_zero())      => E_SAMO_Addr_Align(),
+    // Though prefetches don't raise exceptions, we return a nominal
+    // exception here to propagate the failed address translation to
+    // the execute of the calling instruction so that it could decide
+    // whether to actually proceed with a prefetch.
+    CacheAccess(CB_prefetch(_)) => E_Load_Addr_Align(),
```
Timmmm: Ditto

model/extensions/Zicbom/zicbom_insts.sail
```diff
@@ -78,60 +78,19 @@ private function process_clean_inval(rs1 : regidx, _cbop : cbop_zicbom) -> Execu
...
-          let exc_write = phys_access_check(Store(Data), ep, paddr, cache_block_size, false);
-          match (exc_read, exc_write) {
-            // Access is permitted if read OR write are allowed. If neither
-            // are allowed then we always report a store exception.
-            (Some(_exc_read), Some(exc_write)) => Some(exc_write),
-            _ => None(),
-          }
+    Ext_DataAddr_OK(vaddr) => match translateAddr(vaddr, CacheAccess(CB_manage(cbop))) {
+      Ok(_paddr, _) => {
+          // The model has no caches so there's no action required.
```
Timmmm: I think this change is incorrect - even if you have no caches you still have to raise an exception when accessing memory you don't have permission to access.

model/sys/mem.sail
```diff
@@ -92,16 +92,52 @@ private function pmaCheck forall 'n, 0 < 'n <= max_mem_access .
         _ => {
           // Check read/write/execute permissions and PMAs.
           let canAccess : bool = match access {
-            InstructionFetch() => attributes.executable,
-            Load(_)            => attributes.readable & not(res_or_con & attributes.reservability == RsrvNone),
-            Store(_)           => attributes.writable & not(res_or_con & attributes.reservability == RsrvNone),
+            InstructionFetch()  => attributes.executable,
+            // XXX: remove reservability check from Load/Store
```
Timmmm:
```suggestion
            // TODO: remove reservability check from Load/Store
```

Or did you intend to do that in this PR? I think just removing it is correct?

Maybe add `assert(res_or_con)`/`assert(not(res_or_con))` to the `Load`/`LoadReserved` etc cases temporarily?

model/pmp/pmp_control.sail
```diff
@@ -91,6 +101,14 @@ function pmpCheck forall 'n, 0 < 'n <= max_mem_access . (
 
   let width : xlenbits = to_bits(width);
 
+  // TODO for accesses of type CacheAccess:
+  //
+  // The CMO extensions impose the following constraints on the
+  // physical addresses in a given cache block: "The PMP access control
+  // bits shall be the same for all physical addresses in the cache
+  // block, and if write permission is granted by the PMP access
+  // control bits, read permission shall also be granted."
```
Timmmm, line 116: You've already got this TODO lower down.

model/pmp/pmp_control.sail
```diff
@@ -11,10 +11,20 @@
...
-    LoadStore(_)       => ent[R] == 0b1 & ent[W] == 0b1,
-    InstructionFetch() => ent[X] == 0b1,
+    Load(_)             => ent[R] == 0b1,
+    LoadReserved(_)     => ent[R] == 0b1,
+    Store(_)            => ent[W] == 0b1,
+    StoreConditional(_) => ent[W] == 0b1,
+    Atomic(_)           => ent[R] == 0b1 & ent[W] == 0b1,
+    InstructionFetch()  => ent[X] == 0b1,
+
+    CacheAccess(CB_manage(_))   => bool_bits(ent[R]) | bool_bits(ent[W]),
```
Timmmm: Maybe worth a `TODO: Handle the unspecified case where X is granted but R/W aren't` so we don't forget (same as in `pmaCheck` and `check_PTE_permission`).

model/sys/vmem_ptw.sail
```diff
@@ -53,14 +53,28 @@ function translationException(access : MemoryAccessType(mem_payload),
...
+
+    (CacheAccess(CB_manage(_)), PTW_No_Access()) => E_SAMO_Access_Fault(),
+    (CacheAccess(CB_manage(_)), _)               => E_SAMO_Page_Fault(),
+    (CacheAccess(CB_zero()), PTW_No_Access())    => E_SAMO_Access_Fault(),
+    (CacheAccess(CB_zero()), _)                  => E_SAMO_Page_Fault(),
+    // Though prefetches don't raise exceptions, we return a nominal
+    // exception here to propagate the failed address translation to
+    // the execute of the calling instruction so that it can decide
+    // whether to actually proceed with a prefetch.
+    (CacheAccess(CB_prefetch(_)), _)             => E_Load_Access_Fault(),
```
Timmmm: Ditto

model/sys/mem.sail
pmundkur: Sorry, I missed that, I was planning to do it in this PR.

model/extensions/Zicbom/zicbom_insts.sail
pmundkur: It might be hard to see in this diff, but an exception is being returned in both the `Ext_DataAddr_Error`  and any error from `translateAddr()`.

model/pmp/pmp_control.sail
pmundkur: This TODO occurs only once in this file, but a very similar TODO is in `mem.sail` where it's the same except for PMAs instead of PMPs.

model/extensions/Zicbom/zicbom_insts.sail
Timmmm: `translateAddr()` only checks the PTE permissions though. You still need to check PMAs and PMPs which is what the old code did via `phys_access_check()`.

model/pmp/pmp_control.sail
Timmmm: Oh right you are. Didn't see it was a different file!

model/extensions/Zicbom/zicbom_insts.sail
pmundkur: Good catch, I'll fix it.

pmundkur: Fixed.

model/sys/mem.sail
pmundkur: The asserts need a slight refactor of the `mem_write_*` APIs since they use a default access of `Store` which trips asserting on `not(res_or_con)`.  I've left a TODO for now.

model/extensions/Zicbom/zicbom_insts.sail
Timmmm: Ah nice, a lot simpler now too!

model/sys/mem.sail
Timmmm: I guess we won't need them anyway when we get rid of the flags (I'll try to rebase my branch that does that onto this one).

jordancarlin (approved):

PR #1406, opened 2025-11-24
Factor out a `prelude` module.

pmundkur:
This allows extensions providing types based on simple enums and `bits` to have them included `before` the `core` module in the project file.

Also extract the types needed for the `V` extension from `core` into a submodule of `V`.  

While here, this also updates the ReadingGuide.


model/core/phys_mem_interface.sail
pmundkur, line 1: Perhaps this should be called `phys_mem_interface`?

Timmmm: Yeah I think that's slightly better but either's pretty good.

jordancarlin (approved):

PR #1409, opened 2025-11-26
Fix unused variable warnings.

pmundkur:
These will trigger in the next Sail compiler release. Also:

* Fixes a PMA config typo (`misaligned_fault` -> `MisalignedFault`).
* Remove unused parameter from `pmpWriteCfg`.

Some other functions should probably be changed to remove unused parameters from their signature; perhaps in a later PR.

### DIFF
--- a/model/sys/mem.sail
+++ b/model/sys/mem.sail
@@ -79,7 +79,7 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : MemoryAccessTy
   }) : option((bits(8 * 'n), mem_meta));
   match (t, result) {
     (InstructionFetch(), None()) => Err(E_Fetch_Access_Fault()),
-    (Load(Data), None())         => Err(E_Load_Access_Fault()),
+    (Load(_), None())            => Err(E_Load_Access_Fault()),
     (_, None())                  => Err(E_SAMO_Access_Fault()),
     (_, Some(v, m))              => Ok(v, m),
   }
@@ -88,14 +88,14 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : MemoryAccessTy
 function accessFaultFromAccessType(accTy : MemoryAccessType(ext_access_type)) -> ExceptionType =
   match accTy {
     InstructionFetch() => E_Fetch_Access_Fault(),
-    Load(Data)         => E_Load_Access_Fault(),
+    Load(_)            => E_Load_Access_Fault(),
     _                  => E_SAMO_Access_Fault(),
   }
 
 function alignmentFaultFromAccessType(accTy : MemoryAccessType(ext_access_type)) -> ExceptionType =
   match accTy {
     InstructionFetch() => E_Fetch_Addr_Align(),
-    Load(Data)         => E_Load_Addr_Align(),
+    Load(_)            => E_Load_Addr_Align(),
     _                  => E_SAMO_Addr_Align(),
   }
 
--- a/model/sys/platform.sail
+++ b/model/sys/platform.sail
@@ -123,7 +123,7 @@ function clint_load(t, Physaddr(addr), width) = {
     then print_log("clint[" ^ bits_str(addr) ^ "] -> <not-mapped>");
     match t {
       InstructionFetch() => Err(E_Fetch_Access_Fault()),
-      Load(Data)         => Err(E_Load_Access_Fault()),
+      Load(_)            => Err(E_Load_Access_Fault()),
       _                  => Err(E_SAMO_Access_Fault())
     }
   }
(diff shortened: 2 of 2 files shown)


model/extensions/V/vext_fp_red_insts.sail
```diff
@@ -95,7 +95,7 @@ mapping clause encdec = RFWVVTYPE(funct6, vm, vs2, vs1, vd)
   when (currentlyEnabled(Ext_Zvfh)   & get_sew() == 16)
      | (currentlyEnabled(Ext_Zve64d) & get_sew() == 32)
 
-function clause execute RFWVVTYPE(funct6, vm, vs2, vs1, vd) = {
+function clause execute RFWVVTYPE(_funct6, vm, vs2, vs1, vd) = {
```
Timmmm, line 102: Probably worth a comment explaining why this is unused. I think it distinguishes ordered and unordered and we always do ordered anyway?

model/extensions/V/vext_mem_insts.sail
```diff
@@ -369,7 +369,7 @@ mapping clause encdec = VLUXSEGTYPE(nf, vm, vs2, rs1, width, vd)
   when (currentlyEnabled(Ext_Zve32x) & vlewidth_pow(width) <= 5)
      | (currentlyEnabled(Ext_Zve64x) & vlewidth_pow(width) <= 6 & xlen == 64)
 
-function process_vlxseg(nf : nfields, vm : bits(1), vd : vregidx, EEW_index_bytes : word_width, EEW_data_bytes : word_width, EMUL_index_pow : int, EMUL_data_pow : int, rs1 : regidx, vs2 : vregidx, num_elem : range(1, vlen), mop : int) -> ExecutionResult = {
+function process_vlxseg(nf : nfields, vm : bits(1), vd : vregidx, EEW_index_bytes : word_width, EEW_data_bytes : word_width, EMUL_index_pow : int, EMUL_data_pow : int, rs1 : regidx, vs2 : vregidx, num_elem : range(1, vlen), _mop : int) -> ExecutionResult = {
```
Timmmm, line 372: Any idea what this is?

model/pmp/pmp_regs.sail
```diff
@@ -96,7 +96,7 @@ function pmpLocked(cfg: Pmpcfg_ent) -> bool =
 function pmpTORLocked(cfg: Pmpcfg_ent) -> bool =
    (cfg[L] == 0b1) & (pmpAddrMatchType_encdec(cfg[A]) == TOR)
 
-function pmpWriteCfg(n: range(0, 63), cfg: Pmpcfg_ent, v: bits(8)) -> Pmpcfg_ent =
+function pmpWriteCfg(_n: range(0, 63), cfg: Pmpcfg_ent, v: bits(8)) -> Pmpcfg_ent =
```
Timmmm: Probably shold remove `_n` here... 

I think it's here because in my original code `sys_pmp_usable_count` was handled in this function, and I forgot to remove the parameter when I removed `sys_pmp_usable_count` before upstreaming.

model/sys/mem.sail
```diff
@@ -115,7 +115,7 @@ function pmaCheck forall 'n, 0 < 'n <= max_mem_access .
       // Check if we need to raise an exception for misalignment.
       match attributes.misaligned_fault {
         AccessFault if misaligned => return Some(accessFaultFromAccessType(accTy)),
-        misaligned_fault if misaligned => return Some(alignmentFaultFromAccessType(accTy)),
+        _misaligned_fault if misaligned => return Some(alignmentFaultFromAccessType(accTy)),
```
Timmmm: Oops I think that should actually be `MisalignedFault`...

model/extensions/V/vext_mem_insts.sail
pmundkur: It's the memory addressing mode.  It really should be an enum instead of an int.  In this case, it's again selecting between the ordered and unordered version of indexed loads and stores, and ignored because it's only implementing the ordered version.  I'll make this an enum and add a comment.

pmundkur: I'll pull this part out into a separate PR, since it's a larger diff.

pmundkur: See #1410.

model/sys/mem.sail
```diff
@@ -88,14 +88,14 @@ function phys_mem_read forall 'n, 0 < 'n <= max_mem_access . (t : MemoryAccessTy
...
-    Load(Data)         => E_Load_Access_Fault(),
+    Load(_)            => E_Load_Access_Fault(),
     _                  => E_SAMO_Access_Fault(),
   }
 
 function alignmentFaultFromAccessType(accTy : MemoryAccessType(ext_access_type)) -> ExceptionType =
   match accTy {
     InstructionFetch() => E_Fetch_Addr_Align(),
-    Load(Data)         => E_Load_Addr_Align(),
+    Load(_)            => E_Load_Addr_Align(),
```
Timmmm, line 98: Btw I think this technically changes the behaviour - in the CHERI version you could have `Load(Tagged)` or something like that. However... this looks more correct to me anyway.

PR #1410, opened 2025-11-26
Consolidate the indexed vector load and store instructions.

pmundkur:
Introduce an `indexed_mop` enum to encode the ordered and unordered indexed variants for these instructions, and use it to avoid duplicating almost identical definitions for the two cases. The difference is isolated into the implementing `process_*` functions, which currently only implement the ordered case.


Timmmm (approved):
Nice!

nadime15 (approved):

jordancarlin (approved):

PR #1415, opened 2025-12-04
Plumb trap_reason from Sail traps into C callbacks

ibvqeibob:
Adds a `trap_reason` field to Sail traps and exposes it to the C/C++ callback interface so that the emulator can log more detailed reasons for traps.

Main changes:
- Introduce a debug-only `trap_reason` union and add `reason : trap_reason` to `sync_exception`.
- Add `handle_exception_with_reason(xtval, e, reason)` and keep the existing `handle_exception` as a wrapper using `Trap_reason_none()`.
- Extend `trap_handler`, `exception_handler` and `trap_callback` to take and propagate a `trap_reason`.
- Expose `struct ztrap_reason` from `sail_riscv_model.h` as `typedef struct ztrap_reason trap_reason` and update the C/C++ `trap_callback(bool, fbits, trap_reason)` interface.
- For `Step_Fetch_Failure` and `Step_Execute(Memory_Exception(...))`, pass `Access_not_in_physical_memory()` as an initial example reason.
- `log_callbacks::trap_callback` prints both the trap cause and `reason` for debugging.

Testing:
- Rebuilt `sail_riscv_sim` successfully.
- Ran a small ELF that triggers an instruction fetch access fault and observed:
  `trap: is_interrupt=0 cause=1` and `reason: access_not_in_physical_memory` in the log.

### DIFF
--- a/model/core/types_ext.sail
+++ b/model/core/types_ext.sail
@@ -43,6 +43,36 @@ let init_ext_ptw : ext_ptw = ()
 
 type ext_ptw_fail = unit
 
+union TrapReason = {
+  Trap_Reason_Unclassified                 : unit,
+
+  Access_Not_In_Physical_Memory            : unit,
+
+  Trap_Reason_Invalid_Source_Addr          : unit,
+  Trap_Reason_No_PTE                       : unit,
+  Trap_Reason_No_Permission                : unit,
+  Trap_Reason_Misaligned_Superpage         : unit,
+  Trap_Reason_PTE_Update_Needed_But_Disabled : unit,
+
+  Invalid_Pte_Reserved_Bits_Nonzero        : (physaddrbits, xlenbits),
+
+  Trap_Reason_Ext_PTW_Error                : unit
+}
+
+function trap_reason_to_str(r : TrapReason) -> string = {
+  match r {
+    Trap_Reason_Unclassified() => "unclassified",
+    Access_Not_In_Physical_Memory() => "access-not-in-physical-memory",
+    Trap_Reason_Invalid_Source_Addr() => "invalid-source-addr",
+    Trap_Reason_No_PTE() => "no-pte",
+    Trap_Reason_No_Permission() => "no-permission",
+    Trap_Reason_Misaligned_Superpage() => "misaligned-superpage",
+    Trap_Reason_PTE_Update_Needed_But_Disabled() => "pte-update-needed-but-disabled",
+    Invalid_Pte_Reserved_Bits_Nonzero(_, _) => "invalid-pte-reserved-bits-nonzero",
+    Trap_Reason_Ext_PTW_Error() => "extension-ptw-error"
+  }
+}
+
 // This type can be used for custom errors for page-table-walks,
 // and values in this type are typically generated from the final
 // result of the ext_ptw at the end of the walk.
--- a/model/postlude/step.sail
+++ b/model/postlude/step.sail
@@ -13,7 +13,7 @@ register hart_state : HartState = HART_ACTIVE()
 private union Step = {
   Step_Pending_Interrupt  : (InterruptType, Privilege),
   Step_Ext_Fetch_Failure  : ext_fetch_addr_error,
-  Step_Fetch_Failure      : (virtaddr, ExceptionType),
+  Step_Fetch_Failure      : (virtaddr, (ExceptionType, TrapReason)),
   Step_Execute            : (ExecutionResult, instbits),
   Step_Waiting            : WaitReason,
 }
@@ -93,7 +93,7 @@ private function run_hart_active(step_no: nat) -> Step = {
       // extension error
       F_Ext_Error(e)   => Step_Ext_Fetch_Failure(e),
       // standard error
-      F_Error(e, addr) => Step_Fetch_Failure(Virtaddr(addr), e),
+      F_Error(e, addr) => Step_Fetch_Failure(Virtaddr(addr), (e, Trap_Reason_Unclassified())),
       // non-error cases:
       F_RVC(h) => {
         sail_instr_announce(h);
@@ -205,14 +205,15 @@ function try_step(step_no : nat, exit_wait : bool) -> bool = {
       handle_interrupt(intr, priv)
     },
     Step_Ext_Fetch_Failure(e) => ext_handle_fetch_check_error(e),
-    Step_Fetch_Failure(vaddr, e) => handle_exception(bits_of(vaddr), e),
+    Step_Fetch_Failure(vaddr, (e, r)) => handle_exception_with_reason(bits_of(vaddr), e, r),
     Step_Waiting(_) => assert(hart_is_waiting(hart_state), "cannot be Waiting in a non-Wait state"),
     Step_Execute(Retire_Success(), _) => assert(hart_is_active(hart_state)),
     // This case was handled above when processing the fetch result.
     Step_Execute(ExecuteAs(_), _) => assert(false),
     // standard errors
     Step_Execute(Trap(priv, ctl, pc), _) => set_next_pc(exception_handler(priv, ctl, pc)),
-    Step_Execute(Memory_Exception(vaddr, e), _) => handle_exception(bits_of(vaddr), e),
+    Step_Execute(Memory_Exception(vaddr, (e, r)), _) =>
+      handle_exception_with_reason(bits_of(vaddr), e, r),
     Step_Execute(Illegal_Instruction(), instbits) => handle_exception(zero_extend(instbits), E_Illegal_Instr()),
     Step_Execute(Enter_Wait(wr), instbits) => {
       if wait_is_nop(wr) then {
--- a/model/sys/sys_control.sail
+++ b/model/sys/sys_control.sail
@@ -115,6 +115,8 @@ function dispatchInterrupt(priv : Privilege) -> option((InterruptType, Privilege
   }
 }
 
+register pending_trap_reason : TrapReason = Trap_Reason_Unclassified()
+
 // types of privilege transitions
 
 union ctl_result = {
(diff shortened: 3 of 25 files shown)


model/sys/sys_control.sail
```diff
@@ -284,15 +284,22 @@ function xtval_exception_value(e : ExceptionType, excinfo : xlenbits) -> option(
   } then Some(excinfo) else None()
 }
 
-function handle_exception(xtval : xlenbits, e : ExceptionType) -> unit = {
-  let t : sync_exception = struct { trap    = e,
-                                    excinfo = xtval_exception_value(e, xtval),
-                                    ext     = None() };
+function handle_exception_with_reason(xtval : xlenbits, e : ExceptionType, r : trap_reason) -> unit = {
```
nadime15: Looks good overall but I personally dont think that we should rename the function. `handle_exception()` was good.

ibvqeibob: Thanks for the review!
I've kept `handle_exception(xtval, e)` as the main interface and added `handle_exception_with_reason(xtval, e, r)` as a helper. The old `handle_exception` now just forwards to the helper with `Trap_reason_none()`, so existing call sites can continue to use the original name.

nadime15: Oh, I missed that. But why not just modify handle_exception() function signature and pass the reason instead? Unless I’m missing something, that seems simpler.  As of today, there are three call sites for handle_exception()

model/postlude/step.sail
```diff
@@ -205,14 +205,14 @@ function try_step(step_no : nat, exit_wait : bool) -> bool = {
...
+    Step_Fetch_Failure(vaddr, e) => handle_exception_with_reason(bits_of(vaddr), e, Access_not_in_physical_memory()),
     Step_Waiting(_) => assert(hart_is_waiting(hart_state), "cannot be Waiting in a non-Wait state"),
     Step_Execute(Retire_Success(), _) => assert(hart_is_active(hart_state)),
     // This case was handled above when processing the fetch result.
     Step_Execute(ExecuteAs(_), _) => assert(false),
     // standard errors
     Step_Execute(Trap(priv, ctl, pc), _) => set_next_pc(exception_handler(priv, ctl, pc)),
-    Step_Execute(Memory_Exception(vaddr, e), _) => handle_exception(bits_of(vaddr), e),
+    Step_Execute(Memory_Exception(vaddr, e), _) => handle_exception_with_reason(bits_of(vaddr), e, Access_not_in_physical_memory()),
     Step_Execute(Illegal_Instruction(), instbits) => handle_exception(zero_extend(instbits), E_Illegal_Instr()),
```
nadime15, line 217: Did you miss to pass a reason for Illegal_Instruction()?

model/sys/sys_control.sail
```diff
@@ -284,15 +284,22 @@ function xtval_exception_value(e : ExceptionType, excinfo : xlenbits) -> option(
...
+    ext     = None(),
+    reason  = r
+  };
   set_next_pc(exception_handler(cur_privilege, CTL_TRAP(t), PC))
 }
 
+function handle_exception(xtval : xlenbits, e : ExceptionType) -> unit = {
+  handle_exception_with_reason(xtval, e, Trap_reason_none())
+}
+
```
nadime15: What does Trap_reason_none() represent? A trap should always have a reason, or is this just a temporary placeholder and reasons are added gradually over time?

Unless I am missing something, I would remove it entirely (Trap_reason_none()) or at least wrap it in an option type like option(trap_reason).

model/sys/sys_control.sail
ibvqeibob: Thanks, that makes sense.
I've now removed `handle_exception_with_reason` and changed `handle_exception` to take `(xtval, e, reason : trap_reason)` directly. All existing call sites have been updated to pass an explicit `trap_reason`.
For the memory-related traps (`Step_Fetch_Failure` and `Memory_Exception`) I pass `Access_not_in_physical_memory()`, and for other traps I currently pass `Trap_reason_none()` until we add more specific reasons for them.

model/postlude/step.sail
ibvqeibob: Good catch — I missed that case initially.
`Step_Execute(Illegal_Instruction(), instbits)` now calls: handle_exception(zero_extend(instbits), E_Illegal_Instr(), Trap_reason_none())

model/sys/sys_control.sail
ibvqeibob: `Trap_reason_none()` is intended to mean "there is no additional debug information beyond the architectural ExceptionType".
Right now only a subset of traps populate a more specific `trap_reason` (for example the memory-related fetch/data access faults that motivated this change). For the rest, we still want to plumb a `trap_reason` value through the interfaces so that the C callbacks have a uniform signature, but we don't yet have anything more precise to attach, hence the `Trap_reason_none()` constructor.
I did consider modelling this as `option(trap_reason)` as you suggest, but using an explicit `Trap_reason_none` keeps the generated C interface a bit simpler.

nadime15: But isn't the whole point of this PR to get more clarity on why we trapped? If most call sites just pass Trap_reason_none(), we're not really getting that clarity, we just have extra work. 

It’s just my opinion, but I think we should not use Trap_reason_none(). It would be better to create meaningful messages, for example in the make_landing_pad_exception case. You could simply add a new reason like ‘landing pad exception’ or something like that.

ibvqeibob: You're absolutely right. I've reworked the patch along those lines:
- Introduced `Trap_reason_unclassified` as an explicit "no extra information (yet)" value, and restricted its use to cases
  where we genuinely don't have anything more to say (e.g. interrupts, some generic/legacy paths).
- For fetch and generic memory exceptions, access faults (`E_Fetch_Access_Fault`, `E_Load_Access_Fault`, `E_SAMO_Access_Fault`) now use `Access_not_in_physical_memory()`instead of an uninformative default.
- `make_landing_pad_exception` now uses a dedicated `Landing_pad_exception()` trap reason, so landing-pad-related  software checks are distinguishable on the C side.

model/core/types_ext.sail
```diff
@@ -48,6 +48,15 @@ type ext_ptw_fail = unit
...
 type ext_ptw_error = unit // No extensions for page-table-walk errors
 
+// Additional information about the underlying cause of a trap.
+// This is more fine-grained than ExceptionType / mcause.
+union trap_reason = {
+  Trap_reason_unclassified : unit,
+  Access_not_in_physical_memory : unit,
+  Invalid_pte_reserved_bits_nonzero : (physaddrbits, xlenbits),
+  Landing_pad_exception : unit
+}
```
nadime15:
```suggestion
union TrapReason = {
  Trap_Reason_Unclassified          : unit,
  Access_Not_In_Physical_Memory     : unit,
  Invalid_Pte_Reserved_Bits_Nonzero : (physaddrbits, xlenbits),
  Landing_Pad_Exception             : unit
}
```

Seems like the majority of the code basis uses for union snake_case, with capital letters

model/core/types_ext.sail
```diff
@@ -48,6 +48,15 @@ type ext_ptw_fail = unit
 // result of the ext_ptw at the end of the walk.
 type ext_ptw_error = unit // No extensions for page-table-walk errors
 
+// Additional information about the underlying cause of a trap.
+// This is more fine-grained than ExceptionType / mcause.
+union trap_reason = {
+  Trap_reason_unclassified : unit,
+  Access_not_in_physical_memory : unit,
+  Invalid_pte_reserved_bits_nonzero : (physaddrbits, xlenbits),
```
nadime15: It looks like `Invalid_Pte_Reserved_bits_nonzero` is not used?

model/postlude/step.sail
```diff
@@ -157,6 +157,21 @@ function wait_is_nop(wr : WaitReason) -> bool =
...
+    E_Fetch_Access_Fault() => Access_not_in_physical_memory(),
+    E_Load_Access_Fault()  => Access_not_in_physical_memory(),
+    E_SAMO_Access_Fault()  => Access_not_in_physical_memory(),
+
+    // All other exceptions (alignment, page faults, envcalls,
+    // software checks, extension-specific traps, etc.) are not yet
+    // classified more finely here.
+    _                      => Trap_reason_unclassified()
+  }
+
```
nadime15: Something that confuses me a bit, is that currently we do not provide precise reasons for why an exception was triggered. At the moment, this approach is quite similar to simply enabling and running with --trace-exception.

I think what needs to be done is, for example, in the case of a memory exception during pt_walk, we should return the underlying cause, as defined in PTW_Error. One idea would be to modify translateAddr() and implement something like:

```
Err(f, ext_ptw) => {
  let exc = translationException(ac, f);
  let reason = ptw_error_to_trap_reason(f);
  Err((exc, reason), ext_ptw)
}
```

```
function ptw_error_to_trap_reason(f : PTW_Error) -> trap_reason = {
  match f {
    PTW_No_Access()               => Access_not_in_physical_memory(),
    // the one below is a new union entry
    PTW_Reserved_Bits(addr, pte) => Invalid_pte_reserved_bits_nonzero(bits_of(addr), pte),
    ...
    _                            => Trap_reason_unclassified()
  }
}
```

ibvqeibob: I’ve added a TrapReason union and wired PTW_Error to it as suggested, so PTW-related memory exceptions now carry both the ExceptionType and a more precise TrapReason.

ibvqeibob:
a

nadime15:
Was there a specific reason for closing this PR @ibvqeibob?

ibvqeibob:
> Was there a specific reason for closing this PR @ibvqeibob?

I did not fully understand this part of the process before, so I closed the PR and reviewed it carefully. This revision passes the TrapReason (a more specific trap cause) all the way from the Sail model to the trap_callback in C.The implementation method is as follows: add TrapReason on the Sail side and map the corresponding causes in the exception generation paths (especially the memory exceptions related to page table walk/address translation); meanwhile, update the C callback interface signature to pass the reason parameter to the upper-layer callback.

ibvqeibob:
> This has a lot of unrelated changes from other PRs. Could you rebase properly to isolate the changes for the traps/callbacks?

Thanks for the heads-up! I’ve rebased the branch onto the latest upstream/master and squashed the work into a single commit. The PR now contains only the TrapReason plumbing from Sail to the C++ trap_callback, with no unrelated commits.

PR #1416, opened 2025-12-06
Bump CI version

Arielfoever:
Bump CMake to 3.31.9 and 4.2.0

Add ubuntu-24.04 for test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04, ubuntu-24.04-arm, macos-latest]
-        cmake_version: ["3.20.0", "4.1.2"]
+        os: [ubuntu-22.04, ubuntu-24.04, ubuntu-24.04-arm, macos-latest]
+        cmake_version: ["3.31.9", "4.2.0"]
         include:
-          - os: ubuntu-22.04
-            cmake_version: "4.1.2"
+          - os: ubuntu-24.04
+            cmake_version: "4.2.0"
             run_all_steps: true
 
     steps:
--- a/.github/actions/sail-setup/action.yml
+++ b/.github/actions/sail-setup/action.yml
@@ -4,7 +4,7 @@ inputs:
   cmake-version:
     description: "CMake version to install"
     required: false
-    default: "4.1.2"
+    default: "4.2.0"
   sail-version:
     description: 'Sail version to install (use "latest" to build from source)'
     required: false


.github/workflows/ci.yml
```diff
@@ -8,11 +8,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04, ubuntu-24.04-arm, macos-latest]
-        cmake_version: ["3.20.0", "4.1.2"]
+        os: [ubuntu-22.04, ubuntu-24.04, ubuntu-24.04-arm, macos-latest]
+        cmake_version: ["3.31.9", "4.2.0"]
```
jordancarlin, line 12: Our `CMakeLists.txt` files declare a minimum CMake version of 3.20, so that version should be maintained for testing.

.github/workflows/ci.yml
```diff
@@ -8,11 +8,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04, ubuntu-24.04-arm, macos-latest]
-        cmake_version: ["3.20.0", "4.1.2"]
+        os: [ubuntu-22.04, ubuntu-24.04, ubuntu-24.04-arm, macos-latest]
```
jordancarlin, line 11: I'm not sure there is much benefit to running CI on both Ubuntu 22.04 and Ubuntu 24.04 other than maybe different compiler versions?

jordancarlin (changes requested):

.github/workflows/ci.yml
Arielfoever: gcc and some lib like gmp

.github/workflows/ci.yml
Arielfoever: Just considering if we should bump cmake min version too

jordancarlin: Unless we have a strong reason to bump the minimum CMake version (a new feature we want to use that isn't supported, etc.), we should avoid increasing the minimum version. All that would do is remove support for the system CMake packaged in older distros without any benefit.

.github/workflows/ci.yml
Timmmm: Fair point, but we do already have `ubuntu-24.04-arm`. I think it's ok to test the combinations with a [pairwise testing approach](https://www.pairwise.org/) rather than exhaustively. Doesn't cost us any more I guess but think of the environment! :-D

If we run into an issue that this would have caught we can reevaluate.

jordancarlin:
@Arielfoever seems like the only part of this PR that we want to proceed with is bumping the default latest CMake version. Can you update this accordingly (drop the change to the minimum CMake version and the addition of Ubuntu-24.04)?

jordancarlin:
Closing this due to lack of author response. There was agreement that we don't want to move forward with most of this PR. The CMake version bump can be handled in a separate PR.

PR #1417, opened 2025-12-06
build: add cmake dependency check and update the build script

rpsene:
- Verify 'cmake' is installed before execution.
- Replace manual 'nproc' logic with 'cmake --parallel'.
- Add 'set -euo pipefail' for strict error handling.
- Parameterize build directory and build type.

### DIFF
--- a/build_simulator.sh
+++ b/build_simulator.sh
@@ -1,8 +1,28 @@
-#!/bin/sh
+#!/bin/bash
+set -euo pipefail
 
-set -e
+# 1. Dependency Verification
+# We use 'command -v' to check if cmake exists - POSIX standard (IEEE 1003.1)
+if ! command -v cmake >/dev/null 2>&1; then
+    echo "Error: 'cmake' is not installed or not in your PATH." >&2
+    echo "Please install CMake (https://cmake.org/download/) and try again." >&2
+    exit 1
+fi
+
+# Default values
 : "${DOWNLOAD_GMP:=TRUE}"
 : "${ENABLE_RISCV_TESTS:=TRUE}"
-cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DDOWNLOAD_GMP="${DOWNLOAD_GMP}" -DENABLE_RISCV_TESTS="${ENABLE_RISCV_TESTS}"
-jobs=$( (nproc || sysctl -n hw.ncpu || echo 2) 2>/dev/null)
-cmake --build build -j${jobs}
+: "${BUILD_TYPE:=RelWithDebInfo}"
+: "${BUILD_DIR:=build}"
+
+# 2. Configuration
+echo "Configuring build in '${BUILD_DIR}'..."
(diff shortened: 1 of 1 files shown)


build_simulator.sh
```diff
@@ -1,8 +1,28 @@
...
+echo "Configuring build in '${BUILD_DIR}'..."
+cmake -S . -B "${BUILD_DIR}" \
+      -DCMAKE_BUILD_TYPE="${BUILD_TYPE}" \
+      -DDOWNLOAD_GMP="${DOWNLOAD_GMP}" \
+      -DENABLE_RISCV_TESTS="${ENABLE_RISCV_TESTS}" \
+      "$@"
+
+# 3. Build
+echo "Building project..."
+cmake --build "${BUILD_DIR}" --parallel
```
jordancarlin, line 28: I've had issues on some systems when using an unbounded number of jobs, so would rather we keep the explicit number of jobs using `jobs=$( (nproc || sysctl -n hw.ncpu || echo 2) 2>/dev/null)`.

rpsene: @jordancarlin we can use, for instance: 

cmake --build "${BUILD_DIR}" --parallel 2

OR

: "${JOBS:=2}"
cmake --build "${BUILD_DIR}" --parallel "${JOBS}"

JOBS=8 ./build.sh

OR 

JOBS="$(
    set +e
    nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 2
)"
if [ "${JOBS}" -gt 2 ]; then
    JOBS=2
fi

cmake --build "${BUILD_DIR}" --parallel "${JOBS}"

Timmmm: I feel like we should just hard-code it to 4. It's simple, it'll work fine even on 1-core systems, and higher values don't really help because compile time is dominated by a few serial steps (Sail compiler, compiling the generated C++, final link).

Also this script was really intended to be mostly for documentation about how to build the model, and an easy "getting started" kind of thing, rather than the official way that you're supposed to build the model once you know what you're doing. So the emphasis should be on simplicity over configurability.

`BUILD_TYPE` is probably fine. I don't really think we need `BUILD_DIR` though - if you need that it's time to use CMake directly.

rpsene: @Timmmm @jordancarlin a bit of background: this change is to simplify a " Running SAIL" sample that will be include in the Getting Started with RISC-V training from LF that I'm reviewing right now. 

The improvements are implemented to add some guardrails to help those that are trying to run sail for the first time.

The complete, e2e sample is available here for now: https://github.com/rpsene/riscv-sail-daily-run

Timmmm: Ah nice. I definitely think the CMake check is a good idea, and `BUILD_TYPE` seems reasonable. What's the motivation for `BUILD_DIR` though?

jordancarlin: @rpsene I'm not sure exactly what the intention is for that training, but is there any reason for it not to use the prebuilt binary releases instead of building the model itself?

PR #1419, opened 2025-12-08
Add Hypervisor Extension

trdthg:
### Introduce

- Based on #612 with some refactor
- Also add hypervisor addition for Sstc, Zicfilp, CBO ...

### Test

tests are collected here https://github.com/trdthg/riscv-hypervisor-tests
- [x] https://github.com/trdthg/riscv-hext-asm-tests Basically no modifications, just update submodule. 
- [x] https://github.com/trdthg/riscv-hs-tests The original author didn't maintain it anymore. Fixed some assertions, also added htif support.
- [ ] https://github.com/trdthg/riscv-hyp-tests Only two tests left failed
	- one depends on #1350 , this should be easy to solve
	- another was marked as failed (both spike and sail) by @defermelowie, I haven't checked it out in depth

I would write a CI to distribute binary for them when I have time, like what we did for vector tests

### TODO

So this PR has been checked by multiple test suites, at least from the test situation, there is no big problem with the core address translation. I think it's time to get ready to split this big PR and commit gradually.

There are some areas that need improvement, and I'll list a few of them first. Any refactoring ideas would be greatly appreciated

- This PR refactors the return value of Exception, for example, in the past, when the access an address was failed, we would return an `(ExceptionType, addr)`, but now due to the hypervisor, this PR extends the addr to `ExceptionContext`, now we usually return `(ExceptionType, ExceptionContext)`, The `ExceptionContext` here will be passed from `execute` to `trap_handler`.
- Added 4 x4 variants to satp_mode, it is used in all the code related to address translation, and there is a lot of space for optimization
- For convenience, the virtaddr in the code is represented by bits(64) now, and considering that after the GVA -> GPA conversion, the GPA will continue to be translated to HPA as a VA, I think we need to add more comprehensive type constraints to physaddr and virtaddr

This PR is also constantly updated (you know, rebase to the latest to prevent it from getting too old), and dealing with the few remaining tests, and the refactoring mentioned above

---

## 2026_01_12 UPDATE

Recommended split and review order

- [ ] Other H related
	- [ ] Implement Two Step Translation #1419 
		- [ ] Extend TLB
			- [ ] Extend sv_width
				- [ ] Add H registers #1110
					- [ ] Refactor CSR Access
						- ~#1357~
						- #1402 (this is better)
				- [ ] Extend trap handler #1290
					- [ ] refactor exception #1451

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -170,8 +170,8 @@ bitfield Mstatus : bits(64) = {
   //MDT  : 42,
   MPELP: 41,
 
-  //MPV  : 39,
-  //GVA  : 38,
+  MPV  : 39,
+  GVA  : 38,
 
   MBE  : 37,
   SBE  : 36,
@@ -211,8 +211,8 @@ private function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
   let o = [o with
     // MDT = v[MDT],
     MPELP = if hartSupports(Ext_Zicfilp) then v[MPELP] else 0b0,
-    // MPV = v[MPV],
-    // GVA = v[GVA],
+    MPV = if currentlyEnabled(Ext_H) then v[MPV] else 0b0,
+    GVA = if currentlyEnabled(Ext_H) then v[GVA] else 0b0,
     // We don't currently support changing MBE and SBE.
     // MBE = v[MBE],
     // SBE = v[SBE],
@@ -279,6 +279,51 @@ function clause write_CSR((0x300, value) if xlen == 64) = { mstatus = legalize_m
 function clause write_CSR((0x300, value) if xlen == 32) = { mstatus = legalize_mstatus(mstatus, mstatus.bits[63 .. 32] @ value); Ok(mstatus.bits[31 .. 0]) }
 function clause write_CSR((0x310, value) if xlen == 32) = { mstatus = legalize_mstatus(mstatus, value @ mstatus.bits[31 .. 0]); Ok(mstatus.bits[63 .. 32]) }
 
+// Hypervisor Status Register
+
+bitfield Hstatus : bits(64) = {
+  HUPMM : 49 .. 48,
+  VSXL  : 33 .. 32,
+
+  VTSR  : 22,
+  VTW   : 21,
+  VTVM  : 20,
+  VGEIN : 17 .. 12,
+
+  HU    : 9,
+  SPVP  : 8,
+  SPV   : 7,
+  GVA   : 6,
+  VSBE  : 5
+}
+
+function legalize_hstatus(o : Hstatus, v : bits(64)) -> Hstatus = {
+  let v = Mk_Hstatus(v);
+  [o with
+    VTSR  = v[VTSR],
+    VTW   = v[VTW],
+    VTVM  = v[VTVM],
+    VGEIN = v[VGEIN],
+    HU    = v[HU],
+    SPVP  = v[SPVP],
+    SPV   = v[SPV],
+    GVA   = v[GVA],
+    VSBE  = v[VSBE],
+  ];
+}
+
+register hstatus : Hstatus = [ Mk_Hstatus(zeros()) with
+    // These fields do not exist on RV32 and are read-only zero
+    // if the corresponding mode is not supported.
+    VSXL = architecture_bits(if xlen == 32 then RV32 else RV64)
+  ]
+
+mapping clause csr_name_map = 0x600  <-> "hstatus"
+function clause is_CSR_accessible(0x600, _, _) = currentlyEnabled(Ext_H) // hstatus
+function clause read_CSR(0x600) = hstatus.bits[xlen - 1 .. 0]
+function clause write_CSR((0x600, value)  if xlen == 64) = { hstatus = legalize_hstatus(hstatus, value); Ok(hstatus.bits) }
+function clause write_CSR((0x600, value)  if xlen == 32) = { hstatus = legalize_hstatus(hstatus, hstatus.bits[63 .. 32] @ value); Ok(hstatus.bits[31 .. 0]) }
+
 // architecture and extension checks
 
 function architecture(priv : Privilege) -> Architecture = {
@@ -289,8 +334,8 @@ function architecture(priv : Privilege) -> Architecture = {
     Machine           => misa[MXL],
     Supervisor        => mstatus[SXL],
     User              => mstatus[UXL],
-    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => hstatus[VSXL],
+    VirtualUser       => hstatus[VSXL],
   })
 }
 
@@ -440,6 +485,43 @@ function clause write_CSR((0x30A, value) if xlen == 64) = { menvcfg = legalize_m
 function clause write_CSR((0x31A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, value @ menvcfg.bits[31 .. 0]); Ok(menvcfg.bits[63 .. 32]) }
 function clause write_CSR(0x10A, value) = { senvcfg = legalize_senvcfg(senvcfg, zero_extend(value)); Ok(senvcfg.bits[xlen - 1 .. 0]) }
 
+
+bitfield HEnvcfg : bits(64) = {
+  // Supervisor TimeCmp Extension
+  STCE   : 63,
+  // Page Based Memory Types Extension
+  PBMTE  : 62,
+  // Hardware update of A/D bits Extension
+  ADUE   : 61,
+  // Reserved WPRI bits.
+  wpri_2 : 60 .. 8,
+  // Cache Block Zero instruction Enable
+  CBZE   : 7,
+  // Cache Block Clean and Flush instruction Enable
+  CBCFE  : 6,
+  // Cache Block Invalidate instruction Enable
+  CBIE   : 5 .. 4,
+  // Reserved WPRI bits.
+  wpri_1 : 3,
+  // Landing Pad Enable
+  LPE    : 2,
+  wpri_0 : 1,
+  // Fence of I/O implies Memory
+  FIOM   : 0,
+}
+
+register henvcfg : HEnvcfg
+
+function legalize_henvcfg(o : HEnvcfg, v : bits(64)) -> HEnvcfg = {
+  let v = Mk_HEnvcfg(v);
+  [o with
+    FIOM = if sys_enable_writable_fiom then v[FIOM] else 0b0,
+  ]
+}
+
+mapping clause csr_name_map = 0x60A  <-> "henvcfg"
+mapping clause csr_name_map = 0x61A  <-> "henvcfgh"
+
 // Return whether or not FIOM is currently active, based on the current
 // privilege and the menvcfg/senvcfg settings. This means that I/O fences
 // imply memory fence.
@@ -448,8 +530,8 @@ function is_fiom_active() -> bool = {
     Machine           => false,
     Supervisor        => menvcfg[FIOM] == 0b1,
     User              => (menvcfg[FIOM] | senvcfg[FIOM]) == 0b1,
-    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => henvcfg[FIOM] == 0b1,
+    VirtualUser       => henvcfg[FIOM] == 0b1,
   }
 }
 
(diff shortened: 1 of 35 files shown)


defermelowie:
Which test in https://github.com/trdthg/riscv-hyp-tests is failing? Perhaps I can have a look and explain why I considered it an acceptable failure back then.

trdthg:
Ah, very glad to see you here! @defermelowie It's this one https://github.com/defermelowie/riscv-hyp-tests/blob/f8cf73bb339a88938f09c4bcc9609a95120f82fa/translation_tests.c#L295-L309

defermelowie:
Essentially, [the test](https://github.com/defermelowie/riscv-hyp-tests/blob/f8cf73bb339a88938f09c4bcc9609a95120f82fa/translation_tests.c#L295-L309) attempts to read a word using `hlvx.wu` from a (VS-level) page without execute permissions. It expects: 
1. A load page fault to be triggered - _ok_
2. `GVA` to be zero - **_fail_**
3. `MPV`/`SPV` to be zero - _ok_

The Guest Virtual Address field of `mstatus`/`hstatus` should reflect the presence of a guest virtual address in `mtval`/`stval`. Therefore, when `GVA` is zero, there can't be a guest virtual address in `mtval`/`stval`.

> Field GVA (Guest Virtual Address) is written by the implementation whenever a trap is taken M/HS-mode.
> For any trap (...) that writes a guest virtual address to `mtval`/`stval`, GVA is set to 1.
> For any other trap into M/HS-mode, GVA is set to 0.
>
> ~ Priv. Spec. v20240411 p. 131 (`hstatus`) & p. 147 (`mstatus`)

This leaves two options:
1. There is no address at all in `mtval`/`stval`
2. There is an address other then guest virtual in `mtval`/`stval`

The first case can only occur if `mtval`/`stval` is set zero on page faults (impl. specific behaviour). However, this behaviour itself is not asserted, only the side effect of `GVA` being 0. Furthermore, other tests in the suite expect an address in `mtval`/`stval` on page faults. Therefore, I assumed the suite expects an address in `mtval`/`stval` instead of unconditional zero. 

If the second case was true, it would mean that `mtval`/`stval` contains a guest physical or supervisor physical address. The main reason this is impossible is that due to the failing VS-level address translation, these addresses are simply not available.

allenjbaum:
I'm unclear why the test expects GVA to be zero, if the (implementation
specific ) behavior is to write an address into xtval when a page fault is
signalled,
then shouldn't GVA=1?  GVA is cause specific, and page_fault is one of the
causes that would write an address to Xtval (providing the implementation
specific behavior is to do that)


Field GVA (Guest Virtual Address) is written by the implementation whenever
a trap is taken into HS-mode.
For any trap (breakpoint, address misaligned, access fault, page fault, or
guest-page fault) that writes a guest virtual address to stval, GVA is set
to 1.
For any other trap into HS-mode, GVA is set to 0.


For breakpoint and memory access traps that write a nonzero value to stval,
GVA is redundant with field SPV (the two bits are set the same)
except when the explicit memory access of an HLV, HLVX, or HSV instruction
causes a fault. In that case, SPV=0 but GVA=1


Without knowing more details, that last sentence seems to be the case here.

There is a page fault, GVA should always be set to one, but in this case
SPV=0,

I don't see why a failing VS-level address is not available - its the input
to the translation process, not the output.

nadime15:
As discussed last Monday, I have created a new wiki page to better coordinate our work:

https://github.com/riscv/sail-riscv/wiki/Hypervisor-Implementation-Status

The page tracks our current status, including an overview of extensions that need to be revisited, as well as TODO's, issues, etc. Additions/Improvements are more than welcome!

model/core/sys_regs.sail
```diff
@@ -953,14 +1082,496 @@ val get_16_random_bits = impure {
...
+  // Init H-extension specific HS-mode CSRs
+
+  // Dynamic XLEN changes are not (yet) supported
+  hstatus = [Mk_Hstatus(zeros()) with VSXL = misa[MXL]];
+
+  hedeleg = Mk_Medeleg(zeros());
+  hideleg = Mk_Minterrupts(zeros());
+  hcounteren = Mk_Counteren(zeros());
+  htval = zeros();
+  htval = zeros();
```
nadime15:
```suggestion
  htval = zeros();
```
Remove duplicate

model/core/sys_regs.sail
```diff
@@ -289,8 +334,8 @@ function architecture(priv : Privilege) -> Architecture = {
     Machine           => misa[MXL],
     Supervisor        => mstatus[SXL],
     User              => mstatus[UXL],
-    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualSupervisor => hstatus[VSXL],
+    VirtualUser       => hstatus[VSXL],
```
nadime15: I think this should be: 
```suggestion
    VirtualSupervisor => hstatus[VSXL],
    VirtualUser       => vsstatus[UXL],
```

model/core/sys_regs.sail
```diff
@@ -953,14 +1082,496 @@ val get_16_random_bits = impure {
...
+// Debug/Trace Registers
+mapping clause csr_name_map = 0x6A8  <-> "hcontext"
+
+function clause is_CSR_accessible(0x6A8, _, _) = currentlyEnabled(Ext_H) // hcontext
+
+// TODO: Wait for Debug Extension
+
+// Virtual Supervisor Scratch Register
+register vsie : xlenbits
+register vsip : xlenbits
```
nadime15:
```suggestion
register vsie : xlenbits
register vsip : xlenbits
```

There are not being used, you dont need them.

model/sys/sys_control.sail
```diff
@@ -35,25 +85,46 @@ function check_CSR(csr : csreg, p : Privilege, access_type : CSRAccessType) -> b
...
+  let deleg = if currentlyEnabled(Ext_S) & del_to_hs then
+                if currentlyEnabled(Ext_H) & del_to_vs then
+                  VirtualSupervisor
+                else Supervisor
+              else Machine;
   // We cannot transition to a less-privileged mode.
-  if   privLevel_to_bits(deleg) <_u privLevel_to_bits(p)
+  if (privLevel_to_bits(deleg) <=_u privLevel_to_bits(p)) & (privLevel_to_virt_bits(deleg) >=_u privLevel_to_virt_bits(p))
   then p else deleg
 }
```
nadime15: I think thats not quite correct, you cant check unconditionally hedeleg, you have to check if V == 1 before considering hedeleg.

So more something like this: 

```
  let del_to_vs = privLevel_is_virtual(p) & currentlyEnabled(Ext_H) & bit_to_bool(hedeleg.bits[idx]);
  
  let deleg = if currentlyEnabled(Ext_S) & del_to_hs 
              then if del_to_vs then VirtualSupervisor else Supervisor
              else Machine
```

model/sys/vmem.sail
```diff
@@ -352,87 +463,215 @@ function translate forall 'v, is_sv_mode('v) . (
...
+          // On RV64 paddr can be 34 or 56 bits, so we zero extend to 64.
+          // On RV32 paddr can only be 34 bits. Sail knows this due to
+          // the assertion above.
+          Ok(Physaddr(zero_extend(paddr)), ext_ptw)
+        },
+        Err(f, ext_ptw)  => {
+          Err(f, ext_ptw)
+        }
+      }
+    },
```
nadime15:
```suggestion
function translate_stage(
  vAddr    : bits(64),                          // virtual address
  ac       : MemoryAccessType(ext_access_type), // Read/Write/ReadWrite/Execute
  priv     : Privilege,                         // Machine/Supervisor/User
  stage    : AddressTranslationStage,           // S/VS/G
  ext_ptw  : ext_ptw,                           // ext_ptw
) -> TR_Result(physaddr, PTW_Error) = {
  let mode = translationMode(priv, stage);
  if mode == Bare then return Ok(Physaddr(vAddr[physaddrbits_len - 1 .. 0]), init_ext_ptw);
  match stage {
    Stage_S => {
      // Sv39 -> 39, etc.
      let sv_width = satp_mode_width(mode);
      assert((xlen == 32 & sv_width == 32) | (xlen == 64 & sv_width != 32));
      let satp_sxlen = get_satp(sv_width);
      let svAddr = vAddr[sv_width - 1 .. 0];
      if vAddr != sign_extend(svAddr) then {
        // translationException(ac, stage, PTW_Invalid_Addr())
        return Err(PTW_Invalid_Addr(), init_ext_ptw);
      };

      let mxr    : bool     = mstatus[MXR] == 0b1;
      let do_sum : bool     = mstatus[SUM] == 0b1;
      let asid              = xsatp_to_asid(satp);

      // let ptb    : bits(64) = satp_to_PT_base(satp);
      let satp_sxlen = get_satp(sv_width);
      let base_ppn = satp_to_ppn(satp_sxlen);
      let res = translate(Stage_S,
                          sv_width,
                          zero_extend(asid),
                          base_ppn, // ptb
                          svAddr[sv_width - 1 .. pagesize_bits],
                          ac,
                          priv,
                          mxr,
                          do_sum,
                          ext_ptw);

      // Fixup result PA or exception
      match res {
        Ok(ppn, ext_ptw) => {
          // Step 10 of VATP.
          // Append the page offset. This is now a 34 or 56 bit address.
          let paddr = ppn @ vAddr[pagesize_bits - 1 .. 0];

          // On RV64 paddr can be 34 or 56 bits, so we zero extend to 64.
          // On RV32 paddr can only be 34 bits. Sail knows this due to
          // the assertion above.
          Ok(Physaddr(zero_extend(paddr)), ext_ptw)
        },
        Err(f, ext_ptw)  => {
          Err(f, ext_ptw)
        }
      }
    },
    Stage_VS => {
      let sv_width = satp_mode_width(mode);
      assert((xlen == 32 & sv_width == 32) | (xlen == 64 & sv_width != 32));
      let vsatp_sxlen = get_vsatp(sv_width);
      let svAddr = vAddr[sv_width - 1 .. 0];
      if vAddr != sign_extend(svAddr) then {
        return Err(PTW_Invalid_Addr(), init_ext_ptw);
      };

      let mxr    : bool = mstatus[MXR] == 0b1 | vsstatus[MXR] == 0b1; // HS-level MXR overwrites VS-stage page protections
      let do_sum : bool = vsstatus[SUM] == 0b1;
      let asid          = vsatp_to_asid(vsatp);
      let base_ppn      = vsatp_to_ppn(vsatp_sxlen);
      let res = translate(Stage_VS,
                          sv_width,
                          zero_extend(asid),
                          base_ppn, // ptb
                          svAddr[sv_width - 1 .. pagesize_bits],
                          ac,
                          priv,
                          mxr,
                          do_sum,
                          ext_ptw);
      // Fixup result PA or exception
      match res {
        Ok(ppn, ext_ptw) => {
          // Step 10 of VATP.
          // Append the page offset. This is now a 34 or 56 bit address.
          let paddr = ppn @ vAddr[pagesize_bits - 1 .. 0];
          // On RV64 paddr can be 34 or 56 bits, so we zero extend to 64.
          // On RV32 paddr can only be 34 bits. Sail knows this due to
          // the assertion above.
          Ok(Physaddr(zero_extend(paddr)), ext_ptw)
        },
        Err(f, ext_ptw)  => {
          Err(f, ext_ptw)
        }
      }
    },
    Stage_G => {
      let sv_width = hgatp_mode_width(mode);
      assert((xlen == 32 & sv_width == 34) | (xlen == 64 & sv_width != 34));
      let hgatp_sxlen = get_hgatp(sv_width);
      let svAddr = vAddr[sv_width - 1 .. 0];
      if vAddr != zero_extend(svAddr) then {
        return Err(PTW_Invalid_Addr(), init_ext_ptw);
      };

      let mxr    : bool     = mstatus[MXR] == 0b1;
      let do_sum : bool     = mstatus[SUM] == 0b1;
      let vmid   : bits(16) = zero_extend(hgatp_to_vmid(hgatp));
      let ptb               = hgatp_to_ppn(hgatp_sxlen);
      let res = translate(Stage_G,
                          sv_width,
                          vmid,
                          ptb,
                          svAddr[sv_width - 1 .. pagesize_bits],
                          ac,
                          // All G-stage memory accesses are considered User-level
                          User,
                          mxr,
                          do_sum,
                          ext_ptw);
      // Fixup result PA or exception
      match res {
        Ok(ppn, ext_ptw) => {
          // Step 10 of VATP.
          // Append the page offset. This is now a 34 or 56 bit address.
          let paddr = ppn @ vAddr[pagesize_bits - 1 .. 0];

          // On RV64 paddr can be 34 or 56 bits, so we zero extend to 64.
          // On RV32 paddr can only be 34 bits. Sail knows this due to
          // the assertion above.
          Ok(Physaddr(zero_extend(paddr)), ext_ptw)
        },
        Err(f, ext_ptw)  => {
          Err(f, ext_ptw)
        }
      }
    },
```

This is way to long, the cases are almost all the same

model/sys/vmem_tlb.sail
```diff
@@ -70,56 +70,72 @@ type num_tlb_entries : Int = 64
 type tlb_index_range = range(0, num_tlb_entries - 1)
 
 // PRIVATE
-register tlb : vector(num_tlb_entries, option(TLB_Entry)) = vector_init(None())
+register tlb_S  : vector(num_tlb_entries, option(TLB_Entry)) = vector_init(None())
+register tlb_VS : vector(num_tlb_entries, option(TLB_Entry)) = vector_init(None())
+register tlb_G  : vector(num_tlb_entries, option(TLB_Entry)) = vector_init(None())
```
nadime15: Any reason why we should have 3 TLB's instead of changing the size of one TLB?

jordancarlin:

PR #1420, opened 2025-12-10
Lean: adapt to Sail change

ineol:
We now need to open the `ConcurrencyInterfaceV1` namespace in the handwritten support file.


ineol:
Note that the following sail patch will fix the rest of the model compilation: https://github.com/rems-project/sail/pull/1564

Timmmm (approved):

PR #1421, opened 2025-12-11
Replace hardcoded constant in tlb_hash() with a compile-time constant

nadime15:
Replaced hardcoded constant in `tlb_hash()` with tlb_entries_bits to improve maintainability. This ensures that if the TLB size is changed in the future the hash computation will automatically change to the new number of entries. Before this change smaller values were caught by the type system, while larger ones were allowed, which could have led to unnecessary collisions and wasted TLB space.

Preferably I would have not introduced new type and used log2(), but Sail does not support it.

### DIFF
--- a/model/sys/vmem_tlb.sail
+++ b/model/sys/vmem_tlb.sail
@@ -66,9 +66,9 @@ function tlb_get_ppn forall 'v, is_sv_mode('v) . (
 
 // 64 entries is based on benchmarks of Linux boots and is where
 // you stop seeing performance improvements.
-type tlb_entries_bits : Int = 6
-let  tlb_entries_bits = sizeof(tlb_entries_bits)
-type num_tlb_entries : Int = 2 ^ tlb_entries_bits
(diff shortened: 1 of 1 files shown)


model/sys/vmem_tlb.sail
```diff
@@ -64,9 +64,11 @@ function tlb_get_ppn forall 'v, is_sv_mode('v) . (
   trunc(ppn | (vpn & levelMask))
 }
 
-// 64 entries is based on benchmarks of Linux boots and is where you stop
-// seeing performance improvements.
-type num_tlb_entries : Int = 64
+// 64 entries is based on benchmarks of Linux boots and is where
+// you stop seeing performance improvements.
+type tlb_entries_bits : Int = 6
```
pmundkur: Perhaps this type name can indicate that it is an exponent 
```suggestion
type num_tlb_entries_exp : Int = 6
```

Timmmm: I was thinking the same thing.

PR #1424, opened 2025-12-15
Add miselect CSR (0x350) and handle MSB=0 reserved values

ibvqeibob:
Implements miselect CSR (0x350) in the Sail model:
- Adds miselect register state
- Adds CSR name mapping and read/write clauses
- Defines deterministic handling for reserved values (MSB=0): keep previous value

### DIFF
--- a/model/core/extensions.sail
+++ b/model/core/extensions.sail
@@ -698,6 +698,6 @@ let extensions_ordered_for_isa_string = [
 
   // Machine mode extensions, ordered alphabetically.
   Ext_Smcntrpmf,
-  Ext_Smcntrpmf,
+  Ext_Smcsrind,
 
 ]


model/exceptions/sys_exceptions.sail
```diff
@@ -80,22 +80,39 @@ function set_stvec(value : xlenbits) -> xlenbits = {
...
 mapping clause csr_name_map = 0x141  <-> "sepc"
 mapping clause csr_name_map = 0x305  <-> "mtvec"
 mapping clause csr_name_map = 0x341  <-> "mepc"
+mapping clause csr_name_map = 0x350  <-> "miselect"
 
 function clause is_CSR_accessible(0x105, _, _) = currentlyEnabled(Ext_S) // stvec
 function clause is_CSR_accessible(0x141, _, _) = currentlyEnabled(Ext_S) // sepc
 function clause is_CSR_accessible(0x305, _, _) = true // mtvec
 function clause is_CSR_accessible(0x341, _, _) = true // mepc
+function clause is_CSR_accessible(0x350, _, _) = true // miselect
```
nadime15:
```suggestion
function clause is_CSR_accessible(0x350, _, _) = false // miselect
```

nadime15 (commented):
Unless there is a reason to merge it, I am a bit hesitant to approve this PR with only this single register. miselect is part of Smcsrind or AIA, so this register is not available by default. Given that, what is the point of having it? I think it should not be too hard to add Smcsrind and Sscsrind instead. Maybe you want to give it a try?

Timmmm:
> I think it should not be too hard to Smcsrind or Sscsrind instead.

There's a PR for those [here](https://github.com/riscv/sail-riscv/pull/956). It would be good to get that merged. But I agree we should avoid adding just part of an extension.

model/exceptions/sys_exceptions.sail
ibvqeibob: Thanks! I added a Smcsrind extension toggle (default off) and gated miselect (CSR 0x350) behind currentlyEnabled(Ext_Smcsrind), so it’s not exposed in default configs but becomes available when Smcsrind is enabled.

model/core/extensions.sail
```diff
@@ -693,4 +698,6 @@ let extensions_ordered_for_isa_string = [
 
   // Machine mode extensions, ordered alphabetically.
   Ext_Smcntrpmf,
+  Ext_Smcntrpmf,
```
nadime15:
```suggestion
  Ext_Smcsrind,
```

ibvqeibob:
> > I think it should not be too hard to Smcsrind or Sscsrind instead.
> 
> There's a PR for those [here](https://github.com/riscv/sail-riscv/pull/956). It would be good to get that merged. But I agree we should avoid adding just part of an extension.

Thanks! I agree we should avoid adding only part of an extension by default. In this PR I’ve gated miselect behind a new Smcsrind extension toggle (default off), so default configurations are unchanged. That said, I’m happy to align with the existing Smcsrind/Sscsrind PR you linked: once that lands I can rebase and adjust this PR to reuse the upstream extension definitions (or close this PR and move the changes there—whichever you prefer).

nadime15:
@ibvqeibob Yeah, we discussed this yesterday in our golden meeting and we think it is better to close this PR, if that is ok for you, and rebase on #956. If you are interested, feel free to pick up that PR and incorporate your changes into it. Otherwise, one of us can take a look, but I think this will take a while because of the upcoming holiday break for most people.

PR #1429, opened 2025-12-15
Fix location of the `hartSupports_measure` function.

pmundkur:
It was defined in a location that seems to have preceded the compiler's knowledge of all the variants of the `extension` enum, resulting in redundant match warnings from the latest Sail compiler:
```
Warning: Redundant case core/extensions.sail:25.4-15:
25 |    Ext_Sstvecd => 1, // > S
   |    ^---------^
   |
This match case is never used

Warning: Redundant case core/extensions.sail:26.4-15:
26 |    Ext_Ssu64xl => 1, // > S
   |    ^---------^
   |
This match case is never used
...
```
and so on, due to the first match pattern `Ext_D` being treated as a variable that made all subsequent matches redundant.

This moves the function next to `currentlyEnabled_measure` and adds comments.

### DIFF
--- a/model/core/extensions.sail
+++ b/model/core/extensions.sail
@@ -19,25 +19,6 @@ scattered mapping extensionName
 val hartSupports : extension -> bool
 scattered function hartSupports
 
-function hartSupports_measure(ext : extension) -> int =
-  match ext {
-    Ext_D => 1, // > F
-    Ext_Sstvecd => 1, // > S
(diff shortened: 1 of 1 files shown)


pmundkur:
@Alasdair is this expected?

nadime15:
I also get warnings about unused variables since this was merged, but I assumed its due to my older compiler version and the newest compiler wont show this anymore.

PR #1432, opened 2025-12-17
Add missing `csr_write_callback` in `accrue_fflags`

jordancarlin:
This currently works the same way as other implicit CSR write callbacks - it is always called when the CSR may potentially be changed. In future we may add configuration options, e.g. to allow distinguishing the cases where no new flags need to be set, or when they do but they're already set.

### DIFF
--- a/model/extensions/FD/fdext_regs.sail
+++ b/model/extensions/FD/fdext_regs.sail
@@ -437,5 +437,7 @@ function accrue_fflags(flags : bits(5)) -> unit = {
   then {
     fcsr[FFLAGS] = f;
     dirty_fd_context_if_present();
-  }
+  };
+  csr_write_callback("fflags", zero_extend(fcsr[FFLAGS]));
+  csr_write_callback("fcsr", zero_extend(fcsr.bits));
 }


model/extensions/FD/fdext_regs.sail
arichardson, line 427: Should it be added here as well?

model/extensions/FD/fdext_regs.sail
```diff
@@ -437,5 +437,7 @@ function accrue_fflags(flags : bits(5)) -> unit = {
   then {
     fcsr[FFLAGS] = f;
     dirty_fd_context_if_present();
-  }
+  };
+  csr_write_callback("fflags", zero_extend(fcsr[FFLAGS]));
```
arichardson, line 441: Should this be inside the if?

model/extensions/FD/fdext_regs.sail
jordancarlin: The `write_fcsr` function is only called by `write_CSR`, which is used as part of the Zicsr instruction execute clause and already includes the callback.

model/extensions/FD/fdext_regs.sail
jordancarlin: This matches how other implicit updates are currently handled. For example, the CSR callback is called every time `mstatus.FS` _might_ be set to dirty, even if it is already set to dirty. We can contemplate a global change to only use the callbacks when something actually changes, but for this PR I'd rather keep it as it is.

Timmmm: Yeah in Codasip we ran into issues with fflags when trying to perfectly align the CSR writes recorded by the DUT and the model. There are actually several reasonable options:

1. Record writes for instructions that *might* update fflags (i.e. everything that calls `accrue_fflags`.
2. Record writes when the instruction is setting a flag, even if it's already set.
3. Record writes when fflags actually changes.

This is option 1. I think it's fine, but doesn't it mean that virtually every floating point instruction will have a write to fflags in the log? That might be pretty annoying but I guess nobody has complained about mstatus yet... It's probably fine for now.

Also, slightly off-topic but my conclusion from this in Codasip is that exactly trying to match register writes from a DUT and model is the wrong thing to do. There are other situations where it's ambiguous whether to record a write or not (e.g. for agnostic masked vector registers). What you really want to compare is the *state* not the state changes.

jordancarlin: There are definitely reasons for all three of those approaches (and as much as I don't want to, we might need to add configurability to pick between them eventually). For now the model seems to globally follow option 1, and I think we should stick with that until there is a reason to do otherwise.

Totally agreed regarding comparing state vs. state changes.

PR #1435, opened 2025-12-18
Ensure that `rd` is written by `vset{i}vl{i}` even for illegal and reserved `vsew` and `vlmul`.

### LINKED ISSUE #1434: vsetvli: Reserved VSEW argument leaves destination register (rd) unchanged, should be set to new value of VL (zero).
## Expected Behavior
When a `VTYPE` field has `Reserved` value, the `VL` CSR should be set to zero and the destination register of `vsetvli` should receive the new, updated value.

## Current Behavior
The destination register keeps the old value. 

## Steps to Reproduce
1. Extract the `elf` file `rv64uv-p-vsetvli_vl_lt_vlmax` from the attached zip archive: [rv64-vsetvli-reserved-vsew.zip](https://github.com/user-attachments/files/24246405/rv64-vsetvli-reserved-vsew.zip) which also has the source (`.S`) and the `objdump` (`.dump`) files for reference.
2. Run 
`sail_riscv_sim --trace-instr --trace-reg --config <PATH_TO_SAIL>/share/sail-riscv/config/rv64d_v128_e64.json rv64uv-p-vsetvli_vl_lt_vlmax`
(I had tried the binary release v0.9, please replace `<PATH_TO_SAIL>` with your path.)
3. Note that the destination register `x5` is not updated with the new value of `VL` (zero) after the second `vsetvli` at `test2+8` (whereas after `test1+8` it *is* updated correctly). 

> [77] [U]: 0x00000000800001A4 (0x00100193) addi x3, x0, 0x1                          test_1+0
x3 <- 0x0000000000000001
[78] [U]: 0x00000000800001A8 (0x00200213) addi x4, x0, 0x2                          test_1+4
x4 <- 0x0000000000000002
[79] [U]: 0x00000000800001AC (0x003272D7) vsetvli x5, x4, e8, m8, tu, mu            test_1+8
x5 <- 0x0000000000000002
CSR mstatus (0x300) <- 0x8000000A00006680
CSR vstart (0x008) <- 0x0000000000000000
CSR vtype (0xC21) <- 0x0000000000000003
CSR vl (0xC20) <- 0x0000000000000002
CSR vstart (0x008) <- 0x0000000000000000
[80] [U]: 0x00000000800001B0 (0x00200393) addi x7, x0, 0x2                          test_1+12
x7 <- 0x0000000000000002
[81] [U]: 0x00000000800001B4 (0x02729863) bne x5, x7, 0x30                          test_1+16
[82] [U]: 0x00000000800001B8 (0x00200193) addi x3, x0, 0x2                          test_2+0
x3 <- 0x0000000000000002
[83] [U]: 0x00000000800001BC (0x00200213) addi x4, x0, 0x2                          test_2+4
x4 <- 0x0000000000000002
[84] [U]: 0x00000000800001C0 (0x023272D7) vsetvli x5, x4, 0x23                      test_2+8
CSR vtype (0xC21) <- 0x8000000000000000
CSR vl (0xC20) <- 0x0000000000000000
CSR mstatus (0x300) <- 0x8000000A00006680
CSR vstart (0x008) <- 0x0000000000000000
[85] [U]: 0x00000000800001C4 (0x00000393) addi x7, x0, 0x0                          test_2+12
x7 <- 0x0000000000000000
[86] [U]: 0x00000000800001C8 (0x00729E63) bne x5, x7, 0x1c                          test_2+16
[87] [U]: 0x00000000800001E4 (0x0FF0000F) fence iorw, iorw                          fail+0

## Context (Environment)
I created a test with two instances of `vsetvli` instruction (using the `TEST_CASE` macro of the [riscv-tests](https://github.com/riscv-software-src/riscv-tests) unit test framework). The first with a legal `vtypei` and the second with a reserved `VSEW`. The second test fails because I had passed in zero as the expected value of the new `VL` stored in the destination register of the `TEST_CASEs` but it contained the old value instead.

pmundkur:
Fixes #1434.

Also adds a note to CONTRIBUTING.md on quoting the spec when fixing similar bugs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,9 @@ automated verification that is presented in a way that the reviewers
 of the pull request can run. It is desirable that the pull request
 explains how it relates to the existing RISC-V architectural tests.
 
+When contributing a bug fix, it is often helpful to point to or quote
+the pertinent portion of the spec in the PR or the code when relevant.
+
 We recommend installing pre-commit hooks that ensure certain basic coding
 style issues can be detected and fixed before submitting the pull request.
 To set up these hooks, install [pre-commit](https://pre-commit.com)


model/extensions/V/vext_vset_insts.sail
jordancarlin: Missing setting `rd` here. It might be easier to put this into `handle_illegal_vtype()` and either have it return `vl` (which `rd` gets set to) or pass it `rd` as an input and have it set `rd` directly.

jordancarlin (changes requested):

model/extensions/V/vext_vset_insts.sail
pmundkur: Good catch, thanks!

model/extensions/V/vext_vset_insts.sail
```diff
@@ -49,14 +49,17 @@ mapping vtype_assembly : string <-> (bits(1), bits(1), bits(3), bits(3)) = {
...
-  // NOTE: vstart should NOT be zeroed out if trapping.
+  // NOTE: vstart should NOT be zeroed out and rd
+  //       should NOT be written if trapping.
 
   vtype.bits = 0b1 @ zeros(xlen - 1); // set vtype.vill
   vl = zeros();
+  // The new value of `vl` needs to be written to `rd`.
+  X(rd) = vl;
   csr_write_callback("vtype", vtype.bits);
   csr_write_callback("vl", vl);
```
jordancarlin: Minor nit, but the log will now show the `rd` write before the CSR write, which isn't really what's happening here. Might as well put the `X(rd) = vl` after the `csr_write_callback`.

nadime15:
It would really help if this and similar issues in the future came with a direct quote (and/or section reference) from the spec. Right now the issue described the expected behavior, but without a spec citation it is hard to quickly verify whether this is a real spec violation or just an interpretation issue, and it means we have to dig through the spec ourselves. Ideally, this note should be added to CONTRIBUTING.md as well.

Timmmm:
I think the code was technically correct already because it is handling *reserved* LMUL/SEW values and usually when things are reserved you can do whatever you like.

But it is slightly ambiguous (in RISC-V? no!) because [the spec](https://riscv-isa-manual.timhutt.co.uk/riscv-unprivileged.html#_unsupported_vtype_values) does also say this:

> A set of instructions is provided to allow rapid configuration of the values in vl and vtype to match application needs. The vset{i}vl{i} instructions set the vtype and vl CSRs based on their arguments, and write the new value of vl into rd.

> If the vtype value is not supported by the implementation, then the vill bit is set in vtype, the remaining bits in vtype are set to zero, and the vl register is also set to zero.

And presumably reserved values are "not supported by the implementation"?

Given that valid reserved behaviour is to set `vill`, and zero `vtype` and `vl` and `rd` then that seems like a reasonable default implementation that covers everything. It's definitely weird to set `vl` but not `rd` so I think we should merge this.

pmundkur:
I think the key clause is " The vset{i}vl{i} instructions ... and write the new value of vl into rd."  The writing is done even for illegal and reserved arguments.  I had checked with Spike before fixing this.

nadime15 (approved):

adingank-qualcomm:
> It would really help if this and similar issues in the future came with a direct quote (and/or section reference) from the spec. Right now the issue described the expected behavior, but without a spec citation it is hard to quickly verify whether this is a real spec violation or just an interpretation issue, and it means we have to dig through the spec ourselves. Ideally, this note should be added to CONTRIBUTING.md as well.

Which is "the" spec you'd like reporters to point to? E.g., I use the [PDF from RVI lists](https://lists.riscv.org/g/tech-vector-ext/attachment/691/0/riscv-v-spec-1.0.pdf) as my spec reference for Vector, but I see that Tim points to another URL in his own domain... Should we all refer to the [HTML from RVI docs](https://docs.riscv.org/reference/isa/unpriv/v-st-ext.html) site?

adingank-qualcomm:
I apologize if commenting on a closed PR is seen as not a good practice in this community!

Also, what's the proper way to report sightings for the changes in this PR?

Timmmm:
> Which is "the" spec you'd like reporters to point to?

Good question! We don't have proper spec versioning in the model yet (though it is on my todo list and I have been very slowly working on it), so currently it's just "the most recent ratified one", and we've generally just been adding quotes rather than linking anywhere, because there isn't really an official stable place to link to yet (supposedly that's coming).

> I apologize if commenting on a closed PR is seen as not a good practice in this community!

I think it's fine :-)

> what's the proper way to report sightings for the changes in this PR?

What do you mean by "sightings"?

adingank-qualcomm:
@Timmmm By "sighting" I meant an issue I saw but not wasn't confirmed as a bug yet. 
E.g., the tracer issue that I reported separately. Using the verification terminology... 

> > Which is "the" spec you'd like reporters to point to?
> 
> Good question! We don't have proper spec versioning in the model yet (though it is on my todo list and I have been very slowly working on it), so currently it's just "the most recent ratified one", and we've generally just been adding quotes rather than linking anywhere, because there isn't really an official stable place to link to yet (supposedly that's coming).
> 
> > I apologize if commenting on a closed PR is seen as not a good practice in this community!
> 
> I think it's fine :-)
> 
> > what's the proper way to report sightings for the changes in this PR?
> 
> What do you mean by "sightings"?

Timmmm:
Still not sure exactly what you mean - I've never heard that terminology before! But anyway if there are remaining issues I think it's probably best to open another Github issue, then it won't be forgotten.

adingank-qualcomm:
> Still not sure exactly what you mean - I've never heard that terminology before! But anyway if there are remaining issues I think it's probably best to open another Github issue, then it won't be forgotten.

No outstanding issues, thanks for all the prompt help from everyone!
Happy holidays and a happy new year!

PR #1436, opened 2025-12-19
Add RV32 mieh/miph and implement mtopi

ibvqeibob:
Add RV32 high-half interrupt CSRs mieh (0x314) and miph (0x354) (guarded by xlen==32).
Implement mtopi (0xFB0) as a read-only CSR reporting the highest-priority pending+enabled interrupt using the AIA default order.

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -511,29 +511,84 @@ function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
 
 register mie     : Minterrupts // Enabled
 register mip     : Minterrupts // Pending
+register mieh    : bits(32) // RV32: high half of mie (AIA)
+register miph    : bits(32) // RV32: high half of mip (AIA)
 register medeleg : Medeleg     // Exception delegation to S-mode
 register mideleg : Minterrupts // Interrupt delegation to S-mode
 
+function mie_wide64() -> bits(64) = {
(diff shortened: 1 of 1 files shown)


model/core/sys_regs.sail
```diff
@@ -511,29 +511,84 @@ function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
 
 register mie     : Minterrupts // Enabled
 register mip     : Minterrupts // Pending
+register mieh    : bits(32) // RV32: high half of mie (AIA)
+register miph    : bits(32) // RV32: high half of mip (AIA)
```
nadime15, line 515: mieh, miph, and topi are from the Smaia/Ssaia extensions, right? We should avoid adding registers from extensions we do not support :/

ibvqeibob: Thanks! Yes — mieh/miph and mtopi are AIA CSRs (Smaia/Ssaia).
I agree,I’ll update this PR to introduce an explicit AIA toggle (e.g. Ext_Smaia, default OFF) and gate is_CSR_accessible (and the RV32 high-half CSRs) behind it, so default configurations remain unchanged?

nadime15: We had the exact same issue in #1424 right and we found it to better to close the PR. Unless there is a good reason to keep this open it would be better to actual introduce the extension along with the register.

PR #1437, opened 2025-12-19
c_emulator: validate DTB range against configured PMA regions

### LINKED ISSUE #1308: Better checks for DTB with configured PMA memory regions
When specifying a DTB with a PMA-based configuration after #866, it would be nice to ensure:
- its address is in a memory region
- the DTB fits within that region

ibvqeibob:
Add a pre-write check that the DTB range [dtb_address, dtb_address + dtb_size) is fully contained in a single configured memory.regions PMA memory region; otherwise print an error and exit.

Fixes #1308.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -304,6 +304,28 @@ uint64_t load_sail(const std::string &filename, bool main_file)
 void write_dtb_to_rom(const std::vector<uint8_t> &dtb)
 {
   uint64_t addr = get_config_uint64({"memory", "dtb_address"});
+  uint64_t size = static_cast<uint64_t>(dtb.size());
+
+  // Overflow check for addr + size - 1
+  uint64_t end = addr + size - 1;
+  if (end < addr) {
+    fprintf(stderr,
+            "DTB address/size overflow: addr=0x%0" PRIx64 ", size=0x%0" PRIx64
+            "\n",
+            addr, size);
+    exit(EXIT_FAILURE);
+  }
+
+  // Validate DTB range against configured PMA memory regions.
+  if (!g_model.zdtb_within_configured_pma_memory(addr, size)) {
+    fprintf(
+        stderr,
+        "DTB does not fit in any configured PMA memory region: "
+        "addr=0x%0" PRIx64 ", size=0x%0" PRIx64 " (end=0x%0" PRIx64 ")\n"
+        "Hint: adjust memory.dtb_address or memory.regions in the config.\n",
+        addr, size, end);
+    exit(EXIT_FAILURE);
+  }
 
   for (uint8_t d : dtb) {
     write_mem(addr++, d);
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -192,6 +192,10 @@ function check_pma_regions(pmas : list(PMA_Region), prev_base : bits(64), prev_s
     },
   }
 
+// Return true iff [addr, addr+size) is fully contained in a single configured PMA memory region.
+function dtb_within_configured_pma_memory(addr : bits(64), size : bits(64)) -> bool =
+  is_some(matching_pma_bits_range(pma_regions, addr, size))
+
 // Check that all memory regions are sorted and don't overlap.
 function check_mem_layout() -> bool =
   if pma_regions == [||] then {
--- a/model/sys/pma.sail
+++ b/model/sys/pma.sail
@@ -104,16 +104,22 @@ struct PMA_Region = {
   include_in_device_tree : bool,
 }
 
-// Get the first PMA that matches a given address range.
-function matching_pma(pmas : list(PMA_Region), addr : physaddr, width : mem_access_width) -> option(PMA_Region) = {
+// Get the first PMA region that fully contains a given [base, base+size) range.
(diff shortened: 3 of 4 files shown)


model/sys/pma.sail
Timmmm: Can we delegate this to `matching_pma_range` now?

model/sys/pma.sail
```diff
@@ -116,6 +116,18 @@ function matching_pma(pmas : list(PMA_Region), addr : physaddr, width : mem_acce
   }
 }
 
+// Get the first PMA region that fully contains a given [base, base+size) range.
+function matching_pma_range(pmas : list(PMA_Region), base : bits(64), size : bits(64)) -> option(PMA_Region) = {
```
Timmmm: Maybe should be called `matching_pma_bits` seeing as the distinction is that the arguments are bits instead of integers?

c_emulator/riscv_sim.cpp
```diff
@@ -304,6 +304,31 @@ uint64_t load_sail(const std::string &filename, bool main_file)
 void write_dtb_to_rom(const std::vector<uint8_t> &dtb)
 {
   uint64_t addr = get_config_uint64({"memory", "dtb_address"});
+  uint64_t size = static_cast<uint64_t>(dtb.size());
+
+  if (size == 0) {
+    return;
+  }
```
Timmmm: Probably remove this check; I think it's unnecessary.

c_emulator/riscv_sim.cpp
```diff
@@ -304,6 +304,31 @@ uint64_t load_sail(const std::string &filename, bool main_file)
...
+
+  // Overflow check for addr + size - 1
+  uint64_t end = addr + size - 1;
+  if (end < addr) {
+    fprintf(stderr,
+            "DTB address/size overflow: addr=0x%0" PRIx64 ", size=0x%0" PRIx64 "\n",
+            addr, size);
+    exit(1);
```
```suggestion
    exit(EXIT_FAILURE);
```

We should change these to exceptions at some point probably but this is fine for now.

Timmmm (changes requested):
Nice, thanks!

c_emulator/riscv_sim.cpp
ibvqeibob: Thanks for the review! I updated the code accordingly:
Renamed the helper to matching_pma_bits_range to clarify the bits-based API.
Refactored matching_pma to delegate to matching_pma_bits_range.
Dropped the zero-size DTB early return and switched to exit(EXIT_FAILURE).

model/postlude/validate_config.sail
```diff
@@ -192,6 +192,15 @@ function check_pma_regions(pmas : list(PMA_Region), prev_base : bits(64), prev_s
     },
   }
 
+// Return true iff [addr, addr+size) is fully contained in a single configured PMA memory region.
+function dtb_within_configured_pma_memory(addr : bits(64), size : bits(64)) -> bool = {
+  if size == zeros() then true
```
Timmmm: I think this is unnecessary too?

model/postlude/validate_config.sail
```diff
@@ -192,6 +192,15 @@ function check_pma_regions(pmas : list(PMA_Region), prev_base : bits(64), prev_s
...
+  else match matching_pma_bits_range(pma_regions, addr, size) {
```
Timmmm: I believe there's an `is_some` function you can use, so 

```
function is_within_configured_pma_memory(addr : bits(64), size : bits(64)) -> bool =
  is_some(matching_pma_bits_range(pma_regions, addr, size))
```

Also this isn't really dtb specific so we may as well make the name generic.

c_emulator/riscv_sim.cpp
```diff
@@ -304,6 +304,29 @@ uint64_t load_sail(const std::string &filename, bool main_file)
...
+
+  // Validate DTB range against configured PMA memory regions.
+  if (!g_model.zdtb_within_configured_pma_memory(addr, size)) {
+    fprintf(stderr,
+            "DTB does not fit in any configured PMA memory region: "
+            "addr=0x%0" PRIx64 ", size=0x%0" PRIx64 " (end=0x%0" PRIx64 ")\n",
+            addr, size, end);
+    fprintf(
+        stderr,
+        "Hint: adjust memory.dtb_address or memory.regions in the config.\n");
```
Timmmm: Maybe move this into the same fprintf above?

model/postlude/validate_config.sail
ibvqeibob: Thanks! I dropped the explicit size == 0 case and simplified it to is_some.

c_emulator/riscv_sim.cpp
ibvqeibob: > Maybe move this into the same fprintf above?

Also folded the hint into the same fprintf as suggested.

jordancarlin:
How did this get through the merge queue even though CI is failing? It looks like the CI passed in the merge queue but failed once it was actually merged into master. Pre-commit is the part that failed, so it wasn't just a transient issue or anything like that.

pmundkur:
It looks like CI for PR #1437  failed both on merge queue (Job 769, https://github.com/riscv/sail-riscv/actions/runs/20458313718) as well as once it was on master (Job 770, https://github.com/riscv/sail-riscv/actions/runs/20458784253).  So this looks like a Github actions bug.

PR #1439, opened 2025-12-19
Improve handling of reserved fences

### LINKED ISSUE #1014: Reserved fence handling is not quite right
We have this code:

```
/* The reserved fences are not really hints, but for now they
 * are in this file to be also treated as nops to avoid trapping.
 */

union clause ast = FENCE_RESERVED : (bits(4), bits(4), bits(4), regidx, regidx)

mapping clause encdec = FENCE_RESERVED(fm, pred, succ, rs, rd)
  <-> fm : bits(4) @ pred : bits(4) @ succ : bits(4) @ encdec_reg(rs) @ 0b000 @ encdec_reg(rd) @ 0b0001111
      when (fm != 0b0000 & fm != 0b1000) | rs != zreg | rd != zreg

function clause execute (FENCE_RESERVED(fm, pred, succ, rs, rd)) = RETIRE_SUCCESS
```

However this isn't quite right on a few fronts. The spec says:

> The unused fields in the FENCE instructions--rs1 and rd--are reserved for finer-grain fences in future extensions. For forward compatibility, base implementations shall ignore these fields, and standard software shall zero these fields. Likewise, many fm and predecessor/successor set settings are also reserved for future use. Base implementations shall treat all such reserved configurations as FENCE instructions (with fm=0000), and standard software shall use only non-reserved configurations.

Therefore:

1. They should *ignore* `rs` and `rd`, not check that they are non-zero.
2. The execute clause shouldn't be a nop - it should delegate to `FENCE`, or do the same thing as it (call `sail_barrier()`).
3. Even if `fm == 0b1000` the instruction can still be reserved if it isn't `rw, rw`. In fact it should be executed with `fm = 0b0000`.

This isn't too critical in the Sail model since fences don't really do anything, but I suspect one day we will want to expose them to some kind of verification callback / logging interface and then they should be correct.

Timmmm:
This fixes the handling of reserved fences. When fences have reserved values in their fields they are supposed to be ignored, whereas currently the model will raise an exception.

This also removes the `fence.reserved` assembly syntax. If you try to disassemble a reserved fence you'll get a pattern match failure. There doesn't seem to be a good option here.

While here I also uncommented the `sail_barrier()` call for `fence.i` which seems to compile fine.

Fixes #1014

### DIFF
--- a/model/postlude/insts_end.sail
+++ b/model/postlude/insts_end.sail
@@ -39,6 +39,6 @@ end encdec_compressed
 
 // Some legal instructions do not have any corresponding assembly, e.g. reserved fences.
 function instruction_to_str(insn : instruction) -> string =
-    if assembly_forwards_matches(insn) then assembly(insn) else (".insn " ^ bits_str(encdec(insn)))
+  if assembly_forwards_matches(insn) then assembly(insn) else (".insn " ^ bits_str(encdec(insn)))
 
 overload to_str = {instruction_to_str}


Timmmm:
Unfortunately this doesn't quite compile yet because I get this:

```
Error:
postlude/insts_end.sail:40.9-19:
40 |function print_insn(insn : instruction) -> string = assembly(insn)
   |         ^--------^
   | Function print_insn calls function marked non-executable
```

I'll see if I can make a small reproducer for @Alasdair later.

What do you think about handling `assembly` for instructions that don't have any syntax @pmundkur @jordancarlin @nadime15 ? This is one of the things that I got stuck at when I was trying to verify our assembly against GCC and Clang for all possible instructions.

Timmmm:
Here's the bug: https://github.com/rems-project/sail/issues/1577

model/extensions/I/base_insts.sail
```diff
@@ -429,10 +429,27 @@ mapping clause assembly = SHIFTIWOP(shamt, rs1, rd, op)
...
+  <-> "fence.tso"
+
+// *****************************************************************
+union clause instruction = FENCE : (bits(4), bits(4), bits(4), regidx, regidx)
 
-mapping clause encdec = FENCE(pred, succ)
-  <-> 0b0000 @ pred @ succ @ 0b00000 @ 0b000 @ 0b00000 @ 0b0001111
+// Note: FENCE_TSO is encoded as a FENCE so that must come first. There is also
+// a PAUSE hint that is encoded as a FENCE so if we add support for that
+// (e.g. to improve assembly output), it must also come first.
```
jordancarlin: PAUSE is already supported: `model/extensions/Zihintpause/zihintpause_insts.sail`. Drop this comment.

jordancarlin:
> What do you think about handling `assembly` for instructions that don't have any syntax @pmundkur @jordancarlin @nadime15 ? This is one of the things that I got stuck at when I was trying to verify our assembly against GCC and Clang for all possible instructions.

What about just printing `.insn <encoding>`like the assemblers do?

Timmmm:
> What about just printing .insn <encoding>like the assemblers do?

I did consider that ... it's a bit weird because that's actually a directive rather than an instruction.

Alternatively we could have no match and then guard calls to `assembly()` with `assembly_backwards_matches()` or whatever?

```
function print_insn(insn : instruction) -> string = if assembly_backward_matches(insn) then assembly(insn) else ".insn " ^ bits_str(encdec(insn))
```

The advantage is you can programmatically detect if there's valid assembly which is useful for my script to validate the assembly syntax.

jordancarlin:
> > What about just printing .insn like the assemblers do?
> 
> I did consider that ... it's a bit weird because that's actually a directive rather than an instruction.
> 
> Alternatively we could have no match and then guard calls to `assembly()` with `assembly_backwards_matches()` or whatever?
> 
> ```
> function print_insn(insn : instruction) -> string = if assembly_backward_matches(insn) then assembly(insn) else ".insn " ^ bits_str(encdec(insn))
> ```
> 
> The advantage is you can programmatically detect if there's valid assembly which is useful for my script to validate the assembly syntax.

That seems like a much better solution. Also make it easier to handle this going forward because I'm sure it won't be the only time this comes up.

jordancarlin (approved):
LGTM

model/extensions/I/base_insts.sail
```diff
@@ -441,28 +456,33 @@ function effective_fence_set(set : bits(4), fiom : bool) -> bits(4) = {
   } else set
 }
 
-function clause execute FENCE(pred, succ) = {
+function clause execute FENCE(fm, pred, succ, rs, rd) = {
```
pmundkur:
```suggestion
function clause execute FENCE(_fm, pred, succ, _rs, _rd) = {
```

model/extensions/Zifenci/zifencei_insts.sail
```diff
@@ -13,14 +13,31 @@
...
 $[wavedrom "0 0 FENCE.I 0 MISC-MEM"]
-mapping clause encdec = FENCEI()
-  <-> 0b000000000000 @ 0b00000 @ 0b001 @ 0b00000 @ 0b0001111
-  when currentlyEnabled(Ext_Zifencei)
+mapping clause encdec = FENCEI(imm, rs, rd)
+  <-> imm : bits(12) @ encdec_reg(rs) @ 0b001 @ encdec_reg(rd) @ 0b0001111
 
 // fence.i is a nop for the memory model
-function clause execute FENCEI() = { /* __barrier(Barrier_RISCV_i); */ RETIRE_SUCCESS }
+function clause execute FENCEI(imm, rs, rd) = {
```
pmundkur:
```suggestion
function clause execute FENCEI(_imm, _rs, _rd) = {
```

pmundkur (changes requested):
Perhaps we could also fix the typo in `extensions/Zifenci` -> `extensions/Zifencei` while you are here?  Otherwise LGTM.

I wonder if uncommenting the `Barrier_RISCV_i` will cause an issue in the litmus tests once we get those running.

Timmmm:
> Perhaps we could also fix the typo in extensions/Zifenci -> extensions/Zifencei while you are here?

I'll do that in another commit. Usually best not to mix up renaming files and editing them or Git gets confused.

> I wonder if uncommenting the Barrier_RISCV_i will cause an issue in the litmus tests once we get those running.

Good question. I think we should probably still uncomment it though and then recomment it if needed later, and actually include a comment explaining why!

model/postlude/insts_end.sail
```diff
@@ -37,6 +37,8 @@ end assembly
 end encdec
 end encdec_compressed
 
-function print_insn(insn : instruction) -> string = assembly(insn)
+// Some legal instructions do not have any corresponding assembly, e.g. reserved fences.
+function instruction_to_str(insn : instruction) -> string =
+    if assembly_forwards_matches(insn) then assembly(insn) else (".insn " ^ bits_str(encdec(insn)))
```
pmundkur: Also remove extra indentation
```suggestion
  if assembly_forwards_matches(insn) then assembly(insn) else (".insn " ^ bits_str(encdec(insn)))
```

pmundkur (approved):

PR #1443, opened 2025-12-20
Fix compiler warnings on macOS and enable more warnings.

pmundkur:
Use `PRId64` instead of `%ld` for portability.

Enable more warning flags for the C++ code, and also warnings-as-errors in CI.

### DIFF
--- a/c_emulator/CMakeLists.txt
+++ b/c_emulator/CMakeLists.txt
@@ -124,7 +124,7 @@ target_compile_options(sail_riscv_sim PRIVATE
   -Wno-parentheses-equality
 )
 
-# Make this use CMAKE_COMPILE_WARNING_AS_ERROR once we upgrade
+# TODO: Make this use CMAKE_COMPILE_WARNING_AS_ERROR once we upgrade
 # the cmake_minimum_version to 3.24 or higher.
 option(WARNINGS_AS_ERRORS "Treat C++ compiler warnings as errors." FALSE)
 


pmundkur:
I enabled `-Wno-unused-parameter` since it seems to build cleanly now.

c_emulator/CMakeLists.txt
```diff
@@ -116,15 +116,16 @@ foreach (config_filename IN LISTS generated_config_filenames)
...
+option(IS_CI_BUILD "Is a build in CI" FALSE)
+
+if(IS_CI_BUILD)
+  target_compile_options(sail_riscv_sim PRIVATE
+    -Wall
+    -Werror
+    -Wextra
+    -Wno-unused-parameter
+  )
+endif()
```
Timmmm: If it's relatively warning-clean now I think maybe we should turn on warnings, and just reserve `-Werror` for CI? And also maybe name it `WARNINGS_AS_ERRORS` rather than `IS_CI_BUILD` since that's what it does. I.e.

```
target_compile_options(sail_riscv_sim PRIVATE
  -Wall
  -Wextra
  -Wno-unused-parameter
)

option(WARNINGS_AS_ERRORS "Upgrade warnings to errors" FALSE)

if (WARNINGS_AS_ERRORS)
  target_compile_options(sail_riscv_sim PRIVATE -Werror)
endif()
```

jordancarlin: We should be able to use the built in CMake property for this instead of defining our own option: `CMAKE_COMPILE_WARNING_AS_ERROR` (https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html#variable:CMAKE_COMPILE_WARNING_AS_ERROR).

c_emulator/CMakeLists.txt
```diff
@@ -116,15 +116,12 @@ foreach (config_filename IN LISTS generated_config_filenames)
...
-# generates too many warnings to turn these on globally.
-
-# target_compile_options(sail_riscv_sim PRIVATE
-#     -Wall -Wextra
-#     # Too annoying at the moment.
-#     -Wno-unused-parameter
-# )
+target_compile_options(sail_riscv_sim PRIVATE
+  -Wall
+  -Werror
```
jordancarlin: `-Werror` shouldn’t be in this list. This is always enabling it.

pmundkur: Oops, missed that.

c_emulator/CMakeLists.txt
Timmmm, line 94: Probably need `-Wno-error` here or something, if that works?

c_emulator/CMakeLists.txt
```diff
@@ -116,12 +116,24 @@ foreach (config_filename IN LISTS generated_config_filenames)
     )
 endforeach()
 
+# Make this use CMAKE_COMPILE_WARNING_AS_ERROR once we upgrade
+# the cmake_minimum_version to 3.24 or higher.
```
Timmmm: This should probably go on line 128? I'd also add add `# TODO: Make this...`

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         run: |
           # Ninja is used because the CMake Makefile generator doesn't
           # build top-level targets in parallel unfortunately.
-          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DFIRST_PARTY_TESTS=TRUE -DENABLE_RISCV_TESTS=TRUE -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_COMPILE_WARNING_AS_ERROR=TRUE -DFIRST_PARTY_TESTS=TRUE -DENABLE_RISCV_TESTS=TRUE -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE
```
Timmmm: Does this need to be updated again?

c_emulator/CMakeLists.txt
```diff
@@ -116,15 +116,23 @@ foreach (config_filename IN LISTS generated_config_filenames)
...
-# )
+target_compile_options(sail_riscv_sim PRIVATE
+  -Wall
+  -Wextra
+  -Wno-unused-parameter
+  # For Sail generated code that uses extra parentheses.
+  -Wno-parentheses-equality
+)
+
+# Make this use CMAKE_COMPILE_WARNING_AS_ERROR once we upgrade
```
Timmmm:
```suggestion
# TODO: Make this use CMAKE_COMPILE_WARNING_AS_ERROR once we upgrade
```

Timmmm (changes requested):

Timmmm (approved):

PR #1445, opened 2025-12-20
Cancel in progress CI jobs when new commits are pushed to the same PR

jordancarlin:


Timmmm:
Hmm I'm surprised that's not the default!

jordancarlin:
> Hmm I'm surprised that's not the default!

Agreed,  but if you look through the actions log, it clearly isn’t.

Timmmm:
Hmm seems like it cancelled itself in the merge queue or something?

jordancarlin:
Hmm. It looks like it ran all of the jobs twice when it was added to the merge queue, and the second instance was cancelled. I’ll look into it. Not sure why the jobs are being run twice, but it looks like that’s actually happening for all PRs when we add them to the merge queue.

PR #1451, opened 2025-12-22
add except context for H extension

KotorinMinami:
Since the trap handling of H extension requires bringing corresponding information to the handler during address translation and other process, this PR is submitted to add the corresponding interface first.


jordancarlin:

PR #1453, opened 2025-12-23
Bump minimum Sail version to 0.20.1

jordancarlin:


jordancarlin:
@Alasdair @bacam It looks like the Rocq install is failing because `Package coq-sail-stdpp has no version 0.20.1`. Should we expect a 0.20.1 release of `coq-sail-stdpp` or should that continue to use 0.20?

Alasdair:
I guess we will probably do a 0.20.1 release, but it might have to wait until after Christmas. In the mean time, I think the current HEAD of the coq-sail repo should work if anyone wants to use the latest release with Rocq.

bacam:
There's a new release now; it's just waiting on the rocq opam repository maintainers (who might be away for Christmas).  We could change the CI to fetch the package directly from github, but usually there's enough time before the model adopts a new version that it's not a problem.

Timmmm (approved):

pmundkur (approved):

PR #1454, opened 2025-12-23
Fix clang-format pre-commit error from #1437

jordancarlin:


Timmmm:
Hmm did this get broken on master somehow? I thought that should be impossible because we use a merge queue. Definitely worth investigating.

jordancarlin:
> Hmm did this get broken on master somehow? I thought that should be impossible because we use a merge queue. Definitely worth investigating.

Agreed. See https://github.com/riscv/sail-riscv/pull/1437#issuecomment-3688114406 for some initial insights.

PR #1459, opened 2025-12-27
Make PC a virtaddr

### LINKED ISSUE #669: Make PC a `virtaddr`
We discussed this a little [here](https://github.com/riscv/sail-riscv/pull/615/files#diff-20961e085e0ef6d48e592b88ce9ce59cce9018a99fd202615c3aabd8729bcf71), but I think `PC` should be a `virtaddr` rather than `xlenbits` like it currently is.

ibvqeibob:
Change the architectural PC register from xlenbits to virtaddr (a newtype) to better reflect its role as a virtual address.
Add mk_virtaddr : xlenbits -> virtaddr and use bits_of_virtaddr at boundaries (bit-level ops, logging, callbacks, CSRs).
In tick_pc(), commit nextPC:xlenbits into PC:virtaddr.
UpdateRVFI fetch paths to avoid indexing/adding a virtaddr directly: translation uses virtaddr.
Trap/exception flow: trap_handler takes virtaddr and stores mepc/sepc as bits.

Fixes #669

### DIFF
--- a/model/core/mem_addrtype.sail
+++ b/model/core/mem_addrtype.sail
@@ -15,6 +15,8 @@ let physaddrbits_len = sizeof(physaddrbits_len)
 newtype physaddr = Physaddr : physaddrbits
 newtype virtaddr = Virtaddr : xlenbits
 
+function mk_virtaddr(x : xlenbits) -> virtaddr = Virtaddr(x)
+
 function bits_of_physaddr(Physaddr(paddr) : physaddr) -> physaddrbits = paddr
 
 function bits_of_virtaddr(Virtaddr(vaddr) : virtaddr) -> xlenbits = vaddr
--- a/model/core/pc_access.sail
+++ b/model/core/pc_access.sail
@@ -15,7 +15,7 @@
   The value in the PC register is the absolute virtual address of the instruction
   to fetch.
  */
-function get_arch_pc() -> xlenbits = PC
+function get_arch_pc() -> xlenbits = bits_of_virtaddr(PC)
 
 function get_next_pc() -> xlenbits = nextPC
 
@@ -26,6 +26,8 @@ function set_next_pc(pc : xlenbits) -> unit = {
 }
 
 function tick_pc() -> unit = {
-  PC = nextPC;
-  pc_write_callback(PC);
+  PC = mk_virtaddr(nextPC);
+  pc_write_callback(bits_of_virtaddr(PC));
 }
+
+function get_pc_bits() -> xlenbits = bits_of_virtaddr(PC)
--- a/model/core/regs.sail
+++ b/model/core/regs.sail
@@ -8,7 +8,7 @@
 
 // program counter
 
-register PC       : xlenbits
+register PC       : virtaddr = mk_virtaddr(zeros())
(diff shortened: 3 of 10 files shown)


jordancarlin:
This seems to contain your changes from #1446. Can you update this PR to to just contain the changes relevant to resolving #669?

ibvqeibob:
> This seems to contain your changes from #1446. Can you update this PR to to just contain the changes relevant to resolving #669?

I rebased pc-virtaddr onto upstream/master and removed the unrelated commit.

model/core/pc_access.sail
```diff
@@ -15,7 +15,7 @@
   The value in the PC register is the absolute virtual address of the instruction
   to fetch.
  */
-function get_arch_pc() -> xlenbits = PC
+function get_arch_pc() -> xlenbits = bits_of_virtaddr(PC)
```
arichardson, line 18: Shouldn't these functions also return virtaddr?

model/core/pc_access.sail
```diff
@@ -26,6 +26,8 @@ function set_next_pc(pc : xlenbits) -> unit = {
 }
 
 function tick_pc() -> unit = {
-  PC = nextPC;
-  pc_write_callback(PC);
+  PC = mk_virtaddr(nextPC);
```
arichardson, line 29: Why not make nextPC a virtaddr as well?

model/core/mem_addrtype.sail
```diff
@@ -15,6 +15,8 @@ let physaddrbits_len = sizeof(physaddrbits_len)
 newtype physaddr = Physaddr : physaddrbits
 newtype virtaddr = Virtaddr : xlenbits
 
+function mk_virtaddr(x : xlenbits) -> virtaddr = Virtaddr(x)
```
jordancarlin, line 18: What is the point of this function vs. just using `Virtaddr(x)` directly?

ibvqeibob: Mainly to centralize virtaddr construction in one place. Today it’s just a thin wrapper over Virtaddr(x), but having mk_virtaddr lets us later add canonicalization/masking/sign-extension rules (or assertions) without touching every call site

model/core/pc_access.sail
ibvqeibob: > Why not make nextPC a virtaddr as well?

I left nextPC as xlenbits intentionally: it behaves like an internal staging register (not architecturally visible) and is heavily used in arithmetic/bit-slicing throughout the model. Keeping it as bits avoids pervasive virtaddr <-> xlenbits conversions and keeps compatibility with existing logic.
The architectural boundary is tick_pc(), where we convert nextPC : xlenbits into PC : virtaddr to enforce the “architectural PC is a virtual address” semantics.

model/core/pc_access.sail
ibvqeibob: PC is now a virtaddr, so Sail code that needs the virtual-address semantics can use PC directly. I kept get_arch_pc() returning xlenbits for compatibility with existing call sites (RVFI/logging/callbacks) that expect a bitvector, and to avoid a much larger refactor.

PR #1460, opened 2025-12-27
Add JSON execution trace output format

Timmmm:
This adds support for outputting an execution trace using [JSON-Lines](https://jsonlines.org/), i.e. newline-separated JSON objects.

This isn't an ideal format, but it is at least universally supported and much better than parsing the text-based output which was never intended to be machine readable.

In future I hope we will add a few other formats - RVVI-TEXT, Protobuf, etc. but I think JSON is a bare minimum.

The schema is ad-hoc but quite simple.

### DIFF
--- a/c_emulator/trace_output_jsonl.cpp
+++ b/c_emulator/trace_output_jsonl.cpp
@@ -9,6 +9,50 @@
 
 // See `doc/jsonl` for the schema and documentation. Please update it if you make changes!
 
+namespace {
+
+// Append an lbits value as a JSON array of bytes to `out`. It must
+// be a multiple of 8 bits.
+void append_lbits_array(std::string &out, lbits value) {
+  assert(value.len % 8 == 0);
+  std::vector<uint8_t> data((value.len + 7) / 8);
+  mpz_export(data.data(), nullptr, -1, 1, 0, 0, *value.bits);
+
+  out += '[';
+
+  bool first = true;
+  for (unsigned byte : data) {
+    if (!first) {
+      out += ',';
+    }
+    first = false;
+    out += std::to_string(byte);
+  }
+
+  out += ']';
+}
+
+// Append an sbits value as a JSON array of bytes to `out`. It must
+// be a multiple of 8 bits.
+void append_sbits_array(std::string &out, sbits value) {
+  assert(value.len % 8 == 0);
+
+  out += '[';
+
+  bool first = true;
+  for (unsigned i = 0; i < value.len; i += 8) {
+    if (!first) {
+      out += ',';
+    }
+    first = false;
+    out += std::to_string((value.bits >> i) & 0xFF);
+  }
+
+  out += ']';
+}
+
+} // namespace
+
 void trace_output_jsonl::open(const std::string &filename) {
   m_ofs.open(filename);
   if (!m_ofs) {
@@ -25,7 +69,10 @@ void trace_output_jsonl::pre_step_callback(hart::Model &model, bool waiting) {
     return;
   }
 
-  m_ofs << "{\"pc\":" << model.zPC.bits;
+  m_ofs << "{\"pc\":";
+  std::string pc_str;
+  append_sbits_array(pc_str, model.zPC);
+  m_ofs << pc_str;
 }
 void trace_output_jsonl::post_step_callback(hart::Model &, bool waiting) {
   if (waiting) {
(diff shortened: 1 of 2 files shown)


Timmmm:
This is a work in progress (I haven't actually run it at all yet, and I'd like to make a proper JSON-Schema for it). Just thought I'd put it out there ASAP.

c_emulator/trace_output_jsonl.cpp
```diff
@@ -0,0 +1,227 @@
...
+  hart::Model &model,
+  const char *type,
+  sbits paddr,
+  uint64_t width,
+  lbits value
+) {
+  if (!m_loads.empty()) {
+    m_loads += ',';
+  }
+  assert(value.len % 8 == 0);
```
arichardson: This is duplicated quite a few times, add a helper to convert to std::vector?

c_emulator/trace_output_jsonl.cpp
```diff
@@ -0,0 +1,227 @@
...
+  std::vector<uint8_t> data((value.len + 7) / 8);
+  mpz_export(data.data(), nullptr, -1, 1, 0, 0, *value.bits);
+
+  m_loads += "{" + std::string("\"paddr\":") + std::to_string(paddr.bits) + "," + std::string("\"width\":") +
+             std::to_string(width) + "," + std::string("\"value\":[");
+  bool comma = false;
+  for (unsigned v : data) {
```
arichardson: This loop is also duplicated multiple times, factor into a helper?

arichardson (commented):
Having something structured is great, no opinion on json vs. other formats but it's simple enough so why not

c_emulator/trace_output_jsonl.cpp
```diff
@@ -0,0 +1,205 @@
...
+  append_lbits_array(m_stores, value);
+  m_stores += "}";
+}
+
+void trace_output_jsonl::mem_read_callback(hart::Model &, const char *, sbits paddr, uint64_t width, lbits value) {
+  if (!m_loads.empty()) {
+    m_loads += ',';
+  }
+  assert(value.len % 8 == 0);
+  std::vector<uint8_t> data((value.len + 7) / 8);
```
arichardson: No longer used?

c_emulator/trace_output_jsonl.cpp
```diff
@@ -0,0 +1,205 @@
...
+// Schema follows this Serde definition. Some notes:
+//
+//  * Tuples are represented as arrays.
+//  * Optional values that are None are simply omitted.
+//  * This uses unsigned 64-bit integers which some JSON libraries may not support.
+//    You need to use one that does. This works fine in Python, Rust (serde_json),
+//    and C++ (Nlohmann/JSON for Modern C++). For Javascript (JSON.parse) you
+//    need to use the `reviver` parameter and parse `context.source` into a BigInt.
+//
+// #[derive(Serialize, Deserialize)]
```
arichardson: Would it make sense to place this comment into a separate .rs file so it can be read more easily?

doc/jsonl/README.md
```diff
@@ -0,0 +1,66 @@
+# JSON-Lines Output Format
+
+This model supports a [JSON-Lines](https://jsonlines.org/) structured logging format that should be used whenever a machine-readable execution log is needed.
+
+**DO NOT PARSE THE HUMAN-READABLE OUTPUT**
```
arichardson, line 5: Maybe this should go into the top-level readme as well?

doc/jsonl/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+The schema matches the following Serde definition. Also note:
+
+- Tuples are represented as arrays.
+- Optional values that are `None` are simply omitted.
```
arichardson: `null` in JSON?

doc/jsonl/README.md
```diff
@@ -0,0 +1,66 @@
...
+    /// Opcode may be missing e.g. for fetch faults.
+    opcode: Option<u32>,
+    /// On exception, this is equal to the cause. Mutually exclusive with interrupt.
+    exception: Option<u64>,
+    /// On interrupt, this is equal to the cause. Mutually exclusive with
+    /// exception. The top 'interrupt' bit of mstatus is not included.
+    interrupt: Option<u64>,
+    /// List of X register writes. May be missing if there are none.
+    #[serde(default)]
+    x: Vec<(u8, u64)>,
```
arichardson: Would it make sense to always use a `Vec<u8>` for the value of all register types? That way it will continue to work once CHERI support is added and I guess it would also be helpful for fp128

arichardson (approved):
Looking good just a few minor comments

doc/jsonl/README.md
Timmmm: Yeah maybe but I expect it's a fair bit of overhead and not usually what you want as a user of the logs.

For CHERI we also need tags too so I was thinking of changing the key from `x` to `c`, which means the format can be changed. And an extra tag field can be added to `MemAccess`.

On the other hand it does make the length nice and explicit, so I'm not totally tied to `u64`. And it makes it easier to use in JavaScript. And it's more uniform. I might change it.

arichardson: I guess the other option would be to use arbitrary length hex strings like we do in the sail config file?

Timmmm: Yeah... that feels even more hacky though.

I think I'll change it to all `Vec<u8>`.

doc/jsonl/README.md
Timmmm: Yeah I'm not sure where to put it exactly though. I think probably what we should do is:

1. Release this, and ask people to use it.
2. After some time, break the human-readable trace output by adding a message like "Please do not parse" to it, but include a `--temporarily-allow-parsing-human-text-output` flag to allow people to migrate.
3. After some more time, remove that flag, and then maybe change the format entirely (I think we can make it quite a lot more readable).

PR #1463, opened 2025-12-31
Update documentation and code style guidelines.

### LINKED ISSUE #1464: Partially unreadable ReadingGuide diagram in dark mode
This diagram in the reading guide should handle dark mode better:

Image

I could draw a new diagram as part of #1463 unless there's a simple fix.

pmundkur:
Update code style to
- clarify that line comments should always be preferred to block comments 
- mention C++ as a target for formatting rules

Update the emulator structure diagram in the reading guide to reflect the new C++ dependencies, and illustrate a common workflow.

Update the changelog to mention the switch to the C++ model of the hart.

Replace the class structure diagram to fix #1464.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# RISCV Sail Model
+# Formal Specification of the RISC-V ISA
 
 This repository contains a formal specification of the RISC-V architecture, written in
 [Sail](https://github.com/rems-project/sail). It has been adopted by RISC-V International.


doc/ReadingGuide.md
```diff
@@ -1,4 +1,32 @@
...
+a default RV64 configuration is assumed. The Sail compiler also
+generates a JSON schema (`sail_riscv_config_schema.json`) for the
+configuration file from the Sail sources; every configuration file is
+validated against this schema before use. More information on using
+the emulator is available using its `-h` help command-line option.
+
+The Sail model can also be used to generate `JSON` and `HTML`
+artifacts for documentation. A prototype of their use to annotate the
+unprivileged volume of the RISC-V specification is available
+[here](https://github.com/Timmmm/riscv-isa-manual).
```
pmundkur: @Timmmm I linked to your github repo of the manual instead of the website, to save bandwidth from the pre-commit checks in CI.

Timmmm: The website is hosted on Github Pages anyway so I don't think that will make any difference... But I think this is fine anyway.

doc/ReadingGuide.md
```diff
@@ -1,4 +1,32 @@
-# A guide to reading the Sail RISC-V specification
+# Using the Sail RISC-V specification
```
jordancarlin: This new section is helpful documentation, but I'm not sure that it belongs in the Reading Guide. It seems better suited for the README (which I think could do with an overhaul to make it more user-focused instead of developer-focused—right now it doesn't even mention that we release binaries of the model that we should be pushing most users towards).

pmundkur: Agreed, I had the same thought yesterday that this should go into the README.  I'll do that later today, and look into updating the README as well.

pmundkur: Moved this to the README.

PR #1465, opened 2026-01-05
Fix font colours in class structure diagram

### LINKED ISSUE #1464: Partially unreadable ReadingGuide diagram in dark mode

Timmmm:
The default colour for text is black in light mode and white in dark mode, but we want always black for the text in boxes.

Fixes #1464


Timmmm:
@pmundkur I didn't see you had already remade the diagram. I think yours looks a little nicer; feel free to close this if you like.

pmundkur:
The handling of background fill is nicer in this diagram (black text and arrows turn to white on the black background in dark mode); typst's SVG doesn't do this for a transparent fill and just uses a white background in dark mode.  I just find typst/Tikz source easier to edit, so perhaps we could use that version?

PR #1467, opened 2026-01-06
Create weekly binary release

jordancarlin:
Create weekly pre-release Mondays at 02:35 UTC to make it easy to use up-to-date binaries of the model.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,8 +19,8 @@ on:
         required: false
         type: boolean
   schedule:
-    # Weekly on Mondays at 00:00 UTC
-    - cron: "0 0 * * 1"
+    # Weekly on Mondays at 02:35 UTC
+    - cron: "35 2 * * 1"
 
 permissions:
   id-token: write


.github/workflows/release.yml
```diff
@@ -18,6 +18,9 @@ on:
         default: true
         required: false
         type: boolean
+  schedule:
+    # Weekly on Mondays at 00:00 UTC
+    - cron: "0 0 * * 1"
```
Timmmm: We should probably not do this on the hour - see [here](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule).

> The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. If the load is sufficiently high enough, some queued jobs may be dropped. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.

I'd change it to something more random like 02:35.

jordancarlin: Updated

.github/workflows/release.yml
```diff
@@ -18,6 +18,9 @@ on:
         default: true
         required: false
         type: boolean
+  schedule:
+    # Weekly on Mondays at 02:35 UTC
```
pmundkur, line 22: Perhaps add a comment here explaining this choice with the link that @Timmmm pointed out?

jordancarlin: #1471

PR #1474, opened 2026-01-07
Use a consistent name for variables of MemoryAccessType.

pmundkur:
This makes the code a bit more readable.


pmundkur:
Pulled out from #1405 to reduce its diff.

Timmmm (approved):
Nice

PR #1475, opened 2026-01-07
Centralize memory-access-to-fault conversions.

pmundkur:
There were multiple conversions scattered around the model. Some used wild-cards, making it easy to miss correct handling when adding a new memory access type, or causing duplicated effort when changing memory access handling.


pmundkur:
Again motivated by reducing the diff for #1405.

jordancarlin (approved):
LGTM

nadime15 (approved):

PR #1476, opened 2026-01-11
plic mvp

ibvqeibob:
This PR adds MVP PLIC support to the Sail RISC-V platform model and includes a bare-metal test to validate the basic external interrupt flow. The current implementation covers the end-to-end path (trigger → claim/complete → HTIF exit).

Follow-ups will focus on broader coverage (more interrupt sources and scenarios) and improved OS compatibility over time.

### DIFF
--- a/test/first_party/CMakeLists.txt
+++ b/test/first_party/CMakeLists.txt
@@ -156,3 +156,45 @@ foreach (xlen IN ITEMS 32 64)
         )
     endforeach()
 endforeach()
+
+# ------------------------------------------------------------
+# PLIC MVP smoke test (bare-metal, RV64 only)
+# Sources live in test/plic/
+# ------------------------------------------------------------
+
+set(plic_dir "${CMAKE_CURRENT_SOURCE_DIR}/../plic")
+set(plic_elf "rv64d_plic_basic.elf")
+set(plic_config "${CMAKE_BINARY_DIR}/config/rv64d_v256_e64.json")
+
+add_custom_command(
+  OUTPUT ${plic_elf}
+  DEPENDS
+    "${plic_dir}/plic_basic.c"
+    "${plic_dir}/crt0.S"
+    "${plic_dir}/link.ld"
+  COMMAND ${CROSS_COMPILER_COMMAND_RV64}
+    -march=rv64imac
+    -mabi=lp64
+    -mno-relax
+    -ffreestanding
+    -nostdlib
+    -static
+    -mcmodel=medany
+    -g
+    -Og
+    -Wall
+    -Werror
+    -T "${plic_dir}/link.ld"
+    -o ${plic_elf}
+    "${plic_dir}/plic_basic.c"
+    "${plic_dir}/crt0.S"
+  VERBATIM
+  COMMENT "Compiling PLIC MVP smoke test"
+)
+
+add_custom_target(build_rv64d_plic_basic ALL DEPENDS ${plic_elf} ${plic_config})
+
+add_test(
+  NAME "first_party_rv64d_plic_basic"
+  COMMAND $<TARGET_FILE:sail_riscv_sim> --config ${plic_config} ${plic_elf}
+)
--- a/test/plic/crt0.S
+++ b/test/plic/crt0.S
@@ -1,0 +1,7 @@
+.section .text.start
+.global _start
+_start:
+  la  t0, tohost
+  li  t1, 1
+  sd  t1, 0(t0)
+1: j 1b
--- a/test/plic/link.ld
+++ b/test/plic/link.ld
@@ -1,0 +1,44 @@
+OUTPUT_ARCH(riscv)
+ENTRY(_start)
+
+MEMORY
+{
+  RAM (rwx) : ORIGIN = 0x80000000, LENGTH = 16M
+}
+
+SECTIONS
+{
+  .text : {
+    *(.text.start*)
+    *(.text*)
+  } > RAM
+
+  .rodata : {
+    *(.rodata*)
+    *(.srodata*)
+  } > RAM
+
+  .data : {
+    *(.data*)
+    *(.sdata*)
+  } > RAM
+
+  .bss : {
+    __bss_start = .;
+    *(.bss*)
+    *(.sbss*)
+    *(COMMON)
+    __bss_end = .;
+  } > RAM
+
+  /* HTIF tohost (Sail uses this to detect end of test) */
+  .tohost : ALIGN(8) {
+    PROVIDE(tohost = .);
+    QUAD(0);
+    PROVIDE(fromhost = .);
+    QUAD(0);
+  } > RAM
+
+  /* simple stack at top of RAM */
+  PROVIDE(_stack_top = ORIGIN(RAM) + LENGTH(RAM));
+}
--- a/test/plic/plic_basic.c
+++ b/test/plic/plic_basic.c
@@ -1,0 +1,108 @@
+typedef unsigned int  uint32_t;
+typedef unsigned long uint64_t;
+
+extern volatile uint64_t tohost;
+
+/* CSR */
+static inline uint64_t csrr_mip(void) {
+  uint64_t x;
+  asm volatile("csrr %0, mip" : "=r"(x));
+  return x;
+}
+
+/* mip.MEIP */
+#define MIP_MEIP (1ull << 11)
+
+/* MMIO */
+static inline void write32(uint64_t addr, uint32_t val) {
+  *(volatile uint32_t *)addr = val;
+}
+static inline uint32_t read32(uint64_t addr) {
+  return *(volatile uint32_t *)addr;
+}
+
+/* PLIC MMIO layout (matches Sail MVP) */
+#define PLIC_BASE 0x0c000000UL
+
+#define OFF_PRIORITY   0x000000UL  /* 4*id */
+#define OFF_PENDING    0x001000UL  /* pending injection (non-standard) */
+#define OFF_ENABLE     0x002000UL  /* ctx0=M, ctx1=S (stride 0x80) */
+#define ENABLE_STRIDE  0x80UL
+
+#define OFF_CTX        0x200000UL  /* ctx page */
+#define CTX_STRIDE     0x1000UL
+#define OFF_THRESHOLD  0x0UL
+#define OFF_CLAIM      0x4UL
+
+static inline uint64_t PLIC_PRIORITY(int id) {
+  return PLIC_BASE + OFF_PRIORITY + (uint64_t)id * 4;
+}
+static inline uint64_t PLIC_PENDING_WORD(int id) {
+  return PLIC_BASE + OFF_PENDING + (uint64_t)(id / 32) * 4;
+}
+static inline uint64_t PLIC_ENABLE_WORD(int ctx, int id) {
+  return PLIC_BASE + OFF_ENABLE + (uint64_t)ctx * ENABLE_STRIDE + (uint64_t)(id / 32) * 4;
+}
+static inline uint64_t PLIC_THRESHOLD(int ctx) {
+  return PLIC_BASE + OFF_CTX + (uint64_t)ctx * CTX_STRIDE + OFF_THRESHOLD;
+}
+static inline uint64_t PLIC_CLAIM(int ctx) {
+  return PLIC_BASE + OFF_CTX + (uint64_t)ctx * CTX_STRIDE + OFF_CLAIM;
+}
+
+/* HTIF: 0 = success, non-zero = failure */
+__attribute__((noreturn)) static void fail(void) {
+  tohost = 1;
+  while (1) {}
+}
+__attribute__((noreturn)) static void pass(void) {
+  tohost = 0;
+  while (1) {}
+}
+
+static void wait_meip(int want_set) {
+  for (int i = 0; i < 2000000; i++) {
+    int is_set = (csrr_mip() & MIP_MEIP) ? 1 : 0;
+    if (is_set == want_set) return;
+  }
+  fail();
+}
+
+int main(void) {
+  const int ctx = 0;   /* ctx0 = hart0/M */
+  const int src = 1;   /* IRQ ID */
+
+  /* Configure: priority, threshold, enable */
+  write32(PLIC_PRIORITY(src), 1);
+  if (read32(PLIC_PRIORITY(src)) != 1) fail();
+
+  write32(PLIC_THRESHOLD(ctx), 0);
+
+  uint64_t en_addr = PLIC_ENABLE_WORD(ctx, src);
+  uint32_t en = read32(en_addr);
+  en |= (1u << (src % 32));
+  write32(en_addr, en);
+
+  /* Inject pending (bring-up helper in Sail MVP) */
+  write32(PLIC_PENDING_WORD(src), (1u << (src % 32)));
+
+  /* External interrupt should assert */
+  wait_meip(1);
+
+  /* claim/complete */
+  uint32_t claim = read32(PLIC_CLAIM(ctx));
+  if (claim != (uint32_t)src) fail();
+  write32(PLIC_CLAIM(ctx), claim);
+
+  /* External interrupt should deassert */
+  wait_meip(0);
+
+  /* Gating: threshold > priority blocks */
+  write32(PLIC_THRESHOLD(ctx), 2);
+  write32(PLIC_PENDING_WORD(src), (1u << (src % 32)));
+  for (int i = 0; i < 500000; i++) {
+    if (csrr_mip() & MIP_MEIP) fail();
+  }
+
+  pass();
+}


model/sys/platform.sail
```diff
@@ -69,6 +78,68 @@ let MTIMECMP_BASE_HI : physaddrbits = zero_extend(0x04004)
...
+let PLIC_OFF_PENDING_END  : int = unsigned(0x2000)
+
+let PLIC_OFF_ENABLE_BASE  : int = unsigned(0x2000)
+let PLIC_OFF_ENABLE_END   : int = unsigned(0x3000)
+let PLIC_ENABLE_STRIDE    : int = unsigned(0x80)    // per-context enable stride
+
+let PLIC_OFF_CTX_BASE     : int = unsigned(0x200000)
+let PLIC_CTX_STRIDE       : int = unsigned(0x1000)  // per-context page
+let PLIC_OFF_THRESHOLD    : int = 0
+let PLIC_OFF_CLAIM        : int = 4
```
Timmmm, line 117: Remove all of these `: int` annotations. Sail will infer more precise types (e.g. `let PLIC_OFF_CLAIM : int(4) = 4`).

plic_test/plic_test.map
```diff
@@ -0,0 +1,83 @@
...
+.riscv.attributes
+                0x0000000000000000       0x2f
+ .riscv.attributes
+                0x0000000000000000       0x2d /tmp/cc4CzCYt.o
+ .riscv.attributes
+                0x000000000000002d       0x2f /tmp/ccS5KxDq.o
+
+.comment        0x0000000000000000        0xf
+ .comment       0x0000000000000000        0xf /tmp/ccS5KxDq.o
+                                         0x10 (size before relaxing)
```
Timmmm: I'm guessing you didn't mean to commit this file.

model/sys/platform.sail
```diff
@@ -31,6 +31,15 @@ private function within_clint forall 'n, 0 < 'n <= max_mem_access . (Physaddr(ad
   & (addr_int + sizeof('n)) <= (clint_base_int + clint_size_int)
 }
 
+private function within_plic forall 'n, 0 < 'n <= max_mem_access .
+  (Physaddr(addr) : physaddr, width : int('n)) -> bool = {
```
Timmmm, line 35: Very minor but can you format this the same as `within_clint`. We'll probably use an auto-formatter eventually but we should be consistent where possible.

model/sys/platform.sail
```diff
@@ -69,6 +78,68 @@ let MTIMECMP_BASE_HI : physaddrbits = zero_extend(0x04004)
 let MTIME_BASE       : physaddrbits = zero_extend(0x0bff8)
 let MTIME_BASE_HI    : physaddrbits = zero_extend(0x0bffc)
 
+// Type-level sizes (can be used in bits()/vector())
+type plic_nsrc = 64
+type plic_nctx = 2
+
+// Term-level constants (can be used in loops/arith)
+let PLIC_NSRC : int = 64 // IDs 0..63, ID0 reserved
+let PLIC_NCTX : int = 2  // ctx0 = hart0/M, ctx1 = hart0/S
```
Timmmm: You can use the same name for both; e.g. look how `xlen` is defined.

```
type plic_nsrc = 64
let plic_nsrc = sizeof(plic_nsrc)
```

model/sys/platform.sail
```diff
@@ -69,6 +78,68 @@ let MTIMECMP_BASE_HI : physaddrbits = zero_extend(0x04004)
...
+
+register plic_priority : list(bits(32)) = [||]
```
Timmmm: Use a `vector` instead of a `list`. Lists are awkward for some backends and we try to avoid them wherever possible. In this case I believe the length is constant? Something like this:

```
register plic_priority : vector(plic_nctx, bits(32)) = vector_init(zeros())
```

model/sys/platform.sail
```diff
@@ -121,7 +192,274 @@ function clint_load(access, Physaddr(addr), width) = {
...
+// PLIC helpers: avoid b[i] style indexing (which causes quantifier issues)
+// -------------------------
+private function plic_threshold_of_ctx(ctx:int) -> bits(32) = {
+  if ctx == 0 then plic_threshold_m else plic_threshold_s
+}
+
+private function plic_mask32(idx:int) -> bits(32) = {
+  if idx < 0 | idx >= 32 then zeros()
+  else (zero_extend(32, 0b1) << idx)
+}
```
Timmmm, line 212: Generally using `int` is a sign that the code could be improved by tighter types. With a few exceptions we shouldn't have `int` anywhere.

For example in this case you probably want something like:

```
private function plic_mask32(idx : range(0, 31)) -> bits(32) =
  zero_extend(0b1 @ zeros(idx))
```

Although actually I don't think you need this function at all. Generally in Sail you don't do bitshifts and masking like in C; you use direct bit access and bitvector concatenation instead.

model/sys/platform.sail
```diff
@@ -121,7 +192,274 @@ function clint_load(access, Physaddr(addr), width) = {
...
+
+private function plic_bit_get64(b: bits(plic_nsrc), idx:int) -> bool = {
+  if idx < 0 then false
+  else if idx < 32 then {
+    let lo : bits(32) = b[31..0];
+    (lo & plic_mask32(idx)) != zeros()
```
Timmmm, line 218: You can just access the bit directly like `b[idx] != bitzero`.

Timmmm (changes requested):
Thanks for the contribution! This has been on my todo list for a while.

If you don't mind I can go through the code and fix all the types etc. Then I'll review the actual logic. Let me know so we don't both end up doing it!

plic_test/start.S
```diff
@@ -0,0 +1,59 @@
...
+
+1:
+    andi t0, t0, 0xff        # low bits = cause code
+    li   t1, 11              # Machine external interrupt
+    bne  t0, t1, 9f
+
+    # PLIC claim/complete for ctx0(M): PLIC_BASE + 0x200004
+    li   t2, 0x0c000000
+    li   t3, 0x200004
+    add  t2, t2, t3
```
nadime15: I think it would be better to introduce a header file and move these definitions into it, so they can be reused for the C and assembly files. This would remove the magic numbers and would make it easier to maintain.

```
#ifndef PLIC_DEFS_H
#define PLIC_DEFS_H

#define PLIC_BASE           0x0c000000
#define PLIC_CLAIM_CTX0     (PLIC_BASE + 0x200004)
#define IRQ_ID              10
#define MCAUSE_MEI          11

#endif
```

plic_test/start.S
```diff
@@ -0,0 +1,59 @@
...
+    sd   t1, 8(sp)
+    sd   t2, 16(sp)
+    sd   t3, 24(sp)
+    sd   t4, 32(sp)
+    sd   t5, 40(sp)
+    sd   t6, 48(sp)
+
+    csrr t0, mcause
+    bltz t0, 1f              # interrupt => mcause is negative
+    j    9f
```
nadime15:
```suggestion
halt:  j    halt

    .p2align 2
    .globl trap_entry
trap_entry:
    addi sp, sp, -64
    sd   ra, 56(sp)
    sd   t0, 0(sp)
    sd   t1, 8(sp)
    sd   t2, 16(sp)
    sd   t3, 24(sp)
    sd   t4, 32(sp)
    sd   t5, 40(sp)
    sd   t6, 48(sp)

    csrr t0, mcause
    bltz t0, 1f              # interrupt => mcause is negative
    j    9f
```

Maybe its just me, but I am not a big fan of reusing labels and use the same labels for backward and forward jumps. I think it would be clearer to give these labels explicit names (I modified one label as an example).

model/sys/platform.sail
```diff
@@ -69,6 +78,68 @@ let MTIMECMP_BASE_HI : physaddrbits = zero_extend(0x04004)
 let MTIME_BASE       : physaddrbits = zero_extend(0x0bff8)
 let MTIME_BASE_HI    : physaddrbits = zero_extend(0x0bffc)
 
+// Type-level sizes (can be used in bits()/vector())
+type plic_nsrc = 64
```
jordancarlin: This should be configurable.

plic_test/start.S
Timmmm: Tbh I think it would be easier just to rewrite the test in C. I don't think it really benefits from being in assembly.

ibvqeibob:
> Thanks for the contribution! This has been on my todo list for a while.感谢你的分享！这件事已经在我的待办事项清单上放了一段时间。
> 
> If you don't mind I can go through the code and fix all the types etc. Then I'll review the actual logic. Let me know so we don't both end up doing it!如果你不介意，我可以帮你整理代码，修正所有类型等。然后我会复习实际的逻辑。告诉我，这样我们俩就不会都去做了！

Thanks for the review and suggestions — sorry for the slow response, I was quite busy recently and only got time to pick this back up.
I’m fixing the type/style issues first (including list→vector). Quick question: for “This should be configurable”, do you prefer (1) build-time config via platform_config.sail, or (2) DTB-driven runtime config (derive nsrc/nctx from riscv,ndev / interrupts-extended, clamped to a compiled max)?
Also, would you rather I keep this PR focused on the MVP + cleanup, and send a follow-up PR for OS-ready PLIC (more sources/contexts/harts + gateway/spec behavior), or do it all in one PR?
My end goal is a full OS-usable PLIC. Thanks again!

jordancarlin:
> Quick question: for “This should be configurable”, do you prefer (1) build-time config via platform_config.sail, or (2) DTB-driven runtime config (derive nsrc/nctx from riscv,ndev / interrupts-extended, clamped to a compiled max)?

It should use the runtime config functionality built into Sail using the JSON file. The same way you are currently setting the base address of the PLIC. 

> Also, would you rather I keep this PR focused on the MVP + cleanup, and send a follow-up PR for OS-ready PLIC (more sources/contexts/harts + gateway/spec behavior), or do it all in one PR? My end goal is a full OS-usable PLIC. Thanks again!

I think it’s fine to have an MVP version first and then augment it later, but as long as the PR doesn’t get too large, either is ok.

ibvqeibob:
> > Quick question: for “This should be configurable”, do you prefer (1) build-time config via platform_config.sail, or (2) DTB-driven runtime config (derive nsrc/nctx from riscv,ndev / interrupts-extended, clamped to a compiled max)?
> 
> It should use the runtime config functionality built into Sail using the JSON file. The same way you are currently setting the base address of the PLIC.
> 
> > Also, would you rather I keep this PR focused on the MVP + cleanup, and send a follow-up PR for OS-ready PLIC (more sources/contexts/harts + gateway/spec behavior), or do it all in one PR? My end goal is a full OS-usable PLIC. Thanks again!
> 
> I think it’s fine to have an MVP version first and then augment it later, but as long as the PR doesn’t get too large, either is ok.

Thanks! Got it — I’ll switch to using Sail’s runtime JSON config for nsrc/nctx (same mechanism as the base/size), and keep this PR focused on MVP + cleanup (incl. list→vector). Then I’ll follow up with a second PR for the OS-ready PLIC extensions. Appreciate the guidance!

ibvqeibob:
> > Quick question: for “This should be configurable”, do you prefer (1) build-time config via platform_config.sail, or (2) DTB-driven runtime config (derive nsrc/nctx from riscv,ndev / interrupts-extended, clamped to a compiled max)?
> 
> It should use the runtime config functionality built into Sail using the JSON file. The same way you are currently setting the base address of the PLIC.
> 
> > Also, would you rather I keep this PR focused on the MVP + cleanup, and send a follow-up PR for OS-ready PLIC (more sources/contexts/harts + gateway/spec behavior), or do it all in one PR? My end goal is a full OS-usable PLIC. Thanks again!
> 
> I think it’s fine to have an MVP version first and then augment it later, but as long as the PR doesn’t get too large, either is ok.

I implemented PLIC configurability via Sail runtime JSON config : added plic.nsrc / plic.nctx fields, read them in init_platform() and clamp to the compile-time maxima to preserve Sail’s type-level constraints. I also replaced plic_priority from a list to vector(plic_nsrc, bits(32)), removed the list helpers, and switched priority read/write to vector access/update. This PR remains MVP (MMIO + arbitration + claim/complete); OS-level extensions (multi-hart/contexts, gateway, etc.) will follow in a separate PR.

ibvqeibob:
> > Quick question: for “This should be configurable”, do you prefer (1) build-time config via platform_config.sail, or (2) DTB-driven runtime config (derive nsrc/nctx from riscv,ndev / interrupts-extended, clamped to a compiled max)?
> 
> It should use the runtime config functionality built into Sail using the JSON file. The same way you are currently setting the base address of the PLIC.
> 
> > Also, would you rather I keep this PR focused on the MVP + cleanup, and send a follow-up PR for OS-ready PLIC (more sources/contexts/harts + gateway/spec behavior), or do it all in one PR? My end goal is a full OS-usable PLIC. Thanks again!
> 
> I think it’s fine to have an MVP version first and then augment it later, but as long as the PR doesn’t get too large, either is ok.

Locally I can build successfully and pass the basic CTest targets (unit_tests + smt_properties_rv64d).
However, CI with the ci.yml configuration (enabling FIRST_PARTY_TESTS, riscv-tests, and riscv-vector-tests) still fails: a large set of riscv-vector-tests (rv32 v128 e32) cases end with subprocess aborted / ***Exception: SegFault. This looks like the sail_riscv_sim process crashing rather than a normal test assertion failure.

My recent changes are mainly:

PLIC arbitration rewritten for Lean-friendly termination (fuel/structural recursion), plus an early stop idx >= plic_nsrc_active to avoid out-of-range accesses;

In config/config.json.in, I enlarged the MMIO memory.regions size from 0x0200_0000 to 0x1000_0000 so that platform.plic.base=0x1000_0000 is inside a valid MMIO region.

Questions:

Are these bulk vector-test SegFault/aborted failures a known issue (especially on macOS / certain CMake+clang combos)? Any recommended way to reproduce/debug ?

What is the expected/recommended mapping for platform.plic.base/size vs memory.regions? Is expanding MMIO to cover PLIC acceptable, or should the region layout be split differently?

c_emulator/config_utils.cpp
```diff
@@ -108,3 +110,21 @@ void validate_config_schema(const std::string &conf_file) {
...
+uint64_t config_get_u64_default(const std::vector<const char *> &keypath, uint64_t def) {
+  sail_config_json j = sail_config_get(keypath.size(), keypath.data());
+  if (j == nullptr) {
+    return def;
+  }
+  if (!sail_config_is_int(j)) {
+    return def;
+  }
+
+  sail_int n;
```
trdthg: This might be the reason why CI fails; `sail_int` is an `mpz_t` type, so it should be initialized like this:

```c
CREATE(sail_int)(&n);
```

On my machine
- If you run `sail_riscv_sim test.elf` directly, the test fails randomly, and `--trace-all` catches nothing.
- If you use `gdb sail_riscv_sim test.elf`, the test never fails. 🤷‍♂️

Finally, I chose to checked the coredump information, from which I traced the issue back to here.

```txt
[New LWP 435042]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `./build/c_emulator/sail_riscv_sim --config /home/trdthg/repo/sail/worktree/plic'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  musable (mem=0x5583a20ce850 <hart::zghz34514>) at ./malloc/malloc.c:5240

warning: 5240   ./malloc/malloc.c: No such file or directory
#0  musable (mem=0x5583a20ce850 <hart::zghz34514>) at ./malloc/malloc.c:5240
#1  __GI___libc_realloc (oldmem=0x5583a20ce850 <hart::zghz34514>, bytes=16) at ./malloc/malloc.c:3440
#2  0x00005583a1fb9160 in __gmp_default_reallocate ()
#3  0x00005583a1f99a72 in __gmpz_realloc ()
#4  0x00005583a1f9a00f in __gmpz_set_str ()
#5  0x00005583a1f91c56 in sail_config_unwrap_int (n=<optimized out>, config=<optimized out>) at /home/trdthg/.opam/default/share/sail/lib/sail_config.c:291
#6  0x00005583a1b37adb in config_get_u64_default (keypath=std::vector of length 3, capacity 3 = {...}, def=def@entry=64) at /home/trdthg/repo/sail/worktree/plic/c_emulator/config_utils.cpp:124
#7  0x00005583a1b3438e in ModelImpl::get_config_plic_nsrc (this=<optimized out>) at /home/trdthg/repo/sail/worktree/plic/c_emulator/riscv_model_impl.cpp:246
#8  0x00005583a1c4b514 in hart::Model::zinit_platform (this=this@entry=0x5583a207e8e0 <g_model>, z3zE3962=z3zE3962@entry=0) at /home/trdthg/repo/sail/worktree/plic/build/sail_riscv_model.cpp:448839
#9  0x00005583a1f77fd5 in hart::Model::zinit_model (this=this@entry=0x5583a207e8e0 <g_model>, zconfig_filename=zconfig_filename@entry=0x5583a8ecd620 "/home/trdthg/repo/sail/worktree/plic/build/config/rv32d_v256_e32.json") at /home/trdthg/repo/sail/worktree/plic/build/sail_riscv_model.cpp:939778
#10 0x00005583a1aec311 in init_sail (elf_entry=elf_entry@entry=2147483648, config_file=<optimized out>) at /home/trdthg/repo/sail/worktree/plic/c_emulator/riscv_sim.cpp:316
#11 0x00005583a1af069f in inner_main (argc=5, argv=0x7fff3ba81b58) at /usr/include/c++/14/bits/basic_string.h:227
#12 0x00005583a1ae97d0 in main (argc=<optimized out>, argv=<optimized out>) at /home/trdthg/repo/sail/worktree/plic/c_emulator/riscv_sim.cpp:680
```

c_emulator/riscv_model_impl.cpp
```diff
@@ -239,6 +240,17 @@ bool ModelImpl::get_config_print_instr(unit) {
...
+uint64_t ModelImpl::get_config_plic_nsrc(unit) {
+  std::vector<const char *> keypath = {"platform", "plic", "nsrc"};
+  return config_get_u64_default(keypath, 64);
+}
+
+uint64_t ModelImpl::get_config_plic_nctx(unit) {
+  std::vector<const char *> keypath = {"platform", "plic", "nctx"};
+  return config_get_u64_default(keypath, 2);
+}
+
```
trdthg: The default value is already defined in `json.config.in`, so `config_get_u64_default` appears to be redundant. Could we follow the same pattern as the clint block above?
Maybe you can just read config from sail and return `zplat_plic_base.bits` here?

c_emulator/riscv_platform_if.cpp
```diff
@@ -164,6 +164,14 @@ bool PlatformInterface::get_config_print_clint(unit) {
...
 }
 
+uint64_t PlatformInterface::get_config_plic_nsrc(unit) {
+  return 64;
+}
+
+uint64_t PlatformInterface::get_config_plic_nctx(unit) {
+  return 2;
+}
+
```
trdthg: just return `zplat_plic_xxx` too?

dependencies/elfio/elfio/elfio_modinfo.hpp
```diff
@@ -59,7 +59,7 @@ template <class S> class modinfo_section_accessor_template
     bool get_attribute( const std::string_view& field_name,
                         std::string&            value ) const
     {
-        for ( const auto [first, second] : content ) {
+        for ( const auto& [first, second] : content ) {
```
trdthg, line 62: is this necessary?

model/sys/platform.sail
```diff
@@ -69,6 +78,71 @@ let MTIMECMP_BASE_HI : physaddrbits = zero_extend(0x04004)
...
+
+// Term-level maxima derived from the type-level sizes
+let PLIC_NSRC_MAX = sizeof(plic_nsrc)
+let PLIC_NCTX_MAX = sizeof(plic_nctx)
+
+// Runtime active (from JSON, clamped to MAX)
+register plic_nsrc_active : int = PLIC_NSRC_MAX
+register plic_nctx_active : int = PLIC_NCTX_MAX
+
+private function clamp_int(x:int, lo:int, hi:int) -> int =
```
trdthg, line 93: We'd better add spaces around the colons for all parameters to keep the style.

model/sys/platform.sail
```diff
@@ -125,6 +199,287 @@ function clint_load(access, Physaddr(addr), width) = {
...
+            ((pr == bpr) & ((bid == 0) | (idx < bid)));
+          if better then {
+            bid = idx;
+            bpr = pr;
+          } else {
+            ()
+          }
+        } else {
+          ()
+        };
```
trdthg, line 289:
```suggestion
        if cand then {
          let better =
            (pr > bpr) |
            ((pr == bpr) & ((bid == 0) | (idx < bid)));
          if better then {
            bid = idx;
            bpr = pr;
          };
        };
```
I think sail should support single if statement.

model/sys/platform.sail
```diff
@@ -125,6 +199,287 @@ function clint_load(access, Physaddr(addr), width) = {
...
+
+        idx = idx + 1;
+        k = k - 1;
+      }
+    } else {
+      ()
+    }
```
trdthg, line 296: same as above

model/sys/platform.sail
```diff
@@ -125,6 +199,287 @@ function clint_load(access, Physaddr(addr), width) = {
...
+      plic_threshold_m = wdata;
+      plic_update_mip();
+      Ok(true)
+    } else if off == (PLIC_OFF_CTX_BASE + 4) then {
+      let id = unsigned(wdata);
+      if id < plic_nsrc_active & id != 0 then {
+        plic_claimed = plic_bit_clear64(plic_claimed, id);
+        plic_claimed = plic_bit_clear64(plic_claimed, 0);
+        plic_update_mip()
+      } else ();
```
trdthg, line 463: same as above

model/sys/platform.sail
```diff
@@ -359,6 +720,26 @@ function init_platform() -> unit = {
   htif_exit_code = zeros();
   htif_cmd_write = 0b0;
   htif_payload_writes = zeros();
+
+  // --- PLIC runtime config (JSON) ---
+  // read config values (C++ backend reads from JSON; other backends use defaults)
+  let nsrc_cfg = unsigned(get_config_plic_nsrc());
+  let nctx_cfg = unsigned(get_config_plic_nctx());
```
trdthg: > // read config values (C++ backend reads from JSON; other backends use defaults)

Maybe reading config from json and returning `zplat_plic_base` directly in `config_get_u64_default` could also resolve this? 
- from `sail < c++ < json`
- to `c++/other_backend < sail < json`

model/sys/platform.sail
```diff
@@ -125,6 +199,287 @@ function clint_load(access, Physaddr(addr), width) = {
...
+      } else {
+        let en = plic_enable_of_ctx(ctx);
+        let th = unsigned(plic_threshold_of_ctx(ctx));
+        let pr = unsigned(plic_priority_get(idx));
+
+        let cand =
+          plic_bit_get64(en, idx) &
+          plic_bit_get64(plic_pending, idx) &
+          (not(plic_bit_get64(plic_claimed, idx))) &
+          (pr > th);
```
trdthg: Maybe moving the operators to the start of the lines is better? It could just be my personal preference.

ibvqeibob:
PLIC (and other platform params) are now consolidated as a single source of truth in Sail, read directly from the JSON runtime config; the C++ side no longer parses JSON for platform setup. Could you please take a moment to review my PR? I’d like to build on this and extend toward a more complete, OS-ready PLIC. Thanks! @Timmmm @jordancarlin @nadime15

nadime15:
@ibvqeibob Did you remove the test? It would be nice to keep one so we can see the code in action.

Timmmm:
I've started working on this (based on this code) on my branch [here](https://github.com/Timmmm/sail-riscv/tree/user/timh/plic)

ibvqeibob:
> I've started working on this (based on this code) on my branch [here](https://github.com/Timmmm/sail-riscv/tree/user/timh/plic)

Sorry — it looks like I accidentally closed this PR.
Thanks @Timmmm for picking it up. I’m happy for you to take the lead on a follow-up PR from your branch, and I can help by reviewing/testing and answering any questions.
If you’re happy to, could you open the new PR from Timmmm:user/timh/plic to riscv:master and reference #1476?

PR #1479, opened 2026-01-12
Rename the `ext_access_type` type to `mem_payload`.

pmundkur:
This was initially introduced as a hook for CHERI to override as an external (`ext_`) extension.  However, this type can also be used to support memory accesses for standard extensions such as Zicfiss (and possibly others in addition to CHERI). The type is also  redefined to an `enum` (from `unit`) to support these forthcoming extensions.


jordancarlin (approved):
LGTM as discussed during today's meeting.

PR #1481, opened 2026-01-13
Add Hypervisor interrupt causes

nadime15:
Add interrupt causes for the hypervisor extension.

The new interrupt bits are defined, but are not yet writable via mip/mie. Full interrupt handling will be implemented once the hypervisor CSRs are in place.

### DIFF
--- a/model/core/types.sail
+++ b/model/core/types.sail
@@ -120,50 +120,50 @@ type is_mem_width('w) = 'w in {1, 2, 4, 8}
 enum InterruptType = {
   I_U_Software,
   I_S_Software,
+  I_VS_Software,
   I_M_Software,
-  // I_VS_Software,
   I_U_Timer,
   I_S_Timer,
-  // I_VS_Timer,
+  I_VS_Timer,
   I_M_Timer,
   I_U_External,
   I_S_External,
-  // I_VS_External,
+  I_VS_External,
   I_M_External,
-  // I_G_External,
+  I_SG_External,
 }
 
 mapping interruptType_bits : InterruptType <-> exc_code = {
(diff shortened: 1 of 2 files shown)


model/core/types.sail
```diff
@@ -120,50 +120,50 @@ type is_mem_width('w) = 'w in {1, 2, 4, 8}
...
-  I_U_Timer    <-> 0b000100, // 4
-  I_S_Timer    <-> 0b000101, // 5
-  // I_VS_Timer    <-> 0b00110, // 6
-  I_M_Timer    <-> 0b000111, // 7
-  I_U_External <-> 0b001000, // 8
-  I_S_External <-> 0b001001, // 9
-  // I_VS_External <-> 0b001010, // 10
-  I_M_External <-> 0b001011, // 11
-  // I_G_External  <-> 0b001100, // 12
+  I_U_Software   <-> 0b000000, // 0
```
Timmmm, line 137: Don't feel like you have to do it in this PR, but we should probably switch these U interrupts back to reserved...

Timmmm (approved):

model/core/types.sail
nadime15: Ups, I missed that, let me revert that!

nadime15: Oh wait, I thought I had added a new entry, never mind. I can do this in a follow-up PR and switch these back to reserved.

jordancarlin (approved):

model/core/types.sail
Timmmm: Added #1485 so we don't forget.

PR #1483, opened 2026-01-13
Use `bits(1)` and `0b0`/`0b1` consistently in place of `bit` and `bitone`/`bitzero`

jordancarlin:
The latest version of Sail made `bit` and `bits(1)` equivalent. Replace all instances of `bitzero` and `bitone` with `0b0` and `0b1` for consistency.


model/prelude/prelude.sail
```diff
@@ -117,12 +117,12 @@ mapping bool_int : bool <-> {0, 1} = {
...
-function bool_to_bit(x : bool) -> bit = bool_bit(x)
-function bit_to_bool(x : bit) -> bool = bool_bit(x)
+function bool_to_bit(x : bool) -> bits(1) = bool_bit(x)
+function bit_to_bool(x : bits(1)) -> bool = bool_bit(x)
 function bool_to_bits(x : bool) -> bits(1) = bool_bits(x)
 function bits_to_bool(x : bits(1)) -> bool = bool_bits(x)
-function bit_to_bits(x : bit) -> bits(1) = bool_bits(bit_to_bool(x))
-function bits_to_bit(x : bits(1)) -> bit = bool_to_bit(bits_to_bool(x))
+function bit_to_bits(x : bits(1)) -> bits(1) = bool_bits(bit_to_bool(x))
+function bits_to_bit(x : bits(1)) -> bits(1) = bool_to_bit(bits_to_bool(x))
```
arichardson: Can we remove those now or do we need to wait for a new sail release?

model/prelude/prelude.sail
```diff
@@ -100,9 +100,9 @@ function ones (n) = sail_ones(n)
 val trunc : forall 'm 'n, 'm >= 0 & 'm <= 'n. (implicit('m), bits('n)) -> bits('m)
 function trunc(m, v) = truncate(v, m)
 
-mapping bool_bit : bool <-> bit = {
-  true <-> bitone,
-  false <-> bitzero,
+mapping bool_bit : bool <-> bits(1) = {
```
arichardson: Can we remove one of these mappings?

model/prelude/prelude.sail
jordancarlin: Done

model/prelude/prelude.sail
jordancarlin: Oops. Got carried away with the find-and-replace. Removed.

arichardson (approved):
LGTM. No opinion on `bit` vs `bits(1)`, both of those seem fine to me. If it was really common, I'd vote for `bit` but since there only seem to be about 20 uses, might as well use `bits(1)` for consistency.

Timmmm:
Yeah I agree `bits(1)` is more consistent. I would imagine the `bit` type could be removed from Sail entirely in future (or maybe just converted to a type alias in `prelude.sail`).

model/prelude/prelude.sail
```diff
@@ -117,12 +112,8 @@ mapping bool_int : bool <-> {0, 1} = {
 }
 
 // These aliases make the conversion direction a bit clearer.
-function bool_to_bit(x : bool) -> bit = bool_bit(x)
-function bit_to_bool(x : bit) -> bool = bool_bit(x)
 function bool_to_bits(x : bool) -> bits(1) = bool_bits(x)
 function bits_to_bool(x : bits(1)) -> bool = bool_bits(x)
```
Timmmm: I wonder if these should be called `bit_to_bool`/`bool_to_bit` since it is always a single bit and there's no need to distinguish the name from the original `bit_to_bool` now?

jordancarlin: Could go either way. I left it as `bits` because that more directly maps to the types that are being used, but happy to switch it if you'd rather `bit_to_bool`.

Timmmm: I think `bit_to_bool` is semantically clearer, otherwise you might read it and think "how does it convert multiple bits to a bool". And now we don't use the `bit` type there's no confusion that it would mean a different type altogether.

jordancarlin: Updated to `bit_to_bool`, `bool_to_bit` and `bool_bit`.

PR #1484, opened 2026-01-14
Get semicolon-separated list of source files from Sail

Timmmm:
Instead of getting a space-separated list of source files from the Sail compiler and converting to a CMake list, get a semicolon separated list directly.

This is not strictly correct because it will fail on filenames containing semicolons, but the old code would have failed on filenames containing spaces, so this is an improvement at least.

This is also necessary for it to work at all on Windows, because `UNIX_COMMAND` interprets the backslashes in Windows file paths as escape sequences.


jordancarlin (approved):
LGTM

PR #1486, opened 2026-01-14
Validate vector register groups.

### LINKED ISSUE #1455: Assertion failed: invalid register group: vrid is not a multiple of EMUL
## Context
I'm trying to write a `vle8` test case *ab initio*, validated by Sail RISC-V model. I got the assertion shown in the title, but couldn't find `vrid` anywhere in the [Vector chapter](https://docs.riscv.org/reference/isa/unpriv/v-st-ext.html) of the spec.

## Current
On binary release v0.9 of the sail-riscv model:
> [76] [M]: 0x00000000800001A0 (0x30200073) mret                                      reset_vector+336
CSR mstatus (0x300) <- 0x8000000A00006680
mem[X,0x00000000800001A4] -> 0x7357
mem[X,0x00000000800001A6] -> 0xC031
[77] [U]: 0x00000000800001A4 (0xC0317357) vsetivli x6, 0x2, e8, m8, tu, mu          reset_vector+340
x6 <- 0x0000000000000002
CSR mstatus (0x300) <- 0x8000000A00006680
CSR vstart (0x008) <- 0x0000000000000000
CSR vtype (0xC21) <- 0x0000000000000003
CSR vl (0xC20) <- 0x0000000000000002
CSR vstart (0x008) <- 0x0000000000000000
mem[X,0x00000000800001A8] -> 0x2297
mem[X,0x00000000800001AA] -> 0x0000
[78] [U]: 0x00000000800001A8 (0x00002297) auipc x5, 0x2                             reset_vector+344
x5 <- 0x00000000800021A8
mem[X,0x00000000800001AC] -> 0x8293
mem[X,0x00000000800001AE] -> 0xE582
[79] [U]: 0x00000000800001AC (0xE5828293) addi x5, x5, -0x1a8                       reset_vector+348
x5 <- 0x0000000080002000
mem[X,0x00000000800001B0] -> 0x8107
mem[X,0x00000000800001B2] -> 0x0202
[80] [U]: 0x00000000800001B0 (0x02028107) vle8.v v2, (x5)                           reset_vector+352
Assertion failed: invalid register group: vrid is not a multiple of EMUL

## Expected 
The unit-stride vector load instruction should load values 45-44 (0x2d2c) in v2 since the prior `vsetivlsi` instruction sets up `vl` correctly to `2` and `x5` contains the address of the `input` data as seen in the `dump` file in the ZIP archive. 

## Steps to reproduce
1. Extract the `elf` file `rv64uv-p-vle8` from the attached zip archive: 
[vle8.zip](https://github.com/user-attachments/files/24320163/vle8.zip)
 
2. Run 
`<SAIL_RISCV_PATH>/sail_riscv_sim --trace-instr --trace-reg --trace-mem --config <SAIL_RISCV_CONFIG_PATH>/rv64d_v128_e64.json rv64uv-p-vle8`

3. Note the error message in the title of this issue.

pmundkur:
This adds missed checks to ensure vector register groups are EMUL-aligned.

It also strengthens the existing register group overlap check (which is used when narrowing/widening) to ensure that the register groups are aligned, saving a separate alignment validity check.

A similar group alignment check is added for whole register loads/stores.

This adds a simple first party test to check for vector register alignment in `vadd.vv`.

This fixes #1455.

### DIFF
--- a/model/extensions/V/vext_control.sail
+++ b/model/extensions/V/vext_control.sail
@@ -68,8 +68,8 @@ function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
   let LMUL_pow_reg = if LMUL_pow < 0 then 0 else LMUL_pow;
   let LMUL = 2 ^ LMUL_pow_reg;
   let vrid_end = vrid_val + LMUL;
-  assert(vrid_end <= 32, "invalid register group: group " ^ dec_str(vrid_val) ^ " ends at " ^ dec_str(vrid_end) ^ " and overflows the largest register number (32)");
-  assert(vrid_val % LMUL == 0, "invalid register group: group " ^ dec_str(vrid_val) ^ " is not a multiple of its EMUL " ^ dec_str(LMUL));
+  assert(vrid_end <= 32, "Invalid register group: group " ^ dec_str(vrid_val) ^ " ends at " ^ dec_str(vrid_end) ^ " and overflows the largest register number (32).");
+  assert(vrid_val % LMUL == 0, "Invalid register group: group " ^ dec_str(vrid_val) ^ " is not a multiple of its EMUL " ^ dec_str(LMUL) ^ ".");
 
   var result : vector('n, bits('sew)) = vector_init(zeros());
 


Timmmm:
Can you add [this test](https://github.com/riscv/sail-riscv/issues/1455#issuecomment-3694055455)?

pmundkur:
> Can you add [this test](https://github.com/riscv/sail-riscv/issues/1455#issuecomment-3694055455)?

I ran it and it runs to completion but printed "FAILURE: 1" (but the assertion failure is gone).  I'll take a look again today.   The first test attached to #1455 passes.

nadime15:
Honestly, at this point it would make more sense to define a proper (nested) vector instruction decoder that performs these checks in a single, centralized place, rather than duplicating the same logic more than 150 times. We could introduce a vector decoder function that handles the base checks first, then uses a match block to determine whether the instruction belongs to a specific extension and apply any additional validation as needed. 

This would reduce code duplication, simplify the individual instruction implementations, and make the overall design easier to read and integrate into the specification.

pmundkur:
I can break out the whitespace changes into a separate PR to make this easier to review.

model/extensions/V/vext_mem_insts.sail
```diff
@@ -295,13 +299,14 @@ function clause execute VLSSEGTYPE(nf, vm, rs2, rs1, width, vd) = {
   let SEW_pow = get_sew_pow();
   let LMUL_pow = get_lmul_pow();
   let EMUL_pow = EEW_pow - SEW_pow + LMUL_pow;
-  // TODO: The EMUL_pow calculation here is incorrect. See https://github.com/riscv/sail-riscv/issues/1196
+
   assert(-3 <= EMUL_pow & EMUL_pow <= 3);
```
Timmmm, line 303: We probably still need a TODO at least since you can still trigger this assertion; it is reserved behaviour.

Timmmm:
Yeah I agree the amount of duplication is slightly unfortunate. Maybe we should merge this first anyway though and then try to simplify things?

I wouldn't be overly surprised if there are so many slight differences between the vector instructions that you can't actually get much simpler than this.

pmundkur:
That's the tradeoff with this scattered approach; it centralizes around the per-instruction axis, but scatters the rest.  Centralizing around the decode axis would have different tradeoffs but they would still exist.

nadime15:
> Yeah I agree the amount of duplication is slightly unfortunate. Maybe we should merge this first anyway though and then try to simplify things?

Yeah I agree

> That's the tradeoff with this scattered approach; it centralizes around the per-instruction axis, but scatters the rest. Centralizing around the decode axis would have different tradeoffs but they would still exist.

I mean right now it is basically centralized already, because the `valid_reg_group()` function is now called every time. There are also other checks that could be added to this function, for example some vector crypto instructions have additional checks that are part of the encdec clause or floating point etc.

You could just call a function like `check_vector_insn_legal()` as one of the first lines in each execute clause, or even as part of the encdec clause if we are talking about illegal instructions, and forward all information including the funct6 field. From there, you can quickly decode which vector instruction group it belongs to and then check everything according to whatever the spec says.

pmundkur:
> Can you add [this test](https://github.com/riscv/sail-riscv/issues/1455#issuecomment-3694055455)?

Added.

model/extensions/V/vext_arith_insts.sail
```diff
@@ -1140,7 +1165,19 @@ function clause execute VISG(funct6, vm, vs2, simm, vd) = {
...
+  let group_overlap : bool = match funct6 {
+    VI_VSLIDEUP   => vd == vs2,
+    VI_VSLIDEDOWN => false,
+    VI_VRGATHER   => vd == vs2,
+  };
+  if illegal_normal(vd, vm) |
+     not(valid_reg_group(vs2, LMUL_pow)) |
+     not(valid_reg_group(vd, LMUL_pow)) |
+     group_overlap
+  then return Illegal_Instruction();
```
nadime15: The code has some redundancy. 

1. 
```
 let group_overlap : bool = match funct6 {
    VI_VSLIDEUP   => vd == vs2,
    VI_VSLIDEDOWN => false,
    VI_VRGATHER   => vd == vs2,
  };
```
You now check for that before the main loops starts, which is fine but the main loop down below still has these checks, they need to be removed.


2. I noticed that you modified valid_reg_overlap() and added 
```
  if rs_int % rs_group != 0 | rd_int % rd_group != 0
  then return false;
```

So the following would be enough:

```
if illegal_normal(vd, vm) | 
   not(valid_reg_overlap(vs2, vd, LMUL_pow, LMUL_pow)) | 
   group_overlap
```

model/extensions/V/vext_arith_insts.sail
```diff
@@ -51,7 +51,10 @@ function clause execute VVTYPE(funct6, vm, vs2, vs1, vd) = {
   let LMUL_pow = get_lmul_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_normal(vd, vm) then return Illegal_Instruction();
+  if illegal_normal(vd, vm) |
+     not(valid_reg_overlap(vs1, vd, LMUL_pow, LMUL_pow)) |
+     not(valid_reg_overlap(vs2, vd, LMUL_pow, LMUL_pow))
+  then return Illegal_Instruction();
```
nadime15, line 73: I am not sure if you want to move the checks for if (vs1 == vd | vs2 == vd) up here, similar to what you did below (if so we need to remove these checks from the main loop).

Aside from that, I think we missed a nasty case:

```
VV_VRGATHEREI16  => {
                              if (vs1 == vd | vs2 == vd) then return Illegal_Instruction();
                              // vrgatherei16.vv uses SEW/LMUL for the data in vs2 but EEW=16 and EMUL = (16/SEW)*LMUL for the indices in vs1
                              let vs1_new : vector('n, bits(16)) = read_vreg(num_elem, 16, 4 + LMUL_pow - SEW_pow, vs1);
                              let idx = unsigned(vs1_new[i]);
                              assert(LMUL_pow + vlen_exp - SEW_pow >= 0);
                              let VLMAX = 2 ^ (LMUL_pow + vlen_exp - SEW_pow);
                              assert(VLMAX <= 'n);
                              if idx < VLMAX then vs2_val[idx] else zeros()
                            }
       
```                     
In this case, we compute a new LMUL (EMUL) for `VV_VRGATHEREI16` that is different from the other cases. Because of that, we need to move this logic earlier or refactor a bit more.

One option would be something along the lines of: 

`let VS1_LMUL = if funct6 == VV_VRGATHEREI16 then (4 + get_lmul_pow() - SEW_pow) else get_lmul_pow();`

and then: 

`not(valid_reg_overlap(vs1, vd, VS1_LMUL, LMUL_pow))`

otherwise we miss this case.

model/extensions/V/vext_arith_insts.sail
pmundkur: 1. Good catch! I'll remove the checks in the loop.
2. Although `valid_reg_overlap()` would suffice, it should really be used when checking overlaps for narrowing/widening instructions (as the comment above it says).  When both `vs` and `vd` EMULs are the same, there is no widening or narrowing, so I think it's better to be explicit and just check for aligned register groups, since that's all that's needed.  I thought of changing the name of `valid_reg_overlap` to indicate this, but I couldn't think of a succinct name, and any name change would have just distracted from the rest of the changes.

model/extensions/V/vext_arith_insts.sail
pmundkur: Great catch, thanks!  I'll also hoist the `read_vreg()` out of the loop for ` VV_VRGATHEREI16`, which seems inefficient.

model/extensions/V/vext_arith_insts.sail
```diff
@@ -51,16 +51,40 @@ function clause execute VVTYPE(funct6, vm, vs2, vs1, vd) = {
...
 
   let vm_val  : bits('n)             = read_vmask(num_elem, vm, zvreg);
-  let vs1_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs1);
+  let vs1_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs1); // not used for VV_VRGATHEREI16.
   let vs2_val : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vs2);
   let vd_val  : vector('n, bits('m)) = read_vreg(num_elem, SEW, LMUL_pow, vd);
 
+  // See above for the vs1 operand of vrgatherei16.vv.
+  let vs1_val_vrgatherei16 : vector('n, bits(16)) = if funct6 == VV_VRGATHEREI16
+                                                    then read_vreg(num_elem, 16, 4 + LMUL_pow - SEW_pow, vs1)
```
nadime15:
```suggestion
  let vs1_val_vrgatherei16 : vector('n, bits(16)) = if funct6 == VV_VRGATHEREI16
                                                    then read_vreg(num_elem, 16, vs1_emul_pow, vs1)
```

model/extensions/V/vext_control.sail
```diff
@@ -66,8 +66,10 @@ val read_vreg : forall 'n 'sew 'p, 'n >= 0 & is_sew_bitsize('sew) . (int('n), in
 function read_vreg(num_elem, SEW, LMUL_pow, vrid) = {
   let vrid_val = unsigned(vregidx_bits(vrid));
   let LMUL_pow_reg = if LMUL_pow < 0 then 0 else LMUL_pow;
-  assert(vrid_val + 2 ^ LMUL_pow_reg <= 32, "invalid register group: vrid overflow the largest number");
-  assert(vrid_val % (2 ^ LMUL_pow_reg) == 0, "invalid register group: vrid is not a multiple of EMUL");
+  let vrid_end = vrid_val + 2 ^ LMUL_pow_reg;
+  let LMUL = 2 ^ LMUL_pow_reg;
```
jordancarlin: Minor simplification:
```suggestion
  let LMUL = 2 ^ LMUL_pow_reg;
  let vrid_end = vrid_val + LMUL;
```

model/extensions/V/vext_arith_insts.sail
```diff
@@ -1970,7 +2052,15 @@ function clause execute MVXTYPE(funct6, vm, vs2, rs1, vd) = {
   let LMUL_pow = get_lmul_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_normal(vd, vm) then return Illegal_Instruction();
+  // "The vslide1up instruction requires that the destination vector
+  // register group does not overlap the source vector register
+  // group. Otherwise, the instruction encoding is reserved."
```
nadime15, line 2057: For me this looks good. The only thing I’m personally not a big fan of is copying parts of the spec into the code and adding comments unless it’s really needed and the code would not be comprehensible otherwise or the reasoning is not clear. The code should be self explanatory without needing to add text from the spec.

I think the example I highlighted and the other two above are understandable as they are and the reasoning is clear, so we don’t need these comments. But this does not block the PR from merging.

nadime15 (approved):

model/extensions/V/vext_arith_insts.sail
pmundkur: The reason I'm doing this (quoting spec fragments) is to make a future documentation goal easier: to link from the Sail model back to the spec (e.g. via normative rules).   When those links are eventually added, they would replace the spec comments; the spec comments are just an aid to find the relevant normative rule(s).

PR #1492, opened 2026-01-18
Remove N extension InterruptType values

### LINKED ISSUE #1485: Remove user interrupts
We removed the never-ratified N extension a while ago, but the `InterruptType` type still has user interrupt values from that extension that are currently `Reserved` in the ISA manual. We should rename them to `Reserved_0` etc.

KotorinMinami:
Remove the N extension InterruptType values and replace them with Reserved:

- I_U_Software -> I_Reserved_0
- I_U_Timer      -> I_Reserved_4
- I_U_External  -> I_Reserved_8

Fixes #1485

### DIFF
--- a/model/core/types.sail
+++ b/model/core/types.sail
@@ -122,15 +122,15 @@ type is_mem_width('w) = 'w in {1, 2, 4, 8}
 // architectural interrupt definitions
 
 enum InterruptType = {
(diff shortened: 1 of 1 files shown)


Timmmm (approved):
Nice, thanks!

Issue #1493, opened 2026-01-18
master is broken and apparently the merge queue CI failures don't stop the merge??


Timmmm:
When I merged #1405 CI failed after a couple of minutes, but it merged anyway! The compile error is

```
Error:
extensions/Ssqosid/ssqosid.sail:9.4-26:
9 |let rcidlen : range(1, 12) = config extensions.Ssqosid.rcid_length
  |    ^--------------------^
  | SMT solver returned unexpected status 1
ninja: build stopped: subcommand failed.
```

So I guess it was actually caused by #379 triggering a compiler bug and then getting merged anyway.

We should:

1. Fix `master`
2. Figure out what's going on with our CI config.

Timmmm:
@pmundkur do we have `Require status checks to pass before merging` set? See [this comment](https://github.com/orgs/community/discussions/75568).

Timmmm:
Or possibly [`Only merge non-failing pull requests`](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue)?

Can I have permission to access these settings?

jordancarlin:
Bizarrely, I just checked out b1d73282 (the commit where #379) was merged, and it compiled without issues. Seems like the bug actually was introduced in #1405 even though the error is showing as related to `rcidlen`.

Definitely concerning that this made it through the merge queue even though CI failed.

jordancarlin:
For reasons that are unclear to me, it looks like it is an issue with `zelen_exp` coming from the `check_vlen_elen` function. Not sure why that function is getting interwoven with `rcid_length`.

Output with the `--dsmt-verbose` and `--dtc-verbose` flags:

```
Checking top-level let
wf_typ: range(1, 12) exs: 
|   wf_nexp: 1 exs: 
|   wf_nexp: 12 exs: 
|   Check config extensions.Ssqosid.rcid_length <= range(1, 12)
|   wf_typ: range(1, 12) exs: 
|   |   wf_nexp: 1 exs: 
|   |   wf_nexp: 12 exs: 
Adding type variable '_#rcidlen : Int
wf_constraint: (1 <= '_#rcidlen & '_#rcidlen <= 12) exs: 
wf_constraint: 1 <= '_#rcidlen exs: 
wf_nexp: 1 exs: 
wf_nexp: '_#rcidlen exs: 
[...]

PR #1494, opened 2026-01-19
Fix model compilation

Timmmm:
This fixes a compilation error introduced by the previous commit. The issue was caused because of a race between adding Ssqosid and refactoring the memory access type.

The Ssqosid commit added an extension to the project file, and the memory access type commit needed to modify that to add `requires core`, however since there was no git conflict it was merged based on an earlier passing CI run.

The merge queue is supposed to catch these kinds of races however it seems to be misconfigured in some way.

Separately, there's a bug in the Sail compiler that meant that the missing `requires core` causes weird invalid SMT generation rather than a `foo is not visible` error message.

See #1493

### DIFF
--- a/model/riscv.sail_project
+++ b/model/riscv.sail_project
@@ -429,6 +429,7 @@ extensions {
   }
 
   Ssqosid {
+    requires core
     files extensions/Ssqosid/ssqosid.sail
   }
 


jordancarlin:
Going to go ahead and merge this (assuming CI passes) so that we can get master building again.

PR #1498, opened 2026-01-20
Use C++ chrono for benchmarking instead of gettimeofday

Timmmm:
This is a bit simpler, and also works on Windows. I skipped the floating point calculation on the overall stat reporting for consistency with the other place that didn't do that, and also because we don't really care about fractions of a kIPS.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -1,4 +1,5 @@
 #include <cassert>
+#include <chrono>
 #include <climits>
 #include <cstring>
 #include <ctype.h>
@@ -34,6 +35,12 @@
 #include "rvfi_dii.h"
 #include "sail_riscv_version.h"
 
+using std::chrono::duration_cast;
+using std::chrono::milliseconds;
+using std::chrono::steady_clock;
+
+namespace {
+
 bool do_show_times = false;
 bool do_print_version = false;
 bool do_print_build_info = false;
@@ -46,7 +53,6 @@ bool do_print_isa = false;
 std::string config_file;
 std::string term_log;
 std::string trace_log_path;
-FILE *trace_log = stdout;
 std::string dtb_file;
 int rvfi_dii_port = 0;
 std::optional<rvfi_handler> rvfi;
@@ -62,6 +68,24 @@ uint64_t mem_sig_start = 0;
 uint64_t mem_sig_end = 0;
 const int DEFAULT_SIGNATURE_GRANULARITY = 4;
 int signature_granularity = DEFAULT_SIGNATURE_GRANULARITY;
+
+steady_clock::time_point init_start;
+steady_clock::time_point init_end;
+
+uint64_t total_insns = 0;
+uint64_t insn_limit = 0;
+#ifdef SAILCOV
+char *sailcov_file = nullptr;
+#endif
+
+// Single global model instance.
+// TODO: This shouldn't be global, but it is to facilitate gradual transition
+// from the C Sail output which was necessarily global.
+ModelImpl g_model;
+
+} // namespace
+
+FILE *trace_log = stdout;
 
 bool config_print_instr = false;
 bool config_print_reg = false;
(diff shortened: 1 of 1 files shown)


c_emulator/riscv_sim.cpp
```diff
@@ -458,13 +459,12 @@ void run_sail(void) {
...
-        fprintf(stderr, "Cannot gettimeofday: %s\n", strerror(errno));
-        exit(EXIT_FAILURE);
-      }
-      uint64_t end_us = 1000000 * ((uint64_t)interval_start.tv_sec) + ((uint64_t)interval_start.tv_usec);
-      fprintf(stdout, "kips: %" PRIu64 "\n", ((uint64_t)1000) * 0x100000 / (end_us - start_us));
+      const auto now = steady_clock::now();
+      const auto interval = now - interval_start;
+      interval_start = now;
+
+      uint64_t kips = 0x100000 / duration_cast<milliseconds>(interval).count();
```
arichardson, line 470: Is this to convert to base-2? I think for kIPS/MIPS the convention is to use base 10? million instructions per second?

Timmmm: No it's because it prints the speed every 0x100000 instructions. See line 461. Slightly odd way to do it - I would have done `(total_insns % 0x100000) == 0` which is more normal and obvious. I'll update that in this PR I think.

Timmmm:
Ok this should pass CI now and I manually tested that it works.

c_emulator/riscv_sim.cpp
```diff
@@ -374,17 +382,14 @@ void finish() {
...
-    if (gettimeofday(&run_end, nullptr) < 0) {
-      fprintf(stderr, "Cannot gettimeofday: %s\n", strerror(errno));
-      exit(EXIT_FAILURE);
-    }
-    int init_msecs = (init_end.tv_sec - init_start.tv_sec) * 1000 + (init_end.tv_usec - init_start.tv_usec) / 1000;
-    int exec_msecs = (run_end.tv_sec - init_end.tv_sec) * 1000 + (run_end.tv_usec - init_end.tv_usec) / 1000;
-    double Kips = ((double)total_insns) / ((double)exec_msecs);
-    fprintf(stderr, "Initialization:   %d msecs\n", init_msecs);
-    fprintf(stderr, "Execution:        %d msecs\n", exec_msecs);
+    run_end = steady_clock::now();
```
arichardson: `run_end` only seems to be used here, does it need to be global?

c_emulator/riscv_sim.cpp
```diff
@@ -80,7 +85,10 @@ bool config_enable_rvfi = false;
 
 bool config_enable_experimental_extensions = false;
 
-struct timeval init_start, init_end, run_end;
+steady_clock::time_point init_start;
+steady_clock::time_point init_end;
+steady_clock::time_point run_end;
```
```suggestion
namespace {
steady_clock::time_point init_start;
steady_clock::time_point init_end;
}
```
since it's only used in this file?
Or 
```suggestion
static steady_clock::time_point init_start;
static steady_clock::time_point init_end;
```

c_emulator/riscv_sim.cpp
```diff
@@ -653,10 +650,7 @@ int inner_main(int argc, char **argv) {
 
   init_sail(entry, config_file.c_str());
 
-  if (gettimeofday(&init_end, nullptr) < 0) {
-    fprintf(stderr, "Cannot gettimeofday: %s\n", strerror(errno));
-    exit(EXIT_FAILURE);
-  }
+  init_end = steady_clock::now();
```
arichardson, line 657: I guess we could also just store init_duration since that's all that's printed on exit?

arichardson (approved):
a few very minor nits, LGTM.

c_emulator/riscv_sim.cpp
Timmmm: Good point, I made it non-global

c_emulator/riscv_sim.cpp
Timmmm: Done (I also moved a load of other stuff that didn't need to be `extern` into an anonymous namespace).

c_emulator/riscv_sim.cpp
Timmmm: True... I decided to leave this for now. I'm planning some more extensive refactoring of this file so I reckon it will change anyway.

PR #1500, opened 2026-01-23
Ensure physical memory access checks are not bypassed for failing store-conditionals.

### LINKED ISSUE #1239: Zalrsc (sc.w/d) checking reservation before checking pmp permission.
A region with no permission (XWR-000) from 0x80002000 to 0x80002040 is there and we're trying to lr.w from start of the region.

```
[233] [M]: 0x800004A0 (0x1007A62F) lr.w x12, (x15)
trapping from M to M to handle load-access-fault
handling exc#0x05 at priv M with tval 0x80002000
CSR mstatus (0x300) <- 0x00001800
CSR mcause (0x342) <- 0x00000005
CSR mtval (0x343) <- 0x80002000
CSR mepc (0x341) <- 0x800004A0 
```
Since, there's no permission read permission, it throws access-fault and when we're trying to sc.w at the same address, with no write permission it's checking reservation first and since no load occur at that address, it simply putting 1 in rd. But it should check the pmp permission first and throw access-fault.

```
[345] [M]: 0x800004AC (0x18C7A62F) sc.w x12, x12, (x15)
x12 <- 0x00000001
```

pmundkur:
Also add a first-party test to check this.

Fixes #1239.

### DIFF
--- a/test/first_party/src/test_phys_perms_on_failing_sc.S
+++ b/test/first_party/src/test_phys_perms_on_failing_sc.S
@@ -14,7 +14,6 @@ main:
   # PMA config is 128m, so use the smaller one. To fit in a 32-bit range for RV32, take
   # the last page of RAM as the test region, and shrink the computed size of RAM.
   # Assume a PMP grain of 0 as in the default config.
-#define PMP_SHIFT 2
 #define RAM_START 0x80000000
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):
LGTM.

I've also reached out to one of the people working on the test that discovered #1239 and they will be testing this patch soon.

PR #1501, opened 2026-01-23
Fix a couple of TODOs now that the VU mode and virtual instruction exceptions are defined.

pmundkur:


pmundkur:
Right about `hstatus` which is why I left it.  But for SFENCE.VMA in VU-mode, it's obvious it should raise virtual instruction, but I couldn't find it explicitly in the spec.  Did I miss it?  It looks like it was missed in this table at least:
https://docs.riscv.org/reference/isa/priv/hypervisor.html#sec:hcauses:~:text=Specifically%2C%20a%20virtual,1,-%2E

pmundkur:
> Right about `hstatus` which is why I left it. But for SFENCE.VMA in VU-mode, it's obvious it should raise virtual instruction, but I couldn't find it explicitly in the spec. Did I miss it? It looks like it was missed in this table at least: https://docs.riscv.org/reference/isa/priv/hypervisor.html#sec:hcauses:~:text=Specifically%2C%20a%20virtual,1,-%2E

Found it, I should have grepped for `SFENCE` instead of `SFENCE.VMA`.

PR #1504, opened 2026-01-27
Fix CSR privilege check for hypervisor mode.

### LINKED ISSUE #1503: `check_CSR_priv` breaks for hypervisor
Both HS-mode and VS-mode have the same `nom_priv_bits` encoding, but HS-mode is actually higher privilege than VS-mode.  This breaks `check_CSR_priv` which compares the required privilege using `nom_priv_bits`.    The HS-mode CSR addresses encode privilege bits of `10` which is greater than the `01` used to encode VS and HS modes.  So `check_CSR_priv` will currently always reject accesses for hypervisor CSRs from VS and HS modes.

pmundkur:
When hypervisor is enabled, `Supervisor` (HS-mode) and `VirtualSupervisor` (VS-mode) privileges are encoded as `0b01` in `privLevel_bits()`, but the CSR addresses for hypervisor and VS CSRs encode the required privilege as `0b10`.  Define a private function to correct for this when checking privilege.

Fixes #1503.

### DIFF
--- a/model/sys/sys_control.sail
+++ b/model/sys/sys_control.sail
@@ -28,6 +28,9 @@ private function privLevel_to_CSR_privbits(p : Privilege) -> nom_priv_bits =
   match p {
     User              => 0b00,
     VirtualUser       => 0b00,
+    // Note that this `if` is not strictly necessary. We could return
+    // 0b10 unconditionally if other checks ensure that hypervisor
+    // CSRs do not exist if H is not enabled.
     Supervisor        => if currentlyEnabled(Ext_H) then 0b10 else 0b01,
     VirtualSupervisor => 0b01,
     Machine           => 0b11,


model/sys/sys_control.sail
```diff
@@ -18,9 +18,24 @@ function effectivePrivilege(access : MemoryAccessType(mem_payload), m : Mstatus,
...
+// `privLevel_bits()`, but the CSR addresses for hypervisor and VS
+// CSRs encode the required privilege as `0b10`.  This overrides
+// the encoding in `privLevel_bits()` for the purposes of CSR privilege
+// checks.
+private function privLevel_to_CSR_privbits(p : Privilege) -> nom_priv_bits =
+  match p {
+    User              => 0b00,
+    VirtualUser       => 0b00,
+    Supervisor        => if currentlyEnabled(Ext_H) then 0b10 else 0b01,
+    VirtualSupervisor => 0b10,
```
radimkrcmar: Shouldn't `VirtualSupervisor` be `0b01`?

pmundkur: Good catch, thanks!

nadime15: Yeah thats why I prefer to fetch these bits via privLevel_to_bits() otherwise it's just to easy to miss. 

But the whole check_CSR_priv needs to be modified anyway even after this change, because even with this change and #1402 we still don’t distinguish whether an access should be rejected due to Virtual mode, for example, a VU-mode access to an M-mode CSR.

Something like this is probably necessary:

```
function check_CSR_priv_test(csr : csreg, p : Privilege) -> CSRAccessResult =
  let p_bits : bits(2) = match p {
    User              => 0b00,
    VirtualUser       => 0b00,
    Supervisor        => if currentlyEnabled(Ext_H) then 0b10 else 0b01,
    VirtualSupervisor => 0b01,
    Machine           => 0b11,
  } in (if p_bits >=_u csrPriv(csr) 
    then CSRAccessSuccessfull 
    else (if privLevel_is_virtual(p) 
             then CSRAccess_Virtual
             else CSRAccess_Illegal))
```


It might make sense just to wait.

pmundkur: It might be better to refactor correct code for #1402 rather than the current incorrect code, so perhaps merge this before #1402 and not wait?

nadime15: I mean it’s WIP and the case can’t even happen. My thought was that someone opens a dedicated PR that refactors all the CSR access checks needed for the hypervisor spec, so we don’t end up spreading changes across multiple PRs.

But either way works for me!

model/sys/sys_control.sail
```diff
@@ -18,9 +18,24 @@ function effectivePrivilege(access : MemoryAccessType(mem_payload), m : Mstatus,
...
+// `VirtualSupervisor` (VS-mode) privileges are encoded as `0b01` in
+// `privLevel_bits()`, but the CSR addresses for hypervisor and VS
+// CSRs encode the required privilege as `0b10`.  This overrides
+// the encoding in `privLevel_bits()` for the purposes of CSR privilege
+// checks.
+private function privLevel_to_CSR_privbits(p : Privilege) -> nom_priv_bits =
+  match p {
+    User              => 0b00,
+    VirtualUser       => 0b00,
+    Supervisor        => if currentlyEnabled(Ext_H) then 0b10 else 0b01,
```
Timmmm, line 34: I think this would actually work if it was just `0b10` right? If hypervisor is disabled non of the `0b10` CSRs exist anyway. I dunno if that's necessarily better code though.

Also I think this is correct but I couldn't find anywhere that really explicitly says how this is meant to work in the spec. Just this

> The next two bits (csr[9:8]) encode the lowest privilege level that can access the CSR.

Which is clear without hypervisor but not very clear at all with hypervisor.

pmundkur: Agreed that this is not clear.  Opened https://github.com/riscv/riscv-isa-manual/issues/2630

pmundkur: Hopefully clearer now with https://github.com/riscv/riscv-isa-manual/pull/2631

Timmmm: What do you think about this?


```suggestion
    Supervisor        => 0b10,
```

pmundkur: Agreed that it may be redundant with an existence check, but I'm leaning towards keeping it as it is, to make sure we don't miss the condition when we refactor this for hypervisor in #1402.

Timmmm: Maybe just add a comment for now? `// Note that this `if` is not strictly necessary. We could return 0b10 unconditionally since hypervisor CSRs don't exist anyway if hypervisor is not enabled.`

model/sys/sys_control.sail
```diff
@@ -18,9 +18,27 @@ function effectivePrivilege(access : MemoryAccessType(mem_payload), m : Mstatus,
...
+// CSRs encode the required privilege as `0b10`.  This overrides
+// the encoding in `privLevel_bits()` for the purposes of CSR privilege
+// checks.
+private function privLevel_to_CSR_privbits(p : Privilege) -> nom_priv_bits =
+  match p {
+    User              => 0b00,
+    VirtualUser       => 0b00,
+    // Note that this `if` is not strictly necessary. We could return
+    // 0b10 unconditionally if other checks ensure that hypervisor
+    // CSRs do not exist if H is not enabled.
```
Timmmm, line 33: > if other checks ensure that hypervisor CSRs do not exist if H is not enabled.

Which they do, right? Is there actually a need for the `if currentlyEnabled(Ext_H)`? I can't think of a situation where it would be incorrect to just return `0b10`.

PR #1507, opened 2026-01-29
Add the Zibi extension.

challenger1024:
Add the Zibi extension, which adds conditional branch instructions to compare with small immediates. This extension isn't ratified yet so it is gated behind the `--enable-experimental-extensions` flag.

Spec: https://github.com/riscv/zibi/releases/
More info: https://github.com/riscv/zibi/issues/2

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,6 @@ For booting operating system images, see the information under the
 - RV32I and RV64I base ISAs, v2.1
 - RV32E and RV64E base ISAs, v2.0
 - Zifencei extension for instruction-fetch fence, v2.0
-- Zibi extension for conditional branches with immediate operands, v0.6
 - Zicsr extension for CSR instructions, v2.0
 - Zicntr and Zihpm extensions for counters, v2.0
 - Zicond extension for integer conditional operations, v1.0
@@ -185,9 +184,9 @@ For booting operating system images, see the information under the
 - Physical Memory Protection (PMP)
 - Static memory regions with some static PMAs (Physical Memory Attributes)
 
-<!-- Uncomment the following section when unratified extensions are added
 The following unratified extensions are supported and can be enabled using the `--enable-experimental-extensions` flag:
--  -->
+
+- Zibi extension for conditional branches with immediate operands, v0.6
 
 **For a list of unsupported extensions and features, see the [Extension Roadmap](https://github.com/riscv/sail-riscv/wiki/Extension-Roadmap).**
 
--- a/model/extensions/Zibi/zibi_insts.sail
+++ b/model/extensions/Zibi/zibi_insts.sail
@@ -18,18 +18,13 @@ mapping encdec_biop : biop <-> bits(3) = {
   BNEI  <-> 0b011
 }
 
-function b_imm5(cimm : bits(5)) -> bits(5) =
-  if cimm == 0b00000
-  then 0b11111
-  else cimm
-
 mapping clause encdec = BITYPE(imm @ 0b0, cimm, rs1, op)
   <-> imm[11] @ imm[9..4] @ cimm @ encdec_reg(rs1) @ encdec_biop(op) @ imm[3..0] @ imm[10] @ 0b1100011
   when currentlyEnabled(Ext_Zibi)
 
 $[split op]
 function clause execute BITYPE(imm, cimm, rs1, op) = {
-  let cimmext : xlenbits = sign_extend(b_imm5(cimm));
+  let cimmext : xlenbits = if cimm == zeros() then ones() else zero_extend(cimm);
   let taken : bool = match op {
     BEQI  => X(rs1) == cimmext,
     BNEI  => X(rs1) != cimmext
@@ -45,7 +40,7 @@ mapping bitype_mnemonic : biop <-> string = {
 }
 
 mapping clause assembly = BITYPE(imm, 0b00000, rs1, op)
-  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_signed_5(0b11111) ^ sep() ^ hex_bits_signed_13(imm)
+  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ "-1" ^ sep() ^ hex_bits_signed_13(imm)
 mapping clause assembly = BITYPE(imm, cimm, rs1, op)
-  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_signed_5(cimm) ^ sep() ^ hex_bits_signed_13(imm)
+  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_5(cimm) ^ sep() ^ hex_bits_signed_13(imm)
   when cimm != 0b00000


jrtc27:
You don't need to sell the point of the extension in the commit message / PR description, just what it actually does

jrtc27:
> As this is a scattered union clause, this could also be caused by using a type defined after the 'scattered union' declaration

Did you read this message?

model/extensions/Zibi/zibi_insts.sail
```diff
@@ -0,0 +1,54 @@
...
+// This file defines the "Zibi" extension, which appends conditional branch instructions with an
+// immediate operand. In real world applications, conditional branches comparing with constants are
+// of great universality. With only conditional branch in current RISC-V ISA, we need to load
+// immediate values into registers before branching, which will increase register pressure and
+// potentially cause register spills/reloads.
+
+// The Zibi extension defines 2 instructions named BEQI and BNEI. BEQI and BNEI compare a register
+// with an immediate, and it takes the branch if registers rs1 and cimm are equal or unequal
+
+enum biop = {BEQI, BNEI}
```
jordancarlin: You need to split this into a separate `zibi_types.sail` file that comes before core. See how the Zicond extension is organized for a simple example.

challenger1024:
I’ve fixed it . thank you for the hint!

model/core/extensions.sail
```diff
@@ -70,6 +70,10 @@ enum clause extension = Ext_H
 mapping clause extensionName = Ext_H <-> "h"
 function clause hartSupports(Ext_H) = false
 
+enum clause extension = Ext_Zibi
+mapping clause extensionName = Ext_Zibi <-> "zibi"
+function clause hartSupports(Ext_Zibi) = config extensions.Zibi.supported
+function clause currentlyEnabled(Ext_Zibi) = hartSupports(Ext_Zibi)
```
jordancarlin: Move the `currentlyEnabled` clause into `zibi_insts.sail`.

model/core/extensions.sail
```diff
@@ -70,6 +70,10 @@ enum clause extension = Ext_H
 mapping clause extensionName = Ext_H <-> "h"
 function clause hartSupports(Ext_H) = false
 
+enum clause extension = Ext_Zibi
+mapping clause extensionName = Ext_Zibi <-> "zibi"
+function clause hartSupports(Ext_Zibi) = config extensions.Zibi.supported
```
jordancarlin: Until this is ratified, it needs to be gated on the `--enable-experimental-extensions` flag.

```suggestion
function clause hartSupports(Ext_Zibi) = sys_enable_experimental_extensions() & config extensions.Zibi.supported
```

model/extensions/Zibi/zibi_insts.sail
```diff
@@ -0,0 +1,47 @@
...
+
+mapping encdec_biop : biop <-> bits(3) = {
+  BEQI  <-> 0b010,
+  BNEI  <-> 0b011
+}
+
+function b_imm5(cimm : bits(5)) -> bits(5) =
+  if cimm == 0b00000 then
+    0b11111
+  else cimm
```
jordancarlin:
```suggestion
  if cimm == 0b00000
  then 0b11111
  else cimm
```

Or maybe just make this all one line.

model/extensions/Zibi/zibi_insts.sail
```diff
@@ -0,0 +1,47 @@
...
+}
+
+
+mapping bitype_mnemonic : biop <-> string = {
+  BEQI  <-> "beqi",
+  BNEI  <-> "bnei"
+}
+
+mapping clause assembly = BITYPE(imm, cimm, rs1, op)
+  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ reg_abi_name_raw(cimm) ^ sep() ^ hex_bits_signed_13(imm)
```
jordancarlin: Why is this using `reg_abi_name_raw` for the assembly mapping of `cimm`? Shouldn't it be `hex_bits_signed_5` or something?

jordancarlin: And should this actually be using `b_imm5(cimm)` as the value to print? Based on the LLVM implementation, it looks like the assembly should be using -1 and not 0.

jordancarlin (changes requested):
Several (fairly minor) comments. Also make sure to add the extension to the README and Changelog.

challenger1024:
if use:
```sail
mapping clause assembly = BITYPE(imm, cimm, rs1, op)
  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_signed_5(b_imm5(cimm)) ^ sep() ^ hex_bits_signed_13(imm)
```
we will get:
```bash
Type error:
extensions/Zibi/zibi_insts.sail:48.78-90:
48 |  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_signed_5(b_imm5(cimm)) ^ sep() ^ hex_bits_signed_13(imm)
| b_imm5 is not a union constructor or mapping in mapping-pattern b_imm5(cimm)
```
So the code ended up being written this way.
last commit:
- add zibi extension to README.md and doc/ChangeLog.md;  
- move currentlyEnabled() function to model/extensions/Zibi/zibi_insts.sail
- modify assembly mapping of zibi

model/extensions/Zibi/zibi_insts.sail
```diff
@@ -0,0 +1,51 @@
...
+  then 0b11111
+  else cimm
+
+mapping clause encdec = BITYPE(imm @ 0b0, cimm, rs1, op)
+  <-> imm[11] @ imm[9..4] @ cimm @ encdec_reg(rs1) @ encdec_biop(op) @ imm[3..0] @ imm[10] @ 0b1100011
+  when currentlyEnabled(Ext_Zibi)
+
+$[split op]
+function clause execute BITYPE(imm, cimm, rs1, op) = {
+  let cimmext : xlenbits = sign_extend(b_imm5(cimm));
```
Timmmm: This doesn't seem to match your spec. E.g. if `cimm` is 0b10000` you'll get `0b1111...10000` when the spec says it should be `0b0000...10000`. You want to do this:

```
let cimmext : xlenbits = if cimm == zeros() then ones() else zero_extend(cimm);
```

model/extensions/Zibi/zibi_insts.sail
```diff
@@ -0,0 +1,51 @@
...
+  else RETIRE_SUCCESS
+}
+
+mapping bitype_mnemonic : biop <-> string = {
+  BEQI  <-> "beqi",
+  BNEI  <-> "bnei"
+}
+
+mapping clause assembly = BITYPE(imm, 0b00000, rs1, op)
+  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_signed_5(0b11111) ^ sep() ^ hex_bits_signed_13(imm)
```
Timmmm: Is the assembly really meant to be `0x1F`? Surely `-1`?

challenger1024: Yes, I was mistaken. So how should I print  -1  here?

README.md
```diff
@@ -115,6 +115,7 @@ For booting operating system images, see the information under the
 - RV32I and RV64I base ISAs, v2.1
 - RV32E and RV64E base ISAs, v2.0
 - Zifencei extension for instruction-fetch fence, v2.0
+- Zibi extension for conditional branches with immediate operands, v0.6
```
jordancarlin: Move this down to the experimental extensions section.

model/extensions/Zibi/zibi_insts.sail
pmundkur:
```suggestion
  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ "-1" ^ sep() ^ hex_bits_signed_13(imm)
```
should work.

model/extensions/Zibi/zibi_insts.sail
```diff
@@ -0,0 +1,51 @@
...
+
+mapping encdec_biop : biop <-> bits(3) = {
+  BEQI  <-> 0b010,
+  BNEI  <-> 0b011
+}
+
+function b_imm5(cimm : bits(5)) -> bits(5) =
+  if cimm == 0b00000
+  then 0b11111
+  else cimm
```
pmundkur: This won't be needed after the fix to compute `cimmext` in `execute`.

model/extensions/Zibi/zibi_insts.sail
```diff
@@ -0,0 +1,51 @@
...
+
+mapping bitype_mnemonic : biop <-> string = {
+  BEQI  <-> "beqi",
+  BNEI  <-> "bnei"
+}
+
+mapping clause assembly = BITYPE(imm, 0b00000, rs1, op)
+  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_signed_5(0b11111) ^ sep() ^ hex_bits_signed_13(imm)
+mapping clause assembly = BITYPE(imm, cimm, rs1, op)
+  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_signed_5(cimm) ^ sep() ^ hex_bits_signed_13(imm)
```
pmundkur: Shouldn't this be printed as an unsigned immediate?
```suggestion
  <-> bitype_mnemonic(op) ^ spc() ^ reg_name(rs1) ^ sep() ^ hex_bits_5(cimm) ^ sep() ^ hex_bits_signed_13(imm)
```

model/extensions/Zibi/zibi_insts.sail
Timmmm: Probably just go for `"-1"` for now. The backwards (parsing a string) direction doesn't actually work yet anyway so it doesn't matter too much that it wouldn't parse e.g. `-0x1`.

pmundkur (approved):

Timmmm (approved):
LGTM. I guess we should probably have some kind of policy on how far along the process extensions need to be before they can be merged into this model?

[This page](https://riscv.atlassian.net/wiki/spaces/HOME/pages/16154861/RISC-V+Specs+Under+Development) seems to have a fairly clear 6-step progression (Zibi is at 'Stabilization').

It probably also depends on how large and invasive the code is. I think it's @pmundkur's decision but at least in this case the code is small and neat and doesn't affect anything else, so I'd vote merge.

pmundkur:
Agree that this seems okay to merge.  "Stabilization" seems a reasonable stage to merge small self-contained extensions, more invasive ones could be frozen or ratification ready as the case may be?

PR #1513, opened 2026-01-30
Add Hypervisor 2-Stage Address Translation

nadime15:
This is an initial draft of the two-stage address translation support for the Hypervisor extension and based on #1419  and #1110  (the actual new code is introduced starting from commit 52ba61776b59ef2b8684d74a18573d0d663db867).

The code is still very experimental and I plan to keep improving and changing it. I would appreciate early feedback on the overall direction.

There are two main problems to address.

## 1. Dealing with implicit VS-Stage Page Faults and Exception handling complexity

With two stage translation, exception handling becomes more complicated. Previously it was enough to propagate the faulting VA and the causing exception. Due to virtualization, we now have several relevant addresses that are required to set stval (and htval) correctly.

These include the GVA (Guest Virtual Address) for explicit VS-Stage faults and the GPA (Guest Physical Address) for G-Stage faults or the guest PTE address that triggered an exception during implicit VS-Stage PTE accesses (and maybe even more, I dont see the big picture yet).

This means the exception context must carry more information than before in order to trap correctly.

To pass this information up the call chain, we define a new struct called `Exception_Context` that carries more contextual data. There is also a new PTE_Error entry (`PTW_Implicit_Error)` that propagates the implicit error including the encountered exception during VS-Stage back up the call chain. 

The trap handler is untouched to keep the draft in a manageable size.

## 2. Virtual address definitions introduced by the Hypervisor spec

The Hypervisor spec introduces 34 bit "virtual addresses" (GPA as part of G-Stage) for RV32. I think it's is debatable whether these should be modeled as virtual addresses or treated as physical addresses. This draft treats the address received from the VS-Stage as a physical address and continue working with it directly instead of casting it back to a virtual address.

There is still work to do on typing and refactoring of the TLB lookup, insertion, and flush logic.

Regular one stage address translation (S-Stage) is not affected by these changes and the code works as usual and compiles.


nadime15:
Here are the diagrams from this morning’s meeting. They might be helpful for one or the other.

# 1-Stage/2-Stage Address Translation

X-Stage-Address-Translation

# 2-Stage Address Translation Fault Handling
Hypervisor-Exception-Table

Timmmm (commented):
Ok I've read slightly more of the spec. My first observation is that we don't *really* need to model Sv39x4, etc. separately. Since the only difference is the top two bits of the first VPN (what `satp`/`hgatp` point to), Sv39x4 is actually identical to Sv39 if you just calculate the root page location differently. In other words you can have `translateAddr` work for guest physical addresses with very minor changes something like this:

```
function translateAddr(
  vAddr : virtaddr,
  access : MemoryAccessType(mem_payload),
  translation_type : {Virtual, GuestPhysical},
) -> TR_Result(physaddr, ExceptionType) = {

  ...
  let satp_sxlen = get_satp_or_hgatp(sv_width, translation_type);
  ...

    let base_ppn = satp_to_ppn(satp_sxlen);
    let base_ppn = match translation_type {
      Virtual => base_ppn,
      GuestPhysical => base_ppn + vAddr[sv_width+1 .. sv_width],
    };
```

In other words we can treat Sv39x4 exactly the same as Sv39, if instead of setting the root page to just `satp.PPN` we set it to `hgatp.PPN + (the top two bits of the guest physical address)`.

Then we can totally forget about the whole x4 thing. Does that sound right?

nadime15:
Yes  you are right, we dont actually need the Sv*x entries, but I am not sure it makes sense to get rid of them entirely.

I think your point is that we can effectively squeeze the S-Stage / VS-Stage and G-Stage into a single function (translateAddr), and then derive the correct values based on `translation_type`. PR #1419 already went in that direction by introducing things like `xatp_mode_width`, etc.

(I am a bit confused by the example though, you are using `satp_to_ppn` instead of something like `satp_or_hgatp_to_ppn`, and `vsatp` it is not considered even though its not the same as `satp`. I assume that was just quick pseudocode.)

For now (not 100% sure yet), I would personally prefer to keep the Sv*x entries. If we want to go down the more generic path, I would rather define higher level helper functions, for example, for fetching the base ppn or sv_width (similar to xatp_mode_width), that forwards to the appropriate underlying helpers (get_vsatp, get_satp, get_hgatp etc.) based on the selected mode.

My concern with removing these entries and making everything as “tight” as possible is that it may become slightly harder to understand, especially for someone who is still in the process of learning the spec. But honestly I am not sure either.

@radimkrcmar @trdthg Any thoughts?

Timmmm:
> I assume that was just quick pseudocode.

Yeah exactly; we'll need to rename things to `xatp` or something like that.

> I would rather define higher level helper functions, for example, for fetching the base ppn or sv_width (similar to xatp_mode_width), that forwards to the appropriate underlying helpers (get_vsatp, get_satp, get_hgatp etc.) based on the selected mode.

I think this is a good idea. IMO this would greatly reduce the amount of code and also simplify it a fair bit, which would make it easier to understand. We should add a comment to explain it of course.

Honestly I think the spec would be clearer if it was written like this too. They could have ditched the entire explanation of extra modes and whatnot and just replaced it with a single paragraph:

> To support RV32 34-bit physical addresses, guest physical address translation has a virtual address space four times larger than usual. `hgatp.PPN` points to four root page tables (and must be 16kB-aligned instead of 4-kB aligned), and the two bits above the normal virtual address width select which root page table to use. For example when `hgatp.MODE` selects Sv32, the lowest 32-bits of the guest physical address are translated in the usual way but using the root page table seleected by bits [33:32] of the guest physical address. Similarly if `hgatp.MODE` is configured as Sv39, then bits [40:39] of the guest physical address select the root page table.

Is that not easier to understand than introducing a load of extra modes, address formats, etc?

trdthg:
I think it’s a good idea generally and we should give it a try. But I suspect some of the type issues might be tricky.

```sail
type vpn_bits('v), is_sv_any_mode('v) = bits('v - pagesize_bits)
```

nadime15:
@Timmmm Haha yeah it is actually!

Ok thanks @Timmmm @trdthg, I will take another look and incooperate what we just discussed!

model/sys/vmem.sail
```diff
@@ -473,198 +480,188 @@ mapping hgatp_mode_width : HGATPMode <-> {34, 41, 50, 59} = {
...
+  // When physaddr is wider than virtaddr (RV32: 34 > 32), we check with zero_extend.
+  // When they are equal (RV64: 64 == 64), we use sign_extend for canonical check.
+  // G-stage addresses are guest physical addresses with sign-extended upper bits.
+  // TODO/NOTE: That's a bit ugly, simplify if possible
+  let svAddr = bits_of(addr)[sv_width - 1 .. 0];
+  let valid_addr : bool = match stage {
+    S_Stage  => bits_of(addr) == (if physaddrbits_len > xlen then zero_extend(svAddr) else sign_extend(svAddr)),
+    VS_Stage => bits_of(addr) == (if physaddrbits_len > xlen then zero_extend(svAddr) else sign_extend(svAddr)),
+    G_Stage  => bits_of(addr) == sign_extend(svAddr),
+  };
```
radimkrcmar: I don't follow this logic.  The validity is only defined for virtual address (first stage), and it always sign extends from 39/48/57 bits to 64 bits on RV64, and compares the result.

The output of first stage is 34 or 56 bits.  Bits that are not translatable by the second stage (above 41/50 on Sv39x4/Sv48x4) are required to be 0.
(VS thinks that it addresses a physical address, and it would be very impractical if that address changed based on what hgatp.mode HS picked.)

radimkrcmar:
Btw. how are you thinking to implement HSV/HLV?
Those can execute at M/HS/HU modes, and require two stage translation... Would they be another MemoryAccessType?

nadime15:
> Btw. how are you thinking to implement HSV/HLV?
> Those can execute at M/HS/HU modes, and require two stage translation... Would they be another MemoryAccessType?

PR #1419 addressed this by splitting `translateAddr` into two functions and introducing `translateAddr_pv` (called `translateAddr_eff_priv` in this PR). This new function allows the effective privilege level to be passed explicitly, rather than being computed internally by `translateAddr` (see how HLV is implemented in PR #1419) . As a result, `translateAddr` now simply wraps `translateAddr_eff_priv`.

So the implementation for HLV will just call:

```
translateAddr_pv(
  vaddr,
  Load(Data),
  privLevel_bits(zero_extend(hstatus[SPVP]), 0b1)
)
```

The full error handling happens within the instruction’s execute clause, instead of delegating to vmem_read or vmem_write, which already encapsulate this logic. We currently have a few cases, such as certain atomic instructions where calling translateAddr directly is still needed. So this would actually not change anything besides the fact that we call `translateAddr_eff_priv` instead of `translateAddr` for HSV/HLV.

model/sys/vmem.sail
nadime15: I agree its confusing, thats why I added a TODO. The address validity check in translate_stage is a bit convoluted due to merging S/VS/G stages into one function even though these stages operate under different address types/lengths. The problem is also the transition from GVA to GPA in RV32 from 32bit to 34bit addresses. 

The original code did the canonical check directly on virtaddr:
```
let svAddr = bits_of(vAddr)[sv_width - 1 .. 0];
if bits_of(vAddr) != sign_extend(svAddr)
```

This worked because virtaddr is xlenbits (32 or 64 bits), and comparing these values is straightforward.
The problem now is that `translate_stage` takes physaddr (34 bits on RV32, 64 bits on RV64) to support G-stage GPAs. The wrapper `translate_vs_stage` converts virtaddr to physaddr using `zero_extend`:

On RV32 with Sv32, virtaddr is 32 bits, but physaddr is 34 bits so we zero_extend, which sets upper 2 bits to 0
But sign_extend(svAddr) to 34 bits sets upper 2 bits based on bit 31 so now If bit 31 = 1: zero_extend gives 00, sign_extend gives 11, which would lead to a mismatch.

I don’t fully grasp the big picture yet, but in general it might not be a good idea to squeeze everything into a single function and rely on physical bits. We can have fewer physical bits than GPA (34 bits), so in that case we would already run into limitations and lose bits using translate_stage that uses phys. addresses (I think).

It might be worth it to define a new type only for GPA. 

But to your point, you basically saying in the G_Stage stage we should zero_extend instead of sign_extend.

radimkrcmar: > On RV32 with Sv32, virtaddr is 32 bits, but physaddr is 34 bits so we zero_extend, which sets upper 2 bits to 0
But sign_extend(svAddr) to 34 bits sets upper 2 bits based on bit 31 so now If bit 31 = 1: zero_extend gives 00, sign_extend gives 11, which would lead to a mismatch.

I think passing anything other than XLEN sized type as an input to the first stage should be an internal sail error, because it shouldn't have happened.  The sign extension just compares whether the upper bits of the XLEN sized address are in the proper format (equal to the highest bit that is used in the Sv mode translation).

The output of the translation is then a 34/56 bit address (regardless of what the actual physical address width is).

> But to your point, you basically saying in the G_Stage stage we should zero_extend instead of sign_extend.

Yes.

model/sys/vmem.sail
```diff
@@ -327,36 +473,253 @@ mapping satp_mode_width : SATPMode <-> {32, 39, 48, 57} = {
...
+  if not(valid_addr) then return Err(PTW_Invalid_Addr(), init_ext_ptw);
+
+  let asid     = get_asid(stage);
+  let vmid     = get_vmid(stage);
+  let mxr      = get_mxr(stage);
+  let do_sum   = get_sum(stage);
+
+  // Step 1 of VATP.
+  let base_ppn = xatp_to_ppn(stage);
+
```
Timmmm: What I was trying to say is, change the function signature to this:

```
private function translate_stage(
  // Zero extend for non-VS translations.
  addr   : bits(xlen + 2),
  access : MemoryAccessType(mem_payload),
  priv   : Privilege,
  stage  : TranslationStage
) -> TR_Result(physaddr, PTW_Error) = {
```

And then:

```
let extra_x4_bits = svAddr[sv_width + 1 .. sv_width]; // sv_width is still always 32, 39, 48, 56.
assert(stage != VS_Stage ==> extra_x4_bits == zeros());
let base_ppn = xatp_to_ppn(stage) + zero_extend(extra_x4_bits);
```

Also then I think you only need two values in `TranslationStage`: `VS_Stage` and `S_Stage` because from that point on there's no difference between S and GS (at least none I know about!).

You can also get rid of all the other changes in `vmem_types.sail` I think.

Does that make sense? I might not have explained super well... (Or I might have misunderstood!)

nadime15: Ahh ok, I get it now, that’s quite smart! But I’m not so sure this works, I’m probably missing something here.

1. I think we need to keep `G_Stage` in order to know which xatp register to access. G_Stage sets the effective address to User Mode (see the eff_priv in translate_stage), and we need to know that we are performing G-Stage translation in case of an implicit VS-Stage access (see pt_walk). Edit: You also need G-Stage to know whether you have consider VMID in TLB look ups, I believe.

2. Assuming we keep G_Stage, sv_width won’t return 32 in the case of RV32, right? It will return 34 bits, so this becomes a problem. A potential workaround would be to just do svAddr[xlen + 1 .. xlen].

3. Changing the length of base_ppn won’t work, at least as of now, because base_ppn is typed to be ppn_bits bits wide (24 or 46). So I think it’s a bit misleading to expand ppn_base to suddenly be 26 bits or 48 bits, and this potentially leads to even more non-trivial changes.

4. Even though I think this is quite smart, it’s not really obvious. In the pt_walk function we actually compute the PTE_Addr and dissect the different VPNs, etc. I think it’s probably not straightforward that you basically enlarged the page table base address much earlier in the call chain.

Another issue, unrelated to this specific one, which I forgot to mention, is that we also need to return either a 32 or 34 bit address depending on the stage we are currently in, or 64 bit. So the return type needs to be modified as well.

Timmmm: 1. Good point! Alternatively you could pass `xatp` in as a parameter I guess. Up to you.

2. Not sure I understand this. Do you mean `xatp_mode_width`? I think that should always return 32, 39, 48, or 56.

3. I don't think we need to change the *width* of `ppn_base` - we just need to add a 2 bit value to it (not concatenate).

4. Yeah even though it is a bit of a sneaky trick, it is sooooo much simpler I think we should still do it (assuming it works and there isn't some catch I haven't thought of).

5. Are you sure? Isn't it always 34 or 64 bits?

radimkrcmar: I think the trick breaks TLB lookups, since they ignore the root page table address and would therefore alias addresses with different upper two bits.   It seems that we're trying too hard to work around the type system, rather than using it...

nadime15: > I think the trick breaks TLB lookups, since they ignore the root page table address and would therefore alias addresses with different upper two bits. It seems that we're trying too hard to work around the type system, rather than using it...

Yeah exactly.. I ran into this while implementing the discussed solution. I have a commit I am about to push that shows the progress and the issue that @radimkrcmar has mentionned

nadime15: See https://github.com/riscv/sail-riscv/commit/f8990eaedd3f8ee13f8c2f9df0355edb8b52ac42 and specifically `match_TLB_Entry()`.

We only consider up to 36 bits for the VPN, which won’t work in the case of a GPA address. I think this trick really simplifies the code, but it seems we run into issues here.

> It seems that we're trying too hard to work around the type system, rather than using it...

Any suggestions? In the previous commit we at least made sure that we had the full VPN. It was a bit more complicated, but I think it was more obvious what was actually going on.

Another option would be to expand translate() and pass these two additional bits to the function as well, or do something like G_Stage(0bXX) and pass them down the call chain. Personally, I’m still in favor of the original approach that explicitly extends the VPN, combined with using `gpabits` that were introduced by https://github.com/riscv/sail-riscv/commit/f8990eaedd3f8ee13f8c2f9df0355edb8b52ac42.

I introduced `gpabits` that are either 34 bits or 64 bits long for VS/S/G-stage access. I think this addresses your comment (https://github.com/riscv/sail-riscv/pull/1513#discussion_r2777261063), @radimkrcmar. I also think this is cleaner than using `bits(xlen + 2)`, which feels a bit weird, especially when xlen = 64.

radimkrcmar: I think the extra complications of the previous versions could have been hidden in nicely named helpers, so it seemed acceptable to me.  Losing the information that we're operating with a completely different mode seems worse to me...  I think we'd ideally be expressing mode and deriving sw_width from it, but refactoring might not be worth the effort unless you have to complicate the code too much to handle Bare modes.

I'm not sure about the need for gpabits -- couldn't we express the translation output width as depending on sw_width?
```
Bare: 'v -> 'v, // We have to handle Bare mode specially, since sw_width is not unique: 32 and 34 could be Sv too
SvAB: 32/{39,48,57} -> 34/56,
SvABx4: 34/{41,50,49} -> 34/56,
``` 
First there could be a function to perform input validation from an arbitrary bit width:
```
validate_input_address (address : bits('v), sv_width : int('w)) -> Result(bits('w))
// perform zero/sign expansion based on mode if 'v > 'w
// Also possible to type check for internal sail errors, e.g. first stage input is should always be xlen
```
Which would feed into the translation
```
translate_valid_address(address : bits('w), ...) -> Result(bits(mode_output_bits('w)))
```
For two stages, the output width would be an input to another round of validation and translation, so there wouldn't really be any difference between the stages.
At the end, we'd validate the output into a physical address or change the physical address interface to accept any number of bits and do the validation itself.

Btw. I think a lot the of the finer design depends on TLB.  Do we want to support all architecturally different RISC-V TLB implementations, or are we just using TLB to make sail a bit faster?
The TLB can cache the cumulative result of all stages (satp+hgatp+mmpt+pmp+pma), cache each stage separately, or not cache some stages at all.  (Our current TLB seems to separate satp and hgatp, and does not cache pmp/pma.)
RISC-V can also cache invalid translations, which we don't do, and would require further changes.

pmundkur: The TLB is really needed for Linux boot, i.e. make the model faster.  It would be nice to have enough just enough detail for to implement the VMA fences, to get reasonable simulation speed, but we don't need every possible TLB variation.

Timmmm: Hmm that is slightly awkward. Maybe we use this trick for `pt_walk` and below, and stuff above that is aware of the extra bits?

nadime15: I gave it another try (https://github.com/riscv/sail-riscv/pull/1513/commits/df046c52276225994cc2d54e1dd7c08e29ea1603) but it's not really what we discussed this morning :/

I was thinking about it again, and I think it’s simpler and much more understandable to slightly change the `VPN` type and reintroduce hypervisor specific terms (Sv32x4, Sv39x4, etc.) to accommodate the increased GVA widths and handle this explicitly. The changes are relatively small, but they make it clearer to the reader what is actually going on.

We still keep a fairly generic interface (`translate_stage()`), which works regardless of the current stage we are in. We could also move the code that computes the new PTE_addr (currently part of `translate()`) into a separate function and add a few comments to make the flow more obvious.

model/sys/vmem.sail
```diff
@@ -79,25 +111,42 @@ function pt_walk(
...
   // Convert to a physical address (34 bits for RV32, 64 bits to RV64).
   // The assertion is required so Sail knows that we can't have a
   // pte_addr of 56 bits and physaddr of 34 bits (because you can't
   // use Sv39+ on RV32).
-  assert(sv_width == 32 | xlen == 64);
-  let pte_addr = Physaddr(zero_extend(pte_addr));
+  assert('v == 32 | 'v == 34 | xlen == 64);
+
+  // Translate PTE address if performing a VS-stage access
+  let pte_addr : physaddr = match translate_PTE_addr(zero_extend(pte_addr), access, stage, ext_ptw) {
```
radimkrcmar: The PTEs are translated with a different access type than the original translation.
Access to PTEs during first stage translation is implicit read.

We can drop the access parameter from translate_PTE_addr.
(The translation could be under pte_read if that made anything simpler.)

model/sys/vmem.sail
```diff
@@ -124,15 +173,15 @@ function pt_walk(
...
+            pt_walk(sv_width, vpn, access, priv, mxr, do_sum, ppn, level - 1, global, stage, ext_ptw)
           else {
               // level 0 PTE, but contains a pointer instead of a leaf
               ptw_fail_callback(PTW_Invalid_PTE(), level, bits_of(pte_addr));
               Err(PTW_Invalid_PTE(), ext_ptw)
             }
         } else {
           // Leaf PTE (Step 5 of VATP).
-          let ppn_size_bits = if 'v == 32 then 10 else 9;
+          let ppn_size_bits = if 'v == 32 | 'v == 34 then 10 else 9;
```
radimkrcmar: `vpn_i_size`?

model/sys/vmem.sail
```diff
@@ -79,25 +111,42 @@ function pt_walk(
...
   ext_ptw,
 ) = {
   ptw_start_callback(zero_extend(vpn), access, (priv, ()));
 
-  // Extract the PPN component for this level; 10 bits on Sv32, otherwise 9.
-  let 'vpn_i_size = if 'v == 32 then 10 else 9;
-  let vpn_i = vpn[(level + 1) * vpn_i_size - 1 .. level * vpn_i_size];
-  let 'log_pte_size_bytes = if 'v == 32 then 2 else 3;
+  let 'vpn_i_size = if ('v == 32 | 'v == 34) then 10 else 9;
+  let 'log_pte_size_bytes = if ('v == 32 | 'v == 34) then 2 else 3;
```
radimkrcmar: I think it would be nicer to derive the parameters from each other, so it's a bit more obvious what their connection is, and I'd probably prefer a style where the if/else based on mode is implemented outside of the function, e.g.:
```
type log_pte_size_bytes('v : Int), is_sv_or_svx4_mode('v) = if 'v == 32 | 'v == 34 then 2 else 3
type vpn_i_size('v : Int), is_sv_or_svx4_mode('v) = pagesize_bits - log_pte_size_bytes('v)
```
(A helper for `'v == 32 | 'v == 34` that distinguishes xlen=32 from xlen=64 should be used in many other places too.)

model/core/vmem_types.sail
```diff
@@ -16,17 +18,19 @@ type is_sv_mode('v) -> Bool = 'v in {32, 39, 48, 57}
 //   Sv48 = range(0, 3)
 //   Sv57 = range(0, 4)
 //
-type level_range('v), is_sv_mode('v) =
-  range(0, (if 'v == 32 then 1 else (if 'v == 39 then 2 else (if 'v == 48 then 3 else 4))))
+type level_range('v : Int), is_sv_or_svx4_mode('v) = range(0, if 'v == 32 | 'v == 34 then 1
+                                                              else if 'v == 39 | 'v == 41 then 2
+                                                              else if 'v == 48 | 'v == 50 then 3
+                                                              else 4)
```
radimkrcmar:
```suggestion
type root_level('v : Int), is_sv_or_svx4_mode('v) = if 'v == 32 | 'v == 34 then 1
                                                    else if 'v == 39 | 'v == 41 then 2
                                                    else if 'v == 48 | 'v == 50 then 3
                                                    else 4
type level_range('v : Int), is_sv_or_svx4_mode('v) = range(0, root_level('v))
```
The `root_level` helper can be used for `initial_level` and `root_level` later on.

model/sys/vmem.sail
```diff
@@ -79,25 +111,42 @@ function pt_walk(
...
+                   else if ('v == 48 | 'v == 50) then 3
+                   else 4;
 
-  // Address of PTE in page table. This is 34 bits for Sv32, otherwise 56 bits.
-  let pte_addr = pt_base @ vpn_i @ zeros('log_pte_size_bytes);
+  // Svx4 root level has 2 extra bits
+  let is_svx4_root = ('v == 34 | 'v == 41 | 'v == 50 | 'v == 59) & (level == root_level);
 
+  let vpn_lo = level * vpn_i_size;
+  let vpn_hi = if is_svx4_root
```
radimkrcmar:
```suggestion
  let vpn_lo = level * vpn_i_size;
  let vpn_hi = if is_svx4_mode('v) & level == root_level('v)
```
Computing the page table offsets is still awkward, but at least contained to a single expression.

radimkrcmar:
```suggestion
  let vpn_lo = level * vpn_i_size;
  let vpn_hi = if is_svx4_mode('v) & level == root_level('v)
```
Computing the page table offsets is still awkward, but at least contained to a single expression.

model/sys/vmem.sail
```diff
@@ -327,36 +468,247 @@ mapping satp_mode_width : SATPMode <-> {32, 39, 48, 57} = {
...
+  then Ok(truncated)
+  else Err(PTW_Invalid_Addr())
+}
+
+val validate_g_input : forall 'v, is_svx4_mode('v). (
+  gpabits,
+  int('v),
+) -> result(bits('v), PTW_Error)
+
+function validate_g_input(addr, sv_width) = {
```
radimkrcmar: The validation doesn't really benefit from having both `vs` and `g` variants and can be purely a function of address (arbitrary bit width) and mode (approximated by sw_width for us) into sw_width bits or error. 
This means the validation can be called from translate_stage, also simplifying both `vs` and `g` stage helpers.

Something like
```
function validate_address forall 'a 'v, is_sv_or_svx4_mode('v). (addr : bits('a), sv_width : int('v)) = {
  let sv_wide_address = if 'v < 'a then truncate(addr, 'v) else zero_extend(addr, 'v);
  let extended_address = if is_sv_mode('v) then sign_extend(sv_sized_address) else zero_extend(sv_wide_address);
  if addr == extended_address
  then Ok(sv_wide_address)
  else (PTW_Invalid_Addr())
}
```

(It should be possible to get rid of the `vs` and `g` stage helpers as well, with a similar approach.)

radimkrcmar (changes requested):

model/sys/vmem.sail
nadime15: @radimkrcmar, I have implemented most of the suggestions. One limitation is that Sail type constraints like `is_sv_mode` and `is_svx4_mode` cannot be called as regular functions inside function bodies like you suggested, they can only be used in type definitions and forall clauses. I tried to work around this by defining `is_sv32_mode` as a type constraint used consistently across all type definitions. I think the dependencies between these should be clearer now. See commit c9f52bd for more details on what I changed.

model/core/hypervisor_regs.sail
```diff
@@ -0,0 +1,263 @@
...
+    // Force 16 KiB alignment with PPN[1:0] = 0b00
+    let h = Mk_Hgatp64(v[Mode] @ 0b00 @ v[VMID] @ v[PPN][43 .. 2] @ 0b00);
+    match hgatpMode_of_bits(arch, h[Mode]) {
+      None() => prev_value,
+      Some(mode) => match mode {
+        HBare  if currentlyEnabled(Ext_Svbare) => h.bits,
+        Sv39x4 if currentlyEnabled(Ext_Sv39)   => h.bits,
+        Sv48x4 if currentlyEnabled(Ext_Sv48)   => h.bits,
+        Sv57x4 if currentlyEnabled(Ext_Sv57)   => h.bits,
+        _ => prev_value,
```
radimkrcmar, line 235: hgatp doesn't do atomic updates (because writing a partial state doesn't completely screw the writer like in the case of satp):
```
A write to hgatp with an unsupported MODE value is not ignored as it is for satp.
Instead, the fields of hgatp are WARL in the normal way, when so indicated.
```

model/sys/vmem.sail
radimkrcmar: Looks better, thanks, and I'm sorry for misleading you about the type system.  I could bikeshed a lot more, but I'll try to only comment on bugs from now on, as not to delay the merge too much. :)

model/core/sys_regs.sail
```diff
@@ -784,6 +931,37 @@ private function lower_mstatus(m : Mstatus) -> Sstatus = {
...
+              extStatus_of_bits(s[XS]) == Dirty |
+              extStatus_of_bits(s[VS]) == Dirty;
+  [o with
+    SD   = bool_to_bit(dirty),
+    // We don't support dynamic changes to UXL.
+    // UXL  = s[UXL],
+    // SDT = if currentlyEnabled(Ext_Ssdbltrp) then s[SDT] else 0b0,
+    MXR  = s[MXR],
+    SUM  = s[SUM],
+    XS   = s[XS],
```
radimkrcmar, line 947: XS is read-only, and SPELP is missing.

I think it's a bit unfortunate that we're diverging in the FS/VS field handling as well.
Is there a reason to not share the legalizer with sstatus?

radimkrcmar (changes requested):

model/sys/vmem.sail
nadime15: Don't worry, I thought this would work too, but I had to realize it does not :smile:

model/core/sys_regs.sail
nadime15: Yeah good point, I think I can use the legalizer for sstatus, i just have to reorganize the code a bit. I will update CSR related changes in #1110.

I also need to update this PR with the changes I made in #1110

model/core/hypervisor_regs.sail
nadime15: Addressed, see [here](https://github.com/trdthg/sail-riscv/blob/711074b0b396e4d943d2698586d9b19dc8dc06dc/model/core/sys_regs.sail#L928-L964)

model/core/sys_regs.sail
nadime15: Addressed, see [here](https://github.com/trdthg/sail-riscv/blob/8e08af8300e6d9b042ab0ef15507d35978a4b82a/model/core/sys_regs.sail#L851-L854)

jordancarlin:

PR #1523, opened 2026-02-03
Update test suite.

pmundkur:
This now uses the 2026-02-03 release of `sail-riscv-tests`.


nadime15 (approved):

nadime15:
Not really. If you visit the repository’s [website](https://github.com/riscv-software-src/sail-riscv-tests), you can see the latest commits and [check](https://github.com/riscv-software-src/sail-riscv-tests/pull/30)
 whether any new tests have been added. We could try to create a list by comparing releases, but this quickly becomes non-trivial, for example, the vector test suite added support for `Zvfh`, which didn’t create new files but instead made existing files larger (which is rare are but still).

PR #1524, opened 2026-02-04
Add per field `mutable_misa` config options

### LINKED ISSUE #1517: Additional configurability needed for `misa` writability
Right now, we have a global `writable_misa` config option that makes the entire `misa` CSR read-only or read-write. Implementations can make each individual extension bit read-only or read-write, so we need additional granularity here. For example, Spike makes the U and S bits read-only but the rest of the extension bits read-write.

UDB uses `MUTABLE_MISA_A`, `MUTABLE_MISA_B`, etc. as the parameters that control this. We probably need to do something similar.

jordancarlin:
This is one approach to fix #1517 by adding a parameter per extension that has a field in `misa` indicating whether the bit is writable. I think this is easier to use than a bitmask for the whole `misa` CSR, but that is another approach that we should consider.


allenjbaum:
I really really really prefer the bitmap approach; it is a mechanism that
can be used over and over, and if you're writing an entire CSR, that's how
you want to do it instead of having to decompose it into N separate
mechanisms.
That single mechanism will probably handle the great majority of CSRs. MISA
is just one example; there are hundreds of other CSRs that will need to do
the same thing. Do't special case it.

model/core/sys_regs.sail
```diff
@@ -101,24 +98,24 @@ val ext_veto_disable_C : unit -> bool
 
 private function legalize_misa(m : Misa, v : xlenbits) -> Misa = {
   let v = Mk_Misa(v);
-  // Suppress updates to MISA if MISA is not writable or if by disabling C next PC would become misaligned or an extension vetoes
-  if   not(sys_enable_writable_misa) | (v[C] == 0b0 & (nextPC[1] == 0b1 | ext_veto_disable_C()))
+  // Suppress updates to MISA if by disabling C next PC would become misaligned or an extension vetoes
+  if   (v[C] == 0b0 & (nextPC[1] == 0b1 | ext_veto_disable_C()))
```
Timmmm: If C is immutable and hard-coded to 1, and you write 0, should this still ignore the entire write?

Timmmm (commented):
Yeah I'm still on the fence. I kind of think that although at first glance CSRs *seem* nice and uniform and a generic mechanism would work, if you look closer there are all sorts of horrible special snowflakes like `seed`, `sstatus`, `senvcfg`, etc.

Can we see what the alternative would look like for this PR though?

I think something like this?

```
private function legalize_misa(m : Misa, v : xlenbits) -> Misa = {
  let misa_writable_mask : xlenbits = config ...;
  let v = Mk_Misa(v & misa_writable_mask | m.bits & ~(misa_writable_mask));
  ...
```

allenjbaum:
Some CSRs are in  read_only category (predefiined for the entire CSR by
address)
For CSRs not in that category, if there are bits/fields that are readOnly,
and (possibly) others that are not, the writes to readOnly fields are
ignored

I am a bit puzzled about your snowflake example.
Sstatus is a "restricted view" CSR, a copy of Mstatus with some bits
treated as RdOnly0.
I think that lends itself to a mask/value approach, with the mask having a
few more bits set from the status mask (except the 2bit UXL field, FS, and
VS fields are each enums with the same values taken from Mstatus.
The general approach is mask/value for the, CSR as a whole, and multibit RW
fields defined either by hi/lo ranges, or enum (for widths <=6b)

The trickiest part is not so much legalizing the write, but legalizing the
WARL definition as a subset of the spec definition with the
spec'ed constraints
(e.g. bits in a RW field that are defined as RdOnly in implementation are
OK, or a RW field is only large <param> bits wide.

jordancarlin:
This is becoming an issue for ACTs. @pmundkur @nadime15, any further thoughts on this approach? I can update the PR if we are willing to move forward with it. I prefer this to a bitmask, and based on discussions in the Parameters SIG, it sounds like the plan is to use individual parameters when the fields in a CSR are unrelated (which seems to apply to MISA).

model/core/sys_regs.sail
```diff
@@ -101,24 +98,24 @@ val ext_veto_disable_C : unit -> bool
...
   then m
   else [m with
-    A = if hartSupports(Ext_A) then v[A] else 0b0,
-    B = if hartSupports(Ext_B) then v[B] else 0b0,
-    C = if hartSupports(Ext_C) then v[C] else 0b0,
-    D = if hartSupports(Ext_D) & v[F] == 0b1 then v[D] else 0b0,
-    F = if hartSupports(Ext_F) then v[F] else 0b0,
+    A = if hartSupports(Ext_A) then v[A] else if config extensions.A.mutable_misa then m[A] else 0b0,
```
pmundkur: I'm wondering if this should be something like 
```suggestion
    A = if hartSupports(Ext_A) then { if config extensions.A.mutable_misa  then v[A] else m[A] } else 0b0,
```

pmundkur (commented):
I too prefer something like this to a bitmask.  I think we could move forward with this.  I'd add a validation check to disallow setting `extension.X.mutable_misa` to `true` if `extension.X.supported` is false.

model/core/sys_regs.sail
nadime15: Yeah good catch. It needs to otherwise its a bug.

model/core/sys_regs.sail
```diff
@@ -101,24 +98,24 @@ val ext_veto_disable_C : unit -> bool
...
     E = bool_to_bit(base_E_enabled),
-    M = if hartSupports(Ext_M) then v[M] else 0b0,
+    M = if hartSupports(Ext_M) then v[M] else if config extensions.M.mutable_misa then m[M] else 0b0,
     // Q = if hartSupports(Ext_Q) then v[Q] else 0b0, TODO: Not supported yet
-    S = if hartSupports(Ext_S) & v[U] == 0b1 then v[S] else 0b0,
-    U = if hartSupports(Ext_U) then v[U] else 0b0,
-    V = if hartSupports(Ext_V) & v[F] == 0b1 & v[D] == 0b1 then v[V] else 0b0,
+    S = if hartSupports(Ext_S) & v[U] == 0b1 then v[S] else if config extensions.S.mutable_misa then m[S] else 0b0,
+    U = if hartSupports(Ext_U) then v[U] else if config extensions.U.mutable_misa then m[U] else 0b0,
+    V = if hartSupports(Ext_V) & v[F] == 0b1 & v[D] == 0b1 then v[V] else if config extensions.V.mutable_misa then m[V] else 0b0,
```
nadime15: misa.D/S/V cannot actually be set individually under certain circumstances. For example, even if `misa.V` is mutable, immutable misa.F and misa.D effectively prevent `misa.V` from being changed (because the final value depends on what is being read and now the actual fixed value of misa.F and misa.D).

nadime15 (commented):
I agree with @pmundkur. This seems fine.

model/core/sys_regs.sail
jordancarlin: If `misa.F` is not mutable and is set to read-only one, why wouldn't you still be able to change `misa.D` from 1 to 0?

model/core/sys_regs.sail
jordancarlin: Oops. Yes.

model/core/sys_regs.sail
nadime15: I think that’s a different example that you describe which happens to work? Thats why I meant that under certain circumstances, some values cannot be set.

There are other examples where this causes an issue. For example, if F is immutable and set to 1, while D is mutable and currently 0, we could write csrw misa to set D to 1 (with all other bits zero). The guard checks v[F] == 0b1, which is false because the write only carries the D bit. Now D stays 0.

The thing is, the code should not decide what D becomes based on the current input (v) when dealing with immutable bits.

jordancarlin: Ah. Yes. I agree. It needs to do it based on the actual value that will be read back at the end.

jordancarlin:
I'm not sure why the Rocq CI keeps rerunning on this PR. It seems to be on its 3rd attempt, but I haven't rerun it since it is supposed to be failing.

model/postlude/validate_config.sail
```diff
@@ -15,6 +15,57 @@ private function check_privs() -> bool = {
...
+  valid = valid & check_mutable_misa_supported("U", hartSupports(Ext_U), config extensions.U.mutable_misa : bool);
+  valid = valid & check_mutable_misa_supported("V", hartSupports(Ext_V), config extensions.V.mutable_misa : bool);
+  valid = valid & check_mutable_misa_supported("H", hartSupports(Ext_H), config extensions.H.mutable_misa : bool);
+
+  if (config extensions.F.mutable_misa : bool) &
+     ((hartSupports(Ext_D) & not(config extensions.D.mutable_misa : bool)) |
+      (hartSupports(Ext_V) & not(config extensions.V.mutable_misa : bool)))
+  then {
+    valid = false;
+    print_endline("`extensions.F.mutable_misa` is true but `extensions.D.mutable_misa` and/or `extensions.V.mutable_misa` is false: mutable F requires mutable D and V when they are supported.");
```
pmundkur: Perhaps something like this if not too awkward?
```suggestion
    print_endline("`extensions.F.mutable_misa` is true but `extensions.D.mutable_misa` and/or `extensions.V.mutable_misa` is false: mutable misa.F requires mutable misa.D and misa.V when D and V are supported respectively.");
```

model/postlude/validate_config.sail
```diff
@@ -15,6 +15,57 @@ private function check_privs() -> bool = {
...
+  };
+  if (config extensions.D.mutable_misa : bool) & hartSupports(Ext_V) & not(config extensions.V.mutable_misa : bool)
+  then {
+    valid = false;
+    print_endline("`extensions.D.mutable_misa` is true but `extensions.V.mutable_misa` is false: mutable D requires mutable V if V is supported.");
```
pmundkur:
```suggestion
    print_endline("`extensions.D.mutable_misa` is true but `extensions.V.mutable_misa` is false: mutable misa.D requires mutable misa.V if V is supported.");
```

model/postlude/validate_config.sail
```diff
@@ -15,6 +15,57 @@ private function check_privs() -> bool = {
...
+  };
+  if (config extensions.U.mutable_misa : bool) &
+     ((hartSupports(Ext_S) & not(config extensions.S.mutable_misa : bool)) |
+      (hartSupports(Ext_H) & not(config extensions.H.mutable_misa : bool)))
+  then {
+    valid = false;
+    print_endline("`extensions.U.mutable_misa` is true but `extensions.S.mutable_misa` and/or `extensions.H.mutable_misa` is false: mutable U requires mutable S and H when they are supported.");
```
pmundkur:
```suggestion
    print_endline("`extensions.U.mutable_misa` is true but `extensions.S.mutable_misa` and/or `extensions.H.mutable_misa` is false: mutable misa.U requires mutable misa.S and misa.H when S and H are supported respectively.");
```

model/postlude/validate_config.sail
```diff
@@ -15,6 +15,57 @@ private function check_privs() -> bool = {
...
+  };
+  if (config extensions.S.mutable_misa : bool) & hartSupports(Ext_H) & not(config extensions.H.mutable_misa : bool)
+  then {
+    valid = false;
+    print_endline("`extensions.S.mutable_misa` is true but `extensions.H.mutable_misa` is false: mutable S requires mutable H if H is supported.");
```
pmundkur:
```suggestion
    print_endline("`extensions.S.mutable_misa` is true but `extensions.H.mutable_misa` is false: mutable misa.S requires mutable misa.H if H is supported.");
```

model/core/sys_regs.sail
```diff
@@ -122,24 +119,37 @@ val ext_veto_disable_C : unit -> bool
 
 private function legalize_misa(m : Misa, v : xlenbits) -> Misa = {
   let v = Mk_Misa(v);
-  // Suppress updates to MISA if MISA is not writable or if by disabling C next PC would become misaligned or an extension vetoes
-  if   not(sys_enable_writable_misa) | (v[C] == 0b0 & (nextPC[1] == 0b1 | ext_veto_disable_C()))
+
+  // Determine new values for each field taking dependencies into account
+  let A = if hartSupports(Ext_A) then { if config extensions.A.mutable_misa then v[A] else m[A] } else 0b0;
+  let B = if hartSupports(Ext_B) then { if config extensions.B.mutable_misa then v[B] else m[B] } else 0b0;
+  let C = if hartSupports(Ext_C) then { if config extensions.C.mutable_misa then v[C] else m[C] } else 0b0;
```
pmundkur, line 126: I think this should retain the `nextPC[1] == 0b1` and `ext_veto_disable_C()` checks to prevent disabling `C`.  The latter is needed for CHERI; I don't remember what breaks without the `nextPC` check but best to keep it.

pmundkur (changes requested):

model/core/sys_regs.sail
jordancarlin: I don't think I changed that check? It still does that check before changing `misa` and rejects the entire write if the check fails.

The `nextPC` check is required by the spec because you cannot have run an instruction aligned to a 2-byte boundary if you don't support C.

pmundkur: Ah, I missed that it just moved down.

pmundkur (approved):

model/core/sys_regs.sail
```diff
@@ -122,24 +119,37 @@ val ext_veto_disable_C : unit -> bool
...
+  let M = if hartSupports(Ext_M) then { if config extensions.M.mutable_misa then v[M] else m[M] } else 0b0;
+  let F = if hartSupports(Ext_F) then { if config extensions.F.mutable_misa then v[F] else m[F] } else 0b0;
+  let D = if hartSupports(Ext_D) & F == 0b1 then { if config extensions.D.mutable_misa then v[D] else m[D] } else 0b0;
+  // let Q = if hartSupports(Ext_Q) & F == 0b1 & D == 0b1 then { if config extensions.Q.mutable_misa then v[Q] else m[Q] } else 0b0;
+  let V = if hartSupports(Ext_V) & F == 0b1 & D == 0b1 then { if config extensions.V.mutable_misa then v[V] else m[V] } else 0b0;
+  let U = if hartSupports(Ext_U) then { if config extensions.U.mutable_misa then v[U] else m[U] } else 0b0;
+  let S = if hartSupports(Ext_S) & U == 0b1 then { if config extensions.S.mutable_misa then v[S] else m[S] } else 0b0;
+  let H = if hartSupports(Ext_H) & S == 0b1 & U == 0b1 then { if config extensions.H.mutable_misa then v[H] else m[H] } else 0b0;
+  // Suppress updates to MISA if by disabling C next PC would become misaligned or an extension vetoes
+  if   (v[C] == 0b0 & (nextPC[1] == 0b1 | ext_veto_disable_C()))
```
nadime15: I think this should be:
```suggestion
  if   (C == 0b0 & m[C] == 0b1 & (nextPC[1] == 0b1 | ext_veto_disable_C()))
```

model/postlude/validate_config.sail
```diff
@@ -15,6 +15,53 @@ private function check_privs() -> bool = {
...
 }
 
+private function check_mutable_misa_supported(ext_name : string, supported : bool, mutable : bool) -> bool = {
+  if mutable & not(supported) then {
+    print_endline("`extensions." ^ ext_name ^ ".mutable_misa` is true but " ^ ext_name ^ " is not supported: `mutable_misa` must be false for unsupported extensions.");
+    false
+  } else {
+    true
+  }
+}
```
nadime15: Would it be better to remove `check_mutable_misa_supported()` and all calls to it?
```suggestion
```
If someone disables, say, hypervisor, the whole check will fail. I think this might be a bit too constraining. `legalize_misa` already forces the actual bit to 0 anyway.

On top of that, I think the relationships between the C extension and the other extensions are quite complicated. For example if someone disables Zcd and but keeps D the user won't know why check_mutable_misa_config() fails and has to understand somehow the complex relationship between extensions.

But that's your call. I’m fine with either approach.

nadime15 (commented):
This needs a changelog entry mentioning that the writable_misa option was removed and replaced with individual options for each misa bitfield field.

model/postlude/validate_config.sail
jordancarlin: I could go either way. @pmundkur, any opinion?

pmundkur: Agree with @nadime15 ; lets remove it.  We really need to have a more declarative approach to specifying the dependencies between extensions.

jordancarlin: Updated (and agreed that we need to figure out a better way to express dependencies).

pmundkur (approved):

model/postlude/validate_config.sail
nadime15: Removing `check_mutable_misa_supported()` is fine, but I don't think we should remove all the other checks that were introduced in the validator. I can't find the exact commit anymore, but I mean the checks in this change:

https://github.com/riscv/sail-riscv/compare/dd475507644ab2ea5383684c638946e7e5c6dc4a..987f0bbc3b73cb70237299083350795240bb9968

The problem is now that `reset_misa()` sets the bitfields to 1 for every supported extension. For example, say, we support both F and D, but misa.F is mutable while misa.D is immutable. If we then try to clear F and set D, misa.D can end up being cleared as well, even though it is immutable and was set to 1 during reset.

nadime15 (commented):
We still need to add the changelog entry.

Issue #1527, opened 2026-02-04
`fcvt.s.bf16` returns the incorrect result for improperly NaN-Boxed inputs


jordancarlin:
On an RV64IFD_Zfbfmin processor, an input of `0xffffffff00000000` (which is not properly NaN-Boxed so should be interpreted as the canonical QNaN) returns `0xFFFFFFFF7E000000` instead of the expected single-precision canonical QNaN (`0xffffffff7fc00000`). Spike returns the expected `0xffffffff7fc00000`.

This was discovered using the ACT4 tests. A simplified testcase is below that fits into the first-party test framework.

```
.global main
main:
    # Save return address in temporary register we're not using.
    mv t6, ra

    # Load improperly NaN-boxed bf16 value into f1
    #if __riscv_xlen == 64
    li t0, 0xffffffff00000000
    fmv.d.x f1, t0

    # Convert to bf16 and back to integer
    fcvt.s.bf16 f2, f1

    # Check result
    fmv.x.d t1, f2
    li t2, 0xffffffff7fc00000
    bne t1, t2, fail
    #endif

pass:
    li a0, 0
    mv ra, t6
    ret
fail:
    li a0, 1
    mv ra, t6
    ret
```

nadime15:
Ok so I had a look at that, seems like in this case the FP16 canonical QNaN is returned instead the BF16 canonical QNaN which is 0x7fc0. 

A will create a PR with a fix shortly.

PR #1528, opened 2026-02-05
Fix fcvt.s.bf16 NaN-boxing returning wrong canonical QNaN

nadime15:
rF_BF16 was using the generic nan_unbox which returns the FP16 canonical
QNaN (0x7e00) for invalid inputs. BF16 has a different format than FP16
and requires its own canonical QNaN (0x7fc0).

Added canonical_NaN_BF16 and updated rF_BF16 to use it directly instead
of the generic nan_unbox.

Additionally, change how Clang is installed in CI by removing the standalone update-alternatives
approach because the default Ubuntu package manager installs a static symlink at /usr/bin/clang
that the alternatives system cannot overwrite. Since the system refuses to replace a file it doesn't
already "own," the environment was staying locked to the older version despite Clang 21 being registered.
Now, we install a newer version of Clang separately and simply pass the the path to the installed Clang version
to CMake as a Clang compiler flag. This way, CMake uses the specified version if the install-clang flag
in the CI workflow is set to true.

Fixes https://github.com/riscv/sail-riscv/issues/1527

Co-authored-by: Jordan Carlin <jordanmcarlin@gmail.com>

### DIFF
--- a/model/extensions/FD/fdext_regs.sail
+++ b/model/extensions/FD/fdext_regs.sail
@@ -27,6 +27,7 @@ function canonical_NaN(n) =
     128 => 0b0 @ ones(15) @ 0b1 @ zeros(111),
   }
 
+let canonical_NaN_BF16 : bits(16) = 0x7fc0
 // For backwards compatibility with code that existed before the
 // generic version above.
 function canonical_NaN_H() -> bits(16)  = canonical_NaN(16)
@@ -210,7 +211,12 @@ function wF_bits(Fregidx(i) : fregidx, data: flenbits) -> unit = {
 overload F = {rF_bits, wF_bits, rF, wF}
 
 function rF_BF16(i : fregidx) -> bits(16) = {
-  nan_unbox(F(i))
+  // This is the same as `nan_unbox(F(i))` except that
+  // it returns the correct canonical NaN for BF16.
+  let v = F(i);
+  if v[flen - 1 .. 16] == ones()
+  then v[15 .. 0]
+  else canonical_NaN_BF16
 }
 
 function wF_BF16(i : fregidx, data : bits(16)) -> unit = {
--- a/test/first_party/src/test_bf16_nan_boxing.S
+++ b/test/first_party/src/test_bf16_nan_boxing.S
@@ -1,0 +1,46 @@
+.global main
+main:
+#if __riscv_xlen == 64
+    # Load improperly NaN-boxed bf16 value into f1
+    li t0, 0xffffffff00000000
+    fmv.d.x f1, t0
+
+    # Convert to bf16 and back to fp32
+    fcvt.s.bf16 f2, f1
+
+    # Check result
+    fmv.x.d t1, f2
+    li t2, 0xffffffff7fc00000
+    bne t1, t2, fail
+
+#else
+    # Build improperly NaN-boxed bf16 value on stack
+    addi sp, sp, -8
+
+    li t0, 0x00000000
+    sw t0, 0(sp)
+
+    li t0, 0xffffffff
+    sw t0, 4(sp)
+
+    # Load improperly NaN-boxed value from stack
+    fld f1, 0(sp)
+    addi sp, sp, 8
+
+    # Convert to bf16 and back to fp32
+    fcvt.s.bf16 f2, f1
+
+    # Check result
+    fmv.x.w t1, f2
+    li t2, 0x7fc00000
+    bne t1, t2, fail
+
+#endif
+
+pass:
+    li a0, 0
+    ret
+
+fail:
+    li a0, 1
+    ret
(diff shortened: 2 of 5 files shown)


nadime15:
I included the test @jordancarlin  provided, but CI will fail, when compiling with Clang I get the following error:

`clang: error: invalid arch name 'rv32gcv_zfbfmin', requires '-menable-experimental-extensions' for experimental extension 'zfbfmin'`

Should I add the experimental flag, or is this something that would be resolved with a newer Clang version (if that even makes sense)? Any thoughts?

pmundkur:
Perhaps we should use a newer version of clang for the tests using https://github.com/egor-tensin/setup-clang?  https://llvm.org/docs/RISCVUsage.html says Zfbfmin is not experimental anymore.

test/first_party/src/test_bf16_nan_boxing.S
```diff
@@ -0,0 +1,27 @@
+.global main
+main:
+    # Save return address in temporary register we're not using.
+    mv t6, ra
```
jordancarlin: Doesn't look like this is needed.

test/first_party/src/test_bf16_nan_boxing.S
```diff
@@ -0,0 +1,27 @@
...
+
+    # Load improperly NaN-boxed bf16 value into f1
+    #if __riscv_xlen == 64
+    li t0, 0xffffffff00000000
```
jordancarlin: Might be worth adding an rv32 version of this test as well.

model/extensions/FD/fdext_regs.sail
```diff
@@ -210,7 +211,10 @@ function wF_bits(Fregidx(i) : fregidx, data: flenbits) -> unit = {
 overload F = {rF_bits, wF_bits, rF, wF}
 
 function rF_BF16(i : fregidx) -> bits(16) = {
-  nan_unbox(F(i))
+  let v = F(i);
```
Timmmm, line 216: Probably worth adding

```
// This is the same as `nan_unbox(F(i))` except that
// it returns the correct cannonical NaN for BF16.
```

nadime15:
> Perhaps we should use a newer version of clang for the tests using https://github.com/egor-tensin/setup-clang? https://llvm.org/docs/RISCVUsage.html says Zfbfmin is not experimental anymore.

Seems like this works only for Ubuntu and Windows? I think we will run into the same issue for macOS again :/

jordancarlin:
> > Perhaps we should use a newer version of clang for the tests using https://github.com/egor-tensin/setup-clang? https://llvm.org/docs/RISCVUsage.html says Zfbfmin is not experimental anymore.
> 
> Seems like this works only for Ubuntu and Windows? I think we will run into the same issue for macOS again :/

The latest version of clang should be available in homebrew for macOS.

jordancarlin:
Would be good to get this merged before the 0.10 release. If we can't sort out #1532 beforehand, it might make sense to merge this without the test and then add the test in a subsequent PR (we can open it right away so it doesn't get forgotten).

pmundkur:
Let's discuss this tomorrow.

model/extensions/FD/fdext_regs.sail
```diff
@@ -27,6 +27,7 @@ function canonical_NaN(n) =
     128 => 0b0 @ ones(15) @ 0b1 @ zeros(111),
   }
 
+function canonical_NaN_BF16() -> bits(16) = 0x7fc0
```
pmundkur: Isn't this simpler just as a variable?
```suggestion
let canonical_NaN_BF16 : bits(16) = 0x7fc0
```

model/extensions/FD/fdext_regs.sail
```diff
@@ -210,7 +211,12 @@ function wF_bits(Fregidx(i) : fregidx, data: flenbits) -> unit = {
 overload F = {rF_bits, wF_bits, rF, wF}
 
 function rF_BF16(i : fregidx) -> bits(16) = {
-  nan_unbox(F(i))
+  // This is the same as `nan_unbox(F(i))` except that
+  // it returns the correct canonical NaN for BF16.
+  let v = F(i);
+  if v[flen - 1 .. 16] == ones()
+  then v[15 .. 0]
+  else canonical_NaN_BF16()
```
pmundkur:
```suggestion
  else canonical_NaN_BF16
```

model/extensions/FD/fdext_regs.sail
nadime15: Simpler is relative. It was mainly for consistency...

nadime15:
I had to rewrite some parts, `update-alternatives` did not work because it kept restoring the system clang, even though we linked against a newer version. I think this happens because the alternatives system still “owns” /usr/bin/clang and will enforce whatever version is registered. 

Now we pass the versioned clang directly to CMake works no matter what /usr/bin/clang points to, which I think is cleaner anyway.

Timmmm:
Can you add a description of the Clang changes to the PR message (I would do it but I don't fully understand it)?

PR #1529, opened 2026-02-05
Move CLI options into struct

Timmmm:
`riscv_sim.cpp` currently suffers from "everything is global". As the first step of fixing that, this moves the CLI options into their own struct and passes it around.


arichardson (approved):

pmundkur:
This got removed from the merge queue due to a conflict.  Hopefully this means #1493 is now handled!

Timmmm:
I don't think it implies that unfortunately - I doubt it's possible to configure it to merge something with a merge conflict because it can't actually do the merge.

PR #1531, opened 2026-02-05
Return instead of exit() in `inner_main()`

Timmmm:
`exit()` is a bit of a code smell and we don't need it here.


arichardson (approved):

Timmmm:
> main() currently returns 1, you could replace this with EXIT_FAILURE as well.

Done

pmundkur (approved):

jordancarlin (approved):

PR #1532, opened 2026-02-05
Install newer Clang version on Linux CI for newer extension support

nadime15:
Ubuntu runners ship with an older Clang that doesn't support newer extensions like zfbfmin. By default, we install Clang Version 21 for Ubuntu runners.

### DIFF
--- a/.github/actions/sail-setup/action.yml
+++ b/.github/actions/sail-setup/action.yml
@@ -35,6 +35,7 @@ runs:
         sudo ./llvm.sh $CLANG_VERSION
         sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-$CLANG_VERSION 100
         sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-$CLANG_VERSION 100
+        sudo update-alternatives --install /usr/bin/lld lld /usr/bin/lld-$CLANG_VERSION 100
 
     - name: Install packages (macOS)
       if: runner.os == 'macOS'


nadime15:
I tried using https://github.com/egor-tensin/setup-clang, but its not very straightforward and I ran into a couple of issues:

1. setup-clang installs Clang into a non-standard location and prepends it to $PATH, but CMake’s find_program() searches system paths before $PATH. As a result, it still finds the older Clang version that ships with the runner. A workaround is required to locate the newly installed Clang, otherwise the default runner Clang is used.

2. Rocq jobs fail because the action depends on PowerShell (that setup-clang uses), which is not available in the container environment used by Rocq.

Given these issues, its much simpler to install Clang manually. Once newer runners are available, we can potentially upgrade the runner image and remove the manual installation step.

.github/workflows/ci.yml
```diff
@@ -27,6 +29,17 @@ jobs:
...
+      # Install newer clang manually because current Ubuntu runners don't have
+      # a clang version that supports newer RISC-V extensions.
+      - name: Install Clang (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          wget https://apt.llvm.org/llvm.sh
+          chmod +x llvm.sh
+          sudo ./llvm.sh $CLANG_VERSION
+          sudo ln -sf /usr/bin/clang-$CLANG_VERSION /usr/bin/clang
+          sudo ln -sf /usr/bin/clang++-$CLANG_VERSION /usr/bin/clang++
```
jordancarlin: Should we move this into the `.github/actions/sail-setup` action? That is where we are installing all of the other dependencies.

pmundkur: I'm wondering if forcefully linking just `/usr/bin/clang` to the new version and leaving any other clang toolchain binaries won't create an inconsistent clang install.  @nadime15 and I discussed this, and we were wondering if it would it be better to install in a new directory and tell CMake to find it there? 

It  would be also be nice to pass an option to `sail-setup` to do this install of latest clang; there's no reason to do this install if one is not building tests (e.g. for the Lean/Rocq builds).

Timmmm: I think it's probably better to use https://apt.llvm.org/ and `update-alternatives` if needed.

Timmmm: Oh wait I'm blind. Still I think `update-alternatives` will do the symlinking in an official way.

nadime15: What about https://github.com/riscv/sail-riscv/pull/1532/commits/25fe69f4fc560a228f75646400ae7f8652667b6d ? Additionally, if needed, I could add a new variable that we pass to CMake to search $PATH first, while by default it would continue searching system paths so users aren’t suprised.

nadime15: Ok, the Rocq container comes with almost nothing. I can barely run anything, including llvm.sh, which relies on lsb_release.

nadime15: Third version: https://github.com/riscv/sail-riscv/commit/e64bae79b0098f11f07873b537f20243d85c0217

.github/workflows/ci.yml
```diff
@@ -69,6 +70,8 @@ jobs:
 
       - name: Common setup
         uses: ./.github/actions/sail-setup
+        with:
+          install-clang: "true"
```
jordancarlin: Why is this needed for the Linux build? It doesn’t use clang.

.github/actions/sail-setup/action.yml
```diff
@@ -19,6 +23,19 @@ runs:
...
+      shell: bash
+      run: |
+        # Install newer clang manually because current Ubuntu runners don't have
+        # a clang version that supports newer RISC-V extensions.
+        curl -O https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        CLANG_VERSION=21
+        sudo ./llvm.sh $CLANG_VERSION
+        sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-$CLANG_VERSION 100
+        sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-$CLANG_VERSION 100
```
jordancarlin, line 37: Do we need to worry about the version of lld too?

bacam:
You have probably realised this already, but the Rocq job doesn't need to compile any C code, it just checks that the model builds, so the clang version doesn't matter for it.

nadime15:
I think it would be cleaner if the Rocq workflow ran directly on the latest Ubuntu runner, similar to the Lean workflow, and installed its dependencies within the Rocq workflow itself (I am pretty sure the container can be configured to match latest ubuntu?).

Right now, every workflow goes through action.yaml, and whenever something is added to sail-setup there is is risk of breaking the Rocq workflow because the container is missing even basic tools. This just makes it harder to maintain and we have to use workarounds like this one (assuming it gets merged as-is).

jordancarlin (approved):

jordancarlin:
> Is there a reason the `release` workflow doesn't use `sail-setup`?

`sail-setup` is only written to be compatible with Ubuntu and macOS right now. The release workflow uses a Rocky 8 container. We could (and maybe should) update it to work for RHEL based distros too so that it can be used in the release workflow. Seems like that should be a separate PR though.

pmundkur:
Agreed that it should be a separate PR.  Since the release workflow also runs tests, I'm just worried tests might fail there due to an older compiler.

jordancarlin:
Come to think of it, this method of installing Clang won't work on Rocky 8—it is exclusive to the `apt` package manager.

pmundkur:
Does `copr` work on Rocky?
https://copr.fedorainfracloud.org/coprs/g/fedora-llvm-team/llvm-snapshots/
We really should add Rocky to the regular CI.  It just feels dicey to use a different platform for release than the ones tested in CI.  Things could go wrong at  release time that weren't caught in CI.

nadime15:
The macOS runner used a different Clang version all the time than the current runners. Maybe even ubuntu-arm and the regular Ubuntu version, and probably the Rocky container too. So all runners have different version. I don’t think it’s worth the pain to try to get the same Clang version on the Rocky container now. No matter what, there will be differences anyway.

jordancarlin:
> The macOS runner used a different Clang version all the time than the current runners. Maybe even ubuntu-arm and the regular Ubuntu version, and probably the Rocky container too. So all runners have different version. I don’t think it’s worth the pain to try to get the same Clang version on the Rocky container now. No matter what, there will be differences anyway.

The issue is that we run the tests as part of the release job, so if the Clang version in the Rocky container is too old, the tests will fail to compile, and the release pipeline will fail.

nadime15:
It seems like the release job uses gcc to compile these tests?

PR #1533, opened 2026-02-06
Add configuration option for reserved mtvec[MODE] values

challenger1024:
Some mtvec[MODE] values are currently reserved. Add an option to configure what happens if a reserved value is written.

### DIFF
--- a/model/core/platform_config.sail
+++ b/model/core/platform_config.sail
@@ -81,22 +81,10 @@ enum XenvcfgCbieReservedBehavior = {
   Xenvcfg_ClearPermissions,
 }
 
-mapping XenvcfgCbieReservedBehavior_str : XenvcfgCbieReservedBehavior <-> string = {
-  Xenvcfg_Fatal <-> "Xenvcfg_Fatal",
-  Xenvcfg_ClearPermissions <-> "Xenvcfg_ClearPermissions",
-}
-overload to_str = {XenvcfgCbieReservedBehavior_str}
-
 enum XtvecModeReservedBehavior = {
   Xtvec_Fatal,
   Xtvec_Ignore,
 }
-
-mapping XtvecModeReservedBehavior_str : XtvecModeReservedBehavior <-> string = {
-  Xtvec_Fatal <-> "Xtvec_Fatal",
-  Xtvec_Ignore <-> "Xtvec_Ignore",
-}
-overload to_str = {XtvecModeReservedBehavior_str}
 
 enum RV32ZdinxOddRegisterReservedBehavior = {
   Zdinx_Fatal,


model/core/sys_regs.sail
```diff
@@ -578,12 +578,15 @@ register mtvec : Mtvec  // Trap Vector
...
-   TV_Direct => v,
-   TV_Vector => v,
-   _         => [v with Mode = o[Mode]]
- }
+  let v = Mk_Mtvec(v);
+  match (trapVectorMode_of_bits(v[Mode])) {
+    TV_Direct => v,
+    TV_Vector => v,
+    _         => match mtvec_mode_reserved_behavior {
+      Mtvec_Fatal => reserved_behavior("Tried to write a reserved value to the MODE field of mtvec."),
```
Timmmm: Probably could add the value that was written?

model/core/platform_config.sail
```diff
@@ -104,6 +104,17 @@ mapping XenvcfgCbieReservedBehavior_str : XenvcfgCbieReservedBehavior <-> string
...
 overload to_str = {XenvcfgCbieReservedBehavior_str}
 
+enum MtvecModeReservedBehavior = {
+  Mtvec_Fatal,
+  Mtvec_UseOld,
+}
+
+mapping MtvecModeReservedBehavior_str : MtvecModeReservedBehavior <-> string = {
+  Mtvec_Fatal <-> "Mtvec_Fatal",
+  Mtvec_UseOld <-> "Mtvec_UseOld",
```
Timmmm: Maybe `Mtvec_Ignore`? UseOld sounds a bit weird to me. Very minor though.

model/core/sys_regs.sail
```diff
@@ -578,12 +578,15 @@ register mtvec : Mtvec  // Trap Vector
...
 // PUBLIC: invoked from set_{ms}tvec() [exceptions/sys_exceptions.sail]
 function legalize_tvec(o : Mtvec, v : xlenbits) -> Mtvec = {
- let v = Mk_Mtvec(v);
- match (trapVectorMode_of_bits(v[Mode])) {
-   TV_Direct => v,
-   TV_Vector => v,
-   _         => [v with Mode = o[Mode]]
- }
+  let v = Mk_Mtvec(v);
+  match (trapVectorMode_of_bits(v[Mode])) {
```
```suggestion
  match trapVectorMode_of_bits(v[Mode]) {
```

model/core/platform_config.sail
pmundkur: Are these string mappings used anywhere?

allenjbaum:
My Sail semantics is weak at best, but are you defining a specific
illegal->legal mapping for writing an illegal value to mtvec (UseOld)?
IT is a 2 bit fWAL ield , and that isn't the only possibility that an
implementation is allowed (it could keep the old value, ignore only the
write to bit 1, force mallest legal value, or force largest)

Timmmm:
Yeah it's defining the particular behaviour that the model already had hard-coded. We should probably decide how exactly we want to deal with all the WARL fields before adding more.

challenger1024:
I have changed  UseOld  to  Ignore .
@Timmmm

model/core/platform_config.sail
```diff
@@ -81,6 +81,23 @@ enum XenvcfgCbieReservedBehavior = {
   Xenvcfg_ClearPermissions,
 }
 
+mapping XenvcfgCbieReservedBehavior_str : XenvcfgCbieReservedBehavior <-> string = {
+  Xenvcfg_Fatal <-> "Xenvcfg_Fatal",
+  Xenvcfg_ClearPermissions <-> "Xenvcfg_ClearPermissions",
+}
+overload to_str = {XenvcfgCbieReservedBehavior_str}
```
Timmmm: I think this was removed in `master`. Can you remove it, and `XtvecModeReservedBehavior_str` below?

Timmmm (changes requested):

jordancarlin (approved):

PR #1534, opened 2026-02-06
Add --config-override flag for configuration overrides

### LINKED ISSUE #806: Test can't stop if enavle `-C` option
When -C (disable-rvc) is enabled, the test falls into an infinite loop

- cmd: `./build/c_emulator/riscv_sim_rv64d -C ./build/test/first_party/rv64d_test_empty.S.elf > log.txt`
- test asm:
  ```
  .global main
  main:
     ret
  ```
 
log

```
Running file ./build/test/first_party/rv64d_test_empty.S.elf.
ELF Entry @ 0x80002040
CSR mstatus <- 0x0000000A00000000 (input: 0x0000000000000000)
mem[X,0x0000000000001000] -> 0x0297
mem[X,0x0000000000001002] -> 0x0000
[0] [M]: 0x0000000000001000 (0x00000297) auipc t0, 0x0
x5 <- 0x0000000000001000
mem[X,0x0000000000001004] -> 0x8593
mem[X,0x0000000000001006] -> 0x0202
[1] [M]: 0x0000000000001004 (0x02028593) addi a1, t0, 0x20
x11 <- 0x0000000000001020
mem[X,0x0000000000001008] -> 0x2573
mem[X,0x000000000000100A] -> 0xF140
[2] [M]: 0x0000000000001008 (0xF1402573) csrrs a0, mhartid, zero
CSR mhartid -> 0x0000000000000000
x10 <- 0x0000000000000000
mem[X,0x000000000000100C] -> 0xB283
mem[X,0x000000000000100E] -> 0x0182
[3] [M]: 0x000000000000100C (0x0182B283) ld t0, 0x18(t0)
mem[R,0x0000000000001018] -> 0x0000000080002040
x5 <- 0x0000000080002040
mem[X,0x0000000000001010] -> 0x8067
mem[X,0x0000000000001012] -> 0x0002
[4] [M]: 0x0000000000001010 (0x00028067) jalr zero, 0x0(t0)
mem[X,0x0000000080002040] -> 0x006F
mem[X,0x0000000080002042] -> 0x0D40
[5] [M]: 0x0000000080002040 (0x0D40006F) jal zero, 0xd4
mem[X,0x0000000080002114] -> 0x4081
[6] [M]: 0x0000000080002114 (0x4081) c.illegal 0x4081
trapping from M to M to handle illegal-instruction
handling exc#0x02 at priv M with tval 0x0000000000000000
CSR mstatus <- 0x0000000A00001800
within_phys_mem: 0x0000000000000000 not within phys-mem:
  plat_rom_base: 0x0000000000001000
  plat_rom_size: 0x0000000000001000
  plat_ram_base: 0x0000000080000000
  plat_ram_size: 0x0000000080000000
trapping from M to M to handle fetch-access-fault
handling exc#0x01 at priv M with tval 0x0000000000000000
CSR mstatus <- 0x0000000A00001800
within_phys_mem: 0x0000000000000000 not within phys-mem:
  plat_rom_base: 0x0000000000001000
  plat_rom_size: 0x0000000000001000
  plat_ram_base: 0x0000000080000000
  plat_ram_size: 0x0000000080000000
```

trdthg:
This adds a `--config-override` option to the simulator to override parts of the base `--config` (or the built-in default). This is especially useful for first-party tests, some of which need the config to be slightly tweaked, for example `Zclsd` and `Zcf` are exclusive.

Fixes #806


c_emulator/riscv_sim.cpp
```diff
@@ -114,6 +116,25 @@ std::vector<uint8_t> read_file(const std::string &file_path) {
...
 }
 
+std::string read_file_to_string(const std::string &path) {
+  std::ifstream in(path);
+  std::ostringstream ss;
+  ss << in.rdbuf();
+  return ss.str();
+}
+
+void deep_merge_json(jsoncons::json &base, const jsoncons::json &override) {
```
pmundkur: It would be good to have a comment here about what this is trying to do.  Also, `override` is recognized by C++ syntax highlighters as a keyword and looks strange in the editor, so perhaps use `json_override` or similar?

c_emulator/riscv_sim.cpp
```diff
@@ -175,6 +197,10 @@ static CLIOptions parse_cli(int argc, char **argv) {
   app.add_option("--terminal-log", opts.term_log, "Terminal log output file")->option_text("<file>");
   app.add_option("--test-signature", opts.sig_file, "Test signature file")->option_text("<file>");
   app.add_option("--config", opts.config_file, "Configuration file")->check(CLI::ExistingFile)->option_text("<file>");
+  app.add_option("--config-override", opts.config_overrides, "Configuration override file")
```
pmundkur: We probably need some more documentation for this.  Either in the README or some other usage manual.   Here too, perhaps add some information using `app.usage()` or `app.footer()` (I think, I don't remember if these are useful) to explain a little more?

c_emulator/config_utils.cpp
```diff
@@ -105,6 +96,22 @@ void validate_config_schema(const std::string &conf_file) {
...
+      validate_config_schema_string(get_default_config());
+    } else {
+      std::ifstream is(source);
+      std::ostringstream os;
+      os << is.rdbuf();
+      validate_config_schema_string(os.str());
+    }
+  } catch (const std::exception &ex) {
+    std::string source = source.empty() ? "default configuration" : "configuration in " + source;
     throw std::runtime_error("Schema conformance check failed for the " + source + ".");
```
Timmmm: Doesn't this lose the `ex.what()`, e.g. `Cannot parse configuration: ...`?

c_emulator/config_utils.cpp
```diff
@@ -105,6 +96,22 @@ void validate_config_schema(const std::string &conf_file) {
...
+
+void validate_config_schema(const std::string &source) {
+  try {
+    if (source.empty()) {
+      validate_config_schema_string(get_default_config());
+    } else {
+      std::ifstream is(source);
+      std::ostringstream os;
+      os << is.rdbuf();
+      validate_config_schema_string(os.str());
```
Timmmm: You made a `read_file_to_string` function which could be used here. Maybe we should move that function and also the existing `read_file` function into some kind of `file_utils.cpp/h` file?

c_emulator/riscv_sim.cpp
```diff
@@ -114,6 +116,25 @@ std::vector<uint8_t> read_file(const std::string &file_path) {
   return {std::istreambuf_iterator<char>(instream), std::istreambuf_iterator<char>()};
 }
 
+std::string read_file_to_string(const std::string &path) {
+  std::ifstream in(path);
```
Timmmm: I think we need to check if this succeeds, like

```
if (!in) throw ...
```

Otherwise it silently returns an empty string. Also it looks like I also missed this in `read_file()` actually.

c_emulator/riscv_sim.cpp
```diff
@@ -592,16 +618,28 @@ int inner_main(int argc, char **argv) {
...
+    config_json = jsoncons::json::parse(get_default_config());
   }
 
+  for (const auto &override_path : opts.config_overrides) {
+    jsoncons::json override_item = jsoncons::json::parse(read_file_to_string(override_path));
+    deep_merge_json(config_json, override_item);
+  }
+
+  std::ostringstream os;
+  os << jsoncons::pretty_print(config_json);
```
Timmmm: I don't think you need to pretty_print - you can just `os << config_json`.

We may run into an issue with errors though. If Sail's JSON parser detects an error it's going to give line numbers in this generated code which is a shame. Maybe we should probably pass the original string in unmodified if `--config-override` is unused so at least in the common case the line numbers work?

On the other hand if we do that we might get weird behaviour where stuff works until you add `--config-override`.

Definitely a disadvantage of doing this in the emulator binary rather than externally. But I do also like the convenience.

c_emulator/riscv_sim.cpp
trdthg: I slightly expanded the description of this option.

For the footer, I took the style from manpage. Take `man cat` as an example, there is an `EXAMPLES` section that lists some common usages, which I think is really nice:

```
EXAMPLES
       cat f - g
              Output f's contents, then standard input, then g's contents.

       cat    Copy standard input to standard output.
```

But CLI11’s default footer formatter will strip all whitespace, so I had to use other ways to represent indentation:

```
EXAMPLES: 
1. sail_riscv_sim test.elf 
-> Run a RISC-V ELF using the default configuration (rv64d_v256_e64.json). 

2. sail_riscv_sim --config base.json test.elf 
-> Run a RISC-V ELF using the specified configuration file (base.json). 

3. sail_riscv_sim --config base.json --config-override A.json --config-override 
B.json test.elf 
-> Run with base.json and overrides; (B.json > A.json > base.json).
```

I looked into CLI11 a bit more, since v2.6 (we are currently at v2.5), there is an option `app.get_formatter()->enable_footer_formatting(false);` that allows us to disable the default footer formatting https://github.com/CLIUtils/CLI11/pull/1150
I think I can send a PR to upgrade CLI11, and also switch it to FetchContent if you all agree.

Timmmm: I think the question is more about the override algorithm itself. Merging JSON isn't self-describing; there are multiple ways to do it. Something like:

> JSON objects are merged by replacing/adding fields from the override config. If a given field is an object in the base and override then instead of replacing the field entirely the merging process recurses into it. All other field types (including arrays) are simply replaced.

c_emulator/config_utils.cpp
```diff
@@ -105,6 +97,21 @@ void validate_config_schema(const std::string &conf_file) {
...
+  try {
+    if (file_path.empty()) {
+      validate_config_schema_string(get_default_config());
+    } else {
+      validate_config_schema_string(read_file_to_string(file_path));
+    }
+  } catch (const std::exception &ex) {
+    std::string source_desc = file_path.empty() ? "default configuration" : "configuration in " + file_path;
+    throw std::runtime_error(
+      "Schema conformance check failed for the " + source_desc + ": " + std::string(ex.what()) + "."
```
Timmmm: This will now give

> Schema conformance check failed for the default configuration: Schema conformance check failed for the configuration..

I dunno what the best solution to that is tbh. We should probably drop the `.`s at least. Maybe just change the message on line 100 to something like `Errors during schema validation`? Then it's like:

> Schema conformance check failed for the default configuration: Errors during schema validation

?

Timmmm (commented):
Looks pretty good to me. I'm still not sure what we do about error location reporting though.

Timmmm:
> I'm still not sure what we do about error location reporting though.

For the record, discussed in the meeting. We decided just to go with it. In theory you shouldn't ever get a parse error from Sail because we always parse with jsoncons first. That may not end up being 100% true if there are subtle differences between the parsers, or maybe things that Sail checks that the schema validation doesn't, but fingers crossed...

trdthg:
I've made a clear distinction between parse and validate errors. 

The code first run `json::parse` on every input config file, and then outputs the filename and location info for any caught errors.

```
$ ./build/c_emulator/sail_riscv_sim --config ./rv32d_v128_e32.json --config-override ./test_A.json test.elf
JSON parse error in override file ./test_A.json:
Extra comma at line 4 and column 4
```

Now it doesn't matter whether `os << config_json` uses `pretty_print` or not, because any formatting issues will have already cause an early return. 

Finally, we perform the validation check. `validate_config_schema` now accepts a `jsoncon::json`, we can guarantee it will only throw schema-related exceptions.

```
$ ./build/c_emulator/sail_riscv_sim --config ./rv32d_v128_e32.json --config-override ./test_A.json test.elf
Error: Schema conformance check failed for merged configuration from ./rv32d_v128_e32.json, ./test_A.json:
- /base: Required property 'xlen' not found.
```

test/first_party/CMakeLists.txt
```diff
@@ -106,59 +120,61 @@ foreach (xlen IN ITEMS 32 64)
...
-        add_custom_target(build_${arch}_${test_source} ALL DEPENDS ${elf} ${config})
-
-        add_test(
-            NAME "first_party_${arch}_${test_source}"
-            COMMAND $<TARGET_FILE:sail_riscv_sim> --config ${config} ${elf}
-        )
-    endforeach()
+    set(elf "${arch}_${test_source}.elf")
+    set(override_dir "${CMAKE_CURRENT_BINARY_DIR}/overrides")
+    file(MAKE_DIRECTORY "${override_dir}")
```
Timmmm: Doesn't look like this is used?

```suggestion
```

test/first_party/CMakeLists.txt
```diff
@@ -106,59 +120,61 @@ foreach (xlen IN ITEMS 32 64)
...
+    set(override_file "${CMAKE_CURRENT_BINARY_DIR}/${arch}_${test_source}.override.json")
+    file(GENERATE OUTPUT "${override_file}" CONTENT "${override_config}")
```
Timmmm: Why not just store the override in an optional file instead of putting it in CMake and then writing it to a file? I think that would be a bit simpler. You can use an empty string to indicate that it doesn't need an override file. Or have a `no_overrides.json` file containing `{}`.

test/first_party/CMakeLists.txt
```diff
@@ -87,17 +87,31 @@ set(common_deps
...
-    "test_misaligned_vector_register_groups.S"
-    "test_pmp_access.c"
-    "test_phys_perms_on_failing_sc.S"
-)
-
+set(tests)
+macro(add_first_party_test name march xlen override)
+    list(APPEND tests "${name}" "${march}" "${xlen}" "${override}")
+endmacro()
+macro(add_common name)
```
Timmmm: I'm sure there must be a better name than `add_common`. Maybe

* `add_common` -> `add_first_party_test`
* `add_first_party_test` -> `add_first_party_test_with_config`

Timmmm (changes requested):

test/first_party/CMakeLists.txt
trdthg: `no_overrides.json` looks nice to me.

Timmmm:
We may want to add a proper man page style document for the simulator at this point, rather than just adding everything to the readme.

doc/ChangeLog.md
```diff
@@ -7,6 +7,8 @@
 
 - A configuration parameter to specify the size of the reservation set
   for Zalrsc atomics has been added.
+- A `--config-override` cli option to specify an additional JSON configuration
+  file that overrides the corresponding fields in default or custom configuration.
```
pmundkur:
```suggestion
- A `--config-override` cli option has been added to specify one or more additional JSON configuration
  files that override the corresponding fields in a configuration.
```

README.md
```diff
@@ -111,6 +111,11 @@ use a custom configuration, save the default configuration into a
 file, edit it as needed, and pass it to the simulator using the
 `--config` option.
 
+To override only a small subset of options while using the default configuration
+or a custom configuration file as a base, the `--config-override` option can be
+used. This option allows an additional JSON configuration file to be specified,
```
pmundkur:
```suggestion
used. This option allows one or more additional JSON configuration files to be specified,
```

PR #1535, opened 2026-02-06
Make `config` settings non-global

Timmmm:
Move the `config_..` settings into `CLIOptions` and `ModelImpl` so they are no longer global.


Timmmm:
Note there is still quite a bit of clean-up to do around this code:

1. `flush_logs()` is only called if `opts.config_print_instr` is set, both other options can print things.
2. Some `config_..` options go to `ModelImpl` and some go to `log_callbacks` and it feels kind of arbitrary.
3. The `config_` prefix is quite redundant.
4. We still have the weird `--print-step` option which simply adds a newline between steps. This was for compatibility with people who parse the logs and should be removed when we have JSON output.

But I will do these in future PRs.

jordancarlin (approved):

PR #1536, opened 2026-02-06
Make global `g_model` instance local

Timmmm:
Reduce the scope of `g_model`. This is better code style and will also make supporting multi-hart systems easier.


Timmmm:
~~This depends on #1535 (stacked PRs when Github?).~~

I think the next step after this will be to create a new class `Hart` or something like that (names welcome!) and move a lot of the methods into there. The `Hart` will include `ModelImpl` as a private member, and can be the library interface to the model (with hopefully a stable-ish API).

Issue #1514, opened 2026-01-31
Unable to pretty-print the spec


WojciechMula:
I use the following invocation:

```shell
$ sail --project ${RISCV_SAIL}/model/riscv.sail_project   \
     --output-sail core                                 \
     --config ${config}                                 \
```

where `RISCV_SAIL` and `config` point to proper dir/file. The freshest master fails with the following error:

```
Type error:
/home/wojtek/github/sail-riscv/model/core/types.sail:100.0-104.1:
100 |union cacheop = {
    |^----------------
104 |}
    |^
    | Types are not well-formed within this type definition. Note that recursive types are forbidden.
    | 
    | Caused by /home/wojtek/github/sail-riscv/model/core/types.sail:101.16-27:
    | 101 |  CB_manage   : cbop_zicbom,
    |     |                ^---------^
    |     | Well-formedness check failed for type
    |     | 
    |     | Caused by /home/wojtek/github/sail-riscv/model/core/types.sail:101.16-27:
    |     | 101 |  CB_manage   : cbop_zicbom,
    |     |     |                ^---------^
    |     |     | Undefined type cbop_zicbom
```

The last working commit is `b1d732827175562acf95d192312bb1d21de232cd`.

I have the latest sail installed, 0.20.1.

Alasdair:
Hmm looks like something isn't quite strict enough in the `sail_project` file. I'll fix that.

You can work around this by adding the needed module names to the CLI, like:
```
sail --output-sail --project <project> --config <config> A_types Zicbom_types core
```

For pretty-printing the files, `sail --fmt` might be a better choice as it doesn't do any pre-processing (and it doesn't need to type check) and has a smarter pretty printing approach.

WojciechMula:
Thank you for prompt answer, will try your workaround.

PR #1537, opened 2026-02-06
Strengthen some before statements in sail_project file

Alasdair:
Some of the extensions were using `before core` to make sure their types are available in the core type definitions. This is however not intended - the well-formedness checking of types in Sail was slightly too weak, and was only checking that the types exist, but not checking if they have been explicitly brought into scope via a `requires` statement. I plan on fixing this in Sail, but this PR will need to be merged first to avoid breaking the model.

Note that this doesn't have to be done for the new Zibi extension, as that only needs the type to be used by the `instruction` clause inside `Zibi`.

### DIFF
--- a/model/riscv.sail_project
+++ b/model/riscv.sail_project
@@ -7,7 +7,7 @@ prelude {
 }
 
 core {
-  requires prelude, Zicbop_types
+  requires prelude, A_types, Zicbop_types, Zicbom_types
 
   files
     core/xlen.sail,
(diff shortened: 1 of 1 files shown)


Alasdair:
This is the cause of the error in https://github.com/riscv/sail-riscv/issues/1514

Alasdair:
Currently the only way to avoid that would be to break the offending types into scattered definitions, but that is quite heavyweight.

I do have some ideas for optional module dependencies that could extend existing definitions to avoid this.

pmundkur:
It would help to document the semantics and intended usage of `before`, `after`, and `requires`, wrt type and other dependencies.  The names are suggestive, but that's clearly not enough to go on.

pmundkur (approved):

PR #1538, opened 2026-02-06
Remove some unused string mappings.

pmundkur:


nadime15:
Hmm, weren't those just introduced 3-4 weeks ago by #1422, etc. ?

pmundkur:
> Hmm, weren't those just introduced 3-4 weeks ago by #1422, etc. ?

Right, they're not being used anywhere, so we shouldn't have added these mappings.  The error messages don't need them.

PR #1539, opened 2026-02-07
Legalize xret privilege with respect to misa.

### LINKED ISSUE #1522: One can still mret to U mode with misa.U disabled
0. Suppose we are in `M` mode with `misa.U` writable and presently `1`.
1. We can write to `mstatus` to set `MPP` to `U`.  This will pass legalization: https://github.com/riscv/sail-riscv/blob/0f291dec6f41e70480c733557c098fc2db746881/model/core/sys_regs.sail#L237 checks https://github.com/riscv/sail-riscv/blob/0f291dec6f41e70480c733557c098fc2db746881/model/core/sys_regs.sail#L161 which is just https://github.com/riscv/sail-riscv/blob/0f291dec6f41e70480c733557c098fc2db746881/model/core/sys_regs.sail#L134
2. We can now write to `misa` to clear `misa.U`.  No update to `mstatus` takes place.
3. We can now execute `mret`: neither the instruction body https://github.com/riscv/sail-riscv/blob/0f291dec6f41e70480c733557c098fc2db746881/model/extensions/I/base_insts.sail#L561-L570 nor the `exception_handler` case https://github.com/riscv/sail-riscv/blob/0f291dec6f41e70480c733557c098fc2db746881/model/sys/sys_control.sail#L231-L249 attempt to validate that the mode we are targeting is enabled.
4. At this point, we are in U mode with `misa.U` clear: `cur_privilege` is set to `U`.  (Note that `mstatus.MPP` is set to `M` rather than the expected `U` due to the assignment `exception_handler`).

pmundkur:
Check `misa` for enabled privilege modes at `xret`, since modifications to `misa` might have disabled the privileges in `mstatus.{MPP, SPP}`.  Add a test for this case.

This leaves handling of misa.U == 0 for later, to be addressed with the other misa bits.

While here, use an assert in place of an internal error as it should have been caught by config validation.

Fixes #1522.

### DIFF
--- a/model/sys/sys_control.sail
+++ b/model/sys/sys_control.sail
@@ -238,6 +238,17 @@ function trap_handler(del_priv : Privilege, c : TrapCause, pc : xlenbits, info :
   };
 }
 
+private function legalize_xret_privilege(p : Privilege) -> Privilege =
+  match p {
+    Machine           => Machine,
+    // If S-mode is not legal, then go to M-mode.
+    Supervisor        => if currentlyEnabled(Ext_S) then Supervisor else Machine,
+    // If U-mode is not legal, nor is S-mode.
+    User              => if currentlyEnabled(Ext_U) then User else Machine,
+    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+  }
+
 function exception_handler(cur_priv : Privilege, ctl : ctl_result,
                            pc: xlenbits) -> xlenbits = {
   match ctl {
@@ -252,7 +263,8 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       let prev_priv = cur_privilege;
       mstatus[MIE]  = mstatus[MPIE];
       mstatus[MPIE] = 0b1;
-      cur_privilege = privLevel_bits(mstatus[MPP], 0b0); // TODO: use mstatus[MPV] if hypervisor enabled
+      cur_privilege = legalize_xret_privilege(privLevel_bits(mstatus[MPP], 0b0)); // TODO: use mstatus[MPV] if hypervisor enabled
+      // "xPP is set to the least-privileged supported mode (U if U-mode is implemented, else M)."
       mstatus[MPP]  = privLevel_to_bits(if currentlyEnabled(Ext_U) then User else Machine);
       if   cur_privilege != Machine
       then mstatus[MPRV] = 0b0;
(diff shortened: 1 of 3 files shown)


pmundkur:
cc @nwf

model/sys/sys_control.sail
```diff
@@ -238,6 +238,17 @@ function trap_handler(del_priv : Privilege, c : TrapCause, pc : xlenbits, info :
   };
 }
 
+private function legalize_xret_privilege(p : Privilege) -> Privilege =
+  match p {
+    Machine           => Machine,
+    // If S-mode is not legal, then go to M-mode.
```
pmundkur, line 244: I'm not sure about this.  The spec says:

> When executing an xRET instruction, supposing xPP holds the value y, xIE is set to xPIE; the privilege mode is changed to y;

If `x` is `M`, `y` is `S` and `misa.S` is clear (and assuming misa.S == 0 implies S is not supported), and U is supported, should privilege mode be changed to M or U?

jordancarlin: `mstatus.MPP` is WARL, so based on the CSR Field Modulation rules, this is implementation specific. Presumably that means we need a config option for it.

> If a write to one CSR changes the set of legal values allowed for a field of a second CSR, then unless specified otherwise, the second CSR’s field immediately gets an UNSPECIFIED value from among its new legal values. This is true even if the field’s value before the write remains legal after the write; the value of the field may be changed in consequence of the write to the controlling CSR.

model/sys/sys_control.sail
```diff
@@ -271,8 +283,9 @@ function exception_handler(cur_priv : Privilege, ctl : ctl_result,
       let prev_priv = cur_privilege;
       mstatus[SIE]  = mstatus[SPIE];
       mstatus[SPIE] = 0b1;
-      cur_privilege = if mstatus[SPP] == 0b1 then Supervisor else User;
-      mstatus[SPP]  = 0b0;
+      cur_privilege = legalize_xret_privilege(if mstatus[SPP] == 0b1 then Supervisor else User);
+      // "xPP is set to the least-privileged supported mode (U if U-mode is implemented, else M)."
+      mstatus[SPP]  = if currentlyEnabled(Ext_U) then 0b0 else 0b1;
```
Timmmm, line 288: Probably should use the same style as line 268.

Timmmm (commented):
I think this probably isn't the right way to handle this. The thing that we don't actually do is the CSR field modulation of `mstatus.MPP`:

> xPP fields are WARL fields that can hold only privilege mode x and any implemented privilege mode lower than x. If privilege mode x is not implemented, then xPP must be read-only 0.

So it's not really that we're legalising the xret privilege, instead we should be modulating the xPP field so that it dynamically reads a legal value (or 0 if there are none).

As a concrete example of where this fix would still be broken, if you set `MPP` to S, write `misa.S`=0, and then read `mstatus` using `csrr` you will still read S.

I think unfortunately we need to add a `read_mstatus` function like we did for something else recently that I can't remember. :-D

The other option is to actually modify `mstatus.xPP` when `misa[S/U]` are modified. That is also legal behaviour (so in theory we might want to support it), however it also is a let less elegant and difficult to do in a bug-free way, so I don't think we should - until someone *really* wants it at least.

model/sys/sys_control.sail
Timmmm: Also... seems very weird that you could execute `sret` on an M-only machine but I couldn't see anything disallowing it at a quick glance.

PR #1541, opened 2026-02-10
Lean emulator

Ptival:
Opening this pull request for discussion about what it would take to integrate these changes upstream.

This PR contains:
- a frontend to start an emulator based on Lean extracted from SAIL RISC-V,
- a CI job that builds said emulator, and tests it on some ELF files.

There are, however, a couple of things that ought to be discussed:

1. In order to parse ELF files, this adds a dependency to [ELFSage](https://github.com/draperlaboratory/ELFSage). Well, almost, the project is purposefully kept compatible with an old version of Lean (see [here](https://github.com/draperlaboratory/ELFSage/pull/16)), so we forked and modernized it. This is a new dependency. I have found it fairly easy to maintain, but I wanted to be upfront about it.
2. Our "main" is not the cleanest, I might do a tiny bit of refactoring, but please let us know whether it would require significant changes to merge.
3. I have extended the Lean workflow to test the extracted emulator. Unfortunately, we have to re-extract the Lean code with a smaller set of modules to get less prohibitive compilation times, and then also extract the ELF files. There might be nicer ways of achieving this without all the redundancy here, but I'm neither an expert at CMake or CI, so I'm open to ideas.

Putting this as a draft request for now.


.github/workflows/compile-lean.yml
```diff
@@ -51,3 +51,56 @@ jobs:
...
+        # e.g. floating point operations that are currently not supported.
+        run: |
+          lake exec lean_riscv_emulator -- ../build/test/2025-10-14/riscv-tests/rv64ui-p-add
+          lake exec lean_riscv_emulator -- ../build/test/2025-10-14/riscv-tests/rv64ui-p-jal
+          lake exec lean_riscv_emulator -- ../build/test/2025-10-14/riscv-tests/rv64ui-p-sub
+          lake exec lean_riscv_emulator -- ../build/test/2025-10-14/riscv-tests/rv64ui-v-jal
+          lake exec lean_riscv_emulator -- ../build/test/2025-10-14/riscv-tests/rv64um-p-div
+          lake exec lean_riscv_emulator -- ../build/test/2025-10-14/riscv-tests/rv64um-p-mul
+          lake exec lean_riscv_emulator -- ../build/test/2025-10-14/riscv-tests/rv64um-v-div
+          lake exec lean_riscv_emulator -- ../build/test/2025-10-14/riscv-tests/rv64um-v-mul
```
nadime15: Yeah, as the comment says, you probably don’t want to hardcode these paths.

Would it make more sense to keep this fully inside CMake?

You could do something similar to what we are doing in test/CMakeLists.txt, add lean as a target and pass a CMake flag to run these tests (e.g. -DENABLE_LEAN_TESTS=TRUE)?

test/CMakeLists.txt
```diff
@@ -123,4 +123,10 @@ if (FIRST_PARTY_TESTS)
     add_subdirectory("first_party")
 endif()
 
-add_subdirectory("unit_tests")
+# This is off by default so we don't require people who
+# just want to build the model to have Clang or RISC-V GCC
+# installed.
+option(UNIT_TESTS "Compile & run unit tests (requires Clang or RISC-V GCC).")
```
pmundkur: Either this option should be `ON` by default, or the various CI workflows need to turn it on.  I'd make this on by default, and turn it off in `./build_simulator.sh` which is the user-facing script; wonder what others think?

pmundkur (commented):
This is really great!

.github/workflows/compile-lean.yml
Ptival: I agree it would make sense to move this into CMake. I will attempt it myself, and ask for help if needed.

test/CMakeLists.txt
Ptival: I think I agree with you. I will wait a bit and change it if nobody is against it.

Timmmm:
I agree this is great! I don't see any issue with merging it except potentially that not many of us really know Lean. Would you be willing to maintain it. Also what are you using it for, just out of curiosity?

(And on the other hand, learning Lean has been on my todo list for ages so maybe this is a good excuse to finally get around to it!)

test/CMakeLists.txt
jordancarlin: Maybe I'm missing something, but what about the unit tests requires a RISC-V compiler? These are the native Sail tests and don't involve any C/assembly/etc. code.

jprider63:
> I agree this is great! I don't see any issue with merging it except potentially that not many of us really know Lean. Would you be willing to maintain it. Also what are you using it for, just out of curiosity?
> 
> (And on the other hand, learning Lean has been on my todo list for ages so maybe this is a good excuse to finally get around to it!)

There's a group of us at Galois and Cambridge who have been working on Sail's Lean backend so we can help if things break.

test/CMakeLists.txt
Ptival: Removing this change from this PR, seems unrelated to what I'm doing and I'm not sure what was the intent of the person who made this change.

Ptival:
Alright, I have made a lot of changes, and it's back to being reviewable:

- I have moved the whole logic of building and running the Lean emulator in CMake. Now the CI only calls the appropriate CMake target, much nicer.
- In order to support printing out the logging of the emulator run, I needed to pass `-D PRINT_EFFECTS` to sail. But there was no way to do so from the current CMake setup, so I added an option `PRINT_EFFECTS` that, when set, adds `-D PRINT_EFFECTS` to sail invocations.
- I wanted to use the status code of the emulator run as the status code of the process, so that a failing emulator run would be a failing run for CI. Unfortunately, the original `loop` did not return the result. If this is okay with everyone, I'm suggesting a change where `loop` returns the exit code of the code being looped. I have updated one use that doesn't care about it to throw it away, and I'm demonstrating in the emulator main function its use.
- As part of having `loop` return a status code, I reorganized its control flow to have a clearer single exit point, rather than the current state where the handling of the exit state happens mid-loop.

EDIT: The CI is currently failing because of this limitation:
https://github.com/rems-project/sail/pull/1624

lean_emulator/lake-manifest.json
pmundkur, line 1: This file needs to pass `pre-commit` checks. See
https://github.com/riscv/sail-riscv/blob/c4d314048c891d59bd471d3bbd262cd30278ed14/CONTRIBUTING.md?plain=1#L33

pmundkur (commented):
This needs a pre-commit fix for CI.

Does the Lean emulator need an unusual amount of CPU/memory resources to build?  It's been building for quite a while on my desktop.

lean_emulator/lake-manifest.json
Ptival: I will look at the pre-commit fix.

As for the resources, I have been building it on a powerful laptop so I'm not sure.
Anecdotally, on our Github CI, the Lean emulator build seems to take ~8 minutes.

model/postlude/step.sail
```diff
@@ -268,7 +268,7 @@ function try_step(step_no : nat, exit_wait : bool) -> bool = {
   }
 }
 
-function loop () : unit -> unit = {
+function loop () : unit -> int = {
```
Timmmm:
```suggestion
function loop () : unit -> nat = {
```

Presumably?

lean_emulator/test_lean_emulator.cmake
```diff
@@ -0,0 +1,80 @@
...
+    rv64ui-p-xori
+)
+
+execute_process(
+    COMMAND lake build lean_riscv_emulator
+    WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}
+)
+foreach(file IN LISTS LEAN_EMULATOR_TEST_ELFS)
+    message(STATUS "Testing Lean emulator on file ${file}")
+    execute_process(
```
Timmmm, line 75: Wouldn't it make more sense to add `lake build ...` as a normal CMake target, and then add these as CTest tests? I can sketch out the code if you don't want to learn CMake (which would be very very understandable!)

Ptival: Please do! I am indeed not used to CMake and have been teaching myself in order to produce what you're currently seeing.

model/postlude/step.sail
Ptival: Good point, in fact I'm not sure what a Sail `int` is, so maybe we should be careful.

It seems we start with a `htif_exit_code`, with declared type `bits(64)`, so I guess this is a sign-agnostic view of it.

I wrote `int` because I had in mind the typical C status codes, e.g. `return -1;`, where they are typically considered signed. But this is likey inconsistent with the call to `unsigned` at:
```
let exit_val = unsigned(htif_exit_code);
```

So the question is, either:
- we want to view exit codes as signed, in which case we would leave `int` but replace `unsigned(...)` with `signed(...)`, or,
- we want to view exit codes as unsigned, in which case we would replace `int` with `nat`.

Do you have more knowledge on the reasonable thing to do here?

lean_emulator/lake-manifest.json
Ptival: I fixed the pre-commit problem. As for the slowness of it, I'm open to suggestions, in particular:
- maybe this could be run daily rather than per commit?
- we could definitely trim the list of ELF files to test, it's somewhat arbitrary.

lean_emulator/test_lean_emulator.cmake
Ptival: Hi @Timmmm , just checking on whether you have bandwidth to look into this, or if I should figure it out.

ineol:
@Ptival @Timmmm In the interested of keeping this moving, I updated @Ptival's branch and moved the emulator tests to be cmake tests in this branch https://github.com/ineol/sail-riscv/tree/lean_emulator_ci2. @Ptival if that works for you can you pull it into your PR branch?

.github/workflows/compile-lean.yml
```diff
@@ -51,3 +51,9 @@ jobs:
         uses: leanprover/lean-action@v1
         with:
           lake-package-directory: "build/model/Lean_RV64D/"
+
+      - name: Test the Lean emulator
+        run: |
+          eval $(opam config env)
+          cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;Sscofpmf;postlude" -DPRINT_EFFECTS=true
+          cmake --build build --target test_lean_emulator
```
ineol, line 59: I think this should be deleted now if we're using the test ci action

lean_emulator/test_lean_emulator.cmake
```diff
@@ -0,0 +1,80 @@
+include(${CMAKE_CURRENT_LIST_DIR}/../test/download_riscv_tests.cmake)
```
ineol, line 1: This file is not needed anymore I think, since it has moved to the existing test framework in the `test` directory

test/CMakeLists.txt
```diff
@@ -1,46 +1,4 @@
...
-        if(NOT extract_result EQUAL 0)
-            message(FATAL_ERROR "Failed to extract ${TARBALL_NAME}")
-        endif()
-
-        # Clean up the tar.gz file
-        file(REMOVE "${DOWNLOAD_PATH}/${TARBALL_NAME}.tar.gz")
-        message(STATUS "${TARBALL_NAME} downloaded and extracted successfully")
-    endif()
-endfunction()
+include(download_riscv_tests.cmake)
```
ineol: Since we're doing the Lean tests here we don't need to factor the downloading function in a separate file

ineol:
Things got a bit weird with the rebase/merge I think :(

Also the tests run instantly for me (after the emulator has been built), which seems to contradict some comments above.

test/download_riscv_tests.cmake
pmundkur, line 1: This file looks like it's not needed anymore and can be deleted.

lean_emulator/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Instructions
+
+1. Compile the Sail RISC-V model for Lean which is a dependency for the Lean emulator. Here's an example command to build the Lean backend (from the root directory). You may want to change the set of modules and extensions that are included in your version of the emulator.
+
+```
+cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;postlude" && cmake --build build --target generated_lean_executable_rv64d
+```
+
+2. Compile the Lean emulator.
```
pmundkur, line 11:
```suggestion
2. Compile the Lean emulator (from this directory).
```

lean_emulator/README.md
```diff
@@ -0,0 +1,27 @@
+# Lean emulator for Sail RISC-V
+
+## Instructions
+
+1. Compile the Sail RISC-V model for Lean which is a dependency for the Lean emulator. Here's an example command to build the Lean backend (from the root directory). You may want to change the set of modules and extensions that are included in your version of the emulator.
+
+```
+cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;postlude" && cmake --build build --target generated_lean_executable_rv64d
+```
+
```
pmundkur, line 10: Perhaps also give the commands to enable the tests by adding `-DENABLE_LEAN_EMULATOR_TESTS=TRUE`
```
cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;postlude" -DENABLE_LEAN_EMULATOR_TESTS=TRUE && cmake --build build --target generated_lean_executable_rv64d
```

lean_emulator/README.md
```diff
@@ -0,0 +1,27 @@
...
+```
+lake exec lean_riscv_emulator -- path/to/riscv/binary.elf
+```
+
+You can download the Spike test ELFs by running the following in the `/test` directory.
+
+```
+cmake -S . -B build -DENABLE_RISCV_TESTS=1 && cmake --build build
+```
```
pmundkur, line 27: Is this still needed?
Instead one could say something like
```
4. If tests were enabled for the build (using `-DENABLE_LEAN_EMULATOR_TESTS=TRUE` above), they can be run (from the root directory) with

ctest --test-dir build -R lean_emulator
```

pmundkur (commented):
I'm able to build the emulator with Cmake now, nice!   The instructions may need some minor tweaks.

And the all tests are failing; the failure looks like a missed TODO:

```     
 Start 51: lean_emulator_rv64ui-p-xori
50/50 Test #51: lean_emulator_rv64ui-p-xori ......***Failed    0.02 sec
TODO: cancel_reservation
Error while running the sail program!: Assertion failed: Pattern match failure at extensions/Zicboz/zicboz_insts.sail:11.0-11.71
PC = 0x0000000080000000
htif_tohost = 0x0000000080001000


0% tests passed, 50 tests failed out of 50
```

Looks pretty close though!

.github/workflows/compile-lean.yml
pmundkur, line 59: this seems to work
```suggestion
          eval $(opam config env)     
          cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;Sscofpmf;postlude" -DPRINT_EFFECTS=true -DENABLE_LEAN_EMULATOR_TESTS=TRUE          
          cmake --build build --target build_lean_emulator          
          ctest --test-dir build -R lean_emulator
```

Ptival:
@ineol is going to take over landing this PR because I no longer have time to work on it.

Closing this in favor of #1777 where Léo can lead the discussion and have direct access to the branch to merge from.

PR #1542, opened 2026-02-10
Add the Rocky Linux container build to CI.

pmundkur:
This adds the platform used for the binary releases to the regular CI.

Unfortunately, the Clang on Rocky 8.9 does not support RISC-V targets, so the first party tests cannot be built (and hence cannot be tested).

The matrix strategy in CI now uses the matrix cell for `ubuntu-24.04-arm` and CMake `3.20.0` for the Rocky container build, whereas previously that cell ran on the runner directly.  This means that first party tests are no longer tested in that cell.  This should not be a problem since they are still tested in the cell for `ubuntu-24.04-arm` and CMake `4.1.2`.

The `sail-setup` action has been ported for the Rocky Linux container, and the current support matrix is documented.

### DIFF
--- a/.github/actions/sail-setup/action.yml
+++ b/.github/actions/sail-setup/action.yml
@@ -62,7 +62,7 @@ runs:
       if: runner.os == 'Linux' && job.container != null
       shell: bash
       # This could be a Rocq or Rocky container.  Don't run this on
-      # the Rocq container because it is a Debian and doesn't need
+      # the Rocq container because it is a Debian-based container and doesn't need
       # these packages anyway.
       run: |
         # sudo needed for compatibility with other steps using sudo


pmundkur:
This will need some minor changes once #1528 is merged.   I can adapt the release workflow on top of this separately.

nadime15:
I see in release.yml that GCC is installed, so why not use GCC to compile the tests?

jordancarlin:
> I see in release.yml that GCC is installed, so why not use GCC to compile the tests?

The host (x86 or arm) gcc compiler is installed, not a riscv cross-compiler version of gcc.

nadime15:
Ahh ok, but couldn’t you just install RISC-V GCC along with it?

pmundkur:
> Ahh ok, but couldn’t you just install RISC-V GCC along with it?

The GCC cross-compiler for RISC-V does not have packages in the Rocky 8.9 distro.  It would have to be built from source, which is not worth it for the few first-party tests that are already covered on other platforms in CI.

nadime15:
You can build it once and cache it in CI, avoiding repeated rebuilds.

pmundkur:
> You can build it once and cache it in CI, avoiding repeated rebuilds.

Ok, but I still don't think it is worth it.  In any case, I'd suggest that it not be part of this PR, and we can address that post-release if really needed.

nadime15:
This feels like a workaround that could be fixed right away. If we know we will want these tests on Rocky eventually (which I think we do), then we are creating two steps where one will do. Now we basically, 1. Merge this workaround now and 2. Later undo it and implement the cross-compiler solution. Thats more total work than just implementing the cross-compiler solution now (which should not take much time and would avoid this workaround).

@jordancarlin @Timmmm @trdthg any thoughts?

Timmmm:
I agree with @pmundkur - probably not worth it. The chances that we build a simulator that a) works on Ubuntu, b) works for the riscv-tests on Rocky 8, and c) *doesn't* work for the first party tests on Rocky 8... is basically zero.

.github/actions/sail-setup/action.yml
```diff
@@ -14,17 +14,62 @@ inputs:
...
+#                          | CMake: (4.1.2)               |       x        |       x         |           | CMake: (4.1.2) |    x
+#                          | Clang: latest (apt.llvm.org) |                |                 |           | Clang: -       |
+#--------------------------------------------------------------------------------------------------------------------------------------
+# (Container builds)       |                              |                |                 |           |                |
+# ------------------       |                              |                |                 |           |                |
+#                          |                              |                |                 |           |                |
+# rockylinux:8.9           | Sail:  0.20.1                |                |                 |           |                |
+# runner: ubuntu-latest    | CMake: 3.20.0                |       x        |                 |           |                |
+#                          | Clang: -                     |                |                 |           |                |
```
jordancarlin, line 42: This matrix documentation seems difficult to maintain and I’m not sure that it useful enough to warrant having it.

.github/actions/sail-setup/action.yml
```diff
@@ -37,6 +82,29 @@ runs:
...
+      # So the below does not work:
+      # CLANG_VERSION=21.1.8
+      # curl --location https://github.com/llvm/llvm-project/releases/download/llvmorg-${CLANG_VERSION}/LLVM-${CLANG_VERSION}-Linux-X64.tar.xz | xz -c -d | sudo tar xv --directory=/usr/local --strip-components=1
+      #
+      # So just use whatever the platform has for now, since we cannot run first-party tests anyway
+      # until we upgrade to much newer versions of Rocky.
+      run: |
+        if [ -f /usr/bin/dnf ]; then
+          dnf install -y clang llvm lld
+        fi
```
jordancarlin: Is there any reason to install clang if we don't use it to compile the first-party tests? I think we can just drop this section.

.github/workflows/ci.yml
```diff
@@ -32,13 +47,23 @@ jobs:
         if: ${{ matrix.run_all_steps}}
         run: python3 -m pip install pre-commit && pre-commit run --all-files --show-diff-on-failure --color=always
 
-      - name: Build simulators
+      - name: Build simulator and first party tests
+        if: job.container == null
         run: |
           # Ninja is used because the CMake Makefile generator doesn't
           # build top-level targets in parallel unfortunately.
           cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWARNINGS_AS_ERRORS=TRUE -DFIRST_PARTY_TESTS=TRUE -DENABLE_RISCV_TESTS=TRUE -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE
```
jordancarlin: You can probably inline the `job.container == null` check as the value to use for `FIRST_PARTY_TESTS` so that we don't need two (almost identical) variants of this step.

.github/workflows/ci.yml
```diff
@@ -5,7 +5,6 @@ on: [push, pull_request, workflow_dispatch, merge_group]
 jobs:
   build:
     name: "Build and test model (OS: ${{ matrix.os }}, CMake Version: ${{ matrix.cmake_version }})"
```
jordancarlin: Update the CI job name to include the container so it is obvious which job failed.

.github/actions/sail-setup/action.yml
pmundkur: I don't this will be too hard to maintain, and I suspect it would be useful to folks (like me) who aren't fluent in github actions.  I can maintain it (as I try to with the rest of the docs).

.github/actions/sail-setup/action.yml
pmundkur: I debated this; I think it would be odd API for the action to be passed 'install-clang' set to true and not install clang.   Also, the comments here explain the various issues with clang on Rocky, which might be useful when we revisit this later.

.github/workflows/ci.yml
pmundkur: Sorry, my actions-fu is not good enough to interpret what to do here.

.github/actions/sail-setup/action.yml
jordancarlin: I'd rather have a check in the action that fails if `install-clang` is true on a platform that doesn't support it. I think this is even more misleading.

.github/workflows/ci.yml
nadime15: Line 50-70 are 2 jobs that basically do the same, so I think what he means is to do something like:
```

- name: Build simulator and optionally first party tests
  run: |
    cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWARNINGS_AS_ERRORS=TRUE -DFIRST_PARTY_TESTS=${{ job.container == null }} -DENABLE_RISCV_TESTS=TRUE -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE $CLANG_FLAG
    ninja -C build all generated_sail_riscv_docs generated_smt_rv64d generated_smt_rv32d
```

.github/actions/sail-setup/action.yml
nadime15: In theory, you could get a similar overview by clicking on Actions, selecting a specific workflow, and then clicking on a particular run. On the left-hand side, you would see an overview of all the jobs that were executed, including the OS and CMake versions they used. We could extend this, by modifying the `name` field of the workflow to include the Sail version, Clang version, etc. and achieve a similar overview without having to maintain it manually.

I am not saying this is necessarily better, but in this case, you wouldn’t need to maintain a separate table.

trdthg:
~Why not just download clang from https://github.com/riscv-collab/riscv-gnu-toolchain ?
`setup-riscv-gnu-toolchain`
then we get the latest riscv supported compiler directly.~

Well, its glibc version is also too high. 🤷🏻‍♂️

.github/actions/sail-setup/action.yml
pmundkur: There are advantages to both, I don't see them as mutually exclusive.  There is some benefit to seeing the whole table in an editor without having to go pointy-clicky in a browser for individual runs.  I don't think this table will change often, or by much, so I'm not convinced by the maintenance burden argument.  If it is indeed a burden, we could always remove it later: these are just comments in a file.

.github/workflows/ci.yml
```diff
@@ -4,8 +4,7 @@ on: [push, pull_request, workflow_dispatch, merge_group]
 
 jobs:
   build:
-    name: "Build and test model (OS: ${{ matrix.os }}, CMake Version: ${{ matrix.cmake_version }})"
-    runs-on: ${{ matrix.os }}
+    name: "Build and test model (OS: ${{ matrix.os }}, container: ${{ matrix.container }}, CMake Version: ${{ matrix.cmake_version }})"
```
jordancarlin: The names look a little funky when container is null now. Maybe only include `container: ` if `matrix.container` is not null.

.github/workflows/ci.yml
```diff
@@ -14,8 +13,24 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    container: ${{ matrix.container }}
 
     steps:
+      - name: Install Git (Rocky)
+        # This needs be done before the git checkout step.
+        if: startsWith(matrix.os, 'ubuntu') && job.container != null
+        run: |
+          dnf update -y --enablerepo=devel
+          dnf install -y --enablerepo=devel git
```
jordancarlin: `--enablerepo=devel` is not needed when installing git.

.github/workflows/ci.yml
pmundkur: I have no idea how to do that.  I'm getting yaml pre-commit errors if I try to put expressions in 'name:'.  I'm out of my depth here.

jordancarlin: I think the following should work:
```suggestion
    name: "Build and test model (OS: ${{ matrix.os }}${{ matrix.container && format(', container: {0}', matrix.container) || '' }}, CMake Version: ${{ matrix.cmake_version }})"
```

pmundkur: sweet!

.github/actions/sail-setup/action.yml
```diff
@@ -14,17 +14,62 @@ inputs:
...
 
-    - name: Install latest Clang (Linux)
-      if: runner.os == 'Linux' && inputs.install-clang == 'true'
+    - name: Install packages (Linux container)
+      if: runner.os == 'Linux' && job.container != null
+      shell: bash
+      # This could be a Rocq or Rocky container.
+      run: |
+        # sudo needed for compatibility with other steps using sudo
+        if [ -f /usr/bin/dnf ]; then
```
Timmmm, line 69: Why do we need to check if this exists?

.github/workflows/ci.yml
```diff
@@ -32,7 +47,8 @@ jobs:
         if: ${{ matrix.run_all_steps}}
         run: python3 -m pip install pre-commit && pre-commit run --all-files --show-diff-on-failure --color=always
 
-      - name: Build simulators
+      - name: Build simulator and first party tests
+        if: job.container == null
```
Timmmm: It might be better to have explicit variables for this and set them in the matrix, i.e.

```
        include:
          - os: ubuntu-22.04
            cmake_version: "4.1.2"
            run_all_steps: true
            first_party_tests: true
          - os: ubuntu-latest
            container: rockylinux:8.9
            cmake_version: "3.20.0"
            first_party_tests: false
          - os: ubuntu-24.04-arm
            container: rockylinux:8.9
            cmake_version: "3.20.0"
            first_party_tests: false
```

.github/workflows/ci.yml
```diff
@@ -44,6 +60,15 @@ jobs:
           cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWARNINGS_AS_ERRORS=TRUE -DFIRST_PARTY_TESTS=TRUE -DENABLE_RISCV_TESTS=TRUE -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE $CLANG_FLAG
```
Timmmm: Just do this?

```
cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWARNINGS_AS_ERRORS=TRUE -DFIRST_PARTY_TESTS=${matrix.first_party_tests} -DENABLE_RISCV_TESTS=TRUE -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE $CLANG_FLAG
```

Timmmm (changes requested):
Sorry for slow review! Some minor comments...

.github/actions/sail-setup/action.yml
pmundkur: It's called in the rocq container, which is a Debian, and those packages are not needed for the Rocq workflow.  It's hackish; happy to improve it.

.github/workflows/ci.yml
```diff
@@ -4,18 +4,36 @@ on: [push, pull_request, workflow_dispatch, merge_group]
...
 jobs:
   build:
-    name: "Build and test model (OS: ${{ matrix.os }}, CMake Version: ${{ matrix.cmake_version }})"
-    runs-on: ${{ matrix.os }}
+    name: "Build and test model (OS: ${{ matrix.os }}${{ matrix.container && format(', container: {0}', matrix.container) || '' }}, CMake Version: ${{ matrix.cmake_version }})"
     strategy:
       matrix:
         os: [ubuntu-22.04, ubuntu-24.04-arm, macos-latest]
         cmake_version: ["3.20.0", "4.1.2"]
+        first_party_tests: [true]
```
jordancarlin: This matrix is starting to get complicated enough and have enough overrides that it might make sense to switch fully to explicit include items.

pmundkur: Fine with me.  I'd like to get a release out soon, well before Monday, ideally with this and #1543.  There's no real reason to wait for the two, but it just seems better that way.  @Timmmm could we converge on this soon?

.github/actions/sail-setup/action.yml
jordancarlin: How about checking if `job.container` is `*rockylinux*` instead of non-null? Then you could drop the `dnf` check in the actual code that is run.

pmundkur: Something like
```
if: runner.os == 'Linux' && startswith(job.container, 'rockylinux')
```
?

jordancarlin: Not sure if that exact syntax will work, but yes, something along those lines. Push it and we will see what happens I guess.

pmundkur: It didn't work on my test branch.

pmundkur: https://github.com/pmundkur/sail-riscv/actions/runs/21997830956/job/63562405343

pmundkur: `echo '${{ toJSON(job.container) }}'` gives
```
 echo '{
    "network": "github_network_3410b12beb6047ab9433ad1e288f3db0",
    "id": "1539337ec5945f0b00f749ca12fb8aa100f56a71e41b409e93401e3a14179d16"
  }'
```

jordancarlin: Well that's unfortunate. Never mind then.

.github/workflows/ci.yml
Timmmm: Yeah I think an explicit list would probably be simpler at this point. I'm away from a computer until Monday so feel free to merge this if everyone else is happy. We can always improve it later!

.github/actions/sail-setup/action.yml
```diff
@@ -14,17 +14,64 @@ inputs:
...
       run: |
         sudo apt install -y --no-install-recommends pkg-config libgmp-dev curl ninja-build
 
-    - name: Install latest Clang (Linux)
-      if: runner.os == 'Linux' && inputs.install-clang == 'true'
+    - name: Install packages (Linux container)
+      if: runner.os == 'Linux' && job.container != null
+      shell: bash
+      # This could be a Rocq or Rocky container.  Don't run this on
+      # the Rocq container because it is a Debian and doesn't need
```
jordancarlin:
```suggestion
      # the Rocq container because it is a Debian-based container and doesn't need
```

PR #1544, opened 2026-02-11
Use FetchContent for CLI11 and bump to v2.6.1

trdthg:
I don't think we'd have any reason to modify the CLI11 source code.

> Meanwhile, CLI11 v2.6 allows us to disable the default formatting behavior of the footer (removing all whitespace characters, including ' ' and '\t'). which might be useful


jordancarlin:
Can you post the new output from `--help` so we can compare it to our custom version?

trdthg:
```diff
14,43c14,43
<   -h, --help                                 Print this help message and exit 
<       --show-times                           Show execution times 
<       --version                              Print model version 
<       --build-info                           Print build information 
<       --print-default-config                 Print default configuration 
<       --print-config-schema                  Print configuration schema 
<       --validate-config                      Exit after config validation (it is always validated) 
<       --print-device-tree                    Print device tree 
<       --print-isa-string                     Print ISA string 
<       --enable-experimental-extensions       Enable experimental extensions 
<       --use-abi-names                        Use ABI register names in trace log 
<       --device-tree-blob <file>              Device tree blob file 
<       --terminal-log <file>                  Terminal log output file 
<       --test-signature <file>                Test signature file 
<       --config <file>                        Configuration file 
<       --trace-output <file>                  Trace output file 
<       --signature-granularity <uint>         Signature granularity 
<       --rvfi-dii <int> (within [1 - 65535])  RVFI DII port 
<       --inst-limit <uint>                    Instruction limit 
<       --trace-instr                          Enable trace output for instruction execution 
<       --trace-ptw                            Enable trace output for Page Table walk 
<       --trace-reg                            Enable trace output for register access 
<       --trace-mem                            Enable trace output for memory accesses 
<       --trace-rvfi                           Enable trace output for RVFI 
<       --trace-clint                          Enable trace output for CLINT memory accesses and status 
<       --trace-exception                      Enable trace output for exceptions 
<       --trace-interrupt                      Enable trace output for interrupts 
<       --trace-htif                           Enable trace output for HTIF operations 
<       --trace-pma                            Enable trace output for PMA checks 
<       --trace-platform                       Enable trace output for platform-level events (MMIO, interrupts, 
---
>   -h,  --help                                Print this help message and exit 
>        --show-times                          Show execution times 
>        --version                             Print model version 
>        --build-info                          Print build information 
>        --print-default-config                Print default configuration 
>        --print-config-schema                 Print configuration schema 
>        --validate-config                     Exit after config validation (it is always validated) 
>        --print-device-tree                   Print device tree 
>        --print-isa-string                    Print ISA string 
>        --enable-experimental-extensions      Enable experimental extensions 
>        --use-abi-names                       Use ABI register names in trace log 
>        --device-tree-blob <file>             Device tree blob file 
>        --terminal-log <file>                 Terminal log output file 
>        --test-signature <file>               Test signature file 
>        --config <file>                       Configuration file 
>        --trace-output <file>                 Trace output file 
>        --signature-granularity <uint>        Signature granularity 
>        --rvfi-dii <int> (within [1 - 65535]) RVFI DII port 
>        --inst-limit <uint>                   Instruction limit 
>        --trace-instr                         Enable trace output for instruction execution 
>        --trace-ptw                           Enable trace output for Page Table walk 
>        --trace-reg                           Enable trace output for register access 
>        --trace-mem                           Enable trace output for memory accesses 
>        --trace-rvfi                          Enable trace output for RVFI 
>        --trace-clint                         Enable trace output for CLINT memory accesses and status 
>        --trace-exception                     Enable trace output for exceptions 
>        --trace-interrupt                     Enable trace output for interrupts 
>        --trace-htif                          Enable trace output for HTIF operations 
>        --trace-pma                           Enable trace output for PMA checks 
>        --trace-platform                      Enable trace output for platform-level events (MMIO, interrupts, 
45,46c45,46
<       --trace-step                           Add a blank line between steps in the trace output 
<       --trace-all                            Enable all trace output 
---
>        --trace-step                          Add a blank line between steps in the trace output 
>        --trace-all                           Enable all trace output 
```
The difference is the spacing between long options and short options. Above is the result for 1/6.

But I just tried 1/7 and it could maintains the same distance as before.

c_emulator/riscv_sim.cpp
```diff
@@ -233,10 +233,7 @@ static CLIOptions parse_cli(int argc, char **argv) {
     "file. This is optional with some arguments, e.g. --print-isa-string."
   );
 
-  // long_options_offset() is a local addition, so when updating CLI11,
-  // see how https://github.com/CLIUtils/CLI11/pull/1185 ended up,
-  // and possibly implement the upstream solution.
-  app.get_formatter()->long_options_offset(6);
+  app.get_formatter()->long_option_alignment_ratio(1 / 7.f);
```
pmundkur: Might be worth comparing with #1289 (perhaps you already did?)

trdthg: 6/45 looks better. Does 6 come from `len('  -h, ')`?

pmundkur: @radimkrcmar ?

pmundkur (approved):

c_emulator/riscv_sim.cpp
radimkrcmar: > 6/45 looks better. Does 6 come from len('  -h, ')?

Yes.  Then dividing by the column width to make the offset be 6 characters even if we change the width.

PR #1545, opened 2026-02-12
Add Hypervisor Instructions

nadime15:
This is an initial draft of adding support for the hypervisor instructions, based on https://github.com/riscv/sail-riscv/pull/1419, https://github.com/riscv/sail-riscv/pull/1110 and #1513 (the actual new code is introduced starting from commit https://github.com/riscv/sail-riscv/commit/edf48f74040f11acef7acd59d90e049f97f25180).

I argue it is rather straightforward, there are just a few things that are a bit tricky.

1. HLVX needs different permissions after the address translation has succeeded, and as far as I understand we need Execute and Load access permission. So this draft PR introduces ExecuteLoad.

2. HLV(X)/HSV pass different effective privileges to the address translation, depending on `hstatus[SPVP]`. That is why the original hypervisor PR introduced `translateAddr_pv` (or `translateAddr_eff_priv`() in this PR), which passes the corresponding privilege mode along with it.

3. We probably want to access data byte-wise.


nadime15:
Small success story, the RISC-V test repo includes a simple test for the 2-stage address translation, and it passes!

jordancarlin:

PR #1546, opened 2026-02-12
Add missed stateen checks for the high-half CSRs of `hstateen[0-3]`.

pmundkur:
This checks are the same as for the low-halfs, but at the high-half CSR addresses.


pmundkur:
I'm attempting a simpler version of #1402.

Timmmm (approved):

Issue #1506, opened 2026-01-28
Aligned access config options are confusing


Timmmm:
See #1468 for more discussion.

Currently we have these options that affect the behaviour of misaligned memory accesses:

`memory.misaligned.supported : bool` - this is a global option that applies to all accesses. It checks the alignment of the virtual address before translation and raises a misaligned address exception if it's not aligned. This is intended for CPUs that don't support misaligned access at all.

`memory.misaligned.allowed_within_exp : range(0, 11)` - after the `memory.misaligned.supported` check and still before translation we check if the access is _within_ an aligned region size `2^allowed_within_exp`. If it isn't we split it into multiple aligned memory accesses. This is used to handle page-crossing accesses and probably Zama16b (accesses that don't cross a 16-byte boundary are atomic).

`memory.misaligned.byte_by_byte : bool` - when an access is split into multiple memory operations due to `allowed_within_exp` then it is either split into maximally-sized aligned operations, or single-byte operations depending on the value of this setting.

`memory.misaligned.order_decreasing : bool` - this just controls the order of the operations.

After all of those settings are applied, we do address translation to the N memory operations, look up the PMA, and then apply the PMA option:

`memory.regions[N].attributes.misaligned_fault : "NoFault"/"AccessFault"/"AlignmentFault"` - if the physical address of the memory operation is misaligned and this is not `NoFault` then we raise an exception of the given kind.

I tried to capture most of this in a flow chart:

<img width="1346" height="872" alt="Image" src="https://github.com/user-attachments/assets/f014ce30-fa5a-4cc1-9b3c-06d882bb3293" />

There are some things we need to resolve:

* [ ] We need to document this better.
* [ ] `"AlignmentFault"` is probably the wrong word since it's never referred to as a "fault" in the ISA manual. (What even is a fault?)
* [ ] It's a bit weird that we split into multiple memory operations *before* address translation. For example if `byte_by_byte` is `true` then each byte is translated separately. We should only ever need to do a max of two address translations.
* [ ] The flow of the code is kind of weird and doesn't actually match the above flow chart (though the result is the same). I think it could be simplified.
* [ ] Whether or not these options apply to atomics or instruction fetch is undefined, because we never do misaligned instruction fetch and misaligned atomics aren't supported by the model yet.

pmundkur:
This is great!  It would be good to eventually explicitly differentiate load/stores and atomics in the diagram, to clarify handling for extensions like Zicclsm and Zama16b.

[...]

PR #1548, opened 2026-02-13
rename memory.misaligned.allowed_within_exp

KotorinMinami:
Part of #1506. The names memory.misaligned.supported and memory.misaligned.allowed_within_exp sound very similar and can be confusing. I think we should refine the naming to better reflect the distinction. Suggestions welcome.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -86,7 +86,7 @@
       "supported": true,
       "byte_by_byte": false,
       "order_decreasing": false,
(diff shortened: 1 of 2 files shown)


Timmmm:
Maybe something like `split_granularity_exp` or `split_at_exp`? I don't think `translation_` is right because it isn't only about translation.

KotorinMinami:
The former might be better, I think.

PR #1549, opened 2026-02-13
Add first party memory model with support for memory dumps

### LINKED ISSUE #1490: Support dumping memory
Once #1489 is completed we should augment it to allow dumping memory at the end of a test *or at the end of an RVVI-DII run*. Ideally we will dump to ELF format and set the entry point to the first instruction that was executed.

This is intended to allow generating ELFs from TestRig runs that can be run in lockstep without using DII (i.e. so the fetch unit is exercised).

Timmmm:
Adds a first party memory model and support for dumping memory contents as a core dump (ELF) file.

If you run a test and then dump memory at the end of the test you can run the dump in the emulator again (since the emulator doesn't care if the ELF is a core dump).

The only caveat is it doesn't copy the symbol table to the dump so you don't get `tohost` and the HTIF printing/test ending don't work.

Fixes #1490

### DIFF
--- a/c_emulator/elf_loader.cpp
+++ b/c_emulator/elf_loader.cpp
@@ -2,8 +2,9 @@
 #include "elfio/elf_types.hpp"
 
 #include <elfio/elfio.hpp>
+#include <memory>
 
-ELF::ELF(std::unique_ptr<ELFIO::elfio> reader) : m_reader(std::move(reader)) {
+ELF::ELF(std::unique_ptr<ELFIO::elfio> elf) : m_elf(std::move(elf)) {
 }
 
 ELF ELF::open(const std::string &filename) {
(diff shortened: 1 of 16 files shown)


c_emulator/elf_loader.cpp
```diff
@@ -86,3 +114,27 @@ std::map<std::string, uint64_t> ELF::symbols() const {
   }
   return symbolMap;
 }
+
+void ELF::add_segment(uint64_t address, const std::vector<uint8_t> &data) {
+  // The ELF format doesn't require this but the ELFIO library only lets
+  // you add segments by adding sections first.
+  ELFIO::section *sec = m_elf->sections.add(".text");
+  sec->set_type(ELFIO::SHT_PROGBITS);
+  sec->set_flags(ELFIO::SHF_ALLOC | ELFIO::SHF_EXECINSTR);
```
pmundkur, line 123: If every dumped segment is essentially a `.text` section, will this cause tools like objdump give occasionally scrambled output?  I'm wondering if we should state clearly somewhere what the dumped ELFs will and will not be useful for, and how they diverge from the original.

c_emulator/sail_riscv_sim.cpp
```diff
@@ -28,8 +29,13 @@ void run_model(CLIOptions &opts, ModelImpl &model, uint64_t, const elf_info &elf
   model.register_callback(log_cbs);
 
   do {
-    run_sail(model, opts, loop_detector, elf_info, run_info);
+    run_sail(model, opts, loop_detector, elf_info, run_info, entry);
```
pmundkur, line 32: I wonder if `entry` should be added to `elf_info`; then `dump_memory_to_elf` could take `elf_info` which would also give it the symbols, once we feel we should add them to the dumped ELF.

c_emulator/riscv_platform_if.cpp
```diff
@@ -8,6 +8,28 @@
...
+}
+
+bool PlatformInterface::emulator_write_mem(uint64_t addr_size, sbits addr, const mpz_t n, const lbits data) {
+  return ::emulator_write_mem(addr_size, addr, n, data);
+}
+
+bool PlatformInterface::emulator_write_mem_exclusive(uint64_t addr_size, sbits addr, const mpz_t n, const lbits data) {
+  return ::emulator_write_mem_exclusive(addr_size, addr, n, data);
+}
+
```
pmundkur, line 32: One thing the Sail RTS also has is `emulator_{read,write}_tag` for CHERI.   Once we start working on CHERI integration, the loading and dumping might need some work to take that into account.  You and @arichardson will have a better idea on what will be needed then.  Perhaps some note somewhere would be good.

pmundkur (commented):
I only took a quick look; some initial comments.

c_emulator/elf_loader.cpp
Timmmm: It actually does work - all the tools seem quite happy with duplicate section names. Maybe it would be less confusing to use `.text.0`, `.text.1` etc. though.

c_emulator/sail_riscv_sim.cpp
Timmmm: Yeah I did look at doing that actually but the structure of the code makes it a bit awkward because sometimes `entry` comes from RVFI. It could do with a refactor!

Timmmm:
We discussed this in the meeting and decided it would be nice to support dumping specific address regions too, including getting the addresses to dump from symbols and maybe the PMA regions. This would potentially allow us to remove the signature dumping code. Maybe something like:

```
--dump-memory=modified  // This code.
--dump-memory=all // Based on PMAs.
--dump-memory=0x1000-0x2000,0x4000-0x6000,sig_start-sig_end // Dump those specific regions, supporting symbols too.
```

Does that sound reasonable?

Timmmm:
Also I think I'll split out the memory model as a separate PR.

Issue #1547, opened 2026-02-13
`medeleg` bit 0 should be controlled by `currentlyEnabled(Ext_Zca)`


jordancarlin:
See Spike issue here: https://github.com/riscv-software-src/riscv-isa-sim/issues/2230, https://github.com/riscv-software-src/riscv-isa-sim/pull/2232.

Looks like the Sail model has the same issue:
```sail
private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
  // M-EnvCalls delegation is not supported
  [Mk_Medeleg(v) with MEnvCall = 0b0]
}
```

PR #1550, opened 2026-02-13
Make medeleg.Fetch_Addr_Align read-only zero when Ext_Zca is enabled.

KotorinMinami:
for #1547

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -534,8 +534,12 @@ bitfield Medeleg : bits(64) = {
 }
 
 private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
+  let v = Mk_Medeleg(v);
   // M-EnvCalls delegation is not supported
-  [Mk_Medeleg(v) with MEnvCall = 0b0]
+  [v with
+    MEnvCall = 0b0,
+    Fetch_Addr_Align = if currentlyEnabled(Ext_Zca) then 0b0 else v[Fetch_Addr_Align],
+  ]
 }
 
 register mie     : Minterrupts // Enabled


model/core/sys_regs.sail
```diff
@@ -534,8 +534,12 @@ bitfield Medeleg : bits(64) = {
 }
 
 private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
+  let v = Mk_Medeleg(v);
   // M-EnvCalls delegation is not supported
-  [Mk_Medeleg(v) with MEnvCall = 0b0]
+  [v with
+    MEnvCall = 0b0,
+    Fetch_Addr_Align = if currentlyEnabled(Ext_Zca) then 0b0 else v[Fetch_Addr_Align],
+  ]
```
Timmmm, line 542: Probably worth comments...

```
  [v with
    // M-EnvCalls delegation is not supported
    MEnvCall = 0b0,
    // If compressed instructions are not supported then
    // fetch alignment exceptions are impossible and this
    // must be read-only zero.
    Fetch_Addr_Align = if currentlyEnabled(Ext_Zca) then 0b0 else v[Fetch_Addr_Align],
  ]
```

Also there is a sneaky thing here in that you can disable Zca at runtime by writing misa.C=0, but this function only applies to *writes* to `medeleg`. In other words what should happen if you do:

1. misa.C = 1
2. medeleg.fetch_addr_align = 1
3. misa.C = 0
4. print(medeleg.fetch_addr_align)

I guess it should print 0 but I think it won't currently.

KotorinMinami: I think the corresponding implement should be where the misa field is written, and I will consider editing it accordingly.But I think the right logic is:
- misa.C = 0
- medeleg.fetch_addr_align = 1
- misa.C = 1
- print(medeleg.fetch_addr_align)

And the result should be 0, right?

Timmmm: No, it would be an unspecified legal value (0 or 1) - see the WARL field modulation bit of the spec.

Actually re-reading the spec I am unconvinced that this was their intent anyway.

> For exceptions that cannot occur in less privileged modes, the corresponding medeleg bits should be read-only zero. In particular, medeleg[11] is read-only zero. The medeleg[16] is read-only zero as double trap is not delegatable.

Both examples they give are things that can *never* occur in less privileged modes, no matter the configuration of the hart. In some cases you might not even know, e.g. if you have dynamic PMAs do you have to at runtime make the load/store address misaligned readable/read-only zero whenever someone changes the PMAs to raise address exceptions? That would clearly be silly. Perhaps that's why they say "should" not "must"?

Either way I think we should probably ask for clarification from `riscv-isa-manual`:

1. Does "should" really mean SHOULD here?
2. Was this meant to apply to exceptions that cannot occur in less privileged modes *for the current hart configuration*, or *under any circumstances*?

nadime15: > print(medeleg.fetch_addr_align)

Why does it matter what it prints? There are many CSRs with fields gated by currentlyEnabled(Ext_*). If you print a CSR field immediately after toggling currentlyEnabled(Ext_*) from 1 to 0, the value will read as false as well (e.g. mstatus).

If we need to observe changes in bitfields, we should call legalize_* before.

Timmmm: If disabling misa.C means that medeleg.fetch_addr_align *must* be read-only zero then this code doesn't fix it.

> If we need to observe changes in bitfields, we should call legalize_* before.

Yes the best way to fix it would be a read-legalisation function for `medeleg`. This is exactly the same situation as we've discussed before for `misa` mutability. If you write one CSR field (`misa.C`) and that changes the legal values of another (`medeleg.fetch_addr_align`) then its value becomes *legal but unspecified*. The sensible implementation options are:

1. Modify `medeleg.fetch_addr_align` when `misa.C` is written.
2. Legalise `medeleg.fetch_addr_align` on read.

However, as I said I'm very unconvinced that the spec is actually saying that we need to do this at all.

nadime15: I did not check this, but I think it would generally be better to never read a CSR without legalizing it first. Right now, we often read CSRs directly, which is bad, and in some cases we even access bitfields directly (internally) without cross-checking whether might be gated and entries have changed. 

It would be nice to have something like transparent getter/setter mechanism (e.g. like in Python) that is implicitly (I thing we called about this in the last meeting?) called when reading or writing a field, and which automatically performs the required legalization for CSR.Field.

nadime15:
@Timmmm opened an issue and asked for clarification (https://github.com/riscv/riscv-isa-manual/issues/2675). 

@KotorinMinami Lets see and wait for the answer we get, and then move forward based on that.

PR #1551, opened 2026-02-13
feat(test): create histograms of instruction mnemonics from Sail traces

adingank-qualcomm:
Initial contribution to enable analysis of tests with scripts in a new directory `test/scripts`.
README describes ideas for future development.

### DIFF
--- a/test/scripts/sail-mnemonic-histogram
+++ b/test/scripts/sail-mnemonic-histogram
@@ -1,0 +1,58 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+help() {
+    echo "Location: $0"
+    script_name=$(basename "$0")
+    echo "Usage: $script_name test [dest-dir config]"
+    echo "This script creates a histogram of instruction mnemonics from a simulation trace"
+    echo "of the <test>"
+    echo "in the <dest-dir> directory"
+    echo "by simulating it on the given <config> of the Sail RISC-V model"
+    echo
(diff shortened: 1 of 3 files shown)


test/scripts/sail-mnemonic-histogram
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
```
nadime15, line 3: The model as the README stated is licensed under BSD-2-Clause.

adingank-qualcomm:
Thanks, @nadime15 ! I'll try to find an appropriate home for such work elsewhere.

Timmmm:
Interesting but I think we probably don't want to go in this direction for a few reasons:

1. We'd want something more robust than Bash.
2. We don't want to parse the human-readable log output (see #1460 )
3. It probably makes more sense to use the [ACT4 coverpoints](https://github.com/riscv-non-isa/riscv-arch-test/tree/act4/coverpoints). They do require a commercial simulator unfortunately.

If there's demand for this sort of basic coverage stuff I think it would be better as a separate tool that can ingest Sail traces in machine readable format. E.g. there's probably something [here](https://github.com/sparcians/stf_tools/tree/master/tools) that can do a simple instruction histogram.

Really what we need to do is support some machine readable output formats. We haven't made a lot of progress on that because it's not clear which format(s) to support.

Timmmm:
I'm too slow 😄

adingank-qualcomm:
Thanks @Timmmm for the insights; I was perhaps too trigger happy! :-)

> I'm too slow 😄

PR #1555, opened 2026-02-17
Log Sail exceptions into the execution trace.

### LINKED ISSUE #1553: "Fatal" reserved behavior config options not printing appropriate string
Setting `"xenvcfg_cbie": "Xenvcfg_Fatal"` in the config and then writing 0b10 to `senvcfg.CBIE` results in the following (unhelpful) error instead of the expected `Reserved behavior: xenvcfg.CBIE = 0b10`.

```
Exiting due to uncaught exception: prelude/errors.sail:27.38-27.101
```

pmundkur:
With a thrown Sail exception, the log now terminates with something like:

```
[119] [M]: 0x000000008000200E (0x10A29073) csrrw x0, senvcfg, x5                    main+14
Fatal reserved behavior: xenvcfg.CBIE = 0b10
Exiting due to uncaught exception: prelude/errors.sail:28.2-28.41
```

where previously it would only show
```
[119] [M]: 0x000000008000200E (0x10A29073) csrrw x0, senvcfg, x5                    main+14
Exiting due to uncaught exception: prelude/errors.sail:28.2-28.41
```

This fixes #1553.

### DIFF
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -45,7 +45,7 @@ void ModelImpl::set_reservation_set_size_exp(uint64_t exponent) {
 }
 
 void ModelImpl::print_current_exception() {
-  if (current_exception) {
+  if (current_exception != nullptr) {
     zprint_exception(*current_exception);
   }
 }


model/prelude/errors.sail
```diff
@@ -10,18 +10,22 @@
...
 union exception = {
   Error_not_implemented : string,
-  Error_internal_error  : unit,
+  Error_internal_error  : string,
   Error_reserved_behavior : string,
 }
 
 val not_implemented : forall ('a : Type). string -> 'a
-function not_implemented message = throw(Error_not_implemented(message))
+function not_implemented message = throw Error_not_implemented(message)
```
jordancarlin: While you're modifying this anyway:
```suggestion
function not_implemented(message) = throw Error_not_implemented(message)
```

model/prelude/errors.sail
```diff
@@ -10,18 +10,22 @@
...
 
 val reserved_behavior : forall ('a : Type). string -> 'a
-function reserved_behavior(message) = throw(Error_reserved_behavior("Reserved behavior: " ^ message))
+function reserved_behavior(message) = throw Error_reserved_behavior(message)
+
+function print_exception(e : exception) -> unit =
+  match e {
+    Error_not_implemented(s)   => print_endline("Not implemented error: " ^ s),
+    Error_internal_error(s)    => print_endline("Internal error: " ^ s),
+    Error_reserved_behavior(s) => print_endline("Reserved behavior error: " ^ s),
```
jordancarlin: It's not really an "error" in the reserved behavior case. Maybe the following is better wording:
```suggestion
    Error_reserved_behavior(s) => print_endline("Fatal reserved behavior: " ^ s),
```

jordancarlin:
> With a thrown Sail exception, the log now terminates with:
> 
> ```
> [119] [M]: 0x000000008000200E (0x10A29073) csrrw x0, senvcfg, x5                    main+14
> Reserved behavior error: xenvcfg.CBIE = 0b10
> Exiting due to uncaught exception: prelude/errors.sail:28.2-28.41
> ```
 It would be nice to avoid printing the `prelude/errors.sail:28.2-28.41` part because that is likely to confuse users.

pmundkur:
> It would be nice to avoid printing the `prelude/errors.sail:28.2-28.41` part because that is likely to confuse users.

That's coming from the Sail compiler/runtime handling of `throw`.  We could change how we signal such errors to avoid using `throw` (say using a flag similar to `have_exception`), but that might affect the formal backends that have infrastructure to handle `throw` but would need custom handling for a new flag.

jordancarlin:
> > It would be nice to avoid printing the `prelude/errors.sail:28.2-28.41` part because that is likely to confuse users.
> 
> That's coming from the Sail compiler/runtime handling of `throw`. We could change how we signal such errors to avoid using `throw` (say using a flag similar to `have_exception`), but that might affect the formal backends that have infrastructure to handle `throw` but would need custom handling for a new flag.

That’s what I expected. Not worth changing.

c_emulator/riscv_model_impl.cpp
```diff
@@ -44,6 +44,12 @@ void ModelImpl::set_reservation_set_size_exp(uint64_t exponent) {
   m_reservation_set_addr_mask = ~((1 << exponent) - 1);
 }
 
+void ModelImpl::print_current_exception() {
+  if (current_exception) {
```
Timmmm:
```suggestion
  if (current_exception != nullptr) {
```

(or whatever it is, to avoid the implicit bool coersion)

PR #1556, opened 2026-02-20
Refactor CSR access checking to handle virtual-instruction exceptions.

### LINKED ISSUE #1350: How about having is_CSR_accessible return enum?

pmundkur:
This changes the return type of `is_CSR_accessible` to specify when virtual-instruction exceptions should be thrown instead of illegal-instruction exceptions.  This was difficult with CSR access checkers that were boolean predicates.

Customized checks are now easier for various CSRs such as `seed`, `stimecmp` and CSRs that do not follow HS-qualification (such as those in Smcsrind/Sscsrind).  The check for `seed` now directly matches its specification in the manual.

Default checkers and helpers are collected in `core/csr_access.sail`. These implement HS-qualification checks for hypervisor support.

Further support for hypervisor is pending integration of the `hcounteren` and `henvcfg` CSRs.

The `is_CSR_accessible` name can be changed in a later PR since it now doesn't return a boolean.

This fixes #1350.

### DIFF
--- a/model/core/csr_access.sail
+++ b/model/core/csr_access.sail
@@ -1,0 +1,58 @@
+// =======================================================================================
+// This Sail RISC-V architecture model, comprising all files and
+// directories except where otherwise noted is subject the BSD
+// two-clause license in the LICENSE file.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+// =======================================================================================
+
+// When hypervisor is enabled, `Supervisor` (HS-mode) and
+// `VirtualSupervisor` (VS-mode) privileges are encoded as `0b01` in
+// `privLevel_bits()`, but the CSR addresses for hypervisor and VS
+// CSRs encode the required privilege as `0b10`.  This overrides
+// the encoding in `privLevel_bits()` for the purposes of CSR privilege
+// checks.
+private function privLevel_to_CSR_privbits(p : Privilege) -> nom_priv_bits =
+  match p {
+    User              => 0b00,
+    VirtualUser       => 0b00,
+    // Note that this `if` is not strictly necessary. We could return
+    // 0b10 unconditionally if other checks ensure that hypervisor
+    // CSRs do not exist if H is not enabled.
+    Supervisor        => if currentlyEnabled(Ext_H) then 0b10 else 0b01,
+    VirtualSupervisor => 0b01,
+    Machine           => 0b11,
+  }
+
+function csrAccess(csr : csreg) -> csrRW = csr[11..10]
+function csrPriv(csr : csreg) -> nom_priv_bits = csr[9..8]
+
+// Check that the CSR access is made with sufficient privilege.
+function check_CSR_priv(csr : csreg, p : Privilege) -> bool =
+  privLevel_to_CSR_privbits(p) >=_u csrPriv(csr)
+
+// Check that the CSR access isn't a write to a read-only CSR.
+function check_CSR_access_type(csr : csreg, access : CSRAccessType) -> bool =
+  not((access == CSRWrite | access == CSRReadWrite) & (csrAccess(csr) == 0b11))
+
+// Combine the two checks into a default access check.
+function default_CSR_access_allowed(csr : csreg, p : Privilege, access : CSRAccessType) -> bool =
+  check_CSR_priv(csr, p) & check_CSR_access_type(csr, access)
+
+// This is the default CSR access checker which implements HS-qualified checks
+// by default.  It cannot be used for CSRs that do not use HS-qualification.
+function check_CSR_access(csr : csreg, p : Privilege, access : CSRAccessType, condition : bool) -> result(unit, csr_access_failure) = {
+  if not(condition) then return Err(CSR_Access_Illegal);
+
+  // Stateen defaults to allowing CSRs that are not gated by the xstateen registers.
+  if not(stateen_allows_CSR_access(csr, p, access)) then {
+    return Err(if privLevel_is_virtual(p) & stateen_allows_CSR_access(csr, Supervisor, access)
+               then CSR_Access_Virtual else CSR_Access_Illegal)
+  };
+
+  if default_CSR_access_allowed(csr, p, access)
+  then return Ok(());
+
+  Err(if privLevel_is_virtual(p) & default_CSR_access_allowed(csr, Supervisor, access)
+      then CSR_Access_Virtual else CSR_Access_Illegal)
+}
--- a/model/core/interrupt_regs.sail
+++ b/model/core/interrupt_regs.sail
@@ -102,11 +102,16 @@ mapping clause csr_name_map = 0x302  <-> "medeleg"
 mapping clause csr_name_map = 0x312  <-> "medelegh"
 mapping clause csr_name_map = 0x303  <-> "mideleg"
 
-function clause is_CSR_accessible(0x304, _, _) = true // mie
-function clause is_CSR_accessible(0x344, _, _) = true // mip
-function clause is_CSR_accessible(0x302, _, _) = currentlyEnabled(Ext_S) // medeleg
-function clause is_CSR_accessible(0x312, _, _) = currentlyEnabled(Ext_S) & xlen == 32 // medelegh
-function clause is_CSR_accessible(0x303, _, _) = currentlyEnabled(Ext_S) // mideleg
+function clause is_CSR_accessible(0x304, p, access_type) = // mie
+  check_CSR_access(0x304, p, access_type, true)
+function clause is_CSR_accessible(0x344, p, access_type) = // mip
+  check_CSR_access(0x344, p, access_type, true)
+function clause is_CSR_accessible(0x302, p, access_type) = // medeleg
+  check_CSR_access(0x302, p, access_type, currentlyEnabled(Ext_S))
+function clause is_CSR_accessible(0x312, p, access_type) = // medelegh
+  check_CSR_access(0x312, p, access_type, currentlyEnabled(Ext_S) & xlen == 32)
+function clause is_CSR_accessible(0x303, p, access_type) = // mideleg
+  check_CSR_access(0x303, p, access_type, currentlyEnabled(Ext_S))
 
 function clause read_CSR(0x304) = mie.bits
 function clause read_CSR(0x344) = mip.bits
@@ -168,7 +173,8 @@ private function legalize_sip(m : Minterrupts, d : Minterrupts, v : xlenbits) ->
 }
 
 mapping clause csr_name_map = 0x144  <-> "sip"
-function clause is_CSR_accessible(0x144, _, _) = currentlyEnabled(Ext_S) // sip
+function clause is_CSR_accessible(0x144, p, access_type) = // sip
+  check_CSR_access(0x144, p, access_type, currentlyEnabled(Ext_S))
 function clause read_CSR(0x144) = lower_mip(mip, mideleg).bits
 function clause write_CSR(0x144, value) = { mip = legalize_sip(mip, mideleg, value); Ok(lower_mip(mip, mideleg).bits) }
 
@@ -188,6 +194,7 @@ private function legalize_sie(m : Minterrupts, d : Minterrupts, v : xlenbits) ->
 }
 
 mapping clause csr_name_map = 0x104  <-> "sie"
-function clause is_CSR_accessible(0x104, _, _) = currentlyEnabled(Ext_S) // sie
+function clause is_CSR_accessible(0x104, p, access_type) =  // sie
+  check_CSR_access(0x104, p, access_type, currentlyEnabled(Ext_S))
 function clause read_CSR(0x104) = lower_mie(mie, mideleg).bits
 function clause write_CSR(0x104, value) = { mie = legalize_sie(mie, mideleg, value); Ok(lower_mie(mie, mideleg).bits) }
--- a/model/core/csr_begin.sail
+++ b/model/core/csr_begin.sail
@@ -15,9 +15,17 @@ scattered mapping csr_name_map
 function csr_name(csr : csreg) -> string = csr_name_map(csr)
 overload to_str = {csr_name}
 
-// returns whether a CSR exists
-val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> bool
+// CSR access control
+enum csr_access_failure = { CSR_Access_Illegal, CSR_Access_Virtual }
+
+// returns the result of a CSR access check
+// TODO: change the function name
+val is_CSR_accessible : (csreg, Privilege, CSRAccessType) -> result(unit, csr_access_failure)
 scattered function is_CSR_accessible
+
+// declaration of functions in `extensions/Stateen/stateen_access_checks.sail`.
+val stateen_allows_CSR_access : (csreg, Privilege, CSRAccessType) -> bool
+scattered function stateen_allows_CSR_access
 
 // returns the value of the CSR if it is defined
 val read_CSR : csreg -> xlenbits
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -124,7 +124,8 @@ private function legalize_misa(m : Misa, v : xlenbits) -> Misa = {
 }
 
 mapping clause csr_name_map = 0x301  <-> "misa"
-function clause is_CSR_accessible(0x301, _, _) = true // misa
+function clause is_CSR_accessible(0x301, p, access_type) = // misa
+  check_CSR_access(0x301, p, access_type, true)
 function clause read_CSR(0x301) = misa.bits
 function clause write_CSR(0x301, value) = { misa = legalize_misa(misa, value); Ok(misa.bits) }
 
@@ -269,8 +270,10 @@ register mstatus : Mstatus = {
 mapping clause csr_name_map = 0x300  <-> "mstatus"
 mapping clause csr_name_map = 0x310  <-> "mstatush"
 
-function clause is_CSR_accessible(0x300, _, _) = true // mstatus
-function clause is_CSR_accessible(0x310, _, _) = xlen == 32 // mstatush
+function clause is_CSR_accessible(0x300, p, access_type) = // mstatus
+  check_CSR_access(0x300, p, access_type, true)
+function clause is_CSR_accessible(0x310, p, access_type) = // mstatush
+  check_CSR_access(0x310, p, access_type, xlen == 32)
 
 function clause read_CSR(0x300) = mstatus.bits[xlen - 1 .. 0]
 function clause read_CSR(0x310 if xlen == 32) = mstatus.bits[63 .. 32]
@@ -331,8 +334,10 @@ mapping clause csr_name_map = 0x747  <-> "mseccfg"
 mapping clause csr_name_map = 0x757  <-> "mseccfgh"
 
 // "mseccfg exists if Zkr is implemented, or if it is required by other processor features."
-function clause is_CSR_accessible(0x747, _, _) = currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp) // mseccfg
-function clause is_CSR_accessible(0x757, _, _) = (currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp)) & xlen == 32 // mseccfgh
+function clause is_CSR_accessible(0x747, p, access_type) = // mseccfg
+  check_CSR_access(0x747, p, access_type, currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp))
+function clause is_CSR_accessible(0x757, p, access_type) =  // mseccfgh
+  check_CSR_access(0x757, p, access_type, (currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp)) & xlen == 32)
 
 function clause read_CSR(0x747) = mseccfg.bits[xlen - 1 .. 0]
 function clause read_CSR(0x757 if xlen == 32) = mseccfg.bits[63 .. 32]
@@ -442,12 +447,12 @@ mapping clause csr_name_map = 0x30A  <-> "menvcfg"
 mapping clause csr_name_map = 0x31A  <-> "menvcfgh"
 mapping clause csr_name_map = 0x10A  <-> "senvcfg"
 
-function clause is_CSR_accessible(0x30A, _, _) = currentlyEnabled(Ext_U) // menvcfg
-function clause is_CSR_accessible(0x31A, _, _) = currentlyEnabled(Ext_U) & (xlen == 32) // menvcfgh
-
-// This should ideally also check `xstateen.ENVCFG`, but due to module
-// interdependency issues, that is handled by `stateen_allows_CSR_access`.
-function clause is_CSR_accessible(0x10A, _, _) = currentlyEnabled(Ext_S) // senvcfg
+function clause is_CSR_accessible(0x30A, p, access_type) = // menvcfg
+  check_CSR_access(0x30A, p, access_type, currentlyEnabled(Ext_U))
+function clause is_CSR_accessible(0x31A, p, access_type) = // menvcfgh
+  check_CSR_access(0x31A, p, access_type, currentlyEnabled(Ext_U) & xlen == 32)
+function clause is_CSR_accessible(0x10A, p, access_type) = // senvcfg
+  check_CSR_access(0x10A, p, access_type, currentlyEnabled(Ext_S))
 
 function clause read_CSR(0x30A) = menvcfg.bits[xlen - 1 .. 0]
 function clause read_CSR(0x31A if xlen == 32) = menvcfg.bits[63 .. 32]
@@ -498,7 +503,8 @@ bitfield Mcause : xlenbits = {
 }
 register mcause : Mcause
 mapping clause csr_name_map = 0x342  <-> "mcause"
-function clause is_CSR_accessible(0x342, _, _) = true // mcause
+function clause is_CSR_accessible(0x342, p, access_type) = // mcause
+  check_CSR_access(0x342, p, access_type, true)
 function clause read_CSR(0x342) = mcause.bits
 function clause write_CSR(0x342, value) = { mcause.bits = value; Ok(mcause.bits) }
 
@@ -545,8 +551,10 @@ register mscratch : xlenbits
 mapping clause csr_name_map = 0x343  <-> "mtval"
 mapping clause csr_name_map = 0x340  <-> "mscratch"
 
-function clause is_CSR_accessible(0x343, _, _) = true // mtval
-function clause is_CSR_accessible(0x340, _, _) = true // mscratch
+function clause is_CSR_accessible(0x343, p, access_type) = // mtval
+  check_CSR_access(0x343, p, access_type, true)
+function clause is_CSR_accessible(0x340, p, access_type) = // mscratch
+  check_CSR_access(0x340, p, access_type, true)
 
 function clause read_CSR(0x343) = mtval
 function clause read_CSR(0x340) = mscratch
@@ -574,7 +582,8 @@ private function legalize_scounteren(_c : Counteren, v : xlenbits) -> Counteren 
 
 register scounteren : Counteren
 mapping clause csr_name_map = 0x106  <-> "scounteren"
-function clause is_CSR_accessible(0x106, _, _) = currentlyEnabled(Ext_S) // scounteren
+function clause is_CSR_accessible(0x106, p, access_type) = // scounteren
+  check_CSR_access(0x106, p, access_type, currentlyEnabled(Ext_S))
 function clause read_CSR(0x106) = zero_extend(scounteren.bits)
 function clause write_CSR(0x106, value) = { scounteren = legalize_scounteren(scounteren, value); Ok(zero_extend(scounteren.bits)) }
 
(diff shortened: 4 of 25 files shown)


pmundkur:
It looks like we will have CI failures due to apt.llvm.org until https://github.com/llvm/llvm-project/issues/182462 is fixed.  If this persists, we might need an alternative to using `https://apt.llvm.org/llvm.sh`

pmundkur:
This is more similar to #1357 than #1402.

model/core/csr_access.sail
```diff
@@ -0,0 +1,84 @@
...
+  Err(if privLevel_is_virtual(p) & is_hs_qualified then CSR_Access_Virtual else CSR_Access_Illegal);
+}
+
+// This is the default CSR access checker.
+function check_CSR_access(csr : csreg, p : Privilege, access_type : CSRAccessType, condition : bool) -> result(unit, csr_access_failure) = {
+  if csrPriv(csr) == 0b11
+  then check_machine_CSR_access(csr, p, access_type, condition)
+  // CSRs at other privileges need HS-qualification checks by default.
+  else check_hs_qualified_CSR_access(csr, p, access_type, condition)
+}
```
nadime15: I need to double/tripple check this but is the following not enough?

```
check_CSR_access(csr : csreg, p : Privilege, access_type : CSRAccessType, condition : bool) -> result(unit, csr_access_failure) = {
  if not(condition) then return Err(CSR_Access_Illegal);
  if default_CSR_access_allowed(csr, p, access_type) then return Ok(());
  if inVirtualMode() & default_CSR_access_allowed(csr, Supervisor, access_type)
  then Err(CSR_Access_Virtual)
  else Err(CSR_Access_Illegal)
}
```

check_stateen_guarded_CSR_access() could call check_CSR_access and simplify the whole function to something like:

```
function check_stateen_guarded_CSR_access(csr : csreg, p : Privilege, access_type : CSRAccessType, condition : bool) -> result(unit, csr_access_failure) = {
  check_CSR_access(csr, p, access_type, condition & stateen_allows_CSR_access(csr, p, access_type))
}
```

pmundkur: Yes, I think your version of `check_CSR_access` is equivalent.  I just split out the machine mode version since its simpler, but we could have a unified version.

Your  version of the stateen-guarded checker is _not_ equivalent: the staten checks have to be done at both `priv` and `Supervisor`, and your version only does it at `priv`.

nadime15: > Your version of the stateen-guarded checker is not equivalent: the staten checks have to be done at both priv and Supervisor, and your version only does it at priv.

I am not a big fan of moving from a centralized solution to a decentralized one. I think the previous solution was quite clean, with all checks bundled in check_CSR.

I would prefer to keep it consistent with what we already have in check_CSR and group all checks together centrally:

```
function check_CSR(csr : csreg, p : Privilege, access_type : CSRAccessType) -> result(unit, csr_access_failure) = {
  if not(stateen_allows_CSR_access(csr, p, access_type)) then {
    if privLevel_is_virtual(p) & stateen_allows_CSR_access(csr, Supervisor, access_type)
    then return Err(CSR_Access_Virtual)
    else return Err(CSR_Access_Illegal)
  };
  is_CSR_accessible(csr, p, access_type) // (Calls the proposed version above)
}
``` 

This handles the stateen check at both p and Supervisor correctly, and individual CSR clauses don't need to know about stateen at all (which is also easier to maintain in case someone wants to use a smaller core), they just use plain check_CSR_access. The stateen HS-qualification check is handled once centrally, avoiding the need for a separate helper entirely.

pmundkur: sounds good to me, pushed.

arichardson (commented):
I am not sure I understand why this is better than https://github.com/riscv/sail-riscv/pull/1402? The other PR has the nice side-effect that you can actually figure out why an access was denied rather than just returning true/false. And having that in the logs would be really nice.

pmundkur:
> I am not sure I understand why this is better than #1402? The other PR has the nice side-effect that you can actually figure out why an access was denied rather than just returning true/false. And having that in the logs would be really nice.

This PR focuses on supporting virtual-instruction exceptions cleanly, which involves implementing the HS-qualified check for hypervisor.   For that, a CSR access check needs two kinds of failures/exceptions, illegal-instruction or virtual-instruction.  Access checking functions that returned just booleans got in the way, since all checks needed to be implemented in terms of those boolean functions.  See the handling of the `seed` CSR in `extensions/Zkr/zkr_control.sail` in the two PRs for an example; it's not clear to me how a virtual-instruction exception would be indicated in #1402.  In this PR, the code corresponds to the spec fairly directly.  Another example would be `srmcfg` in `extensions/Ssqosid/ssqosid.sail`.

Also, from the point of view of including the Sail source in the manual, having the debug logging support adds a fair amount of noise.   We could still add it if that's not an issue; I think it could be easier on top of this PR.

arichardson:
> > I am not sure I understand why this is better than #1402? The other PR has the nice side-effect that you can actually figure out why an access was denied rather than just returning true/false. And having that in the logs would be really nice.
> 
> This PR focuses on supporting virtual-instruction exceptions cleanly, which involves implementing the HS-qualified check for hypervisor. For that, a CSR access check needs two kinds of failures/exceptions, illegal-instruction or virtual-instruction. Access checking functions that returned just booleans got in the way, since all checks needed to be implemented in terms of those boolean functions. See the handling of the `seed` CSR in `extensions/Zkr/zkr_control.sail` in the two PRs for an example; it's not clear to me how a virtual-instruction exception would be indicated in #1402. In this PR, the code corresponds to the spec fairly directly. Another example would be `srmcfg` in `extensions/Ssqosid/ssqosid.sail`.
> 
> Also, from the point of view of including the Sail source in the manual, having the debug logging support adds a fair amount of noise. We could still add it if that's not an issue; I think it could be easier on top of this PR.

Thanks for the explanation. I personally don't feel strongly about this either way so am more than happy to defer to your judgement.

In #1402, isn't support for virtual instruction exceptions already included? It's just not been added to the specific cases yet since it's refactoring without functional changes? It already returns a CSRAccessResult value:

```
union CSRAccessResult = {
  CSRAccess_Success : unit,

  // Accessing the CSR is illegal currently and should raise Illegal_Instruction Exception
  CSRAccess_Illegal : CSRAccessIllegalReason,

  // Like CSRAccess_Illegal, but raise Virtual_Instruction Exception
  CSRAccess_Virtual : unit,
}
```

pmundkur:
> In #1402, isn't support for virtual instruction exceptions already included? It's just not been added to the specific cases yet since it's refactoring without functional changes? 

You're right, the support is included.  Although I think it could be made to work, I think the overloaded `&` could be confusing when applied to `CSRAccess_Illegal()` and `CSRAccess_Virtual()`, where an order would need to be picked, and then remembered everywhere that `&` is used.  

I think the additional helpers for debugging also get in the way of including it in the manual.

model/core/interrupt_regs.sail
```diff
@@ -102,11 +102,16 @@ mapping clause csr_name_map = 0x302  <-> "medeleg"
 mapping clause csr_name_map = 0x312  <-> "medelegh"
 mapping clause csr_name_map = 0x303  <-> "mideleg"
 
-function clause is_CSR_accessible(0x304, _, _) = true // mie
-function clause is_CSR_accessible(0x344, _, _) = true // mip
-function clause is_CSR_accessible(0x302, _, _) = currentlyEnabled(Ext_S) // medeleg
-function clause is_CSR_accessible(0x312, _, _) = currentlyEnabled(Ext_S) & xlen == 32 // medelegh
-function clause is_CSR_accessible(0x303, _, _) = currentlyEnabled(Ext_S) // mideleg
+function clause is_CSR_accessible(0x304, p, access_type) = // mie
+  check_CSR_access(0x304, p, access_type, true)
```
Timmmm, line 106: It's very unfortunate that you have to repeat so much for the default case... 

How about this instead:

1. Leave `is_CSR_accessible` as it was. It returns `false` for Illegal *and* Virtual instruction exceptions.
2. Make `check_CSR_access` the outer function. Instead of `if not(condition) then return Err(CSR_Access_Illegal);` do

```
if not(is_CSR_accessible(csr, priv, access_type) then
  // Illegal or virtual instruction exception.
  return Err(if csr_exception_is_virtual(csr, priv, access_type) then CSR_Access_Virtual else CSR_Access_Illegal);
```

Then have a second scattered function for `csr_exception_is_virtual` (naming TBD!) which returns a bool - *assuming* `is_CSR_accessible()` returned false, should the exception be virtual? E.g. for `seed` it would be

```
function clause is_CSR_accessible(0x015, priv, access_type) =
  currentlyEnabled(Ext_Zkr) &
  // Read-only access is not allowed.
  (access_type != CSRRead) &
  (match priv {
    Machine => true,
    Supervisor => mseccfg[SSEED] == 0b1,
    User => mseccfg[USEED] == 0b1,
    VirtualUser => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
  })

function clause csr_exception_is_virtual(0x015, priv, access_type) = 
  is_priv_virtual(priv) & access_type == ReadWrite
```

Then we have a default `csr_exception_is_virtual(_, _, _) = false`

This is maybe mildly more awkward for CSRs with custom conditions for virtual instruction exceptions... but it's vastly simpler for everything else, and also has the not insignificant advantage of making the hypervisor code easy to ignore for people reading the code that don't care about hypervisor stuff.

The only real downside (having not fully worked through things!) I can see is that it's easy to forget to add it when you're adding a new CSR. But I think it's probably worth it.

What do you think?

pmundkur: I agree that the repetition in this PR is unfortunate.  

I'd prefer an approach that captures the spec-defined `HS-qualified` concept explicitly _somewhere_.  That will help from the perspective of including the Sail source in the spec.  In this PR, at least there are a couple of lines in `core/csr_access.sail` that do correspond to it.  (I had comments at those places, but I guess I lost them along the way; I could add them back.)

I suspect that the `csr_exception_is_virtual()` approach will have its own repetition to capture HS-qualification all over the place.  And I'm not sure it will work well that way, since you need to do the full access using S-mode, including any `stateen` checks, before deciding that the exception is virtual; this is harder with different functions performing different parts of the access check.

For CSRs like `stimecmp` in Sstc, having the check be in one function seems simpler than splitting it across two boolean functions `is_CSR_accessible()` and `csr_exception_is_virtual()`.

Timmmm: Sorry, to be clear I think the rest of `check_CSR_access` where it does the stateen check and so on should still exist. So it should only be a small number of CSRs that need special logic. Maybe I should make a PR... (another one? :-D)

pmundkur: > Maybe I should make a PR... (another one? :-D)

Sure, if you have the time!  It's good to explore this design space a little bit.

nadime15: Any update on this, @Timmmm?

nadime15: Here is the link to the alternative approach: 

https://github.com/nadime15/sail-riscv/blob/bf250ccb51c6fb383609c58cbb5fe31672af3925/model/sys/sys_control.sail#L52-L80

Special cases, as far as I am aware are:
```
* satp, 
* counter CSRs, 
* hpm counters,
* seed (which still need to be added to the list)
```

pmundkur: I'm still failing to see the advantage of this approach. The only one I can see is that existing `is_CSR_accessible` clauses do not need to be modified.  

But the list of exceptions will keep growing (this list is already missing `ssp`, `srmcfg`, `scountovf`, which are already in the model, and I haven't looked at extensions that are yet to be added).  I'm also not sure how `seed` can be handled cleanly using this approach.

This PR puts all the logic for a CSR into one function (or uses local helpers if needed), and leaves the top-level pretty clean, with no handling of a long list of special cases.

Timmmm: I was thinking that `match` on line 60 would be a second scattered function, then it doesn't really matter how long the list gets. We can include `access_type` in the function to handle `seed`'s weirdness.

```
function check_CSR(csr : csreg, p : Privilege, access_type : CSRAccessType) -> CSRCheckResult =
  if check_CSR_priv(csr, p)
     & check_CSR_access(csr, access_type)
     & is_CSR_accessible(csr, p, access_type)
     & stateen_allows_CSR_access(csr, p, access_type) then {
    CSR_Check_OK()
  } else if (p == VirtualSupervisor | p == VirtualUser) & is_CSR_exception_virtual(csr, p, access_type) then {
    CSR_Virtual()
  } else {
    CSR_Illegal()
  }
  ...

function clause is_CSR_exception_virtual(0x015, _p, access_type) =
  mseccfg[SSEED] == 0b1 & access_type == CSRReadWrite
```

Seems fairly straightforward?

pmundkur: It still seems rather pointless to have _two_ scattered functions to decide what to return in `{CSR_Check_OK, CSR_Virtual, CSR_Illegal}`, rather than a single one that decides the return value directly.  These two scattered functions would be accessing the same state twice (e.g. `mseccfg[SSEED]` for `seed`).

nadime15: The advantage of this approach is that it is much easier to follow, implement and follows separation of concern principles. In the current PR, the meaning of `is_CSR_accessible` gets somewhat lost, it ends up doing too much. 

In my opinion, `is_CSR_accessible` should only be a gate, it answers whether an access, given the current state of the machine, is permitted or not. Pushing exception classification into it mixes two separate concerns (you basically overloading the function). The name itself suggests this, you would not expect a function called `is_CSR_accessible` to also decide which exception to raise (we could obviously rename it if needed).

Another advantage of our approach is that basically no existing function needs to be modified. The current PR approach has to accommodate functions due to its new return type, and it gets a bit awkward because you end up with a mix of functions returning booleans and now one returning an exception or result type. Generally speaking, it makes more sense to decide what kind of exception to return one level higher, in one dedicated place.

> 

> But the list of exceptions will keep growing (this list is already missing ssp, srmcfg, scountovf, which are already in the model, and I haven't looked at extensions that are yet to be added). I'm also not sure how seed can be handled cleanly using this approach.

The same argument about a growing list applies to stateen as well. Besides that stateen access checking is already self-contained in stateen_allows_CSR_access. Nobody expects that function to also decide which exception to raise. A scattered or regular function, say, `is_CSR_exception_virtual` would follow the exact same pattern, it is a dedicated, self-contained function that answers one specific question, given that access was denied and we are in a virtual privilege mode, should the exception be virtual? Each function owns its own concern, and the hypervisor-specific logic is nicely isolated rather than spread across is_CSR_accessible clauses everywhere.

Something like: 

```
function check_CSR(csr : csreg, p : Privilege, access_type : CSRAccessType) -> CSRCheckResult =
  if check_CSR_priv(csr, p)
     & check_CSR_access(csr, access_type)
     & is_CSR_accessible(csr, p, access_type)
     & stateen_allows_CSR_access(csr, p, access_type) then {
    CSR_Check_OK()
  } else if (p == VirtualSupervisor | p == VirtualUser) 
          & is_CSR_exception_virtual(csr, p, access_type) then {
    CSR_Virtual()
  } else {
    CSR_Illegal()
  }
```

>  I'm also not sure how seed can be handled cleanly using this approach.

As I said last monday, seed is only special when the hypervisor is active and you are in VS or VU mode, which is exactly the case is_CSR_exception_virtual would handle. You can take the same approach as with the performance counters and define a dedicated clause that returns whether a virtual or illegal instruction exception should be raised. That is actually the only moment it matters, and it keeps the logic self-contained.

pmundkur: As this PR message pretty clearly says: "The is_CSR_accessible name can be changed in a later PR since it now doesn't return a boolean."  (It also says this in a TODO in the code itself: https://github.com/pmundkur/sail-riscv/blob/74874131ba02a7242ac3714218295cdfc7a6655d/model/core/csr_begin.sail#L22)

So I wouldn't put too much in the name `is_CSR_accessible`. 

There is already a function that follows the approach used in this PR: `feature_enabled_for_priv`, which was introduced in #1597.  Following the logic above, is the argument now that this function should be split into two, for "separation of concerns" reasons?

I don't think the logic is as self contained as is being claimed.  Like I said, the `mseccfg` register would be checked in two functions for the `seed` CSR.  I find _that_ confusing.  The same would apply for `ssp`, `scountovf`, counters and perhaps others.   This might also complicate RVWMO modelling; for e.g. there would be two reads of the `mseccfg` register when checking a `seed` access where one read would suffice.

The `stateen_allows_CSR_access` is an internal implementation detail of `check_CSR_access`; it can have a boolean return or something else, it is an internal concern of the CSR access check code and not exposed to the rest of the code.

nadime15: Am I missing something here? Checking `is_CSR_accessible` has no side effects, we dont legalize. The double read in the access check path is just reading raw register state like mseccfg[SSEED]. That is not a legalizing read, it is just a field check with no side effects. And if an access fails and we need to identify the correct exception in VS/VU mode, no read_CSR is performed anyway, legalizing reads only happen there, after access is granted.

Also, and I say this as someone who is not an RVWMO expert, does RVWMO even apply here? The name itself says RISC-V Weak Memory Ordering, it concerns memory operations and their ordering guarantees. A combinational field check on a CSR register state is not a memory operation. So the RVWMO concern probably dont apply at all in this context.

(Looking at the RVWMO spec, it explicitly defines the memory model for regular main memory operations only.)

Regeding seed, I would argue the seed case is simply not worth redesigning the code to this extent. What you describe only occurs when Zkr is enabled, the hypervisor is enabled, the hart is in VS or VU mode, and mseccfg[SSEED] == 0b1. Thats a very specific combination of conditions for what is a rare concern.

pmundkur: Sorry, I overthought it, I don't think RVWMO is a concern; two reads of the CSR are equivalent to a single CSR read for the instruction's read footprint.

Could you implement the two function "separation of concerns" version as completely (handling all the CSRs) as this PR, so that we have something concrete to compare?

trdthg: Today someone told me: "Reading sail-riscv is much faster than checking the riscv-isa-manual," and I couldn't agree more.

I also prefer keeping the code as simply/shorter as possible, but things change when we consider integrating these segments into the manual. I'm not sure how the Sail code extraction process works (Is there a link?) , but my point is that I prefer an approach that looks better when integrated into the manual.

In this regard, I still prefer this PR. We have only one entry point: `function clause single_source_of_csr_accessible(csr, ...)`.

Timmmm: I think the split style would actually make it *easier* to insert into the manual. The behaviour of most CSRs is already split into "this is how it works" in the Machine chapter, or whatever, and then "by the way if you have hypervisor also do this" in the Hypervisor chapter.

nadime15: > Could you implement the two function "separation of concerns" version as completely (handling all the CSRs) as this PR, so that we have something concrete to compare?

Sure! I will do that today or in the next few days.

nadime15: Here is an example of the scattered function approach: 

https://github.com/riscv/sail-riscv/compare/master...nadime15:sail-riscv:csr-virtual-exception-refactor

I covered satp, stateen, hpmcounter, seed, etc. 

I have to say I quite like the approach. The exception classification logic is very easy to read, to follow and extensible. Also, since most CSRs simply raise illegal instruction exceptions (the default clause takes care of it), the actual coverage is limited to a handful of Supervisor CSRs (+ counters, timers, etc.). 

I think having a dedicated function that tells the user/reader when a virtual instruction is raised makes it much clearer to understand what the table in the hypervisor spec is actually trying to say (for example, by simply grepping the function and checking the corresponding cases).

pmundkur: I was also hoping for a more complete implementation (e.g. including `ssp`) to help evaluate the two approaches.

I think there is a lot of redundancy here: at the top level in `check_CSR` there is
```
else if (p == VirtualSupervisor | p == VirtualUser)
          & is_CSR_exception_virtual(csr, p, access_type)
```
and then the privilege is checked again in almost every `is_CSR_exception_virtual` clause, which is of the form
```
function clause is_CSR_exception_virtual(0x61D, p, _) = p == VirtualSupervisor
```
Why repeat this privilege check?

I also don't think this is correct.  There are at least two classes of errors:
- When accessing the top-half `*h` CSRs registers on RV64, the exception should be illegal-instruction, regardless of the privilege.  This could erroneously generate virtual-instruction exceptions, e.g. for `cycleh` on RV64 when accessed with a virtual privilege.
- When accessing CSRs that are gated by an extension being supported, e.g. `hstaten*` by S{ms}staten, accessing those CSRs should generate an illegal-instruction exception if that extension is configured to be not supported, regardless of the privilege.

radimkrcmar: Continuing the discussion from call: the seed actually is implementing the generic HS-qualified rule.
Seed always traps in VS-mode, and if the access would have succeeded in HS-mode then it throws virtual, else illegal instruction exception.  (It's bad the the manual makes it look like a normative statement when it's non-normative outcome of other rules.)

We will have a few exceptions to the HS-qualified rule, but I think it would be better to only describe those deviations, and not re-implement the rule for each CSR.

radimkrcmar: To illustrate the idea, I think something like this expresses the HS-qualified rule for CSRs: 
```
else if ((p == VirtualSupervisor | p == VirtualUser) {
  match CSR_hs_qualified_rule_override(csr, mode, access_type) {
    None -> check_CSR(base_csr(csr), Supervisor, access_type) == CSR_Check_Ok() ? CSR_Virtual() : CSR_Illegal()
    Some(x) -> x
  }
```
The `CSR_hs_qualified_rule_override` could handle CSRs like vtype (never virtual) or vsireg (always virtual), and `base_csr` would remap high CSRs to low on RV32.
For satp, we could either proceed with mstatus.TVM=0 before doing the `check_CSR` (more correct, but also more work), or just implement an override.

Note that the HS-qualified rule is supposed to apply to all encodings that can throw illegal-instruction exceptions, not just CSR accesses, so we could theoretically generalize it further, but that might be overdoing it...

pmundkur: @radimkrcmar What do you make of the way HS-qualification is done in this PR, in `model/core/csr_access.sail:check_CSR_access`.  That is the default handler, for cases that don't need extra checks e.g. in `mseccfg`, `*envcfg`, etc.

nadime15: @radimkrcmar I dont know if you saw that (one of the first links in this discussion but I forget to mention this in the meeting today) but my branch of the hypervisor implementation basically has something similar, I forgot to include this because it was just a quick example I made on friday.

See line 75:

https://github.com/nadime15/sail-riscv/blob/bf250ccb51c6fb383609c58cbb5fe31672af3925/model/sys/sys_control.sail#L52-L80

So my approach could be updated and include your suggestion or we could further define and it could look like:

```
function check_CSR(csr : csreg, p : Privilege, access_type : CSRAccessType) -> CSRCheckResult = {
  if check_CSR_priv(csr, p)
     & check_CSR_access(csr, access_type)
     & is_CSR_accessible(csr, p, access_type)
     & stateen_allows_CSR_access(csr, p, access_type) then {
    CSR_Check_OK()
  } else if (p == VirtualSupervisor | p == VirtualUser) then {
    if is_CSR_exception_virtual(csr, p, access_type) then {
      CSR_Virtual()
    } else if check_CSR(csr, Supervisor, access_type) == CSR_Check_OK() then {
      CSR_Virtual()
    } else {
      CSR_Illegal()
    }
  } else {
    CSR_Illegal()
  }
}
```

Or another option, which basically mixes both of ours would be, which I like even more:

```

function check_CSR(csr : csreg, p : Privilege, access_type : CSRAccessType) -> CSRCheckResult = {
  if check_CSR_priv(csr, p)
     & check_CSR_access(csr, access_type)
     & is_CSR_accessible(csr, p, access_type)
     & stateen_allows_CSR_access(csr, p, access_type) then {
    CSR_Check_OK()
  } else if (p == VirtualSupervisor | p == VirtualUser)
          & is_CSR_exception_virtual(csr, p, access_type) then {
    CSR_Virtual()
  } else {
    CSR_Illegal()
  }
}

// Special cases that deviate from the HS-qualified rule
function clause is_CSR_exception_virtual(0x180, p, _) = p == VirtualSupervisor & hstatus[VTVM] == 0b1
function clause is_CSR_exception_virtual(0x015, _, access_type) = mseccfg[SSEED] == 0b1 & access_type == CSRReadWrite
// other special cases
// ..

// Default: implement the general HS-qualified rule
function clause is_CSR_exception_virtual(csr, _, access_type) = 
  check_CSR(csr, Supervisor, access_type) == CSR_Check_OK()
```

The 2. approach would make things even more clear, because if I am not mistaken this would cover any CSR where access fails in VS/VU mode but would succeed in Supervisor mode, this covers H-extension CSRs like hgatp, hstateen, VS CSRs like vsatp, and any future CSRs automatically without needing explicit clauses

radimkrcmar: @pmundkur It's correct, although I think it's a shame that the access checks due to *stateen vs. other state are not abstracted consistently:  it shouldn't matter why the CSR traps.

If we keep the `is_CSR_accessible(csr, mode, access_type) -> bool` function that says when to trap, we can easily generalize stimecmp, ssp, srmcfg, seed, etc. as well.

e.g. accessing stimecmp
if not M-mode & menvcfg.STCE=0 -> trap
if U/VU-mode & senvcfg.STCE=0 -> trap
if VU/VS-mode & henvcfg.STCE=0 -> trap

It's the normal HS-qualified rule to figure out which trap it is.   You're using the Supervisor trick with `stateen_allows_CSR_access(csr, Supervisor, access)` and I think it makes sense to generalize it with `is_CSR_accessible(csr, Supervisor, access)`.

Also, there aren't that many exceptions to the HS-qualified rule, and I think it would be nice to very explicitly mark them as special cases: separating the trap/no-trap vs illegal/virtual exception logic seems clearer.
(I wouldn't do code transformations that could have been made by an optimizing compiler at the cost of code clarity.)

And last, Sail doesn't have syntactic sugar to turn
```
function foo(a,b,c,d,e,f,g) -> bar(a,b,c,d,e,f,g,h)
```
into something like
```
function foo(...) -> bar (..., h)
```
so it looks ugly. :)

radimkrcmar: @nadime15 I like the second approach.  Looks clean and handles all I can think of.

nadime15: I addressed some of the feedback from this morning (see commit message). Here’s a link to the updated version:

https://github.com/riscv/sail-riscv/compare/master...nadime15:sail-riscv:csr-virtual-exception-refactor

The main change is that I added an HS-qualifier check. Besides that, I slightly modified `check_CSR` as well to catch cases that could previously return a virtual instruction exception. For example, in the old version, a VU mode csr access failing due to `stateen_allows_CSR_access` could be mistakenly classified as a virtual instruction exception because of `is_CSR_exception_virtual` combined with the HS-qualifier rule.

pmundkur: Could you create a PR so that we can comment on the code directly?

nadime15:
@Timmmm is this one good to go?

pmundkur:
@nadime15 I think you meant your #1624?

nadime15:
Ah ups! Yes sorry, I mean #1624! :D

pmundkur:
Closing since #1624 has been merged.

Issue #1557, opened 2026-02-21
Changes to support for symbolic execution


msuadOf:
I'm currently interested in RISC-V and the symbolic execution engine `isla`, but it seems like nobody has been paying attention to this for quite some time. Some changes in sail-riscv have caused issues with symbolic execution. I'm not very familiar with sail-riscv, so I was wondering if anyone could help me with some modifications.

msuadOf:
the change in [commit:14172f9b6d model/prelude.sail](https://github.com/riscv/sail-riscv/commit/14172f9b6d78830b7c61cf4570b1a9209048fd5d#diff-28bd1858b2cf208fbd0aed9363c6cb0eb8149603284c78a93f33f13b0edc5cc7) uses an intermediate bitvector whose size depends on the shift, which doesn't translate into SMT. And passes this parameter to the `sign_extend` function.

My requirement is, when calling `sign_extend`, can the first argument of `sign_extend`, which is an implicit parameter, always use a fixed value? For example, `xlen` (which can be fixed to 32/64 during the configuration).

Anyway, thank you for your time!

Timmmm:
Hmm if that doesn't work I'm surprised that's the only case you've run into. Also when you say "it doesn't translate into SMT" what do you mean exactly? Perhaps the Sail compiler could just support this case...

msuadOf:
This shouldn't be the only situation I've encountered; it's just that my test case got stuck at this point. I need to fix this bug before I can move forward and see what other bugs lie ahead.

"It doesn't translate into SMT" means that I'm building a symbolic execution engine based on Isla, which uses Z3 (that's what I meant by "SMT"). Both sign_extend and zero_extend have corresponding functions in Z3, but the extension length cannot be a symbolic value. In the signature val sign_extend : forall 'n 'm, 'm >= 'n. (implicit('m), bits('n)) -> bits('m), that implicit parameter 'm of sign_extend is a concrete, determined value in Sail. In the older version, it was used as an implicit parameter with the value xlen*2.

Timmmm:
Right but I mean what actually happens when you try to translate it to smt2? Does the Sail compiler give an error? Does it give incorrect output? Does it just skip that function?

Alasdair:
I think the solution is probably to replace the function with a primitive that directly generates the `bvashr` SMT primitive.

bacam:
[...]

PR #1558, opened 2026-02-21
Support for Symbolic Excution 

msuadOf:
As mentioned in [#1557](https://github.com/riscv/sail-riscv/issues/1557), I have attempted some modifications, but I am not very familiar with Sail. If there are any issues, I would greatly appreciate your patient assistance. Thank you very much.


bacam:
The change in #1559 should provide a more direct improvement for symbolic execution, because Isla will map it directly to SMT.

PR #1559, opened 2026-02-23
Use Sail standard definitions for shifts by integer amounts

bacam:
Helps with Isla symbolic execution, which doesn't support bitvectors of a symbolic size.  (See #1557.)

### DIFF
--- a/model/prelude/prelude.sail
+++ b/model/prelude/prelude.sail
@@ -181,8 +181,6 @@ val "shift_bits_left"  : forall 'n 'm. (bits('n), bits('m)) -> bits('n)
 overload operator >> = {shift_bits_right, sail_shiftright}
 overload operator << = {shift_bits_left, sail_shiftleft}
 
-// Ideally this would be sail builtin. This implementation is not efficient for large shifts.
-
 val shift_bits_right_arith : forall 'm 'n, 'n >= 1 . (bits('n), bits('m)) -> bits('n)
 function shift_bits_right_arith(value, shift) =
   sail_arith_shiftright(value, unsigned(shift))


model/prelude/prelude.sail
```diff
@@ -178,22 +178,14 @@ infix 7 <<
...
 
-// 'n >= 1 is required due to https://github.com/rems-project/sail/issues/471
-val shift_right_arith : forall 'm 'n, 'n >= 1 & 'm >= 0 . (bits('n), int('m)) -> bits('n)
-function shift_right_arith(value, shift) =
-  sign_extend('n + shift, value)[('n - 1 + shift) .. shift]
-
 val shift_bits_right_arith : forall 'm 'n, 'n >= 1 . (bits('n), bits('m)) -> bits('n)
 function shift_bits_right_arith(value, shift) =
-  shift_right_arith(value, unsigned(shift))
+  sail_arith_shiftright(value, unsigned(shift))
```
Timmmm, line 186: Hmm the definition of this from `vector.sail` is:

```
val sail_arith_shiftright = pure {
  lean: "BitVec.rotateRight",
  _: "arith_shiftr"} : forall 'n ('ord : Order).
    (bitvector('n, 'ord), int) -> bitvector('n, 'ord)
```

Shouldn't `BitVec.rotateRight` be `BitVec.sshiftRight` @Alasdair ?

Otherwise this LGTM.

bacam: I completely failed to notice that when reading it... I'll sort out some testing for it and correct it.

bacam: This is fixed in Sail now.

model/prelude/prelude.sail
```diff
@@ -178,22 +178,14 @@ infix 7 <<
 val "shift_bits_right" : forall 'n 'm. (bits('n), bits('m)) -> bits('n)
 val "shift_bits_left"  : forall 'n 'm. (bits('n), bits('m)) -> bits('n)
 
-val "shiftl" : forall 'm 'n, 'n >= 0. (bits('m), int('n)) -> bits('m)
-val "shiftr" : forall 'm 'n, 'n >= 0. (bits('m), int('n)) -> bits('m)
-
-overload operator >> = {shift_bits_right, shiftr}
-overload operator << = {shift_bits_left, shiftl}
+overload operator >> = {shift_bits_right, sail_shiftright}
+overload operator << = {shift_bits_left, sail_shiftleft}
```
Timmmm, line 182: Btw very weird asymmetry here. Both in the different naming, and in the fact that apparently we only need to declare the type of one of these (and also why do we even need to do that?). Oh well.

Timmmm:
I just removed a comment that was only relevant for the old code.

PR #1561, opened 2026-02-24
Log explicit and implicit updates to `mip`.

### LINKED ISSUE #1560: Missing `mip` CSR callbacks in clint functions
`clint_dispatch` and `clint_store` update `mip` without calling `csr_write_callback`, so the update to `mip` is not captured in the trace file.

pmundkur:
`mip` is updated implicitly in `clint_dispatch()` and explicitly in `clint_store()`. If both updates occur simultaneously, ensure that there is a single `csr_write_callback()`.

Fixes #1560.

### DIFF
--- a/model/sys/platform.sail
+++ b/model/sys/platform.sail
@@ -131,12 +131,11 @@ private function clint_dispatch(mip_was_written : bool) -> unit = {
   if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then {
     mip[STI] = bool_to_bit(stimecmp <=_u mtime);
   };
+  if get_config_print_clint()
+  then print_log("clint mtime " ^ bits_str(mtime));
   if old_mip != mip.bits | mip_was_written then {
     csr_write_callback("mip", mip.bits);
   };
-  if get_config_print_clint()
-  then print_log("clint mtime " ^ bits_str(mtime) ^ " (mip.MTI <- " ^ bits_str(mip[MTI]) ^
-    (if currentlyEnabled(Ext_Sstc) then ", mip.STI <- " ^ bits_str(mip[STI]) else "") ^ ")");
 }
 
 private val clint_store: forall 'n, 'n > 0. (physaddr, int('n), bits(8 * 'n)) -> MemoryOpResult(bool)


model/sys/platform.sail
```diff
@@ -125,11 +125,15 @@ function clint_load(access, Physaddr(addr), width) = {
   }
 }
 
-private function clint_dispatch() -> unit = {
+private function clint_dispatch(mip_was_written : bool) -> unit = {
+  let old_mip = mip.bits;
   mip[MTI] = bool_to_bit(mtimecmp <=_u mtime);
   if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then {
     mip[STI] = bool_to_bit(stimecmp <=_u mtime);
   };
```
jordancarlin, line 133: Unrelated to this PR, but I just noticed that we are handling `stimecmp` from Sstc in `clint_dispatch`. `stimecmp` is not part of the clint (it is a CSR, not MMIO), so it should really be moved elsewhere in the model.

model/sys/platform.sail
```diff
@@ -125,11 +125,15 @@ function clint_load(access, Physaddr(addr), width) = {
...
+  if old_mip != mip.bits | mip_was_written then {
+    csr_write_callback("mip", mip.bits);
+  };
   if get_config_print_clint()
   then print_log("clint mtime " ^ bits_str(mtime) ^ " (mip.MTI <- " ^ bits_str(mip[MTI]) ^
     (if currentlyEnabled(Ext_Sstc) then ", mip.STI <- " ^ bits_str(mip[STI]) else "") ^ ")");
```
jordancarlin: Should we remove this `mip` printing now that it will be captured by the `csr_write_callback`?

model/sys/platform.sail
```diff
@@ -142,43 +146,43 @@ function clint_store(Physaddr(addr), width, data) = {
     if   get_config_print_clint()
     then print_log("clint[" ^ bits_str(addr) ^ "] <- " ^ bits_str(data) ^ " (mip.MSI <- " ^ bits_str(data[0..0]) ^ ")");
```
jordancarlin, line 146: Same as above. Drop the `mip` printing (but retain the `clint` print).

model/sys/platform.sail
```diff
@@ -125,11 +125,15 @@ function clint_load(access, Physaddr(addr), width) = {
...
-private function clint_dispatch() -> unit = {
+private function clint_dispatch(mip_was_written : bool) -> unit = {
+  let old_mip = mip.bits;
   mip[MTI] = bool_to_bit(mtimecmp <=_u mtime);
   if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then {
     mip[STI] = bool_to_bit(stimecmp <=_u mtime);
   };
+  if old_mip != mip.bits | mip_was_written then {
+    csr_write_callback("mip", mip.bits);
+  };
```
jordancarlin, line 138: This should probably go after the print gated on `get_config_print_clint()` so that it is more apparent in the log why `mip.MTI` is changing.

model/sys/platform.sail
pmundkur: An advantage of keeping both here is that it helps differentiate an implicit `mip` update from an explicit write to it.

model/sys/platform.sail
pmundkur: Yes, I think this could go.

model/sys/platform.sail
jordancarlin: The machine-mode bits of `mip` are all read-only (and `mip.STI` is as well when Sstc is enabled), so I'm not sure it actually helps. It should also be clear enough because when it is an implicit update it would always follow a clint message.

model/sys/platform.sail
jordancarlin: Looks like you removed the entire print. I was suggesting removing the `(mip.MTI <- " ^ bits_str(mip[MTI]) ^
    (if currentlyEnabled(Ext_Sstc) then ", mip.STI <- " ^ bits_str(mip[STI]) else "")` portion of it but keeping the `"clint mtime " ^ bits_str(mtime)` portion.

model/sys/platform.sail
pmundkur: I meant an explicit write to the memory-mapped  `mip[MSI]` using the MMIO write to the CLINT.

model/sys/platform.sail
Timmmm: Or we just rename the function...

But actually I need to make some kind of `read_mip()` function for PLIC anyway so maybe it makes sense to just add these on demand anyway rather than pushing them explicitly.

Also `mip` is one of those CSRs where its "value", and implicit writes to it aren't neat and well defined, so whatever prompted this... I wouldn't rely on it always giving the values you expect.

pmundkur: Agreed, I think having a real PLIC would crystallize the mip legalization issues into a more concrete form.

PR #1562, opened 2026-02-24
Add failure reasons for CSR accesses for debugging logs

pmundkur:
A prototype of the debug logging functionality of #1402 but on top of #1556.

It's not as detailed as #1402 since it's just a quick prototype to show the idea.

It avoids the confusing operator override of `&` to operate over a ternary set.

### DIFF
--- a/model/extensions/Zicntr/zicntr_control.sail
+++ b/model/extensions/Zicntr/zicntr_control.sail
@@ -16,40 +16,40 @@ mapping clause csr_name_map = 0xC81  <-> "timeh"
 mapping clause csr_name_map = 0xC82  <-> "instreth"
 
 function clause is_CSR_accessible(0xC00, priv, access_type) = {
-  if not(currentlyEnabled(Ext_Zicntr)) then return Err(CSR_Access_Illegal);
+  if not(currentlyEnabled(Ext_Zicntr)) then return Err(CSR_Access_Illegal, CSR_Access_Disabled_Extension());
   if not(privLevel_is_virtual(priv)) & default_CSR_access_allowed(0xC00, priv, access_type) & counter_enabled(0, priv) then return Ok(());
   // TODO: For VS/VU, check hcounteren and mcounteren.
-  Err(CSR_Access_Illegal)
+  Err(CSR_Access_Illegal, CSR_Access_Failed_Condition())
 }
 function clause is_CSR_accessible(0xC01, priv, access_type) = {
-  if not(currentlyEnabled(Ext_Zicntr)) then return Err(CSR_Access_Illegal);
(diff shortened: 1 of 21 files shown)


model/extensions/Zicntr/zicntr_control.sail
```diff
@@ -36,10 +66,14 @@ mapping clause csr_name_map = 0xB02  <-> "minstret"
...
-function clause is_CSR_accessible(0xB80, _, _) = currentlyEnabled(Ext_Zicntr) & xlen == 32 // mcycleh
-function clause is_CSR_accessible(0xB82, _, _) = currentlyEnabled(Ext_Zicntr) & xlen == 32 // minstreth
+function clause is_CSR_accessible(0xB00, priv, access_type) = // mcycle
+  check_CSR_access(0xB00, priv, access_type, currentlyEnabled(Ext_Zicntr))
+function clause is_CSR_accessible(0xB02, priv, access_type) = // minstret
+  check_CSR_access(0xB02, priv, access_type, currentlyEnabled(Ext_Zicntr))
+function clause is_CSR_accessible(0xB80, priv, access_type) = // mcycleh
+  check_CSR_access(0xB80, priv, access_type, currentlyEnabled(Ext_Zicntr) & xlen == 32)
+function clause is_CSR_accessible(0xB82, priv, access_type) = // minstreth
+  check_CSR_access(0xB82, priv, access_type, currentlyEnabled(Ext_Zicntr) & xlen == 32)
```
arichardson, line 76:
```suggestion
function clause is_CSR_accessible(0xB00, priv, access_type) = // mcycle
  check_CSR_access(0xB00, priv, access_type, Ext_Zicntr)
function clause is_CSR_accessible(0xB02, priv, access_type) = // minstret
  check_CSR_access(0xB02, priv, access_type, Ext_Zicntr)
function clause is_CSR_accessible(0xB80, priv, access_type) = // mcycleh
  check_CSR_access_32(0xB80, priv, access_type, Ext_Zicntr)
function clause is_CSR_accessible(0xB82, priv, access_type) = // minstreth
  check_CSR_access_32(0xB82, priv, access_type, Ext_Zicntr)
```

is adding a _32 version and overloading this with an extension arg too confusing for the spec?

That would allow returning `CSR_Access_Disabled_Extension` if the extension is missing and something like `CSR_Access_XLEN32_Only` for the other cases?

arichardson (commented):
While I like the overloaded operator & for general readability, I do agree with you that it might be confusing for general consumers of the spec. So this approach look sensible to me.

It might be worth adding a special case for the *H registers instead of giving the general "condition failed" error

model/extensions/Zicntr/zicntr_control.sail
pmundkur: That sounds reasonable to me.

pmundkur:
Closing since #1624 has been merged.

PR #1565, opened 2026-02-25
Move interrupt state to its own file

Timmmm:
This reduces the size of the large `sys_regs.sail` file. Also the `mip` logic is about to get super hairy because of SEIP's special snowflake behaviour, so it's nice to isolate things in their own file.


model/riscv.sail_project
```diff
@@ -537,7 +538,7 @@ mops {
 }
 
 postlude {
-  after extensions, mops
+  after extensions, mops, core
```
Timmmm, line 541: I'm not sure if this is meant to be here, or does `requires` imply `after`?

Timmmm:
Note this is giving me weird Sail compiler errors, so it might be hitting a Sail compiler bug. Sometimes I get

    Fatal error: exception Stack overflow

And sometimes I get this (with no source location unfortunately):

```
Error:
Tuple assignment index out of bounds
```

Hopefully it's just my system being weird...

model/core/interrupt_regs.sail
```diff
@@ -0,0 +1,199 @@
...
+  Fetch_Addr_Align  : 0
+}
+
+private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
+  // M-EnvCalls delegation is not supported
+  [Mk_Medeleg(v) with MEnvCall = 0b0]
+}
+
+// Note that the MEI and SEI state from the platform is NOT stored directly
+// in this `mip` register. Instead it is ORed in from `external_interrupts_pending()`
```
pmundkur: Might need some more (temporary) explanation of this `external_interrupts_pending` function until it materializes in tree.

pmundkur (approved):

model/riscv.sail_project
pmundkur: Good catch, it seems better to have `core` in `after` as well.

model/core/interrupt_regs.sail
Timmmm: Oh yeah oops that snuck in from my PLIC branch. I'll remove it.

jordancarlin (approved):

model/core/interrupt_regs.sail
Timmmm: Removed.

Timmmm:
Ok I tried it on another machine and it works, also works in CI, so must just be something weird on my end. :shrug:

nadime15:
@Timmmm I am seeing the exact same issue since yesterday! Were you able to fix it somehow?

I have tried updating the Sail compiler, checking out older commits, switching the build type to debug, and even replacing the linker (to speed up linking), but nothing seems to help. After ~20 minutes I get a Stack overflow.

I am wondering whether this could be related to my GCC version (Ubuntu 15.2.0-4ubuntu4, GCC 15.2.0), even though that doesnt fully make sense to me. As far as I can tell, the only thing I did yesterday was fetch new commits, and I have already tried checking out older ones as well.

nadime15:
Okay, it keeps getting stranger. I hit the stack overflow again, but when I immediately recompiled (without deleting ./build), it worked. This was at commit a33475ae which is before this change got introduced.

Later, I switched to a different branch and hit a stack overflow again. This time, recompiling afterward didn’t work. (This branch is master plus some of my own changes.)

Timmmm:
Hmmm... no I didn't fix it; just tried on a different computer where it worked. To be honest I couldn't even figure out where the stack overflow was coming from. Maybe we can add `--verbose` to the Sail compiler and get more of a clue what's going on?

Alasdair:
You could try running with `OCAMLRUNPARAM=b` to get backtraces on stack overflow

nadime15:
-- Building versions: 2026-02-23-c4d3140-2-g2beb0645-dirty (git), 0.10 (cmake).
-- Found sail: /home/nadime/.opam/4.14.2/bin/sail
-- Sail library directory: /home/nadime/.opam/4.14.2/share/sail
-- jsoncons v1.4.3
-- Sail version: Sail 0.20.1 (sail @ opam-v2.4.1 0.20.1)
-- Configuring done (0.6s)
-- Generating done (0.3s)
-- Build files have been written to: /home/nadime/work/sail/sail-riscv/build
[  0%] Built target default_config
[  1%] Building C++ code from Sail model
[  3%] Built target sail_runtime
[ 94%] Built target softfloat
Fatal error: exception Stack overflow
Raised by primitive operation at Stdlib__Set.Make.singleton in file "set.ml", line 143, characters 26-54
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
...
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__List.map in file "list.ml", line 92, characters 32-39
gmake[2]: *** [model/CMakeFiles/generated_sail_riscv_model.dir/build.make:229: sail_riscv_model.cpp] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:554: model/CMakeFiles/generated_sail_riscv_model.dir/all] Error 2
gmake: *** [Makefile:166: all] Error 2

Alasdair:
Ah that is a super unhelpful backtrace.

However, List.map is tail recursive since OCaml 5.1, so maybe building Sail with a newer OCaml compiler might fix it? (Also gives me a clue to try reproducing it by downgrading).

Alasdair:
Hmm, I can't reproduce even on an older OCaml version. Do you know exactly which version you used? Or if it's the pre-built binary?

You can control the stack size by using the `l` flag in OCAMLRUNPARAM, like `OCAMLRUNPARAM="b,l=1M` will give you a one megabyte stack and backtraces (https://ocaml.org/manual/5.4/runtime.html).

The default is 256k afaik. I tried a few numbers and wasn't able to induce a failure until 50k, so something odd is going on. Does it go away if you do a fully clean build?

nadime15:
Ok yeah that helped, I switched to OCaml 5.2.0 and works, thanks!

This is unrelated to the issue above, but it seems to me that compilation takes noticeably longer than before. I havent measured it, just an impression, but maybe its just in my head.

Timmmm:
Ah interesting. Did you ever see the `Error: Tuple assignment index out of bounds` error @nadime15 ?

We should really get some compile time benchmark going somewhere.

nadime15:
@Alasdair
I was using OCaml version 4.14.2. Before this commit, I tried a fully clean build, which first failed, but then it worked when I ran another build without fully cleaning. After that I tested again with HEAD at the latest commit which did not work at all. I am not sure if this actually answers your question.

@Timmmm
Nope, I did not see the same error message

Alasdair:
If that error occurs again, if should (hopefully) have some additional context provide you have thi: https://github.com/rems-project/sail/pull/1633

Alasdair:
It is an internal compiler error though, so it should not be happening for any reason

nadime15:
Ok yeah I see, thanks!

PR #1571, opened 2026-02-28
Fail CI if one of the jobs fail

jordancarlin:
Another attempt at preventing PRs that fail CI from merging. Apparently GitHub considers skipped jobs as passing, and the `ci_pass` job was getting skipped when one of the jobs it depended on failed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,8 @@ jobs:
   # This is here so that the "Status checks that are required" Github
   # option doesn't need to list all of the matrix jobs (it doesn't seem
   # to work if you just specify the job ID of the matrix itself).
+  # A skipped job is considered passing by GitHub, so this job needs to
+  # always run and then checks the results of all of the jobs it depends on.
   ci_pass:
     runs-on: ubuntu-latest
     needs: [build, linux]
@@ -152,5 +154,4 @@ jobs:
         if: "${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}"
         run: exit 1
       - name: exit success
-        if: "${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}"
         run: exit 0


.github/workflows/ci.yml
```diff
@@ -146,6 +146,11 @@ jobs:
...
+    if: "${{ always() }}"
     steps:
-      - name: Dummy
-        run: echo 'Jobs need at least one step.'
+      - name: exit failure
+        if: "${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}"
+        run: exit 1
+      - name: exit success
+        if: "${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}"
+        run: exit 0
```
Timmmm, line 157: I guess you could simplify this slightly like

```
      - name: exit success
        if: "${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}"
         run: exit 0
      - name: exit failure
         run: exit 1
```

Seems super dumb that the normal thing that surely everyone wants is so hard to set up? (On the other hand people do complain about Github CI a lot so maybe that's just the way it is!)

.github/workflows/ci.yml
```diff
@@ -146,6 +146,11 @@ jobs:
   ci_pass:
     runs-on: ubuntu-latest
     needs: [build, linux]
+    if: "${{ always() }}"
```
Timmmm, line 151: Oh also we should definitely have comments explaining why it's like this.

.github/workflows/ci.yml
jordancarlin: It would always run the second step and fail then.

Timmmm: Oh yeah sorry, the other way around...

pmundkur: Is the `exit success` step really needed? What happens if it is omitted?

jordancarlin: Otherwise the job is considered skipped because none of the steps run. Skipped and passing CI jobs show up differently in the GitHub interface, and I think we want it to show that it is passing.

PR #1573, opened 2026-02-28
add configuration option for reserved behavior vsew >= 100.

challenger1024:


model/extensions/V/vext_regs.sail
```diff
@@ -247,7 +247,12 @@ type SEW_pow = range(3, 6)
 type LMUL_pow = range(-3, 3)
 
 function is_invalid_sew_pow(v : bits(3)) -> bool =
-  v >_u 0b011
+  if v >_u 0b011
+  then match v_sew_reserved_behavior {
+    Vsew_Fatal => reserved_behavior("vsew contains reserved value: " ^ bits_str(v) ^ "."),
+    // The caller will detect the return value and invoke handle_illegal_vtype().
+    Vsew_Illegal => true,
+  } else false
```
Timmmm, line 255: I'm not sure about this, it seems weird that this function would have side effects.

Can we do the check when you actually *write* vsew instead?

challenger1024: In model/extensions/V/vext_vset_insts.sail, the execute_vsetvl_type function calls is_invalid_sew_pow, and the subsequent write only occurs when the check passes. Given this control flow, I was wondering whether my modification could still introduce any unintended side effects.

Timmmm: Hmm, I think I would just have one config option `base.reserved_behavior.vtype`, and put the code in `handle_illegal_vtype`. Although this technically isn't "reserved behaviour" in the sense that implementations are allowed to do anything, so I'm starting to thing we just shouldn't do this at all.

See https://riscv-isa-manual.timhutt.co.uk/riscv-unprivileged.html#_unsupported_vtype_values

challenger1024: So should I close this PR?

Timmmm: Yeah I think so. Sorry about that!

PR #1575, opened 2026-03-01
Enable release builds for Mac

Timmmm:
Build Mac binary releases in addition to Linux. Since the Sail compiler Mac binary release isn't quite ready this uses Sail from OPAM.

This required one fix: don't use `-static-libstdc++` or `-static-libgcc` for non-GCC compilers. Mac uses AppleClang which doesn't support these flags. The resulting binary depends on these libraries, which I think is kind of unavoidable on Mac:

    /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1900.180.0)
    /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)

I also increased the parallelism of ctest since by default it is 1, which takes a long time. I think the default Github runners are 2-core.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -43,10 +43,9 @@ jobs:
             os: ubuntu-24.04-arm
             container: rockylinux:8.9.20231119
 
-          # TODO: Mac. We need a Mac binary release of the Sail compiler first.
-          # - name: mac-arm64
-          #   os: macos-latest
-          #   container: ""
+          - name: mac-arm64
+            os: macos-latest
+            container: ""
 
     runs-on: ${{ matrix.os }}
     container: ${{ matrix.container }}
@@ -66,12 +65,6 @@ jobs:
         run: |
           dnf install -y git
 
-      # TODO: Mac
-      # - name: Install dependencies (Mac)
-      #   if: startsWith(matrix.os, 'macos')
-      #   run: |
-      #     brew install --force --overwrite ...
-
       - name: Check out repository code
         uses: actions/checkout@v6
 
@@ -86,7 +79,7 @@ jobs:
           cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=Release -DSTATIC=TRUE -DENABLE_RISCV_TESTS=TRUE -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE
           cd build
           ninja all
-          ctest
+          ctest -j2
           cpack
 
       - name: Build JSON and HTML
@@ -169,6 +162,7 @@ jobs:
           files: |
             release-Linux-x86_64/sail-riscv-Linux-x86_64.tar.gz
             release-Linux-aarch64/sail-riscv-Linux-aarch64.tar.gz
+            release-mac-arm64/sail-riscv-mac-arm64.tar.gz
             release-Linux-x86_64/model/sail_riscv_model.json
             release-Linux-x86_64/model/sail_riscv_html.tgz
           fail_on_unmatched_files: true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,9 @@ if(POLICY CMP0135)
     cmake_policy(SET CMP0135 OLD)
 endif()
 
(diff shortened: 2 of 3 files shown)


Timmmm:
I did [a test run](https://github.com/Timmmm/sail-riscv/actions/runs/22540916711/job/65295917389) of this and it seems to work. You can download the binary from the Upload Artifacts step there.

CMakeLists.txt
```diff
@@ -80,16 +80,18 @@ if (STATIC)
         list(INSERT CMAKE_FIND_LIBRARY_SUFFIXES 0 .lib .a)
     else()
         set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
-        check_linker_flag(CXX "-static-libstdc++" HAVE_STATIC_LIBSTDCXX)
-        if (NOT HAVE_STATIC_LIBSTDCXX)
-            message(FATAL_ERROR "Building with -DSTATIC=TRUE requires static C++ std library installation")
+        if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
jordancarlin: What about a static build using clang on Linux? Seems like that should be possible but is no longer possible with this approach.

CMakeLists.txt
```diff
@@ -80,16 +80,18 @@ if (STATIC)
...
+            if (NOT HAVE_STATIC_LIBSTDCXX)
+                message(FATAL_ERROR "Building with -DSTATIC=TRUE requires static C++ std library installation")
+            endif()
+            add_link_options(-static-libstdc++)
+            check_linker_flag(CXX "-static-libgcc" HAVE_STATIC_LIBGCC)
+            if (NOT HAVE_STATIC_LIBGCC)
+                message(FATAL_ERROR "Building with -DSTATIC=TRUE requires static libgcc installation")
+            endif()
+            add_link_options(-static-libgcc)
         endif()
```
jordancarlin, line 95: This now silently ignores the `STATIC` option if the compiler is not GCC. Seems like it should at least print a warning and probably even an error.

CMakeLists.txt
Timmmm: Hmm that's a good point. Maybe we should just not print an error instead? So it will add those flags for all compilers that support them? (Clang and GCC but not AppleClang)

CMakeLists.txt
Timmmm: It still does `set(CMAKE_FIND_LIBRARY_SUFFIXES .a)` which links GMP statically on Linux and Mac.

jordancarlin: Oops. Missed that. Sounds good.

CMakeLists.txt
jordancarlin: Seems reasonable to me. Or we could condition it on platform instead of compiler ID and explicitly skip it only for macOS.

Timmmm: Hmm yeah maybe that is the best way. I can't see how you'd ever use those flags on Mac.

.github/workflows/release.yml
Timmmm, line 167: Also I forgot to add Mac here.

Timmmm: Added.

CMakeLists.txt
Timmmm: I went for adding the flags if the compiler supports them. Hopefully that will be fine.

jordancarlin: Seems like a good solution.

pmundkur:
What is the backwards compatibility situation on MacOS?  Is building on `macos-latest` an impediment here, and should we be (testing and) releasing a build on `macos-14`?

jrtc27:
Use -mmacosx-version-min= if you want to support older versions than whatever GitHub's runner is

Timmmm:
> Use -mmacosx-version-min= if you want to support older versions than whatever GitHub's runner is

I added this although it isn't strictly necessary at the moment since it seems the default is to support MacOS 11.0+ which is the first version that came with ARM Macs. I guess it can't hurt to be specific though, and it will make a difference if we ever support x86 Macs.

PR #1576, opened 2026-03-01
Fix simulator output when run without arguments

### LINKED ISSUE #1572: Emulator output with no arguments is a bit weird
```
% ./sail_riscv_sim   
Sail RISC-V Model 


Usage: [OPTIONS]


OPTIONS:
  -h,     --help              Print this help message and exit 

```

Kind of implies the *only* option is `-h`. Might have been me that broke this 😬 . It's not too bad though - if you run it with `-h` it does print all of the options properly.

KotorinMinami:
The simulator output when you run it without any arguments was missing all the options except `-h`.

Fixes #1572


Timmmm (approved):
Oops! Thanks for fixing it.

PR #1583, opened 2026-03-02
Allow running RVFI-DII without any ELFs

Timmmm:
If you use RVFI you shouldn't need to provide any ELFs. This code is a little janky but I'm working on refactoring this whole file separately so I will fix it later.

This makes TestRig work again (with some patches to how it thinks you run the emulator).

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -695,8 +695,8 @@ int inner_main(int argc, char **argv) {
     return EXIT_SUCCESS;
   }
 
-  // If we get here, we need to have ELF files to run.
-  if (opts.elfs.empty()) {
+  // If we get here, we need to have ELF files to run (except in RVFI mode).
+  if (opts.elfs.empty() && !rvfi.has_value()) {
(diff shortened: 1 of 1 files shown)


c_emulator/riscv_sim.cpp
```diff
@@ -713,7 +713,7 @@ int inner_main(int argc, char **argv) {
 
   init_start = steady_clock::now();
 
-  if (rvfi) {
+  if (rvfi.has_value()) {
```
arichardson, line 716: The existing code was fine, but if you want to be explicit about this being a std::optional, that also sounds good to me.

jordancarlin:
> I wonder if it makes sense to have a CI job that does something like a "one-test run of testrig" of sail against itself to ensure the infrastructure still works?

See #933 and #1584.

c_emulator/riscv_sim.cpp
Timmmm: Yeah I'm not a fan of implicit bool conversion in general. Makes it harder to follow the code and increases the chance of missing type errors.

PR #1587, opened 2026-03-03
Improve handling of `ebreak`.

pmundkur:
`ebreak` was handled as a `MemoryException` when it really should be a `Trap`.

While here, remove a type annotation that is no longer needed.

### DIFF
--- a/model/extensions/I/base_insts.sail
+++ b/model/extensions/I/base_insts.sail
@@ -604,9 +604,9 @@ mapping clause encdec = EBREAK()
   <-> 0b000000000001 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
 
 function clause execute EBREAK() = {
-  let e = E_Breakpoint(Brk_Software);
-  let t : sync_exception = struct { trap    = e,
-                                    excinfo = xtval_exception_value(e, PC),
+  let trap = E_Breakpoint(Brk_Software);
+  let t : sync_exception = struct { trap    = trap,
+                                    excinfo = xtval_exception_value(trap, PC),
                                     ext     = None() };
   Trap(cur_privilege, CTL_TRAP(t), PC)
 }


model/extensions/I/base_insts.sail
```diff
@@ -603,8 +603,13 @@ union clause instruction = EBREAK : unit
 mapping clause encdec = EBREAK()
   <-> 0b000000000001 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
 
-function clause execute EBREAK() =
-  Memory_Exception(Virtaddr(PC), E_Breakpoint(Brk_Software))
+function clause execute EBREAK() = {
+  let e = E_Breakpoint(Brk_Software);
```
jordancarlin: ecall calls this variable `trap` instead of `e`. We should probably be consistent.

model/extensions/I/base_insts.sail
```diff
@@ -603,8 +603,13 @@ union clause instruction = EBREAK : unit
 mapping clause encdec = EBREAK()
   <-> 0b000000000001 @ 0b00000 @ 0b000 @ 0b00000 @ 0b1110011
 
-function clause execute EBREAK() =
-  Memory_Exception(Virtaddr(PC), E_Breakpoint(Brk_Software))
+function clause execute EBREAK() = {
+  let trap = E_Breakpoint(Brk_Software);
+  let t : sync_exception = struct { trap    = trap,
```
Timmmm, line 608: Very minor but you could probably inline this.

```suggestion
  let t : sync_exception = struct { trap    = E_Breakpoint(Brk_Software),
```

pmundkur: Inlining `trap` would need `E_breakpoint(Brk_Software)` to repeated in the `excinfo` line below that calls `xtval_exception_value(trap, PC)`.  

But agreed that the exception and trap handling flow is weird and could be improved; it is pretty old code that hasn't aged well.

Timmmm: Oh yeah I missed that.

PR #1588, opened 2026-03-03
Add a partial implementation of Sdtrig.

pmundkur:
This adds the icount, itrigger and etrigger triggers, leaving mcontrol6 for a later PR due to its complexity.  Tracing and external triggers need additional infrastructure and are also left for later.  Also left for a later PR are additional matching functionality using the `textra{32,64}` and `{s,m,h}context` CSRs.

A `--trace-trigger` command-line option has been added, and callbacks have been added to track trigger matching and firing.


pmundkur:
Making this a draft while I test it.

model/extensions/Sdtrig/sdtrig_control.sail
```diff
@@ -0,0 +1,197 @@
...
+  let (t, action) : (Trigger, Trig_Action) = match triggers[idx] {
+    T_icount(t) => {
+      let t = if icount_count_is_hardwired_one
+               then {t with m = false, s = false, u = false, vs = false, vu = false}
+               else t;
+      (T_icount({t with pending = false}), t.action)
+    },
+    T_itrigger(t) => {
+      (T_itrigger({t with match_val = None()}), t.action)
+    },
```
nadime15, line 186: I think, right now `icount` can never fire. In `find_highest_priority_pending_trigger`, the condition is:

```
t.pending & t.hit & priv_match_icount(...)
```

but nothing ever sets `hit` for `icount`. `match_icount_triggers()` only sets `count` and `pending`, and `fire_trigger()` doesnt set it either. Thats why `hit` remains 0 unless software explicitly writes it, so the trigger just arms and sits there.

In the spec, `hit` is not a condition for firing, it's only a status bit. For `icount`, it says that 

> hardware sets it when the trigger fires

and the firing rule is:

> When `pending` is set, the trigger fires just before any further instructions are executed in a mode where the trigger is enabled.

It also says that if the `hit` bit is not implemented, it always reads as 0. That means it cannot be a condition for firing, otherwise implementations that dont support `hit` could never fire an `icount` trigger.

It seems to be working already for `itrigger` and `etrigger`.

The following should fix this (I modified line 164 and 182):
```suggestion
  foreach (i from 0 to (sizeof(trigger_vector_length) - 1)) {
    match triggers[i] {
      T_disabled(_) => (),
      T_icount(t) =>
        // icount triggers fire in a mode where the trigger is enabled.
        if t.pending & priv_match_icount(t, priv, t.action) then return Some(t.action, i, zeros()),
      T_itrigger(t) =>
        if t.hit & is_some(t.match_val) & get_reentrancy_control(t.action, priv) then return Some(t.action, i, zeros()),
      T_etrigger(t) =>
        if t.hit & is_some(t.match_val) & get_reentrancy_control(t.action, priv) then return Some(t.action, i, zeros()),
    }
  };
  None()
}

// firing

function fire_trigger(idx : trigger_index) -> unit = {
  let (t, action) : (Trigger, Trig_Action) = match triggers[idx] {
    T_icount(t) => {
      let t = if icount_count_is_hardwired_one
               then {t with m = false, s = false, u = false, vs = false, vu = false}
               else t;
      (T_icount({t with pending = false, hit = true}), t.action)
    },
    T_itrigger(t) => {
      (T_itrigger({t with match_val = None()}), t.action)
    },
```

nadime15 (commented):
While the Hypervisor PR is being reviewed, I started cross checking this PR in the meantime. I just started, so I may find more but here is one thing I noticed:

model/extensions/Sdtrig/sdtrig_types.sail
```diff
@@ -0,0 +1,589 @@
...
+
+// TODO: support enabling tdata1.dmode when Debug Mode is supported with Sdext.
+function make_trigger(tdata1 : TData1, tdata2 : xlenbits, tdata3 : xlenbits) -> Trigger = {
+  let tt = tdata1[trig_type];
+  if   not(is_supported_trigger_type(tt))
+  then return make_disabled_trigger(tdata1, tdata2, tdata3);
+  match trig_type(tt) {
+    TT_icount => {
+      let v = Mk_ICount_bits(tdata1.bits);
+      let act = v[action][3..0];
```
nadime15:
```suggestion
      let act = v[action];
```

model/extensions/Sdtrig/sdtrig_types.sail
```diff
@@ -0,0 +1,589 @@
...
+          dmode = false,
+          vs = if currentlyEnabled(Ext_H) then bool_bit(v[vs]) else false,
+          vu = if currentlyEnabled(Ext_H) then bool_bit(v[vu]) else false,
+          hit = bool_bit(v[hit]),
+          count = if icount_count_is_hardwired_one then zero_extend(0x1) else v[count],
+          m = bool_bit(v[m]),
+          pending = bool_bit(v[pending]),
+          s = if currentlyEnabled(Ext_S) then bool_bit(v[s]) else false,
+          u = if currentlyEnabled(Ext_U) then bool_bit(v[u]) else false,
+          action = trig_action(act),
```
nadime15, line 506:
```suggestion
          action = trig_action(act[3 .. 0]),
```

model/extensions/Sdtrig/sdtrig_types.sail
```diff
@@ -0,0 +1,589 @@
...
+        T_itrigger(struct {
+          dmode = false,
+          hit = bool_bit(v[hit]),
+          vs = if currentlyEnabled(Ext_H) then bool_bit(v[vs]) else false,
+          vu = if currentlyEnabled(Ext_H) then bool_bit(v[vu]) else false,
+          nmi = bool_bit(v[nmi]),
+          m = bool_bit(v[m]),
+          s = if currentlyEnabled(Ext_S) then bool_bit(v[s]) else false,
+          u = if currentlyEnabled(Ext_U) then bool_bit(v[u]) else false,
+          action = trig_action(act),
```
nadime15, line 527:
```suggestion
          action = trig_action(act[3 .. 0]),
```

model/extensions/Sdtrig/sdtrig_types.sail
```diff
@@ -0,0 +1,589 @@
...
+      if is_supported_action(act) then {
+        T_etrigger(struct {
+          dmode = false,
+          hit = bool_bit(v[hit]),
+          vs = if currentlyEnabled(Ext_H) then bool_bit(v[vs]) else false,
+          vu = if currentlyEnabled(Ext_H) then bool_bit(v[vu]) else false,
+          m = bool_bit(v[m]),
+          s = if currentlyEnabled(Ext_S) then bool_bit(v[s]) else false,
+          u = if currentlyEnabled(Ext_U) then bool_bit(v[u]) else false,
+          action = trig_action(act),
```
nadime15, line 548:
```suggestion
          action = trig_action(act[3 .. 0]),
```

model/extensions/Sdtrig/sdtrig_types.sail
```diff
@@ -0,0 +1,589 @@
...
+          tdata2 = tdata2,
+          tdata3 = tdata3,
+        })
+      } else {
+        make_disabled_trigger(tdata1, tdata2, tdata3)
+      }
+    },
+    TT_itrigger => {
+      let v = Mk_IETrigger_bits(tdata1.bits);
+      let act = v[action][3..0];
```
nadime15:
```suggestion
      let act = v[action];
```

model/extensions/Sdtrig/sdtrig_types.sail
```diff
@@ -0,0 +1,589 @@
...
+          tdata3 = tdata3,
+          match_val = None(),
+        })
+      } else {
+        make_disabled_trigger(tdata1, tdata2, tdata3)
+      }
+    },
+    TT_etrigger => {
+      let v = Mk_IETrigger_bits(tdata1.bits);
+      let act = v[action][3..0];
```
nadime15:
```suggestion
      let act = v[action];
```

model/extensions/Sdtrig/sdtrig_types.sail
```diff
@@ -0,0 +1,589 @@
...
+  TA_Trace_Off     <-> "Trace Off",
+  TA_Trace_Notify  <-> "Trace Notify",
+  TA_Ext_Trigger_0 <-> "External Trigger Output 0",
+  TA_Ext_Trigger_1 <-> "External Trigger Output 1",
+}
+
+function is_supported_action(a : bits(4)) -> bool = {
+  let act = trig_action(a) in
+  act == TA_Breakpoint // | (currentlyEnabled(Ext_Sdext) & (act == TA_Enter_Debug))
+}
```
nadime15: Safer to check first, otherwise we call trig_action with a value that has no mapping and the model crashes. Software can write any value to the action field, so this is reachable. I also check bits 5:4 because the field is 6 bits but the mapping only takes 4, so without it action=16 gets read as breakpoint (because the code before truncated the upper 2 bits, so 0b10000 became 0b0000 which is a breakpoint).


```suggestion
function is_supported_action(a : bits(6)) -> bool = {
  if a[5 .. 4] != zeros() and not(trig_action_forwards_matches(a[3 .. 0])) then return false;
  let act = trig_action(a[3 .. 0]) in
    act == TA_Breakpoint // | (currentlyEnabled(Ext_Sdext) & (act == TA_Enter_Debug))
   }
}
```

pmundkur:
Thanks for reviewing this!  I'll pick this up again when I get back next week.

model/extensions/Sdtrig/sdtrig_control.sail
pmundkur: Good catch; fixed.

model/extensions/Sdtrig/sdtrig_types.sail
pmundkur: I changed `trig_action` to be `bits(6) <-> Trig_Action` and added the `forwards_match` check.

model/extensions/Sdtrig/sdtrig_types.sail
pmundkur: Not needed now, see above.

model/extensions/I/base_insts.sail
```diff
@@ -612,6 +617,8 @@ function clause execute SRET() = {
   else if not(ext_check_xret_priv (Supervisor))
   then Ext_XRET_Priv_Failure()
   else {
+    match_icount_triggers(cur_privilege);
+
     let prev_priv = cur_privilege;
```
nadime15, line 622: > If more than one of the above events occur during a single instruction execution, the trigger still only
> matches once for that instruction.

You call match_icount_triggers in MRET/SRET but then again in Step_Execute(Retire_Success(), _) after returning successfully. 

I was thinking we might be able to delete these two calls, but that alone wouldn’t fix the issue, since xret also changes mstatus.mie and tcontrol.mte, which go into get_reentrancy_control. Does it make sense to snapshot that state before execution in try_step, next to the existing minstret_increment line, and have match_icount_triggers use the snapshot?

model/extensions/Sdtrig/sdtrig_control.sail
```diff
@@ -0,0 +1,197 @@
...
+  foreach (i from 0 to (sizeof(trigger_vector_length) - 1)) {
+    match triggers[i] {
+      T_disabled(_) => (),
+      T_icount(t) =>
+        // icount triggers fire in a mode where the trigger is enabled.
+        if t.pending & priv_match_icount(t, priv, t.action) then return Some(t.action, i, zeros()),
+      T_itrigger(t) =>
+        if t.hit & is_some(t.match_val) & get_reentrancy_control(t.action, priv) then return Some(t.action, i, zeros()),
+      T_etrigger(t) =>
+        if t.hit & is_some(t.match_val) & get_reentrancy_control(t.action, priv) then return Some(t.action, i, zeros()),
```
nadime15, line 168: Reentrancy gets evaluated twice once at match time in trap_handler (sys_control.sail, still pre trap MIE and source privilege) and again here at fire time. On a trap into M-mode mie is cleared at so this check fails and the trigger does not fire, but match_val stays Some. It then fires at a random step later once mie is set to, for example after the handler's mret, in U-mode with no connection to the original trap.

See:

> Both options prevent etrigger and itrigger from having any effect on exceptions and interrupts that are handled in M-mode.


Could the pending match be dropped when reentrancy denies the fire, rather than carried forward? Same applies to the itrigger arm at 166.

model/extensions/Sdtrig/sdtrig_control.sail
```diff
@@ -0,0 +1,197 @@
...
+
+function find_highest_priority_pending_trigger(priv : Privilege) -> option((Trig_Action, trigger_index, xlenbits)) = {
+  if not(currentlyEnabled(Ext_Sdtrig)) then return None();
+
+  foreach (i from 0 to (sizeof(trigger_vector_length) - 1)) {
+    match triggers[i] {
+      T_disabled(_) => (),
+      T_icount(t) =>
+        // icount triggers fire in a mode where the trigger is enabled.
+        if t.pending & priv_match_icount(t, priv, t.action) then return Some(t.action, i, zeros()),
```
nadime15: The return exits on the first pending trigger, so only one fires per step and the rest fire on later steps. For example: two icount triggers, both count=1, u=1, m=0. One retire in U-mode sets pending on
  both. Trigger 0 fires and traps to M-mode, trigger 1 does not match there, so it waits for the mret and then raises a second breakpoint at an unrelated instruction.

  The spec treats this as one event:
 
  

> When multiple triggers in the same priority fire at once, hit (if implemented) is set for all of them. If more than one of these triggers has action=0 then tval is updated in accordance with one of them, but which one is UNSPECIFIED.

  
  Picking the first one for the action is fine, but the others need their pending/match_val cleared and hit set at the same time. For etrigger and itrigger hit is already set at match time in line 133, only icount misses it

model/extensions/Sdtrig/sdtrig_types.sail
```diff
@@ -0,0 +1,637 @@
...
+  let typ = trig_type_bits(tt);
+  (typ == TT_none) | (typ == TT_disabled) |
+  (typ == TT_icount & sdtrig_has_icount) |
+  (typ == TT_etrigger & sdtrig_has_etrigger) |
+  (typ == TT_itrigger & sdtrig_has_itrigger)
+}
+
+function supported_triggers_info() -> bits(16) = {
+  var info : bits(16) = zeros();
+  let flag = zero_extend(16, 0b1);
```
pmundkur, line 76: I'm not sure why I did this.  It's saying "no trigger" is supported, which is an odd thing to say..

model/extensions/I/base_insts.sail
pmundkur: Yes, I think this makes sense.

model/extensions/Sdtrig/sdtrig_control.sail
pmundkur: Right, this is an issue.

> Could the pending match be dropped when reentrancy denies the fire, rather than carried forward? Same applies to the itrigger arm at 166.

I didn't understand this.  I was wondering if we could just drop the `get_reentrancy_control` check for `itrigger` and `etrigger` here, since they were checked when the triggers matched.  (Perhaps that's what you meant?)

model/extensions/Sdtrig/sdtrig_regs.sail
```diff
@@ -0,0 +1,152 @@
...
+  let t = trigger_write_tdata3(idx, triggers[idx], value);
+  triggers[idx] = t;
+  Ok(tdata3_of_trigger(t))
+}
+
+mapping clause csr_name_map = 0x7a4  <-> "tinfo"
+function clause is_CSR_accessible(0x7a4, _, _) = currentlyEnabled(Ext_Sdtrig)
+function clause read_CSR(0x7a4) = {
+  // indicate that the trigger does not exist.
+  if sizeof(num_triggers) == 0 then return zero_extend(0x1);
```
nadime15, line 91: All the `sizeof(num_triggers) == 0` guards look unreachable to me. validate_config.sail rejects num_triggers == 0 when Sdtrig is supported, and when it is not supported `is_CSR_accessible` on the line above already makes the CSR illegal. So no CSR access sequence can reach these branches.

The tinfo write clause just below is also missing the same guard, so it would hit the assert where the read returns early. I think we can either drop the guards everywhere or adding the missing one, so read and write have the same behavior.

pmundkur: Good point; perhaps they can be replaced by '!= 0' asserts.

model/extensions/Sdtrig/sdtrig_control.sail
nadime15: I think dropping would not work. Without the check the etrigger fires on the next step, which is the first instruction of the M-mode handler with mie already 0. That is the case option 1 (The hardware prevents triggers...) tries to prevent:

> The hardware prevents triggers with action=0 from matching or firing while in M-mode and while MIE in mstatus is 0.

What I meant is to keep the check but clear `match_val` when it denies, so the trigger is dropped instead of staying pending for a subsequent step.

I think we need both checks. The m/s/u bits belong to the mode the trap was taken from, so they have to be evaluated at match time. Reentrancy belongs to the mode where the fire would happen, so it can only be evaluated after the trap state is set.

pmundkur:
I just pushed a rebase, but haven't addressed the feedback...

davidharrishmc:
Comments from Claude review:
* tcontrol doesn't update on trap entry
* writes to tdata1 of a configured trigger only update hit.  Spec says they should also write count.
* write to icount should not match against the new trigger.  Spec says count must hold the written value afterward
* etrigger and itrigger matches should be dropped, not deferred during reentrancy.  Spec says both reentrancy options "prevent etrigger and itrigger from having any effect on exceptions and interrupts that are handled in M-mode."   
* Simultaneous triggers should all fire, not just one per step.
* VSEI delegated to VSmode should be seen as cause 9, but itrigger checks bit 10
* tdata1 write that sets type masks tdata2

model/postlude/step.sail
```diff
@@ -95,26 +97,47 @@ private function run_hart_waiting(_step_no : nat, wr: WaitReason, instbits : ins
...
+  };
+
   match dispatchInterrupt(cur_privilege) {
     Some(intr, priv) => return Step_Pending_Interrupt(intr, priv),
     None() => (),
   };
+
+  // `icount` triggers match based on the privilege used for fetching,
+  // but the match is performed after the instruction retires, by which time
+  // the privilege might have changed (e.g. the RET instructions).
```
nadime15:
```suggestion
  // the privilege might have changed (e.g. the xRET instructions).
```

model/extensions/Sdtrig/sdtrig_type_utils.sail
```diff
@@ -0,0 +1,56 @@
...
+  supported = supported | make_etrigger_flag(E_Load_Page_Fault());
+  supported = supported | make_etrigger_flag(E_SAMO_Page_Fault());
+  supported = supported | make_etrigger_flag(E_Software_Check());
+  supported = supported | make_etrigger_flag(E_Fetch_GPage_Fault());
+  supported = supported | make_etrigger_flag(E_Load_GPage_Fault());
+  supported = supported | make_etrigger_flag(E_Virtual_Instr());
+  supported = supported | make_etrigger_flag(E_SAMO_GPage_Fault());
+  supported = supported | make_etrigger_flag(E_Breakpoint(Brk_Software));
+  supported
+}
```
nadime15: This might be a bit hard to parse through because its not obvious which exceptions are not valid triggers, what about: 

```
// Support all codes from 0 to 23 except reserved and hardware error codes
let supported_etrigger_mask : xlenbits =
  zero_extend(ones() : bits(24))
  & ~(  make_etrigger_flag(E_Reserved_14())
      | make_etrigger_flag(E_Reserved_16())
      | make_etrigger_flag(E_Reserved_17())
      | make_etrigger_flag(E_Hardware_Error()))
```

pmundkur: I think that makes sense, but the reason it was done this way was to support configuration for individual interrupts:

> Hardware may support only a subset of exceptions. A debugger must read back tdata2 after writing it
to confirm the requested functionality is actually supported.

I don't know whether implementations commonly support only subsets.   If not, we could go with the more succinct version.

nadime15: Ahh ok I see, makes sense. 

I believe we could extend this to:

```
let supported_etrigger_mask : xlenbits =
  zero_extend(ones() : bits(24))
  & ~(  make_etrigger_flag(E_Reserved_14())
      | make_etrigger_flag(E_Reserved_16())
      | make_etrigger_flag(E_Reserved_17())
      | make_etrigger_flag(E_Hardware_Error()))
  & sdtrig_etrigger_supported_bits 
```

and by default all are supported, similar to `delegatable_bits`. So a user could, if desired, support fewer exceptions. Btw. I think the same applies to itrigger.

pmundkur: That's true, except we'll need some additional validation logic similar to that for `delegatable_bits`.    I'm okay either way, but perhaps the mask approach is more consistent in this case.

pmundkur: > This might be a bit hard to parse through because its not obvious which exceptions are not valid triggers, what about:
> 
> ```
> // Support all codes from 0 to 23 except reserved and hardware error codes
> let supported_etrigger_mask : xlenbits =
>   zero_extend(ones() : bits(24))
>   & ~(  make_etrigger_flag(E_Reserved_14())
>       | make_etrigger_flag(E_Reserved_16())
>       | make_etrigger_flag(E_Reserved_17())
>       | make_etrigger_flag(E_Hardware_Error()))
> ```

I'm in the middle of adding this and the corresponding mask. What was the reasoning behind excluding `Hardware_Error`?   It doesn't occur in the model, so the trigger will never match it, so perhaps there is no downside in saying the trigger for it is supported?

nadime15: No idea to be honest :D I just noticed E_Hardware_Error was not added to the original version of `supported_etrigger_mask` thats why I excluded it. I would need to double check the spec and look if Hardware Errors are mentioned somewhere.

nadime15: Hmm I could not find anything, I think it's fine to support `Hardware_Error`, since triggers may fire on up to XLEN of the Exception Codes defined in mcause:

> This trigger may fire on up to XLEN of the Exception Codes defined in mcause (described in the Privileged Spec, with Interrupt=0). Those causes are configured by writing the corresponding bit in tdata2. (E.g. to trap on an illegal instruction, the debugger sets bit 2 in tdata2.)

model/core/interrupt_regs.sail
```diff
@@ -184,7 +183,9 @@ private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
     VSEnvCall         = if currentlyEnabled(Ext_H) then v[VSEnvCall]         else 0b0,
     // M-EnvCalls and DoubleTrap cannot be delegated.
     Double_Trap = 0b0,
-    MEnvCall    = 0b0
+    MEnvCall    = 0b0,
+    // medeleg[3] is hardwired to 0 when using `tcontrol`-based reentrancy handling for breakpoints.
+    Breakpoint         = if sdtrig_breakpoint_reentrancy == Use_tcontrol then 0b0 else v[Breakpoint],
```
```suggestion
    Breakpoint         = if currentlyEnabled(Ext_Sdtrig) & sdtrig_breakpoint_reentrancy == Use_tcontrol then 0b0 else v[Breakpoint],
```

model/extensions/Sdtrig/sdtrig_control.sail
```diff
@@ -0,0 +1,241 @@
...
+  // before delivery.  Since this function is called with the
+  // delivered interrupts, translate them back before deciding whether
+  // the trigger matches.
+  let intr = if del_priv == VirtualSupervisor then match intr {
+               I_S_External => I_VS_External,
+               I_S_Software => I_VS_Software,
+               I_S_Timer    => I_VS_Timer,
+               _            => intr,
+             } else intr;
+
```
nadime15: I don't think this is correct. The spec says:

  > The interrupt number is interpreted in the mode that the trap handler executes in. (E.g. virtualized interrupt numbers are not the same in every mode.)

For example if, say, `I_VS_Software` becomes pending and is delegated to VS-mode (via hideleg), getPendingSet translates it to  `I_S_Software`. The guest now has `I_S_Software` pending, not `I_VS_Software`, and since that is the interrupt the VS-mode handler sees, the trigger has to match on `I_S_Software` without translating it back (VS-mode never dealt with I_VS_Software in the first place).

`I_VS_Software` is only seen when the interrupt is not delegated (the corresponding hideleg bit is 0) and the hypervisor handles it in HS-mode.

```suggestion
```

(`del_priv` can be removed too)

doc/ChangeLog.md
```diff
@@ -1,9 +1,19 @@
 # Release notes for the next version
 
+- The following extensions have been added:
+  - Sdtrig
+    - This support is partial: only icount, itrigger, and etrigger
+      triggers are supported; tdata3 and context CSRs are supported
```
nadime15: double "yet"

nadime15: ups, see one line below.

model/extensions/Sdtrig/sdtrig_control.sail
pmundkur: I think this makes sense, thanks.

model/extensions/Sdtrig/sdtrig_control.sail
```diff
@@ -0,0 +1,223 @@
...
+            if t.count == zeros()
+            then t
+            else
+              // As implemented, this trigger cannot match for
+              // multiple events during a single instruction.
+              if t.count != zero_extend(0b1)
+              then {t with count = t.count - 1}
+              else {t with
+                     count = if icount_count_is_hardwired_one then t.count else zeros(),
+                     pending = true};
```
nadime15: What do you think about the following (potential) simplification?
```suggestion
          // As implemented, this trigger cannot match for
          // multiple events during a single instruction.
          let new_t : ICount = match unsigned(t.count) {
            // count 0 stays at 0 until explicitly written
            0 => t,
            1 => {t with
                   count = if icount_count_is_hardwired_one then t.count else zeros(),
                   pending = true},
            _ => {t with count = t.count - 1},
          };
```

model/core/sys_regs.sail
```diff
@@ -1018,12 +1018,17 @@ function clause is_CSR_accessible(0x200, _, _) = currentlyEnabled(Ext_H) // vsst
...
 function clause write_CSR(0x200, value) = { vsstatus = legalize_sstatus(vsstatus, value, true); Ok(lower_mstatus(vsstatus, true).bits[xlen - 1 .. 0]) }
 
+function read_real_sstatus() -> Sstatus =
+  lower_mstatus(mstatus, false)
+
+function read_nominal_sstatus() -> Sstatus =
+  if inVirtualPrivilege()
+  then lower_mstatus(vsstatus, true)
+  else lower_mstatus(mstatus, false)
+
```
nadime15: Im not sure we need the two new helpers. The only use of read_real_sstatus() reads SIE in the Supervisor case (in prevent_reentrance), where V=0, so mstatus[SIE] says the same thing directly. read_nominal_sstatus() is then just the old read_sstatus() again. I would drop both and use mstatus[SIE] in prevent_reentrance and revert back to upstream.

pmundkur: I think it's preferable to avoid accessing these bits directly when there are such lowering functions involved.  It's also clearer with respect to the spec, since the spec talks about `SIE in sstatus`; the fact that we implement `sstatus` inside `mstatus` is sort of an implementation detail.

nadime15: Fair enough but can we go back to read_sstatus()?

pmundkur: `read_sstatus` is ambiguous, since it could mean the real one, or the nominal one that depends on the virtualization mode.  I think it's better to be more explicit about which case one has in mind.

nadime15: Yeah, I’m not sure. I think that’s just error prone and harder to maintain. Also, the whole codebase already accesses the Supervisor fields directly in mstatus, for example in SRET, the trap handler, and getPendingSet. In theory you would need to refactor now the whole code basis.

It would be good to get a third opinion on this.

@jordancarlin any thoughts on this? I know you haven’t been following the discussion, but maybe you have some insights.

nadime15: Or alternatively, we could get rid of both these helpers, since legalization for sstatus/vsstatus takes place in lower_mstatus anyway. This would at least resolve the discussion about the functions :D

jordancarlin: I haven't looked through all of the code yet, but based on a quick skim, I think I agree with @nadime15. I am in favor of dropping these helpers. I don't think they make the code easier to follow.

pmundkur: I think the question is matching the code to the spec; the spec talks about `SIE in sstatus`, the model doesn't really have `sstatus`, so the `read_real_sstatus()` is the replacement; I'm arguing that accessing `mstatus.SIE` when the spec says `SIE in sstatus` is exposing an implementation detail.

jordancarlin:
It’s not really an implementation detail though because the spec defines sstatus.SIE to be an alias of mstatus.SIE. So it’s not exposing an internal Sail model detail as much as it is using a fact that is clearly spelled out in the spec.

model/core/sys_regs.sail
pmundkur: Removed the helpers.

PR #1589, opened 2026-03-04
Add atomic support to the PMA config.

pmundkur:
This was previously commented out pending support in the memory access type.

Atomic accesses are now checked against the underlying PMAs.

In the default configuration, RAM is given full atomic support (`AMOCASQ`), and MMIO and ROM regions are given none (`AMONone`).

This also removes a `TODO` on marking `phys_access_check` private that is no longer applicable due to its use in the Zicbo{p,m} extensions.

### DIFF
--- a/model/sys/pma.sail
+++ b/model/sys/pma.sail
@@ -49,8 +49,8 @@ function pma_allows_atomic_op forall 'n, 0 < 'n <= max_mem_access . (pma : Atomi
     AMOLogical    => op == AMOSWAP | op == AMOAND | op == AMOOR | op == AMOXOR,
     AMOArithmetic => op != AMOCAS,
     AMOCASW       => op != AMOCAS | width == 4,
-    AMOCASD       => op != AMOCAS | width == 8,
-    AMOCASQ       => op != AMOCAS | width == 16,
+    AMOCASD       => op != AMOCAS | width == 4 | width == 8,
+    AMOCASQ       => op != AMOCAS | width == 4 | width == 8 | width == 16,
   }
 
 enum Reservability = { RsrvNone, RsrvNonEventual, RsrvEventual }


model/sys/mem.sail
```diff
@@ -97,23 +97,17 @@ private function pmaCheck forall 'n, 0 < 'n <= max_mem_access .
             // TODO: assert(not(res_or_con)). This requires a slight refactor of the mem_write_* APIs,
             // since the use of the default Store access conflicts with the passed in aq/rl/con flags.
             Store(Data)            => attributes.writable,
-            // TODO: For LR/SC, we currently don't differentiate between the presence (RsrvEventual)
```
jordancarlin: This should still be a TODO.

model/extensions/A/aext_types.sail
```diff
@@ -6,8 +6,27 @@
...
 // =======================================================================================
 
-enum amoop = {AMOSWAP, AMOADD, AMOXOR, AMOAND, AMOOR,
-              AMOMIN, AMOMAX, AMOMINU, AMOMAXU, AMOCAS}
+// These enums are arranged according to the level of PMA support they require:
+// AMOSwap < AMOLogical < AMOArithmetic < AMOCASW < AMOCASD < AMOCASQ
+enum amoop = {
+  AMOSWAP,
+  AMOAND, AMOOR, AMOXOR,
+  AMOADD, AMOMIN, AMOMAX, AMOMINU, AMOMAXU,
```
arichardson: why is add grouped with min/max? If we're doing manual grouping I'd be inclined to either keep it on a single line or merge with the bit ops.

I would just use whatever line wrapping sail -fmt uses (1 per line?).

model/sys/mem.sail
```diff
@@ -182,7 +178,6 @@ private function highestPriorityAlignmentOrAccessFault  (l : ExceptionType, r :
   if alignmentOrAccessFaultPriority(l) > alignmentOrAccessFaultPriority(r) then l else r
 
 // Check if access is permitted according to PMPs and PMAs.
-// TODO: this can be marked private after https://github.com/riscv/sail-riscv/pull/1405
```
arichardson, line 185: Since you removed this, should it be marked private?

model/sys/pma.sail
```diff
@@ -42,6 +42,18 @@ mapping atomic_support_str : AtomicSupport <-> string = {
 }
 overload to_str = {atomic_support_str}
 
+function pma_allows_atomic_op(pma : AtomicSupport, op : amoop) -> bool =
+  match pma {
+    AMONone       => false,
+    AMOSwap       => op == AMOSWAP,
+    AMOLogical    => op == AMOSWAP | op == AMOAND | op == AMOOR | op == AMOXOR,
+    AMOArithmetic => op != AMOCAS,
+    // TODO: the Atomic memory access does not carry AMOCAS widths
```
Timmmm: The `pmaCheck` function has a `width` parameter though - could you use that?

model/sys/mem.sail
pmundkur: I updated the PR: it is now used in Zicbo{p,m} and so no longer can be marked private.

model/sys/pma.sail
pmundkur: Good point, I used it.

model/extensions/A/aext_types.sail
pmundkur: I just made it a single line.

model/sys/pma.sail
```diff
@@ -42,6 +42,17 @@ mapping atomic_support_str : AtomicSupport <-> string = {
...
 
+function pma_allows_atomic_op forall 'n, 0 < 'n <= max_mem_access . (pma : AtomicSupport, op : amoop, width : int('n)) -> bool =
+  match pma {
+    AMONone       => false,
+    AMOSwap       => op == AMOSWAP,
+    AMOLogical    => op == AMOSWAP | op == AMOAND | op == AMOOR | op == AMOXOR,
+    AMOArithmetic => op != AMOCAS,
+    AMOCASW       => op != AMOCAS | width == 4,
+    AMOCASD       => op != AMOCAS | width == 8,
+    AMOCASQ       => op != AMOCAS | width == 16,
```
Timmmm:
```suggestion
    AMOCASW       => op != AMOCAS | width <= 4,
    AMOCASD       => op != AMOCAS | width <= 8,
    AMOCASQ       => op != AMOCAS | width <= 16,
```

I think?

pmundkur: I used the exact widths for the CAS instructions.

Timmmm: Right, I mean you should be able to do a width 4 AMOCAS if the PMA is AMOCASD. I think your current code won't allow that.

Timmmm: Oh you fixed it, never mind. I feel like `<=` is simpler and clearer but that works too.

PR #1592, opened 2026-03-05
Use a wrapper to ensure that `xtval_exception_value` is used for constructed traps.

pmundkur:
It was easy to forget to call `xtval_exception_value` when constructing a `sync_exception` struct.  The wrapper addresses this and also makes it more succinct.

This fixes the missed calls to `xtval_exception_value` in the traps for Zicfiss and Zicfilp.

### DIFF
--- a/model/extensions/I/base_insts.sail
+++ b/model/extensions/I/base_insts.sail
@@ -542,7 +542,6 @@ function clause execute ECALL() = {
     VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
     VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
   };
-  // No need for the `make_sync_exception` wrapper since there is no excinfo for tval.
   let t : sync_exception = struct {
     trap    = trap,
     excinfo = None(),


model/sys/sys_control.sail
```diff
@@ -323,10 +323,13 @@ function xtval_exception_value(e : ExceptionType, excinfo : xlenbits) -> option(
   } then Some(excinfo) else None()
 }
 
+function make_sync_exception(e : ExceptionType, xtval : xlenbits, ext : option(ext_exception)) -> sync_exception =
+  struct { trap    = e,
+           excinfo = xtval_exception_value(e, xtval),
+           ext     = ext }
```
jordancarlin: Is it worth simplifying this to drop the `ext` argument and have the wrapper always set `ext` to `None()`? Anywhere that needs to actually use `ext_exception` can create the struct directly.

pmundkur: `ext` is used in the Codasip version of CHERI, so I'm in favor of keeping it in the wrapper until we're sure we don't need it.  @Timmmm ?

nadime15:
I have been working on the hypervisor branch (https://github.com/nadime15/sail-riscv/tree/hypervisor) which overlaps heavily with this area. I have already spent a lot of time on unnecessary rebasing because of changes that need to be refactored anyway once hypervisor support is added, and this PR would add to that.

Since this area is going to be refactored anyway, I would really prefer we hold off on merging this for now. Can we sync on timing so we can avoid duplicating the effort?

pmundkur:
I'm not sure it's reasonable to expect the master branch to stall on any changes that affect the hypervisor until that is fully ready.  It's unfortunate, but I think dealing with rebasing is unavoidable.

nadime15:
Just to clarify, I am not asking for the master branch to stall on _any_ changes until the hypervisor is ready. I'm asking about one specific PR that adds no new functionality and touches code that will be refactored anyway.

Those are two very different things. Rebasing on meaningful changes is a normal part of the process and I have no issue with that. But rebasing on a pure code reorganisation that provides no value and will be redone during the refactor is avoidable overhead, and that distinction matters when the work it's blocking is the team's top priority.

As you know, hypervisor support is our top priority, so it makes sense to take a long-term perspective and focus on facilitating integration rather than creating obstacles.

I don't think that's an unreasonable ask.

If you take a look at this branch (https://github.com/nadime15/sail-riscv/tree/hypervisor) you will see that your proposed change is already included here, so the change this PR does is not totally lost.

pmundkur:
> If you take a look at this branch (https://github.com/nadime15/sail-riscv/tree/hypervisor) you will see that your proposed change is already included here, so the change this PR does is not totally lost.

It's looking good, but that branch doesn't contain the fix in this PR: "This fixes the missed calls to xtval_exception_value in the traps for Zicfiss and Zicfilp."

This isn't a "pure code reorganisation that provides no value"; it's actually fixing a bug.

nadime15:
After reading both the PR description and the code, it's really hard to tell there's an actual bug being fixed here, it reads like a pure cleanup. The Zicfiss and Zicfilp cases were ignoring `software_check_fault_writes_xtval`, which is a real issue, fair enough. Arguably just removing the hardcoded `Some()` would have been enough of a fix without the broader refactor, but I understand the intention.

model/sys/sys_control.sail
Timmmm: I think it would be ok to drop the ext argument from `make_sync_exception`. CHERI can add `make_sync_exception_ext()` or something like that that does have the `ext` argument.

model/extensions/I/base_insts.sail
```diff
@@ -542,11 +542,8 @@ function clause execute ECALL() = {
     VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
     VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
   };
-  let t : sync_exception = struct {
-    trap    = trap,
-    excinfo = None(),
-    ext     = None(),
-  };
+  // xtval (i.e. zeros()) is an ignored value for ECALL exceptions.
+  let t : sync_exception = make_sync_exception(trap, zeros(), None());
```
Timmmm: I'm not a fan of the `zeros()` here though. Maybe just leave this instance as-is? Or change the argument to `option()` but I think that would be more verbose since this is the only time you want `none`.

model/sys/sys_control.sail
```diff
@@ -323,10 +323,13 @@ function xtval_exception_value(e : ExceptionType, excinfo : xlenbits) -> option(
...
+function make_sync_exception(e : ExceptionType, xtval : xlenbits) -> sync_exception =
+  struct { trap    = e,
+           excinfo = xtval_exception_value(e, xtval),
+           ext     = None() }
+
 function handle_exception(xtval : xlenbits, e : ExceptionType) -> unit = {
-  let t : sync_exception = struct { trap    = e,
-                                    excinfo = xtval_exception_value(e, xtval),
-                                    ext     = None() };
+  let t : sync_exception = make_sync_exception(e, xtval);
```
Timmmm: I think you could probably inline this one (and the one in `EBREAK`), if you wanted. Not a deal-breaker though.

model/extensions/I/base_insts.sail
```diff
@@ -542,6 +542,7 @@ function clause execute ECALL() = {
     VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
     VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
   };
+  // No need for the `make_sync_exception` wrapper since there is no excinfo for tval.
```
arichardson: It seems like we should just call the wrapper everywhere. We can pass an arbitrary tval argument (e.g. zero) and let the wrapper set it to None() (which it already does). This seems more consistent than adding this comment.

Timmmm: Ha I dunno if you saw but I said the opposite in a previous thread. I don't like passing the arbitrary tval argument if it isn't used; not as type correct.

arichardson: That's a good point, I just didn't like the need for a comment why it's not being used. Maybe we could add a `make_sync_exception_no_ttval` wrapper?

Timmmm: Yeah could do but it would only be used here...

Timmmm: Could just delete the comment tbf.

pmundkur: I agree; that wrapper would make sense when there is more than one place where it is used.  Right now, there would be only one.
I think the comment is useful in that it's explaining why the usual wrapper is not used (perhaps not obvious to a casual spec reader).

arichardson: Yeah I don't mind either way, I just thought the comment was actually more confusing than helpful.

pmundkur: Ok, I removed the comment.

PR #1593, opened 2026-03-05
add unratified RISC-V Zbr0p93 ext

KotorinMinami:
Add unratified RISC-V Zbr0p93 ext
This is part of B extension.
We can optional support it.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -348,6 +348,9 @@
     "Zbs": {
       "supported": false
     },
+    "Zbr": {
+      "supported": false
+    },
     "Zbc": {
       "supported": true
     },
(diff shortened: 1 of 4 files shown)


jrtc27:
Spec link? And what pre-ratification state is it in?

KotorinMinami:
Simply inspired by this [patch](https://patchew.org/QEMU/20260304121113.117299-1-james.wainwright@lowrisc.org/20260304121113.117299-3-james.wainwright@lowrisc.org)

jordancarlin:
Based on the conversation on that patch, it looks like this is not something that RISC-V plans to ratify, so I don’t think it belongs in the sail model.

jordancarlin:
@KotorinMinami, can we close this?

KotorinMinami:
Oh! yes

PR #1595, opened 2026-03-05
Add a missed check for invalid PTE.

pmundkur:
"For non-leaf PTEs, the D, A, and U bits are reserved for future standard use."

"... if any bits or encodings that are reserved for future standard use are set within pte, stop and raise a page-fault exception corresponding to the original access type."

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -8,7 +8,8 @@
   - Zicfiss
 
 - D, A, U and other bits that are reserved in non-leaf PTEs now raise a page-fault exception.
-  This is a backwards incompatible change.
+  This is a backwards incompatible change, and is required since version 1.12 (also known as version 20211203)
+  of the privileged specification.
 
 - A Mac ARM binary release is now available.
 


jordancarlin:
See #677. I think we should go forward with this even without spec versioning and then gate it on the spec version once we support configurable versions, but it is a potentially breaking change.

pmundkur:
Thanks, I'd missed that.  I just saw this missed check when working on Svnapot.  I agree: we should add version gating later.

Timmmm:
Yeah sorry I never got around to finishing that. Maybe just remove the added line below the comment instead of before it?

model/sys/vmem_pte.sail
```diff
@@ -86,6 +86,7 @@ private function pte_is_invalid(pte_flags : PTE_Flags, pte_ext : PTE_Ext) -> boo
   // TODO: handle V=1 and henvcfg.SSE=0 at VS/VU levels
   | (pte_flags[R] == 0b0 & pte_flags[W] == 0b1 & pte_flags[X] == 0b0 & menvcfg[SSE] == 0b0)
   | (pte_flags[R] == 0b0 & pte_flags[W] == 0b1 & pte_flags[X] == 0b1)
+  | (pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1))
```
Timmmm: Also you need this:

```suggestion
  | (pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros())
```

doc/ChangeLog.md
```diff
@@ -1,5 +1,8 @@
 # Release notes for the next version
 
+- D, A, U and other bits that are reserved in non-leaf PTEs now raise a page-fault exception.
+  This is a backwards incompatible change.
```
jordancarlin: Maybe mention that this is required starting in priv 1.12?

PR #1596, opened 2026-03-06
Add the Svnapot extension.

pmundkur:


model/postlude/validate_config.sail
```diff
@@ -294,7 +294,17 @@ private function check_misc_extension_dependencies() -> bool = {
...
   if hartSupports(Ext_Ssqosid) & not(hartSupports(Ext_Zicsr)) then {
     valid = false;
-    print_endline("The Ssqosid extensions is enabled but Zicsr is disabled: supporting Ssqosid requires Zicsr.");
+    print_endline("The Ssqosid extension is enabled but Zicsr is disabled: supporting Ssqosid requires Zicsr.");
+  };
+  if (config extensions.Svnapot.supported : bool) then {
+    if xlen == 32 then {
+      valid = false;
+      print_endline("The Svnapot extension is enabled: Svnapot is not supported on RV32");
+    };
```
ved-rivos: Same question here. If extensions.Svnapot.support is already factoring in @CONFIG_XLEN_IS_64@

jordancarlin: @CONFIG_XLEN_IS_64@ is used when creating our default configurations, but a user can provide any configuration, so we need to validate it.

model/sys/vmem_pte.sail
```diff
@@ -90,7 +90,7 @@ private function pte_is_invalid(pte_flags : PTE_Flags, pte_ext : PTE_Ext) -> boo
   // Early versions do not require this check. This will need to be behind
   // a flag when the Sail model allows specifying the spec version.
   | (pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros()))
-  | (pte_ext[N] != 0b0 & not(currentlyEnabled(Ext_Svnapot)))
+  | (pte_ext[N] != 0b0 & (not(currentlyEnabled(Ext_Svnapot)) | level > 0))
```
Timmmm: I don't think you need to change this function. It already checks that `N` is 0 for non-leaf PTEs, so you know that N can only be 1 for the leaf PTE. Then ... (follow me to the other thread)

model/sys/vmem.sail
```diff
@@ -154,6 +154,13 @@ function pt_walk(
                 // Superpage PPN @ lower VPNs @ page-offset
                 let low_bits = ppn_size_bits * level;
                 ppn[length(ppn) - 1 .. low_bits] @ vpn[low_bits - 1 .. 0]
```
Timmmm, line 159: (continuing from the other thread)... you can just check N is not set here.

```
if pte_ext[N] == 0b1 {
  // Reserved for level > 0
  return Err(PTW_Invalid_PTW(), ext_ptw);
}
```

model/sys/vmem.sail
```diff
@@ -154,6 +154,13 @@ function pt_walk(
...
+              } else if currentlyEnabled(Ext_Svnapot) & pte_ext[N] == 0b1 then {
+                // Only 64K NAPOT PTEs are valid.
+                let pte_napot_bits = 4;
+                if ppn[pte_napot_bits - 1 .. 0] != 0b1000
+                then return Err(PTW_Invalid_PTE(), ext_ptw);
+
+                ppn[length(ppn) - 1 .. pte_napot_bits] @ vpn[pte_napot_bits - 1 .. 0]
```
Timmmm:
```suggestion
                [ppn with pte_napot_bits - 1 .. 0 = vpn[pte_napot_bits - 1 .. 0]]
```

Maybe better?

pmundkur: Maybe, but that's not valid syntax.

model/sys/vmem.sail
pmundkur: That works, but I think it's better to centralize the PTE-related checks in `vmem_pte`.

model/sys/vmem.sail
Timmmm: Ah you just need brackets:

```
  [ppn with (pte_napot_bits - 1) .. 0 = vpn[pte_napot_bits - 1 .. 0]]
```

model/sys/vmem.sail
Timmmm: Shouldn't we do the PPN check in that function too then? It seems weird to do half the check inside and half outside.

Timmmm: I think doing it outside has the advantage that we could support the extended scheme easily if it is ever specified:

{D847E032-89BA-4210-A735-E6CB500B7C88}

pmundkur: Yes, it's suboptimal, but the `pte_is_invalid` function doesn't have the PPN, it's just checking the flags.  The PPN check here is similar to the super-page alignment check for the PPN above.

Timmmm: Yeah that's kind of why I think it should be checked outside that function. @jordancarlin @nadime15 any opinion?

nadime15: @Timmmm I am not sure I follow, are you saying you dont want to move the napot ppn pattern check (`ppn[3..0] == 0b1000`) into `pte_is_invalid` because it's tied to the translation logic and potential future encodings? And that the superpage misalignment check could move in since its just a validity check? 

So something like this?

```
match napot_page_size(pte_ext, ppn) {
  None()       => return Err(PTW_Invalid_PTE(), ext_ptw),
  Some(size)   => {
    let napot_bits = size - pagesize_bits;  // 16 - 12 = 4
    [ppn with (napot_bits - 1) .. 0 = vpn[napot_bits - 1 .. 0]]
  }
}

private val napot_page_size : forall 'v, is_sv_mode('v) . (
  PTE_Ext,
  ppn_bits('v),
) -> option(nat)
function napot_page_size(pte_ext, ppn) =
  if pte_ext[N] != 0b1 then None()
  else match ppn[3..0] {
    0b1000 => Some(16),  // 64K = 2^16
    _      => None()     // invalid or future encoding
  }
```

pmundkur: If we pull the (level > 0) check for Svnapot out of `vmem_pte::pte_is_invalid()`, then won't we have to repeat the check twice in `vmem::ptw_walk()`, once in both branches of `pte_is_non_leaf(pte_flags)`?

Timmmm: No, because we already verify that for non-leaf entries `N` must be zero. Let me make a commit with my suggestion for clarify, 1 sec...

Timmmm: Like this: https://github.com/riscv/sail-riscv/compare/master...Timmmm:sail-riscv:user/timh/svnapot_suggestion

And actually what I think we *should* write is something like this:

```
            PTE_Check_Success(ext_ptw) => {
              // Bits replaced with VPN due to level or Svnapot.
              let low_bits = if pte_ext[N] == 0b0 then {
                ppn_size_bits * level
              } else {
                count_trailing_zeros(ppn) + 1
              };

              // Only 64k NAPOT PTEs are valid.
              if pte_ext[N] == 0b1 & low_bits != 4 then return Err(PTW_Invalid_PTE(), ext_ptw);

              let ppn = [ppn with low_bits >.. 0 = vpn[low_bits >.. 0];

              ptw_success_callback(zero_extend(ppn), level);

              Ok(struct {ppn=ppn, pte=pte, pteAddr=pte_addr, level=level, global=global}, ext_ptw)
            }
```

That captures (I think) the intended encoding for Svnapot and the connection with non-Svnapot superpages.

It's a bit awkward without support for `>..` though so let's not actually do that now!

Timmmm: I guess we could do

```
            PTE_Check_Success(ext_ptw) => {
              // Bits replaced with VPN due to level or Svnapot.
              let low_bits = if pte_ext[N] == 0b0 then {
                ppn_size_bits * level
              } else {
                count_trailing_zeros(ppn) + 1
              };

              // Only 64k NAPOT PTEs are valid.
              if pte_ext[N] == 0b1 & low_bits != 4 then return Err(PTW_Invalid_PTE(), ext_ptw);

              let ppn = if low_bits == 0 then ppn else [ppn with (low_bits - 1) .. 0 = vpn[low_bits - 1 .. 0]];

              ptw_success_callback(zero_extend(ppn), level);

              Ok(struct {ppn=ppn, pte=pte, pteAddr=pte_addr, level=level, global=global}, ext_ptw)
            }
```

What do you think?

pmundkur: > No, because we already verify that for non-leaf entries `N` must be zero. 

> Like this: [master...Timmmm:sail-riscv:user/timh/svnapot_suggestion](https://github.com/riscv/sail-riscv/compare/master...Timmmm:sail-riscv:user/timh/svnapot_suggestion)

That check only ensures that N must be 0 at all levels if Svnapot is not supported.  If Svnapot is supported, it will incorrectly allow N to be 1 on PTEs that are non-leaf pointers.  (i.e. the `if pte_is_non_leaf(pte_flags) then ... pt_walk()` branch)

Timmmm: This line checks that N is zero for non-leaf PTEs:

```
| (pte_is_non_leaf(pte_flags) & (pte_flags[A] == 0b1 | pte_flags[D] == 0b1 | pte_flags[U] == 0b1 | pte_ext.bits != zeros()))
```

pmundkur: Ah, I missed that.  I agree: your approach should adapt better to the extended schemes in the future.

Timmmm (approved):
Thanks, LGTM.

PR #1597, opened 2026-03-06
Refactor feature_enabled_for_priv() to return tri-state result

nadime15:
Replace the bool return type with `FeatureEnabledResult` to distinguish between illegal-instruction and virtual-instruction exceptions for VS/VU mode accesses.

This is required for correct hypervisor extension support, specifically for counter CSR accesses from VS/VU mode where the exception type depends on the mcounteren/hcounteren state (see priv 15.6.1. Trap Cause Codes).

The same fix will also benefit Zicbom and Zicboz once henvcfg is implemented (see unpriv 19.5). Existing callers are migrated to feature_enabled_for_priv_bool for backward compatibility.

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -840,10 +840,7 @@ function feature_enabled_for_priv(p : Privilege, machine_enable_bit : bits(1), s
 // do not yet distinguish between illegal-instruction and virtual-instruction
 // exceptions. Required for upcoming Hypervisor extension support.
 function feature_enabled_for_priv_bool(p : Privilege, m : bits(1), s : bits(1), h : bits(1)) -> bool =
-  match feature_enabled_for_priv(p, m, s, h) {
-    FEATURE_ENABLED => true,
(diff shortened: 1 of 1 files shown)


model/core/sys_regs.sail
```diff
@@ -817,16 +817,35 @@ val get_16_random_bits = impure {
...
+}
+
+// Backward-compatible wrapper around feature_enabled_for_priv for callers that
+// do not yet distinguish between illegal-instruction and virtual-instruction
+// exceptions. Required for upcoming Hypervisor extension support.
+function feature_enabled_for_priv_bool(p : Privilege, m : bits(1), s : bits(1), h : bits(1)) -> bool =
+  match feature_enabled_for_priv(p, m, s, h) {
+    FEATURE_ENABLED => true,
+    _               => false,
+  }
```
```suggestion
  feature_enabled_for_priv(p, m, s, h) == FEATURE_ENABLED
```

Timmmm (approved):

PR #1601, opened 2026-03-09
Gate ptw_fail_callback on config_print_ptw flag

nadime15:
`ptw_fail_callback` was unconditionally printing PTW failures unlike the other PTW callbacks which are gated on `config_print_ptw`.

### DIFF
--- a/c_emulator/riscv_callbacks_log.cpp
+++ b/c_emulator/riscv_callbacks_log.cpp
@@ -154,8 +154,7 @@ void log_callbacks::ptw_fail_callback(
   int64_t level,
   sbits pte_addr
 ) {
-  // failed trace is always available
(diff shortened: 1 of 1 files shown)


nadime15:
Yeah, I think so too. However, its quite confusing because I can’t tell whether these logs are coming from the binary I am testing or from sail. TLB misses arent actual errors, but they can spam the logs so much that it becomes impossible to see anything useful, so I dont see why they should be printed every time.

Timmmm:
I imagine it was like this because normally when debugging you are only really interested in failed PTWs. But I agree it probably shouldn't be printed by default. We might want an extra flag (`--trace-ptw-failures` or whatever) but let's do this for now.

PR #1624, opened 2026-03-24
Add Hypervisor Virtual Instruction Exception Handling for CSR Accesses

nadime15:
Virtual instruction exceptions for CSR accesses from VS/VU mode were previously not handled. This PR introduces a scattered function `is_CSR_exception_virtual` whose default clause implements the general HS-qualified rule from the H extension spec. CSRs that deviate from this rule have explicit clauses: satp, seed, stimecmp, ssp, cycle/time/instret and hpmcounters, and scountovf.

`check_CSR` is refactored to ensure failures in access type, stateen, or privilege checks always raise illegal instruction exceptions and are never promoted to virtual instruction exceptions by the HS-qualified rule.

A simpler alternative to #1556.

### DIFF
--- a/model/extensions/K/zkr_control.sail
+++ b/model/extensions/K/zkr_control.sail
@@ -52,7 +52,5 @@ function clause is_CSR_accessible(0x015, priv, access_type) =
     VirtualUser       => false,
   })
 
-function clause is_CSR_exception_virtual(0x015, _, access_type) = mseccfg[SSEED] == 0b1 & access_type == CSRReadWrite
-
 function clause read_CSR(0x015) = read_seed_csr()
 function clause write_CSR(0x015, _value) = Ok(write_seed_csr())
--- a/model/extensions/K/zkr_control.sail
+++ b/model/extensions/K/zkr_control.sail
@@ -43,15 +43,16 @@ private function write_seed_csr () -> xlenbits = zeros()
 mapping clause csr_name_map = 0x015  <-> "seed"
 function clause is_CSR_accessible(0x015, priv, access_type) =
   currentlyEnabled(Ext_Zkr) &
-  // Read-only access is not allowed.
-  (access_type != CSRRead) &
-  (match priv {
-    Machine => true,
-    Supervisor => mseccfg[SSEED] == 0b1,
-    User => mseccfg[USEED] == 0b1,
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualUser => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-  })
+  access_type != CSRRead &
+  match priv {
+    Machine           => true,
+    User              => mseccfg[USEED] == 0b1,
+    Supervisor        => mseccfg[SSEED] == 0b1,
+    VirtualSupervisor => false,
+    VirtualUser       => false,
+  }
+
+function clause is_CSR_exception_virtual(0x015, p, access_type) = mseccfg[SSEED] == 0b1 & access_type == CSRReadWrite
 
 function clause read_CSR(0x015) = read_seed_csr()
 function clause write_CSR(0x015, _value) = Ok(write_seed_csr())
--- a/model/extensions/K/zkr_control.sail
+++ b/model/extensions/K/zkr_control.sail
@@ -44,15 +44,15 @@ mapping clause csr_name_map = 0x015  <-> "seed"
 function clause is_CSR_accessible(0x015, priv, access_type) =
   currentlyEnabled(Ext_Zkr) &
   access_type != CSRRead &
-  match priv {
+  (match priv {
     Machine           => true,
     User              => mseccfg[USEED] == 0b1,
     Supervisor        => mseccfg[SSEED] == 0b1,
     VirtualSupervisor => false,
     VirtualUser       => false,
-  }
+  })
 
-function clause is_CSR_exception_virtual(0x015, p, access_type) = mseccfg[SSEED] == 0b1 & access_type == CSRReadWrite
+function clause is_CSR_exception_virtual(0x015, _, access_type) = mseccfg[SSEED] == 0b1 & access_type == CSRReadWrite
 
 function clause read_CSR(0x015) = read_seed_csr()
 function clause write_CSR(0x015, _value) = Ok(write_seed_csr())
--- a/model/extensions/Sscofpmf/sscofpmf.sail
+++ b/model/extensions/Sscofpmf/sscofpmf.sail
@@ -78,7 +78,7 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
 }
 
 // scountovf
-function clause is_CSR_accessible(0xDA0, p, _) =
+function clause is_CSR_accessible(0xDA0, _, _) =
   currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
 // TODO: when Smcdeleg/Ssccfg is implemented, extend to:
 //   & not(p == VirtualSupervisor & currentlyEnabled(Ext_Smcdeleg)
--- a/model/extensions/Sscofpmf/sscofpmf.sail
+++ b/model/extensions/Sscofpmf/sscofpmf.sail
@@ -78,11 +78,13 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
 }
 
 // scountovf
-function clause is_CSR_accessible(0xDA0, p, _) = currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
+function clause is_CSR_accessible(0xDA0, priv, _) =
+  currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S) &
+  (match priv {
+    VirtualSupervisor => false, /* TODO: not(menvcfg[CDE] == 0b1) */
+    VirtualUser       => false, /* TODO: not(menvcfg[CDE] == 0b1) */
+    _                 => true,
+  })
 
-// TODO: Smcdeleg is not yet implemented (including menvcfg[CDE]).
-// Once implemented, the is_CSR_exception_virtual clause will
-// correctly raise virtual instruction exceptions when menvcfg.CDE=1.
-// function clause is_CSR_exception_virtual(0xDA0, p, _) = currentlyEnabled(Ext_Smcdeleg) & menvcfg[CDE] == 0b1 & (p == VirtualSupervisor | p == VirtualUser)
 function clause read_CSR(0xDA0) = zero_extend(get_scountovf(cur_privilege))
 // scountovf is read-only.
--- a/model/extensions/Sscofpmf/sscofpmf.sail
+++ b/model/extensions/Sscofpmf/sscofpmf.sail
@@ -78,13 +78,7 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
 }
 
 // scountovf
-function clause is_CSR_accessible(0xDA0, p, _) =
-  currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S) &
-  (match p {
-    VirtualSupervisor => true,
-    VirtualUser       => false,
-    _                 => true,
-  })
+function clause is_CSR_accessible(0xDA0, p, _) = currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
 
 // TODO: Smcdeleg is not yet implemented (including menvcfg[CDE]).
 // Once implemented, the is_CSR_exception_virtual clause will
--- a/model/extensions/Sscofpmf/sscofpmf.sail
+++ b/model/extensions/Sscofpmf/sscofpmf.sail
@@ -78,13 +78,11 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
 }
 
 // scountovf
-function clause is_CSR_accessible(0xDA0, priv, _) =
-  currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S) &
-  (match priv {
-    VirtualSupervisor => false, /* TODO: not(menvcfg[CDE] == 0b1) */
-    VirtualUser       => false, /* TODO: not(menvcfg[CDE] == 0b1) */
-    _                 => true,
-  })
+function clause is_CSR_accessible(0xDA0, p, _) =
+  currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
+// TODO: when Smcdeleg/Ssccfg is implemented, extend to:
+//   & not(p == VirtualSupervisor & currentlyEnabled(Ext_Smcdeleg)
+//   & currentlyEnabled(Ext_Ssccfg) & menvcfg[CDE] == 0b1)
 
 function clause read_CSR(0xDA0) = zero_extend(get_scountovf(cur_privilege))
 // scountovf is read-only.
--- a/model/extensions/Sscofpmf/sscofpmf.sail
+++ b/model/extensions/Sscofpmf/sscofpmf.sail
@@ -71,13 +71,21 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
   match priv {
     Machine           => overflow,
     Supervisor        => overflow & mcounteren.bits,
+    VirtualSupervisor => overflow & mcounteren.bits & hcounteren.bits,
     User              => internal_error(__FILE__, __LINE__, "scountovf not readable from User mode"),
-    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualUser       => internal_error(__FILE__, __LINE__, "scountovf not readable from VirtualUser mode"),
   }
 }
 
 // scountovf
-function clause is_CSR_accessible(0xDA0, _, _) = currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
+function clause is_CSR_accessible(0xDA0, p, _) =
+  currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S) &
+  match p {
+    VirtualSupervisor => true,
+    VirtualUser       => false,
+    _                 => true,
+  }
+
+function clause is_CSR_exception_virtual(0xDA0, p, _) = currentlyEnabled(Ext_Smcdeleg) & menvcfg[CDE] == 0b1 & (p == VirtualSupervisor | p == VirtualUser)
 function clause read_CSR(0xDA0) = zero_extend(get_scountovf(cur_privilege))
 // scountovf is read-only.
--- a/model/extensions/Sscofpmf/sscofpmf.sail
+++ b/model/extensions/Sscofpmf/sscofpmf.sail
@@ -71,7 +71,7 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
   match priv {
     Machine           => overflow,
     Supervisor        => overflow & mcounteren.bits,
-    VirtualSupervisor => overflow & mcounteren.bits & hcounteren.bits,
+    VirtualSupervisor => overflow & mcounteren.bits /* TODO: & hcounteren.bits */,
     User              => internal_error(__FILE__, __LINE__, "scountovf not readable from User mode"),
     VirtualUser       => internal_error(__FILE__, __LINE__, "scountovf not readable from VirtualUser mode"),
   }
@@ -80,12 +80,15 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
 // scountovf
 function clause is_CSR_accessible(0xDA0, p, _) =
   currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S) &
-  match p {
+  (match p {
     VirtualSupervisor => true,
     VirtualUser       => false,
     _                 => true,
-  }
+  })
 
-function clause is_CSR_exception_virtual(0xDA0, p, _) = currentlyEnabled(Ext_Smcdeleg) & menvcfg[CDE] == 0b1 & (p == VirtualSupervisor | p == VirtualUser)
+// TODO: Smcdeleg is not yet implemented (including menvcfg[CDE]).
+// Once implemented, the is_CSR_exception_virtual clause will
+// correctly raise virtual instruction exceptions when menvcfg.CDE=1.
+// function clause is_CSR_exception_virtual(0xDA0, p, _) = currentlyEnabled(Ext_Smcdeleg) & menvcfg[CDE] == 0b1 & (p == VirtualSupervisor | p == VirtualUser)
 function clause read_CSR(0xDA0) = zero_extend(get_scountovf(cur_privilege))
 // scountovf is read-only.
--- a/model/extensions/Sstc/sstc.sail
+++ b/model/extensions/Sstc/sstc.sail
@@ -11,14 +11,14 @@ mapping clause csr_name_map = 0x15D  <-> "stimecmph"
 
 private function stimecmp_accessible(priv : Privilege) -> bool =
   currentlyEnabled(Ext_S) & currentlyEnabled(Ext_Sstc) &
-  match priv {
+  (match priv {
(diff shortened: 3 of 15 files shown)


model/extensions/cfi/zicfiss_regs.sail
```diff
@@ -30,31 +30,32 @@ function zicfiss_xSSE(priv : Privilege) -> bool =
...
-
+function clause is_CSR_accessible(0x011, p, _) = ssp_accessible(p)
+function clause is_CSR_exception_virtual(0x011, p, _) =
+  // NOTE: Double check conditions
+  currentlyEnabled(Ext_Zicfiss) &
+  menvcfg[SSE] == 0b1 & (match p {
+    VirtualSupervisor => false /* TODO: read_henvcfg()[SSE] == 0b0 */,
+    VirtualUser       => /* TODO: read_henvcfg()[SSE] == 0b0 | */ read_senvcfg()[SSE] == 0b0,
+    _                 => false,
+  })
```
radimkrcmar: `ssp` doesn't need a special case.

model/extensions/Sscofpmf/sscofpmf.sail
```diff
@@ -71,13 +71,24 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
...
 
 // scountovf
-function clause is_CSR_accessible(0xDA0, _, _) = currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
+function clause is_CSR_accessible(0xDA0, p, _) =
+  currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S) &
+  (match p {
+    VirtualSupervisor => true,
+    VirtualUser       => false,
+    _                 => true,
+  })
```
radimkrcmar: Shouldn't the access from VirtualUser be denied by check_CSR_priv?

model/extensions/Sstc/sstc.sail
```diff
@@ -5,24 +5,37 @@
...
-function clause is_CSR_accessible(0x14D, priv, _) =
-    currentlyEnabled(Ext_S) & currentlyEnabled(Ext_Sstc) &
-    sstc_CSRs_accessible(priv)
+// stimecmp raises virtual instruction exception in VS mode when
+// mcounteren/menvcfg allow access but hcounteren/henvcfg block it
+function clause is_CSR_exception_virtual(0x14D, p, _) =
+  p == VirtualSupervisor &
+  currentlyEnabled(Ext_S) & currentlyEnabled(Ext_Sstc) &
+   mcounteren[TM] == 0b1  & menvcfg[STCE] == 0b1
+   /* TODO: & (hcounteren[TM] == 0b0  | henvcfg[STCE] == 0b0) */
```
radimkrcmar: I think `stimecmp` is following the HS-qualified rule.

model/extensions/Sstc/sstc.sail
```diff
@@ -5,24 +5,37 @@
...
 
-function clause is_CSR_accessible(0x15D, priv, _) =
-    currentlyEnabled(Ext_S) & currentlyEnabled(Ext_Sstc) & xlen == 32 &
-    sstc_CSRs_accessible(priv)
+function clause is_CSR_exception_virtual(0x15D, p, _) =
+  is_CSR_exception_virtual(0x14D, p, CSRRead) & xlen == 32
```
radimkrcmar: This should be done for all high CSRs, so I think it would be good to abstract it.  We don't need to do anything right now since the current implementation only breaks when Supervisor is 64 bit while Virtual mode is 32 bit, which we don't yet support.

model/extensions/Zicntr/zicntr_control.sail
```diff
@@ -22,6 +22,13 @@ function clause is_CSR_accessible(0xC80, priv, _) = currentlyEnabled(Ext_Zicntr)
 function clause is_CSR_accessible(0xC81, priv, _) = currentlyEnabled(Ext_Zicntr) & xlen == 32 & counter_enabled(1, priv) // timeh
 function clause is_CSR_accessible(0xC82, priv, _) = currentlyEnabled(Ext_Zicntr) & xlen == 32 & counter_enabled(2, priv) // instreth
 
+function clause is_CSR_exception_virtual(0xC00, p, _) = feature_enabled_for_priv(p, mcounteren.bits[0], scounteren.bits[0], 0b0 /* TODO: hcounteren.bits[0] */) == FEATURE_VIRTUAL
+function clause is_CSR_exception_virtual(0xC01, p, _) = feature_enabled_for_priv(p, mcounteren.bits[1], scounteren.bits[1], 0b0 /* TODO: hcounteren.bits[1] */) == FEATURE_VIRTUAL
+function clause is_CSR_exception_virtual(0xC02, p, _) = feature_enabled_for_priv(p, mcounteren.bits[2], scounteren.bits[2], 0b0 /* TODO: hcounteren.bits[2] */) == FEATURE_VIRTUAL
```
radimkrcmar: All these CSRs (+ the hpm counters lower) seem to follow the HS-qualified rule as well.

model/sys/sys_control.sail
```diff
@@ -44,11 +44,29 @@ function check_CSR_priv(csr : csreg, p : Privilege) -> bool =
...
+  CSR_Virtual  : unit,
+}
+
+function check_CSR(csr : csreg, p : Privilege, access_type : CSRAccessType) -> CSRCheckResult = {
+  let access_ok  = check_CSR_access(csr, access_type);
+  let stateen_ok = stateen_allows_CSR_access(csr, p, access_type);
+  let priv_ok    = check_CSR_priv(csr, p);
+  // NOTE: stateen-blocked accesses are never virtual
+  //       read-only violations are never virtual
+  //       lower privilege mode accessing higher privilege mode csr are never virtual
```
radimkrcmar: stateen and priv failures can lead to virtual exceptions.
(Read only can't, but we don't really need to special-case if it only complicates the code -- the end result will be the same as S-mode can't write to read-only CSRs either.)

model/sys/vmem.sail
```diff
@@ -186,9 +186,19 @@ termination_measure pt_walk(_,_,_,_,_,_,_,level,_, _) = level
...
+    VirtualSupervisor => currentlyEnabled(Ext_H) /* TODO: & hstatus[VTVM] == 0b0 */,
+    User              => false,
+    VirtualUser       => false,
+  }
+
 register satp : xlenbits
 mapping clause csr_name_map = 0x180  <-> "satp"
-function clause is_CSR_accessible(0x180, priv, _) = currentlyEnabled(Ext_S) & not(priv == Supervisor & mstatus[TVM] == 0b1)
+function clause is_CSR_accessible(0x180, priv, _) = satp_accessible(priv)
+function clause is_CSR_exception_virtual(0x180, p, _) = p == VirtualSupervisor /* TODO: & hstatus[VTVM] == 0b1 */
```
radimkrcmar: It is a virtual exception even when accessed in U-mode.

model/sys/vmem.sail
```diff
@@ -186,9 +186,19 @@ termination_measure pt_walk(_,_,_,_,_,_,_,level,_, _) = level
 // ****************************************************************
 // Architectural SATP CSR
 
+private function satp_accessible(priv : Privilege) -> bool =
+  match priv {
+    Machine           => true,
```
radimkrcmar: `satp` doesn't have to exist without S.

model/sys/vmem.sail
```diff
@@ -186,9 +186,19 @@ termination_measure pt_walk(_,_,_,_,_,_,_,level,_, _) = level
...
+    Supervisor        => currentlyEnabled(Ext_S) & mstatus[TVM]  == 0b0,
+    VirtualSupervisor => currentlyEnabled(Ext_H) /* TODO: & hstatus[VTVM] == 0b0 */,
```
radimkrcmar: Can't we assume that S is enabled when we are in Supervisor mode (same for VS-mode and H)?

(I think that the failure to meet the extension dependency should be omitted or expressed as an internal error.)

radimkrcmar (changes requested):
The exceptions for the F and V CSRs are missing.  (They do not throw virtual exceptions even when HS-qualified.)

model/extensions/K/zkr_control.sail
```diff
@@ -43,15 +43,16 @@ private function write_seed_csr () -> xlenbits = zeros()
...
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualUser => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    Machine           => true,
+    User              => mseccfg[USEED] == 0b1,
+    Supervisor        => mseccfg[SSEED] == 0b1,
+    VirtualSupervisor => false,
+    VirtualUser       => false,
   })
 
+function clause is_CSR_exception_virtual(0x015, _, access_type) = mseccfg[SSEED] == 0b1 & access_type == CSRReadWrite
```
radimkrcmar: No need for the special case, the seed CSR is following HS-qualified rule as well.

model/sys/vmem.sail
nadime15: Yeah sure, I can remove that, it’s just extra verbosity.

model/extensions/K/zkr_control.sail
nadime15: Ok that was first non obvious to me, but you are right. Thats a nice simplification!

model/sys/vmem.sail
nadime15: How can you access a S-mode register in U-mode?

model/extensions/K/zkr_control.sail
radimkrcmar: This is not obvious because we have a bug in `is_CSR_accessible`: the unprivileged volume is only talking about throwing an exception on read-only accesses, yet the privileged volume also throws illegal exception on write-only accesses from non-M modes.

To conform to the manual, we should take the privilege into account when checking access_type.
(I don't think this behavior was intended...)

model/sys/vmem.sail
radimkrcmar: You can express all CSRs in accesses. The access will trigger an exception, yet that exception is virtual instruction when in VU mode.

There is a even a clarification about that:
* in VU-mode, attempts to access an implemented non-high-half supervisor CSR when the same access (read/write) would be allowed in HS-mode, assuming mstatus.TVM=0;

The purpose of HS-qualified rule is to not bother M-mode with illegal instruction exceptions that would be just forwarded to HS-mode.

nadime15: Ahh yeah, it’s a bit confusing. I’m not sure what the best way to solve this is, maybe I need to rewrite it:

```
function clause is_CSR_exception_virtual(0x180, p, _) =  ( p == VirtualUser | (p == VirtualSupervisor & hstatus[VTVM] == 0b1)
```

and at the same time

change check_CSR back to:

```
function check_CSR(csr : csreg, p : Privilege, access_type : CSRAccessType) -> CSRCheckResult = {
  if check_CSR_priv(csr, p) & check_CSR_access(csr, access_type)
     & is_CSR_accessible(csr, p, access_type)
     & stateen_allows_CSR_access(csr, p, access_type) then
    CSR_Check_OK()
  else if (p == VirtualSupervisor | p == VirtualUser)
        & is_CSR_exception_virtual(csr, p, access_type) then
    CSR_Virtual()
  else
    CSR_Illegal()
}
```

This also would handle the stateen case you described above?

model/extensions/Sscofpmf/sscofpmf.sail
nadime15: Yes correct, I guess I can simply this to: 
```
function clause is_CSR_accessible(0xDA0, p, _) = currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
```

model/sys/vmem.sail
radimkrcmar: The changes to `check_CSR` look correct, and `satp` can be simplified to:
```
function clause is_CSR_exception_virtual(0x180, _, _) = true
```

nadime15:
Thanks @radimkrcmar for the feedback! I think I addressed all of your comments, let me know if something is missing.

model/extensions/Sscofpmf/sscofpmf.sail
```diff
@@ -71,13 +71,18 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
...
     User              => internal_error(__FILE__, __LINE__, "scountovf not readable from User mode"),
-    VirtualUser       => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
-    VirtualSupervisor => internal_error(__FILE__, __LINE__, "Hypervisor extension not supported"),
+    VirtualUser       => internal_error(__FILE__, __LINE__, "scountovf not readable from VirtualUser mode"),
   }
 }
 
 // scountovf
-function clause is_CSR_accessible(0xDA0, _, _) = currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
+function clause is_CSR_accessible(0xDA0, p, _) = currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
```
radimkrcmar:
```suggestion
function clause is_CSR_accessible(0xDA0, _, _) = currentlyEnabled(Ext_Sscofpmf) & currentlyEnabled(Ext_S)
```

model/extensions/Sscofpmf/sscofpmf.sail
```diff
@@ -71,13 +71,18 @@ private function get_scountovf(priv : Privilege) -> bits(32) = {
...
+
+// TODO: Smcdeleg is not yet implemented (including menvcfg[CDE]).
+// Once implemented, the is_CSR_exception_virtual clause will
+// correctly raise virtual instruction exceptions when menvcfg.CDE=1.
+// function clause is_CSR_exception_virtual(0xDA0, p, _) = currentlyEnabled(Ext_Smcdeleg) & menvcfg[CDE] == 0b1 & (p == VirtualSupervisor | p == VirtualUser)
```
radimkrcmar:
```suggestion
// Once implemented, the default is_CSR_exception_virtual clause will
// correctly raise virtual instruction exceptions when menvcfg.CDE=1.
```
We'll make `is_CSR_accessible` return false in VS/VU modes, and the rest is covered by the HS-qualified rule.

radimkrcmar (approved):
LGTM.

model/sys/vmem.sail
```diff
@@ -186,9 +186,19 @@ termination_measure pt_walk(_,_,_,_,_,_,_,level,_, _) = level
...
+    VirtualSupervisor => false /* TODO: hstatus[VTVM] == 0b0 */,
+    User              => false,
+    VirtualUser       => false,
+  }
+
 register satp : xlenbits
 mapping clause csr_name_map = 0x180  <-> "satp"
-function clause is_CSR_accessible(0x180, priv, _) = currentlyEnabled(Ext_S) & not(priv == Supervisor & mstatus[TVM] == 0b1)
+function clause is_CSR_accessible(0x180, priv, _) = satp_accessible(priv)
+function clause is_CSR_exception_virtual(0x180, _, _) = true
```
Timmmm, line 226: I think this should be `= mstatus[TVM] == 0b0`? Or something like that.

model/postlude/csr_end.sail
```diff
@@ -23,3 +23,10 @@ function clause write_CSR(csr, _) = {
    internal_error(__FILE__, __LINE__, "Write to CSR that does not exist: " ^ bits_str(csr));
 }
 end write_CSR
+
+// Implements the general HS-qualified rule from the H extension spec.
+// If an access fails in VS/VU mode but would succeed in HS mode, raise a
+// virtual instruction exception instead of an illegal instruction exception.
+// CSRs that deviate from this rule have explicit clauses defined.
+function clause is_CSR_exception_virtual(csr, _, access_type) = check_CSR(csr, Supervisor, access_type) == CSR_Check_OK()
```
Timmmm: The spec has a load of talk about high-half and low-half CSRs. It's not super clear but I presume this is because you might have different XLEN in each mode? I guess we don't support that currently so maybe it doesn't matter. Do we need to think about this?

Timmmm (commented):
I definitely prefer this, assuming it's correct. Sorry @pmundkur! 😬

model/sys/vmem.sail
nadime15: We know that accesses from VU always raise a virtual instruction exception. If an access from VS fails, then hstatus[VTVM] must be 1, otherwise the access would have succeeded.

Therefore, the more verbose check

`is_CSR_exception_virtual(0x180, _, _) = (p == VirtualUser | (p == VirtualSupervisor & hstatus[VTVM] == 0b1))`

can be simplified. Since failure in VS already implies hstatus[VTVM] == 1, the explicit check is redundant.

This simplifies to:

`function clause is_CSR_exception_virtual(0x180, _, _) = true`

model/postlude/csr_end.sail
nadime15: >  It's not super clear but I presume this is because you might have different XLEN in each mode? I guess we don't support that currently so maybe it doesn't matter.

Yeah this is about mixed XLEN where HS mode is RV64 but VS is RV32. We dont support that yet so we dont need to worry about (at least for now).

model/extensions/cfi/zicfiss_regs.sail
```diff
@@ -30,30 +30,22 @@ function zicfiss_xSSE(priv : Privilege) -> bool =
   })
 
 // See "Shadow Stack Pointer (ssp) CSR access control".
-function is_ssp_accessible(priv : Privilege) -> bool =
-  match priv {
+private function ssp_accessible(priv : Privilege) -> bool =
```
pmundkur: Why change the name? `is_ssp_accessible` is a better name for a function returning a boolean, just like the `is_CSR_exception_virtual` added  in this PR.
```suggestion
private function is_ssp_accessible(priv : Privilege) -> bool =
```

model/extensions/cfi/zicfiss_regs.sail
```diff
@@ -30,30 +30,22 @@ function zicfiss_xSSE(priv : Privilege) -> bool =
...
+  })
 
 register ssp : xlenbits = zeros()
 
 // "There is no high CSR defined as the ssp is always as wide as the
 // XLEN of the current privilege mode."
 mapping clause csr_name_map = 0x011  <-> "ssp"
-// TODO: handle throwing VirtualInstruction when required by `is_ssp_accessible()`.
-function clause is_CSR_accessible(0x011, _, _) = currentlyEnabled(Ext_Zicfiss) & is_ssp_accessible(cur_privilege)
+function clause is_CSR_accessible(0x011, p, _) = ssp_accessible(p)
```
pmundkur:
```suggestion
function clause is_CSR_accessible(0x011, p, _) = is_ssp_accessible(p)
```

model/extensions/Sscofpmf/sscofpmf.sail
pmundkur: Should we leave a comment here like below?
```
// TODO: when adding Smcdeleg/Sscfg:
//function clause is_CSR_exception_virtual(0xDA0, _, _) = currentlyEnabled(Ext_Smcdeleg) & currentlyEnabled(Ext_Sscfg) & currentlyEnabled(Ext_Sscofpmf) & menvcfg[CDE] == 0b1
```

model/extensions/Sstc/sstc.sail
```diff
@@ -5,24 +5,26 @@
...
 mapping clause csr_name_map = 0x14D  <-> "stimecmp"
 mapping clause csr_name_map = 0x15D  <-> "stimecmph"
 
-function sstc_CSRs_accessible(priv : Privilege) -> bool =
-    priv == Machine | (priv == Supervisor & mcounteren[TM] == 0b1 & menvcfg[STCE] == 0b1)
-
-function clause is_CSR_accessible(0x14D, priv, _) =
-    currentlyEnabled(Ext_S) & currentlyEnabled(Ext_Sstc) &
-    sstc_CSRs_accessible(priv)
+private function stimecmp_accessible(priv : Privilege) -> bool =
```
pmundkur:
```suggestion
private function is_stimecmp_accessible(priv : Privilege) -> bool =
```

model/extensions/Sstc/sstc.sail
```diff
@@ -5,24 +5,26 @@
...
+    VirtualSupervisor => mcounteren[TM] == 0b1 & menvcfg[STCE] == 0b1,
+                         /* TODO: & hcounteren[TM] == 0b1 & henvcfg[STCE] == 0b1 */
+    VirtualUser       => false,
+  })
 
-function clause is_CSR_accessible(0x15D, priv, _) =
-    currentlyEnabled(Ext_S) & currentlyEnabled(Ext_Sstc) & xlen == 32 &
-    sstc_CSRs_accessible(priv)
+function clause is_CSR_accessible(0x14D, priv, _) = stimecmp_accessible(priv)
+function clause is_CSR_accessible(0x15D, priv, _) = stimecmp_accessible(priv) & xlen == 32
```
pmundkur:
```suggestion
function clause is_CSR_accessible(0x14D, priv, _) = is_stimecmp_accessible(priv)
function clause is_CSR_accessible(0x15D, priv, _) = is_stimecmp_accessible(priv) & xlen == 32
```

pmundkur (changes requested):
AFAICT, there should be special handling for `srmcfg`:
> If mstateen0.SRMCFG is 1 or if extension Smstateen is not implemented, attempts to access srmcfg when V=1 raise a virtual-instruction exception.

I don't think this correctly handles the "or if extension Smstateen is not implemented" case: it will allow access instead of generating a virtual-instruction exception.  It does seem an odd exception though; perhaps I'm reading the spec or the code wrong?

nadime15:
> AFAICT, there should be special handling for srmcfg:
> 
>     If mstateen0.SRMCFG is 1 or if extension Smstateen is not implemented, attempts to access srmcfg when V=1 raise a virtual-instruction exception.
> 
> I don't think this correctly handles the "or if extension Smstateen is not implemented" case: it will allow access instead of generating a virtual-instruction exception. It does seem an odd exception though; perhaps I'm reading the spec or the code wrong?

No special handling needed, I modified the `is_CSR_accessible` clause and prohibited access for VS and VU (they always raise an exception).

This should cover all 3 scenarios with the default HS-qualifier rule:

* if Smstateen not implemented -> stateen_allows_CSR_access returns true -> Virtual Instruction Exception
* if mstateen0.SRMCFG=1 -> stateen_allows_CSR_access returns true -> Virtual Instruction Exception  
* if mstateen0.SRMCFG=0 -> stateen_allows_CSR_access returns false -> Illegal Instruction Exception

model/extensions/Sscofpmf/sscofpmf.sail
nadime15: No, its already handled implicitly by the HS-qualifier rule. Like the other cases that were removed, it follows the same pattern, if the CSR is not accessible, it implies that menvcfg[CDE] == 0b1 (otherwise it would have been accessible). Therefore, we dont need to check this explicitly.

pmundkur (commented):
LGTM, modulo the termination measures for the recursive calls (`check_CSR` -> `is_CSR_exception_virtual` ->  `check_CSR`) needed to get the Lean and Rocq builds to work.

The gain in succinctness is being paid for by the recursive call of `check_CSR` re-doing some of the checks (e.g. xlen and extension enabled checks) all over again, but that may be okay.

nadime15:
@bacam If you have the time, could you take a look at the failing Lean [job](https://github.com/riscv/sail-riscv/actions/runs/23801054413/job/69361299756?pr=1624)? It's not really obvious to me why it fails or what I should do to address it.

bacam:
> @bacam If you have the time, could you take a look at the failing Lean [job](https://github.com/riscv/sail-riscv/actions/runs/23801054413/job/69361299756?pr=1624)? It's not really obvious to me why it fails or what I should do to address it.

The error that's there is due to a bug in the Lean backend, which should be easy to fix.  However, manually sorting that out reveals the Lean doesn't like the termination measure for `check_CSR`.  It is possible to make it happy by changing the `if` expressions testing `p` to pattern matches.  However, it appears to me that you could factor out the main test in `check_CSR` into its own function and avoid the recursion entirely.  If that works I think it would be a nicer solution.

pmundkur (approved):
LGTM.  Thanks for trying this approach!

Timmmm (approved):
Sorry for the delay. My github inbox... :-D

LGTM!

PR #1630, opened 2026-03-26
Add the Ziccif extension.

pmundkur:
This updates the fetcher to handle atomic fetches for NAPOT PCs with ILEN=32.  This implementation will have to be updated when ILEN is increased to 48 and above to handle discarded bytes that straddle memory protection boundaries.

This removes an obsolete comment about fetch extension checks returning a transformed PC: they currently return an optional error.

### DIFF
--- a/doc/ReadingGuide.md
+++ b/doc/ReadingGuide.md
@@ -255,6 +255,24 @@ The `unit_tests` module collects Sail unit tests for the specification.
 The `main` module provides a [`main()`](../model/main/main.sail)
 function that is used in other Sail backends.
 
+### Miscellaneous notes
+
+When implementing atomic accesses (e.g. for AMOs and instructions
+fetches in Ziccif), there are two approaches that can be taken in the
+Sail model. One approach is to implement an atomic memory access
+using a single memory transaction at the lowest level (i.e. in
+[phys_mem_interface.sail](../model/core/phys_mem_interface.sail); this
+is the approach taken in this model. The second approach splits an
+atomic access into multiple smaller intra-instruction memory
+transactions at the level of
+[phys_mem_interface.sail](../model/core/phys_mem_interface.sail);
+however, these would have to be accompanied by additional rules in the
+RVWMO memory model specifying that no other memory accesses can
+intervene in between these transactions. Both approaches are
+observationally equivalent for the C++ emulator since it currently
+implements a single-hart system, but the first approach is preferred
+to simplify the rules in the memory model.
+
 ## Structure of the C++ emulator
 
 The diagram below illustrates how the C++ emulator is built from the
--- a/doc/ReadingGuide.md
+++ b/doc/ReadingGuide.md
@@ -260,15 +260,15 @@ function that is used in other Sail backends.
 When implementing atomic accesses (e.g. for AMOs and instructions
 fetches in Ziccif), there are two approaches that can be taken in the
 Sail model. One approach is to implement an atomic memory access
-using a single memory transaction at the lowest level (i.e. in
+using a single memory operation at the lowest level (i.e. in
 [phys_mem_interface.sail](../model/core/phys_mem_interface.sail); this
 is the approach taken in this model. The second approach splits an
 atomic access into multiple smaller intra-instruction memory
-transactions at the level of
+operations at the level of
 [phys_mem_interface.sail](../model/core/phys_mem_interface.sail);
 however, these would have to be accompanied by additional rules in the
 RVWMO memory model specifying that no other memory accesses can
-intervene in between these transactions. Both approaches are
+intervene in between these operations. Both approaches are
 observationally equivalent for the C++ emulator since it currently
 implements a single-hart system, but the first approach is preferred
 to simplify the rules in the memory model.
--- a/model/core/extensions.sail
+++ b/model/core/extensions.sail
@@ -250,15 +250,20 @@ function clause hartSupports(Ext_Zbs) = config extensions.Zbs.supported
 
 // Main memory supports all atomics in Zaamo
 enum clause extension = Ext_Ziccamoa
-mapping clause extensionName= Ext_Ziccamoa <-> "Ziccamoa"
+mapping clause extensionName= Ext_Ziccamoa <-> "ziccamoa"
 function clause hartSupports(Ext_Ziccamoa) = config extensions.Ziccamoa.supported
 // Main memory supports atomics in Zacas
 enum clause extension = Ext_Ziccamoc
-mapping clause extensionName= Ext_Ziccamoc <-> "Ziccamoc"
+mapping clause extensionName= Ext_Ziccamoc <-> "ziccamoc"
 function clause hartSupports(Ext_Ziccamoc) = config extensions.Ziccamoc.supported
+// Instruction-Fetch Atomicity
+enum clause extension = Ext_Ziccif
+mapping clause extensionName = Ext_Ziccif <-> "ziccif"
+function clause hartSupports(Ext_Ziccif) = config extensions.Ziccif.supported
+function clause currentlyEnabled(Ext_Ziccif) = hartSupports(Ext_Ziccif)
 // Main memory supports forward progress on LR/SC sequences
 enum clause extension = Ext_Ziccrse
-mapping clause extensionName= Ext_Ziccrse <-> "Ziccrse"
+mapping clause extensionName= Ext_Ziccrse <-> "ziccrse"
 function clause hartSupports(Ext_Ziccrse) = config extensions.Ziccrse.supported
 
 // Scalar & Entropy Source Instructions: NIST Suite: AES Decryption
@@ -654,6 +659,7 @@ let extensions_ordered_for_isa_string = [
   Ext_Zicboz,
   Ext_Ziccamoa,
   Ext_Ziccamoc,
+  Ext_Ziccif,
   Ext_Ziccrse,
   Ext_Zicfilp,
   Ext_Zicfiss,
--- a/model/postlude/fetch.sail
+++ b/model/postlude/fetch.sail
@@ -11,14 +11,47 @@
 
 private function isRVC(h : half) -> bool = not(h[1 .. 0] == 0b11)
 
+// This instruction byte fetcher is used for partial fetches, and
+// hence doesn't return the FetchResult type that a full fetcher does.
+// All the bytes fetched go through memory protection and PMA checks;
+// this just barely works for Ziccif with ILEN=32 (bits) since there
+// are no discarded bytes that lie outside protection boundaries.
+// This will have to be generalized when ILEN is increased above 32
+// (e.g. to 48): there could be discarded bytes that straddle
+// protection boundaries in that case, and memory protection and PMA
+// checks do not apply to discarded bytes.
+
+private union FetchBytes_Result('width : Int), 'width >= 0 = {
+  FetchBytes_Ext_Error : ext_fetch_addr_error,
+  FetchBytes_Exception : ExceptionType,
+  FetchBytes_Success   : bits('width * 8),
+}
+
+private function fetch_bytes forall 'n, 'n in {2, 4} . (fetch_start : xlenbits, granule_start : xlenbits, width : int('n)) -> FetchBytes_Result('n) = {
+  match ext_fetch_check_pc(fetch_start, granule_start) {
+    Some(e) => return FetchBytes_Ext_Error(e),
+    None()  => (),
+  };
+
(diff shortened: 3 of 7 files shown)


model/postlude/fetch.sail
```diff
@@ -11,14 +11,47 @@
...
+  let paddr : physaddr = match translateAddr(Virtaddr(granule_start), InstructionFetch()) {
+    Err(e, _)    => return FetchBytes_Exception(e),
+    Ok(paddr, _) => paddr,
+  };
+
+  match mem_read(InstructionFetch(), paddr, width, false, false, false) {
+    Err(e)    => FetchBytes_Exception(e),
+    Ok(bytes) => FetchBytes_Success(bytes),
+  }
+}
```
nadime15, line 45: Given the discussion in #1614, it might make more sense (assuming we go with the suggested approach or similar) to not bundle address translation and memory access in a single function, and instead split them up:

```
function translate_insn(vaddr : xlenbits) -> result(physaddr, FetchResult) =
  match translateAddr(Virtaddr(vaddr), InstructionFetch()) {
    Err(e, _) => Err(F_Error(e, vaddr)),
    Ok(paddr, _) => Ok(paddr),
  }
...
```

But this is up to you, if you prefer to merge Ziccif first and then (potentially) refactor this part of the code afterwards once we found an agreement, that works for me as well.

pmundkur: I'd prefer to merge this first.  We're going to need a general approach for misaligned access, including fetch, and once that is resolved, other existing code is going to need similar refactoring anyway.  Merging this will help us do the refactor keeping in mind how it would apply for Ziccif.

model/core/extensions.sail
```diff
@@ -256,6 +256,11 @@ function clause hartSupports(Ext_Ziccamoa) = config extensions.Ziccamoa.supporte
 enum clause extension = Ext_Ziccamoc
 mapping clause extensionName= Ext_Ziccamoc <-> "Ziccamoc"
 function clause hartSupports(Ext_Ziccamoc) = config extensions.Ziccamoc.supported
+// Instruction-Fetch Atomicity
+enum clause extension = Ext_Ziccif
+mapping clause extensionName = Ext_Ziccif <-> "ziccif"
```
nadime15, line 261:
```suggestion
mapping clause extensionName = Ext_Ziccif <-> "Ziccif"
```

model/postlude/fetch.sail
```diff
@@ -27,35 +60,28 @@ private function fetch() -> FetchResult = {
...
+      FetchBytes_Ext_Error(e) => F_Ext_Error(e),
+      FetchBytes_Exception(e) => F_Error(e, PC),
+      FetchBytes_Success(ilo) =>
+        if   isRVC(ilo)
+        then F_RVC(ilo)
+        else match fetch_bytes(PC, PC + 2, 2) {
+          FetchBytes_Ext_Error(e) => F_Ext_Error(e),
+          FetchBytes_Exception(e) => F_Error(e, PC + 2),
+          FetchBytes_Success(ihi) => F_Base(append(ihi, ilo)),
+        },
```
nadime15, line 103: The spec says: 

> An implementation with the Ziccif extension fetches instructions in a manner **_equivalent_** to the following state machine.

If I understand correctly, you are trying to model the code strictly according to the spec, but I am not sure we actually need to do so as long as the observable behavior remains the same. I think, removing the if branch would still be Ziccif compliant.

I also think the if branch could trigger an exception if a 32 bit read crosses a page boundary and the actual instruction turns out to be RVC. In theory the upper bits should be discarded which would still be valid. If we keep the suggestion (down below), it should work fine, we just keep fetching 2 bytes at the time, make it easier to extend to ILEN=48, and would simplify integration of misaligned instruction fetches that cross PMP boundaries?

I am not 100% sure, I am curious to know what others think.

```suggestion
match fetch_bytes(PC, PC, 2) {
  FetchBytes_Ext_Error(e) => F_Ext_Error(e),
  FetchBytes_Exception(e) => F_Error(e, PC),
  FetchBytes_Success(ilo) =>
    if isRVC(ilo)
    then F_RVC(ilo)
    else match fetch_bytes(PC, PC + 2, 2) {
      FetchBytes_Ext_Error(e) => F_Ext_Error(e),
      FetchBytes_Exception(e) => F_Error(e, PC + 2),
      FetchBytes_Success(ihi) => F_Base(append(ihi, ilo)),
    },
}
```

model/postlude/fetch.sail
```diff
@@ -11,14 +11,47 @@
...
 private function fetch() -> FetchResult = {
-  // fetch PC check for extensions: extensions return a transformed PC to fetch,
-  // but any exceptions use the untransformed PC.
   // TODO: Add a parameter to try_step() to allow forcing the fetch result and use that instead.
 
   if get_config_rvfi()
   then return rvfi_fetch();
 
+  // Doing this check here assumes the external fetch checks are higher priority
+  // then the following alignment check.
```
nadime15:
```suggestion
  // than the following alignment check.
```

model/postlude/fetch.sail
nadime15: Ups, I think I was wrong, its an aligned addr! But I still think it could turn out to be easier if we always use 2-byte granules.

model/core/extensions.sail
pmundkur: It looks like lower case is actually consistent with the others, it is `ziccrse` and `ziccamo{ac}` that are incorrectly uppercase.

model/postlude/fetch.sail
pmundkur: > But I still think it could turn out to be easier if we always use 2-byte granules.

That's what the earlier fetcher was doing.  The Ziccif spec requires

> any instruction fetches of naturally aligned power-of-2 sizes of at most min(ILEN,XLEN) bits are atomic.

And since the unprivileged spec defines ILEN as 

> We use the term ILEN (measured in bits) to refer to the maximum instruction length supported by an implementation

for now ILEN is 32.  This means naturally aligned 32-bit fetches need to be atomic, and cannot be split.

nadime15: > for now ILEN is 32. This means naturally aligned 32-bit fetches need to be atomic, and cannot be split.

I dont think this is what the spec says. I would say its more that the fetch must appear atomic to the rest of the system. The spec says that a 32-bit access must be atomic, but it does not say how this atomicity is implemented (observable behavior vs microarchitectural implementation). 

In principle, a 32-bit access could be broken down into four byte reads, as long as the overall operation is atomic. If we were dealing with a concurrent multi-core system this would need more consideration, since intermediate states could become observable.

pmundkur: Since this is an architectural model, when the ISA says the access should be atomic, the model should implement an atomic access.   I'm not sure what the value would be of doing otherwise?  It would also be problematic from the point of view of integrating the Sail model into the prose specification.

nadime15: I dont understand why you say, “when the ISA says the access should be atomic, the model should implement an atomic access.” Why is splitting a 32-bit access into two 16-bit atomic accesses not atomic in this case? As long as theres no concurrency or the series of accesses cannot be interrupted, it seems atomic to me. 

The only issue I could see is if we accessed memory mapped registers that trigger some kind of action on a partial read (but I guess MMIO regions are usually not executable). 

Can you give an example where performing two 16-bit atomic accesses (to model a 32-bit atomic access), would fail to be atomic and why the example above is not atomic?

>  I'm not sure what the value would be of doing otherwise?

The PR as it is, looks good to me, I am just wondering whether we could simplify the implementation, generalize it and still model the correct behavior at the same time.

pmundkur: The Sail model includes an implementation of the RVWMO specification, which is used for e.g. in https://github.com/rems-project/isla.  See `core/phys_mem_interface.sail` for the relevant memory interface.  It is important that this physical memory interface accurately capture memory accesses, including the various kinds of atomic accesses.  Splitting an ISA-specified atomic access would break this model.

nadime15: It seems that instruction fetches simply issue a Read_plain type of operation. Looking at the Isla code, this doesn’t appear to have any significant semantics (?) unlike operations such as Read_RISCV_reserved_acquire or similar.

Interestingly, we dont use the Ak_ifetch() variant, which, according to this Isla [configuration file](https://github.com/rems-project/isla/blob/7f6882b3f468fe34df38ae3ffc0aede5baa0e7d6/configs/riscv32.toml#L7), has no defined semantics for RISC-V.

Besides that Chapter 17.1. RVWMO Memory Consistency Model, Version 2.0 says:
> The interaction of the memory model with I/O memory, **instruction fetches**, FENCE.I, page-table walks, and SFENCE.VMA is not (yet) formalized.

Not really my domain, but I think this wouldn break anything.

pmundkur: It's been done for ARM by the Isla folks; I hope it gets done for RISC-V too.
https://link.springer.com/chapter/10.1007/978-3-030-44914-8_23
I don't think we should pre-emptively break the model just because it's not yet done for RISC-V yet.  Let's try to keep things atomic when the spec says it should be atomic.

pmundkur:
@Timmmm @jordancarlin could you review this?  I've added the comments discussed in the Monday meeting.

model/postlude/fetch.sail
```diff
@@ -27,35 +60,40 @@ private function fetch() -> FetchResult = {
...
+  // intra-instruction fetches.  The former approach is currently simpler
+  // since it does not require any additional rules in the memory model.
+  // However, if/when ILEN is increased to 48 and higher, it may be that
+  // the complexities around memory protection could be addressed using
+  // smaller fetches and augmenting the memory model with these additional
+  // rules for intra-instruction split fetches.
+  if is_aligned_addr(Virtaddr(PC), 4) & currentlyEnabled(Ext_Ziccif) then {
+    match fetch_bytes(PC, PC, 4) {
+      FetchBytes_Ext_Error(e)   => F_Ext_Error(e),
+      FetchBytes_Exception(e)   => F_Error(e, PC),
```
Timmmm, line 86: I think it's worth a comment here something like:

```
// If an exception is raised for the upper two bytes but the instruction is only 16 bits
// then the exception should be ignored. In this implementation we avoid situations
// where this can happen - PMP's minimum granularity is 4 bytes, and for simplicity
// we restrict PMA granularity to 4096 bytes. This situation may still occur of you
// configure MMIO devices incorrectly (this is not checked currently).
```

model/postlude/fetch.sail
```diff
@@ -27,35 +60,40 @@ private function fetch() -> FetchResult = {
...
-              Err(e, _) => F_Error(e, PC_hi),
-              Ok(ppchi, pbmt, _) =>
-                match mem_read(InstructionFetch(), pbmt, ppchi, 2, false, false, false) {
-                  Err(e)  => F_Error(e, PC_hi),
-                  Ok(ihi) => F_Base(append(ihi, ilo))
-                }
-            }
-          }
-      }
+  // For a PC aligned at ILEN=32, perform an atomic fetch for Ziccif.
```
Timmmm, line 63: I think we probably need a comment explaining why we need this even though the current two-fetch implementation is atomic for the C++ emulator. I think the reason was it isn't atomic for the memory model with multiple harts?

pmundkur: I'm tempted to put this in the doc/ReadingGuide since this is not specific to instruction fetches; even AMOs could be implemented byte-wise (for e.g.) in the C++ emulator  and still be "atomic";  the atomicity is required for the multi-hart memory-model.

jordancarlin (approved):

Timmmm (approved):

doc/ReadingGuide.md
```diff
@@ -255,6 +255,24 @@ The `unit_tests` module collects Sail unit tests for the specification.
 The `main` module provides a [`main()`](../model/main/main.sail)
 function that is used in other Sail backends.
 
+### Miscellaneous notes
+
+When implementing atomic accesses (e.g. for AMOs and instructions
+fetches in Ziccif), there are two approaches that can be taken in the
+Sail model. One approach is to implement an atomic memory access
+using a single memory transaction at the lowest level (i.e. in
```
Timmmm:
```suggestion
using a single memory operation at the lowest level (i.e. in
```

^-- that's the official term btw.

doc/ReadingGuide.md
```diff
@@ -255,6 +255,24 @@ The `unit_tests` module collects Sail unit tests for the specification.
...
+[phys_mem_interface.sail](../model/core/phys_mem_interface.sail); this
+is the approach taken in this model. The second approach splits an
+atomic access into multiple smaller intra-instruction memory
+transactions at the level of
```
Timmmm:
```suggestion
operations at the level of
```

PR #1635, opened 2026-03-29
Use ExecuteAs to make execute non-recursive

Timmmm:
Currently execute() is recursive due to these remaining calls that we apparently missed. I verified that with these changes Sail now considers it to be non-recursive, and consequently it can hoist GMP allocations outside of the function. This should be a little faster, though I did not measure a significant speedup.

### DIFF
--- a/model/extensions/FD/zcd_insts.sail
+++ b/model/extensions/FD/zcd_insts.sail
@@ -16,7 +16,7 @@ mapping clause encdec_compressed = C_FLDSP(uimm, rd)
 
 function clause execute C_FLDSP(uimm, rd) = {
   let imm : bits(12) = zero_extend(uimm @ 0b000);
-  execute(LOAD_FP(imm, sp, rd, 8))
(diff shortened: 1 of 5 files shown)


Timmmm:
I'm not 100% sure about the sfence using `ExecuteAs`. It seems like if we allow much more of that it could get difficult to verify we don't have any double-indirections. Maybe we should just split that out into a function instead?

jordancarlin:
> I'm not 100% sure about the sfence using `ExecuteAs`. It seems like if we allow much more of that it could get difficult to verify we don't have any double-indirections. Maybe we should just split that out into a function instead?

Agreed. Especially because we might want the behavior to differ slightly down the road, which would be easiest with a function.

arichardson (approved):

pmundkur (approved):
LGTM.  We could do the `sfence` function separately since it's otherwise unrelated to the core change of this PR.

jordancarlin (approved):

PR #1641, opened 2026-03-31
Fix command path in README for simulator help

davidharrishmc:


nadime15 (approved):

README.md
```diff
@@ -117,7 +117,7 @@ used. This option allows one or more additional JSON configuration files to be s
 whose fields take precedence over those in the base configuration.
 
 Information on other options for the simulator is available from
-`build/c_emulator/sail_riscv_sim -h`.
```
pmundkur: To be consistent with the rest of the section, perhaps this sentence could be reworded as
```
Information on other options for the simulator is available using the `-h` option.
```
The path is also mentioned earlier in this section; that sentence could also be reworded similarly to remove mention of the path/executable.

Timmmm (approved):
I think it's probably fine as-is. Tbh we could probably just delete this sentence. Who of our users doesn't know about `-h`?

README.md
pmundkur: I removed the earlier path, and removed this line; it was redundant and didn't belong in the section on "Configuring platform options."

pmundkur (approved):

jordancarlin (approved):

Timmmm (approved):

PR #1642, opened 2026-03-31
Fix markdown link-check whitelist url.

pmundkur:
Append what seems to be a missed 'raw=true' specifier after the terminating '?' in the URL for `overview-sail.png` in the README; adjust the whitelist pattern accordingly.  Also re-enable link checks for `isla-axiomatic.cl.cam.ac.uk`.

### DIFF
--- a/.markdown-link-check.config
+++ b/.markdown-link-check.config
@@ -13,7 +13,7 @@
       "reason": "random access issues from github.com CI"
     },
     {
-      "pattern": "^http://www.cl.cam.ac.uk/~pes20/sail/overview-sail.png",
+      "pattern": "^http://www.cl.cam.ac.uk/~pes20/sail/overview-sail.png?",
       "reason": "random access issues from github.com CI"
     }
   ]
--- a/.markdown-link-check.config
+++ b/.markdown-link-check.config
@@ -1,10 +1,6 @@
 {
   "ignorePatterns": [
     {
-      "pattern": "^https://isla-axiomatic.cl.cam.ac.uk",
-      "reason": "expired certificate"
-    },
-    {
       "pattern": "^https://gmplib.org",
       "reason": "access issues from github.com"
     },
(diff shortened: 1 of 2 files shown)


jordancarlin (approved):

.markdown-link-check.config
```diff
@@ -13,7 +9,7 @@
       "reason": "random access issues from github.com CI"
     },
     {
-      "pattern": "^http://www.cl.cam.ac.uk/~pes20/sail/overview-sail.png",
+      "pattern": "^http://www.cl.cam.ac.uk/~pes20/sail/overview-sail.png?",
```
Timmmm: Is this a regex? Surely you meant this?

```suggestion
      "pattern": "^http://www.cl.cam.ac.uk/~pes20/sail/overview-sail.png$",
```

jordancarlin: It's a literal question mark at the end of the link...

pmundkur: Yes, it's indeed a regex, that's trying to match the "https://www.cl.cam.ac.uk/~pes20/sail/overview-sail.png?" in the README.  I just pushed a fix using  a '*' to match the trailing literal '?'.

Timmmm: Ah right. `*` is also wrong because that will match `gggg...`. You want `.png\\?"`

Timmmm: Or better, just get right of the `?` in the link!

pmundkur: Just for background, the motivation for this PR was this CI failure:
https://github.com/riscv/sail-riscv/actions/runs/23801739242/job/69363792474
where the link checker's `ignorePattern` did not work.   Since the failures are so intermittent, it's hard to test these reliably.

It looks like the "?" was specifically introduced in 112f1263d85eb7879e52700f9 , but I'm wondering if it was a typo, and the intention was to add "?raw=true".  The Sail repo also had a similar commit
https://github.com/rems-project/sail/commit/bc5b4f2597fc6f06c031d639fdc0585850e37959, but that was modified later to now have "?raw=true" (but using a github link for (almost) the same image).  Perhaps we could add "?raw=true" to the README link and use that full link also  in the `ignorePattern`?

Timmmm (approved):

PR #1646, opened 2026-04-01
Add a flag to the address translator to enable a pure lookup.

pmundkur:
A pure lookup enables the translation of virtual addresses to physical addresses without any side-effects (e.g. PTE writes, TLB updates).  This is sometimes needed to handle misalignment e.g. to lookup the MAG PMA for a virtual address, or check whether a misaligned atomic could complete except for the misalignment (for software emulation).  This can also be useful for Ziccif when ILEN is increased to 48 or higher.

To make the Sail code cleaner, the C++ PTW callbacks are passed a parameter to indicate whether they are being called for such a lookup.  This is currently used to suppress PTW trace logging in this case.

C++ callbacks for PTE reads during a lookup are still called. These should be bracketed by the PTW callbacks, so they could in theory be handled correctly.  But the logging callbacks currently don't handle this.

### DIFF
--- a/c_emulator/riscv_callbacks_if.cpp
+++ b/c_emulator/riscv_callbacks_if.cpp
@@ -102,7 +102,8 @@ void callbacks_if::ptw_start_callback(
   [[maybe_unused]] hart::Model &model,
   [[maybe_unused]] uint64_t vpn,
   [[maybe_unused]] hart::zMemoryAccessTypezIEmem_payloadz5zK access_type,
-  [[maybe_unused]] hart::ztuple_z8z5enumz0zzPrivilegezCz0z5unitz9 privilege
+  [[maybe_unused]] hart::ztuple_z8z5enumz0zzPrivilegezCz0z5unitz9 privilege,
+  [[maybe_unused]] bool is_pure_lookup
 ) {
 }
 
@@ -110,14 +111,16 @@ void callbacks_if::ptw_step_callback(
   [[maybe_unused]] hart::Model &model,
   [[maybe_unused]] int64_t level,
   [[maybe_unused]] sbits pte_addr,
-  [[maybe_unused]] uint64_t pte
+  [[maybe_unused]] uint64_t pte,
+  [[maybe_unused]] bool is_pure_lookup
 ) {
 }
 
 void callbacks_if::ptw_success_callback(
   [[maybe_unused]] hart::Model &model,
   [[maybe_unused]] uint64_t final_ppn,
-  [[maybe_unused]] int64_t level
+  [[maybe_unused]] int64_t level,
+  [[maybe_unused]] bool is_pure_lookup
 ) {
 }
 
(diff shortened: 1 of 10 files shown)


Timmmm:
> This is sometimes needed to handle misalignment e.g. to lookup the MAG PMA for a virtual address, or check whether a misaligned atomic could complete except for the misalignment (for software emulation).

Can you expand on this? For the MAG I would expect that we do a single address translation and PMA lookup for the address of AMO.

For example for a 2-byte AMO at address 0x3 we translate 0x3 and look up its PMA. If it says the MAG is 8 then we know that it can' proceed. If it says 4 then we know it should fail.

pmundkur:
> > This is sometimes needed to handle misalignment e.g. to lookup the MAG PMA for a virtual address, or check whether a misaligned atomic could complete except for the misalignment (for software emulation).
> 
> Can you expand on this? For the MAG I would expect that we do a single address translation and PMA lookup for the address of AMO.
> 
> For example for a 2-byte AMO at address 0x3 we translate 0x3 and look up its PMA. If it says the MAG is 8 then we know that it can' proceed. If it says 4 then we know it should fail.

I was wondering: when it fails, or the optional MAG PMA is not present, and we need to determine whether the misaligned atomic "could complete except for the misalignment", should the side-effects of that address translation be visible?

This also matters for misaligned loads/stores and the MAG in `vmem_utils`.  They currently use a global `allowed_within` to decide split misaligned accesses based on the virtual address, but the `allowed_within` should be coming from the MAG (if present) of the physical address.  If we restrict the maximum MAG granule to a page size,  we could look up the MAG (without any translation side-effects)  and use it for `allowed_within`.    (An alternative would be to move the whole splitting logic from `vmem_utils` into `mem` to work on physical addresses; but that would re-open the page-crossing issue.)

Timmmm:
> I was wondering: when it fails, or the optional MAG PMA is not present, and we need to determine whether the misaligned atomic "could complete except for the misalignment", should the side-effects of that address translation be visible?

Yeah I think side-effects are fine based on this bit of the spec:

> The PTE update is not required to be atomic with respect to the memory access that caused the update and a trap may occur between the PTE update and the memory access that caused the PTE update. If a trap occurs then the A and/or D bit may be updated but the memory access that caused the PTE update might not occur.

----

> If we restrict the maximum MAG granule to a page size, we could look up the MAG (without any translation side-effects) and use it for allowed_within.

Yeah I think that's what we should do. But we don't have to worry about side-effects, because of the above statement. We might want to just split atomic accesses into a separate function since we're never going to want to split them into multiple memory operations - it should make the code a bit simpler. We want something like:

```
if res then {
  let (paddr, pma) = translateAddr(vaddr);
  if paddr fits inside pma.mag then {
    // Do single memory operation
  } else {
    // access fault.
  }
} else {
  // existing code with the for loop.
}
```

We potentially also do want to keep the global MAG since it can be checked before address translation.

pmundkur:
I thought this would be needed for the MAG PMA, but don't anymore.

PR #1648, opened 2026-04-03
Add explicit main/IO memory PMA

pmundkur:
Add a memory type attribute to distinguish main memory and IO memory. The other attributes are checked for consistency with the memory type. It is also used for PBMT.

Also fix some confusing variable naming that conflated 'PMA' with 'PMA_Region'.

### DIFF
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -258,7 +258,7 @@ private function check_pma_regions(regions : list(PMA_Region), prev_base : bits(
         return false;
       };
       let attributes = region.attributes;
-      if pma_is_main_memory(attributes) & attributes.cacheable & attributes.coherent then {
+      if memory_region_type(attributes) == MainMemory & attributes.cacheable & attributes.coherent then {
         if check_opts.ziccamoa & attributes.atomic_support < AMOArithmetic then {
           print_endline("Memory region starting at " ^ bits_str(region.base) ^ " is coherent and cacheable with " ^ atomic_support_str(attributes.atomic_support) ^ " atomicity support, but Ziccamoa is enabled which requires AMOArithmetic support.");
           return false;
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -235,6 +235,21 @@ private function check_vext_config() -> bool = {
   valid;
 }
 
+private function check_pma_region(region : PMA_Region) -> bool = {
+  let pma = region.attributes;
+  // TODO: this could be extended, e.g. to check that read_idempotent => readable, etc.
+  match pma.mem_type {
+    MainMemory =>
+      if not(pma.readable & pma.writable & pma.read_idempotent & pma.write_idempotent) then {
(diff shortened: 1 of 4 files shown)


model/postlude/validate_config.sail
```diff
@@ -226,6 +226,9 @@ private function check_vext_config() -> bool = {
   valid;
 }
 
+private function pma_is_main_memory(attributes : PMA) -> bool =
+  attributes.readable & attributes.read_idempotent & attributes.writable & attributes.write_idempotent
```
jordancarlin: Add a TODO for memory ordering.

jordancarlin (approved):

model/sys/pma.sail
```diff
@@ -114,6 +114,10 @@ struct PMA = {
   // TODO: Memory Ordering PMA
 }
 
+// TODO: Add Memory Ordering
+function pma_is_main_memory(attributes : PMA) -> bool =
+  attributes.readable & attributes.read_idempotent & attributes.writable & attributes.write_idempotent
```
Timmmm: I think maybe we should import the code from #1468 

```
enum MemoryRegionType = { MainMemory, IOMemory }

// Technically you can have a region that satisfies these properties
// but is still declared as IOMemory but we don't support that yet.
// TODO: We should add an explicit memory type PMA rather than inferring this.
// TODO: Consider memory ordering if that PMA is added.
function memory_region_type(pma : PMA) -> MemoryRegionType =
  if pma.readable & pma.writable & pma.read_idempotent & pma.write_idempotent
  then MainMemory
  else IOMemory
```

model/postlude/validate_config.sail
```diff
@@ -244,48 +259,51 @@ private struct pma_check_opts = {
...
-          print_endline("Memory region starting at " ^ bits_str(pma.base) ^ " is coherent and cacheable with " ^ reservability_str(attributes.reservability) ^ " reservability support, but Ziccrse is enabled which requires RsrvEventual support.");
+          print_endline("Memory region starting at " ^ bits_str(region.base) ^ " is coherent and cacheable with " ^ reservability_str(attributes.reservability) ^ " reservability support, but Ziccrse is enabled which requires RsrvEventual support.");
           return false;
         };
         if check_opts.ssccptr & not(attributes.supports_pte_read) then {
-          print_endline("Memory region starting at " ^ bits_str(pma.base) ^ " is coherent and cacheable without hardware page-table read support, but Ssccptr is enabled which requires this support.");
+          print_endline("Memory region starting at " ^ bits_str(region.base) ^ " is coherent and cacheable without hardware page-table read support, but Ssccptr is enabled which requires this support.");
           return false;
         };
       };
```
Timmmm, line 296: Minor nit: I don't understand why some of these checks are in `check_pma_region` and others are outside it?

Otherwise this LGTM!

Timmmm (approved):

model/postlude/validate_config.sail
pmundkur: I was hoping to have `check_pma_region` eventually check for a consistent PMA without regard to any extensions that were being supported, i.e. have it essentially implement #1613.

Timmmm: Feels like a slightly arbitrary distinction to me but 🤷  :-)

PR #1649, opened 2026-04-06
`hpmcounter{3-31}` do not depend on U

jordancarlin:
The `hpmcounter{3-31}` CSRs are unprivileged CSRs that should be available from any priv mode (contigent on `mcounteren`). They do not require User mode to be implemented. This updates them to match the behavior of `cycle`/`time`/`instret`.

From the spec:
> The Zihpm extension comprises up to 29 additional unprivileged 64-bit hardware performance counters, hpmcounter3-hpmcounter31. When XLEN=32, the upper 32 bits of these performance counters are accessible via additional CSRs hpmcounter3h- hpmcounter31h. The Zihpm extension depends on the Zicsr extension.

Discovered using ACT4 tests.

### DIFF
--- a/model/extensions/Zihpm/zihpm.sail
+++ b/model/extensions/Zihpm/zihpm.sail
@@ -251,8 +251,8 @@ function clause write_CSR((0b1011100 /* 0xB80 */ @ index : bits(5), value) if xl
 }
 
 // Hardware Performance Monitoring user mode counters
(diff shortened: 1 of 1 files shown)


Timmmm (approved):
Ooo good find. I think that was my mistake 😬

pmundkur (approved):

PR #1651, opened 2026-04-06
Speed up PMP checks

nadime15:
Speed up PMP checks by caching effective addresses and skipping check when no entries active.

pmpReadAddrReg() was being called on every memory access to compute the effective address, despite PMP registers almost never changing at runtime. This adds two optimizations:

1. Precompute and cache the effective address for each PMP entry on every CSR write

2. Track the number of active entries and fast-exit pmpCheck when none are configured

Measured 7-9% overall KIPS improvement on a Linux boot workload.

### DIFF
--- a/model/pmp/pmp_control.sail
+++ b/model/pmp/pmp_control.sail
@@ -118,6 +118,11 @@ function pmpCheck forall 'n, 0 < 'n <= max_mem_access . (
 
   if sys_pmp_count == 0 then return None();
 
+  if pmp_active_count == 0 then return (
+    if priv == Machine then None()
+    else Some(accessFaultFromAccessType(access))
+  );
+
   let width : xlenbits = to_bits(width);
 
   // TODO for accesses of type CacheAccess:
@@ -128,11 +133,12 @@ function pmpCheck forall 'n, 0 < 'n <= max_mem_access . (
   // block, and if write permission is granted by the PMP access
   // control bits, read permission shall also be granted."
 
+  var prev_pmpaddr : xlenbits = zeros();
   foreach (i from 0 to sys_pmp_count - 1) {
-    let prev_pmpaddr = if i > 0 then pmpReadAddrReg(i - 1) else zeros();
+    let cur_pmpaddr = pmpaddr_n_eff[i];
     let cfg = pmpcfg_n[i];
 
-    match pmpMatchAddr(addr, width, cfg, pmpReadAddrReg(i), prev_pmpaddr) {
+    match pmpMatchAddr(addr, width, cfg, cur_pmpaddr, prev_pmpaddr) {
       PMP_NoMatch      => (),
       PMP_PartialMatch => return Some(accessFaultFromAccessType(access)),
       PMP_Match        => return (
@@ -141,6 +147,7 @@ function pmpCheck forall 'n, 0 < 'n <= max_mem_access . (
         else Some(accessFaultFromAccessType(access))
       ),
     };
+    prev_pmpaddr = cur_pmpaddr
   };
   if priv == Machine then None() else Some(accessFaultFromAccessType(access))
 }
@@ -149,6 +156,8 @@ function reset_pmp() -> unit = {
   foreach (i from 0 to 63) {
     // On reset the PMP register's A and L bits are set to 0 unless the platform
     // mandates a different value.
-    pmpcfg_n[i] = [pmpcfg_n[i] with A = pmpAddrMatchType_encdec(OFF), L = 0b0];
+    pmpcfg_n[i]     = [pmpcfg_n[i] with A = pmpAddrMatchType_encdec(OFF), L = 0b0];
+    pmpaddr_n_eff[i] = zeros();
   };
+  pmp_active_count = 0
 }
--- a/model/pmp/pmp_control.sail
+++ b/model/pmp/pmp_control.sail
@@ -134,7 +134,7 @@ function pmpCheck forall 'n, 0 < 'n <= max_mem_access . (
   // control bits, read permission shall also be granted."
 
   var prev_pmpaddr : xlenbits = zeros();
-  foreach (i from 0 to sys_pmp_count - 1) {
+  foreach (i from 0 to sys_pmp_usable_count - 1) {
     let cur_pmpaddr = pmpaddr_n_eff[i];
     let cfg = pmpcfg_n[i];
 
@@ -156,7 +156,7 @@ function reset_pmp() -> unit = {
   foreach (i from 0 to 63) {
     // On reset the PMP register's A and L bits are set to 0 unless the platform
     // mandates a different value.
-    pmpcfg_n[i]     = [pmpcfg_n[i] with A = pmpAddrMatchType_encdec(OFF), L = 0b0];
+    pmpcfg_n[i] = [pmpcfg_n[i] with A = pmpAddrMatchType_encdec(OFF), L = 0b0];
     pmpaddr_n_eff[i] = zeros();
   };
   pmp_active_count = 0
--- a/model/pmp/pmp_regs.sail
+++ b/model/pmp/pmp_regs.sail
@@ -20,6 +20,9 @@ let sys_pmp_usable_count : range(0, 64) = config memory.pmp.usable_count
 // G=0 -> 4 bytes, G=10 -> 4096 bytes.
 let sys_pmp_grain : range(0, 63) = config memory.pmp.grain
 
+// Number of active PMP entries
+register pmp_active_count : nat
+
 // PMP configuration entries
 
 private enum PmpAddrMatchType = {OFF, TOR, NA4, NAPOT}
@@ -43,6 +46,8 @@ bitfield Pmpcfg_ent : bits(8) = {
 
 register pmpcfg_n : vector(64, Pmpcfg_ent)
 register pmpaddr_n : vector(64, xlenbits)
+// Cached effective PMP addresses
+register pmpaddr_n_eff : vector(64, xlenbits)
 
 // Packing and unpacking pmpcfg regs for xlen-width accesses
 
@@ -87,6 +92,17 @@ private function pmpReadAddrReg(n : range(0, 63)) -> xlenbits = {
 
     _ => addr,
   }
+}
+
+// Update cached PMP effective addresses and active entry count (after CSR writes)
+private function pmpUpdateCache() -> unit = {
+  pmp_active_count = 0;
+  foreach (i from 0 to sys_pmp_count - 1) {
+    pmpaddr_n_eff[i] = pmpReadAddrReg(i);
+    if pmpAddrMatchType_encdec(pmpcfg_n[i][A]) != OFF then {
+      pmp_active_count = pmp_active_count + 1;
+    }
+  };
 }
 
 // Helpers to handle locked entries
@@ -138,7 +154,7 @@ private function pmpWriteCfgReg(n : range(0, 15), v : xlenbits) -> unit = {
     foreach (i from 0 to 3) {
       let idx = n*4 + i;
       if idx < sys_pmp_usable_count then
-        pmpcfg_n[idx]  = pmpWriteCfg(pmpcfg_n[idx],  v[8*i+7 .. 8*i]);
+        pmpcfg_n[idx] = pmpWriteCfg(pmpcfg_n[idx], v[8*i+7 .. 8*i]);
     }
   }
   else {
@@ -146,9 +162,10 @@ private function pmpWriteCfgReg(n : range(0, 15), v : xlenbits) -> unit = {
     foreach (i from 0 to 7) {
       let idx = n*4 + i;
       if idx < sys_pmp_usable_count then
-        pmpcfg_n[idx]  = pmpWriteCfg(pmpcfg_n[idx],  v[8*i+7 .. 8*i]);
+        pmpcfg_n[idx] = pmpWriteCfg(pmpcfg_n[idx], v[8*i+7 .. 8*i]);
     }
-  }
+  };
+  pmpUpdateCache()
 }
 
 private function pmpWriteAddr(locked: bool, tor_locked: bool, reg: xlenbits, v: xlenbits) -> xlenbits =
@@ -164,6 +181,7 @@ private function pmpWriteAddrReg(n : range(0, 63), v : xlenbits) -> unit = {
       pmpaddr_n[n],
       v,
     );
+  pmpUpdateCache()
 }
 
 // PMP CSRs


model/pmp/pmp_control.sail
```diff
@@ -118,6 +118,11 @@ function pmpCheck forall 'n, 0 < 'n <= max_mem_access . (
 
   if sys_pmp_count == 0 then return None();
 
+  if pmp_active_count == 0 then return (
+    if priv == Machine then None()
+    else Some(accessFaultFromAccessType(access))
+  );
+
```
pmundkur, line 125: Couldn't this be achieved by just changing the upper bound of the `foreach` loop below to `pmp_active_count`?

model/pmp/pmp_regs.sail
```diff
@@ -43,6 +46,8 @@ bitfield Pmpcfg_ent : bits(8) = {
 
 register pmpcfg_n : vector(64, Pmpcfg_ent)
 register pmpaddr_n : vector(64, xlenbits)
+// Cached effective PMP addresses
+register pmpaddr_n_eff : vector(64, xlenbits)
```
pmundkur, line 50: I don't think we should be adding this kind of extra state to the model, but wonder what others think.

Timmmm: I was pondering doing this caching too... I agree it's not ideal to add extra state, but in a way it's not really different to the TLB. Real designs will probably do exactly this.

However I was also going to try change the code to not use `nat` everywhere, which should provide a substantial speedup. I got part way through that; I haven't checked all the overflows and wrapping and stuff yet though which is going to be the awkward bit. I'll push my code anyway.

pmundkur: I think the TLB is relevant to capturing `sfence` and exposing software bugs; this is just performance-related state that makes the model noisier.  I'd rather not add this irrelevant state and improve the compiler instead.

nadime15: > I don't think we should be adding this kind of extra state to the model, but wonder what others think.

I understand, but `perf` shows (z)pmpCheck dropping from 43% to 2.4% of cycles, which I think justifies it. As @Timmmm said, its basically what the TLB does.

> I was pondering doing this caching too... I agree it's not ideal to add extra state, but in a way it's not really different to the TLB. Real designs will probably do exactly this.
> 
> However I was also going to try change the code to not use nat everywhere, which should provide a substantial speedup. I got part way through that; I haven't checked all the overflows and wrapping and stuff yet though which is going to be the awkward bit. I'll push my code anyway.

That will likely generate better (backend) code, but for PMP the bottleneck is the call frequency rather than the call cost. The PMP configuration is basically static after boot, but we recomputed the effective addresses on every single memory access multiple times. We should try to avoid any computation.

This PR basically hits a performance threshold, I think after that, any further changes don’t matter much anymore.

model/pmp/pmp_control.sail
nadime15: Nope because `pmp_active_count` only tells you how many PMPs are set, not which one. Once you enter the main loop, you still have to check each PMP individually. (`pmp_active_count` could be 1, but the 13th PMP register might actually be the one that’s active).

I had an idea to keep a list of active PMP registers instead of iterating over all of them, but I ended up reverting those changes because 1) the type system kept complaining when I tried to dynamically store PMP register IDs in a list and use them to fetch the corresponding entries and 2) even after I worked around the type system, there was no performance improvement, just uglier code :/

pmundkur: Ah, I missed it for `pmp_usable_count`.  Regardless of this PR, the upper bound for the `foreach` loop should really be `pmp_usable_count`.

model/pmp/pmp_regs.sail
Timmmm: Here's my code to not use `nat`: #1662 - it's probably wrong in some ways for the end-of-memory/overflow cases. Also IIRC it doesn't entirely avoid `nat` anyway due to Sail using `nat` for some small generic functions (@Alasdair 's proposed solution there was to have Sail inline them).

nadime15: Ok cool! I think your change and mine could potentially be combined. The only thing I noticed which surprised me is that in your PR the Linux CI job, specifically booting Linux, takes around [6 minutes](https://github.com/riscv/sail-riscv/actions/runs/24331801764/job/71039187988?pr=1662). As far as I can tell, thats slower than the current baseline (which is around 5:20–5:30).

Maybe this was just an unlucky run due to noise or because the PR is still in progress, but I would expect it to be at least a bit faster than the baseline (once your PR is stable). This PR gets it down to around [5:00–5:10 minutes](https://github.com/nadime15/sail-riscv/actions/runs/24342611836/job/71075041597).

pmundkur: Could we get times for when `memory.pmp.usable_count` is set to 0 (and the bound of the `for` loop in `pmpCheck` is changed to `sys_pmp_usable_count`), without any other changes?

nadime15: [Here](https://github.com/nadime15/sail-riscv/actions/runs/24345889188/job/71086511560), no idea why but it failed after 5:09 (locally it worked).

Timmmm: Yeah I think given my code didn't really speed things up much I think we don't really have a choice but to cache the PMP addresses.

I feel like we should probably merge my PR anyway (once I've double checked all of the edge cases), since it will potentially be faster in future, and also I feel like it's more useful for people reading the spec to have code that could actually be real hardware rather than `nat` (it probably helps for things like the SystemVerilog backend too).

model/pmp/pmp_control.sail
```diff
@@ -118,6 +118,11 @@ function pmpCheck forall 'n, 0 < 'n <= max_mem_access . (
 
   if sys_pmp_count == 0 then return None();
 
+  if pmp_active_count == 0 then return (
```
Timmmm, line 121:
```suggestion
  // This is a performance optimisation for the
  // case where all PMPs are OFF.
  if pmp_all_off then return (
```

model/pmp/pmp_control.sail
```diff
@@ -150,5 +157,7 @@ function reset_pmp() -> unit = {
     // On reset the PMP register's A and L bits are set to 0 unless the platform
     // mandates a different value.
     pmpcfg_n[i] = [pmpcfg_n[i] with A = pmpAddrMatchType_encdec(OFF), L = 0b0];
+    pmpaddr_n_eff[i] = zeros();
   };
+  pmp_active_count = 0
```
Timmmm, line 162: I would maybe do this instead. Not as efficient but resets are rare and it is clearer.

```suggestion
    pmpUpdateCache();
```

And you can remove line 160.

model/pmp/pmp_regs.sail
```diff
@@ -20,6 +20,9 @@ let sys_pmp_usable_count : range(0, 64) = config memory.pmp.usable_count
 // G=0 -> 4 bytes, G=10 -> 4096 bytes.
 let sys_pmp_grain : range(0, 63) = config memory.pmp.grain
 
+// Number of active PMP entries
+register pmp_active_count : nat
```
Timmmm, line 24: Since you never actually use the number you could just replace this with a `bool`:

```suggestion
register pmp_all_off : bool
```

model/pmp/pmp_regs.sail
```diff
@@ -138,17 +154,18 @@ private function pmpWriteCfgReg(n : range(0, 15), v : xlenbits) -> unit = {
...
     assert(n % 2 == 0, "Unexpected pmp config reg write");
     foreach (i from 0 to 7) {
       let idx = n*4 + i;
       if idx < sys_pmp_usable_count then
-        pmpcfg_n[idx]  = pmpWriteCfg(pmpcfg_n[idx],  v[8*i+7 .. 8*i]);
+        pmpcfg_n[idx] = pmpWriteCfg(pmpcfg_n[idx], v[8*i+7 .. 8*i]);
     }
-  }
+  };
+  pmpUpdateCache()
```
Timmmm, line 168:
```suggestion
  pmpUpdateCache();
```

OCD :-D

model/pmp/pmp_regs.sail
```diff
@@ -164,6 +181,7 @@ private function pmpWriteAddrReg(n : range(0, 63), v : xlenbits) -> unit = {
       pmpaddr_n[n],
       v,
     );
+  pmpUpdateCache()
```
Timmmm, line 184:
```suggestion
  pmpUpdateCache();
```

model/pmp/pmp_regs.sail
```diff
@@ -89,6 +94,17 @@ private function pmpReadAddrReg(n : range(0, 63)) -> xlenbits = {
   }
 }
 
+// Update cached PMP effective addresses and active entry count (after CSR writes)
+private function pmpUpdateCache() -> unit = {
+  pmp_active_count = 0;
+  foreach (i from 0 to sys_pmp_count - 1) {
+    pmpaddr_n_eff[i] = pmpReadAddrReg(i);
+    if pmpAddrMatchType_encdec(pmpcfg_n[i][A]) != OFF then {
+      pmp_active_count = pmp_active_count + 1;
```
Timmmm, line 103:
```suggestion
      pmp_all_off = false;
      return;
```

model/pmp/pmp_regs.sail
```diff
@@ -89,6 +94,17 @@ private function pmpReadAddrReg(n : range(0, 63)) -> xlenbits = {
   }
 }
 
+// Update cached PMP effective addresses and active entry count (after CSR writes)
+private function pmpUpdateCache() -> unit = {
+  pmp_active_count = 0;
```
Timmmm, line 99:
```suggestion
  pmp_all_off = true;
```

Timmmm (changes requested):
LGTM with some minor changes.

pmundkur:
I still don't think we should be adding extra state like this.
- This is not like a TLB; TLBs are specifically discussed in the spec, and have specific instructions targeting its management like `sfence`.   A PMP cache is not mentioned in the spec.
- It complicates formal reasoning using the model with state not specified in the spec.
- It creates noise that gets in the way of adding other PMP extensions like Smepmp.
- If one needs to boot OSes like Linux, one can just turn _off_ PMP to get a faster boot.  The  opposite is the case with a TLB, where to boot Linux faster, you need to turn _on_ a TLB.

nadime15:
> It complicates formal reasoning using the model with state not specified in the spec.
> It creates noise that gets in the way of adding other PMP extensions like Smepmp.

Thats really hard to imagine and I think this is not really true. The cached state is 100% derived from existing registers, contains zero new information, and is always in sync. Its not "extra state" in any meaningful architectural sense, its just a performance optimization of existing state. I dont see how it complicates formal reasoning or Smepmp any more than the existing TLB does.

Could you provide details how this complicates the formal reasoning or how this gets in the way of the implemation of Smepmp?

> If one needs to boot OSes like Linux, one can just turn off PMP to get a faster boot. The opposite is the case with a TLB, where to boot Linux faster, you need to turn on a TLB.

But maybe I actually want to test it with tons of PMPs enabled? Its not only about Linux, I can imagine a lot of test suites that all use the same start.S/boot.S etc. file that potentially configures PMP (like the [RISC-V Test repository](https://github.com/riscv/riscv-test-env/blob/a1c373ec89a3500630bafabf406108a8fc568bcc/v/vm.c#L295) partially does. So it would literally improve the run of time of all these suits. If this PR already provides performance improvements in the single digit percentages for workloads with only 1–2 PMP regions enabled, I can only imagine the gains will be even more noticeable when using more than 10.

But to keep things simple, we can just vote. If people think its useful, we merge it, if not, we dont. Either way works for me.

model/pmp/pmp_regs.sail
nadime15: Yeah sure, sounds good to me!

Timmmm:
I think even though the TLB is mentioned in the spec, it is also technically not required in the model. The TLB is really pretty similar to this PMP cache. The spec *does* talk about PMP caching, which isn't necessarily always invisible:

> Implementations with virtual memory are permitted to perform address translations speculatively and earlier than required by an explicit memory access, and are permitted to cache them in address translation cache structures—including possibly caching the identity mappings from effective address to physical address used in Bare translation modes and M-mode. The PMP settings for the resulting physical address may be checked (and possibly cached) at any point between the address translation and the explicit memory access. Hence, when the PMP settings are modified, M-mode software must synchronize the PMP settings with the virtual memory system and any PMP or address-translation caches. This is accomplished by executing an SFENCE.VMA instruction with rs1=x0 and rs2=x0, after the PMP CSRs are written. 

However, I also expect that both the PMP cache and TLB will make formal work a bit more difficult. I don't think anyone has done any serious formal work with the TLB enabled. Ibex is machine mode only. LowRISC & Capabilities Ltd are working on end-to-end verification of the CVA6 (+CHERI) so I guess we might see there.

I'd be curious to know what @Alasdair thinks. But I do think the performance of the model currently with PMP enabled (which is a totally reasonable thing to do) is bad enough that we should probably do this.

If it ends up being painful for formal we could potentially add `$iftarget c cpp` around the code.

Alasdair:
Inlining the signed and unsigned comparisons in the prelude seems to offer a speed up for just the test cases:

From
```
Total Test time (real) =  60.23 sec
```
to
```
Total Test time (real) =  48.17 sec
```
will need to do some actual profiling too, but there are probably a few other easy things that would help.

Alasdair:
I did a bit of further optimisation, but the structure of the PMP check functions as written mostly stops any attempt to really make them low cost. At the start of the function:
```
  let addr = unsigned(addr);
  let width = unsigned(width);
```
`addr` and `width` are both potentially 64-bit, but because the top bit could be `1` unsigned cannot convert them to a signed `int64_t`. Leaving them as bit vectors would naively seem to change the semantics as we add them together in `pmpRangeMatch`, so values near the top of the address range would wrap when using bitvector addition. Even if we did convert to `int64_t`, this addition would require an `int65_t` to avoid overflow, so things would get cast to big integers there by Sail anyway.

Edit: Actually we know width is small in the function that calls `pmpMatchAddr`, so we could just avoid converting it to xlenbits. The problem is still that `addr + width` can wrap though.

Timmmm:
> `addr` and `width` are both potentially 64-bit, but because the top bit could be `1` unsigned cannot convert them to a signed `int64_t`. t `addr + width` can wrap though.

Oh is sbits specifically signed? I didn't realise that. Perhaps the `len` field could also be used to encode the sign somehow. Bit of a large change though.

I removed the `unsigned()`s in #1662 but as you say it means it might wrap and I haven't thought about that edge case yet. Might be worth profiling that anyway since I'm like 60% sure we can make it work.

We should definitely add some test cases for memory accesses that wrap the end of memory anyway.

PR #1655, opened 2026-04-07
Introduce a Vector payload for memory accesses.

pmundkur:
For now, these behave identically to scalar (`Data`) payloads.

However, these accesses can be treated differently for purposes of misalignment.  As per the vector spec,

> Support for misaligned vector memory accesses is independent of an implementation’s support for misaligned scalar memory accesses.

Adding this payload enables this decoupling of misaligned support for scalar and vector accesses.


jordancarlin (approved):
LGTM. We might want to rename `Data` to something more descriptive eventually as we break more and more out of it.

Timmmm:
LGTM. I *think* this is still usable with CHERI since the other access type CHERI currently uses (at least the last time I touched it) is for capability loads/stores, and IIRC you can't do vector capability loads/stores.

Timmmm (approved):

Timmmm:
@bacam @Alasdair do you want take a look at this Lean failure? It looks like the change to `vmem_types.sail` caused an issue but it looks pretty innocuous to me.

jordancarlin:
> @bacam @Alasdair do you want take a look at this Lean failure? It looks like the change to `vmem_types.sail` caused an issue but it looks pretty innocuous to me.

Already being discussed in https://github.com/rems-project/sail/issues/1654.

bacam:
The Lean problem should be fixed now.  I'll try rerunning the CI job.

PR #1656, opened 2026-04-10
Detect potentially infinite trap loops.

pmundkur:
This helps in automated tested environments where an infinite loop caused by a failing test can cause issues.

An `xret_callback` is used in combination with the existing `trap_callback` to detect trap loops.  The `trap_callback` is now called after the state changes due to the trap are visible.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -610,7 +610,7 @@ void run_sail(ModelImpl &model, const CLIOptions &opts, traploop_detector &loop_
       model.ztick_clock(UNIT);
     }
 
-    if (!is_waiting && loop_detector.loop_detected()) {
+    if (loop_detector.loop_detected()) {
       fprintf(
         stdout,
         "FAILURE: possible trap loop detected with MEPC=0x%" PRIx64 " and SEPC=0x%" PRIx64 "\n",
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -36,6 +36,7 @@
 #include "riscv_model_impl.h"
 #include "rvfi_dii.h"
 #include "sail_riscv_version.h"
+#include "traploop_detector.h"
 
 using std::chrono::duration_cast;
 using std::chrono::milliseconds;
@@ -127,6 +128,7 @@ struct CLIOptions {
   bool do_print_isa = false;
 
   bool use_rv32_default = false;
+  bool disable_trap_loop_detection = false;
   std::string config_file;
   std::vector<std::string> config_overrides;
   std::string term_log;
@@ -186,6 +188,11 @@ static CLIOptions parse_cli(int argc, char **argv) {
   );
   app.add_flag("--use-abi-names", opts.config_use_abi_names, "Use ABI register names in trace log");
   app.add_flag("--rv32", opts.use_rv32_default, "Use the default RV32 configuration");
+  app.add_flag(
+    "--disable-trap-loop-detection",
+    opts.disable_trap_loop_detection,
+    "Disable detection of potentially infinite trap loops"
+  );
 
   app.add_option("--device-tree-blob", opts.dtb_file, "Device tree blob file")
     ->check(CLI::ExistingFile)
@@ -506,7 +513,7 @@ void flush_logs() {
   fflush(trace_log);
 }
 
-void run_sail(ModelImpl &model, const CLIOptions &opts) {
+void run_sail(ModelImpl &model, const CLIOptions &opts, traploop_detector &loop_detector) {
   bool is_waiting = false;
   // The emulator tick increments time by 1 at every step, so the number
   // of steps to wait is equal to the needed increment in the time CSR.
@@ -602,6 +609,16 @@ void run_sail(ModelImpl &model, const CLIOptions &opts) {
     } else if (wait_steps_remaining > 0) {
       model.ztick_clock(UNIT);
     }
+
+    if (!is_waiting && loop_detector.loop_detected()) {
+      fprintf(
+        stdout,
+        "FAILURE: possible trap loop detected with MEPC=0x%" PRIx64 " and SEPC=0x%" PRIx64 "\n",
+        loop_detector.mepc(),
+        loop_detector.sepc()
+      );
+      exit(EXIT_FAILURE);
+    }
   }
 
   // This is reached if there is a Sail exception, HTIF has indicated
@@ -687,6 +704,11 @@ int inner_main(int argc, char **argv) {
   model.set_config_use_abi_names(opts.config_use_abi_names);
 
   model.set_config_print_step(opts.config_print_step);
+
+  traploop_detector loop_detector;
+  if (!opts.disable_trap_loop_detection) {
+    model.register_callback(&loop_detector);
+  }
 
   std::string config_json_string;
   if (!opts.config_file.empty()) {
@@ -803,11 +825,12 @@ int inner_main(int argc, char **argv) {
   init_end = steady_clock::now();
 
   do {
-    run_sail(model, opts);
+    run_sail(model, opts, loop_detector);
     // `run_sail` only returns in the case of rvfi.
     if (rvfi) {
       /* Reset for next test */
       reinit_sail(model, entry, opts.config_file.c_str());
+      loop_detector.reset();
     }
   } while (rvfi);
 
--- a/c_emulator/CMakeLists.txt
+++ b/c_emulator/CMakeLists.txt
@@ -75,6 +75,8 @@ add_executable(sail_riscv_sim
     riscv_callbacks_log.h
     riscv_callbacks_rvfi.cpp
     riscv_callbacks_rvfi.h
+    traploop_detector.cpp
+    traploop_detector.h
 )
 
 # The generated C model is not warnings-clean, silence them.
--- a/c_emulator/riscv_callbacks_if.cpp
+++ b/c_emulator/riscv_callbacks_if.cpp
@@ -91,6 +91,9 @@ void callbacks_if::trap_callback(
 ) {
 }
 
+void callbacks_if::xret_callback([[maybe_unused]] hart::Model &model, [[maybe_unused]] bool is_mret) {
+}
+
 // Page table walk callbacks
 void callbacks_if::ptw_start_callback(
   [[maybe_unused]] hart::Model &model,
--- a/c_emulator/riscv_callbacks_if.h
+++ b/c_emulator/riscv_callbacks_if.h
@@ -46,6 +46,8 @@ public:
 
   virtual void trap_callback(hart::Model &model, bool is_interrupt, fbits cause);
 
+  virtual void xret_callback(hart::Model &model, bool is_mret);
+
   // Page table walk callbacks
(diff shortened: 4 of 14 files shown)


pmundkur:
This is to help with situations like https://github.com/riscv/riscv-arch-test/issues/1238

c_emulator/riscv_sim.cpp
```diff
@@ -185,6 +186,7 @@ static CLIOptions parse_cli(int argc, char **argv) {
   );
   app.add_flag("--use-abi-names", opts.config_use_abi_names, "Use ABI register names in trace log");
   app.add_flag("--rv32", opts.use_rv32_default, "Use the default RV32 configuration");
+  app.add_flag("--detect-trap-loops", opts.detect_trap_loops, "Exit if a potentially infinite trap loop is detected");
```
nadime15: I think thats quite useful! Could we extend this further by supporting exits in different modes? Its not uncommon that [tests](https://github.com/riscv/riscv-test-env/blob/6de71edb142be36319e380ce782c3d1830c65d68/p/riscv_test.h#L233-L239) delegate exceptions. Ideally, it would allow specifying the modes (M, S and possibly VS once supported). 

or at least 

```
app.add_flag("--detect-trap-loops", opts.detect_trap_loops, 
    "Exit if a potentially infinite trap loop is detected (M, S, and VS mode)");
```

and then later in detected_trap_loop something like:
```

    switch (zcur_privilege) {
        case Machine:
            if(...)
              current_pc = zmepc.bits;
               ...
            break;
        case Supervisor:
            current_pc = zsepc.bits;
            ...
            break;
        case VirtualSupervisor:
            current_pc = vsepc.bits;
            ...
            break;
        default:
            return false;
    }
```

pmundkur: Actually I was wondering if the instret check would suffice?

pmundkur: `minstret` is incorrect due to `mcountinhibit`, perhaps we could just check if the PC has changed?

Timmmm: I don't quite understand the `instret` check. Won't the trap handler itself retire instructions and therefore increment `instret`? What situations is this meant to detect exactly?

Timmmm (commented):
As an alternative we could implement [Smdbltrap](https://riscv-isa-manual.timhutt.co.uk/riscv-privileged.html#machine-double-trap). That might cover most of the use cases for this, and we're going to want to do it eventually anyway.

c_emulator/riscv_sim.cpp
pmundkur: The case I was seeing was where the trap handler wasn't able to fetch the instruction at `mtvec` and entered a trap loop, but I don't know what other kinds of loops are occurring in ACT4.    I don't think ACT4 can assume Smdbltrap yet.  Perhaps we could have callbacks for `mret/sret` and track the difference between a count of `trap_callbacks` and `xret_callbacks`?

allenjbaum:
The case where an attempted fetch of the trap handler entry address in
Xtvec itself trap wouldn't.
It is also the subject of a security vulnerability in RISC-V; it makes an
ideal launching pad for glitch attacks, and has been demonstrated to work.

c_emulator/riscv_sim.cpp
Timmmm: ACT4 doesn't need to assume Smdbltrp - merely implementing it allows us to halt the simulator when there is a double trap:

> When the Smrnmi extension is not implemented, or if the Smrnmi extension is implemented and mnstatus.NMIE is 0, the hart enters a critical-error state without updating any architectural state, including the pc. This state involves ceasing execution, disabling all interrupts (including NMIs), and asserting a critical-error signal to the platform.

The only potential issues are if ACT4 checks the `mstatus.MDT` bit because it will currently be read-only 0 but if you implement Smdbltrp it will normally be 1. @jordancarlin would that cause any issues? Do ACT4s rely on mstatus *exactly* matching Spike?

If that would be an issue then I think we should still implement Smdbltrp but then we can just use its machinery to provide this functionality even when Smdbltrp is actually disabled. This is pretty similar to [Ibex's double fault detection](https://ibex-core.readthedocs.io/en/latest/03_reference/exception_interrupts.html#double-fault-detection).

nadime15:
Ok I think I misunderstood the use case :D

On my side, I was thinking more generally about infinite trapping (trap → mret → resume execution → trap again at the same PC). However I believe that this is not reliably detectable without false positives.

But I think @pmundkur would be covered with @Timmmm suggestion, assuming it does not call mret before trapping again.

c_emulator/riscv_sim.cpp
jordancarlin: It would definitely cause issues because we will eventually have tests that try writing to that bit. We need the CSRs to match exactly.

jordancarlin:
I'm wondering if a simpler version of this that is likely to catch the common case is just checking if we try to execute an instruction from address 0x0 3+ times in a row. That would stop the core if it trapped when `mtvec` was not initialized.

jordancarlin:
We also might want to consider making this the default behavior instead of needing a flag. Realistically, no one wants the simulation to keep running if it is stuck in a trap loop and several other simulators already terminate when that happens (at least Whisper and Imperas).

pmundkur:
@jordancarlin for uninitialized `mtvec` this PR should work as is; that's the case I was seeing.   Is that the only kind of trap loop that happens in ACT4?  If so, we could use this PR until we merge Smdbltrp.

jordancarlin:
> @jordancarlin for uninitialized `mtvec` this PR should work as is; that's the case I was seeing. Is that the only kind of trap loop that happens in ACT4? If so, we could use this PR until we merge Smdbltrp.

It's the only one I've seen. Some of the unpriv tests don't initialize mtvec because we are not guaranteed to be running on a core with support for priv features like `mtvec`.

Timmmm:
I might have a go at Smdbltrp tonight. It looks very simple.

c_emulator/riscv_sim.cpp
Timmmm: Ok I implemented Smdbltrp in #1659 but having done that and given @jordancarlin 's point I think maybe this PR is the best option after all.

I think the implementation could potentially be made more robust though by counting traps and xrets. If you see N traps in a row without an xret that's a pretty good sign something has gone horribly wrong. We do currently have callbacks for traps but not xret as far as I can see so that would need to be added.

Or we could just go with this.

pmundkur: I too prefer the approach with xret callbacks as it's likely to be more robust.  I'll add those.

jordancarlin: Maybe an initial version should drop the `minstret` check and add a check for the PC being zero in addition to it not changing. Once we add an `xret` callback, we could drop the PC=0 check in favor of checking for no `xret`.

jordancarlin: > I too prefer the approach with xret callbacks as it's likely to be more robust. I'll add those.

Even better!

nadime15: @Timmmm Couldn we use #1659 to enable Smdbltrp implicitly, but mask mstatus.mdt on reads when running with --detect-trap-loops, so that mstatus.mdt always appears as 0? So your PR should be enough then to implement the desired behaviour

jordancarlin: That wouldn't quite work because `mstatus.MDT` could be left as 1 if the test stays in machine mode so the double trap checking would never be enabled. As mentioned above, I also think we should avoid needing a flag to enable this and should instead make it the default behavior.

nadime15: > That wouldn't quite work because mstatus.MDT could be left as 1 if the test stays in machine mode so the double trap checking would never be enabled.

I am not sure I follow, mstatus.mdt=1 would trigger the detection not disable it? Could you clarify what scenario you are thinking of?

Unless Im missing something, I think we could always run the double-trap detection (#1659) without requiring a flag regardless of whether the extension is enabled or disabled. If the extension is disabled, we could simply mask the corresponding bit in mstatus on read.

Timmmm: Yeah I thought about that but I think actually it would just be simpler to do it separately.

jordancarlin: > I am not sure I follow, mstatus.mdt=1 would trigger the detection not disable it? Could you clarify what scenario you are thinking of?

Sorry. I got that backwards. I suppose it could work, but I think I agree with Tim that it seems simpler to just make them two separate features. Especially for purposes of inclusion in the ISA manual; it would be kind of strange if the Smdbltrp code had bits that were always enabled.

c_emulator/riscv_sim.cpp
```diff
@@ -602,6 +609,16 @@ void run_sail(ModelImpl &model, const CLIOptions &opts) {
     } else if (wait_steps_remaining > 0) {
       model.ztick_clock(UNIT);
     }
+
+    if (!is_waiting && loop_detector.loop_detected()) {
```
Timmmm: Probably don't need `!is_waiting` since you're not going to see any traps/xrets while waiting anyway?

```suggestion
    if (loop_detector.loop_detected()) {
```

Timmmm (approved):
One minor nit but otherwise LGTM!

Timmmm (approved):

jordancarlin (approved):

pmundkur:
So the threshold used in this PR is too low for some ACT4 tests, funnily enough the ones testing traps and exceptions :-)

Here are the trap/xret counts for at the end of some of these tests:

- ExceptionsS/ExceptionsS-00.elf: traps=1912, xrets=1823
- ExceptionsZicboS/ExceptionsZicboS-00.elf: traps=160, xrets=97
- ExceptionsZicboU/ExceptionsZicboU-00.elf: traps=76, xrets=33

@jordancarlin Are these numbers expected?  Perhaps the trap loop threshold should be made configurable?

jordancarlin:
That doesn't seem quite right. I'll look into it.

arichardson:
Tests can legitimately just jump somewhere else from the trap handler, so they might not be paired up.

Since this is just trying to detect the common infinite trap case, maybe the less false-positive prone check would be to detect a trap when executing the first few instructions of mtvec?

In CHERI QEMU I believe the logic I used inside handle_exception was just `fault_pc == mtvec`. That handles the case where you keep jumping to zero and failing the ifetch.
It doesn't handle something like a trap handler that fails to load something on the nth instruction, but that could be done by counting "num_traps" + "instret since trap" and if you say 5 traps each with less than 5 instruction executed since the jump to mtvec you conclude we are in an infinite loop. Or we could even track the last N program counters and compare if they are the same.

pmundkur:
An initial version of this did use `instret` (very naively though), but I removed it since it can be affected by `mcountinhibit` and `Smcntrpmf`.

arichardson:
> An initial version of this did use `instret` (very naively though), but I removed it since it can be affected by `mcountinhibit` and `Smcntrpmf`.

I didn't mean using instret, but instead using a callback for each retire that is not affected by csrs.

pmundkur:
> I didn't mean using instret, but instead using a callback for each retire that is not affected by csrs.

Ahh, makes sense.

PR #416, opened 2024-02-25
Double trap - Add unratified Ssdbltrp and Smdbltrp extensions

ved-rivos:
Spec: https://github.com/riscv/riscv-double-trap
Tests: https://github.com/riscv-non-isa/riscv-arch-test/pull/438

PR #1659, opened 2026-04-10
Add support for Smdbltrp (double trap)

Timmmm:
This adds support for the Smdbltrp and Ssdbltrp extensions which allow detecting and handling double traps.

When a double trap in machine mode is detected the hart is supposed to halt execution and raise an error to the platform. For now, for simplicity I've just used an assertion. In future we should probably change this to return from try_step() in a way that indicates to C++ that there has been a double trap.

The Smdbltrap extension requires the `mtval2` CSR (also used by hypervisor) so I added that too.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -613,6 +613,15 @@
     "Smcntrpmf": {
       "supported": true
     },
+    // The double trap extensions are not enabled by default because
+    // mstatus[MDT] is reset to 1, which means unless your boot code zeroes
+    // it the first trap you take will be an unexpected double trap.
+    "Smdbltrp": {
+      "supported": false
+    },
+    "Ssdbltrp": {
+      "supported": false
+    },
     "Svbare": {
       "supported": true,
       "sfence_vma_illegal_if_svbare_only": true
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -777,6 +777,15 @@
     "Smcntrpmf": {
       "supported": true
     },
+    // The double trap extensions are not enabled by default because
+    // mstatus[MDT] is reset to 1, which means unless your boot code zeroes
+    // it the first trap you take will be an unexpected double trap.
+    "Smdbltrp": {
+      "supported": false
+    },
+    "Ssdbltrp": {
+      "supported": false
+    },
     "Svbare": {
       "supported": true,
       "sfence_vma_illegal_if_svbare_only": true
--- a/model/core/extensions.sail
+++ b/model/core/extensions.sail
@@ -534,11 +534,20 @@ function clause hartSupports(Ext_Svvptc) = config extensions.Svvptc.supported
 enum clause extension = Ext_Smcntrpmf
 mapping clause extensionName = Ext_Smcntrpmf <-> "smcntrpmf"
 function clause hartSupports(Ext_Smcntrpmf) = config extensions.Smcntrpmf.supported
+// Double Trap (Machine)
+enum clause extension = Ext_Smdbltrp
+mapping clause extensionName = Ext_Smdbltrp <-> "smdbltrp"
+function clause hartSupports(Ext_Smdbltrp) = config extensions.Smdbltrp.supported
+function clause currentlyEnabled(Ext_Smdbltrp) = hartSupports(Ext_Smdbltrp)
 // Machine-mode view of the state-enable extension
 enum clause extension = Ext_Smstateen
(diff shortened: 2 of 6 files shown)


Timmmm:
Not tested of course. I should probably write a first party test.

Timmmm:
Wait I just realised I totally missed Ssdbltrp and the `menvcfg.DTE` bit. Also I should have checked but Ved already had an implementation in #416 - I'll cross check mine with his.

jordancarlin:
@Timmmm Does this build off of https://github.com/riscv/sail-riscv/pull/416?

jordancarlin:
> Wait I just realised I totally missed Ssdbltrp and the `menvcfg.DTE` bit. Also I should have checked but Ved already had an implementation in #416 - I'll cross check mine with his.

> @Timmmm Does this build off of #416?

Oops. Sent this before seeing your latest comment.

Timmmm:
Ok I added Ssdbltrp. @ved-rivos wrote some tests for the old riscv-arch-tests in https://github.com/riscv/riscv-arch-test/pull/438 - is it possible to get those into ACT4 somehow?

jordancarlin:
> @ved-rivos wrote some tests for the old riscv-arch-tests in [riscv/riscv-arch-test#438](https://github.com/riscv/riscv-arch-test/pull/438) - is it possible to get those into ACT4 somehow?

We will get those ported eventually, but they are fairly low priority right now unless someone else wants to take the porting work on since they are not part of RVA23.

Timmmm:
I tweaked this a little, and also I had to disable it by default because it's one of the few RISC-V extensions that isn't backwards compatible: `mstatus[MDT]` is reset to 1, which causes the next trap to be an unexpected double trap.

I presume the intention is that firmware sets `mstatus[MDT]=0` after it has set up an appropriate trap handler, but our tests don't do that.

Timmmm:
Mmm formal backends are failing because I made `trap_handler` recursive. I will refactor things so that it isn't (I don't think it needs to be).

Timmmm:
Ok I think this is ready. No tests I'm afraid though. @jordancarlin do you know if any already exist for double trap?

nadime15:
@Timmmm The new Alibaba test suite might be helpful:

https://github.com/XUANTIE-RV/damo-rv-priv-ats/blob/main/README_en.md

It's really easy to set up. Just clone the repository and run:

`make -C Ssdbltrp  sail CROSS_COMPILER=riscv64-unknown-elf- SAIL=/path_to_sail/sail-riscv/build/c_emulator/sail_riscv_sim`

The test suite is still fairly new, so it's not perfect yet. It also includes tests for Smdbltrp, but they don't work out of the box, you'll need to do a bit of tinkering to get them running. I tried the Ssdbltrp tests, and they worked.

Timmmm:
Ah nice test suite. Maybe it's worth adding to our CI? I feel like we need an index of these test suites, I'm starting to lose track.

I fixed the only failing Ssdbltrp tests which were due to #1803.

```
========================================
  Test Summary
========================================
  Total:   33
  Passed:  16
  Failed:  0
  Skipped: 17
----------------------------------------
  Assertions: 19 total, 19 passed, 0 failed
----------------------------------------
```

Timmmm:
Ah I didn't see it was already fixed in #1808. I'll revert that.

jordancarlin:
The ACTs don't have tests for this extension yet. The focus is on RVA23 for now and then we will get caught up with the other extensions. The Alibaba test suite is one of the newer releases and parts of it may be folded into the ACTs. I think @nadime15 already added it to CI in his Hypervisor PR (#1419).

ved-rivos:
Some tests are here : https://github.com/riscv/riscv-arch-test/pull/438

jordancarlin:
Thanks for the reminder about the ACT3 tests for them @ved-rivos. I am not in a position to run ACT3 tests at the moment, but we will definitely use those as a starting point when we add ACT4 tests for these extensions.

jordancarlin:
This is going to be an important extension to get added soon for sake of ACTs even though it isn't part of RVA23 because of how it influences the behavior of a number of other extensions.

nadime15:
I can take a look at it and update this PR.

pmundkur:
@nadime15 you might have to create a new PR to pass DCO since @Timmmm's commit wasn't signed (unless @Timmmm sees this and repushes a signed commit).

nadime15:
Closing this PR since #1961 was based on this PR and has now been merged.

PR #1663, opened 2026-04-14
Require PMA regions to be page-aligned.

pmundkur:
This ensures that PMAs do not change within intra-page accesses; this simplifies PMA checks, and most implementations are expected to conform with this restriction.

### DIFF
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -241,13 +241,13 @@ private function check_pma_region(region : PMA_Region) -> bool = {
   // will not change within intra-page accesses.  Although this constraint
   // is not imposed by the specification, most implementations are
   // expected to conform with it.
-  let align : bits(64) = ones(64) << 12;
-  if (region.base & align) != region.base then {
+  if region.base[pagesize_bits - 1 .. 0] != zeros() then {
     print_endline("Memory region starting at " ^ bits_str(region.base) ^ " does not start on a 4K (page) boundary.");
     return false;
   };
-  if (region.size & align) != region.size then {
-    print_endline("Memory region starting at " ^ bits_str(region.base) ^ " does not end on a 4K (page) boundary.");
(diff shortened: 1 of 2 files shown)


model/postlude/validate_config.sail
```diff
@@ -236,6 +236,20 @@ private function check_vext_config() -> bool = {
 }
 
 private function check_pma_region(region : PMA_Region) -> bool = {
+  // We currently require PMAs to be aligned to page boundaries.  This
+  // simplifies memory protection and alignment checks: PMA properties
+  // will not change within intra-page accesses.  Although this constraint
+  // is not imposed by the specification, most implementations are
+  // expected to conform with it.
+  let align : bits(64) = ones(64) << 12;
+  if (region.base & align) != region.base then {
```
Timmmm:
```suggestion
  if region.base[pagesize_bits - 1 .. 0] != zeros() then {
```

model/postlude/validate_config.sail
```diff
@@ -236,6 +236,20 @@ private function check_vext_config() -> bool = {
...
+    print_endline("Memory region starting at " ^ bits_str(region.base) ^ " does not start on a 4K (page) boundary.");
+    return false;
+  };
+  if (region.size & align) != region.size then {
```
Timmmm:
```suggestion
  if region.size[pagesize_bits - 1 .. 0] != zeros() then {
```

model/postlude/validate_config.sail
```diff
@@ -236,6 +236,20 @@ private function check_vext_config() -> bool = {
...
+    print_endline("Memory region starting at " ^ bits_str(region.base) ^ " does not end on a 4K (page) boundary.");
+  };
```
Timmmm, line 251: Think you forgot a `return false;` here?

Timmmm (commented):
Might also be worth a brief comment in config.json.in.

Timmmm (approved):

jordancarlin (approved):

PR #1664, opened 2026-04-14
Remove the panic branch in translate_TLB_hit.

nadime15:
When Svade and Svadu are disabled, a store to a page with PTE_D clear triggers a PTE writeback. The TLB entry must already be present for this to hit translate_TLB_hit. If PMP blocks the page table entry, this should raise an access fault rather than panic.

The added test triggers this panic pre-change when Svadu and Svade are disabled.

Fixes https://github.com/riscv/sail-riscv/issues/1633


Timmmm (approved):
Awesome, thanks!

jordancarlin (approved):

pmundkur (approved):

PR #1666, opened 2026-04-15
Tweak trap loop detector heuristic.

pmundkur:
The previous heuristic did not suffice for a few ACT4 tests where the trap count grew slowly but steadily over the xret count.

Adjust by introducing an instruction retired callback and resetting the loop detector when a sufficient number of instructions have retired since the last trap.

This still catches tight trap loops but allows the above tests to complete successfully.

### DIFF
--- a/c_emulator/traploop_detector.cpp
+++ b/c_emulator/traploop_detector.cpp
@@ -7,34 +7,46 @@ traploop_detector::traploop_detector() {
   reset();
 }
 
+void traploop_detector::reset_loop() {
+  mepc_at_first_trap = 0;
+  sepc_at_first_trap = 0;
+  instret_at_last_trap = 0;
+  nested_trap_count = 0;
+}
+
 void traploop_detector::reset() {
-  trap_count = 0;
-  trap_mepc = 0;
-  trap_sepc = 0;
-  xret_count = 0;
+  instret = 0;
+  reset_loop();
 }
 
 void traploop_detector::trap_callback(hart::Model &model, bool, fbits) {
-  trap_count++;
-  trap_mepc = model.zmepc.bits;
-  trap_sepc = model.zsepc.bits;
+  if (nested_trap_count == 0) {
+    mepc_at_first_trap = model.zmepc.bits;
+    sepc_at_first_trap = model.zsepc.bits;
+  }
+  nested_trap_count++;
+  instret_at_last_trap = instret;
 }
 
 void traploop_detector::xret_callback(hart::Model &, bool) {
-  xret_count++;
+  reset_loop();
 }
 
-bool traploop_detector::loop_detected() {
-  // It would be very unusual to have xrets exceeding trap-count by a
-  // large amount.  Should this be handled as well?
-  uint64_t diff = (trap_count > xret_count) ? trap_count - xret_count : 0;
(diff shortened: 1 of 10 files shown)


Timmmm:
Rather than adding another callback, can't you just use `post_step_callback()`?

pmundkur:
`post_step_callback()` is called on every step, regardless of whether an instruction retired or not, so you'd have to probe the model to find out.  The `instret_callback()` is gated by the `retired` variable in `step()` so as long as that is accurate, the `instret_callback()` is more reliable.

jordancarlin:
> `post_step_callback()` is called on every step, regardless of whether an instruction retired or not, so you'd have to probe the model to find out. The `instret_callback()` is gated by the `retired` variable in `step()` so as long as that is accurate, the `instret_callback()` is more reliable.

Is that extra granularity needed for the trap loop detector though? Seems like it would work just as well using `post_step_callback`, even if it does increment more often.

pmundkur:
> > `post_step_callback()` is called on every step, regardless of whether an instruction retired or not, so you'd have to probe the model to find out. The `instret_callback()` is gated by the `retired` variable in `step()` so as long as that is accurate, the `instret_callback()` is more reliable.
> 
> Is that extra granularity needed for the trap loop detector though? Seems like it would work just as well using `post_step_callback`, even if it does increment more often.

I'd rather that the heuristics were based on accurate counts, rather than rough numbers, otherwise heuristics based on inaccurate counts will be rather fragile.  And the `instret_callback` seems like a useful thing to have anyway, given the limitations with using the `minstret` register?

c_emulator/traploop_detector.h
```diff
@@ -16,11 +16,21 @@ class traploop_detector : public callbacks_if {
...
 
 private:
-  uint64_t trap_count = 0;
-  uint64_t trap_mepc = 0;
-  uint64_t trap_sepc = 0;
-  uint64_t xret_count = 0;
-  const uint64_t trap_count_threshold = 10;
+  void reset_loop();
+
+  uint64_t instret;
```
arichardson: Keep the initialization to 0?

c_emulator/traploop_detector.cpp
```diff
@@ -7,34 +7,46 @@ traploop_detector::traploop_detector() {
...
 }
 
 void traploop_detector::xret_callback(hart::Model &, bool) {
-  xret_count++;
+  reset_loop();
+}
+
+void traploop_detector::instret_callback(hart::Model &) {
+  instret++;
+  if ((instret_at_last_trap != 0) & (instret - instret_at_last_trap > instrets_to_reset_loop)) {
```
arichardson:
```suggestion
  if ((nested_trap_ count != 0) && instret_since_last_trap > instrets_to_reset_loop) {
```

How about just storing instret_since_last_trap and resetting that to zero on trap?

arichardson (commented):
Minor suggestions but generally looks good to me

c_emulator/traploop_detector.cpp
pmundkur: Good idea, done.

jordancarlin (approved):

arichardson (approved):

PR #1680, opened 2026-04-24
Renable the SMT builds and test.

pmundkur:
The trigger for the mentioned bug doesn't exist anymore, so the test and builds can be re-enabled.


Timmmm (approved):

pmundkur:
Hmm, it looks like the trigger for this bug in #1647 got removed in the boiler-plate removal, so we don't need to wait for a new Sail compiler release.

jordancarlin (approved):

Timmmm:
Maybe we should take the Markdown Link checker out of CI. It's very unreliable.

jordancarlin:
> Maybe we should take the Markdown Link checker out of CI. It's very unreliable.

At a minimum, we should update the GitHub Action so that it automatically retries that check a few times since it seems to be transient network issues. If that doesn't solve things, then maybe we try disabling it for external links only.

jordancarlin:
This one seems to be having very bad luck in the merge queue. This time it failed to download CLI11.

PR #1682, opened 2026-04-27
Move alignment helpers from vmem_utils to mem_utils to allow use with physical addresses.

pmundkur:
These helpers use `bits` and `xlenbits`, and this allows them to be used to implement checks for the MAG PMA. 

The API for `allowed_misaligned` has been generalized by adding a parameter for its allowed-within region, but its width parameter has been restricted to the widest possible memory access.

### DIFF
--- a/model/sys/mem_utils.sail
+++ b/model/sys/mem_utils.sail
@@ -29,8 +29,7 @@ private function prop_access_within_single(addr : bits(32)) -> bool = {
   access_within(addr, 1, 0)
 }
 
-private val allowed_misaligned : (xlenbits, word_width_wide, range(0, 12)) -> bool
-function allowed_misaligned(addr, width, region_width_exp) = {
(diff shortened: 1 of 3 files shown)


model/sys/mem_utils.sail
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// A 1-byte access is always within a 2^0 = 1-byte region.
+$[property]
+private function prop_access_within_single(addr : bits(32)) -> bool = {
+  access_within(addr, 1, 0)
+}
+
+private val allowed_misaligned : (xlenbits, word_width_wide, range(0, 12)) -> bool
+function allowed_misaligned(addr, width, region_width_exp) = {
```
jordancarlin: Probably worth combining these lines into a single function signature in this case.

jordancarlin (approved):
LGTM.

PR #1683, opened 2026-04-27
Refactor riscv_sim.cpp by moving initialization and driver logic into dedicated files

nadime15:
As of today it is impractical to link against the model because riscv_sim.cpp is quite messy, relies on global state, and contains runtime logic that is coupled with its own main(). This commit pulls the run loop, callback setup, etc. into new files (simulator.h/cpp and riscv_sim_utils.h/cpp) so other binaries can link against the model and drive the simulator without copying main's setup code.

A few notable changes:

- A new defined simulator class that owns the run loop and its setup (logs, callbacks, etc.).
- Library code no longer calls exit(), run() returns a RunStatus/RunResult, and setup paths throw ConfigError. Main (inner_main()) handles termination.
- Eliminate global state by grouping related state into structs and passing them explicitly through function interfaces.
- load_sail returns its results as a struct instead of writing globals.
- inner_main split into a few named helpers.

Open this as a draft pull request for now.

### DIFF
--- a/c_emulator/riscv_sim_utils.cpp
+++ b/c_emulator/riscv_sim_utils.cpp
@@ -1,0 +1,120 @@
+#include "riscv_sim_utils.h"
+#include "elf_loader.h"
+#include "rts.h"
+#include "symbol_table.h"
+#include <stdexcept>
+
+namespace riscv_sim {
+
+LoadResult load_sail(ModelImpl &model, const std::string &filename, bool main_file) {
+  ELF elf = ELF::open(filename);
+
+  switch (elf.architecture()) {
+  case Architecture::RV32:
+    if (model.zxlen != 32) {
+      throw ConfigError("32-bit ELF not supported by RV" + std::to_string(model.zxlen) + " model.");
+    }
+    break;
+  case Architecture::RV64:
+    if (model.zxlen != 64) {
+      throw ConfigError("64-bit ELF not supported by RV" + std::to_string(model.zxlen) + " model.");
+    }
+    break;
+  }
+
+  // Load into memory.
+  elf.load([](uint64_t address, const uint8_t *data, uint64_t length) {
+    // TODO: We could definitely improve on rts.c's memory implementation
+    // (which is O(N^2)) and writing one byte at a time here.
+    for (uint64_t i = 0; i < length; ++i) {
+      write_mem(address + i, data[i]);
+    }
+  });
+
+  // Load the entire symbol table.
+  const auto symbols = elf.symbols();
+
+  // Save reversed symbol table for log symbolization.
+  // If multiple symbols from different ELF files have the same value the first
+  // one wins.
+  const auto reversed_symbols = reverse_symbol_table(symbols);
+  g_symbols.insert(reversed_symbols.begin(), reversed_symbols.end());
+
+  LoadResult result;
+  result.entry = elf.entry();
+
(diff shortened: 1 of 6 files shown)


nadime15:
#1674 turned out to be more “complicated” than expected. I initially thought I could simply run a CI job and execute the tests using our existing infrastructure. However it seems I also need to link the model against a c++ library (google benchmark) and define a new main() that drives the simulator and handles (what Codspeed/Google Benchmark refers to as) benchmark registration.

The idea is to slightly refactor riscv_sim.cpp so users can provide their own entry point and link it against the model. Otherwise a lot of code must be copied to actually drive the model.

I think this approach could also be useful for users who want to drive the model in their own way, and it may make the codebase cleaner and easier to understand.

c_emulator/riscv_sim_utils.h
```diff
@@ -0,0 +1,28 @@
...
+public:
+  using std::runtime_error::runtime_error;
+};
+
+struct LoadResult {
+  uint64_t entry;
+  std::optional<uint64_t> htif_tohost;
+  std::optional<uint64_t> sig_start;
+  std::optional<uint64_t> sig_end;
+};
```
pmundkur, line 21: The ELF symbols could be part of the load result.  Right now, the `g_symbols` used by the tracer is populated as a side-effect of `load_sail`; it would be nice to have an explicit API in `symbol_table.h` to initialize it using a field of `LoadResult`.

c_emulator/simulator.h
```diff
@@ -0,0 +1,94 @@
...
+namespace riscv_sim {
+
+enum class RunStatus {
+  HtifSuccess,
+  HtifFailure,
+  TrapLoop,
+  SailException,
+  InstructionLimit,
+  RvfiEof,
+  RvfiEndTrace,
```
pmundkur, line 21: Perhaps not in this PR but we could split out RVFI into its own simulator class, and simplify this one.

c_emulator/riscv_sim_utils.cpp
```diff
@@ -0,0 +1,120 @@
...
+  model.zinit_boot_requirements(UNIT);
+}
+
+// reinitialize to clear state and memory, typically across tests runs
+void reinit_sail(ModelImpl &model, uint64_t elf_entry, std::optional<uint64_t> htif_tohost, const char *config_file) {
+  model.model_fini();
+  model.model_init();
+  init_sail(model, elf_entry, htif_tohost, config_file);
+}
+
```
pmundkur, line 87: Since these functions are basically just manipulating `model`, perhaps these functions should actually be implemented as members of `ModelImpl`?

c_emulator/simulator.cpp
```diff
@@ -0,0 +1,273 @@
...
+    std::fclose(trace_log);
+    trace_log = stdout;
+  }
+  if (term_fd >= 0) {
+    close(term_fd);
+    term_fd = -1;
+  }
+}
+
+RunResult Simulator::run() {
```
pmundkur, line 111: Again perhaps not in this PR, but this run loop could be split into a single-stepping function, and a run loop over that.

pmundkur (commented):
I think such a refactor would be very useful, for e.g. in #1102.  Some initial thoughts, but I'd defer to @Timmmm and @arichardson for better suggestions.

PR #1686, opened 2026-04-29
Add configuration flags to enable platforms without CLINT or SIG devices.

### LINKED ISSUE #1685: Add config options to disable platform MMIO devices
Right now, you are required to specify valid addresses for the CLINT and the SIG in the config file. Depending on your device's memory map, this may be difficult or impossible. It also adds unnecessary complexity to the configuration for DUTs that don't support interrupts. Each of these platform MMIO devices should have a `supported`/`enabled`/etc. config option that determines whether it exists at all. If it is not supported, then the address should not be validated.

pmundkur:
Fixes #1685.

### DIFF
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -370,10 +370,14 @@ private function check_mem_layout() -> bool =
       svadu    = config extensions.Svadu.supported,
     };
     let pmas_ok  = check_pma_regions(pma_regions, zeros(), zeros(), check_opts, false);
-    let clint_ok = within_configured_pma_memory("CLINT (platform.clint)", Some(IOMemory),
+    let clint_supported : bool = config platform.clint.supported;
+    let clint_ok = not(clint_supported) |
+                   within_configured_pma_memory("CLINT (platform.clint)", Some(IOMemory),
                                                 to_bits_checked(config platform.clint.base : int),
                                                 to_bits_checked(config platform.clint.size : int));
-    let sig_ok   = within_configured_pma_memory("simple interrupt generator (platform.simple_interrupt_generator)", Some(IOMemory),
+    let sig_supported : bool = config platform.simple_interrupt_generator.supported;
+    let sig_ok   = not(sig_supported) |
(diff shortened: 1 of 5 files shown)


model/postlude/validate_config.sail
```diff
@@ -370,10 +370,14 @@ private function check_mem_layout() -> bool =
...
-    let clint_ok = within_configured_pma_memory("CLINT (platform.clint)", Some(IOMemory),
+    let clint_supported : bool = config platform.clint.supported;
+    let clint_ok = not(clint_supported) |
+                   within_configured_pma_memory("CLINT (platform.clint)", Some(IOMemory),
                                                 to_bits_checked(config platform.clint.base : int),
                                                 to_bits_checked(config platform.clint.size : int));
-    let sig_ok   = within_configured_pma_memory("simple interrupt generator (platform.simple_interrupt_generator)", Some(IOMemory),
+    let sig_supported : bool = config platform.simple_interrupt_generator.supported;
+    let sig_ok   = not(sig_supported) |
+                   within_configured_pma_memory("simple interrupt generator (platform.simple_interrupt_generator)", Some(IOMemory),
```
nadime15, line 380: Is this enough? I would expect to modify within_clint/within_sig as well and do something like:
```
let sig_supported : bool = config platform.simple_interrupt_generator.supported;

private function within_sig forall 'n, 0 < 'n <= max_mem_access . (Physaddr(addr) : physaddr, width : int('n)) -> bool = {
  ...
 sig_supported & (addr_int + sizeof('n)) <= (sig_base_int + sig_size_int)
}
```

Otherwise the device is still available

pmundkur: Good point, fixed.

nadime15 (approved):

jordancarlin (approved):

Timmmm (approved):

nazar-pc:
More broadly, there are more examples of a similar configuration challenges.

For example, `extensions.Ssqosid.supported: false` still requires `rcid_length` and `mcid_length` to be non-zero, even though they are not applicable by definition. There are probably others, but I remember that one from yesterday.

Issue #1687, opened 2026-04-29
vrgatherei16.vv allowing reserved behavior vs1=vs2 at SEW=32/64


georgiatai:
Sail is allowing `vs1` and `vs2` to overlap for `vrgatherei16.vv` when SEW is set to 32 or 64. I believe that this should be reserved considering the following 2 quotes:

> The vrgatherei16.vv form uses SEW/LMUL for the data in vs2 but EEW=16 and EMUL = (16/SEW)*LMUL for the indices in vs1.

>  An encoding that would result in the same vector register being read with two or more different EEWs, including when the vector register appears at different positions within two or more vector register groups, is reserved.

Example test case:
```li x5, 1                                          # Load desired vl value
vsetvli x0, x5, e32, m1, tu, mu
la x4, vd_random_base_097                         # Load address of desired value
vle32.v v24, (x4)                                 # Load desired value from memory into v24
la x31, vs2_random_base_097                       # Load address of desired value
vle32.v v1, (x31)                                 # Load desired value from memory into v1
la x30, vs1_random_base_097                       # Load address of desired value
vle32.v v1, (x30)                                 # Load desired value from memory into v1
vrgatherei16.vv v24, v1, v1
```
Output from test case:
```mem[X,0x08000360C] -> 0x3A108A57
[2977] [M]: 0x8000360C (0x3A108A57) vrgatherei16.vv v20, v1, v1
CSR mstatus (0x300) <- 0x80006600
CSR mstatush (0x310) <- 0x00000000
v20 <- 0x0000000D0000000D0000000D00000000
CSR mstatus (0x300) <- 0x80006600
CSR mstatush (0x310) <- 0x00000000
CSR vstart (0x008) <- 0x00000000
```

jordancarlin:
~Likely the same issue as #1056.~

pmundkur:
#1056 is about source/dest overlaps, whereas this is about source/source overlaps with different EEWs.

jordancarlin:
> [#1056](https://github.com/riscv/sail-riscv/issues/1056) is about source/dest overlaps, whereas this is about source/source overlaps with different EEWs.

Oops. Good catch.

pmundkur:
It looks like Spike has the same issue.  Fixing this causes some of the vector-tests to fail, because they were generated using Spike and it allows this overlap.

PR #1688, opened 2026-04-29
Fix a missed check of vector source/dest overlap.

pmundkur:
This is the case mentioned in #1056, which seems to have been missed in #1486.


Timmmm (approved):

nadime15 (approved):
Looks good to me. Would it be possible to add a test covering any of those overlap cases? I dont think we currently have any coverage for this.

Maybe this could also be handled in a separate PR. I think there are five cases to cover:

1. dest EEW > src EEW, overlap at high end of dst, src EMUL ≥ 1 (legal)
2. dest EEW > src EEW, overlap not at high end (illegal)
3. dest EEW < src EEW, overlap at low end of src (legal)
4. dest EEW < src EEW, overlap elsewhere (illegal)
5. dest EEW == src EEW, any overlap (legal)

pmundkur:
I'm hoping these tests will come out of ACT.  They've already found overlap issues like #1687; these tests would be similar.

jordancarlin:
> I'm hoping these tests will come out of ACT. They've already found overlap issues like #1687; these tests would be similar.

Yes. They will be covered by the SsstrictV test suite from ACT4 (not yet published).

jordancarlin (approved):

PR #1692, opened 2026-04-30
Optionally enable link-time optimization via CheckIPOSupported

nadime15:
Adds an opt-in `ENABLE_LTO` CMake option that allows the linker to aggressively inline functions into the generated model code, and compiles GMP with link-time optimization as well.

Off by default to keep incremental builds fast during local development. Enabled in the CI and release workflows.

Can reduce runtime by up to ~50% on Linux boot (from ~4m25s to ~2m11s, locally tested with GCC 15.2.0), with similar improvements observed across other workloads.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,12 @@ endif()
 option(DOWNLOAD_GMP "Download libgmp and build the library locally instead of using a system installation" OFF)
 if (DOWNLOAD_GMP)
     include(ExternalProject)
+    # Build GMP with LTO when ENABLE_LTO is on.
+    if (ENABLE_LTO)
+        set(GMP_CFLAGS "CFLAGS=-std=c11 -O3 -flto=auto -ffat-lto-objects")
+    else()
+        set(GMP_CFLAGS "CFLAGS=-std=c11")
+    endif()
     ExternalProject_Add(
         gmp
         URL https://ftp.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz https://gmplib.org/download/gmp/gmp-6.3.0.tar.xz https://ftpmirror.gnu.org/gnu/gmp/gmp-6.3.0.tar.xz
@@ -129,13 +135,14 @@ if (DOWNLOAD_GMP)
             --enable-static
             --with-pic
             # Newer compilers default to C23 which causes GMP build errors.
-            CFLAGS=-std=c11
+            "${GMP_CFLAGS}"
         # The libgmp.a output is created by the install step but since GMP::GMP
         # depends on the entire ExternalProject build, we can pretend it is
         # created by the build.
         # TODO: Once we depend on CMake 3.26, use INSTALL_BYPRODUCTS
         BUILD_BYPRODUCTS "<INSTALL_DIR>/lib/libgmp.a"
     )
+
     ExternalProject_Get_property(gmp INSTALL_DIR)
     add_library(GMP::GMP STATIC IMPORTED GLOBAL)
     # Work around `Imported target "GMP::GMP" includes non-existent path`.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,8 +121,8 @@ if (DOWNLOAD_GMP)
     # Build GMP with LTO when ENABLE_LTO is on, using the same toolchain and
     # IPO flags as the main project so the LTO bytecode is compatible.
     if (ENABLE_LTO)
-        string(REPLACE ";" " " GMP_LTO_CFLAGS "${CMAKE_C_COMPILE_OPTIONS_IPO}")
-        string(REPLACE ";" " " GMP_LTO_LDFLAGS "${CMAKE_C_LINK_OPTIONS_IPO}")
+        list(JOIN CMAKE_C_COMPILE_OPTIONS_IPO " " GMP_LTO_CFLAGS)
+        list(JOIN CMAKE_C_LINK_OPTIONS_IPO " " GMP_LTO_LDFLAGS)
         set(GMP_EXTRA_CONFIGURE_ARGS
             "CFLAGS=-std=c11 -O3 ${GMP_LTO_CFLAGS}"
             "LDFLAGS=${GMP_LTO_LDFLAGS}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,11 +118,19 @@ endif()
 option(DOWNLOAD_GMP "Download libgmp and build the library locally instead of using a system installation" OFF)
 if (DOWNLOAD_GMP)
     include(ExternalProject)
-    # Build GMP with LTO when ENABLE_LTO is on.
+    # Build GMP with LTO when ENABLE_LTO is on, using the same toolchain and
+    # IPO flags as the main project so the LTO bytecode is compatible.
     if (ENABLE_LTO)
-        set(GMP_CFLAGS "CFLAGS=-std=c11 -O3 -flto=auto -ffat-lto-objects")
+        string(REPLACE ";" " " GMP_LTO_CFLAGS "${CMAKE_C_COMPILE_OPTIONS_IPO}")
+        string(REPLACE ";" " " GMP_LTO_LDFLAGS "${CMAKE_C_LINK_OPTIONS_IPO}")
+        set(GMP_EXTRA_CONFIGURE_ARGS
+            "CFLAGS=-std=c11 -O3 ${GMP_LTO_CFLAGS}"
+            "LDFLAGS=${GMP_LTO_LDFLAGS}"
+            "CC=${CMAKE_C_COMPILER}"
+            "CXX=${CMAKE_CXX_COMPILER}"
+        )
     else()
-        set(GMP_CFLAGS "CFLAGS=-std=c11")
+        set(GMP_EXTRA_CONFIGURE_ARGS "CFLAGS=-std=c11")
     endif()
     ExternalProject_Add(
         gmp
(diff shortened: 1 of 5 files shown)


pmundkur (approved):
Nice!  LGTM.

jordancarlin (approved):

Timmmm:
Do we want to do this unconditionally? LTO has a pretty bad effect on compile times in my experience. We should probably only enable it `if (CMAKE_BUILD_TYPE STREQUAL Release OR CMAKE_BUILD_TYPE STREQUAL RelWithDebInfo)`.

Timmmm (changes requested):
Yeah I did a quick test and as I suspected it's pretty bad. A one-line change to `riscv_sim.cpp` gives these incremental compile times:

Before (already not great!):

```
real    0m14.787s
user    0m14.104s
sys     0m0.799s
```

After:

```
real    1m31.810s
user    1m52.201s
sys     0m4.202s
```

This is with `RelWithDebInfo` with Clang 20 on Linux on a fairly beefy desktop.

I think we need to either:

1. Only enable it for Release/RelWithDebInfo builds.
2. Give it its own flag, and enable it in the binary release Github workflow.

I normally work with the `RelWithDebInfo` config, and it is the default in `build_simulator.sh`, so I think I would prefer the latter.

In a previous PR I had an `extra_fast.cmake` file that was enabled by a flag and also did things like `-march=x86-64-v3`, maybe `-fomit-frame-pointer`, etc. I'll see if I can find it.

Timmmm:
> I'll see if I can find it.

Hmm doesn't look like I ever published it, but it was very short so never mind.

nadime15:
Good point. I tested this locally (and noticed minimal difference, 18 minutes :cry:) and checked the CI runs. LTO seems to add ~2–3 minutes to build time for some CI jobs, but overall some CI Jobs are up to ~15 minutes faster once test runtime is included (i.e. [before](https://github.com/riscv/sail-riscv/actions/runs/25184023914/job/73836442613) and [after](https://github.com/riscv/sail-riscv/actions/runs/25181183724/job/73826548318?pr=1692)). (The linux job does not seem to be effected at all [before](https://github.com/riscv/sail-riscv/actions/runs/25184023914/job/73836442512) and [after](https://github.com/riscv/sail-riscv/actions/runs/25181183724/job/73826548184?pr=1692))

What do you think about keeping the flag optional, but enabling it for release builds and for CI jobs that build and run the tests (instead of just enabling it for release builds), since the small build time overhead is offset by faster test execution?

Timmmm:
I would probably benchmark on a local machine rather than relying on Github CI VMs. They're running on machines with lots of other simultaneous jobs so the runtime is probably *very* noisy. But if it makes the overall jobs faster I think it makes sense.

I still don't think it should be enabled by default though. We could enable `-flto=thin` by default for Clang though. That doesn't affect compile times as badly IIRC. But it doesn't seem like CMake provides an abstracted interface for that, and it's Clang-only. Apparently GCC has something similar called `-fwhopr` but I've never used it.

nadime15:
I usually benchmark locally and via CI. The 35% (the one mentioned in the PR message) number was measured locally, not from CI.

> I still don't think it should be enabled by default though.

I don't know if you have seen it but its disabled by default now. I just enabled it in CI. See https://github.com/riscv/sail-riscv/commit/8fa894bbaba6cf8438c118970c798c506a608e1f

Alasdair:
If there are some small functions in the Sail library that benefit a lot from inlining we could move the bodies into the header and mark them as `static inline` so they get inlined without LTO.

nadime15:
So I was playing around a bit and gathered some stats.

Some base assumptions: I boot Linux as defined in os-boot/linux. I also switched my laptop to performance mode (I forgot it was on power saving before ...), so the numbers will look a bit different.

My Compiler: GCC 15.2.0

| Configuration                                      | Compile Time | Linux Boot Time |
|---------------------------------------------------|-------------|-----------------|
| Base                                              | 7m29s   | 4m25s           |
| Base + Inline                                     | 7m24s       | 4m10s               |
|Base + LTO                                               | 6m18s       | 3m47s           |
|Base + LTO + GMP CFLAGS="-std=c11 -O3 -fno-plt" ( and no --with-pic) | X    | 3m36s           |

Notes:
1. Regarding inlining, I tried @Alasdair’s suggestion and inlined a few selected functions into sail.h (eq_bits, neq_bits, bitvector_access, bitvector_access_inc).
2. I think the numbers in my PR message actually came from Base + LTO + modified CFLAGS. I also changed the instruction limit for Linux boot from 20M to 10M, so I need to fix the PR message.

For some reason, compiling with LTO is even faster than without. I would have expected the opposite, so this probably depends on the compiler and its version.

For inlining, I suspect that if I inline more and more functions, I could get closer to LTO performance? Interestingly, I think a number of functions in sail.c could be rewritten (removing some for loops) to improve performance. I tried that yesterday and saw a 5–10% improvement (assuming the replacements are equivalent).

We could probably get even better performance by tuning the CFLAGS for GMP. I could add that as well, assuming there are no side effects I am aware of.

IMHO I would add LTO as an optional feature (disabled by default) like discussed. It’s a simple change and gives an easy win locally (at least on my machine, as well as in CI). Later, we could experiment more with inlining and rewriting functions, and potentially drop LTO if people prefer that. Thoughts?

arichardson:
Would it be possible to forward the LTO flags to the bundled libgmp build? I imagine this could result in even more significant speedups.

nadime15:
Ok wow! Good idea @arichardson, that’s a significant additional speedup.

Here are the stats:

| Configuration               | Linux Boot Time   |
|-----------------------------|--------|
| Base                    | 4m25s  |
| Base +  LTO (simulator only) | 3m34s (~20% improvement) |
| Base + LTO (simulator + GMP) | **2m11s** (~50% improvement) |

Compile time is around the same as Base + LTO (see the comment above). 

For fun I run Linux boot time with PMP disabled and the runtime dropped to 1m42.

CMakeLists.txt
```diff
@@ -102,6 +118,12 @@ endif()
 option(DOWNLOAD_GMP "Download libgmp and build the library locally instead of using a system installation" OFF)
 if (DOWNLOAD_GMP)
     include(ExternalProject)
+    # Build GMP with LTO when ENABLE_LTO is on.
+    if (ENABLE_LTO)
+        set(GMP_CFLAGS "CFLAGS=-std=c11 -O3 -flto=auto -ffat-lto-objects")
```
arichardson: This will not work when building with clang.

A quick search suggested to me that we can use `CMAKE_{C,CXX}_COMPILE_OPTIONS_IPO` to get the flags that e.g. Clang https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Compiler/Clang.cmake#L79 or GCC https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Compiler/GNU.cmake#L109 pass normally. Similary, we should add `LDFLAGS=${CMAKE_C_LINK_OPTIONS_IPO}`.

With this change we should also be able to pass `CC=${CMAKE_C_COMPILER}` and `CXX=${CMAKE_CXX_COMPILER}`, to ensure that building with clang gets compatible LTO object files if we are building sail with clang.

CMakeLists.txt
```diff
@@ -113,13 +135,14 @@ if (DOWNLOAD_GMP)
             --enable-static
             --with-pic
             # Newer compilers default to C23 which causes GMP build errors.
-            CFLAGS=-std=c11
+            "${GMP_CFLAGS}"
```
arichardson:
```suggestion
            "${GMP_CFLAGS}"
            "CC=${CMAKE_C_COMPILER}"
            "CXX=${CMAKE_CXX_COMPILER}"
```

CMakeLists.txt
arichardson: But since those flags exist unconditionally, we should only pass them with `ENABLE_LTO`.

arichardson (approved):
Small comments on the GMP integration, but this is great!

CMakeLists.txt
nadime15: Good point, thanks! I had initially only tested this locally with GCC. I have now verified it with both GCC and Clang, and it works see: https://github.com/riscv/sail-riscv/pull/1692/changes/d70ae17c49457b5361c972e0d5b25044f43ce91d.

It’s actually even faster with Clang. The baseline is around 4 minutes, and with LTO enabled it drops to about 1m42s, roughly a ~57% reduction in runtime (linux boot)!

arichardson (approved):

pmundkur (approved):

CMakeLists.txt
```diff
@@ -102,6 +118,20 @@ endif()
 option(DOWNLOAD_GMP "Download libgmp and build the library locally instead of using a system installation" OFF)
 if (DOWNLOAD_GMP)
     include(ExternalProject)
+    # Build GMP with LTO when ENABLE_LTO is on, using the same toolchain and
+    # IPO flags as the main project so the LTO bytecode is compatible.
+    if (ENABLE_LTO)
+        string(REPLACE ";" " " GMP_LTO_CFLAGS "${CMAKE_C_COMPILE_OPTIONS_IPO}")
+        string(REPLACE ";" " " GMP_LTO_LDFLAGS "${CMAKE_C_LINK_OPTIONS_IPO}")
```
Timmmm: Minor but it's a bit cleaner to do this like this:


```suggestion
        list(JOIN CMAKE_C_COMPILE_OPTIONS_IPO " " GMP_LTO_CFLAGS)
        list(JOIN CMAKE_C_LINK_OPTIONS_IPO " " GMP_LTO_LDFLAGS)
```

Timmmm (approved):
One minor comment but otherwise LGTM, thanks!

PR #1696, opened 2026-05-05
Add configuration for the alignment of the bases of mtvec and stvec.

### LINKED ISSUE #1693: Request xtvec alignment parameters
I need parameters to describe the alignment of xtvec per the spec below.

UDB has two MTVEC parameters that I suggest we use.  It does not yet have STVEC.
It appears that STVEC_BASE_ALIGNMENT_DIRECT is always 4, so I think we only need
to add STVEC_BASE_ALIGNMENT_VECTORED.  UDB expresses the alignment in bytes
(e.g. MTVEC_BASE_ALIGNMENT_DIRECT = 64) and it would be convenient for Sail to
use the same units.  Thus I suggest:

MTVEC_BASE_ALIGNMENT_DIRECT
MTVEC_BASE_ALIGNMENT_VECTORED
STVEC_BASE_ALIGNMENT_VECTORED

mtvec
```
The value in the BASE field must always be
aligned on a 4-byte boundary, and the MODE setting may impose additional alignment constraints on the
value in the BASE field.
An implementation may have different alignment constraints for different modes. In particular,
MODE=Vectored may have stricter alignment constraints than MODE=Direct.
```

stvec
```
The BASE field in stvec is a field that can hold any valid virtual or physical address, subject to the
following alignment constraints: the address must be 4-byte aligned, and MODE settings other than Direct
might impose additional alignment constraints on the value in the BASE field.
```

pmundkur:
The constraints for the direct and vectored modes can be specified independently for mtvec, as powers of 2 upto a maximum of 24.  2^24 should be a reasonable maximum across most RV32 and RV64 implementations. The alignment can similarly be specified for the vectored mode of stvec.

Fixes #1693.

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -8,9 +8,8 @@
     by the same hart to the reservation set has been added; see
     `platform.reservation.invalidate_on_same_hart_store`. This defaults
     to false to match previous behavior.
-  - The mode-dependent alignment constraints of the bases of the
-    `mtvec` and `stvec` CSRs can now be specified; see
-    `base.{m,s}tvec.base_alignment`.
+  - The alignment constraints of the bases of the `mtvec` and `stvec`
+    CSRs can now be specified; see `base.{m,s}tvec.base_alignment`.
 
 # Release notes for version 0.11
 
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -8,6 +8,9 @@
     by the same hart to the reservation set has been added; see
     `platform.reservation.invalidate_on_same_hart_store`. This defaults
     to false to match previous behavior.
+  - The mode-dependent alignment constraints of the bases of the
+    `mtvec` and `stvec` CSRs can now be specified; see
+    `base.{m,s}tvec.base_alignment`.
 
 # Release notes for version 0.11
 
--- a/model/exceptions/sys_exceptions.sail
+++ b/model/exceptions/sys_exceptions.sail
@@ -71,12 +71,12 @@ function get_stvec() -> xlenbits =
   stvec.bits
 
 function set_mtvec(value : xlenbits) -> xlenbits = {
-  mtvec = legalize_tvec(mtvec, value, Mtvec);
+  mtvec = legalize_tvec(mtvec, value, plat_mtvec_base_alignment_direct_exp, plat_mtvec_base_alignment_vectored_exp);
   mtvec.bits
 }
 
 function set_stvec(value : xlenbits) -> xlenbits = {
-  stvec = legalize_tvec(stvec, value, Stvec);
+  stvec = legalize_tvec(stvec, value, 2, plat_stvec_base_alignment_vectored_exp);
   stvec.bits
 }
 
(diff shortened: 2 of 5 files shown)


model/exceptions/sys_exceptions.sail
```diff
@@ -71,12 +71,12 @@ function get_stvec() -> xlenbits =
...
 
 function set_mtvec(value : xlenbits) -> xlenbits = {
-  mtvec = legalize_tvec(mtvec, value);
+  mtvec = legalize_tvec(mtvec, value, Mtvec);
   mtvec.bits
 }
 
 function set_stvec(value : xlenbits) -> xlenbits = {
-  stvec = legalize_tvec(stvec, value);
+  stvec = legalize_tvec(stvec, value, Stvec);
```
Timmmm: I think if it would be clearer just to pass in the alignment requirements...

```
function set_mtvec(value : xlenbits) -> xlenbits = {
  mtvec = legalize_tvec(mtvec, value, plat_mtvec_base_alignment_direct_exp, plat_mtvec_base_alignment_vectored_exp);
  mtvec.bits
}
```

Then the legalize function still doesn't need to care which one it is legalising.

doc/ChangeLog.md
```diff
@@ -8,6 +8,9 @@
     by the same hart to the reservation set has been added; see
     `platform.reservation.invalidate_on_same_hart_store`. This defaults
     to false to match previous behavior.
+  - The mode-dependent alignment constraints of the bases of the
+    `mtvec` and `stvec` CSRs can now be specified; see
+    `base.{m,s}tvec.base_alignment`.
```
jordancarlin: "mode-dependent" is implied by the CSRs and not really relevant to the Changelog entry.

```suggestion
  - The alignment constraints of the bases of the
    `mtvec` and `stvec` CSRs can now be specified; see
    `base.{m,s}tvec.base_alignment`.
```

allenjbaum:
My wish is that there be a single legalize function that takes
 - a field (not the entire register),
 - a value, and
 - a json derived description of what is allowed to be written (e.g.
mask_value pair, range, or enum list),
and just writes the field (or not if it's an illegal value and we haven't
implemented illegal->legal mapping functionality)
In that case, any time the legal values change based on some other CSR
state (in this case mtvec.mode) there is a call to the legalize function
for each
different format case.there is explicit code that selects which json
description is used, e.g.
  if (mtvec.mode == mvtec_direct) legalize_CSR_field(mtvec.base,
mtvec_base_direct_legalval)
  else
 legalize_CSR_field(mtvec.base, mtvec_base_vector_legalval).

Separately, there would be          legalize_CSR_field(mtvec.mode,
mtvec_mode_legalval).
So, the only special cases that need wrapping are those that have legal
values that change depending on other state.

Stvec is more complicated, because what is allowed to be written may also
depend on the current satp.mode as well (zero extend if bare, sign extend
otherwise)
  I'm not sure if the sign/zero extension point is allowed to variable, is
simply the max(PA_width, SVxx),
or if only HW has to worry about value restrictions but SW can write as
many bits  as it has implemented (which must be supplied by  the vendor as
a parameter)

Timmmm (approved):

jordancarlin (approved):

PR #1697, opened 2026-05-05
Add riscv,cbo{m,z,p}-block-size properties to the device tree

nadime15:
The device tree generator lists Zicbom/Zicboz/Zicbop in riscv,isa, but omits the corresponding block-size properties, which causes Linux to disable these extensions at boot. This change emits the appropriate properties in the device tree when the extensions are supported, using 2^plat_cache_block_size_exp to match the instruction semantics.

At the same time add Zic64b to both the device tree ISA string and the extension list.

### DIFF
--- a/model/core/extensions.sail
+++ b/model/core/extensions.sail
@@ -658,6 +658,7 @@ let extensions_ordered_for_isa_string = [
 
   // Z extensions, ordered by category and then alphabetically.
   // Zi
+  Ext_Zic64b,
   Ext_Zibi,
   Ext_Zicbom,
   Ext_Zicbop,
--- a/model/postlude/device_tree.sail
+++ b/model/postlude/device_tree.sail
@@ -119,6 +119,15 @@ private function generate_dts_memories(pmas : list(PMA_Region)) -> string =
     }
   }
 
+// Generates a `riscv,<prefix>-block-size` property line if the corresponding
+// cache-block extension is supported. The size is 2^plat_cache_block_size_exp
+// bytes, the same value used by the Zicbom/Zicboz/Zicbop instruction semantics.
+private function generate_cbo_block_size(prefix : string, ext : extension) -> string = {
+  if hartSupports(ext)
+  then "      riscv," ^ prefix ^ "-block-size = <" ^ dec_str(2 ^ plat_cache_block_size_exp) ^ ">;\n"
+  else ""
+}
+
 function generate_dts() -> string = {
   let clock_freq : nat = config platform.clock_frequency;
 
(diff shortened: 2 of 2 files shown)


model/postlude/device_tree.sail
```diff
@@ -156,6 +165,9 @@ function generate_dts() -> string = {
 ^ "      riscv,isa = \"" ^ generate_isa_string(Canonical_Lowercase) ^ "\";\n"
 ^ "      riscv,isa-extensions = " ^ generate_isa_string(DeviceTree_ISA_Extensions) ^ ";\n"
 ^ "      mmu-type = \"riscv," ^ mmu_type() ^ "\";\n"
+^ generate_cbo_block_size("cbom", Ext_Zicbom)
+^ generate_cbo_block_size("cboz", Ext_Zicboz)
+^ generate_cbo_block_size("cbop", Ext_Zicbop)
```
pmundkur, line 170: The kernel treating them as independent parameters indicates we should similarly have separate configurable cache block sizes for these extensions, and Zic64b support just ensures that all of them are 64b.

nadime15: If I understand correctly, you are saying each extension can have its own set of parameters. If so, I agree, that makes sense, and I am aware of that. If you would prefer to introduce three separate parameters, I can do this in a follow-up PR or do you wanna handle this in this PR?

pmundkur: Yes, I meant a per-extension cache-block size for Zicbo{m,p,z}.  Your choice re: PRs, but it seems pretty straightforward to me, might as well do it in this PR, since this PR is just a few lines anyway.

nadime15: Sure, will do.

pmundkur (approved):
LGTM, thanks!

I'm wondering if we should do anything special if Zic64b is supported but none of the Zicbo* extensions are.   But this is good for now.

Timmmm (changes requested):
I don't understand this. These instructions are operating on the same cache blocks, so how could they be different sizes? 

> A cache block is identified by any of the physical addresses corresponding to the underlying memory locations.

So how would you have different cache blocks sizes for the same address?

Also the spec currently explicitly prohibits different sizes so this seems like it's complicating things for no benefit?

> In the initial set of CMO extensions, **the size of a cache block shall be uniform throughout the system**.

pmundkur:
> Also the spec currently explicitly prohibits different sizes so this seems like it's complicating things for no benefit?
> 
> > In the initial set of CMO extensions, **the size of a cache block shall be uniform throughout the system**.

Ah, I missed that.  My suggestion was based on device tree having three different bindings for these cache sizes.

nadime15:
Ahh good catch. I reverted it to the original commit.

model/core/extensions.sail
```diff
@@ -658,6 +658,7 @@ let extensions_ordered_for_isa_string = [
 
   // Z extensions, ordered by category and then alphabetically.
   // Zi
+  Ext_Zic64b,
```
Timmmm, line 661: I think you accidentally removed the check that this enables?

nadime15: There is already a check in:

```
private function check_extension_param_constraints() -> bool = {
  var valid : bool = true;
  if (config extensions.Zic64b.supported : bool) & (plat_cache_block_size_exp != 6) then {
    valid = false;
    print_endline("The Zic64b extension is enabled but the cache block size is not 64 bytes.");
  };
...
```

Timmmm: Oh I see, I misread this code. LGTM.

Timmmm (approved):

jordancarlin (approved):

pmundkur (approved):

PR #1698, opened 2026-05-06
Add checks to ensure valid overlaps between vector source register groups.

### LINKED ISSUE #1687: vrgatherei16.vv allowing reserved behavior vs1=vs2 at SEW=32/64

pmundkur:
The spec requires:

> An encoding that would result in the same vector register being read with
> two or more different EEWs, including when the vector register appears at
> different positions within two or more vector register groups, is reserved.

These checks include overlaps between a source register group and the mask vector v0 when masking is in effect.

Currently, the reserved encoding results in an illegal instruction.  This reserved behavior could be parameterized like other reserved behavior in a subsequent PR.

This should fix #1687.

### DIFF
--- a/model/extensions/V/vext_arith_insts.sail
+++ b/model/extensions/V/vext_arith_insts.sail
@@ -59,18 +59,29 @@ function clause execute VVTYPE(funct6, vm, vs2, vs1, vd) = {
   // source vector register groups, otherwise the instruction encoding
   // is reserved."
 
+  let vs1_emul_pow = if funct6 == VV_VRGATHEREI16 then 4 + LMUL_pow - SEW_pow else LMUL_pow;
+
   // Check for source destination overlaps for vrgather instructions.
-  let src_dst_overlap : bool = if funct6 == VV_VRGATHER | funct6 == VV_VRGATHEREI16
+  let src_dst_overlap : bool = if funct6 == VV_VRGATHER
                                then vs1 == vd | vs2 == vd
+                               else if funct6 == VV_VRGATHEREI16
+                               then not(valid_disjoint_reg_groups(vs1, vd, vs1_emul_pow, LMUL_pow)) | vs2 == vd
                                else false;
 
-  let vs1_emul_pow = if funct6 == VV_VRGATHEREI16 then 4 + LMUL_pow - SEW_pow else LMUL_pow;
+  let src_mask_overlap = not(valid_vm_src_overlap(vm, vs1, SEW)) |
+                         not(valid_vm_src_overlap(vm, vs2, SEW));
+
+  let src_src_overlap : bool = if funct6 == VV_VRGATHEREI16
+                               then not(valid_src_overlap(vs1, 16, vs1_emul_pow, vs2, SEW, LMUL_pow))
+                               else false;
 
   if illegal_vstart(VSTART_ARITH) |
      illegal_normal(vd, vm) |
      not(valid_reg_overlap(vs1, vd, vs1_emul_pow, LMUL_pow)) |
      not(valid_reg_group(vs2, LMUL_pow)) |
-     src_dst_overlap
+     src_dst_overlap |
+     src_src_overlap |
+     src_mask_overlap
   then return Illegal_Instruction();
 
   let 'n = num_elem;
@@ -217,7 +228,9 @@ function clause execute NVSTYPE(funct6, vm, vs2, vs1, vd) = {
   if illegal_vstart(VSTART_ARITH) |
      illegal_variable_width(vd, vm, SEW_widen, LMUL_pow_widen) |
      not(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow)) |
-     not(valid_reg_group(vs1, LMUL_pow))
+     not(valid_src_overlap(vs2, SEW_widen, LMUL_pow_widen, vs1, SEW, LMUL_pow)) |
+     not(valid_vm_src_overlap(vm, vs1, SEW)) |
+     not(valid_vm_src_overlap(vm, vs2, SEW_widen))
   then return Illegal_Instruction();
 
   assert(SEW_widen <= 64);
@@ -293,7 +306,9 @@ function clause execute NVTYPE(funct6, vm, vs2, vs1, vd) = {
   if illegal_vstart(VSTART_ARITH) |
      illegal_variable_width(vd, vm, SEW_widen, LMUL_pow_widen) |
      not(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow)) |
-     not(valid_reg_group(vs1, LMUL_pow))
+     not(valid_src_overlap(vs2, SEW_widen, LMUL_pow_widen, vs1, SEW, LMUL_pow)) |
+     not(valid_vm_src_overlap(vm, vs1, SEW)) |
+     not(valid_vm_src_overlap(vm, vs2, SEW_widen))
   then return Illegal_Instruction();
 
   assert(SEW_widen <= 64);
@@ -368,7 +383,9 @@ function clause execute MASKTYPEV(vs2, vs1, vd) = {
      illegal_vd_masked(vd)  |
      not(valid_reg_group(vs1, LMUL_pow)) |
      not(valid_reg_group(vs2, LMUL_pow)) |
-     not(valid_reg_group(vd, LMUL_pow))
+     not(valid_reg_group(vd, LMUL_pow)) |
+     not(valid_vm_src_overlap(0b0, vs1, SEW)) |
+     not(valid_vm_src_overlap(0b0, vs2, SEW))
   then return Illegal_Instruction();
 
   let 'n = num_elem;
@@ -485,7 +502,8 @@ function clause execute VXTYPE(funct6, vm, vs2, rs1, vd) = {
   if illegal_vstart(VSTART_ARITH) |
      illegal_normal(vd, vm) |
      not(valid_reg_group(vs2, LMUL_pow)) |
-     not(valid_reg_group(vd, LMUL_pow))
+     not(valid_reg_group(vd, LMUL_pow)) |
+     not(valid_vm_src_overlap(vm, vs2, SEW))
   then return Illegal_Instruction();
 
   let 'n = num_elem;
@@ -611,7 +629,8 @@ function clause execute NXSTYPE(funct6, vm, vs2, rs1, vd) = {
 
   if illegal_vstart(VSTART_ARITH) |
      illegal_variable_width(vd, vm, SEW_widen, LMUL_pow_widen) |
-     not(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow))
+     not(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow)) |
+     not(valid_vm_src_overlap(vm, vs2, SEW_widen))
   then return Illegal_Instruction();
 
   assert(SEW_widen <= 64);
@@ -686,7 +705,8 @@ function clause execute NXTYPE(funct6, vm, vs2, rs1, vd) = {
 
   if illegal_vstart(VSTART_ARITH) |
      illegal_variable_width(vd, vm, SEW_widen, LMUL_pow_widen) |
-     not(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow))
+     not(valid_reg_overlap(vs2, vd, LMUL_pow_widen, LMUL_pow)) |
+     not(valid_vm_src_overlap(vm, vs2, SEW_widen))
   then return Illegal_Instruction();
 
   assert(SEW_widen <= 64);
@@ -772,6 +792,7 @@ function clause execute VXSG(funct6, vm, vs2, rs1, vd) = {
      illegal_normal(vd, vm) |
      not(valid_reg_group(vs2, LMUL_pow)) |
      not(valid_reg_group(vd, LMUL_pow)) |
+     not(valid_vm_src_overlap(vm, vs2, SEW)) |
      src_dst_overlap
   then return Illegal_Instruction();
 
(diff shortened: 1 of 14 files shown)


pmundkur:
This breaks several tests in `riscv-vector-tests` due to the generation of reserved behavior, around a 110 tests per VLEN/ELEN combination.  I'll file an issue for it there.

model/extensions/V/vext_utils_insts.sail
```diff
@@ -85,6 +85,48 @@ function valid_reg_overlap(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_p
...
+  then return false;
+
+  rs_int + rs_group <= rd_int | rs_int >= rd_int + rd_group
+}
+
+// Check for valid overlaps between source register groups: overlaps
+// are not allowed if the EEWs are different for the two groups; a
+// mask register is considered to have EEW of 1. This also checks if
+// the register groups are valid.
+function valid_src_overlap(rs1 : vregidx, EEW_rs1 : nat, EMUL_pow_rs1 : int,
```
nadime15, line 137: Arent there tons of functions that need to check for source and destination register overlap? So maybe we keep it just as `valid_overlap`?

model/extensions/V/vext_utils_insts.sail
```diff
@@ -85,6 +85,48 @@ function valid_reg_overlap(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_p
...
+  let rs2_group = if EMUL_pow_rs2 > 0 then 2 ^ EMUL_pow_rs2 else 1;
+  let rs1_int = unsigned(vregidx_bits(rs1));
+  let rs2_int = unsigned(vregidx_bits(rs2));
+
+  if rs1_int % rs1_group != 0 | rs2_int % rs2_group != 0
+  then return false;
+
+  if EEW_rs1 == EEW_rs2 then return true;
+
+  rs1_int + rs1_group <= rs2_int | rs2_int >= rs2_int + rs2_group
```
nadime15: Did you mean the following?
```suggestion
  rs1_int + rs1_group <= rs2_int | rs1_int >= rs2_int + rs2_group
```
Because otherwise its always true.

I think the following is even easier to read:

`  rs1_int + rs1_group <= rs2_int | rs2_int + rs2_group <= rs1_int`

model/extensions/V/vext_utils_insts.sail
pmundkur: Source/dest overlap checks involve the respective EMULs when they could differ; source/source overlaps involve EEWs when they could differ.  I think it might be clearer to keep them separate?  Or perhaps I didn't understand your point...

model/extensions/V/vext_utils_insts.sail
pmundkur: Oops, I'll fix it, thanks.

model/extensions/V/vext_utils_insts.sail
nadime15: `valid_disjoint_reg_groups` and `valid_src_overlap` have basically the same logic. The only difference is that `valid_src_overlap` additionally checks the case EEW_rs1 == EEW_rs2. Could `valid_src_overlap` just call `valid_disjoint_reg_groups`?

`valid_disjoint_reg_groups(rs1, rs2, EMUL_pow_rs1, EMUL_pow_rs2) | EEW_rs1 == EEW_rs2`

Besides that, all of these functions (valid_reg_overlap, valid_src_overlap, and valid_disjoint_reg_groups) share partially the same logic:

```
  let rs_group = if EMUL_pow_rs > 0 then 2 ^ EMUL_pow_rs else 1;
  let rd_group = if EMUL_pow_rd > 0 then 2 ^ EMUL_pow_rd else 1;
  let rs_int = unsigned(vregidx_bits(rs));
  let rd_int = unsigned(vregidx_bits(rd));

  if rs_int % rs_group != 0 | rd_int % rd_group != 0
  then return false;
```
  
I think this deserves its own helper function because at least for me, it makes these functions much harder to parse and understand. When reading them, I initially assume there are important differences, so I end up comparing them line by line, only to realize that this entire section is identical in all of them.

pmundkur: > `valid_disjoint_reg_groups` and `valid_src_overlap` have basically the same logic. The only difference is that `valid_src_overlap` additionally checks the case EEW_rs1 == EEW_rs2. Could `valid_src_overlap` just call `valid_disjoint_reg_groups`?
> 
> `valid_disjoint_reg_groups(rs1, rs2, EMUL_pow_rs1, EMUL_pow_rs2) | EEW_rs1 == EEW_rs2`

This won't be correct: even if `EEW_rs1 == EEW_rs2`,  `rs1` and `rs2` still have to be aligned for their EMUL otherwise they are not valid.

> Besides that, all of these functions (valid_reg_overlap, valid_src_overlap, and valid_disjoint_reg_groups) share partially the same logic:
> 
> ```
>   let rs_group = if EMUL_pow_rs > 0 then 2 ^ EMUL_pow_rs else 1;
>   let rd_group = if EMUL_pow_rd > 0 then 2 ^ EMUL_pow_rd else 1;
>   let rs_int = unsigned(vregidx_bits(rs));
>   let rd_int = unsigned(vregidx_bits(rd));
> 
>   if rs_int % rs_group != 0 | rd_int % rd_group != 0
>   then return false;
> ```
> 
> I think this deserves its own helper function because at least for me, it makes these functions much harder to parse and understand. When reading them, I initially assume there are important differences, so I end up comparing them line by line, only to realize that this entire section is identical in all of them.

I agree, these checkers could have this common code factored out.  But that is unrelated to the issue this PR is addressing though, so I'll do it in a followup one.

model/extensions/V/vext_utils_insts.sail
```diff
@@ -85,6 +85,48 @@ function valid_reg_overlap(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_p
...
+function valid_disjoint_reg_groups(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_pow_rd : int) -> bool = {
+  let rs_group = if EMUL_pow_rs > 0 then 2 ^ EMUL_pow_rs else 1;
+  let rd_group = if EMUL_pow_rd > 0 then 2 ^ EMUL_pow_rd else 1;
+  let rs_int = unsigned(vregidx_bits(rs));
+  let rd_int = unsigned(vregidx_bits(rd));
+
+  if rs_int % rs_group != 0 | rd_int % rd_group != 0
+  then return false;
+
+  rs_int + rs_group <= rd_int | rs_int >= rd_int + rd_group
```
nadime15:
```suggestion
  rs_int + rs_group <= rd_int | rd_int + rd_group <= rs_int
```

model/extensions/V/vext_utils_insts.sail
```diff
@@ -85,6 +85,48 @@ function valid_reg_overlap(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_p
   } else true;
 }
 
+function valid_disjoint_reg_groups(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_pow_rd : int) -> bool = {
+  let rs_group = if EMUL_pow_rs > 0 then 2 ^ EMUL_pow_rs else 1;
+  let rd_group = if EMUL_pow_rd > 0 then 2 ^ EMUL_pow_rd else 1;
```
nadime15:
```suggestion
  let rs_group = 2 ^ max(EMUL_pow_rs, 0);
  let rd_group = 2 ^ max(EMUL_pow_rd, 0);
```

Same for `valid_src_overlap`

model/extensions/V/vext_utils_insts.sail
nadime15: Ah yeah you are right. 

I think something like this would be nice and should be equivalent: 
```
function aligned(r : vregidx, EMUL_pow : int) -> bool =
  unsigned(vregidx_bits(r)) % (2 ^ max(EMUL_pow, 0)) == 0

function disjoint(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_pow_rd : int) -> bool = {
  let rs_group = 2 ^ max(EMUL_pow_rs, 0);
  let rd_group = 2 ^ max(EMUL_pow_rd, 0);
  let rs_int = unsigned(vregidx_bits(rs));
  let rd_int = unsigned(vregidx_bits(rd));
  rs_int + rs_group <= rd_int | rd_int + rd_group <= rs_int
}

function valid_disjoint_reg_groups(rs, rd, EMUL_pow_rs, EMUL_pow_rd) -> bool =
  aligned(rs, EMUL_pow_rs) & aligned(rd, EMUL_pow_rd) & disjoint(rs, rd, EMUL_pow_rs, EMUL_pow_rd)

function valid_src_overlap(rs1, EEW_rs1, EMUL_pow_rs1, rs2, EEW_rs2, EMUL_pow_rs2) -> bool =
  aligned(rs1, EMUL_pow_rs1) & aligned(rs2, EMUL_pow_rs2)
  & (EEW_rs1 == EEW_rs2 | disjoint(rs1, rs2, EMUL_pow_rs1, EMUL_pow_rs2))
```

I would prefer to this in the PR because you touch this anyway (and actually introduces it to the codebase, so why not get it right away from the start) and it makes it easier to understand but your call.

pmundkur: The problem here is that there would be repeated computations of `unsigned(vregidx_bits(*))` and `(2 ^ max(EMUL_*, 0))`  for `rs1` and `rs2` in every call to `aligned()` and `disjoint()`.   So the succinctness hides some inefficiency.  We should refactor to something that is both succinct and efficient.

nadime15: I dont think this is really a deal breaker since this is not part of the hot path, and the instructions themselves seem quite straightforward and wont add much overhead. So this inefficiency is basically negligible in the grand scheme of things, but we can keep it as it is.

model/extensions/V/vext_arith_insts.sail
```diff
@@ -1976,11 +2014,11 @@ function clause execute MVVCOMPRESS(vs2, vs1, vd) = {
 
   // vcompress should always be executed with a vstart of 0
   if start_element != 0 |
-     vs1 == vd |
+     not(valid_disjoint_reg_groups(vs1, vd, 1, LMUL_pow)) |
```
nadime15: I think vs1 is the mask here (consisting of only one register), so we would need to pass the exponent of a power of two. So 2^1, the exponent would be 0.
```suggestion
     not(valid_disjoint_reg_groups(vs1, vd, 0, LMUL_pow)) |
```

model/extensions/V/vext_mem_insts.sail
```diff
@@ -368,14 +370,16 @@ mapping clause encdec = VLXSEGTYPE(nf, vm, vs2, rs1, width, vd, mop)
...
   let EEW_index_bytes = 2 ^ (EEW_index_pow - 3);
   let EEW_data_pow = get_sew_pow();
   let EEW_data_bytes = get_sew_bytes();
   let EMUL_data_pow = get_lmul_pow();
   let EMUL_index_pow = EEW_index_pow - EEW_data_pow + EMUL_data_pow;
 
   if illegal_indexed_load(vd, vm, nf, 2 ^ EEW_index_pow, EMUL_index_pow, EMUL_data_pow) |
-     not(valid_reg_overlap(vs2, vd, EMUL_index_pow, EMUL_data_pow))
+     not(valid_disjoint_reg_groups(vs2, vd, EMUL_index_pow, EMUL_data_pow)) |
+     not(valid_vm_src_overlap(vm, vs2, EEW_index))
```
nadime15, line 389: If i understand correctly non-segment indexed loads and segment indexed load have different rules

For vector indexed loads: 

> For vector indexed segment loads, the destination vector register groups cannot overlap the source vector register group (specified by vs2), else the instruction encoding is reserved.


and I think that non-segment indexed load is covered by the generel overlap rule 



> A destination vector register group can overlap a source vector register group only if one of the following holds:
> 
>     The destination EEW equals the source EEW.
> 
>     The destination EEW is smaller than the source EEW, and the lowest-numbered register in the destination vector register group is the same as the lowest-numbered register in the source vector register group. (For example, when LMUL=1, vnsrl.wi v0, v0, 3 is legal, but a destination of v1 is not).
> 
>     The destination EEW is greater than the source EEW, the source EMUL is at least 1, and the highest-numbered register in the destination vector register group is the same as the highest-numbered register in the source vector register group. (For example, when LMUL=8, vzext.vf4 v0, v6 is legal, but a source of v0, v2, or v4 is not).
> 
> For the purpose of determining register group overlap constraints, mask elements have EEW=1.

Something like this?
```suggestion
let index_dst_overlap : bool = if nf == 1
then not(valid_reg_overlap(vs2, vd, EMUL_index_pow, EMUL_data_pow)) // non-segment indexed load
else not(valid_disjoint_reg_groups(vs2, vd, EMUL_index_pow, EMUL_data_pow)); // segment indexed load
  
     if illegal_indexed_load(vd, vm, nf, 2 ^ EEW_index_pow, EMUL_index_pow, EMUL_data_pow) |
     index_dst_overlap |
     not(valid_vm_src_overlap(vm, vs2, EEW_index))
```

pmundkur: I think you're right, good catch!

nadime15 (approved):
Nice, looks good to me!

pmundkur:
I'll merge this later today unless someone suggests changes before then.

PR #1704, opened 2026-05-09
Fix building tests with clang < 21

arichardson:
The mcyclecfg csr name is not supported in older versions, so use the define from encoding.h instead.


jordancarlin (approved):
Might be worth using an older version of Clang in CI to avoid issues like this.

Timmmm (approved):

pmundkur (approved):

PR #1581, opened 2026-03-02
Add unratified Zvfbfa extension 

TinyuengKwan:
Implement the extension described in:
https://github.com/aswaterman/riscv-misc/blob/main/isa/zvfbfa.adoc

This commit introduces an initial implementation of the Zvfbfa extension. Note that this extension does not introduce new instructions; instead, it modifies the behavior and semantics of existing vector floating-point instructions when operating on the relevant element types. As a result, the changes are largely integrated into existing execution paths rather than implemented as separate opcodes.

Due to this behavioral nature of the extension, and the large number of affected instructions, there may be omissions or subtle semantic mismatches. I am not fully confident that all modified behaviors strictly conform to the specification. The current state should therefore be considered a draft and leaves room for further refinement and cleanup.

Future work will gradually add support for the extensions described in:
- https://github.com/aswaterman/riscv-misc/blob/main/isa/zvfofp4min.adoc
- https://github.com/aswaterman/riscv-misc/blob/main/isa/zvfofp8min.adoc

PR #1705, opened 2026-05-12
Add unratified Zvqwdota8i extension

rpsene:
Implement Phase 1 of the unratified Zvdota family of dot-product extensions, adding the two `Zvqwdota8i` instructions `vqwdotau.vv` and `vqwdotas.vv` (major opcode 0x77, funct3=000, funct6 0x26/0x27).

Spec: https://github.com/aswaterman/riscv-misc/blob/main/isa/ldot-bdot/ldot-bdot.adoc

The spec uses `vtype.altfmt` to select vs1 signedness, but `altfmt` is not yet in sail-riscv master — it depends on the Zvfbfa extension (#1581). This PR pins vs1 to unsigned (the spec's `altfmt=0` case) and leaves a `TODO(Zvfbfa)` marker for the follow-up.

Enabled via `--enable-experimental-extensions`. Includes a new first-party test covering both opcodes on rv32 and rv64.

---

### AI tool disclosure (per CONTRIBUTING.md)

This PR was produced with AI assistance (Anthropic's Claude). The work was human-directed: I selected the extension and scope (Phase 1 only, vs1 pinned to `altfmt=0`), chose the encoding-space approach (OP-VE at 0x77, funct3=000), identified Zvabd as the structural template, and diagnosed every build/runtime failure (Sail typecheck constraint propagation requiring an explicit `assert(SEW == 8)`, the missing `mstatus[VS]` enable that traps `vsetvli` as illegal, and the LLVM 18 `-march=…zfbfmin` quirk) against actual Docker builds — not hand-waved. The AI produced initial drafts of the Sail code, the assembly test, and the documentation under that direction; I reviewed every line before commit. All commits are DCO-signed off by me, and I accept authorial responsibility for the content.

### DIFF
--- a/model/extensions/Zvqwdota8i/zvqwdota8i_insts.sail
+++ b/model/extensions/Zvqwdota8i/zvqwdota8i_insts.sail
@@ -1,0 +1,104 @@
+// =======================================================================================
+// This Sail RISC-V architecture model, comprising all files and
+// directories except where otherwise noted is subject the BSD
+// two-clause license in the LICENSE file.
+//
+// SPDX-License-Identifier: BSD-2-Clause
+// =======================================================================================
+
+// Zvqwdota8i: Quad-widening 8-bit integer dot-product (unratified, v0.2).
+// Two instructions, both in major opcode 0x77 (OP-VE) with funct3=000:
+//   vqwdotau.vv  vd, vs2, vs1, vm   funct6=0x26   vs2 unsigned
+//   vqwdotas.vv  vd, vs2, vs1, vm   funct6=0x27   vs2 signed
+// Computes  vd[0] += sum_{i=vstart..vl-1} (mask[i] ? vs2[i] * vs1[i] : 0)
+// modulo 2^32.  vs1, vs2 are EEW=8, EMUL=LMUL; vd is EEW=32, EMUL=1, accumulator in vd[0].
+// Other elements of vd are tail.
+//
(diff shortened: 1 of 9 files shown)


model/extensions/Zvqwdota8i/zvqwdota8i_insts.sail
```diff
@@ -0,0 +1,104 @@
...
+  then return Illegal_Instruction();
+
+  // "The destination vector register cannot overlap with the source vector register groups."
+  let vd_int      = unsigned(vregidx_bits(vd));
+  let vs2_int     = unsigned(vregidx_bits(vs2));
+  let vs1_int     = unsigned(vregidx_bits(vs1));
+  let src_group_size : int = if LMUL_pow > 0 then 2 ^ LMUL_pow else 1;
+  if (vd_int >= vs2_int & vd_int < vs2_int + src_group_size) |
+     (vd_int >= vs1_int & vd_int < vs1_int + src_group_size)
+  then return Illegal_Instruction();
```
nadime15, line 56: I think you can shorten the code by doing the following:

```suggestion
 if illegal_reduction() |
   not(valid_reg_group(vs1, LMUL_pow)) |
   not(valid_reg_group(vs2, LMUL_pow)) |
   // "The destination vector register cannot overlap with the source vector register groups."
   vd == vs1 | vd == vs2
then return Illegal_Instruction();
```

Thats similar to what we do with vslide1up in vext_arith_insts.sail

PR #1706, opened 2026-05-12
Inline single-use `process*` helpers in `vext_mem_insts` into their corresponding execute clauses.

### LINKED ISSUE #1700: Inline process_* helpers in vext_mem_insts.sail
Each instruction in `vext_mem_insts.sail` has a private process_* function that is called exactly once from its execute clause (and never reused). Unless I miss something, inlining them would make each instruction self-contained and easier to read.

KotorinMinami:
This inlining makes the execute clauses more self-contained and easier to read.

Fixes #1700.


nadime15 (approved):
Looks good, thanks!

pmundkur (approved):

PR #1707, opened 2026-05-12
Refactor `inner_main` by splitting it into smaller initialization utilities.

pmundkur:
Simulator initialization and cli option processing is done in three steps, roughly:

- before model initialization (`preinit_args()`)
- configuring model options (`preinit_model()`)
- completing model initialization (`init_model()`)

`init_platform_constants()`, `{re}init_sail()` are moved into member functions of the model. The outer run loop is moved into `run_model()`.

Subsequent refactors can address reducing/eliminating the globals in `riscv_sim.cpp` and splitting out rvfi into a separate simulation loop.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -624,29 +624,32 @@ void init_logs(const CLIOptions &opts) {
 #endif
 }
 
+// Initialization result used during startup.
+enum class InitResult {
+  ExitFailure,
+  ExitSuccess,
+  Continue,
+};
+
 // Processes options that don't need an initialized model and gets the
 // json configuration string; returns whether to continue with model
 // initialization.
-bool preinit_args(CLIOptions &opts, int &exit_code, std::string &config_json_string) {
+InitResult preinit_args(CLIOptions &opts, std::string &config_json_string) {
   if (opts.do_print_version) {
     std::cout << version_info::release_version << std::endl;
-    exit_code = EXIT_SUCCESS;
-    return false;
+    return InitResult::ExitSuccess;
   }
   if (opts.do_print_build_info) {
     print_build_info();
-    exit_code = EXIT_SUCCESS;
-    return false;
+    return InitResult::ExitSuccess;
   }
   if (opts.do_print_default_config) {
     printf("%s", opts.use_rv32_default ? get_default_rv32_config() : get_default_config());
-    exit_code = EXIT_SUCCESS;
-    return false;
+    return InitResult::ExitSuccess;
   }
   if (opts.do_print_config_schema) {
     printf("%s", get_config_schema());
-    exit_code = EXIT_SUCCESS;
-    return false;
+    return InitResult::ExitSuccess;
   }
 
   if (opts.do_show_times) {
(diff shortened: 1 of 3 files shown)


pmundkur:
A more incremental approach to what #1683 is trying to do.

c_emulator/riscv_sim.cpp
```diff
@@ -652,31 +624,31 @@ void init_logs(const CLIOptions &opts) {
...
-  ModelImpl model;
-
+// Processes options that don't need an initialized model and gets the
+// json configuration string; returns whether to continue with model
+// initialization.
+bool preinit_args(CLIOptions &opts, int &exit_code, std::string &config_json_string) {
   if (opts.do_print_version) {
     std::cout << version_info::release_version << std::endl;
-    return EXIT_SUCCESS;
+    exit_code = EXIT_SUCCESS;
```
Timmmm: This is pretty awkward. I'd prefer if we use an enum instead like

```
enum class ProcessResult {
  ExitFailure,
  ExitSuccess,
  Continue,
};
```

Same for `preinit_model`.

Otherwise this LGTM.

nadime15 (commented):
Looks good! Should we move `load_sail()` to riscv_model_impl.cpp as well? @Timmmm @pmundkur

pmundkur:
> Looks good! Should we move `load_sail()` to riscv_model_impl.cpp as well? @Timmmm @pmundkur

I think eventually we may want to, but I'd hold off on it due to the calls to `exit` on load failure.  I have a branch that moves most of these functions to `sim_utils.{h,cpp}` so that it's easy to have multiple binaries each with their own `main()`.   That branch doesn't deal with such bare `exit()` calls yet; once we get rid of them, we could move this then.

nadime15:
So I took a look at load_sail, and you basically just need to replace the two exits with thrown exceptions. This could be something simple like what is already done in `elf_loader.cpp`:

`throw std::runtime_error("File '" + filename + "' is not found or it is not an ELF file");`

pmundkur:
That would work, but that would be the first `throw` from `ModelImpl`, which makes me leery.  What do you think @Timmmm ?

nadime15:
If had C++ 23 we could use expected similar to rusts Result<T, E> :D or maybe c++17 with optional. 

We need to refactor `load_sail` anyway like i did in #1683. And we could also extend this a bit more into something like

```
struct LoadResult {
    uint64_t entry;
    std::optional<uint64_t> htif_tohost;
    std::optional<uint64_t> sig_start;
    std::optional<uint64_t> sig_end;
    bool valid = true;
    std::string error_message;
};
```
And the caller would simply check whether the load was successful and then handle the result accordingly, which would avoid throwing an exception. There are many ways to approach this.

pmundkur:
>  There are many ways to approach this.

Agreed, which is why I proposed above to deal with this later, along with the other functions that have the same issue.

Timmmm (approved):

Timmmm:
Yeah I kind of hate exceptions... I've used [my own Rust-style Result in C++ before](https://github.com/Timmmm/robocut/blob/develop/src/Result.h) and it worked pretty nicely (especially if you use [statement expressions](https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html) for `try`... which doesn't work with MSVC unfortunately).

I'd be in favour of C++23. It is fairly recent in C++ terms, and I know a lot of hardware people use ancient RHEL/CentOS distros, but those have an easy way to install recent compilers so maybe it's fine?

pmundkur:
The Rocky container we use for our release has g++ 8.5, which is too old for C++23.   Even my main dev box has g++ 13.3 which doesn't support all of C++23 :-(

nadime15 (approved):

jordancarlin (approved):

PR #1708, opened 2026-05-12
Reduce globals in the simulator in favor of struct parameters.

pmundkur:
`trace_log` and `rvfi_cbs` are left as global for now to be handled later.

### DIFF
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -357,7 +357,7 @@ void ModelImpl::set_config_print_step(bool on) {
   m_config_print_step = on;
 }
 
-void ModelImpl::set_elf_symbols(std::map<uint64_t, std::string> &&symbols) {
+void ModelImpl::set_elf_symbols(std::map<uint64_t, std::string> symbols) {
   m_symbols = std::move(symbols);
 }
 
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -8,13 +8,6 @@
 #include "riscv_callbacks_if.h"
 #include "symbol_table.h"
 
-int term_fd = 1; // set during startup
-void plat_term_write_impl(char c) {
-  if (write(term_fd, &c, sizeof(c)) < 0) {
-    fprintf(stderr, "Unable to write to terminal!\n");
-  }
-}
-
 void ModelImpl::register_callback(callbacks_if *cb) {
   if (std::find(m_callbacks.begin(), m_callbacks.end(), cb) == m_callbacks.end()) {
     m_callbacks.push_back(cb);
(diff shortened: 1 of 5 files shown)


c_emulator/riscv_model_impl.cpp
```diff
@@ -361,6 +361,10 @@ void ModelImpl::set_config_print_step(bool on) {
   m_config_print_step = on;
 }
 
+void ModelImpl::set_elf_symbols(std::map<uint64_t, std::string> &&symbols) {
```
Timmmm: I'm 90% sure you don't need want an r-value reference here. Just pass by value:


```suggestion
void ModelImpl::set_elf_symbols(std::map<uint64_t, std::string> symbols) {
```

c_emulator/riscv_sim.cpp
```diff
@@ -326,7 +328,7 @@ static CLIOptions parse_cli(int argc, char **argv) {
   return opts;
 }
 
-uint64_t load_sail(ModelImpl &model, const std::string &filename, bool main_file) {
+uint64_t load_sail(ModelImpl &model, const std::string &filename, bool main_file, elf_info &elf_info) {
```
Timmmm, line 332: Probably makes sense to add `uint64_t entry;` to `elf_info` and then return `elf_info` instead of using a reference parameter.

Ah wait actually I see that is awkward because of our multiple ELF file support. That is a pain. We could definitely clean this up but I think probably what you have here is an improvement at least so let's go with that for now.

c_emulator/riscv_model_impl.cpp
pmundkur: Won't that copy the map and its strings when initializing the function argument?  The goal here was to avoid the copy.

pmundkur: You are right, apparently the move constructor will be used and copies avoided.

c_emulator/riscv_model_impl.h
```diff
@@ -33,10 +32,13 @@ class ModelImpl final : public hart::Model {
   void set_config_print_interrupt(bool on);
   void set_config_print_htif(bool on);
   void set_config_print_pma(bool on);
+  void set_config_print_step(bool on);
+
   void set_config_rvfi(bool on);
   void set_config_use_abi_names(bool on);
 
-  void set_config_print_step(bool on);
+  void set_elf_symbols(std::map<uint64_t, std::string> g_symbols);
```
Timmmm: Also this should just be `symbols`:

```suggestion
  void set_elf_symbols(std::map<uint64_t, std::string> symbols);
```

Timmmm (approved):

jordancarlin (approved):

Timmmm:
Huh copilot actually didn't suck. We should fix the potential hang/division by zero in another PR.

jordancarlin:
> Huh copilot actually didn't suck. We should fix the potential hang/division by zero in another PR.

I've found it's pretty good on more common languages (C++, Python, etc.). Might be worth turning on the setting to enable it automatically for the repo at this point.

PR #1712, opened 2026-05-13
Guard against a reserved shamt encoding that aliased with `add.uw`.

### LINKED ISSUE #1711: add.uw decoded for funct7=0100000 instead of only funct7=0000100
Sail incorrectly decodes the encoding `0x207383BB` as `add.uw x7, x7, x7` and executes it without trapping. The correct `add.uw` encoding requires `funct7 = 0000100` (bit 29 set), but this encoding has `funct7 = 0100000` (bit 25 set), which is a reserved/illegal encoding in the RV64 W-type instruction space (`opcode = 0111011`).

**Comparison:**

| Encoding | funct7 | Instruction | Sail | Spike |
|---|---|---|---|---|
| `0x087383BB` | `0000100` | `add.uw t2, t2, t2` | Executes | Executes  |
| `0x207383BB` | `0100000` | Reserved/illegal | Executes as `add.uw`  | Traps (illegal instruction)  |

**Spike (correct):**
```
core   0: 0x80007110 (0x207383bb) unknown
core   0: exception trap_illegal_instruction, epc 0x80007110
core   0:           tval 0x00000000207383bb
```

**Sail (incorrect):**
```
0x80007110 (0x207383BB) add.uw x7, x7, x7
x7 <- 0x0000000000000000
```

Sail's decoder appears to check only a subset of the funct7 bits when matching `add.uw`, allowing `0100000` to match when only `0000100` should be accepted. This was discovered during the Ssstrict illegal instruction sweep (`cp_rwtype`), which exhaustively sweeps all `funct7 × funct3` combinations in the W-type opcode space.

**Environment:** Sail (latest), RV64 with Zba enabled

pmundkur:
Fixes #1711.

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -16,6 +16,9 @@
     (e.g. Linux boot time reduced by up to ~50%). Off by default to keep
     incremental builds fast, enabled in CI and release builds.
 
+- Important issues addressed and bugs fixed:
+  - https://github.com/riscv/sail-riscv/issues/1711 : A reserved instruction encoding was decoded as `add.uw`.
+
 # Release notes for version 0.11
 
 - Updates to the [configuration file](../config/config.json.in):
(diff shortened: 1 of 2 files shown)


jordancarlin (approved):
Might be worth a Changelog entry since this is a somewhat significant bug.

Timmmm:
Not technically a bug since we don't support Ssstrict yet! :-D

Timmmm (approved):

PR #1720, opened 2026-05-17
Access sail-generated model structures via the C++ wrapper.

pmundkur:
This removes use of the z-encoded names from the outer simulator and insulates it from the Sail C API.

The callbacks still need to use the z-encoding since their API gives them access to `hart::Model` instead of `hart::ModelImpl`.  It would be nice to fix those as well.

Fix a `step_no` type to `nat` instead of `int` in the stepper.

### DIFF
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -46,12 +46,6 @@ void ModelImpl::set_reservation_invalidate_on_same_hart_store(bool invalidate_on
   m_reservation_invalidate_on_same_hart_store = invalidate_on_same_hart_store;
 }
 
-void ModelImpl::print_current_exception() {
-  if (current_exception != nullptr) {
-    zprint_exception(*current_exception);
-  }
-}
-
 unit ModelImpl::fetch_callback(sbits opcode) {
   for (auto c : m_callbacks) {
     c->fetch_callback(*this, opcode);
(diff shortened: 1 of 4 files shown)


c_emulator/riscv_model_impl.cpp
```diff
@@ -399,3 +393,62 @@ void ModelImpl::reinit_sail(
...
+  std::string isa(c_isa);
+  KILL(sail_string)(&c_isa);
+  return isa;
+}
+
+void ModelImpl::print_current_exception() {
+  if (current_exception != nullptr) {
+    zprint_exception(*current_exception);
+  }
+}
```
Timmmm, line 425: We should probably have this actually return a `std::option<std::string>` of the current exception if that is possible. Don't have to do it in this PR though.

c_emulator/riscv_model_impl.h
```diff
@@ -15,12 +15,16 @@ extern FILE *trace_log;
 // Model wrapped with an implementation of its platform callbacks.
 class ModelImpl final : public hart::Model {
```
Timmmm, line 16: Can you make the model private with this change?

```suggestion
class ModelImpl final : private hart::Model {
```

Timmmm (commented):
LGTM. Couple of minor comments and I guess we should fix the printf specifiers as copilot says.

c_emulator/riscv_model_impl.cpp
pmundkur: Makes sense, I'll try it in a subsequent PR.  I think I fixed the copilot printfs.

c_emulator/riscv_model_impl.h
pmundkur: I had tried that, but it didn't work because of `rvfi`: it's constructor needs ` rvfi_handler(int port, hart::Model &model)`, and `hart::Model` becomes inaccessible.  It might be possible after we change callbacks to take `hart::ModelImpl&`.

Timmmm: Yeah I think that's exactly the move. (In a future PR.)

Timmmm (approved):

arichardson (approved):
Much cleaner

PR #1726, opened 2026-05-26
Remove the `trace_log` extern global from the `riscv_model` library.

pmundkur:
This also now allows `riscv_model` to be compiled as a shared library if needed.

Make `trace_log` part of `run_info` in `riscv_sim` and close the terminal file-descriptor as well when closing logs.  Use a more explicit `STDOUT_FILENO` for `term_fd` instead of `1`, and use a flag to track whether it needs to be closed.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -56,15 +56,15 @@ struct elf_info {
 
 struct run_info {
   std::optional<rvfi_handler> rvfi;
-  int term_fd = 1;
+  // Terminal output goes to stdout unless changed via the `--terminal-log` option.
+  int term_fd = STDOUT_FILENO;
   steady_clock::time_point init_start;
   steady_clock::time_point init_end;
   uint64_t total_insns = 0;
+  FILE *trace_log = stdout;
 };
 
 } // namespace
-
-FILE *trace_log = stdout;
 
 static void print_build_info() {
   std::cout << "Sail RISC-V release: " << version_info::release_version << std::endl;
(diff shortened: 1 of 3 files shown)


c_emulator/riscv_sim.cpp
```diff
@@ -438,19 +437,22 @@ void write_signature(const std::string &file, unsigned signature_granularity, co
   fclose(f);
 }
 
-void close_logs() {
+void close_logs(run_info &run_info) {
+  if (run_info.term_fd != 1) {
+    close(run_info.term_fd);
+  }
```
nadime15:
```suggestion
  if (run_info.term_fd != -1) {
    close(run_info.term_fd);
  }
```

 I suggest changing term_fd = 1 to term_fd = -1, which is the standard Unix value for "no file descriptor assigned". The current value 1 is fragile, as 1 is reserved for stdout. It reads like you close the fd only if term_fd is not stdout.

arichardson: How about we use STDOUT_FILENO or a separate should_close bool?

pmundkur: `run_info.term_fd` is initialized to 1 (its default value), so the idea is to close it only if it was expliclitly `open()`-ed (i.e. `!= 1`).   I could change the initialization to use `STDOUT_FILENO` to be more explicit.

nadime15: `init_logs` exits the function if open returns -1 thus term_fd can only have another value than -1 if it was explicitly opened successfully (assuming term_fd is initially set to -1). Seems strange to check if its 1, if in theory this does not indicate whether it was successfully opened or not when the initial is set to 1 and open() can in theory return 1 too.

pmundkur: In theory it can, but in practice it won't since we aren't doing anything funky with file descriptors.  Would a comment help?

pmundkur: I put in a boolean flag as suggested by @arichardson which seems safest.

nadime15: Ok nice

nadime15 (approved):

arichardson (approved):
Looks good. I guess another option would be to always use a new file descriptor for term_fd (either `open()` or `dup()`), but this seems simpler.

PR #1728, opened 2026-05-26
Add an implementation of the remote serial protocol to support debuggers.

pmundkur:
This allows GDB and LLDB to interact with the simulator. This is lightly  tested with both GDB and LLDB.

This implementation does not touch any Sail code, and hence works independent of whether the model supports Sdext/Sdtrig. However, when Sdext/Sdtrig are  supported, they could be used to provide an alternative internal implementation for breakpoints and triggers.  This does add a dependency on the `asio` library for asynchronous networking.

The implementation supports basic functionality for the base integer and floating point registers, including breakpoints and watchpoints.

CSRs and vector registers are not yet supported; adding vector registers should be straightforward, but CSRs will need some infrastructure.

It would be nice to have unit tests for GDB and LLDB, but perhaps these can be done in a later PR.


dependencies/asio/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+include(FetchContent)
+
+# Allow downloads to be turned off (in particular, for targets that
+# don't need it).  Downloads may not be needed for systems that have
+# asio installed (e.g.  via `libasio-dev` on Ubuntu-based systems and
+# `asio-devel` on Fedora-based systems).
+
+option(DOWNLOAD_ASIO "Download asio" ON)
```
Timmmm: What happens if you set this to FALSE/OFF? Does it still work somehow? What creates the `asio` target?

c_emulator/riscv_model_impl.h
```diff
@@ -72,21 +72,36 @@ class ModelImpl final : private hart::Model {
...
+  // accessors not being marked const in hart::Model.
+  uint64_t peek_xreg(int64_t reg);
+  uint64_t peek_freg(int64_t reg);
   // returns std::nullopt if the model has not thrown an exception.
   std::optional<std::string> string_of_current_exception();
 
+  // write access to model state
+
+  void poke_xreg(int64_t reg, uint64_t val);
+  void poke_freg(int64_t reg, uint64_t val);
```
Timmmm: Can we have `set_xreg`/`xreg` instead of peek/poke? They're needlessly obscure IMO! Also inconsistent with the existing names.

pmundkur:
Marked this as ready for review.

c_emulator/riscv_model_impl.cpp
```diff
@@ -513,3 +525,42 @@ bool ModelImpl::htif_done() const {
...
+  sail_val.bits = val;
+  (void)zwF(reg, sail_val);
+}
+
+void ModelImpl::set_pc(uint64_t val) {
+  zPC.bits = val;
+}
+
+void ModelImpl::set_fcsr(uint64_t val) {
+  zfcsr.zbits = val;
```
Timmmm: Should we maybe call `write_fcsr()`? Otherwise this skips any legalisation, and also `dirty_fd_context_if_present()`. I double checked the RISC-V debug spec and it should call that:

> If `write` is set and `transfer` is set, then copy data from the `arg0` region of `data` into the register
specified by `regno`, **and perform any side effects that occur when this register is written from M-mode.**

`poke_freg` handles this correctly because it calls `wF()` which calls `dirty_fd_context_if_present()`.

c_emulator/riscv_model_impl.cpp
```diff
@@ -513,3 +525,42 @@ bool ModelImpl::htif_done() const {
...
+void ModelImpl::poke_freg(int64_t reg, uint64_t val) {
+  assert(reg < 32);
+  sbits sail_val;
+  sail_val.len = zxlen;
+  sail_val.bits = val;
+  (void)zwF(reg, sail_val);
+}
+
+void ModelImpl::set_pc(uint64_t val) {
+  zPC.bits = val;
```
Timmmm: Similarly to `set_fcsr` should we call `zset_next_PC()` here? We don't currently do anything special when PC is written, but when configurable physical address width is supported, or with CHERI we would.

c_emulator/CMakeLists.txt
```diff
@@ -93,6 +93,38 @@ add_executable(sail_riscv_sim
     traploop_detector.h
 )
 
+option(ENABLE_GDBSERVER
+    "Enable gdb-server for interaction with debuggers as a remote target."
+    OFF
+)
```
Timmmm: Does this need to be a configure time option? I would have thought we could just always enable it to be honest. What are the downsides?

c_emulator/gdb/triggers.cpp
```diff
@@ -0,0 +1,155 @@
...
+  if (m_run_info.enable_trace) {
+    fprintf(m_run_info.trace_log, "%s.\n", os.str().c_str());
+  }
+}
+
+bool triggers::at_watchpoint(AccessType t, uint64_t addr, int64_t width) {
+  for (const auto &w : m_watchpoints) {
+    // Match an overlapping access of the right type.  This assumes
+    // that the provided ranges don't wrap if width-1 is used.
+    if ((addr + width - 1 < w.first) | (w.first + w.second.width - 1 < addr)) {
```
Timmmm:
```suggestion
    if ((addr + width <= w.first) | (w.first + w.second.width <= addr)) {
```
Maybe?

c_emulator/gdb/gdb_run_info.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include <stdio.h>
+
+struct gdb_run_info {
+  bool enable_trace;
+  FILE *trace_log;
```
Timmmm:
```suggestion
  bool enable_trace = false;
  FILE *trace_log = nullptr;
```

Just to be safe...

c_emulator/gdb/gdbserver.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+void accept(asio::ip::tcp::acceptor &acceptor, ModelImpl &model, gdb_run_info &info) {
+  acceptor.async_accept([&model, &info](asio::error_code ec, asio::ip::tcp::socket socket) {
+    if (ec) {
+      std::cerr << "Error on accept: " << ec << ": " << ec.value() << " " << ec.message() << std::endl;
+      return;
+    }
+
+    auto conn = std::make_shared<connection>(std::move(socket), model, info);
+    conn->read();
```
Timmmm, line 19: What happens if you connect to the port multiple times? I think we should only accept one connection. Presumably you can use `accept` instead of `async_accept`? (I haven't used asio for at least a decade...)

c_emulator/gdb/connection.cpp
```diff
@@ -0,0 +1,93 @@
...
+    m_handler{std::make_shared<protocol_handler>(m_socket.get_executor(), *this, model, info)} {
+
+  // This cannot be done in the handler's constructor due to the
+  // restrictions of `std::enable_shared_from_this`.
+  m_handler->attach_callbacks();
+}
+
+void connection::read() {
+  auto self(shared_from_this());
+  async_read_until(
```
Timmmm, line 29: Do we need anything to be async? My mental model for how this should work (and this is actually what I've implemented for driving JTAG in a SV simulation), is that you pause the simulation and do blocking reads until you get a command that would allow the simulation to continue (step or resume). Then you run until something happens (breakpoint hit etc.) without reading from GDB, and then repeat.

That should be fully deterministic I think? The thing I think we should avoid is allowing GDB to be processing stuff *in parallel* with the Sail simulation running. That would make it non-deterministic which means we can't use it for testing etc.

(I may have misunderstood what `async_read_until` does - the `async` just made me worried...)

c_emulator/gdb/connection.h
```diff
@@ -0,0 +1,31 @@
...
+#include <asio.hpp>
+#include <deque>
+#include <memory>
+
+struct gdb_run_info;
+class ModelImpl;
+
+class connection : public std::enable_shared_from_this<connection> {
+public:
+  explicit connection(asio::ip::tcp::socket socket, ModelImpl &model, gdb_run_info &info);
```
Timmmm, line 14: Should we maybe do the "Best" option from here?

https://en.cppreference.com/cpp/memory/enable_shared_from_this

c_emulator/gdb/notes.md
```diff
@@ -0,0 +1,55 @@
+Some notes about this implementation of the Remote Serial Protocol:
+
+- It depends on `asio` for asynchronous networking. Despite being a
+  mostly synchronous request-response protocol, it does have an
+  asynchronous interrupt (`0x03`) notification, which could appear
```
Timmmm, line 13: Ah that's awkward. I guess we can make it fully deterministic except for this one thing, and then just not use that in tests?

c_emulator/gdb/notes.md
```diff
@@ -0,0 +1,55 @@
...
+
+- Binary data in packet payloads needs to be escaped. The server
+  doesn't send any such packets _yet_, so no escaping is done on
+  transmission.
+
+- The server only serves one client connection and then exits when
+  that connection closes. For the server to stay alive to serve
+  another client, the model will likely need to be re-initialized and
+  the ELF file re-loaded, and current APIs would need to be extended
+  if this use case really needs to be supported.
```
Timmmm, line 63: Yeah I can't imagine ever needing this tbh. *Maybe* if you're debugging a really long process like Linux and accidentally disconnect or something?

c_emulator/gdb/parse_utils.cpp
```diff
@@ -0,0 +1,10 @@
+#include "parse_utils.h"
+
+std::optional<uint64_t> string_to_opt_uint64t(const std::string &istr) {
+  try {
+    unsigned long long num = std::stoull(istr, nullptr, 16);
```
Timmmm: Unfortunately this will stop at the first unconvertable character, e.g. `   12three` will return 12.

Kind of blows my mind that C++ *still* doesn't have a sane string-to-int function... it did take them several decades to get `std::string::starts_with()`... so maybe in another 20 years? :-D

[`from_chars`](https://en.cppreference.com/cpp/utility/from_chars) may be slightly easier to use and at least it doesn't skip whitespace. I think this does the job:

```
std::optional<uint64_t> string_to_opt_uint64t(const std::string &str) {
  uint64_t result{};
  const char* end = str.data() + str.size();
  auto [ptr, ec] = std::from_chars(str.data(), end, result, 16);
  if (ec == std::errc() && ptr == end) {
    return result;
  }
  return std::nullopt;
}
```

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+      m_parse_buffer.erase(0, 1);
+      auto resp = response::interrupt();
+      resp.dispatch(*this, m_run_info);
+      continue;
+    }
+    if (c == '+') {
+      m_parse_buffer.erase(0, 1);
+      continue;
+    }
+    // buffer begins with '$'.
```
Timmmm: I think `switch (c) { ...` might be a bit more readable here.

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+    auto hash_idx = m_parse_buffer.find('#');
+    if (hash_idx == std::string::npos || hash_idx + 2 >= m_parse_buffer.length()) {
+      if (m_run_info.enable_trace) {
+        fprintf(m_run_info.trace_log, "Incomplete request: [%s]\n", m_parse_buffer.c_str());
+      }
```
Timmmm: This is a perfectly valid thing to happen no? I think it probably shouldn't print anything...

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+      break;
+    }
+    // Handle checksum.
+    int checksum = 0;
+    for (std::string::size_type i = 1; i < hash_idx; ++i) {
+      checksum += m_parse_buffer[i];
+    }
+    checksum &= 255;
```
Timmmm: Probably just `uint8_t checksum` instead?

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+    auto pkt_checksum_str = std::string(m_parse_buffer, hash_idx + 1, hash_idx + 3);
+    int pkt_checksum = std::stoi(pkt_checksum_str, nullptr, 16);
```
Timmmm: Probably should use `string_to_opt_uint64t` here!

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+    if (pkt_checksum != checksum) {
+      std::cerr << "Checksum mismatch: computed " << checksum << " but got " << pkt_checksum << "(" << pkt_checksum_str
+                << ")" << std::endl;
+      std::cerr << " for msg " << m_parse_buffer << std::endl;
+      std::exit(EXIT_FAILURE);
```
Timmmm: Can we throw an exception instead?

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+      }
+      break;
+    }
+    // Handle checksum.
+    int checksum = 0;
+    for (std::string::size_type i = 1; i < hash_idx; ++i) {
+      checksum += m_parse_buffer[i];
+    }
+    checksum &= 255;
+    auto pkt_checksum_str = std::string(m_parse_buffer, hash_idx + 1, hash_idx + 3);
```
Timmmm: `m_parse_buffer.substr(hash_idx + 1, 2)`?

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+    // Skip the leading '$' and checksum.
+    auto raw_req = m_parse_buffer.substr(1, hash_idx - 1);
+    m_parse_buffer.erase(0, hash_idx + 3);
+    process_raw_request(std::move(raw_req));
+  }
+}
+
+void protocol_handler::process_raw_request(std::string req) {
+  // Unescape the buffer.
+  std::string::size_type read_idx = 0, write_idx = 0, len = req.length();
```
Timmmm: Minor nit but I find variable declarations with commas are a footgun in C++ - can you put them on separate lines?

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+  while (read_idx < len) {
+    if (req[read_idx] == '}') {
+      ++read_idx;
+      req[write_idx] = static_cast<char>(req[read_idx] ^ 0x20);
```
Timmmm, line 105: I think you need to check again that `read_idx` is `< len` otherwise you have an out-of-bound read.

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+    }
+  }
+  if (resp.has_value()) {
+    queue_response(std::move(resp.value()));
+  } else {
+    std::cerr << "Request was not recognized!" << std::endl;
+    // TODO: We should generally send an unsupported response.  But we
+    // don't send any response for now since the interaction stops
+    // without a response and this helps to find remaining protocol
+    // packets that may need to be supported.
```
Timmmm, line 136: Can't you just print the packet that wasn't recognised to find the packets that need to be supported? I'm not sure I understand this point.

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+  // TODO: escape special characters in `resp`.  This will be needed
+  // if any response sends binary data, but there is no such response
+  // as yet (e.g. the `x` request is responded to as not-supported).
+  data.append(resp);
+
+  data.append("#");
+  {
+    // Compute checksum.
+    int checksum = 0;
```
Timmmm:
```suggestion
    uint8_t checksum = 0;
```

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+void protocol_handler::send_data(const std::string &data) {
+  if (m_run_info.enable_trace) {
+    fprintf(m_run_info.trace_log, "gdbserver --> %s\n", data.c_str());
+  }
+  m_connection.send_data(data);
+}
+
+void protocol_handler::send_stop_reply() {
+  if (m_model.htif_done()) {
+    // Send `W AA`, where AA is the exit status in big-endian hex.
```
Timmmm, line 178: Is it actually without the space?

```suggestion
    // Send `WAA`, where AA is the exit status in big-endian hex.
```

c_emulator/gdb/protocol_handler.cpp
```diff
@@ -0,0 +1,304 @@
...
+    auto exit_code = m_model.htif_exit_code();
+    int n = static_cast<int>(sizeof(exit_code));
+    std::ostringstream buf;
+    buf << "W" << std::hex << std::setfill('0');
+    for (int i = n - 1; i >= 0; --i) {
+      unsigned byte = (exit_code >> (i * 8)) & 0xFF;
+      buf << std::setw(2) << byte;
+    }
```
Timmmm: Err maybe I'm missing something but isn't this just doing the same as `buf << exit_code`?

c_emulator/gdb/responses.h
```diff
@@ -0,0 +1,251 @@
...
+class write_register : public response_handler {
+public:
+  write_register(uint64_t regidx, uint64_t regval) : m_regidx(regidx), m_regval(regval) {
+  }
+
+  void dispatch(protocol_handler &, gdb_run_info &) override;
+
+private:
+  uint64_t m_regidx;
+  uint64_t m_regval;
```
Timmmm: I'd probably add `= 0;` to all the `uint64_t` members in this file just to be safe. (Yet another C++ thing they have never fixed.)

c_emulator/gdb/requests.h
```diff
@@ -0,0 +1,45 @@
...
+class request_parsers {
+public:
+  request_parsers();
+  ~request_parsers() = default;
+
+  const std::vector<request_parser_ptr> &get_parsers() const;
+
+private:
+  std::vector<request_parser_ptr> parsers;
+};
```
Timmmm: I think this class doesn't really do anything - you could replace it with a function

```
std::vector<request_parser_ptr> create_packet_parsers() {
  return {
    new q...
  }
}
```

And then

```
request_parsers m_parsers = create_packet_parsers();
```

c_emulator/gdb/requests.h
```diff
@@ -0,0 +1,45 @@
...
+
+namespace request {
+
+class request_parser {
+public:
+  virtual ~request_parser() = default;
+
+  virtual std::optional<response_handler_ptr> parse(const std::string &, gdb_run_info &enable_trace) const {
+    return std::nullopt;
+  }
```
Timmmm: Probably makes more sense to be pure since every parser is going to parse...

```suggestion
  virtual std::optional<response_handler_ptr> parse(const std::string &, gdb_run_info &enable_trace) const = 0;
```

Timmmm (commented):
Lots of code! I've added some initial comments but I haven't looked at `protocol_handler.h`, `request.cpp` or `responses.cpp`. Generally looks good; my main worry is that we make sure it is deterministic (as long as you don't use the interrupt packet).

c_emulator/gdb/connection.cpp
pmundkur: Right, the `async_read` is needed to handle the interrupt packet, which might arrive while the simulator is executing (e.g. because the user pressed control-C).

c_emulator/gdb/notes.md
pmundkur: I would like to have a test for the interrupt packet, to make sure it works (e.g. have the simulator execute an infinite loop, and then have the debugger send an interrupt packet, and get a response).  But otherwise the tests would be purely deterministic.

c_emulator/gdb/requests.h
pmundkur: ~The advantage of the class is that the list of request parsers is constructed once in the constructor and then returned every time the protocol parser needs it.  Unless the function used a static, it would construct the list from scratch every time?~

c_emulator/gdb/protocol_handler.cpp
pmundkur: I was not sending a response when the request packet did not parse, as a fail-stop way of finding such packets.  Good point though, I'll also print the request packet to `std::cerr`.  (I always had the trace turned on, so was seeing it that way in such cases.)  I'd keep the fail-stop for now, but perhaps not phrase it as a `TODO`?

c_emulator/gdb/protocol_handler.cpp
pmundkur: Yes, it's without the space, but the spec uses spaces for clarity.  I added a comment about this to the notes.

c_emulator/gdb/gdbserver.cpp
pmundkur: Yes, it accepts only one connection since it doesn't re-issue the accept after a connection is accepted, and if there is an error on the first accept, it just prints an error and exits.  I guess I could use a synchronous accept, but this seems in keeping with the rest of the async code.  I can change it if you prefer.

c_emulator/gdb/triggers.cpp
pmundkur: I was trying to address wraparound as the comment above suggests; maybe I should have done this instead
```suggestion
    if ((addr + (width - 1) < w.first) | (w.first + (w.second.width - 1) < addr)) {
```
but perhaps there is a better way?

c_emulator/riscv_model_impl.cpp
pmundkur: I set the bits directly and called `dirty_fd_context_if_present()`.  This allows the debugger to set reserved bits of `fcsr` if it really wants to, whereas you can't do that with `write_fcsr()`.

pmundkur (commented):
Thanks for the review!  I've addressed most of them, and will look into the remainder soon.

c_emulator/gdb/parse_utils.cpp
pmundkur: Thanks for this pointer!

c_emulator/gdb/protocol_handler.cpp
pmundkur: Good catch!

c_emulator/CMakeLists.txt
pmundkur: I added the option pretty early, when I had manually installed `asio` and before I added it to the CMake build.  I can remove it; I can't think of any downsides now.

dependencies/asio/CMakeLists.txt
pmundkur: Good question.  The build fails unless `libasio-dev` is installed (on Ubuntu); if it is installed, the build picks it up.  

But the build still fails with `libasio-dev 1.28.1`, which is surprising since I had started working on this with that version installed.  I then removed that package when moving the dependency install to CMake, and used the latest stable upstream version.  Perhaps I should remove this option along with `ENABLE_GDBSERVER`?

pmundkur: Removed.

c_emulator/gdb/connection.h
pmundkur: As far as I can tell, the "Best" pattern is good for external clients of the class to get shared pointers to its objects.  Here (for `class connection`), we are creating the shared pointers for connection objects for internal use.  The same is true for `class protocol_handler`: the shared pointers are created internally and used for callbacks; no external client class is calling `connection` or `protocol_handler` to get a shared pointer.  So I think the "Good" way is good enough?

pmundkur:
@Timmmm @jordancarlin @nadime15 @arichardson Any issues with merging this?  This is fairly functional, and though there are likely to be some issues, it'll be easier to find them once people start using it, and these issues don't impact any Sail code or the main emulator.

c_emulator/riscv_model_impl.cpp
```diff
@@ -523,3 +535,50 @@ bool ModelImpl::htif_done() const {
...
+  sbits sail_val;
+  sail_val.len = zxlen;
+  sail_val.bits = val;
+  (void)zwX(reg, sail_val);
+}
+
+void ModelImpl::set_freg(int64_t reg, uint64_t val) {
+  assert(reg < 32);
+  sbits sail_val;
+  sail_val.len = zxlen;
```
nadime15:
```suggestion
  sail_val.len = zflen;
```

This probably should be zflen?

pmundkur: Indeed, thanks!

nadime15:
@pmundkur Yeah thats fine with me since its quite isolated.

arichardson (commented):
I likely won't have time to review this in the coming days but if everyone else is happy I think we should merge it. Skimming the diff it looks like it's very self contained and an important feature. Since it won't break anything else I think we can go ahead with this.

jordancarlin (approved):
Approving because this seems noninvasive and is a useful feature to have. We can fix any issues in follow up PRs.

PR #1730, opened 2026-05-28
Hoist the cmake logic for a vector test suite into a function.

pmundkur:
This makes it easier to change VLEN/ELEN combinations in the future.


nadime15 (approved):

Timmmm:
Doesn't seem easier to me yet but I'll take your word for it that it will be in future!

PR #1731, opened 2026-05-29
Refactor the callback API to avoid z-encoded internal model types.

pmundkur:
This builds on the previous refactors of #1720 and #1724. The callbacks now access model internals via `ModelImpl`. This allows `hart::Model` to be a private inheritance of `ModelImpl`.  Callbacks are still exposed to basic Sail C types (i.e. `{lsf}bits`).

Unfortunately, RVFI requires more complex access and for now its implementation classes are made friends of `ModelImpl`.


pmundkur:
@Timmmm @arichardson would you have time to review this?

arichardson (approved):
Much better not using those internal names. No opinion on whether we need to change to private inheritance to prevent accessing them. So far it seems like there is very little boilerplate caused by this so LGTM

Timmmm (approved):
Much better!

PR #1737, opened 2026-06-01
Handle reserved `vtype` bits in `vsetvli`/`vsetivli` instruction encodings.

### LINKED ISSUE #1699: Add configuration parameter for vsetvli/vsetvl behavior on reserved vtype values
When `vsetvli` or `vsetvl` is executed with a reserved vtype encoding (e.g., `vlmul = 100`), the RISC-V V specification states that the implementation should set `vtype.vill = 1` and `vl = 0`. Sail currently raises an illegal-instruction exception (`mcause = 2`) instead.

Both behaviors can be argued as legal:
- Setting `vill = 1`: the instruction is a valid `vsetvli` — it's the *requested configuration* that's unsupported, not the instruction itself
- Trapping: the encoding contains reserved field values, which could be treated as illegal. Implementations could choose
to trap when illegal values are written to vtype instead of setting vill, to allow emulation to
support new configurations for forward-compatibility.

QEMU implement the `vill = 1` behavior. This prevents Sail from being used as a reference model for compliance tests that include vector configuration sweeps.

**Proposed solution:**

Add two configuration parameters under `extensions.V`:

```json
"V": {
    "trap_on_reserved_vtype_imm": false,
    "trap_on_reserved_vtype_reg": false
}
```

Two separate parameters are needed because the two cases have different characteristics:

1. **`trap_on_reserved_vtype_imm`** — controls `vsetvli` and `vsetivli` behavior. The reserved vtype value is encoded as an immediate in the instruction, so it is known at decode time. An implementation could choose to treat this as an illegal encoding.

2. **`trap_on_reserved_vtype_reg`** — controls `vsetvl` behavior. The vtype value comes from a register (`rs2`), so it is only known at execution time. Trapping here would mean the instruction's legality depends on a register value, which is unusual.

**When `false` (default, matching QEMU):**
- `vsetvli`/`vsetivli`/`vsetvl` with reserved vtype sets `vtype.vill = 1`, `vl = 0`, and writes `rd = 0`
- No exception is raised

**When `true` (current Sail behavior):**
- The instruction raises `mcause = 2` (illegal instruction)

pmundkur:
Though the bits may be reserved, it is more flexible to still decode the instructions in order to decide how to handle the reserved behavior.

This PR just sets the `vill` bit in `vtype`.  A subsequent PR will introduce the parameters to handle the reserved behavior.

Partially fixes #1699.

### DIFF
--- a/model/extensions/V/vext_vset_insts.sail
+++ b/model/extensions/V/vext_vset_insts.sail
@@ -40,15 +40,6 @@ mapping ma_flag : string <-> bits(1) = {
   sep() ^ "mu" <-> 0b0,
 }
 
-mapping vtype_assembly : string <-> (bits(1), bits(1), bits(3), bits(3)) = {
-  // Mnemonics are preferable but there's no mnemonic for lmul=0b100.
-  sew_flag(sew) ^ maybe_lmul_flag(lmul) ^ ta_flag(ta) ^ ma_flag(ma) <-> (ma, ta, sew, lmul)
-    when sew[2] != 0b1 & lmul != 0b100,
-  // However that is still potentially a legal instruction so you can
-  // use an integer for the whole vtype in that case.
-  hex_bits_8(ma @ ta @ sew @ lmul)  <-> (ma, ta, sew, lmul),
-}
-
 // For vsetvli and vsetivli, we absorb a reserved vtype fragment
 // (`vtr`) to parameterize reserved behavior, but this fragment cannot
 // be represented in assembly, so a custom mapper is used for the
@@ -56,6 +47,15 @@ mapping vtype_assembly : string <-> (bits(1), bits(1), bits(3), bits(3)) = {
 val vtr_mnemonic : bits(3) <-> string
 function vtr_mnemonic_forwards(_vtr) = ""
 function vtr_mnemonic_forwards_matches(_vtr) = true
+
+mapping vtype_assembly : (bits(3), bits(1), bits(1), bits(3), bits(3)) <-> string = {
+  // Mnemonics are preferable but there's no mnemonic for lmul=0b100.
+  (vtr, ma, ta, sew, lmul) <-> sew_flag(sew) ^ maybe_lmul_flag(lmul) ^ ta_flag(ta) ^ ma_flag(ma) ^ vtr_mnemonic(vtr)
+    when sew[2] != 0b1 & lmul != 0b100 & vtr == 0b000,
+  // However that is still potentially a legal instruction so you can
+  // use an integer for the whole vtype in that case.
+  (vtr, ma, ta, sew, lmul) <-> hex_bits_11(vtr @ ma @ ta @ sew @ lmul)
+}
 
 private function handle_illegal_vtype(rd : regidx) -> unit = {
   // Note: Implementations can set vill or trap with an illegal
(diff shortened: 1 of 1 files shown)


model/extensions/V/vext_vset_insts.sail
```diff
@@ -138,13 +146,16 @@ private function execute_vsetvl_type(
...
 
 // ********************************** vsetvli ***********************************
-union clause instruction = VSETVLI : (bits(1), bits(1), bits(3), bits(3), regidx, regidx)
+union clause instruction = VSETVLI : (bits(3), bits(1), bits(1), bits(3), bits(3), regidx, regidx)
 
-mapping clause encdec = VSETVLI(ma, ta, sew, lmul, rs1, rd)
-  <-> 0b0000 @ ma @ ta @ sew @ lmul @ encdec_reg(rs1) @ 0b111 @ encdec_reg(rd) @ 0b1010111
+mapping clause encdec = VSETVLI(vtr, ma, ta, sew, lmul, rs1, rd)
+  <-> 0b0 @ vtr @ ma @ ta @ sew @ lmul @ encdec_reg(rs1) @ 0b111 @ encdec_reg(rd) @ 0b1010111
   when currentlyEnabled(Ext_Zve32x)
```
nadime15, line 151: Would it make sense to do something like:

```
mapping clause assembly = VSETVLI(vtr, ma, ta, sew, lmul, rs1, rd)
    <-> "vsetvli" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ vtype_assembly(ma, ta, sew, lmul)
      when vtr == zeros()
mapping clause assembly = VSETVLI(vtr, ma, ta, sew, lmul, rs1, rd)
    <-> "vsetvli" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(rs1) ^ sep() ^ hex_bits_11(vtr @ ma @ ta @ sew @ lmul)
      when vtr != zeros()
```

Do match Spikes output in case of vtr != zeros()? 

This would also git rid of the above:
```

// For vsetvli and vsetivli, we absorb a reserved vtype fragment
// (`vtr`) to parameterize reserved behavior, but this fragment cannot
// be represented in assembly, so a custom mapper is used for the
// mnemonic.
val vtr_mnemonic : bits(3) <-> string
function vtr_mnemonic_forwards(_vtr) = ""
function vtr_mnemonic_forwards_matches(_vtr) = true
```

I think this would be a bit cleaner and we already have cases (C_NOP) with multiple mapping clauses for a given instruction.

pmundkur: It won't match Spike's output, which was posted in #1699.  But if it is legal assembly, then yes I think this may be a better approach.  @Timmmm didn't you have an assembly checker that could check this?

nadime15: I don't understand, why wouldn't it match Spike's output?

`vsetvli s8, a0, 0x494`

Isn't that exactly what the clause above does?

pmundkur: When `vtr != zeros()` Spike doesn't print `hex_bits_11(vtr @ ma @ ta @ sew @ lmul)`.  Or am I missing something?

pmundkur: Actually, we should use `hex_bits_12()` and then it would work I think.

nadime15: I think `hex_bits_11()` is fine they produce the same result just the argument width is different.

pmundkur: Unfortunately, the first clause (when `vtr == zeros())` doesn't work when trying this.  You still need the `vtr_mnemonic` hack. I just folded it into the `vtype_assembly` mnemonic.

nadime15: Ahh yeah I see

model/extensions/V/vext_vset_insts.sail
```diff
@@ -40,13 +40,21 @@ mapping ma_flag : string <-> bits(1) = {
   sep() ^ "mu" <-> 0b0,
 }
 
-mapping vtype_assembly : string <-> (bits(1), bits(1), bits(3), bits(3)) = {
+// For vsetvli and vsetivli, we absorb a reserved vtype fragment
+// (`vtr`) to parameterize reserved behavior, but this fragment cannot
+// be represented in assembly, so a custom mapper is used for the
+// mnemonic.
```
nadime15: I think this needs to be updated now (it works now I guess with your new change?)

nadime15 (approved):

model/extensions/V/vext_vset_insts.sail
pmundkur: Good catch, fixed.

pmundkur:
I'd like to merge this soon so that we can rebase #1732 on top of this to fix #1699.

PR #1738, opened 2026-06-02
Update to the latest (2026-05-29) test suite release.

pmundkur:
The main highlight of this release is the vector test suites that have been fixed to avoid generating instructions that were reserved due to source register groups that overlapped with different EEWs. See: chipsalliance/riscv-vector-tests#92 and chipsalliance/riscv-vector-tests#93

The riscv-tests and riscv-arch-test suites were also updated to their latest releases.

Built with:
riscv-software-src/sail-riscv-tests: 222fbdd
riscv-tests: 1eb47d946c55f55cab8653c224c2993acc0276bd 
riscv-vector-tests: f76bff121dce91b6b23e1d37be77a5e44af914c3 
riscv-arch-test: cedc9279fae6636b526d210ae47c0b5eb7654305 
Spike: 20feb9c2bf2a7deab964d8190b0cbd4b4131bec3
Toolchain release: 2026.03.13
Sail RISC-V release: 0.11
```
$ ./compare-releases.py -p releases/2026-04-17 -c releases/2026-05-29 
riscv-tests has 667 test files, with 12 added to and 0 removed from the previous release. 
riscv-arch-tests has 2080 test files, with 554 added to and 0 removed from the previous release. 
riscv-vector-tests-v128x32 has 1944 test files, with 0 added to and 0 removed from the previous release. 
riscv-vector-tests-v128x64 has 3042 test files, with 0 added to and 0 removed from the previous release. 
riscv-vector-tests-v256x32 has 1945 test files, with 0 added to and 0 removed from the previous release. 
riscv-vector-tests-v256x64 has 3043 test files, with 0 added to and 0 removed from the previous release. 
riscv-vector-tests-v512x32 has 1945 test files, with 0 added to and 0 removed from the previous release. 
riscv-vector-tests-v512x64 has 3043 test files, with 0 added to and 0 removed from the previous release.
```

Some tests in `riscv-arch-tests` fail for reasons documented in `test/CMakeLists.txt`; these have been excluded from test runs.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -125,37 +125,37 @@ endif()
 option(ENABLE_RISCV_ARCH_TESTS "Enable riscv-arch-tests" OFF)
 
 function(filter_arch_tests test_elfs elfs_to_use_varname)
-  # These tests pass with a config matching the one used in
-  # `riscv-arch-test`, but which differs from the ones generated in
-  # our build.  They can be re-enabled once we support more granular
-  # test-specific configurations in CI.
-  list(FILTER test_elfs EXCLUDE REGEX ".*/InterruptsS-00\\.elf$")
-  list(FILTER test_elfs EXCLUDE REGEX ".*/InterruptsSm-00\\.elf$")
-  list(FILTER test_elfs EXCLUDE REGEX ".*/S-00\\.elf$")
-  list(FILTER test_elfs EXCLUDE REGEX ".*/SsstrictS-01\\.elf$")
-  list(FILTER test_elfs EXCLUDE REGEX ".*/UV-00\\.elf$")
+    # These tests pass with a config matching the one used in
+    # `riscv-arch-test`, but which differs from the ones generated in
+    # our build.  They can be re-enabled once we support more granular
+    # test-specific configurations in CI.
+    list(FILTER test_elfs EXCLUDE REGEX ".*/InterruptsS-00\\.elf$")
+    list(FILTER test_elfs EXCLUDE REGEX ".*/InterruptsSm-00\\.elf$")
+    list(FILTER test_elfs EXCLUDE REGEX ".*/S-00\\.elf$")
+    list(FILTER test_elfs EXCLUDE REGEX ".*/SsstrictS-01\\.elf$")
+    list(FILTER test_elfs EXCLUDE REGEX ".*/UV-00\\.elf$")
 
-  # These tests fail because
-  # https://github.com/riscv/sail-riscv/pull/1690 strengthened illegal
-  # instruction checks for vector instructions (like vbrev8).
-  # Instructions that executed with the 0.11 release now raise illegal
-  # instruction exceptions with the current version.
-  list(FILTER test_elfs EXCLUDE REGEX ".*/SsstrictSm-14\\.elf$")
+    # These tests fail because
+    # https://github.com/riscv/sail-riscv/pull/1690 strengthened illegal
+    # instruction checks for vector instructions (like vbrev8).
+    # Instructions that executed with the 0.11 release now raise illegal
+    # instruction exceptions with the current version.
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -124,6 +124,41 @@ endif()
 # Download and add RISC-V arch tests.
 option(ENABLE_RISCV_ARCH_TESTS "Enable riscv-arch-tests" OFF)
 
+function(filter_arch_tests test_elfs elfs_to_use)
+  # These tests pass with a config matching the one used in
+  # `riscv-arch-test`, but which differs from the ones generated in
+  # our build.  They can be re-enabled once we support more granular
+  # test-specific configurations in CI.
```
Timmmm: What's the significant difference in config?

test/CMakeLists.txt
```diff
@@ -124,6 +124,41 @@ endif()
...
+  list(FILTER test_elfs EXCLUDE REGEX ".*/InterruptsS-00.elf$")
+  list(FILTER test_elfs EXCLUDE REGEX ".*/InterruptsSm-00.elf$")
+  list(FILTER test_elfs EXCLUDE REGEX ".*/S-00.elf$")
+  list(FILTER test_elfs EXCLUDE REGEX ".*/SsstrictS-01.elf$")
+  list(FILTER test_elfs EXCLUDE REGEX ".*/UV-00.elf$")
```
Timmmm: Technically...

```suggestion
  list(FILTER test_elfs EXCLUDE REGEX ".*/UV-00\\.elf$")
```

But it's not going to actually matter in this case so up to you.

test/CMakeLists.txt
```diff
@@ -124,6 +124,41 @@ endif()
...
+  # These RV64 tests fail because
+  # https://github.com/riscv/sail-riscv/pull/1712 fixed an incorrect
+  # decoding of add.uw, and now correctly treats that encoding as
+  # reserved and traps with an illegal instruction exception.
+  list(FILTER test_elfs EXCLUDE REGEX ".*/sail-rv64-max/.*/SsstrictS-06.elf$")
+  list(FILTER test_elfs EXCLUDE REGEX ".*/sail-rv64-max/.*/SsstrictSm-06.elf$")
+  list(FILTER test_elfs EXCLUDE REGEX ".*/sail-rv64-max/.*/SsstrictU-02.elf$")
+
+  set(elfs_to_use "${test_elfs}")
+  return(PROPAGATE elfs_to_use)
```
Timmmm: I think `PROPAGATE` isn't supported by CMake 3.20 (which is the earliest version we claim to support). But also I don't think you need it here.

Also I think this should be like this:

```
function(filter_arch_tests test_elfs elfs_to_use_varname)
  ...
  set(${elfs_to_use_varname} "${test_elfs}")
  return(PROPAGATE ${elfs_to_use_varname})
```

So that the variable name isn't hard-coded in the function. (Yeay CMake)

test/CMakeLists.txt
pmundkur: Their config use a much higher `max_time_to_wait` of 200, which I think the interrupt tests need; they also enable a few more extensions.  I hadn't realized that we had some extensions off by default that I don't think have incompatibility issues.   I haven't looked into why yet.

Timmmm: Ah fair enough. We should probably just change our config to match theirs then I guess if it works.

Timmmm (approved):

jordancarlin:
> Some tests in riscv-arch-tests fail for reasons documented in test/CMakeLists.txt; these have been excluded from test runs.

Happy to help look into these in a few weeks once I’m back if they’re not resolved by then.

jordancarlin (approved):

pmundkur:
Merging this since it will unblock #1698.

PR #1741, opened 2026-06-02
Support single-mode trap vector configurations for `mtvec` and `stvec`.

### LINKED ISSUE #1729: Allow configuring mtvec/stvec MODE as fixed (single-mode) and its reset value
Sail always implements a writable tvec MODE that legalizes both direct and vectored, so it can't model cores that support only one mode or reset to a non-zero mode. 

This surfaced during ACT runs: a vectored-only core (cv32e20) and a core that resets to vectored (cv32e40p) both diverged from the Sail reference on trap-vector signatures. Those were patched on the test-framework side (riscv/riscv-arch-test#1574, riscv/riscv-arch-test#1590), but the cleaner fix is to express the core's actual mode behavior.

pmundkur:
Also set `{ms}tvec` to a supported mode on reset.  This is needed to match DUTs that have a hard-wired single-mode.

Fixes #1729.

### DIFF
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -575,6 +575,7 @@ private function check_mmio_devices() -> bool = {
 
 function config_is_valid() -> bool = {
     check_privs()
+  & check_tvecs()
   & check_mmu_config()
   & check_mem_layout()
   & check_mmio_devices()
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -15,6 +15,22 @@ private function check_privs() -> bool = {
   true
 }
 
+private function check_tvecs() -> bool = {
+  var valid : bool = true;
+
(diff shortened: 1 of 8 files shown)


model/postlude/validate_config.sail
```diff
@@ -15,6 +15,22 @@ private function check_privs() -> bool = {
...
+  };
+
+  if (config extensions.S.supported : bool) & not(config base.stvec.direct.supported : bool) & not(config base.stvec.vectored.supported : bool) then {
+    valid = false;
+    print_endline("`stvec` specifies neither `direct` or `vectored` is supported; one of the modes must be supported.");
+  };
+
+  valid
+}
+
```
nadime15, line 33: You dont seem to use it?

pmundkur: Hah! Forgot about that :-)

nadime15 (approved):

jordancarlin (approved):

PR #1742, opened 2026-06-03
Make WFI optionally available to user-mode.

### LINKED ISSUE #1739: [bug?] WFI in U-mode on M/U-platform
### TL;DR

M/U-platforms executing `WFI` in user-mode with `mstatus.TW=0` should not throw an illegal instruction exception.

-----------

I am porting the new [riscv-arch-test](https://github.com/riscv/riscv-arch-test) framework to a custom RISC-V core that implements `M` and `U` privilege modes.

The `InterruptsU` tests fail and while debugging the logs and traces I came across this:

According to the trace log Sail **immediately** raises an illegal instruction fault when executing `WFI` in user-mode (`mstatus.TW = 0`):

```
[3981] [U]: 0x80000B08 (0x10500073) wfi                                             InterruptsU_cg_cp_wfi_mie_0_tw_0+0
trapping from U to M to handle illegal-instruction
```

This also happens when `mstatus.TW = 1`, which seems fine:

```
[5097] [U]: 0x8000105C (0x10500073) wfi                                             InterruptsU_cg_cp_wfi_timeout_mie_0_mtie_1+0
trapping from U to M to handle illegal-instruction
```

In the according M-mode tests (`InterruptsSM`) Sail actaully enters a wait state (ignoring `mstatus.TW`), which seems fine, too:

```
[14161] [M]: 0x800043F4 (0x10500073) wfi                                            InterruptsSm_cg_cp_wfi_mie_0_tw_0+0
entering WAIT-WFI state at PC 0x800043F4
```

Looking at Sail's source code, it seems like an illegal instruction exception is always thrown when executing `WFI` U-mode:

https://github.com/riscv/sail-riscv/blob/ac28d43b242ecdded6c61dc592cfb1132be85c1b/model/extensions/I/base_insts.sail#L654-L663

Now the question is: As far as I understand the RISC-V specification, on an M/U platform (no S-mode), `WFI` in user-mode with `mstatus.TW=0` should not raise an illegal instruction exception:

> **When TW=0, the WFI instruction may execute in modes less privileged than M when not prevented for some other reason.** When TW=1, then if WFI is executed in any less-privileged mode, and it does not complete within an implementation-specific, bounded time limit, the WFI instruction causes an illegal-instruction exception.
> [...]
> When S-mode is implemented, then executing WFI in U-mode causes an illegal-instruction exception, regardless of the value of the TW bit, unless the instruction completes within an implementation-specific, bounded time limit.

Source: https://github.com/riscv/riscv-isa-manual/blob/b60694bbfcc362fe89d007028513b087f955656b/src/priv/machine.adoc?plain=1#L751

Did I miss something here?

pmundkur:
According to the spec,

> WFI is available in all privileged modes, and optionally available to U-mode.

Add a configuration parameter for this case, and set it to false by default for backwards compatibility.

Fixes #1739.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -361,7 +361,10 @@
     "clock_frequency": 1000000000,
     "instructions_per_tick": 2,
     "wfi_is_nop": false,
-    // WFI is optionally available to User mode.
+    // WFI is optionally available to User mode.  Note that even if
+    // `wfi_is_nop` is set to `true` above, it will still not be
+    // available to User mode unless `wfi_available_to_user_mode` is
+    // also `true`.
     "wfi_available_to_user_mode": false,
     // The maximum increment in the time CSR before a wait instruction
     // (e.g. WFI, WRS.NTO, WRS.STO) that is not a NOP expires its
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -361,6 +361,8 @@
     "clock_frequency": 1000000000,
     "instructions_per_tick": 2,
     "wfi_is_nop": false,
+    // WFI is optionally available to User mode.
+    "wfi_available_to_user_mode": false,
     // The maximum increment in the time CSR before a wait instruction
     // (e.g. WFI, WRS.NTO, WRS.STO) that is not a NOP expires its
     // wait. It is not possible to wait indefinitely.
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -19,6 +19,9 @@
   - Writable bits of the `mcounteren` CSR can now be specified;
     see `base.mcounteren_writable_bits`. (This was previously controlled
     by `base.writable_hpm_counters`, which was incorrect.)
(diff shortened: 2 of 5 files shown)


pmundkur:
I'm not sure about the `wfi_is_nop=true` and `wfi_available_to_user_mode=false` case, since it would imply a nop not available to user mode.

nadime15:
Seems like we first check whether U-Mode can execute WFI (wfi_available_to_user_mode) and then whether it should behave as a NOP. So if someone uses your config choice, it would rightfully throw an illegal instruction exception (which is what I would expect in this case). It's probably worth leaving a comment above the wfi_is_nop option to make this clear.

pmundkur:
Added a comment to clarify.

Timmmm (approved):

nadime15 (approved):

jordancarlin (approved):

PR #1743, opened 2026-06-03
Support configuring legal values for `mstatus.{FS,VS}`.

### LINKED ISSUE #1722: Allow for read only 0 FS and VS bits in mstatus
The spec allows for `mstatus.FS` to be read only 0 if `F` extension is not enabled, but sail does not currently support the behaviour.
Similar clause would be for `mstatus.VS`.

pmundkur:
This uses an enum to support three likely configurations for these extension context states:

- `Off`: implying read-only zero (or the set `{Off}`)
- `TwoState`: containing `{Off, Dirty}`
- `FourState`: containing all the possible states `{Off, Initial, Clean, Dirty}`

since these seem the most likely implementation choices.

It handles the following clauses of the spec:

> If S-mode is implemented but the F extension is not, FS may optionally be read-only zero.

> If S-mode is implemented but the v registers are not, VS may optionally be read-only zero.

It also moves the supported extension validation checks to `validate_config` where it is done once, and then CSR value legalization is just done against the configured enum above.

This fixes #1722.

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -91,6 +91,9 @@ let sys_enable_writable_fiom : bool = config base.writable_fiom
 // Which HPM counters are supported (as a bit mask). Bits [2 .. 0] are ignored.
 let sys_writable_hpm_counters : bits(32) = config base.writable_hpm_counters
 
+// The mstatus.FS legal values list.
+let mstatus_fs_legal_values : list(int) = config base.mstatus.mstatus_fs_legal_values
+
 // Supervisor timecmp
 function clause currentlyEnabled(Ext_Sstc) = hartSupports(Ext_Sstc)
 
@@ -149,6 +152,18 @@ function virtual_memory_supported() -> bool = {
 // This measure needs to be > currentlyEnabled_measure(Ext_Sv{32,39,48,57}),
 // currently all equal to 2, see currentlyEnabled_measure in extensions.sail
 termination_measure virtual_memory_supported(_) = 3
+
+function match_fs_legal_values(legal_values : list(int), v : int) -> bool = {
+  match legal_values {
+    [||] => false,
+    value :: rest => (value == v) | match_fs_legal_values(rest, v)
+  }
+}
+
+function is_vs_read_only() -> bool = {
+  let mstatus_vs_writable : bool = config base.mstatus.mstatus_vs_writable;
+  not(hartSupports(Ext_Zve32x)) | (not(hartSupports(Ext_S)) & not(hartSupports(Ext_V)) & mstatus_vs_writable)
+}
 
 //
 // Illegal values legalized to least privileged mode supported.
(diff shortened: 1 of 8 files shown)


pmundkur:
Adapted from #1413.

model/postlude/validate_config.sail
```diff
@@ -567,15 +567,53 @@ private function check_stateen_config() -> bool = {
...
+  };
+
+  let legal_vs : ExtContext = config base.mstatus.vs_legal_states;
+  if vector_support_level >= Integer then {
+    if legal_vs == ExtContext_Off then {
+      valid = false;
+      print_endline("The vector registers are enabled but `base.mstatus.vs_legal_states` is set to `ExtContext_Off`, which makes `mstatus.VS` read-only zero: it cannot be read-only zero.");
+    }
+  } else if not(config extensions.S.supported : bool) then {
+    if legal_fs != ExtContext_Off then {
```
```suggestion
    if legal_vs != ExtContext_Off then {
```

model/core/types.sail
```diff
@@ -365,6 +365,14 @@ mapping extStatus_bits : ExtStatus <-> ext_status = {
 function extStatus_to_bits(e : ExtStatus) -> ext_status = extStatus_bits(e)
 function extStatus_of_bits(b : ext_status) -> ExtStatus = extStatus_bits(b)
 
+function is_extStatus_valid(ctx : ExtContext, status : ExtStatus) -> bool = {
+  match ctx {
+    ExtContext_Off       => status == Off,
+    ExtContext_TwoState  => status == Off | status == Dirty,
```
nadime15: Is there actually a use case for TwoState? And would it make sense to implement TwoState case like this:
```suggestion
    ExtContext_TwoState  => if extStatus_of_bits(v) == Off
                            then extStatus_to_bits(Off)
                            else extStatus_to_bits(Dirty),
```

I think this is spec compliant and would allow software that relies on the Initial/Clean states to enable the FP/V unit to continue working. Otherwise its somewhat hard to imagine what kind of software would actually make use of this behavior?

I am aware of lazy context switching, but my understanding is that this assumes the full 4-state model, where the FP/V unit can be enabled via the Initial/Clean states. If "traditional" software cannot enable FP in this 2 state case, it seems to imply that software could unexpectedly trap even though misa[F] is set (Im just guessing here).

pmundkur: Your suggested legalization makes sense and is explicitly called out in the spec, so I used it.
I don't know how much software can rely on though, since these are WARL fields after all.

nadime15: Yeah these are WARL, but software is written with that in mind. it does not rely on reading back the exact value it wrote. Both linux and xvisor enable FP by writing initial, without assuming it stays initial. Linux only ever reads FS to check for dirty/off (Xvisor also writes dirty instead off initial directly in its save/restore path).

model/postlude/validate_config.sail
```diff
@@ -567,15 +567,53 @@ private function check_stateen_config() -> bool = {
...
 
+private function check_mstatus_fields() -> bool = {
+  var valid : bool = true;
+
+  let legal_fs : ExtContextPolicy = config base.mstatus.fs_legal_states;
+  if (config extensions.Zfinx.supported : bool) then {
+    if legal_fs != ExtContext_Off then {
+      valid = false;
+      print_endline("The Zfinx extension is enabled but `base.mstatus.fs_legal_states` is not set to `ExtContext_Off`: `mstatus.FS` needs to be read-only zero (i.e. `base.mstatus.fs_legal_states` should be set to `ExtContext_Off`).")
+    }
```
nadime15, line 584: Up to you if you wanna use plat_mstatus_legal_fs/vs (just wanted to point it out) 
```suggestion
  let legal_fs : ExtContextPolicy = plat_mstatus_legal_fs;
     ...
   }
```

pmundkur: That was deliberate.  The goal is to have `validate_config.sail` stand as much on its own as possible, with minimal reference to the rest of the model, so that the parameter and config validation could be synched with the minimal validation language that was being discussed in the Parameter SIG.   I don't know how far that language will go, but lets try to keep this file as self-contained as possible.  There are still usages of `hartSupports` that it would be good to remove eventually.

nadime15: Ahh ok I see, makes sense.

nadime15 (approved):
Looks good!

model/core/sys_regs.sail
```diff
@@ -227,29 +227,25 @@ private function legalize_mstatus(o : Mstatus, v : bits(64)) -> Mstatus = {
...
     SUM = if virtual_memory_supported() then v[SUM] else 0b0,
     MPRV = if currentlyEnabled(Ext_U) then v[MPRV] else 0b0,
+
     // We don't have any extension context yet.
-    XS = extStatus_to_bits(Off),
-    // "If neither the F extension nor S-mode is implemented, then FS is read-only zero.
-    // If S-mode is implemented but the F extension is not, FS may optionally be read-only zero."
-    // TODO: This should be made a platform parameter for the latter case.
-    FS = if hartSupports(Ext_Zfinx) | (not(hartSupports(Ext_F)) & not(currentlyEnabled(Ext_S))) then 0b00 else v[FS],
+    XS = extStatus_map(Off),
```
jordancarlin, line 232: While we are here, should we also add support for writing `mstatus.XS`? While any hardware the writes the bits are custom, the bits themselves are not custom and the model should be able to match software writes to the CSR. We can leave the default for it as `ExtContext_Off`.

jordancarlin (approved):

model/core/sys_regs.sail
pmundkur: Let's keep it as it is.

> In harts without additional user extensions requiring new state, the XS field is read-only zero. 

and we don't have additional user extensions.

Timmmm (approved):

PR #1744, opened 2026-06-05
Enforce stricter vstart bounds checking

### LINKED ISSUE #1104: Stricter checking of `vstart` bounds
The model currently uses 16 bits for the `vstart` value.  The spec (Vector Start Index (vstart) Register) requires a stricter check:

```
The use of vstart values greater than the largest element index for the current vtype setting is reserved.
It is recommended that implementations trap if vstart is out of bounds. It is not
required to trap, as a possible future use of upper vstart bits is to store imprecise
trap information.
```

The model currently does not have an option to trap if vstart is written to with reserved values.

nadime15:
The specification states that a `vstart` greater than the last valid element index is reserved. Vector instructions now can raise an illegal instruction in this case.

To support this, `illegal_vstart` requires the number of elements the instruction operates on. `get_num_elem` cannot be used directly as the bound, because for fractional LMUL (less than 1) it returns a value larger than the real limit. The `EMUL` is therefore passed as well and used to derive the correct bound. For `LMUL` of 1 or greater the value is already correct and behavior is unchanged.

The element-group crypto instructions (Zvkned, Zvknha/b, Zvksed, Zvksh, Zvkg) perform this check in their shared decode predicate (`zvk_check_encdec`) rather than in `illegal_vstart`. They continue to permit a non-zero `vstart` provided it is aligned to an element group.

The specification recommends but does not require trapping, so the behavior is configurable via the new option
`extensions.V.reserved_behavior.vstart_out_of_bounds`. The default `Vstart_Illegal` raises an illegal instruction exception, while `Vstart_Ignore` lets the instruction run as a no-op that writes no elements.

Follow up PR for https://github.com/riscv/sail-riscv/pull/1732 and fixes https://github.com/riscv/sail-riscv/issues/1104.

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -34,6 +34,15 @@
     when `vstart` is non-zero. The default value of true introduces a backward-
     incompatible change compared to previous versions, but can be changed to
     preserve the earlier behaviour.
+  - New option `extensions.V.reserved_behavior.vstart_out_of_bounds`
+    (default `Vstart_Illegal`) controls how an out-of-bounds `vstart`
+    (one greater than the largest element index for the current `vtype`,
+    `vstart` >= VLMAX) is handled. This applies to all vector
+    instructions, including the vector crypto extensions. `Vstart_Illegal`
+    raises an illegal instruction exception, `Vstart_Ignore` treats the
+    instruction as a no-op that writes no elements. The model now enforces
+    this bound by default, correctly accounting for fractional LMUL, which
+    introduces a backward-incompatible change compared to previous versions.
   - The handling of attempts to set an unsupported or reserved `vtype`
     can now be configured; see `extensions.V.reserved_behavior.illegal_vtype`.
   - The supported values for `mstatus.FS` and `mstatus.VS` can be
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -36,8 +36,8 @@
     preserve the earlier behaviour.
   - New option `extensions.V.reserved_behavior.vstart_out_of_bounds`
     (default `Vstart_Illegal`) controls how an out-of-bounds `vstart`
-    (one greater than the largest element index for the current `vtype`,
-    `vstart` >= VLMAX) is handled. This applies to all vector
+    (`vstart` >= VLMAX, i.e. one greater than the largest element index
+    for the current `vtype`) is handled. This applies to all vector
     instructions, including the vector crypto extensions. `Vstart_Illegal`
     raises an illegal instruction exception, `Vstart_Ignore` treats the
     instruction as a no-op that writes no elements. The model now enforces
--- a/model/core/platform_config.sail
+++ b/model/core/platform_config.sail
@@ -173,6 +173,11 @@ enum IllegalVtypeReservedBehavior = {
   IllegalVtype_Fatal,
 }
 
+enum VstartReservedBehavior = {
+  Vstart_Illegal,
+  Vstart_Ignore,
+}
+
 let amocas_odd_register_reserved_behavior : AmocasOddRegisterReservedBehavior = config base.reserved_behavior.amocas_odd_register
 let fcsr_rm_reserved_behavior : FcsrRmReservedBehavior = config base.reserved_behavior.fcsr_rm
 let pmp_write_only_reserved_behavior : PmpWriteOnlyReservedBehavior = config base.reserved_behavior.pmpcfg_write_only
@@ -180,3 +185,4 @@ let xenvcfg_cbie_reserved_behavior : XenvcfgCbieReservedBehavior = config base.r
 let xtvec_mode_reserved_behavior : XtvecModeReservedBehavior = config base.reserved_behavior.xtvec_mode
 let rv32zdinx_odd_register_reserved_behavior : RV32ZdinxOddRegisterReservedBehavior = config base.reserved_behavior.rv32zdinx_odd_register
 let illegal_vtype_reserved_behavior : IllegalVtypeReservedBehavior = config extensions.V.reserved_behavior.illegal_vtype
+let vstart_reserved_behavior : VstartReservedBehavior = config extensions.V.reserved_behavior.vstart_out_of_bounds
--- a/model/core/platform_config.sail
+++ b/model/core/platform_config.sail
@@ -173,7 +173,7 @@ enum IllegalVtypeReservedBehavior = {
   IllegalVtype_Fatal,
 }
 
-enum VstartReservedBehavior = {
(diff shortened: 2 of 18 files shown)


nadime15:
We now strictly enforce a trap when vstart >= VLMAX. Should I add a config option to make this trap optional? With it off, an out-of-bounds vstart would just be a no-op the instruction writes nothing instead of trapping.

pmundkur:
@nadime15 could you push a rebased version so that the tests can run?

model/extensions/V/vext_utils_insts.sail
```diff
@@ -47,12 +48,23 @@ let vstart_scalar_move_zero_required : bool = config extensions.V.vstart.zero_re
...
-    VSTART_ARITH       => vstart_arith_zero_required & not(assert_vstart(0)),
-    VSTART_LOAD_STORE  => false,
-    VSTART_SCALAR_MOVE => vstart_scalar_move_zero_required & not(assert_vstart(0)),
-    VSTART_MANDATORY   => not(assert_vstart(0))
+//
+// num_elem is what get_num_elem returned, and EMUL_pow is the EMUL used
+// to compute it. A vstart past the largest element index (VLMAX - 1) is
+// reserved. We need `EMUL_pow` to work out VLMAX when LMUL is fractional.
+val illegal_vstart : forall 'n 'p, 'n >= 0. (vstart_class, int('n), int('p)) -> bool
+function illegal_vstart(class, num_elem, EMUL_pow) = {
```
Timmmm: I think this is equivalent to this...

```suggestion
function illegal_vstart(class : vstart_class, num_elem : nat, EMUL_pow : int) = {
```

But we can probably use tighter types than `nat` and `int`.

model/extensions/V/vext_utils_insts.sail
```diff
@@ -47,12 +48,23 @@ let vstart_scalar_move_zero_required : bool = config extensions.V.vstart.zero_re
...
+  let vs = unsigned(vstart);
+  // For fractional LMUL, get_num_elem returns a whole register of elements,
+  // which is bigger than VLMAX, so divide to get VLMAX. When EMUL_pow >= 0,
+  // num_elem already equals VLMAX and we use it as-is.
+  let VLMAX = if EMUL_pow >= 0 then num_elem else num_elem / (2 ^ (0 - EMUL_pow));
+  match class {
+    // Out-of-bounds vstart is reserved, in-bounds non-zero traps per config.
+    VSTART_ARITH       => vs >= VLMAX | (vstart_arith_zero_required & vs !=0 ),
```
Timmmm:
```suggestion
    VSTART_ARITH       => vs >= VLMAX | (vstart_arith_zero_required & vs != 0),
```

model/extensions/V/vext_utils_insts.sail
```diff
@@ -47,12 +48,23 @@ let vstart_scalar_move_zero_required : bool = config extensions.V.vstart.zero_re
 //                  all reductions, vcompress, vcpop.m, vfirst.m,
 //                  vmsbf/sif/sof.m, and viota.m. Always traps, not
 //                  configurable.
-function illegal_vstart(cls : vstart_class) -> bool = {
-  match cls {
-    VSTART_ARITH       => vstart_arith_zero_required & not(assert_vstart(0)),
```
Timmmm, line 52: Not in this PR, but `assert_vstart` is a weird function. It doesn't actually assert anything and it's literally just `vstart == 0`. We should remove it an inline the comparisons.

nadime15: Ah yeah, good point.

model/extensions/V/vext_utils_insts.sail
nadime15: Ah cool, much simpler!

pmundkur:
> We now strictly enforce a trap when vstart >= VLMAX. Should I add a config option to make this trap optional? With it off, an out-of-bounds vstart would just be a no-op the instruction writes nothing instead of trapping.

Yes, I think we should provide an option under  "extensions.V.reserved_behavior".  We could move the existing "extensions.V.illegal_vtype" to "extensions.V.reserved_behavior.illegal_vtype".

nadime15:
I added a new config option, extensions.V.reserved_behavior.vstart_out_of_bounds. I have not moved extensions.V.illegal_vtype yet. Do we want to do that in this PR or leave it for a follow up PR?

model/core/platform_config.sail
```diff
@@ -173,10 +173,16 @@ enum IllegalVtypeReservedBehavior = {
   IllegalVtype_Fatal,
 }
 
+enum VstartReservedBehavior = {
```
pmundkur: We could probably be more specific about this being out-of-bounds, perhaps just use `OOB` and explain in a comment?  Maybe something like
```suggestion
// Handling of out-of-bounds (OOB) vstart
enum OOBVStartReservedBehavior = {
```

doc/ChangeLog.md
```diff
@@ -34,6 +34,15 @@
     when `vstart` is non-zero. The default value of true introduces a backward-
     incompatible change compared to previous versions, but can be changed to
     preserve the earlier behaviour.
+  - New option `extensions.V.reserved_behavior.vstart_out_of_bounds`
+    (default `Vstart_Illegal`) controls how an out-of-bounds `vstart`
+    (one greater than the largest element index for the current `vtype`,
+    `vstart` >= VLMAX) is handled. This applies to all vector
```
pmundkur: To me this is clearer if written as
```suggestion
    (`vstart` >= VLMAX, i.e. one greater than the largest element index
    for the current `vtype`) is handled. This applies to all vector
```

pmundkur (commented):
Nice!  Minor nits, but otherwise LGTM.

pmundkur:
> I added a new config option, extensions.V.reserved_behavior.vstart_out_of_bounds. I have not moved extensions.V.illegal_vtype yet. Do we want to do that in this PR or leave it for a follow up PR?

Let's do it in a follow up.

pmundkur (approved):

pmundkur:
@nadime15 are you working on the `extensions.V.illegal_vtype` -> `extensions.V.reserved_behavior.illegal_vtype` move or should I take care of it?

nadime15:
Yeah I started, I am going to open a PR in a few minutes.

pmundkur:
I'd like to merge this in a few hours and get a release out by the end of the day...

nadime15:
I'll read through the code again in 2–3 hours and do a final check. If I don't change anything, just assume it's good on my end. :D

pmundkur:
I think this will need rebasing on top of #1746 when that merges, due to conflicts.

PR #1745, opened 2026-06-05
Simplify the `reinit_sail()` API.

pmundkur:
`reinit_sail()` should not have to provide again the same parameters that were already provided in `init_sail()`.  This allows an initialized model to be reinitialized without having to know the initialization parameters.


Timmmm (approved):
LGTM. These could potentially be constructor arguments and then you only have a single `init_sail()` method.

PR #1752, opened 2026-06-08
Reject vector segment loads/stores that increment past v31

### LINKED ISSUE #1748: Segmented Loads Going Past v31 are Reserved
For segmented loads and stores, the spec says that “If the vector register numbers accessed by the segment load or store would increment past 31, then the instruction encoding is reserved.” This means that if we have SEW = 8, and we execute `vlseg2e32 v28, (t0), v0.t`, we should deduce that because nf = 2 and emul = 4, the load must access up to register 28 + 2*4 = 36, so the encoding is reserved. 

I observe this expected behavior in spike, which has an explicit check for this case (and I have verified that this was the cause of the illegal instruction trap). Here is the spike trace:
```
core   0: 0x80001eac (0x0c007e57) vsetvli t3, zero, e8, m1, ta, ma
core   0: 3 0x80001eac (0x0c007e57) c8_vstart 0x00000000 x28 0x00000010 c3104_vl 0x00000010 c3105_vtype 0x000000c0
core   0: 0x80001eb0 (0xfffe0e13) addi    t3, t3, -1
core   0: 3 0x80001eb0 (0xfffe0e13) x28 0x0000000f
core   0: 0x80001eb4 (0x008e1073) csrw    vstart, t3
core   0: 3 0x80001eb4 (0x008e1073) c8_vstart 0x0000000f
core   0: 0x80001eb8 (0x0005b297) auipc   t0, 0x5b
core   0: 3 0x80001eb8 (0x0005b297) x5  0x8005ceb8
core   0: 0x80001ebc (0x18828293) addi    t0, t0, 392
core   0: 3 0x80001ebc (0x18828293) x5  0x8005d040
core   0: 0x80001ec0 (0x2002ee07) vlseg2e32.v v28, (t0), v0.t
core   0: exception trap_illegal_instruction, epc 0x80001ec0
core   0:           tval 0x2002ee07
```

In the same place, Sail does not trap, and instead I observed this trace:
```
[4124] [M]: 0x80001EAC (0x0C007E57) vsetvli x28, x0, e8, m1, ta, ma                 ExceptionsVls_vlseg2e32_v_cg_cp_vstart+156
x28 <- 0x00000020
CSR mstatus (0x300) <- 0x80006680
CSR vstart (0x008) <- 0x00000000
CSR vtype (0xC21) <- 0x000000C0
CSR vl (0xC20) <- 0x00000020
CSR vstart (0x008) <- 0x00000000
     ...
mem[X,0x080001EC0] -> 0x2002EE07
[4129] [M]: 0x80001EC0 (0x2002EE07) vlseg2e32.v v28, (x5), v0.t                     ExceptionsVls_vlseg2e32_v_cg_cp_vstart_duplicate_2+0
CSR mstatus (0x300) <- 0x80006680
v28 <- 0x8F453376F6B54386BBA5E8852A66AE7CC28D3507357191EFB1A6CB509A6C0F13
CSR mstatus (0x300) <- 0x80006680
v0 <- 0x8F453376F6B54386BBA5E8852A66AE7CC28D3507357191EFB1A6CB509A6C0F13
CSR mstatus (0x300) <- 0x80006680
     ...
v31 <- 0x739706D4F6B54386BBA5E8852A66AE7CC28D3507357191EFB1A6CB509A6C0F13
mem[R,0x08005D13C] -> 0x83A9537B
CSR mstatus (0x300) <- 0x80006680
v3 <- 0x83A9537BF68C8AF7D9BBB8263BC3E60C486F78365823937B4D00F0DBD45E0CC9
CSR mstatus (0x300) <- 0x80006680
CSR vstart (0x008) <- 0x00000000
```
Sail doesn't seem to check this case, and instead executes the reserved instruction by wrapping around the register file. While sail is not required to trap on a reserved instruction, this behavior does not seem intended. If it is, it would be very helpful to get a configuration that allows us to have a sail configuration that matches spike on this test. 

For reference, this is the test that I was running [ExceptionsVls_rv32.sig.elf.gz](https://github.com/user-attachments/files/28656471/ExceptionsVls_rv32.sig.elf.gz), this is the sail configuration I was using [sail.json](https://github.com/user-attachments/files/28656511/sail.json), and I observed this on sail 0.11.

nadime15:
Segment load/stores encodings whose register numbers increment past v31 are reserved, but the model silently wrapped around the register file instead of trapping. This adds `valid_segment_no_overflow()` to the (indexed) load and store illegal check functions, so that these raise an illegal instruction exception.

Fixes #1748

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -2,6 +2,7 @@
 
 - Important issues addressed and bugs fixed:
   - https://github.com/riscv/sail-riscv/issues/1750 : Non-segmented indexed loads were trapping on legal overlaps
+  - https://github.com/riscv/sail-riscv/issues/1748 : Segment loads/stores whose register numbers increment past v31 were not treated as reserved.
 
 # Release notes for version 0.12
 
--- a/model/extensions/V/vext_mem_insts.sail
+++ b/model/extensions/V/vext_mem_insts.sail
@@ -66,7 +66,7 @@ function clause execute VLSEGTYPE(nf, vm, rs1, width, vd) = {
   if illegal_vstart(VSTART_LOAD_STORE, num_elem, EMUL_pow) | illegal_load(vd, vm, nf, EEW, EMUL_pow) | not(valid_reg_group(vd, EMUL_pow))
   then return Illegal_Instruction();
 
-  let EMUL_reg = if EMUL_pow <= 0 then 1 else 2 ^ EMUL_pow;
+  let EMUL_reg = 2 ^ max(EMUL_pow, 0);
   let vm_val = read_vmask(num_elem, vm, zvreg);
   let vd_seg = read_vreg_seg(num_elem, load_width_bytes * 8, EMUL_pow, nf, vd);
 
@@ -123,7 +123,7 @@ function clause execute VLSEGFFTYPE(nf, vm, rs1, width, vd) = {
   if illegal_vstart(VSTART_LOAD_STORE, num_elem, EMUL_pow) | illegal_load(vd, vm, nf, EEW, EMUL_pow) | not(valid_reg_group(vd, EMUL_pow))
   then return Illegal_Instruction();
 
-  let EMUL_reg = if EMUL_pow <= 0 then 1 else 2 ^ EMUL_pow;
+  let EMUL_reg = 2 ^ max(EMUL_pow, 0);
   let vm_val = read_vmask(num_elem, vm, zvreg);
   let vd_seg = read_vreg_seg(num_elem, load_width_bytes * 8, EMUL_pow, nf, vd);
   let tail_ag = get_vtype_vta();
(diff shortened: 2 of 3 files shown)


pmundkur (commented):
LGTM other than some naming issues.

The pre-existing `valid_segment()` is now misleading since it is only doing a partial check on whether a segment is valid. Perhaps we could rename that to `valid_segment_group_size()`?  (`segment_group` to take into account all the segments.)

Also, `valid_segment_group_without_overflow()` reads better to me than `valid_segment_no_overflow()` but you don't have to change it if you don't agree.

nadime15:
Yeah I good point, I improved the naming. 

I think that once ACT4 has achieved good coverage of the RVV code and all its edge cases, it would be a good opportunity to clean it up. I think the code could benefit from some cleanup in the future, given all the additions that have accumulated over time.

pmundkur (approved):

PR #1755, opened 2026-06-09
Split out functions in `riscv_sim.cpp` into smaller files.

pmundkur:
This just moves the CLI related code from `riscv_sim.cpp` into `cli_options.{h,cpp}` and the main function to `sail_riscv_sim.cpp` to allow alternative `main()` functions to be more easily written.  This code is just moved without any changes.

Some utility functions call `exit()`; this is not ideal and can be addressed in a later refactor.

### DIFF
--- a/c_emulator/cli_options.h
+++ b/c_emulator/cli_options.h
@@ -1,0 +1,59 @@
+#pragma once
+
+#include <cstdint>
+#include <string>
+#include <vector>
+
+const unsigned DEFAULT_SIGNATURE_GRANULARITY = 4;
+
+struct CLIOptions {
+  bool do_show_times = false;
+  bool do_print_version = false;
+  bool do_print_build_info = false;
+  bool do_print_default_config = false;
+  bool do_print_config_schema = false;
+  bool do_print_dts = false;
+  bool do_validate_config = false;
+  bool do_print_isa = false;
+
+  bool use_rv32_default = false;
+  bool disable_trap_loop_detection = false;
+  std::string config_file = {};
+  std::vector<std::string> config_overrides = {};
+  std::string term_log = {};
+  std::string trace_log_path = {};
+  std::string dtb_file;
+  unsigned rvfi_dii_port = 0;
+  std::vector<std::string> elfs;
+  uint64_t insn_limit = 0;
+
+  std::string sig_file = {};
+  unsigned signature_granularity = DEFAULT_SIGNATURE_GRANULARITY;
+
+#ifdef SAILCOV
+  std::string sailcov_file = {};
+#endif
+
+  bool config_print_instr = false;
+  bool config_print_gpr = false;
+  bool config_print_fpr = false;
+  bool config_print_vreg = false;
+  bool config_print_csr = false;
+  bool config_print_mem_access = false;
+  bool config_print_clint = false;
+  bool config_print_exception = false;
+  bool config_print_interrupt = false;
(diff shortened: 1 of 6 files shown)


arichardson (changes requested):
Overall the result seems good to me but I think doing this as one change will confuse git.


This is not great for git blame or any downstreams. How about just pulling out the main function into a separate sim_main.cpp (or some other name) and keeping this file as is?

And if you want to rename it that could be done as a separate follow up that just does the renaming so git doesn't get confused.

Timmmm (commented):
I also think the CLI stuff (`CLIOptions` and `parse_cli`) could go in its own file. I was planning to do that. It's nicely isolated.

pmundkur:
>  How about just pulling out the main function into a separate sim_main.cpp (or some other name) and keeping this file as is?
> 

Good idea, done.

> I also think the CLI stuff (`CLIOptions` and `parse_cli`) could go in its own file. I was planning to do that. It's nicely isolated.

Also a good idea, done.

Although now the diff of `riscv_sim.cpp` is harder to read.

c_emulator/cli_options.h
```diff
@@ -0,0 +1,57 @@
+#include <cstdint>
```
Timmmm, line 3:
```suggestion
#pragma once

#include <cstdint>
```

c_emulator/riscv_sim.h
```diff
@@ -0,0 +1,77 @@
+#include "cli_options.h"
```
Timmmm:
```suggestion
#pragma once

#include "cli_options.h"
```

Timmmm (commented):
LGTM other than the missing `#pragma once`s. It would be good to preserve Git blame but that's tricky in a single commit so tbh I wouldn't worry too much about it.

Timmmm (approved):

arichardson (commented):
I know this is annoying busy work, but could you split the extracting functions and moving functions around within one file onto two separate commits? I'm struggling to review the riscv_sim.cpp changes here since it's not obvious if anything changed in this large diff. 

Result looks good to me though

pmundkur:
> I know this is annoying busy work, but could you split the extracting functions and moving functions around within one file onto two separate commits? I'm struggling to review the riscv_sim.cpp changes here since it's not obvious if anything changed in this large diff.
> 

No worries, thanks for taking the time to review this!

Timmmm:
> I'm struggling to review the riscv_sim.cpp changes here since it's not obvious if anything changed in this large diff.

I've spend ages looking for a nice diff tool that will show moved/modified lines between files and it basically doesn't exist. Kind of blows my mind because it must be such a common thing to want to do!

The only things I've seen that even tries is Phabricator which will show moved lines in its diff view, and `git diff --color-moved` (which you can make more readable like [this](https://gist.github.com/0xdevalias/1245fe66ac6953d490114b417a0075e3)) which is a lot better than nothing but you're still reading raw diffs in the terminal which isn't exactly fun. :-/

arichardson (approved):
I haven't checked that the moved functions are line by line the same but I trust that this is correct. LGTM now. It might be good to have both the move+rearrange commits in the history (i.e. two separate PRs that get merged immediately) but I'll let you decide how best to handle this.

pmundkur:
Sounds good; I'll put the rearrangement into a separate PR for a cleaner history.

PR #1757, opened 2026-06-11
Add support for VLEN=64 in config generation and vector tests.

pmundkur:
VLEN=64 is not included under the full V extension and so provides an additional test point.

Remove testing of the VLEN=512/ELEN=32 combination since the test suite for this combination takes around 6 hours to generate which often exceeds the maximum Github CI execution time limit (6 hours).

The VLEN=64 suites are in the latest `sail-riscv-tests` release:
```
$ ./compare-releases.py -p releases/2026-05-29 -c releases/2026-06-10
riscv-tests has 667 test files, with 0 added to and 0 removed from the previous release.
riscv-arch-tests has 2080 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v64x32 with 1941 test files was added.
riscv-vector-tests-v64x64 with 3025 test files was added.
riscv-vector-tests-v128x32 has 1944 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v128x64 has 3042 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v256x32 has 1945 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v256x64 has 3043 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v512x32 with 1945 test files was removed.
riscv-vector-tests-v512x64 has 3043 test files, with 0 added to and 0 removed from the previous release.
```

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -84,7 +84,7 @@ endfunction()
 # supported by the RISC-V Vector Test suite.
 foreach(vlen IN ITEMS 64 128 256 512)
     foreach(elen IN ITEMS 32 64)
-        if((${vlen} STREQUAL 512) AND (${elen} STREQUAL 32))
+        if((${vlen} EQUAL 512) AND (${elen} EQUAL 32))
             # The test suite for this combination is not built due to
             # its long execution time.
             continue()
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -84,7 +84,7 @@ endfunction()
 # supported by the RISC-V Vector Test suite.
 foreach(vlen IN ITEMS 64 128 256 512)
     foreach(elen IN ITEMS 32 64)
-        if(("$vlen" STREQUAL "512") AND ("$elen" STREQUAL "32"))
+        if(("${vlen}" STREQUAL "512") AND ("${elen}" STREQUAL "32"))
             # The test suite for this combination is not built due to
             # its long execution time.
             continue()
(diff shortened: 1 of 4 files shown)


test/CMakeLists.txt
```diff
@@ -82,8 +82,13 @@ endfunction()
...
 # NOTE: Setting XLEN independently of ELEN is not
 # supported by the RISC-V Vector Test suite.
-foreach(vlen IN ITEMS 128 256 512)
+foreach(vlen IN ITEMS 64 128 256 512)
     foreach(elen IN ITEMS 32 64)
+        if(("$vlen" STREQUAL "512") AND ("$elen" STREQUAL "32"))
+            # The test suite for this combination is not built due to
+            # its long execution time.
+            continue()
+        endif()
```
nadime15, line 91:
```suggestion
        if(vlen EQUAL 512 AND elen EQUAL 32)
            # The test suite for this combination is not built due to
            # its long execution time.
            continue()
```

I tried running it locally with (-DENABLE_RISCV_VECTOR_TESTS_V512_E32=TRUE) and it did not skip this case. You compared the literal "$vlen" to "512".

pmundkur: Thanks for testing this.  Fixed.

nadime15: Sure. Out of curiosity, is there a reason why you perform a string comparison instead of the integer comparison? This seems to be cleaner:

`if(vlen EQUAL 512 AND elen EQUAL 32)`

pmundkur: No reason other than it worked :-)  Agreed that integer comparison makes more sense.  The CMake language seems quite weird.

nadime15 (approved):

test/CMakeLists.txt
nadime15: > The CMake language seems quite weird.

Haha yeah I agree :D

Timmmm: Yeah it's quite bad. The reason there's STREQUAL and EQUAL is that it's stringly typed. `100 EQUAL 100.0` will be true but `100 STREQUAL 100.0` won't.

pmundkur: So perhaps I should use EQUAL instead of STREQUAL, or should I go back to doing a string comparison?

Timmmm: I think EQUAL makes the most sense when you're using numbers.

Timmmm (approved):

PR #1758, opened 2026-06-11
Add build rule for HOL4 theorem prover

bacam:
This includes a small tweak to the handwritten Lem file to sidestep a an issue with the Sail library.

It requires the current development versions of Sail and Lem to build.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -565,4 +565,36 @@ foreach (xlen IN ITEMS 32 64)
             "${CMAKE_BINARY_DIR}/isabelle/${arch}/ROOT"
     )
     add_custom_target(generated_isabelle_${arch} DEPENDS "${CMAKE_BINARY_DIR}/isabelle/${arch}/${arch_thy}.thy")
+
+    ############# HOL4 #############
+
+    # Build HOL4 definitions from Lem
+    file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/hol4/${arch}")
+    configure_file("${CMAKE_SOURCE_DIR}/handwritten_support/Holmakefile.in" "${CMAKE_BINARY_DIR}/hol4/${arch}/Holmakefile" @ONLY)
+    add_custom_command(
+        DEPENDS
+          "${arch}.lem"
+          "${arch}_types.lem"
+          "${CMAKE_SOURCE_DIR}/handwritten_support/riscv_extras.lem"
+          "${CMAKE_SOURCE_DIR}/handwritten_support/riscv_extras_fdext.lem"
+        VERBATIM
+        OUTPUT
+          "${CMAKE_BINARY_DIR}/hol4/${arch}/${arch}Script.thy"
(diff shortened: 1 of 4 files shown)


pmundkur (approved):

jordancarlin (approved):
LGTM. It would be good to get more of the theorem prover targets into CI.

bacam:
Doing full builds for HOL4 and Isabelle in github CI might not work because they don't cope well with large data structures and use a disproportional amount of time and memory.  However, I am thinking of checking that we can produce the prover output (i.e., running Sail and Lem), and packing up the results as a artifact so that users don't have to have a full Sail install.

model/CMakeLists.txt
```diff
@@ -565,4 +565,43 @@ foreach (xlen IN ITEMS 32 64)
...
+            ${CMAKE_SOURCE_DIR}/handwritten_support/riscv_extras.lem
+            ${CMAKE_SOURCE_DIR}/handwritten_support/riscv_extras_fdext.lem
+            ${arch}_types.lem
+            ${arch}.lem
+        COMMAND
+            sed
+              -e "s/ARCH/${arch}/g"
+               ${CMAKE_SOURCE_DIR}/handwritten_support/Holmakefile.in
+               >
+               ${CMAKE_BINARY_DIR}/hol4/${arch}/Holmakefile
```
Timmmm: We could probably use `configure_file()` for this maybe? If you search the repo there are a few examples.

Timmmm (commented):
LGTM apart from the one comment.

model/CMakeLists.txt
bacam: Thanks.  I thought that `configure_file` was only applied in the initial configuration, but trying it out I see that changes to the source file still get picked up at build.

model/CMakeLists.txt
```diff
@@ -565,4 +565,37 @@ foreach (xlen IN ITEMS 32 64)
...
+          "${arch}_types.lem"
+          "${CMAKE_SOURCE_DIR}/handwritten_support/riscv_extras.lem"
+          "${CMAKE_SOURCE_DIR}/handwritten_support/riscv_extras_fdext.lem"
+        VERBATIM
+        OUTPUT
+          "${CMAKE_BINARY_DIR}/hol4/${arch}/${arch}Script.thy"
+          "${CMAKE_BINARY_DIR}/hol4/${arch}/${arch}_typesScript.thy"
+          "${CMAKE_BINARY_DIR}/hol4/${arch}/riscv_extrasScript.thy"
+          "${CMAKE_BINARY_DIR}/hol4/${arch}/riscv_extras_fdextScript.thy"
+          "${CMAKE_BINARY_DIR}/hol4/${arch}/Holmakefile"
```
Timmmm: I guess this file should be in `DEPENDS` now?

bacam: This rule doesn't actually invoke HOL, so we don't need the `Holmakefile` in `DEPENDS`.  However, it shouldn't be in `OUTPUT`, so I'll just fix that.

Timmmm (approved):

Issue #1006, opened 2025-06-06
Vector load instructions result in repeated output to the log


jordancarlin:
Vector load instructions cause a very verbose and repetitive output to the log. For example, executing a `vle8.v` instruction shows `mstatus` and `v1` being updated 64 times each with the same values every time. This quickly blows up the log file and makes parsing it much more difficult than it should be.

@nadime15 I know you've been running a lot of vector tests. Have you run into this?

```
...
[636] [M]: 0x800009D8 (0x02030087) vle8.v v1, (x6)
CSR mstatus (0x300) <- 0x0000000080000600
CSR mstatush (0x310) <- 0x0000000000000000
CSR vstart (0x008) <- 0x0000000000000000
mem[R,0x08002E620] -> 0x81
CSR mstatus (0x300) <- 0x0000000080000600
CSR mstatush (0x310) <- 0x0000000000000000
v1 <- 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000081
CSR mstatus (0x300) <- 0x0000000080000600
CSR mstatush (0x310) <- 0x0000000000000000
v1 <- 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000081
CSR mstatus (0x300) <- 0x0000000080000600
CSR mstatush (0x310) <- 0x0000000000000000
v1 <- 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000081
CSR mstatus (0x300) <- 0x0000000080000600
CSR mstatush (0x310) <- 0x0000000000000000
v1 <- 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000081
CSR mstatus (0x300) <- 0x0000000080000600
CSR mstatush (0x310) <- 0x0000000000000000
v1 <- 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000081
CSR mstatus (0x300) <- 0x0000000080000600
CSR mstatush (0x310) <- 0x0000000000000000
v1 <- 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000081
CSR mstatus (0x300) <- 0x0000000080000600
CSR mstatush (0x310) <- 0x0000000000000000
[...]

PR #1759, opened 2026-06-11
Correctly update vxsat in vssubu and add vxsat to trace when changed

### LINKED ISSUE #1069: VSSUBU doesn't set vxsat
This code is missing `vcsr[vxsat] = 0b1;`

```
        VV_VSSUBU        => {
                              if unsigned(vs2_val[i]) < unsigned(vs1_val[i]) then zeros()
```

That code looks a bit suss to be honest - why is `VV_VSSUBU` so different to `VV_VSSUB`?

Same for `VX_VSSUBU`.

fourth-bit:
In vssubu, when the value in vs2 was greater than rs1 or the value in vs1, the previous behavior was to clamp the result to zero and return. The spec says: "If the result would overflow the destination, the result is replaced with the closest representable value, and the vxsat bit is set," so vxsat should have been set in these cases. This PR sets vxsat in these cases. Fixes #1069.

This PR also adds vxsat updates to the trace. Previously, the only updates to vxsat that were logged in the trace were updates in csr write instructions. The addition of vxsat to the trace is modeled after the way that fflags handled, in that every instruction that can update these csrs issues a write callback for both vcsr and vxsat.

### DIFF
--- a/model/extensions/V/vext_regs.sail
+++ b/model/extensions/V/vext_regs.sail
@@ -346,6 +346,12 @@ private function write_vcsr(vxrm_val : bits(2), vxsat_val : bits(1)) -> unit = {
   dirty_v_context()
 }
 
+function write_vxsat(vxsat_val: bits(1)) -> unit = {
+  write_vcsr(vcsr[vxrm], vxsat_val);
+  csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));
+  csr_write_callback("vcsr", zero_extend(vcsr.bits));
+}
+
 function clause write_CSR(0x008, value) = { set_vstart(value[15 .. 0]); Ok(vstart) }
 function clause write_CSR(0x009, value) = { write_vcsr (vcsr[vxrm], value[0 .. 0]); Ok(zero_extend(vcsr[vxsat])) }
 function clause write_CSR(0x00A, value) = { write_vcsr (value[1 .. 0], vcsr[vxsat]); Ok(zero_extend(vcsr[vxrm])) }
--- a/model/extensions/V/vext_regs.sail
+++ b/model/extensions/V/vext_regs.sail
@@ -347,9 +347,11 @@ private function write_vcsr(vxrm_val : bits(2), vxsat_val : bits(1)) -> unit = {
 }
 
 function write_vxsat(vxsat_val: bits(1)) -> unit = {
-  write_vcsr(vcsr[vxrm], vxsat_val);
-  csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));
-  csr_write_callback("vcsr", zero_extend(vcsr.bits));
+  if vcsr[vxsat] != vxsat_val then {
+    write_vcsr(vcsr[vxrm], vxsat_val);
+    csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));
+    csr_write_callback("vcsr", zero_extend(vcsr.bits));
+  }
 }
 
 function clause write_CSR(0x008, value) = { set_vstart(value[15 .. 0]); Ok(vstart) }
--- a/model/extensions/V/vext_regs.sail
+++ b/model/extensions/V/vext_regs.sail
@@ -348,10 +348,11 @@ private function write_vcsr(vxrm_val : bits(2), vxsat_val : bits(1)) -> unit = {
 
 function write_vxsat(vxsat_val: bits(1)) -> unit = {
   if vcsr[vxsat] != vxsat_val then {
-    write_vcsr(vcsr[vxrm], vxsat_val);
-    csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));
-    csr_write_callback("vcsr", zero_extend(vcsr.bits));
-  }
+    vcsr[vxsat] = vxsat_val;
+    dirty_v_context();
+  };
+  csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));
+  csr_write_callback("vcsr", zero_extend(vcsr.bits));
 }
 
 function clause write_CSR(0x008, value) = { set_vstart(value[15 .. 0]); Ok(vstart) }
--- a/model/extensions/V/vext_utils_insts.sail
+++ b/model/extensions/V/vext_utils_insts.sail
@@ -609,6 +609,7 @@ function unsigned_saturation(len, elem) = {
     write_vxsat(0b1);
     ones('m)
   } else {
+    write_vxsat(vcsr[vxsat]);
     elem['m - 1 .. 0]
   }
 }
@@ -623,6 +624,7 @@ function signed_saturation(len, elem) = {
     write_vxsat(0b1);
     0b1 @ zeros('m - 1)
   } else {
+    write_vxsat(vcsr[vxsat]);
     elem['m - 1 .. 0]
   };
 }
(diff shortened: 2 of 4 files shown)


model/extensions/V/vext_utils_insts.sail
```diff
@@ -606,9 +606,10 @@ function get_fixed_rounding_incr(vec_elem, shift_amount) = {
 val unsigned_saturation : forall ('m 'n: Int), ('n >= 'm > 1). (int('m), bits('n)) -> bits('m)
 function unsigned_saturation(len, elem) = {
   if unsigned(elem) > unsigned(ones('m)) then {
-    vcsr[vxsat] = 0b1;
+    write_vxsat(0b1);
     ones('m)
   } else {
+    write_vxsat(vcsr[vxsat]);
```
nadime15, line 612: The else branches here and below (line 624) now trigger the callback even though no value has changed. Is this intentional? I think the callback should only be triggered when vxsat is actually written. I would remove these calls from the else branches.

model/extensions/V/vext_regs.sail
```diff
@@ -346,6 +346,12 @@ private function write_vcsr(vxrm_val : bits(2), vxsat_val : bits(1)) -> unit = {
   dirty_v_context()
 }
 
+function write_vxsat(vxsat_val: bits(1)) -> unit = {
+  write_vcsr(vcsr[vxrm], vxsat_val);
+  csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));
+  csr_write_callback("vcsr", zero_extend(vcsr.bits));
+}
+
```
nadime15, line 357: I think in the worst case you could trigger the callbacks (VX_VSSUBU and VV_VSSUBU) for every vector element (assuming the if branch is taken every time), even though no new state has been reached after the first write.

Would it make sense to do something like this?
 
```suggestion
function write_vxsat(vxsat_val : bits(1)) -> unit = {
  if vcsr[vxsat] != vxsat_val then {
    write_vcsr(vcsr[vxrm], vxsat_val);
    csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));
    csr_write_callback("vcsr",  zero_extend(vcsr.bits));
  }
}
```

This way, the callbacks are only triggered when vcsr.vxsat actually changes.

model/extensions/V/vext_utils_insts.sail
fourth-bit: Fixed. Triggering the callback on every write, regardless of whether or not there was a change, is the current behavior of fflags. Is that something that should be changed too?

pmundkur: I can't tell whether the sail_to_rvvi script depends on that current behavior for fflags.
https://github.com/riscv/riscv-arch-test/blob/act4/framework/src/act/sail_to_rvvi.py
If it doesn't, we could change that too.

jordancarlin: I'd like to get a native RVVI output in before the next release, so that script will be removed soon anyway. I'd say go ahead with the change to fflags (in a separate PR).

fourth-bit: There is a dependence on this behavior, but it shouldn't depend on it. When sail executes a csrrwi instruction, like is performed to reset fflags, the trace only emits an update to fflags, but not to fcsr which is also written. This means that without fflags and fcsr updating on every relevant floating point operation, the rvvi and coverage implementation associated with it will not be able to correctly track the value fcsr. So, if the accrue_fflags is changed alone, it will break the ACT4 coverage, but if these are changed together it shouldn't break ACT4.

jordancarlin: When fcsr is written the tracer should probably print both fcsr and fflags/frm. Can you open an issue for this?

nadime15: Hmm I was checking the code a bit more, and we have many cases where we call callbacks unconditionally even if the underlying state has not actually changed, such as `dirty_fd_context`() or `set_vstart()`.

I'm not entirely sure what the contract is here. Do we want callbacks to track actual state changes only, or should they be triggered for every write, regardless of whether the state value has changed? 

Because if we start to change it now we would need to change all these cases and probably more.

jordancarlin: Maybe for this PR we should have it log unconditionally, and we can look into a larger refactor after to only log when the state changes.

nadime15: @jordancarlin Yeah, I agree. So I guess matching accrue_fflags logic would turn out to be something like:
```
function write_vxsat(vxsat_val : bits(1)) -> unit = {
  if vcsr[vxsat] != vxsat_val then {
    vcsr[vxsat] = vxsat_val;
    dirty_v_context();
  };
  csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));   // unconditional
  csr_write_callback("vcsr",  zero_extend(vcsr.bits));
}
```

model/extensions/V/vext_regs.sail
```diff
@@ -347,9 +347,11 @@ private function write_vcsr(vxrm_val : bits(2), vxsat_val : bits(1)) -> unit = {
 }
 
 function write_vxsat(vxsat_val: bits(1)) -> unit = {
-  write_vcsr(vcsr[vxrm], vxsat_val);
-  csr_write_callback("vxsat", zero_extend(vcsr[vxsat]));
-  csr_write_callback("vcsr", zero_extend(vcsr.bits));
+  if vcsr[vxsat] != vxsat_val then {
```
Timmmm, line 350: We don't use this behaviour for other identity CSR writes, e.g. see `accrue_fflags()`. It's the same for GPR writes - they get reported even if the value doesn't change.

fourth-bit:
It seems like we've landed on always reporting the values of `vcsr` and `vxsat` regardless of whether or not a change has occurred. I brought back the calls to `write_vxsat` even when nothing changes, so that the values can be reported in all cases, and changed the function to only write to `vcsr` and call `dirty_v_context` when the value changes, which matches the behavior of `accrue_fflags`.

model/extensions/V/vext_regs.sail
```diff
@@ -346,6 +346,15 @@ private function write_vcsr(vxrm_val : bits(2), vxsat_val : bits(1)) -> unit = {
   dirty_v_context()
 }
 
+function write_vxsat(vxsat_val: bits(1)) -> unit = {
+  if vcsr[vxsat] != vxsat_val then {
+    vcsr[vxsat] = vxsat_val;
+    dirty_v_context();
```
pmundkur, line 352: I'm wondering if we need the `dirty_v_context()` here.  It is indeed safer; but the arithmetic instructions from which this is called will always `set_vstart()` which will also call `dirty_v_context()`. 
Having it here will increase the vector log verbosity mentioned in #1006.

pmundkur (commented):
LGTM other than whether the call to `dirty_v_context()` is needed.

model/extensions/V/vext_regs.sail
fourth-bit: It seems clearer to include `dirty_v_context` here. It also doesn't seem like removing it does much to address the verbosity, because running an instruction that calls this function emits at most one extra trace update to mstatus, while there is a trace entry for every element with vcsr and vxsat.

pmundkur: That's true, it's clearer to have it here.

pmundkur (approved):

nadime15 (approved):

jordancarlin (approved):

jordancarlin:
@fourth-bit Looks like this fixes #1069. Can you add an entry to the significant bugs fixed section of `doc/ChangeLog.md` for that?

Issue #1761, opened 2026-06-15
How to fix `mem_write_ea`?


pmundkur:
According to my understanding of a previous discussion https://github.com/riscv/sail-riscv/discussions/392, `mem_write_ea`  is needed for the concurrency model and should be called after the read of the register feeding into the address of the write, and before the read of the register that feeds into the value to be written.  [~Is this understanding correct?~ Updated: From discussion in the golden model meeting, it seems like this is correct.]

The current model does not follow this sequence for `STORE` in `extensions/I/base_insts`: 
```
function clause execute STORE(imm, rs2, rs1, width) = {
  ...
  let data = X(rs2)[width * 8 - 1 .. 0];
  match vmem_write(rs1, offset, width, data, Store(Data), false, false, false) {
  ...
```
i.e. the register `rs2` feeding the value is read before the register `rs1` feeding into the address is read in `vmem_write` (where `mem_write_ea` is called). 

(The model also does not follow it for AMOs in `extensions/A/zaamo_insts`, but a simple re-ordering can fix it there.)

However, `mem_write_ea` terminates in `write_ram_ea` in `core/phys_mem_interface` which was made into a nop for all backends
```
val write_ram_ea : forall 'n, 0 < 'n <= max_mem_access. (write_kind, physaddr, int('n)) -> unit
function write_ram_ea(_wk, Physaddr(_addr), _width) = ()
```
in this commit 59447f06.

Does this mean that `write_mem_ea` can be removed?

If it cannot be removed, then there may be another problem in addition to the `STORE` issue above.  Since `mem_write_ea` is called before `mem_write_value`, and `mem_write_value` does PMP/PMA checks which might prevent the write from being performed, `mem_write_ea` might announce the effective address prematurely unless it does the same PMP/PMA checks that `mem_write_value` does.   Is this premature and possibly erroneous announcement a problem for the concurrency model?  Or should `mem_write_ea` duplicate the same checks that `mem_write_value` does? [Updated: From discussion in the golden model meeting, it seems that `mem_write_ea` and `mem_write_value` should behave consistently and `mem_write_ea` should do the same PMP/PMA checks.]

PR #1767, opened 2026-06-17
Add the `aq` and `rl` flags for LR/SC and AMOs to the access type.

pmundkur:
This merely adds the flags but does not use them.  A subsequent PR will use these flags where needed instead of the three bool arguments in the memory access functions.

The access type is also made an argument for `mem_write_ea` which only looked at the `aq`/`rl` flags; having the full access type will make it possible for `mem_write_ea` and `mem_write` to enforce the same PMP/PMA checks on the access (see #1761).

### DIFF
--- a/model/core/mem_type_utils.sail
+++ b/model/core/mem_type_utils.sail
@@ -12,12 +12,12 @@ function accessFaultFromAccessType(access : MemoryAccessType(mem_payload)) -> Ex
     Load(Data)                          => E_Load_Access_Fault(),
     Load(Vector)                        => E_Load_Access_Fault(),
     Load(PageTableEntry)                => E_Load_Access_Fault(),
-    LoadReserved(Data)                  => E_Load_Access_Fault(),
+    LoadReserved(_, _, Data)            => E_Load_Access_Fault(),
     Store(Data)                         => E_SAMO_Access_Fault(),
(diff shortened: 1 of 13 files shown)


pmundkur:
This should complete the remaining bit of #925 and help to unblock #1734.

arichardson (approved):
It looks like this should allow for some nice future improvements. 

Also while looking at this I was wondering if it would make sense to use an enum for aq/rl instead of the boolean flags (RELAXED/AQ/RL/AQRL)? But not sure if that really makes sense.

pmundkur:
We could change it to enums later.  For now, imho it'll be easier to review the switch from the three boolean arguments to the bools embedded in the access type if they are kept as bools.  When we do switch to enums, it would be good to match the manual which uses `acquire-RCsc`, `release-RCsc`, etc.  But the concurrency model uses `{Read,Write}_RISCV_{conditional_}{strong_}{acquire,release}`, so there would be some renaming/name-mapping somewhere anyway.

Timmmm:
Yeah it would also be good to get some documentation of what the Sail orderings mean exactly @Alasdair @bacam

jordancarlin (approved):
I definitely agree that enums would be better for this long term, but the bools seem fine as a first step to match what we already have.

nadime15 (approved):

Timmmm (approved):

PR #1768, opened 2026-06-18
Fixes to Vector Illegal Instruction Traps

fourth-bit:
This PR adds a few fixes to the illegal instruction detection in the vector instructions:

1. This changes the number of elements that are being considered in the vstart check for scalar move instructions, to align it with the way that arithmetic instructions that permit nonzero vstart are handled. This allows sail to match the behavior that I have observed on other implementations that allow nonzero vstart for scalar moves.
2. This adds mask overlap checks for vs3 in VSSEG type instructions and to vs1 in floating point reductions. Both of these follow from the same rule in the spec: "A vector register cannot be used to provide source operands with more than one EEW for a single instruction. A mask register source is considered to have EEW=1 for this constraint. An encoding that would result in the same vector register being read with two or more different EEWs, including when the vector register appears at different positions within two or more vector register groups, is reserved," because mask registers are read at EEW=1.
3. Also following from that rule, this PR adds an overlap constraint to integer widening reductions because the scalar and vector registers are read at different EEWs. 
4. It expands the segmented indexed load overlap check to take the number of segments into account, so that all possible overlaps are checked.

### DIFF
--- a/model/extensions/V/vext_arith_insts.sail
+++ b/model/extensions/V/vext_arith_insts.sail
@@ -1925,7 +1925,7 @@ function clause execute VMVXS(vs2, rd) = {
   let LMUL_pow = get_lmul_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, 0) |
+  if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, LMUL_pow) |
      illegal_vd_unmasked()
   then return Illegal_Instruction();
 
@@ -2453,7 +2453,7 @@ function clause execute VMVSX(rs1, vd) = {
   let LMUL_pow = get_lmul_pow();
   let num_elem = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, 0) |
+  if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, LMUL_pow) |
      illegal_vd_unmasked()
   then return Illegal_Instruction();
 
--- a/model/extensions/V/vext_arith_insts.sail
+++ b/model/extensions/V/vext_arith_insts.sail
@@ -1922,10 +1922,13 @@ mapping clause encdec = VMVXS(vs2, rd)
 
 function clause execute VMVXS(vs2, rd) = {
   let SEW      = get_sew();
+  let num_elem = get_num_elem(0, SEW);
+  // Scalar move instructions disregard LMUL, but for the purposes of finding an illegal vstart,
+  // some implementations use LMUL for the bounds check, and this helps match those implementations.
   let LMUL_pow = get_lmul_pow();
-  let num_elem = get_num_elem(LMUL_pow, SEW);
+  let num_elem_vstart = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, LMUL_pow) |
+  if illegal_vstart(VSTART_SCALAR_MOVE, num_elem_vstart, LMUL_pow) |
      illegal_vd_unmasked()
   then return Illegal_Instruction();
 
@@ -2450,10 +2453,12 @@ mapping clause encdec = VMVSX(rs1, vd)
 
 function clause execute VMVSX(rs1, vd) = {
   let SEW      = get_sew();
+  let num_elem = get_num_elem(0, SEW);
+  // See above comment in VMVXS for why LMUL is used here
   let LMUL_pow = get_lmul_pow();
-  let num_elem = get_num_elem(LMUL_pow, SEW);
+  let num_elem_vstart = get_num_elem(LMUL_pow, SEW);
 
-  if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, LMUL_pow) |
+  if illegal_vstart(VSTART_SCALAR_MOVE, num_elem_vstart, LMUL_pow) |
      illegal_vd_unmasked()
   then return Illegal_Instruction();
 
@@ -2480,7 +2485,7 @@ function clause execute VMVSX(rs1, vd) = {
     }
   };
 
-  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
+  write_vreg(num_elem, SEW, 0, vd, result);
   set_vstart(zeros());
   RETIRE_SUCCESS
 }
--- a/model/extensions/V/vext_arith_insts.sail
+++ b/model/extensions/V/vext_arith_insts.sail
@@ -1922,7 +1922,8 @@ mapping clause encdec = VMVXS(vs2, rd)
 
 function clause execute VMVXS(vs2, rd) = {
   let SEW      = get_sew();
-  let num_elem = get_num_elem(0, SEW);
+  let LMUL_pow = get_lmul_pow();
+  let num_elem = get_num_elem(LMUL_pow, SEW);
 
   if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, 0) |
      illegal_vd_unmasked()
@@ -2449,7 +2450,8 @@ mapping clause encdec = VMVSX(rs1, vd)
 
 function clause execute VMVSX(rs1, vd) = {
   let SEW      = get_sew();
-  let num_elem = get_num_elem(0, SEW);
+  let LMUL_pow = get_lmul_pow();
+  let num_elem = get_num_elem(LMUL_pow, SEW);
 
   if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, 0) |
      illegal_vd_unmasked()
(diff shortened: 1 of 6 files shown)


model/extensions/V/vext_red_insts.sail
```diff
@@ -32,6 +32,7 @@ function clause execute RIVVTYPE(funct6, vm, vs2, vs1, vd) = {
 
   if illegal_vstart(VSTART_MANDATORY, num_elem_vs, LMUL_pow) |
      illegal_widening_reduction(SEW_widen) |
+     not(valid_src_overlap(vs2, SEW_widen, 0, vs1, SEW, LMUL_pow)) |
```
nadime15: I think the order is reversed?
```suggestion
     not(valid_src_overlap(vs2, SEW, LMUL_pow, vs1, SEW_widen, 0)) |
```

model/extensions/V/vext_fp_red_insts.sail
```diff
@@ -112,6 +113,7 @@ function clause execute RFWVVTYPE(_funct6, vm, vs2, vs1, vd) = {
   if illegal_vstart(VSTART_MANDATORY, num_elem_vs, LMUL_pow) |
      illegal_fp_widening_reduction(SEW, rm_3b, SEW_widen) |
      not(valid_reg_group(vs2, LMUL_pow)) |
+     not(valid_vm_src_overlap(vm, vs1, SEW)) |
```
nadime15: I think you have to add valid_src_overlap like you did in vext_red_insts.sail?
```suggestion
     not(valid_src_overlap(vs2, SEW, LMUL_pow, vs1, SEW_widen, 0))
     not(valid_vm_src_overlap(vm, vs1, SEW)) |
```

model/extensions/V/vext_red_insts.sail
nadime15: To match `vfwredosum.vs` and `vfwredusum.vs add valid_reg_group(...)`
```suggestion
     not(valid_reg_group(vs2, LMUL_pow))
     not(valid_src_overlap(vs2, SEW_widen, 0, vs1, SEW, LMUL_pow)) |
```

model/extensions/V/vext_fp_red_insts.sail
```diff
@@ -112,6 +113,7 @@ function clause execute RFWVVTYPE(_funct6, vm, vs2, vs1, vd) = {
...
      not(valid_vm_src_overlap(vm, vs2, SEW))
```
nadime15:
```suggestion
     not(valid_vm_src_overlap(vm, vs1, SEW_widen)) |
not(valid_vm_src_overlap(vm, vs2, SEW))
```

model/extensions/V/vext_arith_insts.sail
```diff
@@ -1922,7 +1922,8 @@ mapping clause encdec = VMVXS(vs2, rd)
 
 function clause execute VMVXS(vs2, rd) = {
   let SEW      = get_sew();
-  let num_elem = get_num_elem(0, SEW);
+  let LMUL_pow = get_lmul_pow();
+  let num_elem = get_num_elem(LMUL_pow, SEW);
 
   if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, 0) |
```
nadime15: I think if you change `get_num_elem` to pass LMUL_pow instead of 0, you also need to change `illegal_vstart` to pass LMUL_pow instead of 0:

  ```suggestion
if illegal_vstart(VSTART_SCALAR_MOVE, num_elem, LMUL_POW) |
```

`get_num_elem` can't return fractions, so for fractional LMUL it rounds num_elem up to a whole register. illegal_vstart uses that third arg to divide it back down to the real VLMAX, so it has to be the same LMUL_pow you passed to get_num_elem, otherwise VLMAX comes out too big for fractional LMUL (for example, LMUL=1/4 gives 4 instead of 1, see the comment in function `illegal_vstart`).

**Comment by fourth-bit** [comment]
Thanks for the feedback. Those changes should be applied everywhere now.

File: model/extensions/V/vext_arith_insts.sail [hunk 5]
```diff
@@ -1922,9 +1922,10 @@ mapping clause encdec = VMVXS(vs2, rd)
 
 function clause execute VMVXS(vs2, rd) = {
   let SEW      = get_sew();
-  let num_elem = get_num_elem(0, SEW);
+  let LMUL_pow = get_lmul_pow();
+  let num_elem = get_num_elem(LMUL_pow, SEW);
```
pmundkur: You might want to add a comment here about using `LMUL`, since the spec says "The instructions ignore LMUL ...".  Something like "Although these instructions are supposed to ignore `LMUL`, some implementations use `LMUL` for bounds checks for `vstart`, and this helps match those implementations."   
For `VMVXS` you could just say "See `VMVXS` above for the use of LMUL."  
And for `VFMVFS` and `VFMVSF` you could similarly say "see VMVXS for the use of LMUL."

model/extensions/V/vext_arith_insts.sail
```diff
@@ -2478,7 +2480,7 @@ function clause execute VMVSX(rs1, vd) = {
     }
   };
 
-  write_vreg(num_elem, SEW, 0, vd, result);
+  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
```
pmundkur: This seems problematic since the `read_vreg` of `vd` above was done with `LMUL_pow` of 0 but the write here is done with `LMUL_pow`.  Shouldn't this be `0` as before?  

```suggestion
  write_vreg(num_elem, SEW, 0, vd, result);
```

The spec says

> The other elements in the destination vector register ( 0 < index < VLEN/SEW)  ...

which implies the previous 0 for `LMUL_pow` was correct.

model/extensions/V/vext_fp_insts.sail
```diff
@@ -1454,7 +1456,7 @@ function clause execute VFMVSF(rs1, vd) = {
     }
   };
 
-  write_vreg(num_elem, SEW, 0, vd, result);
+  write_vreg(num_elem, SEW, LMUL_pow, vd, result);
```
pmundkur: Again, this change looks wrong; I think `LMUL_pow` should be 0 like before.

model/extensions/V/vext_red_insts.sail
```diff
@@ -32,6 +32,8 @@ function clause execute RIVVTYPE(funct6, vm, vs2, vs1, vd) = {
 
   if illegal_vstart(VSTART_MANDATORY, num_elem_vs, LMUL_pow) |
      illegal_widening_reduction(SEW_widen) |
+     not(valid_reg_group(vs2, LMUL_pow)) |
```
pmundkur: The `valid_reg_group(vs2, LMUL_pow)` check is redundant here and can be removed, since the `valid_src_overlap(vs2, SEW, LMUL_pow, ...)` check on the next line includes a check for valid register groups (see the doc comment for `valid_src_overlap()`).

model/extensions/V/vext_utils_insts.sail
```diff
@@ -142,6 +142,19 @@ function valid_disjoint_reg_groups(rs : vregidx, rd : vregidx, EMUL_pow_rs : int
   rs_int + rs_group <= rd_int | rd_int + rd_group <= rs_int
 }
 
+// Check for valid overlaps in segmented instructions that have an additional number of data fields
+function valid_disjoint_reg_groups_segmented(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_pow_rd : int, data_fields: int) -> bool = {
```
pmundkur: `nfields` is a stronger type than `int` here.
```suggestion
function valid_disjoint_reg_groups_segmented(rs : vregidx, rd : vregidx, EMUL_pow_rs : int, EMUL_pow_rd : int, data_fields: nfields) -> bool = {
```

pmundkur (changes requested):
Thanks for catching all these cases!   I think there are some minor fixes needed.

model/extensions/V/vext_red_insts.sail
nadime15: That was essentially just matching the implementation in vext_fp_red_insts.sail, but we can remove the call to valid_reg_group from vext_fp_red_insts.sail as well.

model/extensions/V/vext_arith_insts.sail
nadime15: This won't work and can trigger an assertion in write_vreg() with the code in its current form (see LMUL = 4) I think we could restore the 0 in write_vreg() as you proposed, but we would also need to adjust the code above.

I was thinking of something along these lines:

```
let num_elem = get_num_elem(0, SEW);
let num_elem_vstart = get_num_elem(LMUL_pow, SEW);
 
 if illegal_vstart(VSTART_SCALAR_MOVE, num_elem_vstart, LMUL_pow) |  // use num_elem_vstart only here and use for the remaining part num_elem
    illegal_vd_unmasked()
```

Would that make sense?

pmundkur: Yeah, that makes more sense.

fourth-bit:
Thank you both for the feedback. I've applied the changes now.

pmundkur (approved):

nadime15 (approved):

jordancarlin (approved):

PR #1772, opened 2026-06-22
Bump Sail compiler release to 0.20.2.

pmundkur:
This updates the download location for the binary releases.

### DIFF
--- a/.github/actions/sail-setup/action.yml
+++ b/.github/actions/sail-setup/action.yml
@@ -8,7 +8,7 @@ inputs:
   sail-version:
     description: 'Sail version to install (use "latest" to build from source)'
     required: false
-    default: "0.20.1"
+    default: "0.20.2"
   install-clang:
     description: "Install latest Clang"
     required: false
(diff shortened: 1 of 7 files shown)


pmundkur:
Still needs the new version to be available from opam.

Timmmm (approved):

.github/actions/sail-setup/action.yml
```diff
@@ -111,7 +111,7 @@ runs:
       shell: bash
       run: |
         sudo mkdir -p /usr/local
-        curl --location https://github.com/rems-project/sail/releases/download/${{ inputs.sail-version }}/sail-$(uname)-$(arch).tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
+        curl --location https://github.com/rems-project/sail/releases/download/${{ inputs.sail-version }}-binary/sail-$(uname)-$(arch).tar.gz | sudo tar xvz --directory=/usr/local --strip-components=1
```
jordancarlin, line 114: @Alasdair Is there a reason the binary release was split out from the regular release again? There was a lot less confusion with the 0.20.1 release when there was a single release.

pmundkur:
@bacam It looks like opam doesn't find the new version of `coq-sail-stdpp`.  Where is that package published anyway?  I don't see even the previous version with `opam list -A`.

bacam:
It's not published yet - I'm just trying to get a new-style `rocq-` package together to accompany it.  Rocq has its own opam repository which the CI picks up from the docker image, rather than requiring explicit configuration.

bacam:
The repository for Rocq packages has merged the new coq-sail version, and the CI works now.

jordancarlin (approved):

PR #1773, opened 2026-06-22
Convert the model callback API to use `shared_ptr`.

pmundkur:
This fixes a TODO and removes the `rvfi_cbs` global. It also addresses some warnings from a slightly strengthened `clang-tidy` config (not checked in yet since too many other warnings need fixing).

### DIFF
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -8,13 +8,13 @@
 #include "riscv_callbacks_if.h"
 #include "symbol_table.h"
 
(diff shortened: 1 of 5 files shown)


Timmmm (approved):

arichardson:
Do we really need a shared_ptr here? Can't we use unique_ptr + std::move?

pmundkur:
> Do we really need a shared_ptr here? Can't we use unique_ptr + std::move?

I do need a `shared_ptr` in #1728, since the `protocol_handler` has callbacks registered with the model and is also referenced from the `connection` object.

PR #1774, opened 2026-06-23
Enforce more fault consistency between `mem_write_ea` and `mem_write_value`.

pmundkur:
There is still a corner case for MMIO, which may need revisiting when the concurrency model is extended to MMIO regions.

This fixes one of the issues mentioned in #1761.

### DIFF
--- a/model/extensions/A/zaamo_insts.sail
+++ b/model/extensions/A/zaamo_insts.sail
@@ -94,7 +94,7 @@ function clause execute AMO(op, aq, rl, rs2, rs1, width, rd) = {
     Ok(addr, pbmt, _) => (addr, pbmt),
   };
 
-  let loaded : bits('width * 8) = match mem_write_ea(addr, width, access, aq & rl, rl, true) {
+  let loaded : bits('width * 8) = match mem_write_ea(addr, width, access, pbmt, aq & rl, rl, true) {
     Err(e) => return memory_exception(vaddr, e),
     Ok(_)  => match mem_read(access, pbmt, addr, width, aq, aq & rl, true) {
       Err(e)     => return memory_exception(vaddr, e),
--- a/model/extensions/Zicboz/zicboz_insts.sail
+++ b/model/extensions/Zicboz/zicboz_insts.sail
@@ -51,7 +51,7 @@ function clause execute ZICBOZ(rs1) = {
         // equal to rs1, but pointer masking can change that.
         Err(e, _) => memory_exception(vaddr - negative_offset, e),
         Ok(paddr, pbmt, _) => {
-          match mem_write_ea(paddr, cache_block_size, access, false, false, false) {
+          match mem_write_ea(paddr, cache_block_size, access, pbmt, false, false, false) {
             Err(e) => memory_exception(vaddr - negative_offset, e),
             Ok(_)  => {
               match mem_write_value(paddr, cache_block_size, zeros(), access, pbmt, false, false, false) {
--- a/model/extensions/cfi/zicfiss_insts.sail
+++ b/model/extensions/cfi/zicfiss_insts.sail
@@ -199,7 +199,7 @@ function clause execute SSAMOSWAP(aq, rl, rs2, rs1, width, rd) = {
  // See https://github.com/riscv/sail-riscv/issues/1761.
 
   let 'width = width;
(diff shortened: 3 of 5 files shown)


Timmmm:
Btw I actually read (ok skimmed) the RVWMO stuff in the ISA manual and it does have an explanation of `mem_write_ea`. Wish I'd found this years ago!

> Notice that writing to memory is split into two steps, Store_ea and Store_memv: the first one makes the memory footprint of the store provisionally known, and the second one adds the value to be stored. We ensure these are paired in the pseudocode (Store_ea followed by Store_memv), but there may be other steps between them.
>
> It is observable that the Store_ea can occur before the value to be stored is determined. For example, for the litmus test [LB+fence.r.rw+data-po](https://github.com/litmus-tests/litmus-tests-riscv/blob/master/tests/non-mixed-size/HAND/LB%2Bfence.r.rw%2Bdata-po.litmus) to be allowed by the operational model (as it is by RVWMO), the first store in Hart 1 has to take the Store_ea step before its value is determined, so that the second store can see it is to a non-overlapping memory footprint, allowing the second store to be committed out of order without violating coherence.

```
{
0:x6=x; 0:x8=y;
1:x6=y; 1:x8=z; 1:x10=x;
}
 P0          | P1           ;
 lw x5,0(x6) | lw x5,0(x6)  ;
 fence r,rw  | xor x7,x5,x5 ;
 ori x7,x0,1 | ori x7,x7,1  ;
 sw x7,0(x8) | sw x7,0(x8)  ;  <-- first store in Hart 1
             | ori x9,x0,1  ;
             | sw x9,0(x10) ;  <--- second store
```

I *think* what they're saying is that this can happen (you need an out-of-order CPU for this of course):

P1: determine store address `0(x8)=z` for `sw x7,0(x8)`
P1: determine store address `0(x10)=x` for `sw x9,0(x10)`
P1: execute `ori x9,x0,1`
P1: address and value are determined for the second store and it doesn't overlap previous stores so we can now commit `sw x9,0(x10)` (`*x = 0`)
P2: execute `lw x5,0(x6)` (`x5 = *x`) - loads the value P1 just wrote
P2: execute all remaining instructions, writes `*y = 0`
P1: Now execute `lw x5,0(x6)` which reads `y`...

Actually maybe I don't get it because that load gets squashed to 0 by the `xor` anyway so how is it observable? Maybe @PeterSewell can give us a minimal example?

PeterSewell:
@Timmmm  y, that's the kind of thing (which is already a minimal example :-) ).   You might want to look at @fshaked 's text on p140ff of https://www.cl.cam.ac.uk/~pes20/multicore-sewell-notes-2026.pdf ("Address-induced control dependency").  

Dependencies through a self-xor (and other arithmetic) are architecturally respected irrespective of the values, even if one might think the h/w could optimise them away entirely - that's quite important for relaxed architecture models to be sane.

pmundkur:
@PeterSewell how do the system-level concurrency models handle memory-mapped IO?  I had to punt on this in this PR, since MMIO regions are typically quite coarsely specified, and one needs to actually do a write to find out whether some fault/exception occurs.  So the `write_ea` could succeed but the `write_value` could still fail.

PeterSewell:
For RV, we've not so far looked at that - maybe Nils will. For Arm, our ISCA2025 paper talks about external aborts - it's an interesting question

jordancarlin (approved):

nadime15 (approved):
This looks good to me. I just wanted to point out that phys_access_check is called twice now in case of a successful store. I think this could lead to a noticeable performance regression (would be interesting to benchmark it via linux boot), since pmp/pma checks are already quite expensive (unless I miss something?). Once spmp is added, it'll probably get even worse.

Caching may help here 😁

pmundkur:
That's a good point.  One way to address this could be to add a config option like `base.enable_concurrency_model` to gate those checks.  So `mem_write_ea` would be a nop when that option is false.  That would also allow us to include `jalr_rmem` and `jalr_seq` in the same clause, to avoid `jalr_rmem` bit-rotting.   If this makes sense, I could do this in a subsequent PR.

jordancarlin:
I was concerned about the performance implications too, but the Linux boot time in CI seems unchanged. It would be good to get #1674 finished so that this is easier to track.

pmundkur:
At the golden model meeting today, we decided to leave the duplicated checks in for now, and not add a config option.  If there are no objections, I'll merge this tomorrow.

PR #1776, opened 2026-06-24
Lean: update the handwritten support to new Sail

ineol:
The namespaces of the generated Lean changed, so the handwritten files need to open the right namespaces.


nadime15 (approved):

pmundkur:
Thanks, it builds locally now!

pmundkur (approved):

PR #1777, opened 2026-06-25
Add an emulator implemented in Lean.

ineol:
This adds:
- a frontend to start an emulator based on Lean extracted from SAIL RISC-V,
- a CI job that builds said emulator, and tests it on some ELF files.


lean_emulator/lake-manifest.json
```diff
@@ -0,0 +1,53 @@
...
+  "version": "1.1.0",
+  "packagesDir": ".lake/packages",
+  "packages": [
+    {
+      "type": "path",
+      "scope": "",
+      "name": "Lean_RV64D_executable",
+      "manifestFile": "lake-manifest.json",
+      "inherited": false,
+      "dir": "../build/model/Lean_RV64D_executable/",
```
pmundkur, line 11: Perhaps in a followup PR, we should fix this hardcoding of the build directory to use the one actually specified in the first `cmake -B <build_dir>` command.  One could use CMake's `configure_file`, a bit like what was done in #1758.

lean_emulator/lakefile.toml
```diff
@@ -0,0 +1,19 @@
...
+root = "Main"
+
+[[require]]
+name = "ELFSage"
+git = "https://github.com/GaloisInc/ELFSage.git"
+rev = "7e68e68e836025df741de04a5bb83439fe88d956"
+
+[[require]]
+name = "Lean_RV64D_executable"
+path = "../build/model/Lean_RV64D_executable/"
```
pmundkur, line 18: Same here about hard-coded build paths.

pmundkur (commented):
Some suggestions for a followup PR, but this LGTM once it passes CI.

pmundkur:
It looks like this needs rebasing to pick up #1776.

ineol:
I just tweaked the emulator imports to fix it and it seems to build on my machine.

pmundkur:
It seems to have passed CI!

One thing I noticed is these warnings during the emulator build, perhaps due to the subsetting of `SAIL_MODULES`:
```
Warning: Incomplete pattern match statement at extensions/Zicboz/zicboz_insts.sail:11.0-8:
11 |function clause currentlyEnabled(Ext_Zicboz) = hartSupports(Ext_Zicboz)
   |^------^
   | 
The following expression is unmatched: Ext_A

Warning: Incomplete pattern match statement at postlude/insts_end.sail:23.0-8:
23 |function clause execute C_ILLEGAL(_s) = Illegal_Instruction()
   |^------^
   | 
The following expression is unmatched: AES32DSI(undefined)
```

This could be fixed in a followup PR.

lean_emulator/README.md
```diff
@@ -0,0 +1,35 @@
+# Lean emulator for Sail RISC-V
+
+## Instructions
+
+1. Compile the Sail RISC-V model for Lean which is a dependency for the Lean emulator. Here's an example command to build the Lean backend (from the root directory). You may want to change the set of modules and extensions that are included in your version of the emulator.
+
+```
+cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;postlude" -DENABLE_LEAN_EMULATOR_TESTS=TRUE && cmake --build build --target generated_lean_executable_rv64d
```
pmundkur: I think this could be simplified to
```suggestion
cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;postlude" -DENABLE_LEAN_EMULATOR_TESTS=TRUE
```
since the `lean_emulator` target depends on `generated_lean_executable_rv64d` so it will get built as a dependency.

lean_emulator/README.md
```diff
@@ -0,0 +1,35 @@
...
+cmake --build build --target build_lean_emulator
+```
+
+3. Run the Lean emulator on an ELF binary.
+
+```
+lake exec lean_riscv_emulator -- path/to/riscv/binary.elf
+```
+
+You can download the Spike test ELFs by running the following
```
pmundkur: This won't actually download the ELFs since that will be done in the first `cmake -S .` step.  You could perhaps say instead
```suggestion
You can test the emulator with the `riscv-test` suite by running the following
```

pmundkur (approved):
Minor fixes to the README but otherwise LGTM.

lean_emulator/Main.lean
```diff
@@ -0,0 +1,17 @@
+import LeanRiscv
+
+def main (args : List String) : IO UInt32 := do
+  if args.length != 1 then do
+    IO.println "usage: run-riscv-lean <elf_file>"
```
pmundkur: Maybe 
```suggestion
    IO.println "usage: lean_riscv_emulator <elf_file>"
```

ineol:
Thank you for the feedback! I'll try to fix the hardcoded build paths in a later PR.

How do you think I can fix the compilation warnings?

pmundkur:
> How do you think I can fix the compilation warnings?

I'm not sure.  I'm guessing this arises from the subset use of SAIL_MODULES.  It looks harmless, so probably not urgent to fix?  Any thoughts @Alasdair @bacam?

Alasdair:
There isn't a super good way around it, other than splitting up the enumeration so each element also lives in the same module which I don't think we want to do.

Maybe the rules for when to emit warnings should be changed, but its a bit of a tradeoff between getting useful warnings in the common case and making sure they never appear in cases like this.

.github/workflows/run-full-tests.yml
```diff
@@ -48,3 +48,23 @@ jobs:
...
+      - name: Check out repository code
+        uses: actions/checkout@v6
+        with:
+          # Git history is needed for `git describe` in the build to work.
+          fetch-depth: 0
+      - name: Common setup
+        uses: ./.github/actions/sail-setup
+      - name: Build Lean emulator
+        run: |
+          cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;Sscofpmf;postlude" -DPRINT_EFFECTS=true -DENABLE_LEAN_EMULATOR_TESTS=on
```
jordancarlin: If you want to change this to build the full Lean emulator, then it might be useful to include it here, but otherwise I don’t see a reason to add the exact same Lean tests that are already in the other workflow. This workflow is intended to run longer tests that we don’t want running on every PR.

.github/workflows/compile-lean.yml
```diff
@@ -51,3 +51,10 @@ jobs:
         uses: leanprover/lean-action@v1
         with:
           lake-package-directory: "build/model/Lean_RV64D/"
+
+      - name: Test the Lean emulator
+        run: |
+          eval $(opam config env)
+          cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;Sscofpmf;postlude" -DPRINT_EFFECTS=true -DENABLE_LEAN_EMULATOR_TESTS=TRUE
```
jordancarlin, line 88: Since this is reconfiguring and rebuilding the model anyway, how about breaking this into a different job (in the same workflow file) so that it can run in parallel with the standard Lean build?

test/CMakeLists.txt
```diff
@@ -197,6 +197,32 @@ if(ENABLE_RISCV_ARCH_TESTS)
...
+    set(DOWNLOAD_URL "${TEST_DOWNLOAD_URL}/${TEST_DOWNLOAD_VERSION}/${TARBALL_NAME}.tar.gz")
+
+    download_riscv_tests(
+        "${DOWNLOAD_PATH}"
+        "${TARBALL_NAME}"
+        "${DOWNLOAD_URL}"
+    )
+
+    include(lean_emulator_test_files.cmake)
+    set(lean_emulator ${PROJECT_SOURCE_DIR}/lean_emulator/.lake/build/bin/lean_riscv_emulator)
```
jordancarlin: Does this mean the final Lean emulator binary is produced in the source tree instead of the build tree? Is there a way to generate it in the build tree instead?

lean_emulator/test_lean_emulator.cmake
jordancarlin, line 1: It seems like almost all of this is now duplicated in the CMake files in the test directory. Should this file be removed?

lean_emulator/README.md
```diff
@@ -0,0 +1,35 @@
...
+```
+cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;postlude" -DENABLE_LEAN_EMULATOR_TESTS=TRUE
+```
+
+2. Compile the Lean emulator
+
+Either, in this directory:
+
+```
+lake update && lake build
```
jordancarlin, line 17: This won’t work since the previous CMake command only configures the build and doesn’t actually produce the Lean files. I’d drop this and just stick with the CMake instructions.

lean_emulator/README.md
```diff
@@ -0,0 +1,35 @@
+# Lean emulator for Sail RISC-V
+
+## Instructions
+
+1. Compile the Sail RISC-V model for Lean which is a dependency for the Lean emulator. Here's an example command to build the Lean backend (from the root directory). You may want to change the set of modules and extensions that are included in your version of the emulator.
```
jordancarlin, line 5: This step only configures the build, it doesn’t actually compile anything.

lean_emulator/lakefile.toml
```diff
@@ -0,0 +1,19 @@
+name = "lean-riscv"
+version = "0.1.0"
```
jordancarlin: Should this use the version number of the sail-riscv model instead? I’d rather not introduce additional version numbers.

lean_emulator/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+add_custom_command(
+  DEPENDS
+      generated_lean_executable_rv64d
```
jordancarlin, line 14: It would be good to generalize this to work with the RV32 version eventually, but not as part of this PR.

jordancarlin (changes requested):
I didn’t review the actual Lean files, but here is feedback on the rest of the PR.

lean_emulator/lakefile.toml
ineol: Is there a way to get/updated it automatically or is it done by hand?

.github/workflows/run-full-tests.yml
ineol: doesn't the next line build the lean emulator? (which is pretty slow to build)

.github/workflows/compile-lean.yml
ineol: Should we remove it from here and only do it in the full-test action?

test/CMakeLists.txt
ineol: from what I can see a build dir can be set in the `lakefile.toml` but it's not clear to me how to set it dynamically with the `lake` build system.

lean_emulator/test_lean_emulator.cmake
ineol: yes, it was an oversight!

lean_emulator/README.md
ineol: I'm pretty sure the line above generates the Lean files
```
cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;postlude" -DENABLE_LEAN_EMULATOR_TESTS=TRUE && cmake --build build --target generated_lean_executable_rv64d
```
At least it does on my machine.

lean_emulator/README.md
ineol: See above, the step has `--build`

.github/workflows/run-full-tests.yml
jordancarlin: Sorry. To clarify, I meant if you want to change this to build the module with all (or an expanded) set of modules, then it might be useful to include, but as is, this is already covered in the standard CI flow, so there is probably no reason to include it here.

.github/workflows/compile-lean.yml
jordancarlin: Assuming it doesn't take too long in CI, I'd rather keep it in the workflow that runs on every PR. Let's try splitting it into two separate jobs in this workflow and see how that goes and then we can decide.

test/CMakeLists.txt
jordancarlin: The best solution might be to have a `lakefile.toml.in` that we run CMake's `configure_file` on to set the build directory dynamically in the `lakefile.toml` file itself.

lean_emulator/README.md
jordancarlin: The instructions in the README currently exclude the `cmake --build` portion:
```
cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DSAIL_MODULES="H;Zca;Zicbom_insts;Zicboz;Zicsr_insts;Zkr;postlude" -DENABLE_LEAN_EMULATOR_TESTS=TRUE
```

lean_emulator/README.md
jordancarlin: The version currently in the README does not have `--build`.

lean_emulator/lakefile.toml
jordancarlin: This can be done using `configure_file` in CMake. See `c_emulator/sail_riscv_version.cpp.in`.

test/CMakeLists.txt
Timmmm: Or I wonder if we could symlink `build` to somewhere in the CMake build directory. Quite unfortunate that you can't set the build directory via an environment variable or something.

Worth a comment if we don't fix it now anyway.

model/postlude/step.sail
```diff
@@ -296,20 +296,21 @@ function loop () : unit -> unit = {
...
       }
     }
-  }
+  };
+
+  let exit_val = unsigned(htif_exit_code);
+  if exit_val == 0
+    then print_endline("SUCCESS")
+    else print_int("FAILURE: ", exit_val);
+  return exit_val
```
Timmmm:
```suggestion
  exit_val
```

pmundkur:
@ineol would you have time to address the feedback from the reviews next week?

ineol:
> Léo Stefanesco (@ineol) would you have time to address the feedback from the reviews next week?

yes! I'll start on Monday (I was very busy this past week)

model/postlude/step.sail
ineol: done

lean_emulator/lakefile.toml
ineol: I just removed the version line, since it does not do anyone any good.

lean_emulator/README.md
ineol: I had another version on my local repo, I fixed it!

test/CMakeLists.txt
ineol: I did the symlink approach: it symlinks the whole `.lake`.

.github/workflows/compile-lean.yml
ineol: I split into two jobs.

ineol:
I think I addressed all the comments. In summary,
- I removed the version for the emulator
- I symlinked its `.lake` directory into the build directory
- Fixed the README
- Removed the spurious `return` in the `step` function.
- Made the emulator build and run in its own job
- Merged with the latest `master` branch (though github says it's out-of-date?)

.github/workflows/compile-lean.yml
```diff
@@ -51,3 +51,40 @@ jobs:
...
+        with:
+          sail-version: "latest"
+
+      - name: Setup Lean
+        uses: leanprover/lean-action@v1
+        with:
+          auto-config: false
+          lake-package-directory: "lean_emulator"
+
+      - name: Test the Lean emulator
```
pmundkur:
```suggestion
      - name: Build and test the Lean emulator
```
Alternatively, perhaps split the build and test into two steps.

lean_emulator/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+add_custom_command(
+  OUTPUT ${CMAKE_BINARY_DIR}/lean-emulator-build
+  COMMENT "Creating symlink for the Lean emulator output."
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_BINARY_DIR}/lean-emulator-build
+  COMMAND ${CMAKE_COMMAND} -E create_symlink
```
pmundkur, line 6: This step failed because I had an existing `.lake` and I had to delete it, but it shouldn't be a problem for a new build I think.

pmundkur (approved):
Minor suggestions, but LGTM.

.github/workflows/compile-lean.yml
ineol: I changed the title as suggested.

lean_emulator/CMakeLists.txt
ineol: Yeah I thought it was not worth the additional cmake complexity.

pmundkur:
@jordancarlin merging this is blocked pending your review.

.github/workflows/compile-lean.yml
jordancarlin: Should we drop `generated_lean_executable_rv64d` from here now that the executable version is tested in the other job?

jordancarlin (commented):
One last minor comment. Otherwise, LGTM.

.github/workflows/compile-lean.yml
pmundkur: @jordancarlin @ineol I made this suggested change.  If this is the last suggested change, we could merge this.

jordancarlin (approved):

.github/workflows/compile-lean.yml
ineol: Thank you!

PR #1778, opened 2026-06-25
Split out splitting functions into `sys/split_access_utils.sail`.

pmundkur:
This moves the address alignment checkers from `sys/mem.sail` and the access splitting functions from `sys/vmem_utils.sail`.  The functions are moved without any changes.

This consolidates the functions used for handling misalignment into one file.  This will help their use to support the MAG PMA, which will involve splitting accesses to physical memory instead of virtual memory.

### DIFF
--- a/model/riscv.sail_project
+++ b/model/riscv.sail_project
@@ -71,6 +71,7 @@ sys {
     sys/platform.sail,
     sys/pma.sail,
     sys/mem_utils.sail,
+    sys/split_access_utils.sail,
     sys/mem.sail,
     sys/vmem_pte.sail,
     sys/vmem_ptw.sail,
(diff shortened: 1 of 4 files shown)


pmundkur:
This is split out from #1734 to reduce the size of that PR.

jordancarlin (approved):
LGTM. Does the reading guide need to be updated at all based on these changes? I haven't looked, but I think I recall a significant portion of it being related to the memory files.

pmundkur:
Yeah it will need to be updated, but I was thinking we could update it once the hypervisor is in, since it's going to need a big update then anyway.

nadime15 (approved):

Timmmm (approved):

PR #1779, opened 2026-06-28
Add a `--dump-memory` CLI option.

mschlaegl:
When set, raw memory dump files are written for each MainMemory region after execution ends (similar to Spike's `mem.0x<base>.bin`).

This accompanies the `--stop-at-pc` CLI option; these two options will help the integration of the tests from the RVVTS test framework.

The [RVVTS Framework](https://github.com/ics-jku/RVVTS) is a modular, open-source framework designed for comprehensive testing of RISC-V Vector (RVV) implementations.  This will allow RVV comparisons between the Spike and Sail-RISCV versions. In the medium term, this will also make the automated use of RVVTS in a Sail-RISCV CI workflow possible.

### DIFF
--- a/c_emulator/riscv_model_impl.h
+++ b/c_emulator/riscv_model_impl.h
@@ -11,6 +11,11 @@
 #include "sail.h"
 #include "sail_riscv_model.h"
 
+struct MemoryRegion {
+  uint64_t base = 0;
+  uint64_t size = 0;
+};
+
 // Model wrapped with an implementation of its platform callbacks.
 class ModelImpl final : private hart::Model {
 public:
@@ -69,6 +74,7 @@ public:
 
   bool config_is_valid();
   bool dtb_within_configured_pma_memory(uint64_t addr, uint64_t size);
+  std::vector<MemoryRegion> main_memory_regions() const;
   std::string generate_dts();
   std::string generate_isa_string();
 
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -15,7 +15,11 @@
 #include "symbol_table.h"
 #include "traploop_detector.h"
 
+#include <algorithm>
+#include <cerrno>
+#include <cstring>
 #include <fcntl.h>
+#include <sstream>
 
 using std::chrono::duration_cast;
 using std::chrono::milliseconds;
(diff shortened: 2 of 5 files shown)


Timmmm:
I probably should have followed through on this code but I also have written some support for dumping memory: https://github.com/riscv/sail-riscv/compare/master...Timmmm:sail-riscv:user/timh/memory

It dumps to an ELF (which you can actually run again). Would that work for you? I'll update it and make a PR anyway...

The `until_pc` stuff seems reasonable anyway - can you separate that out into a separate PR?

mschlaegl:
> I probably should have followed through on this code but I also have written some support for dumping memory: [master...Timmmm:sail-riscv:user/timh/memory](https://github.com/riscv/sail-riscv/compare/master...Timmmm:sail-riscv:user/timh/memory)
> 
> It dumps to an ELF (which you can actually run again). Would that work for you? I'll update it and make a PR anyway...

I'd have to take a closer look at it. If the ELF contained the entire memory image, it would almost certainly work. However, this would require me to extend RVVTS with an ELF reader, which would add overhead that would then affect the test time.
To be honest, at first glance I prefer the RAW dump because it works similarly to spike and could therefore potentially create compatibility with other tools as well.

> 
> The `until_pc` stuff seems reasonable anyway - can you separate that out into a separate PR?

I created a new PR for the `--until_pc` patch -> #1783

Timmmm:
Here's the updated PR: #1549  - I don't see any issue with supporting multiple formats tbh. I'm not too sure about just outputting all of the PMA regions though because quite often that is all 2^64 bytes of memory. We could do it on a page based basis or something (i.e. every page that has been written gets dumped).

The code in #1549 dumps each byte, but it's not efficient in some cases and it's not going to be fun to read if you are reading it manually...

jordancarlin:
We discussed this during the tech-golden-model meeting today, and agree that having an option to dump the raw memory contents would be useful separately from the ELF dump. If possible, it would be nice combine this with the existing logic for dumping the signature region of memory. One possibility is to have a flag that takes a list of labels and dumps the memory between the labels. Maybe if no labels are provided to the flag it can default to dumping all of memory.

mschlaegl:
@jordancarlin 

> We discussed this during the tech-golden-model meeting today, and agree that having an option to dump the raw memory contents would be useful separately from the ELF dump.

Ok, great!


> If possible, it would be nice combine this with the existing logic for dumping the signature region of memory. 

I see! Yes this makes totally sense :-)
~~Maybe a write function that can take arbitrary memory addresses (+len) which is then used by both.~~ -> UPDATE: At first glance, a generalization might seem obvious, but it is actually quite difficult: `write_signature` takes `signature_granularity` into account and reads memory in reverse order (output MSB first) -- which doesn't fit well with the raw memory dump implemented in `write_memory_dump` ;-)

> One possibility is to have a flag that takes a list of labels and dumps the memory between the labels. Maybe if no labels are provided to the flag it can default to dumping all of memory.

Not sure about this. Could you describe your ideas on this in a little more detail? - Thanks!

pmundkur:
Good point about `signature_granularity` and byte ordering being an issue; perhaps we should keep it separate.  Now that `--stop-at-pc` has been split out into #1783, would you prefer to remove that portion from this PR, or create a separate one for `--dump-memory`?

mschlaegl:
@pmundkur I removed the --stop-at-pc patch (handled in #1783) from this PR.

c_emulator/riscv_model_impl.h
```diff
@@ -11,6 +11,11 @@
 #include "sail.h"
 #include "sail_riscv_model.h"
 
+struct MemoryDumpRegion {
```
pmundkur: We could use a more generic name, since this might be useful for non-dumping purposes.
```suggestion
struct MemoryRegion {
```

c_emulator/riscv_sim.cpp
```diff
@@ -117,6 +121,41 @@ void write_signature(const std::string &file, unsigned signature_granularity, co
   fclose(f);
 }
 
+void write_memory_dump(const MemoryDumpRegion &region) {
+  std::ostringstream file_os;
+  file_os << "mem.0x" << std::hex << region.base << ".bin";
```
pmundkur: It would be good to allow the user to specify the filename prefix.

pmundkur (commented):
LGTM, other than some minor suggestions.

c_emulator/riscv_model_impl.h
mschlaegl: Addressed in new version.

c_emulator/riscv_sim.cpp
mschlaegl: Addressed in new version.
`--dump-memory <prefix>` -> output: `<prefix>.0x<base>.bin`

pmundkur (approved):

jordancarlin (approved):
LGTM. We should update the PR title and description before merging.

pmundkur:
I updated the PR title and description.  I hope this is okay @mschlaegl @jordancarlin.  I left some context in because the commit message for #1783 refers to this PR.

PR #1783, opened 2026-06-29
Add --stop-at-pc CLI option

mschlaegl:
Add `--stop-at-pc <addr>`. When specified, execution terminates as soon as the specified program counter is reached (similar to Spike's until pc debug command).

See #1779

### DIFF
--- a/c_emulator/cli_options.h
+++ b/c_emulator/cli_options.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <optional>
 #include <string>
 #include <vector>
 
@@ -26,6 +27,7 @@ struct CLIOptions {
   unsigned rvfi_dii_port = 0;
   std::vector<std::string> elfs;
   uint64_t insn_limit = 0;
+  std::optional<uint64_t> stop_at_pc;
 
   std::string sig_file = {};
   unsigned signature_granularity = DEFAULT_SIGNATURE_GRANULARITY;
--- a/c_emulator/riscv_callbacks_stop_at_pc.h
+++ b/c_emulator/riscv_callbacks_stop_at_pc.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include <cstdint>
+
+#include "riscv_callbacks_if.h"
+#include "sail.h"
+
+class stop_at_pc_callbacks : public callbacks_if {
+public:
+  explicit stop_at_pc_callbacks(uint64_t pc) : m_pc(pc) {
+  }
+
+  bool stop_requested() const {
+    return m_stop_requested;
+  }
+
+  void pc_write_callback(ModelImpl &, sbits new_pc) override {
+    if (new_pc.bits == m_pc) {
+      m_stop_requested = true;
+    }
+  }
+
+private:
+  uint64_t m_pc = 0;
+  bool m_stop_requested = false;
+};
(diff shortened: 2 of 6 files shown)


c_emulator/cli_options.h
```diff
@@ -26,6 +26,8 @@ struct CLIOptions {
   unsigned rvfi_dii_port = 0;
   std::vector<std::string> elfs;
   uint64_t insn_limit = 0;
+  uint64_t until_pc = 0;
+  bool has_until_pc = false;
```
Timmmm: Probably better like this. Also maybe `stop_at_pc` is a better name?

```suggestion
  std::optional<uint64_t> until_pc;
```

pmundkur:
~Instead of adding a new callback, could this use the existing `pc_write_callback`?~  I see that it did, sorry, reading too quickly :p

Timmmm (commented):
One minor nit but otherwise LGTM.

I did create a fancy expression system for the fork we had in Codasip where you could do things like `--stop 'pc == 0x1234 || x10 == 0 && cycle > 10000'`, and also set the exit code based on an expression... but it was quite complicated and had measurable overhead since it was checked every cycle. Also I never pushed the code anywhere.

I'd say this is fine instead.

jordancarlin (approved):
LGTM once @Timmmm's comment is addressed.

c_emulator/riscv_callbacks_stop_at_pc.h
pmundkur, line 1: If the option is renamed to `--stop-at-pc`, perhaps this file could be named accordingly (e.g. `stop_at_pc.h`), and the class correspondingly called `stop_at_pc`.

c_emulator/cli_options.h
mschlaegl: Thank you very much for the feedback. I've uploaded a new version of the patch with the changes you suggested (std::optional).
However, I've kept "until_pc" for now -- the idea was to give it a name similar to the one in Spike (Spike's “until pc” debug command). Please let me know if this should be changed. Thanks!

c_emulator/riscv_callbacks_stop_at_pc.h
mschlaegl: It seems like I was too quick to roll out the new version ;-)
As written above in the answer to @Timmmm's comment: I've kept "until_pc" for now -- the idea was to give it a name similar to the one in Spike (Spike's "until pc" debug command).
However, I have no problem changing the parameter and code from `until_pc` to `stop-at-pc` if there are no other votes for `until_pc.` :-)

mschlaegl:
Since there were two suggestions for `stop-at-pc` (@Timmmm and @pmundkur) and no other votes for `until-pc`, I changed the parameter (and the code) to `stop-at-pc`.
I will also adjust this in #1779

pmundkur (approved):

mschlaegl:
Thanks for the reviews!
All requests should be addressed now, and CI is passing.

Timmmm (approved):

Timmmm:
If you're wondering why I edited the PR message/title it's because our merge flow uses those for the commit message/title. Also we have a system where we add the `will-be-merged` label and then merge it a bit later in case anyone has any last minute objections (unlikely in this case).

Thanks for the contribution!

PR #1784, opened 2026-06-30
Refactor `pmaCheck()` to simplify forthcoming MAG PMA checks.

pmundkur:
The specification is silent about the priority order of PMA and PMP checks; this rearranges the order of the PMA checks to move the misalignment checks after all the other PMA checks, and the PMA checks after the PMP checks.

This seems more in line with the language of the specification which often uses the phrase

> a memory access that would otherwise be able to complete except for the misalignment

The initial checks ensure that the access `would otherwise be able to complete except for the misalignment`, after which the access is checked for misalignment. 

The  "Synchronous exception priority in decreasing priority order." table in the ISA manual also prioritizes access faults above misaligned exceptions (when done after address translation); this change conforms to that priority.

### DIFF
--- a/model/sys/mem.sail
+++ b/model/sys/mem.sail
@@ -174,18 +174,6 @@ private function pmaCheck forall 'n, 0 < 'n <= max_mem_access .
   opt_exception
 }
 
-private function alignmentOrAccessFaultPriority(exc : ExceptionType) -> range(0, 1) = {
-  match exc {
-    E_Fetch_Access_Fault() => 1,
-    E_Load_Access_Fault()  => 1,
-    E_SAMO_Access_Fault()  => 1,
-    E_Fetch_Addr_Align()   => 0,
-    E_Load_Addr_Align()    => 0,
-    E_SAMO_Addr_Align()    => 0,
-    _                      => internal_error(__FILE__, __LINE__, "Invalid exception: " ^ to_str(exc))
-  }
-}
-
 // Check if access is permitted according to PMPs and PMAs.
 function phys_access_check forall 'n, 0 < 'n <= max_mem_access . (
   access : MemoryAccessType(mem_payload),
--- a/model/sys/mem.sail
+++ b/model/sys/mem.sail
@@ -77,8 +77,11 @@ private function pmaCheck forall 'n, 0 < 'n <= max_mem_access .
       override_PMA(attributes, pbmt),
   };
 
-  // The spec is silent about the order in which PMA checks are done.
-  // This implementation checks for misalignment after the other PMAs.
+  // The spec is silent about the order in which PMA checks are done,
+  // but after address translation (i.e. here), access faults have
+  // higher priority than misaligned exceptions so those are checked
+  // first. See the "Synchronous exception priority in decreasing
+  // priority order." table in the ISA manual.
 
   let canAccess : bool = match access {
     InstructionFetch()     => attributes.executable,
@@ -183,9 +186,6 @@ private function alignmentOrAccessFaultPriority(exc : ExceptionType) -> range(0,
   }
 }
 
-private function highestPriorityAlignmentOrAccessFault  (l : ExceptionType, r : ExceptionType) -> ExceptionType =
-  if alignmentOrAccessFaultPriority(l) > alignmentOrAccessFaultPriority(r) then l else r
-
 // Check if access is permitted according to PMPs and PMAs.
 function phys_access_check forall 'n, 0 < 'n <= max_mem_access . (
   access : MemoryAccessType(mem_payload),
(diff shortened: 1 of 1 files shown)


pmundkur:
This will be easier to review if `Hide whitespace` is selected in the view settings under `Files changed`.

allenjbaum:
I can't tell (because of the rather large diff) if you are intending to
select the fault based on an option that is specified in the config file,
or not

But the spec  is not silent about this, and (my reading) is that is a
parameter that is specified in the config file..
It explicitly states that misaligned checks can **optionally** be either
high or lower priority than page fault+access fault. (table 17 in sec 3.1
of priv spec)
It further describes that this option is intended to "flexibly cater to two
design points.".
The wording you point to also reinforces this:

For a memory access that would otherwise be able to complete except for the
misalignment,

an access-fault exception *can *be raised instead of an address-misaligned
exception

if the misaligned access should not be emulated, e.g., if accesses to the
memory region have side effects.

 The keyword "can" is important here, as it often (and in this case) is a
marker for optional, implementation defined behavior.

pmundkur:
This PR is changing the order in which the various PMA checks are done.  The misaligned fault is already chosen from the config file, which describes how to configure the high and low priority cases of misalignment checks; this has been possible since the 0.11 release.

Timmmm:
Hmm yeah I think actually the original code was incorrect because the spec does say (in the exception priority table) that alignment checks are lower priority than access checks (after address translation).

We handled that correctly for PMP/PMA checks (see `highestPriorityAlignmentOrAccessFault`) but not within PMA checks.

So I think this code is correct and I would change the comment to something like

```
// After address translation (i.e. here), access faults have higher priority than
// misaligned exceptions so those are checked first. See the
// "Synchronous exception priority in decreasing priority order." table in the ISA manual.
```

Also, now that I look at it, I think the `phys_access_check` function could be simplified from

```
function phys_access_check forall 'n, 0 < 'n <= max_mem_access . (
  access : MemoryAccessType(mem_payload),
  pbmt : page_based_mem_type,
  priv : Privilege,
  paddr : physaddr,
  width : int('n),
  res_or_con : bool,
) -> option(ExceptionType) = {
  let pmpError : option(ExceptionType) = pmpCheck(paddr, width, access, priv);
  let pmaError : option(ExceptionType) = pmaCheck(paddr, width, access, pbmt, res_or_con);
  match (pmpError, pmaError) {
    (None(), None())     => None(),
    (Some(e), None())    => Some(e),
    (None(), Some(e))    => Some(e),
    (Some(e0), Some(e1)) => Some(highestPriorityAlignmentOrAccessFault(e0, e1)),
  }
}
```

To:

```
function phys_access_check forall 'n, 0 < 'n <= max_mem_access . (
  access : MemoryAccessType(mem_payload),
  pbmt : page_based_mem_type,
  priv : Privilege,
  paddr : physaddr,
  width : int('n),
  res_or_con : bool,
) -> option(ExceptionType) =
  match pmpCheck(paddr, width, access, priv) {
     Some(e) => e,
     None() => pmaCheck(paddr, width, access, pbmt, res_or_con),
  }
```

A PMP check can never raise a misaligned exception, so as long as we do PMP checks before PMA checks, and PMA access checks before PMA misaligned checks, then everything works out.

pmundkur:
Good suggestion, I've simplified `phys_access_check()` as suggested, and also now removed no longer needed helper functions.

Timmmm (approved):
LGTM!

model/sys/mem.sail
```diff
@@ -184,13 +183,16 @@ function phys_access_check forall 'n, 0 < 'n <= max_mem_access . (
...
-    (None(), None())     => None(),
-    (Some(e), None())    => Some(e),
-    (None(), Some(e))    => Some(e),
-    (Some(e0), Some(e1)) => Some(highestPriorityAlignmentOrAccessFault(e0, e1)),
+
+  // Although the relative ordering of PMA and PMP checks is
+  // unspecified, access faults do have a higher priority than
+  // misaligned exceptions (see comment in `pmaCheck()` above).  This
+  // implementation prioritizes PMP checks before PMA checks, as PMP
+  // checks do not raise misaligned exceptions.
```
jordancarlin, line 191: We might need to add a config option for this eventually, but this seems ok for now.

jordancarlin (approved):

model/sys/mem.sail
Timmmm: How so? I believe this is the only allowed behaviour.

jordancarlin: I believe an implementation can technically prioritize misaligned exceptions above access faults even in this scenario. Hopefully no-one actually does that though.

PR #1785, opened 2026-06-30
Enable some `readability` checks in `clang-tidy`.

pmundkur:
Other than some minor cleanup, this interestingly found a mismatch in a parameter name for `tlb_add_callback` and `tlb_flush_callback`, which used `index` in `riscv_platform_if.h` (correctly) and `level` in `riscv_platform_if.cpp` (incorrectly).  This was caught by `readability-inconsistent-declaration-parameter-name`.

`rvfi_handler::gen_entry` which returns a constant was made `static` to satisfy `readability-convert-member-functions-to-static`.

`6.f` was made `6.F` due to `readability-uppercase-literal-suffix`.

One `else` block was removed to satisfy `readability-else-after-return`.

One `auto` was changed to `auto const *` to satisfy `readability-qualified-auto`.

Other fixes were due to `readability-implicit-bool-conversion`.

`readability-identifier-length`, `readability-magic-numbers`,  `readability-isolate-declaration` and `readability-named-parameter` were silenced since they each raise too many warnings.

`readability-avoid-nested-conditional-operator` was disabled since it complains over nested use of `?` e.g

```
const char *annotation = is_entry_selected ? (is_flush ? "  <- flushed" : "  <- added") : "";
```

Satisfying `readability-function-cognitive-complexity` will require refactoring `riscv_sim::run_sail()`.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -7,7 +7,14 @@ Checks: >
   misc-*,
   -misc-include-cleaner,
   -misc-const-correctness,
-  -misc-no-recursion
+  -misc-no-recursion,
+  readability-*,
+  -readability-identifier-length,
+  -readability-isolate-declaration,
+  -readability-named-parameter,
+  -readability-avoid-nested-conditional-operator,
+  -readability-magic-numbers,
+  -readability-function-cognitive-complexity
 
 HeaderFilterRegex: ""
 ...
--- a/c_emulator/cli_options.cpp
+++ b/c_emulator/cli_options.cpp
@@ -148,7 +148,7 @@ CLIOptions parse_cli(int argc, char **argv) {
   );
 
   std::size_t column_width = 45;
-  app.get_formatter()->long_option_alignment_ratio(6.f / static_cast<float>(column_width));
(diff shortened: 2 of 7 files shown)


.clang-tidy
```diff
@@ -7,7 +7,14 @@ Checks: >
...
   -misc-include-cleaner,
   -misc-const-correctness,
-  -misc-no-recursion
+  -misc-no-recursion,
+  readability-*,
+  -readability-identifier-length,
+  -readability-isolate-declaration,
+  -readability-named-parameter,
+  -readability-avoid-nested-conditional-operator,
+  -readability-magic-numbers,
```
arichardson, line 16: The magic number one seems like it might be a good idea, is it too noisy?

arichardson (approved):
Looks great just wondering if there are more we should add.

.clang-tidy
pmundkur: Yeah, it's too noisy, and will need to be enabled in a separate PR.

Timmmm (approved):

jordancarlin (approved):

PR #1791, opened 2026-07-02
Make -DFIRST_PARTY_TESTS=ON work by default on macOS

arichardson:
On macOS the clang binary in $PATH (/usr/bin/clang) does not support
RISC-V ELF targets, so we default to the homebrew installation instead.
If it's not found via homebrew users still need to pass a path to
CLANG_BIN manually. Additionally, the check for RISC-V support was using
the default clang in $PATH instead of $CLANG_BIN, so it failed on macOS.

### DIFF
--- a/test/first_party/CMakeLists.txt
+++ b/test/first_party/CMakeLists.txt
@@ -6,10 +6,22 @@
 # 1. Clang, which is sensibly designed to be able to cross-compile to any architecture it supports.
 #    Unfortunately it is possible to build it without RISC-V support, e.g. RHEL 8 does this, so we
 #    need to check that. You can download a binary from https://github.com/llvm/llvm-project/releases
+#    On macOS, /usr/bin/clang is in $PATH by default but does not support the RISC-V target, so we default to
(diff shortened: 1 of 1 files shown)


test/first_party/CMakeLists.txt
```diff
@@ -31,7 +43,7 @@ set(CROSS_COMPILER_COMMAND_RV64)
 if (CLANG_BIN)
     # Check it supports RISC-V.
     execute_process(
-        COMMAND clang -print-targets
+        COMMAND ${CLANG_BIN} -print-targets
```
arichardson, line 46: This is the actual fix, CLANG_BIN was being ignored. I could split this out as a separate commit.

pmundkur (approved):
Looks good to me.   I wonder if you have looked into getting clang-tidy working (-DENABLE_CLANG_TIDY=ON)?

jordancarlin (approved):

Timmmm (approved):

arichardson:
Okay to hit merge or should I wait?

PR #1796, opened 2026-07-07
Restrict the ISA version configuration to the privileged ISA.

pmundkur:
Also use the 1.13, 1.12, etc. nomenclature instead of the dates, since that seems to be what users are reporting.

For now, a single privileged version is used for both Machine and Supervisor mode, since this suffices for the specifications to date.  This might have to change if future specifications diverge from this practise.

### DIFF
--- a/model/core/isa_version.sail
+++ b/model/core/isa_version.sail
@@ -6,46 +6,44 @@
 /*  SPDX-License-Identifier: BSD-2-Clause                                                */
 /*=======================================================================================*/
 
-// Privileged and unprivileged spec versions. The versioning system for specifications
-// is unclear so we specify the riscv-isa-manual Git commit hashes in comments. Any commit you need
-// to refer to should be added to this enum IN ORDER FROM OLDEST TO NEWEST.
-enum IsaVersion = {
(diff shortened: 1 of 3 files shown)


nadime15 (approved):

model/core/isa_version.sail
```diff
@@ -6,46 +6,44 @@
...
+// "Supervisor ISA", and just the "Machine" and "Supervisor" privilege
+// levels in the latest version of the spec.  It is conceivable that
+// in the future, the Machine and Supervisor portions could get
+// different versions.  But for now, there is a single version for
+// both for simplicity.
+
+enum Privileged_ISA_Version = {
+  Privileged_ISA_1_11,
+  Privileged_ISA_1_12,
+  Privileged_ISA_1_13,
```
Timmmm, line 26: Can we figure out which commits these are? If not I guess we'll just have to live with it. We can make sure we do it for future releases at least.

Timmmm (approved):

model/core/isa_version.sail
pmundkur: Added.

PR #1797, opened 2026-07-07
Ensure that `SEW` cannot be configured to exceed `ELEN`.

### LINKED ISSUE #1794: Sail not setting vill when executing illegal vector configuration-setting instruction
Hello, I have been using sail to test my own implementation of RISC-V Vector core. My core has ELEN = 32 and VLEN = 128 and aims to support Zve32x extension, so I have (or atleast I think I have) configured sail with the same settings:
```
"V": {
 // `support_level` must be one of the below strings:
      // . "Disabled" for no vector extension support
      // . "Integer" for Zve*x support
      // . "Float_single" for Zve*f support
      // . "Float_double" for Zve*d support
      // . "Full" for the full V extension
      // Note that `support_level` requires compatible
      // values for `vlen_exp` and `elen_exp` below.
      // For e.g. Zve32x and Zve32f require `elen_exp` >= 5,
      // Zve64x, Zve64f, and Zve64d require `elen_exp` >= 6,
      // and V requires `elen_exp` >= 6 and `vlen_exp` >= 7.
      "support_level": "Integer",
      // `vlen_exp` and `elen_exp` should be between 3 and 16,
      // inclusive.  If `support_level` is `"Disabled"`, then these
      // should be left as 3 (or any other legal value).
      "vlen_exp": 7,
      "elen_exp": 5,
      "reserved_behavior": {
        // This determines how to handle attempts by vsetvli, vsetivli,
        // and vsetvl to set an unsupported or reserved vtype.
        // "IllegalVtype_SetVill" - set vtype.vill, set vl to zero, clear vstart, and write rd.
        // "IllegalVtype_Illegal" - treat it as an illegal instruction.
        // "IllegalVtype_Fatal"   - raise a Sail exception, stopping execution.
        "illegal_vtype": "IllegalVtype_SetVill",
        // This determines how to handle an out-of-bounds vstart (vstart >= VLMAX), which the
        // vector spec marks reserved and recommends, but does not require, trapping.
        // "Vstart_Illegal" - raise an illegal instruction exception.
        // "Vstart_Ignore"  - treat as a no-op, the instruction writes nothing.
        "vstart_out_of_bounds": "Vstart_Illegal"
      },
      "vl_use_ceil": false,
      // The maximum index EEW for indexed vector addressing mode, as
      // a power of 2.  This must be at least 3 but must not exceed
      // the supported ELEN.
      "max_index_eew_exp": 5,
      "vstart": {
        // The vector specification permits implementations to raise an
        // illegal instruction exception when `vstart` is non-zero for
        // instructions that cannot produce such a `vstart` value through
        // normal execution.
        "zero_required": {
          // Whether vector arithmetic instructions raise an illegal instruction
          // exception when `vstart` is non-zero. This also governs instructions
          // from the Zvbb, Zvbc, Zvabd, Zvfbfmin, and Zvfbfwma extensions.
          "arith": true,
          // Whether the vector scalar move instructions (vmv.x.s, vmv.s.x,
          // vfmv.f.s and vfmv.s.f) raise an illegal instruction exception when
          // `vstart` is non-zero.
          "scalar_move": true
        }
      }
    },
```
I have my own assembly tests that I use to test different instructions, one of them being vset{i}vl{i} instruction. In this test I issue the following instruction: vsetvli x0, a0, e64, m2, ta, ma, which, as of my understanding, should result into an illegal configuration attempt since maximum element width supported is 32bit. Instead, sail will execute this instruction as if it is supported and it will set vtype to 0xd9, instead of the expected 0x80000000. I am using latest sail version (0.12).
Probably I am missing some configuration detail or misunderstanding something but I couldn't find the fix. Thank you in advance.

pmundkur:
This can happen on systems with ELEN=32; add a check to prevent this.

Fixes #1794.


model/extensions/V/vext_vset_insts.sail
```diff
@@ -112,11 +112,13 @@ private function execute_vsetvl_type(
...
-  // "For a given supported fractional LMUL setting, implementations must
-  // support SEW settings between SEWMIN and LMUL * ELEN, inclusive."
-  //
-  // Also some cases require that VLMAX doesn't change (i.e. LMUL/SEW is constant).
-  if SEW_pow_new > (LMUL_pow_new + elen_exp) |
+  // Check the following:
+  // . SEW cannot exceed ELEN.
+  // . "For a given supported fractional LMUL setting, implementations must
+  //   support SEW settings between SEWMIN and LMUL * ELEN, inclusive."
+  // . Also some cases require that VLMAX doesn't change (i.e. LMUL/SEW is constant).
```
nadime15:
```suggestion
  // Check the following:
  // - SEW cannot exceed ELEN.
  // - "For a given supported fractional LMUL setting, implementations must
  //   support SEW settings between SEWMIN and LMUL * ELEN, inclusive."
  // - Also some cases require that VLMAX doesn't change (i.e. LMUL/SEW is constant).
```

nadime15 (approved):

Timmmm (approved):

PR #1799, opened 2026-07-08
Test for configuration to raise access faults on misaligned access

Timmmm:
This adds a first party test to check that it is possible to configure a memory region to raise an access fault on misaligned access. I added an extra memory region to the default config for this but it could also go in a test specific override (maybe better).

Also this changes the default `allowed_within_exp` to 12 which is way less surprising - otherwise *all* misaligned accesses are split into 1-byte accesses so you don't get any exception at all in this test.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -223,7 +223,7 @@
       // 2^allowed_within_exp sized regions will be split into multiple
       // memory operations. 0 means all misaligned accesses will be split.
       // The maximum value is 12 (one page).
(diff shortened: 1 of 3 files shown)


pmundkur (commented):
Yeah, I'd move the region into a test-specific override, since the default config is also used for OS boot.

The global `allowed_within_exp` change will be overridden by a specified MAG by #1734 (in that PR, the global `allowed_within_exp` only kicks in if no MAG is specified for a region).

PR #1801, opened 2026-07-09
Fix some clang-tidy issues caused by recent commits.

pmundkur:
Fixes a couple of `readability-qualified-auto`, `readability-implicit-bool-conversion` and `bugprone-implicit-widening-of-multiplication-result` warnings.

### DIFF
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -417,7 +417,7 @@ bool ModelImpl::dtb_within_configured_pma_memory(uint64_t addr, uint64_t size) {
 
 std::vector<MemoryRegion> ModelImpl::main_memory_regions() const {
   std::vector<MemoryRegion> regions;
-  for (auto region = zpma_regions; region != nullptr; region = region->tl) {
+  for (auto *region = zpma_regions; region != nullptr; region = region->tl) {
     if (region->hd.zattributes.zmem_type == hart::zMainMemory) {
       regions.push_back({region->hd.zbase, region->hd.zsizze});
     }
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -417,7 +417,7 @@ bool ModelImpl::dtb_within_configured_pma_memory(uint64_t addr, uint64_t size) {
 
 std::vector<MemoryRegion> ModelImpl::main_memory_regions() const {
   std::vector<MemoryRegion> regions;
-  for (auto *region = zpma_regions; region != nullptr; region = region->tl) {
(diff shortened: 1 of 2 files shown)


pmundkur:
I'm not sure why these were not caught in CI even though `WARNINGS_AS_ERRORS` is set.

c_emulator/riscv_model_impl.cpp
```diff
@@ -417,7 +417,7 @@ bool ModelImpl::dtb_within_configured_pma_memory(uint64_t addr, uint64_t size) {
 
 std::vector<MemoryRegion> ModelImpl::main_memory_regions() const {
   std::vector<MemoryRegion> regions;
-  for (auto region = zpma_regions; region != nullptr; region = region->tl) {
+  for (auto *region = zpma_regions; region != nullptr; region = region->tl) {
```
Timmmm: Probably should be `const` too?

```suggestion
  for (const auto *region = zpma_regions; region != nullptr; region = region->tl) {
```

c_emulator/riscv_sim.cpp
```diff
@@ -128,12 +128,12 @@ void write_memory_dump(const MemoryRegion &region, const std::string &prefix) {
...
 
   FILE *f = fopen(file.c_str(), "wb");
-  if (!f) {
+  if (f == nullptr) {
     fprintf(stderr, "Cannot create memory dump '%s': %s\n", file.c_str(), strerror(errno));
     return;
   }
 
-  constexpr size_t buffer_size = 64 * 1024;
+  constexpr size_t buffer_size = static_cast<size_t>(64 * 1024);
```
Timmmm, line 139: Maybe add a comment:
```suggestion
  // TODO: In C++23 we can use a literal suffix: 64 * 1024zu;
  constexpr size_t buffer_size = static_cast<size_t>(64 * 1024);
```

Timmmm (approved):

c_emulator/riscv_sim.cpp
```diff
@@ -128,12 +128,15 @@ void write_memory_dump(const MemoryRegion &region, const std::string &prefix) {
...
+  if (f == nullptr) {
     fprintf(stderr, "Cannot create memory dump '%s': %s\n", file.c_str(), strerror(errno));
     return;
   }
 
-  constexpr size_t buffer_size = 64 * 1024;
+  // TODO: In C++23 we can use a literal suffix: 64 * 1024ZU
+  // (upper-case due to the clang-tidy
+  // `readability-uppercase-literal-suffix` check).
+  constexpr size_t buffer_size = static_cast<size_t>(64 * 1024);
```
arichardson, line 139: This warning seems a bit annoying but if there is only one place we need to fix this then lgtm

arichardson (approved):

PR #1802, opened 2026-07-09
Add configuration option to control when mstatus[FD] is dirtied

### LINKED ISSUE #1798: Sail needs a configurable mstatus.FS dirty-tracking policy, or ACT must avoid precise FS/SD expectations when the DUT profile says HW_MSTATUS_FS_DIRTY_UPDATE=imprecise.
[smf00_fs_dirty_imprecise_evidence.zip](https://github.com/user-attachments/files/29796646/smf00_fs_dirty_imprecise_evidence.zip)

SmF-00 Test fails on VF2/U74 because the DUT uses spec-legal imprecise mstatus.FS dirty tracking, but ACT/Sail reference signature still expects precise FS behavior.

Failure point:
fcvt.w.s x11, f0

Generated test sequence sets:
mstatus.FS = 1 / Initial
Then executes fcvt.w.s x11,f0, then reads mstatus.

VF2/U74 result:

Actual mstatus:   0x8000000a00006080
Expected mstatus: 0x0000000a00002080

Interpretation:
Expected: FS=Initial, SD=0
Actual:   FS=Dirty,   SD=1

Logs snippet!

========================================================================================

RVCP-SUMMARY: TEST FAILED - Test File "SmF-00.S"
RVCP: DEBUG INFORMATION FOLLOWS
RVCP: Test Info: "test: 90; cg: SmF_fcsr_cg; cp: cp_mstatus_FS_transition_fs1; bin: fcvt.w.s x11, f0"
RVCP: Instruction: 0xc00075d3
RVCP: Approximate address (failure may be slightly after this): 0x00000000800023d4
RVCP: Register: x11
RVCP: Bad Value:      0x8000000a00006080
RVCP: Expected Value: 0x0000000a00002080
RVCP: MEPC:    0x000000008000215c
RVCP: MCAUSE:  0x0000000000000002
RVCP: MTVAL:   0x000000004010f053
RVCP: MSTATUS: 0x8000000a00006080
RVCP: END OF DEBUG INFORMATION

=============================================================================

I changed the U74 profile to: **HW_MSTATUS_FS_DIRTY_UPDATE: imprecise**  and regenerated/reran SmF-00, but the generated Sail/ACT expected signature did not change.

The DUT behavior is spec-legal imprecise FS dirty tracking. ACT has a profile knob for this (HW_MSTATUS_FS_DIRTY_UPDATE: imprecise), but SmF-00 still generates a Sail/reference signature assuming precise FS tracking. Either the SmF generator should account for the imprecise profile setting, or Sail needs a configurable FS dirty-update mode so the reference signature can model DUT behavior.

Hardware behavior is explicitly allowed at any privileged-spec version, both old and current. 

▎ "Implementations may choose to track the dirtiness of the floating-point register state imprecisely by reporting the state to be dirty even when it has not been modified."

And it goes further — some implementations skip Clean entirely:

▎ "On other implementations, dirtiness might not be tracked at all, in which case the valid FS states are Off and Dirty, and an attempt to set FS to Initial or Clean causes it to be set to Dirty."

So U74 marking FS=Dirty after fcvt.w.s (which only reads f0 and writes an integer register) is spec-compliant conservative behavior, not a bug and not an artifact of an older revision. Real cores commonly do this because precise per-instruction dirty tracking (only-if-an-f-register-or-fcsr-actually-changed) is expensive to implement; it's simpler to just mark the FPU state dirty on any instruction that touches the FPU pipeline at all.

Timmmm:
Currently the model sets the FD (Float Dirty) bit only when absolutely required, but real implementations have a couple of other realistic options. This adds a configuration option that allows selecting them.

Note that the set of legal behaviours for dirty bits is in general unbounded. It would technically be legal to hard-code it to 1. But these are the most likely behaviours in actual hardware because they remove dependencies of setting FD, which makes things faster/easier on pipelined/OoO designs:

* Fflags_Dirty_Flag removes dependency on the value of `fflags`.
* Fflags_Dirty_Instruction removes dependency on execution of the instruction.

Fixes #1798

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -426,7 +426,17 @@
       "supported": true
     },
     "F": {
-      "supported": true
+      "supported": true,
+      // When to set mstatus[FS] and mstatus[SD] to 1, as a result of fflags
+      // being set. There are three options:
+      //
+      // - Fflags_Dirty_Precise: Only set FS/SD when absolutely required, because
+      //                         fflags has actually changed value.
+      // - Fflags_Dirty_Flag: Set FS/SD when the instruction sets a flag, even
+      //                      if that flag is already set.
+      // - Fflags_Dirty_Instruction: Set FS/SD when we execute an instruction that
+      //                             might set a flag, even if it actually doesn't.
+      "fflags_dirty_policy": "Fflags_Dirty_Precise"
     },
     "D": {
       "supported": true
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -5,6 +5,8 @@
     specified; see `base.isa_version`.
   - Delegatable subsets of `medeleg` and `mideleg` can be specified;
     see `base.medeleg.delegatable_bits` and `base.mideleg.delegatable_bits`.
+  - Add the `extensions.F.fflags_dirty_policy` option to configure exactly
+    when mstatus[FS/SD] is set by floating point instructions.
 
 - Important issues addressed and bugs fixed:
   - https://github.com/riscv/sail-riscv/issues/1794 : `vtype.vill` was not set when SEW was configured to exceed ELEN
--- a/model/core/platform_config.sail
+++ b/model/core/platform_config.sail
@@ -190,3 +190,14 @@ let xtvec_mode_reserved_behavior : XtvecModeReservedBehavior = config base.reser
 let rv32zdinx_odd_register_reserved_behavior : RV32ZdinxOddRegisterReservedBehavior = config base.reserved_behavior.rv32zdinx_odd_register
 let illegal_vtype_reserved_behavior : IllegalVtypeReservedBehavior = config extensions.V.reserved_behavior.illegal_vtype
 let vstart_reserved_behavior : OOBVstartReservedBehavior = config extensions.V.reserved_behavior.vstart_out_of_bounds
+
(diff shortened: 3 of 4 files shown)


pmundkur (commented):
There are some instructions that write F/D registers that don't write `fflags`, like `FMV*`s to floating point register destinations, and `FSGN*`.   There are also vector instructions like `vfmv.f.s` that would need updating.  I didn't check comprehensively, so there might be others.

This might also mean that the name of the config option should change to not mention `fflags`?

Timmmm:
This setting only affects fflags accumulation though. When FPR `fN` registers are modified, or when `fcsr/fflags` is written directly with `csrw`, the model always sets the dirty bit.

You are technically allowed to make those precise too:

> If an instruction explicitly or implicitly writes a floating-point register or the fcsr but does not alter its contents, and FS=Initial or FS=Clean, it is implementation-defined whether FS transitions to Dirty.

However I think it's extremely unlikely that any implementation would actually do that, so I don't think it makes sense to preemptively add a setting.

pmundkur:
Sounds good, we can add a separate setting for that if needed.

pmundkur (approved):
There is a typo in the flag names in the PR text, but otherwise this LGTM.

nadime15 (approved):

PR #1804, opened 2026-07-13
Add some `cppcoreguidelines` checks for `clang-tidy`.

pmundkur:
This adds another set of general clang-tidy checks.  Useful ones among those disabled could be enabled one-by-one with appropriate fixes in subsequent PRs.

It fixes the `init-variables` and `avoid-c-arrays` checks.

`avoid-magic-numbers` causes too many warnings currently.
`avoid-do-while` requires refactoring the small `do-while` loop in `sail_riscv_sim.cpp::run_model()`; probably not worth it.
`owning-memory` is disabled since it suggests the use of `gsl` (guidelines-support-library).
`bounds-avoid-unchecked-container-access` could be enabled in a later PR.
`bounds-array-to-pointer-decay` is triggered by accesses to `gmplib`'s `mpz_t` integers in `lbits` when using `gmp_fprintf`.
`bounds-pointer-arithmetic` is triggered when accessing an element in the Sail vector of TLB entries through the vector's `data` pointer.
`type-cstyle-cast` is disabled due to `struct sockaddr` usage for the socket in RVFI.
`type-reinterpret-cast` is disabled due to its use when writing bytes to memory from the ELF loader.
`type-union-access` is disabled since `option(TLB_Entry)` in the TLB callbacks is a union.
`type-vararg` is disabled due to the use of `fprintf`.

### DIFF
--- a/c_emulator/riscv_callbacks_log.h
+++ b/c_emulator/riscv_callbacks_log.h
@@ -51,5 +51,5 @@ private:
   bool config_print_ptw;
   bool config_print_tlb;
   FILE *trace_log;
-  std::vector<uint64_t> pending_flush_indices{};
+  std::vector<uint64_t> pending_flush_indices;
 };
--- a/c_emulator/riscv_callbacks_log.h
+++ b/c_emulator/riscv_callbacks_log.h
@@ -51,4 +51,5 @@ private:
   bool config_print_ptw;
   bool config_print_tlb;
   FILE *trace_log;
+  std::vector<uint64_t> pending_flush_indices{};
 };
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -448,7 +448,7 @@ std::optional<std::string> ModelImpl::string_of_current_exception() {
   if (!have_exception) {
     return std::nullopt;
   }
-  sail_string str{};
+  sail_string str = nullptr;
   CREATE(sail_string)(&str);
   zstring_of_exception(&str, *current_exception);
   std::string exception_str(str);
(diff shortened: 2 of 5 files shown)


jrtc27:
I mean, avoid-do-while is a stupid rule; when that's the exact loop behaviour you want you should use the right tool for the job, not shoehorn it into another type...

Timmmm (approved):

Timmmm:
I think it's just seen as error-prone, which is fair but also I'm not sure I agree with that anyway (I can't recall ever writing any bugs due to getting confused about do-while).

> macro-usage requires removing the SOFTFLOAT_PRELUDE macro in riscv_softfloat.cpp.

There's no reason for it to be a macro - I made a PR to change it to a function: #1806

pmundkur:
Rebased and enabled the macro check after #1806.

jordancarlin (approved):

c_emulator/riscv_callbacks_log.h
```diff
@@ -51,4 +51,5 @@ class log_callbacks : public callbacks_if {
   bool config_print_ptw;
   bool config_print_tlb;
   FILE *trace_log;
+  std::vector<uint64_t> pending_flush_indices{};
```
arichardson:
```suggestion
  std::vector<uint64_t> pending_flush_indices;
```

The init seems redundant since it will be default constructed anyway?

c_emulator/riscv_model_impl.cpp
```diff
@@ -448,7 +448,7 @@ std::optional<std::string> ModelImpl::string_of_current_exception() {
   if (!have_exception) {
     return std::nullopt;
   }
-  sail_string str;
+  sail_string str{};
```
arichardson: This is just a const char*, right? Might be easier to read with `= nullptr`, but no strong preference.

arichardson (approved):
One minor nit otherwise LGTM.

arichardson (approved):

PR #1808, opened 2026-07-14
Improve handling of the delegatable bits configuration for `medeleg` and `mideleg`.

### LINKED ISSUE #1803: medeleg should have some read-only 0 bits
Per priv spec 3.1.8, medeleg[16] (double-trap) should be read-only zero.  

The other reserved and custom fields probably should be ROZ too, but that is a Sail implementation decision.

The attached file runs in Spike and writes all 1s to medeleg.  Spike reads back 000cb3fe because double trap, ecall from M-mode, and misaligned instructions are not delegable and reserved and custom fields are ROZ.  The expected value is generated by Sail during compilation and shows fffff7ff because only ecall from M-mode is ROZ.

```
This log file generated by running:
spike --isa=rv64imafdcbv_zicbom_zicboz_zicbop_zicfilp_zicfiss_zicond_zicsr_zicntr_zicclsm_ziccif_zifencei_zihintntl_zihintpause_zihpm_zimop_zabha_zacas_zawrs_zfa_zfbfmin_zfh_zcb_zcmop_zbc_zkn_zks_zkr_zvfbfmin_zvfbfwma_zvfh_zvbb_zvbc_zvkg_zvkned_zvknha_zvknhb_zvksed_zvksh_zvkt_sscofpmf_smcntrpmf_sstc_svinval_svade_svadu_svnapot_svpbmt_smstateen_ssnpm_smmpm_smnpm /home/harris/cvw/addins/riscv-arch-test-dh/work/spike-rv64-max/elfs/priv/Sm/Sm_mcsr-00.elf


RVCP-SUMMARY: TEST FAILED - Test File "Sm_mcsr-00.S"
RVCP: DEBUG INFORMATION FOLLOWS
RVCP: Test Info: "test: 248; cg: Sm_mcsr_cg; cp: cp_mcsr_access; bin: medeleg_csrrw1"
RVCP: Instruction: 0x30241073
RVCP: Approximate address (failure may be slightly after this): 0x0000000080000150
RVCP: Register: x15
RVCP: Bad Value:      0x00000000000cb3fe
RVCP: Expected Value: 0xfffffffffffff7ff
RVCP: END OF DEBUG INFORMATION
```

[Sm_mcsr-00.elf.gz](https://github.com/user-attachments/files/29865548/Sm_mcsr-00.elf.gz)

pmundkur:
Ensure that bits for reserved exceptions in `medeleg` are read-only zero, and that the double-trap exception cannot be delegated.  Bits for custom exceptions are allowed to be set but ignored.

Validate the delegated bits against reserved bits. For `mideleg`, bits for interrupts designated for platform use are allowed to be set but ignored.

`legalize_mideleg` now allows {`MEI`, `MTI`, `MSI`} to be set in `mideleg` to account for this clarification:
https://github.com/riscv/riscv-isa-manual/issues/3162

This fixes #1803.

### DIFF
--- a/model/core/interrupt_regs.sail
+++ b/model/core/interrupt_regs.sail
@@ -68,6 +68,9 @@ private function legalize_mideleg(o : Minterrupts, v : xlenbits) -> Minterrupts 
 }
 
 bitfield Medeleg : bits(64) = {
+  HardwareError     : 19,
+  SoftwareCheck     : 18,
+  DoubleTrap        : 16,
   SAMO_Page_Fault   : 15,
   Load_Page_Fault   : 13,
   Fetch_Page_Fault  : 12,
@@ -84,9 +87,29 @@ bitfield Medeleg : bits(64) = {
   Fetch_Addr_Align  : 0
 }
 
-private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
-  // M-EnvCalls delegation is not supported
-  [Mk_Medeleg(v & plat_medeleg_delegatable_bits) with MEnvCall = 0b0]
+function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
+  let v = Mk_Medeleg(v & plat_medeleg_delegatable_bits);
+  // M-EnvCalls and DoubleTrap cannot be delegated, and reserved and
+  // custom fields should be read-only zero.
+  [Mk_Medeleg(zeros()) with
+    HardwareError      = v[HardwareError],
+    SoftwareCheck      = v[SoftwareCheck],
+    DoubleTrap         = 0b0,
+    SAMO_Page_Fault    = v[SAMO_Page_Fault],
+    Load_Page_Fault    = v[Load_Page_Fault],
+    Fetch_Page_Fault   = v[Fetch_Page_Fault],
+    MEnvCall           = 0b0,
+    SEnvCall           = v[SEnvCall],
+    UEnvCall           = v[UEnvCall],
+    SAMO_Access_Fault  = v[SAMO_Access_Fault],
+    SAMO_Addr_Align    = v[SAMO_Addr_Align],
+    Load_Access_Fault  = v[Load_Access_Fault],
+    Load_Addr_Align    = v[Load_Addr_Align],
+    Breakpoint         = v[Breakpoint],
+    Illegal_Instr      = v[Illegal_Instr],
+    Fetch_Access_Fault = v[Fetch_Access_Fault],
+    Fetch_Addr_Align   = v[Fetch_Addr_Align],
+  ]
 }
 
 // Note that mip's MEIP and SEIP had special handling. MEIP is always
--- a/model/core/interrupt_regs.sail
+++ b/model/core/interrupt_regs.sail
@@ -62,13 +62,11 @@ private function legalize_mideleg(_o : Minterrupts, v : xlenbits) -> Minterrupts
   // in `mip` and `mie` (see `sys/sys_control.sail:getPendingSet()`,
   // they will not be used (see
   // `sys/sys_control.sail:findPendingInterrupt()`).
+  //
+  // Note also that M-mode interrupts (MTI, MSI, MEI) can be delegated;
+  // see https://github.com/riscv/riscv-isa-manual/issues/3162
   [v with
     LCOFI = if currentlyEnabled(Ext_Sscofpmf) then v[LCOFI] else 0b0,
-    // M-mode interrupts can be delegated;
-    // see https://github.com/riscv/riscv-isa-manual/issues/3162
-    MEI   = v[MEI],
-    MTI   = v[MTI],
-    MSI   = v[MSI],
     SEI   = if currentlyEnabled(Ext_S) then v[SEI] else 0b0,
     STI   = if currentlyEnabled(Ext_S) then v[STI] else 0b0,
     SSI   = if currentlyEnabled(Ext_S) then v[SSI] else 0b0,
@@ -76,9 +74,9 @@ private function legalize_mideleg(_o : Minterrupts, v : xlenbits) -> Minterrupts
 }
 
 bitfield Medeleg : bits(64) = {
-  HardwareError     : 19,
-  SoftwareCheck     : 18,
-  DoubleTrap        : 16,
+  Hardware_Error    : 19,
+  Software_Check    : 18,
+  Double_Trap       : 16,
   SAMO_Page_Fault   : 15,
   Load_Page_Fault   : 13,
   Fetch_Page_Fault  : 12,
@@ -102,7 +100,7 @@ private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
   // `sys/sys_control.sail:exception_delegatee()`).
   [Mk_Medeleg(v & plat_medeleg_delegatable_bits) with
     // M-EnvCalls and DoubleTrap cannot be delegated.
-    DoubleTrap         = 0b0,
+    Double_Trap        = 0b0,
     MEnvCall           = 0b0,
   ]
 }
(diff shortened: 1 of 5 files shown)


model/core/interrupt_regs.sail
jordancarlin, line 1: Unrelated to this PR, but I just noticed, we should rename or split this file. `medeleg` is not related to interrutps.

model/postlude/validate_config.sail
```diff
@@ -550,6 +550,33 @@ private function check_extension_param_constraints() -> bool = {
...
+    };
+    if valid & (medelegation.bits != (legalize_medeleg(Mk_Medeleg(zeros()), config base.medeleg.delegatable_bits)).bits) then {
+      valid = false;
+      print_endline("Bits for reserved or custom exceptions are set in `base.medeleg.delegatable_bits`.");
+    };
+
+    let midelegation = Mk_Minterrupts(config base.mideleg.delegatable_bits);
+    if midelegation[MEI] != 0b0 | midelegation[MTI] != 0b0 | midelegation[MSI] != 0b0 then {
+      valid = false;
+      print_endline("One or more bits for machine-mode interrupts (i.e. bits 3, 7, 11) are set in `base.mideleg.delegatable_bits`; machine-mode interrupts cannot be delegated.");
```
jordancarlin: Unfortunately, this is not actually true. See https://github.com/riscv/riscv-isa-manual/issues/3162.

model/postlude/validate_config.sail
```diff
@@ -550,6 +550,33 @@ private function check_extension_param_constraints() -> bool = {
...
+      valid = false;
+      print_endline("Bit 16 (Double Trap) in `base.medeleg.delegatable_bits` is set; Double Trap cannot be delegated.");
+    };
+    if medelegation[MEnvCall] != 0b0 then {
+      valid = false;
+      print_endline("Bit 11 (Environment call from M-mode) in `base.medeleg.delegatable_bits` is set; this environment call cannot be delegated.");
+    };
+    if valid & (medelegation.bits != (legalize_medeleg(Mk_Medeleg(zeros()), config base.medeleg.delegatable_bits)).bits) then {
+      valid = false;
+      print_endline("Bits for reserved or custom exceptions are set in `base.medeleg.delegatable_bits`.");
```
jordancarlin: I'm not convinced we should prohibit someone from making the custom bits writable in the config if that is what their implementation does, but could go either way.

jordancarlin (changes requested):

model/core/interrupt_regs.sail
pmundkur: Yeah, I noticed that too, we should move `medeleg` back to `sys_regs`.

model/postlude/validate_config.sail
pmundkur: Huh, I missed that; it's a bit odd but anyway...

model/postlude/validate_config.sail
pmundkur: Agreed; I'll remove the constraint for custom bits.

model/postlude/validate_config.sail
jordancarlin: It's very odd, but I don't think we can prohibit something that the spec says is allowed...

model/core/interrupt_regs.sail
jordancarlin: Agreed. Something for a future PR.

jordancarlin (approved):
LGTM. The PR description needs to be updated to remove the mention of custom bits.

model/postlude/validate_config.sail
```diff
@@ -550,6 +550,30 @@ private function check_extension_param_constraints() -> bool = {
...
+      print_endline("Bits for reserved exceptions are set in `base.medeleg.delegatable_bits`.");
+    };
+
+    let midelegation = Mk_Minterrupts(config base.mideleg.delegatable_bits);
+    // Ignore bits set for interrupts designated for platform use.
+    let reserved_interrupt_mask : xlenbits = zero_extend(0xD555);
+    if valid & (midelegation.bits & reserved_interrupt_mask) != zeros() then {
+      valid = false;
+      print_endline("Bits for reserved interrupts are set in `base.mideleg.delegatable_bits`.");
+    }
```
```suggestion
    if (medelegation.bits & reserved_exception_mask) != zeros() then {
      valid = false;
      print_endline("Bits for reserved exceptions are set in `base.medeleg.delegatable_bits`.");
    };

    let midelegation = Mk_Minterrupts(config base.mideleg.delegatable_bits);
    // Ignore bits set for interrupts designated for platform use.
    let reserved_interrupt_mask : xlenbits = zero_extend(0xD555);
    if (midelegation.bits & reserved_interrupt_mask) != zeros() then {
      valid = false;
      print_endline("Bits for reserved interrupts are set in `base.mideleg.delegatable_bits`.");
    }
```
Is there a reason why this is gated on valid? None of the other cases in this or the other files do this. I think it would be better to remove it to keep things consistent with the rest of the file.

model/core/interrupt_regs.sail
```diff
@@ -84,9 +88,12 @@ bitfield Medeleg : bits(64) = {
   Fetch_Addr_Align  : 0
 }
 
-private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
-  // M-EnvCalls delegation is not supported
-  [Mk_Medeleg(v & plat_medeleg_delegatable_bits) with MEnvCall = 0b0]
+function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
```
```suggestion
private function legalize_medeleg(_o : Medeleg, v : bits(64)) -> Medeleg = {
```

model/postlude/validate_config.sail
pmundkur: Yeah, I was planning to clean this up today.  I was initially using `legalize_medeleg` to do the validation; this was left over from that and I forgot to remove it.  Same for the change in `private` below.

nadime15 (approved):

model/core/interrupt_regs.sail
```diff
@@ -53,21 +53,32 @@ private function legalize_mie(o : Minterrupts, v : xlenbits) -> Minterrupts = {
...
+  [v with
     LCOFI = if currentlyEnabled(Ext_Sscofpmf) then v[LCOFI] else 0b0,
-    MEI   = 0b0,
-    MTI   = 0b0,
-    MSI   = 0b0,
+    // M-mode interrupts can be delegated;
+    // see https://github.com/riscv/riscv-isa-manual/issues/3162
+    MEI   = v[MEI],
+    MTI   = v[MTI],
+    MSI   = v[MSI],
```
Timmmm:
```suggestion
```

These lines don't do anything.

model/core/interrupt_regs.sail
```diff
@@ -53,21 +53,32 @@ private function legalize_mie(o : Minterrupts, v : xlenbits) -> Minterrupts = {
...
     SEI   = if currentlyEnabled(Ext_S) then v[SEI] else 0b0,
     STI   = if currentlyEnabled(Ext_S) then v[STI] else 0b0,
     SSI   = if currentlyEnabled(Ext_S) then v[SSI] else 0b0,
   ]
 }
 
 bitfield Medeleg : bits(64) = {
+  HardwareError     : 19,
+  SoftwareCheck     : 18,
+  DoubleTrap        : 16,
```
Timmmm:
```suggestion
  Hardware_Error     : 19,
  Software_Check     : 18,
  Double_Trap        : 16,
```

To match the formatting of `union ExceptionType` (also maybe update `Reserved_19` there?)

model/postlude/validate_config.sail
```diff
@@ -550,6 +550,30 @@ private function check_extension_param_constraints() -> bool = {
...
+    };
+    if medelegation[MEnvCall] != 0b0 then {
+      valid = false;
+      print_endline("Bit 11 (Environment call from M-mode) in `base.medeleg.delegatable_bits` is set; this environment call cannot be delegated.");
+    };
+    let reserved_exception_mask : bits(64) = zero_extend(0x0_FFFF_00F2_4400);
+    if (medelegation.bits & reserved_exception_mask) != zeros() then {
+      valid = false;
+      print_endline("Bits for reserved exceptions are set in `base.medeleg.delegatable_bits`.");
+    };
```
Timmmm, line 567: I'm not sure about this. Where does it say that reserved bits can't be delegated?

Timmmm (changes requested):

model/postlude/validate_config.sail
pmundkur: They could be, but this doesn't support it for now.  I clarified the comments in the config file.

Timmmm (approved):

PR #1810, opened 2026-07-16
Q Extension

fourth-bit:
Adds support for the Q extension, including the relevant instructions in Zfa and Zfh. The bindings for softfloat pass each value as two uint64s, so they go through an intermediate function unlike the other bindings. This was chosen to avoid dealing with unpacking and repacking sail internals as 128-bit values would be passed as `lbits`. It also widens the `mem_read` and  `mem_write` functions to support the 128-bit memory accesses, and expands the number of bits passed to `freg_write_callback`.

This PR builds off of @jordancarlin's work in #445, resolving all conflicts from the past year and modified according to the other changes to the floating point extensions that has happened since then.


davidharrishmc:
Some bugs from a Claude review

1. fmul.q computes addition — model/core/softfloat_interface.sail:64

sail
val extern_f128Mul = pure {cpp: "softfloat_f128add", lem: "softfloat_f128_add"} : ...

2. feq.q / flt.q / fle.q decode to the D-format instruction — model/extensions/FD/qext_insts.sail:684,688,692

sail
mapping clause encdec = F_BIN_X_TYPE_D(rs2, rs1, rd, FEQ_D)
  <-> 0b101_0011 @ encdec_freg(rs2) @ encdec_freg(rs1) @ 0b010 @ encdec_reg(rd) @ 0b101_0011
  when currentlyEnabled(Ext_Q)

funct7 1010011 is the Q (fmt=11) encoding, but it constructs F_BIN_X_TYPE_D with FEQ_D/FLT_D/FLE_D. Consequences:
- The quad compares run the 64-bit semantics on F_or_X_D(rs), which now unboxes — so any operand that isn't a NaN-boxed double turns into a canonical NaN. Results and fflags are wrong for essentially every real quad input.
- F_BIN_X_TYPE_Q's three execute clauses (lines 772/784/796) are dead code; nothing constructs that union member.
- Disassembly prints feq.d, and F_BIN_X_TYPE_D(_,_,_,FEQ_D) now has two encdec clauses, so the encode direction of the mapping is ambiguous.


3. Quad NaN quiet-bit tested at the wrong index — qext_insts.sail:101,108

sail
function f_is_SNaN_Q(x128) = (exp == ones()) & (mant [51] == bitzero) & (mant != zeros())
function f_is_QNaN_Q(x128) = (exp == ones()) & (mant [51] == bitone)


4. fmvh.x.d breaks once FLEN=128 — zfa_insts.sail:704

rF_D was changed to unbox (fdext_regs.sail:250), which is right for arithmetic, but FMVH.X.D is a transfer instruction:

sail
let rs1_val_D = F_D(rs1)[63..32];   // now nan_unbox(F(rs1))[63..32]


5. legalize_misa still has Q stubbed out — model/core/sys_regs.sail:118

sail
// Q = if hartSupports(Ext_Q) then v[Q] else 0b0, TODO: Not supported yet

6. FP trace lines get a doubled 0x prefix — c_emulator/riscv_callbacks_log.cpp:75

fprintf(trace_log, "f%d <- 0x", reg);
gmp_fprintf(trace_log, "0x%0*ZX\n", value.len / 4, *value.bits);

7. Lem stub arity mismatch — handwritten_support/riscv_extras_fdext.lem:101

val softfloat_f128_sqrt : bitvector -> bitvector -> bitvector -> (bitvector * bitvector * bitvector)
let softfloat_f128_sqrt _ _ = fail

Three arguments declared, two bound. (softfloat_f128_round_to_int right above gets it right with four.)

8. fclass.q disassembles as fclass.d — qext_insts.sail:860.


Q defaults to enabled — config/config.json.in:451 adds "Q": {"supported": true} unconditionally. Since flen_bytes is derived directly from ext_q_supported, this makes the default model FLEN=128: misa bit 16 set at reset, every f register trace line 32 hex digits wide, and every F/D value NaN-boxed to 128 bits. That silently changes the reference behavior for all existing users, and no profile (RVA22/RVA23) includes Q. It should default to false, or be parameterized through default_config.h.in/CMake the way V is.

flen is derived from raw config, not hartSupports. Q.supported=true, D.supported=false yields FLEN=128 with misa.Q=misa.D=0 — an F-only hart with 128-bit registers. The existing D/F pair has the same latent hole, but the model already internal_errors on the F+Zfinx contradiction; a matching check for Q-without-D would be cheap and worth adding while you're here.

Inconsistent RV64-only gating within the PR. FCVT.L.Q/FCVT.LU.Q/FCVT.Q.L/FCVT.Q.LU use the static xlen >= 64 (matching the D extension), while FMVH.X.Q/FMVP.Q.X use the new dynamic in64BitMode() (mirroring FMVH_X_D's in32BitMode()). Both idioms are defensible; mixing them in one PR is not. Pick one and say why.

Naming. Every new Zfa instruction is RISCV_FLI_Q, RISCV_FMINM_Q, RISCV_FMVH_X_Q… while all the existing ones are FLI_D, FMINM_D, FMVH_X_D. The RISCV_ prefix is the legacy style this model moved away from. Also the section comment above RISCV_FMVP_Q_X says // FMVP.X.Q (should be FMVP.Q.X).

davidharrishmc:
@fourth-bit I suggest trying to build Sail with this extension and run it to generate expected results for Q tests using ACT4.  Run those tests on CVW, which has a robust Q implementation.  Also test other precisions. That will catch many of the kinds of bugs above.

jordancarlin:
Spike also already supports Q and can be used to compare results with this implementation.

PR #1813, opened 2026-07-16
Fix hard-coded build paths in Lean emulator build.

pmundkur:


pmundkur:
Hmm, this works locally but not in CI.   Will look at this later...

lean_emulator/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+configure_file("lakefile.toml.in" "${CMAKE_CURRENT_SOURCE_DIR}/lakefile.toml" @ONLY)
+configure_file("lake-manifest.json.in" "${CMAKE_CURRENT_SOURCE_DIR}/lake-manifest.json" @ONLY)
```
jordancarlin, line 2: Does it make sense to output the configured files in the build directory? Then you could run the whole build from there and avoid the symlink step.

pmundkur: I don't know much about the Lean build system.  From what I can tell, `lake` needs a `.lake` directory in the source tree.  Perhaps there is a way of changing that, but I'll leave that to the Lean folks.

ineol: I'm afraid that it it will make it difficult to work on the emulator if the `.lake` directory is not where it should. Is there a reason why the symlink approach was not satisfactory?

pmundkur: It shouldn't be a problem I think, though perhaps `.lake` should be added to `.gitignore` (that's also what `lake new` seems to do).

jordancarlin: The symlink approach is fine.

jordancarlin (approved):

nadime15 (approved):

PR #1816, opened 2026-07-17
Report physical address for physical memory exceptions.

pmundkur:
Since physical memory accesses can now be split, the precise physical memory address needs to be reported for any exception encountered by a physical memory operation. This adds this physical memory address to the `Err` case of `MemoryOpResult`.

The helpers in `vmem_utils` have been updated to appropriately offset any virtual memory addresses reported in exceptions.

Accesses that do not split (e.g. AMOs and cache operations) assert this for physical addresses in exceptions.

All this addresses the `xtval` regression mentioned in the previous commit of #1734. 

Two obsolete FIXMEs have been removed since they were fixed by #1734.

### DIFF
--- a/model/extensions/cfi/zicfiss_insts.sail
+++ b/model/extensions/cfi/zicfiss_insts.sail
@@ -199,11 +199,6 @@ function clause execute SSAMOSWAP(aq, rl, rs2, rs1, width, rd) = {
   // release consistency semantics, using the aq and rl bits, to help
   // implement multiprocessor synchronization."
 
-  // FIXME: the AMO could lie outside a MAG PMA region, which will be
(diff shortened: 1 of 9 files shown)


pmundkur:
This fixes the regression in #1734.

model/extensions/cfi/zicfiss_insts.sail
jordancarlin: Should this FIXME be dropped too?

pmundkur: Yes!

nadime15 (approved):

jordancarlin (approved):

PR #1819, opened 2026-07-18
Add privileged ISA version checks for CSRs since version 1.12.

### LINKED ISSUE #1818: `mconfigptr` should depend on priv >= 1.12
See https://github.com/riscv/riscv-arch-test/issues/1900.

pmundkur:
Checks for version 1.11 and earlier can be added as needed.

Version checks for other features can be similarly added as needed.

Fixes #1818.

### DIFF
--- a/model/core/isa_version.sail
+++ b/model/core/isa_version.sail
@@ -26,6 +26,12 @@ enum Privileged_ISA_Version = {
   Privileged_ISA_1_13,
 }
 
+mapping privileged_isa_version_name : Privileged_ISA_Version <-> string = {
+  Privileged_ISA_1_11 <-> "1.11",
+  Privileged_ISA_1_12 <-> "1.12",
+  Privileged_ISA_1_13 <-> "1.13",
+}
+
 // For convenience, version comparison operators. This relies on the order of values in the enum.
 function privileged_isa_version_lt(x : Privileged_ISA_Version, y : Privileged_ISA_Version) -> bool =
   num_of_Privileged_ISA_Version(x) < num_of_Privileged_ISA_Version(y)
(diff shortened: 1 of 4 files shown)


model/core/isa_version.sail
```diff
@@ -50,6 +56,15 @@ let priv_isa_version : Privileged_ISA_Version = config base.privileged_isa_versi
...
 // available in earlier versions.
 let xenvcfg_csrs_are_defined = priv_isa_version >= Privileged_ISA_1_12
+let mconfigptr_is_defined = priv_isa_version >= Privileged_ISA_1_12
+let mseccfg_csrs_are_defined = priv_isa_version >= Privileged_ISA_1_12
+let mstatush_is_defined = priv_isa_version >= Privileged_ISA_1_12
+// PMP register count is checked in validate_config() by check_version_constraints().
+
+// CSRs added in version 1.13. These CSRs are not
+// available in earlier versions.
+let delegh_csrs_are_defined = priv_isa_version >= Privileged_ISA_1_13
```
jordancarlin, line 70: Now that we are using the version numbers instead of hashes and this is easier to read, it might make more sense to inline these. It seems more readable to see which version each CSR requires where the CSR is defined instead of needing to refer back to this file.

pmundkur: One advantage of the centralized approach is that it's easier to see the things affected by a version in one place, a bit like the prefaces of the spec.  For someone trying to change the version in the config file, it's also easier to see this `isa_version` file for the things that would be affected.  So I'd prefer the current approach unless you strongly prefer inlining the checks.

jordancarlin: Not a strong opinion. Fine to leave it as is.

jordancarlin (approved):

model/core/sys_regs.sail
```diff
@@ -334,9 +334,9 @@ register mseccfg : Seccfg = legalize_mseccfg(Mk_Seccfg(zeros()), zeros())
 mapping clause csr_name_map = 0x747  <-> "mseccfg"
 mapping clause csr_name_map = 0x757  <-> "mseccfgh"
 
-// "mseccfg exists if Zkr is implemented, or if it is required by other processor features."
-function clause is_CSR_accessible(0x747, _, _) = currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp) | currentlyEnabled(Ext_Smmpm) // mseccfg
-function clause is_CSR_accessible(0x757, _, _) = (currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp)) & xlen == 32 // mseccfgh
+// For Sm1p12 and later, "mseccfg exists if Zkr is implemented, or if it is required by other processor features."
+function clause is_CSR_accessible(0x747, _, _) = mseccfg_csrs_are_defined & currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp) | currentlyEnabled(Ext_Smmpm) // mseccfg
```
nadime15:
```suggestion
function clause is_CSR_accessible(0x747, _, _) = mseccfg_csrs_are_defined & (currentlyEnabled(Ext_Zkr) | hartSupports(Ext_Zicfilp) | currentlyEnabled(Ext_Smmpm)) // mseccfg
```

nadime15 (approved):

PR #1820, opened 2026-07-18
Add support for the Zabhlrsc extension (byte/halfword LR/SC)

arichardson:
Widens lrsc_width_valid to accept byte/halfword widths when Zabhlrsc is enabled. Gated behind --enable-experimental-extensions since it is not yet ratified.

Specification: https://github.com/riscv/riscv-isa-manual/pull/3234

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,7 @@ The following unratified extensions are supported and can be enabled using the `
 
 - Zibi extension for conditional branches with immediate operands, v0.6
 - Zvabd extension for vector absolute difference, v0.7
+- Zabhlrsc extension for byte and halfword load reserved/store conditional instructions
 
 **For a list of unsupported extensions and features, see the [Extension Roadmap](https://github.com/riscv/sail-riscv/wiki/Extension-Roadmap).**
 
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -600,6 +600,9 @@
     "Zabha": {
       "supported": true
     },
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -215,6 +215,7 @@ The following unratified extensions are supported and can be enabled using the `
 
 - Zibi extension for conditional branches with immediate operands, v0.6
 - Zvabd extension for vector absolute difference, v0.7
+- Zabhlrsc extension for byte and halfword load reserved/store conditional instructions
```
jordancarlin, line 218: What is the current version number?

jordancarlin (approved):
Seems like a simple and noninvasive extension. No concerns from me regarding merging before ratification.

README.md
arichardson: That is the one open question. We don't have an assigned version since it was split out from the RVY spec a while ago. Maybe best to go with 0.7?

nadime15 (commented):
Looks good, I would just add a change log entry (/doc/ChangeLog.md). Maybe something like:

> - The following unratified extensions have been added:
>     - Zabhlrsc

README.md
pmundkur: It looks like this got ARC approval very recently and hasn't yet made it into official minutes?  We could leave the version number unspeciied.

pmundkur (commented):
LGTM other than adding to `doc/Changelog.md`.

PR #1821, opened 2026-07-20
Fix the condition for `currentlyEnabled(Ext_Svpbmt)`.

pmundkur:
Svpbmt is dynamically enabled by `menvcfg[PBMTE]`; this bit was not examined when checking whether Svpbmt was currently enabled when checking PTE bits for validity.

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -419,7 +419,7 @@ private function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     STCE = if currentlyEnabled(Ext_Sstc) then v[STCE] else 0b0,
     PMM = if currentlyEnabled(Ext_Smnpm) & is_supported_pmm(PM_SMNPM, pmm_mode(v[PMM])) then v[PMM] else o[PMM],
     ADUE = if currentlyEnabled(Ext_Svadu) then v[ADUE] else 0b0,
-    PBMTE = if currentlyEnabled(Ext_Svpbmt) then v[PBMTE] else 0b0,
+    PBMTE = if hartSupports(Ext_Svpbmt) & currentlyEnabled(Ext_Sv39) then v[PBMTE] else 0b0,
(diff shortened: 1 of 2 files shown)


pmundkur:
Extracted from the original #677 into a separate PR.

model/core/sys_regs.sail
```diff
@@ -419,7 +419,7 @@ private function legalize_menvcfg(o : MEnvcfg, v : bits(64)) -> MEnvcfg = {
     STCE = if currentlyEnabled(Ext_Sstc) then v[STCE] else 0b0,
     PMM = if currentlyEnabled(Ext_Smnpm) & is_supported_pmm(PM_SMNPM, pmm_mode(v[PMM])) then v[PMM] else o[PMM],
     ADUE = if currentlyEnabled(Ext_Svadu) then v[ADUE] else 0b0,
-    PBMTE = if currentlyEnabled(Ext_Svpbmt) then v[PBMTE] else 0b0,
+    PBMTE = if hartSupports(Ext_Svpbmt) & currentlyEnabled(Ext_Sv39) then v[PBMTE] else 0b0,
```
jordancarlin, line 422: If we want to go this direction, it seems like the same thing applies to several of the other items in this list. Definitely Sstc and probably also Svadu, Zicfiss, Zicfilp, and Zicboz.

pmundkur: I don't think those extensions are currently wrong (they check the appropriate bits of menvcfg where needed), as far as I can tell.  Svpbmt was wrong since it didn't check its relevant bit.  I think we can fix the consistency issue later?

jordancarlin (approved):

model/core/sys_regs.sail
jordancarlin: That's fine with me.

pmundkur: It looks like the bit was being checked prior to #677, and @Timmmm refactored the logic.  I restored the original logic, which makes this PR unnecessary.  We could make things more consistent for all the extensions later.

PR #1823, opened 2026-07-21
Require privileged ISA version 1.12+ for Svnapot and Svpbmt

nadime15:
Both extensions were ratified alongside priv 1.12, which also introduced the requirement that reserved PTE bits and encodings fault. Rejecting these extensions on 1.11 configs avoids ambiguity in the version gating of the reserved-encoding checks from #677.

### DIFF
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -658,6 +658,21 @@ function check_version_constraints() -> bool = {
                   privileged_isa_version_name(priv_isa_version) ^ " which only allows 16 PMP registers.");
   };
 
+  // These extensions were ratified alongside priv 1.12, which is also when
+  // reserved PTE bits/encodings were required to fault. Allowing them on 1.11
+  // would make the reserved-encoding checks ambiguous.
+  if hartSupports(Ext_Svnapot) & priv_isa_version < Privileged_ISA_1_12 then {
+    valid = false;
+    print_endline("Svnapot is enabled, but it requires privileged ISA version 1.12 or later (version is set to " ^
+                  privileged_isa_version_name(priv_isa_version) ^ ").");
+  };
+
+  if hartSupports(Ext_Svpbmt) & priv_isa_version < Privileged_ISA_1_12 then {
+    valid = false;
+    print_endline("Svpbmt is enabled, but it requires privileged ISA version 1.12 or later (version is set to " ^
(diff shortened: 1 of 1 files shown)


jordancarlin:
Based on the spec preface, it looks like Svnapot was introduced alongside priv 1.12, which is when reserved PTEs were required to fault. I wonder if we should require priv 1.12 when Svnapot is enabled to avoid any ambiguity here?

pmundkur:
@jordancarlin could you open a clarification request on the spec for  priv version dependence of extensions like Svnapot?

nadime15:
Yeah good point, thats cleaner. I checked the preface and both `svnapot` and `svpbmt` first appear in the 20211203 release together with priv 1.12. I replaced the fix with a validate_config check that rejects both extensions if priv version < 1.12.

pmundkur:
I'm back, I'll open an issue on the spec shortly...

jordancarlin:
> I'm back, I'll open an issue on the spec shortly...

Thanks. Looks like you beat me to it.

pmundkur:
Looks like this is a corner case, and pending a response soon to https://github.com/riscv/riscv-isa-manual/issues/3267, should we release without this?

jordancarlin:
> Looks like this is a corner case, and pending a response soon to [riscv/riscv-isa-manual#3267](https://github.com/riscv/riscv-isa-manual/issues/3267), should we release without this?

Fine with me. I doubt anyone is actually using Priv 1.11 with Svnapot.

nadime15:
Maybe we can wait another 1-2 hours. If we dont get any clarification by then, we can create a new release without this change.

pmundkur:
@nadime15  I'm afraid I have to head out soon.  I'll merge the version bump now and put out a release tonight.

nadime15:
@pmundkur Sure, go ahead!

jordancarlin:
Based on https://github.com/riscv/riscv-isa-manual/issues/3267, it sounds like we don't want to add this requirement. @nadime15, are you ok with closing?

nadime15:
@jordancarlin Sure!

Closing this since Svnapot and Svpbmt are compatible with Privileged Spec v1.11, even though they are defined in Privileged Spec v1.12. See https://github.com/riscv/riscv-isa-manual/issues/3267.

PR #1824, opened 2026-07-21
Do not mask translated physical addresses to the implemented width.

nadime15:
The privileged spec explicitly disallows this, see the following quote: "The algorithm does not admit the possibility of ignoring high-order PPN bits for implementations with narrower physical addresses." A leaf PTE with PPN bits above the implemented width must produce the full physical address, which then faults at the PMA/PMP checks since no memory region can exist there (enforced by check_pma_regions). Masking instead silently aliased such accesses into implemented memory, and was also inconsistent with Bare mode and non-leaf PTE pointers, which are not masked.

That's a quick follow-up to #1412


pmundkur (approved):
Good catch!

jordancarlin (approved):

PR #1830, opened 2026-07-22
Fix usage of `get_16_random_bits` for the `seed` CSR.

### LINKED ISSUE #1829: `seed` CSR `OPST` field intermittently returning `DEAD`
With the latest 0.13 release, reading the `seed` CSR gives `DEAD` instead of `ES16` about 50% of the time. The relevant Sail code (below) seems to indicate we should always be getting `ES16`, so something strange is going on here. Version 0.12 of the model always returned `ES16`.

```sail
private function read_seed_csr() -> xlenbits = {
  let reserved_bits : bits(6) = 0b000000;
  let custom_bits : bits(8) = 0x00;
  let seed : bits(16) = get_16_random_bits();
  zero_extend(opst_code(ES16) @ reserved_bits @ custom_bits @ seed)
}
```

The attached zip has a test that reads from `seed` 100 times along with the results for version 0.12 and 0.13 of the model.

[sail-zkr-seed-issue.zip](https://github.com/user-attachments/files/30284791/sail-zkr-seed-issue.zip)

pmundkur:
The Sail declaration is:
```
val get_16_random_bits = impure {
    interpreter: "Platform.get_16_random_bits",
    cpp: "plat_get_16_random_bits",
    lem: "plat_get_16_random_bits"
} : unit -> bits(16)
```

while the C++ implementation is:

```
mach_bits ModelImpl::plat_get_16_random_bits(unit) {
  return m_gen64();
}
```

This resulted in the generated C++ or-ing more than 16 bits in `read_seed_csr`:
```
  zseed = plat_get_16_random_bits(UNIT);
  z2zE19672 = (zcustom_bits << 16) | zseed;        // this has 64 random bits
  z2zE19673 = (zreserved_bits << 24) | z2zE19672;
  z2zE19674 = (UINT64_C(0b10) << 30) | z2zE19673;  // which corrupt the 0b10 (ES16) bits.

```

This fixes `plat_get_16_random_bits` to return 16 random bits as required.

Fixes #1829.

### DIFF
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -206,7 +206,7 @@ unit ModelImpl::tlb_flush_end_callback(TLB tlb) {
 mach_bits ModelImpl::plat_get_16_random_bits(unit) {
   // This function can be changed to support deterministic sequences of
   // pseudo-random bytes. This is useful for testing.
-  return m_gen64();
+  return m_gen64() & 0xFFFF;
 }
 
 // Note: Store-Conditionals are allowed to spuriously fail. If you want
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -1,5 +1,8 @@
 # Release notes for the next version
 
+- Important issues addressed and bugs fixed:
+  - https://github.com/riscv/sail-riscv/issues/1829 : seed CSR OPST field contained random values
+
 # Release notes for version 0.13
 
 The main highlights of this release are the addition of the misaligned


pmundkur:
I've not yet looked into when the regression happened.

jordancarlin:
It seems like this might be indicative of a bug in the Sail compiler. Shouldn't `let seed : bits(16) = get_16_random_bits();` force it to only use 16 bits even if `get_16_random_bits` returns 64 bits? Or raise an error?

jordancarlin (approved):

jordancarlin:
Also interesting that our regression tests never caught this. I ran into it immediately when trying to use it with ACT4.

pmundkur:
Well, the compiler could play it safe and mask extra bits, but it is entitled to assume that the
```
val get_16_random_bits = impure {
    cpp: "plat_get_16_random_bits"
} : unit -> bits(16)
```
contract is satisfied.

Alasdair:
Yes, the implementation of the primitive has to respect whatever type it has been given in Sail.

Alasdair (approved):

Timmmm (approved):
Might be nice if the Sail compiler had an option to add assertions to check return values conform to the expectation. Dunno how hard that would be though.

PR #1831, opened 2026-07-23
Add a `--trace-pmp` option to aid debugging access faults due to PMP.

pmundkur:
This complements the existing `--trace-pma` option.


jordancarlin (approved):
LGTM.

It would be nice to also print out the contents of the pmpcfg/pmpaddr CSRs that caused the check to fail, but that can happen in a future enhancement to this logging.

Timmmm (approved):
Very useful!

PR #1833, opened 2026-07-24
Change Zama16b validation check to require a MAG value of at least 16 bytes.

### LINKED ISSUE #1832: Zama16b should accommodate misaligned_atomicity_granule_size_exp >= 4
With Zama16b = true, when I change misaligned_atomicity_granule_size_exp to 6 to match a cache line size, I get:

```
    Configuration in /home/harris/cvw/addins/riscv-arch-test-dh/config/imperas/imperas-rv64-max/sail.json is invalid.
    Main memory region starting at 0x0000000080000000 is coherent and cacheable with a misaligned atomicity granule size of 2^6, but Zama16b is enabled which requires granule size of 2^4 bytes.
```

Zama16b is defined as:
```
Zama16b Misaligned loads, stores, and AMOs to main memory regions that do not cross a
naturally aligned 16-byte boundary are atomic.
```

So I think the constraint should be granule size of >= 2^4 bytes, not exactly 2^4 bytes.

pmundkur:
Fixes #1832.

### DIFF
--- a/doc/ChangeLog.md
+++ b/doc/ChangeLog.md
@@ -1,7 +1,7 @@
 # Release notes for the next version
 
 - Important issues addressed and bugs fixed:
-  - https://github.com/riscv/sail-riscv/issues/1832 : Zama16b validation check should require at least 16 bytes
+  - https://github.com/riscv/sail-riscv/issues/1832 : Zama16b validation check should require at least 16 bytes, not exactly 16 bytes
   - https://github.com/riscv/sail-riscv/issues/1829 : seed CSR OPST field contained random values
 
 # Release notes for version 0.13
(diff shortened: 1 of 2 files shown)


doc/ChangeLog.md
```diff
@@ -1,6 +1,7 @@
 # Release notes for the next version
 
 - Important issues addressed and bugs fixed:
+  - https://github.com/riscv/sail-riscv/issues/1832 : Zama16b validation check should require at least 16 bytes
```
jordancarlin: The bug is unclear with the current wording.
```suggestion
  - https://github.com/riscv/sail-riscv/issues/1832 : Zama16b validation check should require at least 16 bytes, not exactly 16 bytes
```

jordancarlin (commented):
Minor Changelog nit. Otherwise, LGTM.

jordancarlin (approved):

nadime15 (approved):

PR #1835, opened 2026-07-24
Fix merging config overrides for option values in the config.

pmundkur:
The recursive merge fails for option objects because they have different head keys. So
```
 {"load_store" : {"None": null}}
```
when overridden by
```
 {"load_store" : {"Some": "AlignmentException"}}
```
created
```
 {"load_store" : {"Some": "AlignmentException", "None": null}}
```
which was rejected by the schema validation check.

This updates the merging algorithm used for config overrides to handle such changes to variant-valued parameters. 

It also unifies the description of the configuration when used for reporting configuration errors and the result of `--validate-config`.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -47,20 +47,88 @@ jsoncons::json parse_json_or_exit(const std::string &json_text, const std::strin
   }
 }
 
-// JSON objects are merged by replacing/adding fields from the override config.
-// If a given field is an object in the base and override then instead of
-// replacing the field entirely the merging process recurses into it.
-// All other field types (including arrays) are simply replaced.
-void deep_merge_json(jsoncons::json &base, const jsoncons::json &json_override) {
-  for (const auto &entry : json_override.object_range()) {
-    const auto &key = entry.key();
-    const auto &value = entry.value();
-    if (base.contains(key) && base[key].is_object() && value.is_object()) {
-      deep_merge_json(base[key], value);
-    } else {
-      base[key] = value;
-    }
+std::string make_path_string(const std::vector<std::string> &path) {
+  // This uses '/' separators for the path elements to be consistent
+  // with the error messages from the jsoncons schema validator.
+  if (path.empty()) {
+    return "path '/'";
   }
+
+  std::ostringstream buf;
+  buf << "path '";
+  for (const auto &seg : path) {
+    buf << "/" << seg;
+  }
+  buf << "'";
+  return buf.str();
+}
+
+// Merge a JSON object `json_override` into an existing JSON object `base`. The
+// algorithm is as follows:
+//
+// 1. If both `json_override` and `base` are objects we merge them as
+//    follows:
+//   a) If both objects have a single field and that key has a
+//      different name, this must be a union whose variant is being
+//      changed (e.g. `None` to `Some`). In this case we replace the
+//      entire `base` with `json_override`.
+//      Note: the key might be different due to a typo in the override,
+//      but that will be caught by the schema validation check.
+//   b) Otherwise this is a struct, or a union with an unchanged
+//      variant. In this case we loop through the keys of
+//      `json_override` and recurse for each one, merging into the
+//      corresponding key of `base` which must exist or an error is
+//      thrown (you cannot create new keys with this function).
+//
+// 2. If either are not objects, then we replace `base` with
+//    `json_override`. Additionally we check for type errors,
+//    e.g. replacing an object with an array.
+
+void deep_merge_json(jsoncons::json &base, const jsoncons::json &json_override, std::vector<std::string> &path) {
+  if (base.type() == jsoncons::json_type::object && json_override.type() == jsoncons::json_type::object) {
+    // If both are objects with size 1 this must be a union that is
+    // changing its variant.  Replace the entire object.  See above
+    // for the effects of a typo.
+    if (base.size() == 1 && json_override.size() == 1 &&
+        base.object_range().begin()->key() != json_override.object_range().begin()->key()) {
+      base = json_override;
+      return;
+    }
+
+    // Otherwise this is a struct or a union with the same
+    // variant. Merge each key as long as it exists in the base.
+    for (const auto &entry : json_override.object_range()) {
+      const auto &key = entry.key();
+      const auto &value = entry.value();
+
+      path.emplace_back(key);
+
+      if (!base.contains(key)) {
+        std::ostringstream msg;
+        msg << "Cannot merge override: key at " << make_path_string(path)
+            << " does not exist in base object; config overrides cannot add new keys to the configuration.";
+        throw std::runtime_error(msg.str());
+      }
+
+      deep_merge_json(base[key], value, path);
+      path.pop_back();
+    }
+    return;
+  }
+
+  // At least one of the arguments is not an object. Replace the base
+  // with the override, but as long as the types are consistent.
+  auto base_type = base.type();
+  auto override_type = json_override.type();
+
+  if (base_type != override_type) {
+    std::ostringstream msg;
+    msg << "Cannot override config at " + make_path_string(path) + " of type '" << base_type
+        << "' with incompatible type '" << override_type << "'.";
+    throw std::runtime_error(msg.str());
+  }
+
+  base = json_override;
 }
 
 void write_dtb_to_rom(ModelImpl &model, const std::vector<uint8_t> &dtb) {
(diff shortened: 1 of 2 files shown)


pmundkur:
We don't have unit tests for this kind of thing, but I'm working on something which needs this so it will be tested that way.

c_emulator/riscv_sim.cpp
```diff
@@ -47,16 +47,48 @@ jsoncons::json parse_json_or_exit(const std::string &json_text, const std::strin
...
+  int key_count = 0;
+  bool is_none = false;
+  bool is_some = false;
+  for (const auto &entry : object.object_range()) {
+    is_none = entry.key() == "None";
+    is_some = entry.key() == "Some";
+    ++key_count;
+  }
+  return key_count == 1 && (is_none || is_some);
+}
```
jordancarlin: Since we are checking that there is a single key, no need to count `None` and `Some` separately. Fine to leave it this way if you prefer, but I think the other approach is probably slightly simpler.

c_emulator/riscv_sim.cpp
```diff
@@ -47,16 +47,48 @@ jsoncons::json parse_json_or_exit(const std::string &json_text, const std::strin
...
-// replacing the field entirely the merging process recurses into it.
-// All other field types (including arrays) are simply replaced.
+// JSON objects are merged by replacing/adding fields from the
+// override config. If a given field is a non-option object in the
+// base and in the override then instead of replacing the field
+// entirely the merging process recurses into it. Option objects and
+// all other field types (including arrays) are simply replaced.
+//
+// The exception for option objects is for the following reason:
+//   {"load_store" : {"Null": null}}
```
```suggestion
//   {"load_store" : {"None": null}}
```

I think?

jordancarlin (approved):

Timmmm:
Hmm I dunno if this works in general - as far as I understand it options are just unions, so we'd have the same issue with e.g.

```
 {"some_union" : {"Variant_A": null}}
```

Being overridden with `Variant_B`. I guess the fundamental problem is there's no way to tell just from the JSON whether something is a struct or a union.

This is probably the pragmatic option since as far as I can remember we don't actually have any other unions in the config... But it's a little unsatisfying that it's a special case. I wonder if there's a way we can make it work in general? Like maybe the JSON itself says whether it should replace the entire object or just fields of it?

Actually I guess could you instead just use "does the replacement key already exist in the object? If not then it must be an enum and we replace the entire object"?

pmundkur:
> Hmm I dunno if this works in general - as far as I understand it options are just unions, so we'd have the same issue with e.g.
> 
> ```
>  {"some_union" : {"Variant_A": null}}
> ```
> 
> Being overridden with `Variant_B`. I guess the fundamental problem is there's no way to tell just from the JSON whether something is a struct or a union.

Right, we would need to add similar checks for other variants if we expose them in the config.

> 
> This is probably the pragmatic option since as far as I can remember we don't actually have any other unions in the config... But it's a little unsatisfying that it's a special case. I wonder if there's a way we can make it work in general? Like maybe the JSON itself says whether it should replace the entire object or just fields of it?

I think the most general way would be to use the constraints in the config schema to guide the replacement.  That will be a lot more complicated though, I think.

> 
> Actually I guess could you instead just use "does the replacement key already exist in the object? If not then it must be an enum and we replace the entire object"?

That might work, but I worry about new unexpected failure cases.   This is a narrow targeted fix.

c_emulator/riscv_sim.cpp
pmundkur: I don't follow: you could have a single unrelated key:
```
"sto": {
   "is_nop": false
 }
```
I'd prefer to make sure it is one of "Some" or "None".  But perhaps I'm missing what you have in mind?

jordancarlin: I was just saying you could do `entry.key() == "None" || entry.key() == "Some"` and save it to a single variable. No big deal either way though.

Timmmm:
Hmm I think I would prefer the general solution that works for the future too. Also I noticed that we are a bit more lax in the merge algorithm than necessary - e.g. it should never be necessary to add fields to an object in the first place, or to change the type of a field, both of which are currently permitted. I think we should do it like this:

```

// Merge a JSON object `json_override` into an existing JSON object `base`. The
// algorithm is as follows:
//
// 1. Both `json_override` and `base` are objects we merge them as follows.
//    a) If both objects have a single field and that key has a different name, this must
//       be a union whose variant is being changed (e.g. `none` to `some`). In this case
//       we replace the entire `base` with `json_override`.
//    b) Otherwise this is a struct, or a union with an unchanged variant. In this
//       case we loop through the keys of `json_override` and recurse for each one,
//       merging into the corresponding key of `base` which must exist or an error is thrown
//       (you cannot create new keys with this function).
// 2. If either are not objects, then we replace `base` with `json_override`.
//    Additionally we check for type errors, e.g. replacing an object with an array.
void deep_merge_json(jsoncons::json &base, const jsoncons::json &json_override) {
  if (base.is_object() && json_override.is_object()) {
    // If both are objects with size 1 this must be a union changing variant.
    // Replace the entire object.
    if (base.size() == 1 && json_override.size() == 1 && base.object_range().begin()->key() != json_override.object_range().begin()->key()) {
      base = json_override;
    } else {
      // Otherwise this is a struct or a union with the same variant. Merge each key.
      for (const auto &entry : json_override.object_range()) {
        const auto &key = entry.key();
        const auto &value = entry.value();
        // We never want to create new keys - all keys must already exist in the base.
        if (!base.contains(key)) {
          throw std::runtime_error("Cannot merge JSON: key '" + key + "' does not exist in base object. You cannot add keys via JSON override.");
        }
        deep_merge_json(base[key], value);
      }
    }
  } else {
    // At least one of the arguments is not an object. Replace the base with the override,
    // but also verify that we aren't doing something silly with the types like
    // replacing an array with a string.

    auto base_type = simple_json_type(base);
    auto override_type = simple_json_type(json_override);

    if (base_type != override_type) {
      // TODO: It would be much nicer to have the full JSON path here!
      throw std::runtime_error("Cannot override JSON value with type '" + std::string(simple_json_type_name(base_type)) + "' with incompatible type '" + std::string(simple_json_type_name(override_type)));
    }

    base = json_override;
  }
}
```

(Not tested, sorry!)

pmundkur:
This sounds right to me.  I'll test this out, thanks!

pmundkur:
@Timmmm your algo works great!  In step 1a), there could be an erroneous merge due to a typo in the override, but that is caught by the schema validation.

pmundkur:
I'll merge this today if there are no objections.

c_emulator/riscv_sim.cpp
```diff
@@ -47,20 +47,86 @@ jsoncons::json parse_json_or_exit(const std::string &json_text, const std::strin
...
+// algorithm is as follows:
+//
+// 1. If both `json_override` and `base` are objects we merge them as
+//    follows:
+//   a) If both objects have a single field and that key has a
+//      different name, this must be a union whose variant is being
+//      changed (e.g. `None` to `Some`). In this case we replace the
+//      entire `base` with `json_override`.
+//      Note: the key might be different due to a typo in the override,
+//      but that will be caught by the schema validation check.
```
nadime15, line 74: I tried this on your branch, and `sail_riscv_sim` accepted the following (see the typo "Sme"):
```
~/work/sail/sail-riscv$ echo '{"memory":{"misaligned":{"exceptions":{"load_store":{"Sme":"AlignmentException"}}}}}' > test.json
~/work/sail/sail-riscv$ build/c_emulator/sail_riscv_sim --config-override test.json --validate-config
The default configuration merged with test.json is valid.
```

Prior to that change, this was rejected

pmundkur: Hmm, this seems to be a bug in the jsoncons schema validation.  It had worked for the few different cases I tried.

pmundkur: So if you make the change directly to `config.json.in` (and hence bypass this algorithm) it still will build and erroneously pass validation checks, I think.

nadime15: I have no idea how this works internally, but could this actually be a Sail bug? It looks like Sail generates an incomplete JSON Schema, which would explain why the validator accepts arbitrary variant names. @Alasdair

pmundkur: Right, that's the other possibility.  I've opened #1850 to track this as a separate issue.

Alasdair: Maybe there is a missing maxProperties on the schema? JSON schema will let you add extra keys to an object without it.

pmundkur: `memory.misaligned.exceptions.load_store` has both `minProperties` and `maxProperties` set to 1.

c_emulator/riscv_sim.cpp
```diff
@@ -47,20 +47,86 @@ jsoncons::json parse_json_or_exit(const std::string &json_text, const std::strin
...
+  // with the override, but as long as the types are consistent.
+  auto base_type = base.type();
+  auto override_type = json_override.type();
+
+  if (base_type != override_type) {
+    std::ostringstream msg;
+    msg << "Cannot override config at path '" + make_path_string(path) + "' of type '" << base_type
+        << "' with incompatible type '" << override_type << "'.";
+    throw std::runtime_error(msg.str());
+  }
```
Timmmm, line 127: The only thing I wasn't 100% sure about here was if we ever want to replace `null` with other things. But I'm happy to go with it and fix it later if that situation ever turns up.

Timmmm (approved):

c_emulator/riscv_sim.cpp
Timmmm: Sorry ignore this comment - I forgot to submit it and then I couldn't figure out how to remove it from the review.

pmundkur:
@nadime15  any objections to merging this?

nadime15:
@pmundkur Nope, go ahead :)

nadime15 (approved):

PR #1837, opened 2026-07-24
Add a test for the `physaddr_bits` config parameter.

pmundkur:
This is currently only used with the `satp` and PMP CSRs.


Timmmm (approved):
Not read in detail but LGTM.

Issue #1807, opened 2026-07-14
Setting misa.S=0 with mideleg non-zero leads to assertion failure on interrupts


davidharrishmc:
Sail throws an assertion failure when an interrupt is taken if misa.S is cleared while mideleg is nonzero:

```
✗ FAILED: SsstrictSm_CSR-00
  Output: work/sail-rv64-max/build/priv/SsstrictSm/SsstrictSm_CSR-00.sig
  Error output:
    Assertion failed: sys/sys_control.sail:107.58-107.59
    begin_signature: 0x8007ad70
    end_signature: 0x800c70a0

  Log file: work/sail-rv64-max/build/priv/SsstrictSm/SsstrictSm_CSR-00.sig.log
  Reproduce with:
    /opt/riscv/bin/sail_riscv_sim \
      --config /home/harris/cvw/addins/riscv-arch-test-dh/config/sail/sail-rv64-max/sail.json \
      --test-signature=/home/harris/cvw/addins/riscv-arch-test-dh/work/sail-rv64-max/build/priv/SsstrictSm/SsstrictSm_CSR-00.sig \
      --signature-granularity 8 \
      /home/harris/cvw/addins/riscv-arch-test-dh/work/sail-rv64-max/build/priv/SsstrictSm/SsstrictSm_CSR-00.sig.elf
```

[Edited by jrtc27 to entirely replace the slop wall:]

Presumably we need to do one of relax the assertion, treat the write as illegal, or clear mideleg on the write, depending on what the spec says.

jrtc27:
Holy wall of AI slop. Can we not, please?

jrtc27:
(You are free to use as much AI as you want locally, but the moment you go interacting with other people it is your responsibility to present it in an appropriate, and validated, manner)

nadime15:
At this point, I have the impression (I am exaggerating a bit since I have not actually checked) that these Ssstrict cases all end up failing eventually (an assertion failure is obviously worse than a simple test failure). There have been so many PRs that broke them that maybe we should exclude them for a while?

Timmmm:
Here's the assertion

```
  // mideleg can only be non-zero if we support Supervisor mode.
  assert(currentlyEnabled(Ext_S) | mideleg.bits == zeros());
```

Since `currentlyEnable(Ext_S)` basically just returns `misa[S]`, that means if you set `mideleg` and then clear `misa[S]` this will assert.

[...]

PR #1839, opened 2026-07-27
Fix unused `mideleg_bits` binding in `getPendingSet()`.

nadime15:
The binding from #1812 never took effect, since `pending_m`/`pending_s` still read `mideleg.bits` directly, so interrupts were delegated to a disabled S-mode (#1807).


pmundkur (approved):

Timmmm (approved):
Oops. Would some kind of `-Werror` in CI have caught this?

jordancarlin (approved):

nadime15:
> Oops. Would some kind of -Werror in CI have caught this?

Good question, I'm not sure. :/

jordancarlin:
There is no -Werror equivalent for Sail itself, and it may have been optimized out by the time it got to C++.

PR #1840, opened 2026-07-27
Implement a basic event framework.

### LINKED ISSUE #1733: Sscofpmf depends on a way to increment a mhpmcounter.
Sscofpmf has a behavior that xip.LCOFIP is set when a hpmcounter overflows.  This is tricky to test because the counter incrementing depends on mhpmevent and on platform-specific behavior of the counters.

I don’t see any Sail code that increments any mhpmcounter

https://github.com/riscv/sail-riscv/blob/master/model/extensions/Zihpm/zihpm.sail

Is there a way in Sail to cause a counter to increment and cause xip.LCOFIP to rise in response?
If not, could this be added so there is a deterministic way to exercise LCOFIP?

For example, if mhpmevent3[31:0] = 1, mhpcounter3 could increment every time an instruction retires.

pmundkur:
This defines a few basic events (TLB misses and misaligned load/store exceptions) around which to build the framework.  The implementation-specific selector codes for the events can be specified in the config file. After a selector value is written to an event selector CSR, the generation of its event results in an increment of the corresponding event counter CSR (subject to Sscofpmf).

To improve efficiency, the event to CSR mapping is implemented in the C++ harness, as Sail doesn't have efficient sets and maps.  Event config validation is also done in the harness.

This framework could be extended in several ways:

- The registered selector codes could be used for WARL legalization of the `hpmevent` CSRs.

- Event selector codes are restricted to 56 bits, as made available by Sscofpmf.  When Sscofpmf is not supported, the use of all 64-bits should be supported.

- The events are counted based on exact matching of selector codes.  It would be good to support bit-mask-based event selection.

Fixes #1733.

### DIFF
--- a/c_emulator/riscv_model_impl.cpp
+++ b/c_emulator/riscv_model_impl.cpp
@@ -2,10 +2,12 @@
 #include <algorithm>
 #include <cassert>
 #include <random>
+#include <set>
 #include <unistd.h>
 
 #include "config_utils.h"
 #include "riscv_callbacks_if.h"
+#include "sail.h"
 #include "symbol_table.h"
 
 void ModelImpl::register_callback(std::shared_ptr<callbacks_if> cb) {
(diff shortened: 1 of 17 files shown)


pmundkur:
Depends on https://github.com/rems-project/sail/pull/1695

jordancarlin:
@Alasdair any chance of a new Sail compiler release soon so we can merge this? This is needed for ACT4 Sscofpmf tests.

PR #1841, opened 2026-07-27
Add validation of the `writable_fiom` config option.

### LINKED ISSUE #1834: Missing validation of `writable_fiom` config option
The ISA manual says
> If S-mode is not supported, or if satp.MODE is read-only zero (always Bare), the implementation may make FIOM read-only zero.

Right now, we allow `base.writable_fiom` to be set to false even if Sv32/Sv39 are not implemented. This should be checked in `validate_config.sail`.

KotorinMinami:
Fixes #1834

### DIFF
--- a/model/postlude/validate_config.sail
+++ b/model/postlude/validate_config.sail
@@ -817,6 +817,17 @@ private function check_extension_param_constraints() -> bool = {
       print_endline("`extensions.H.geilen` is " ^ dec_str(geilen) ^ " but cannot be greater than " ^ dec_str(xlen - 1) ^ " for RV" ^ dec_str(xlen) ^ ".");
     };
   };
+  if not(sys_enable_writable_fiom) then {
+    if xlen == 64 & (config extensions.Sv39.supported : bool) then {
+      valid = false;
+      print_endline("`base.writable_fiom` is false but Sv39 is supported: FIOM may only be read-only zero when satp.MODE is always Bare.");
+    }
+    else if xlen == 32 & (config extensions.Sv32.supported : bool) then {
+      valid = false;
+      print_endline("`base.writable_fiom` is false but Sv32 is supported: FIOM may only be read-only zero when satp.MODE is always Bare.");
+    }
+  };
+
   valid
 }
 


model/postlude/validate_config.sail
```diff
@@ -671,6 +671,21 @@ private function check_physaddr_bits() -> bool = {
...
+    }
+    else if xlen == 64 & not(config extensions.Sv39.supported : bool) then {
+      print_endline("base.writable_fiom is true but only Svbare is enabled.");
+      false
+    }
+    else { true }
+  }
+  else { true }
+}
+
```
nadime15, line 688: I think this needs to be implemented the other way around? When sys_enable_writable_fiom is true, we should allow writes if either Sv32 or Sv39 is enabled. What we should not allow is sys_enable_writable_fiom being false while Sv32 or Sv39 is true.
```suggestion
private function check_fiom_writable() -> bool = {
    let vm_supported = if xlen == 64 then config extensions.Sv39.supported : bool
                       else config extensions.Sv32.supported : bool;
    if not(sys_enable_writable_fiom) & vm_supported then {
      print_endline("`base.writable_fiom` is false but address translation is supported: FIOM may only be
  read-only zero when satp.MODE is always Bare.");
      false
    } else true
  }

```

@jordancarlin

jordancarlin: Agreed. @nadime15's implementation seems correct.

model/postlude/validate_config.sail
```diff
@@ -861,6 +861,18 @@ private function check_physaddr_bits() -> bool = {
...
 
+private function check_fiom_writable() -> bool = {
+  if not(sys_enable_writable_fiom) & xlen == 64 & (config extensions.Sv39.supported : bool) then {
+    print_endline("`base.writable_fiom` is false but Sv39 is supported: FIOM may only be read-only zero when satp.MODE is always Bare.");
+    false
+  }
+  else if not(sys_enable_writable_fiom) & xlen == 32 & (config extensions.Sv32.supported : bool) then {
+    print_endline("`base.writable_fiom` is false but Sv32 is supported: FIOM may only be read-only zero when satp.MODE is always Bare.");
+    false
+  }
```
pmundkur: `not(sys_enable_writable_fiom)` could be hoisted into its own `if` 

```suggestion
  valid = true;
  if not(sys_enable_writable_fiom)) then
    if xlen == 64 & (config extensions.Sv39.supported : bool) then {
      print_endline("`base.writable_fiom` is false but Sv39 is supported: FIOM may only be read-only zero when satp.MODE is always Bare.");
      valid  = false
    } else if not(xlen == 32 & (config extensions.Sv32.supported : bool) then {
    print_endline("`base.writable_fiom` is false but Sv32 is supported: FIOM may only be read-only zero when satp.MODE is always Bare.");
      valid = false
    }
  };
  return valid
```
and this check could be made part of `check_extension_param_constraints()` instead of adding a new function.

pmundkur (approved):

jordancarlin (approved):

jordancarlin:
@KotorinMinami looks like this one needs a rebase.

pmundkur:
@KotorinMinami you will need sign the commit and force-push for this (and in future) to pass the DCO
https://github.com/riscv/sail-riscv/pull/1841/checks?check_run_id=100843531172

model/postlude/validate_config.sail
```diff
@@ -817,6 +817,17 @@ private function check_extension_param_constraints() -> bool = {
       print_endline("`extensions.H.geilen` is " ^ dec_str(geilen) ^ " but cannot be greater than " ^ dec_str(xlen - 1) ^ " for RV" ^ dec_str(xlen) ^ ".");
     };
   };
+  if not(sys_enable_writable_fiom) then {
```
```suggestion
if not(sys_enable_writable_fiom) & (config extensions.S.supported : bool) {
```

nadime15 (approved):

PR #1847, opened 2026-07-29
Use freg_name for LOAD_FP assembly rd

nadime15:


pmundkur (approved):
Great! How did you catch this?

nadime15:
That's actually a follow up for #1827. It was just a coincidence. I wanted to take a screenshot of some of our FP code for the presentation last Monday, and while looking at the FP code I noticed the issue. After that, I went through each case individually (which is just a handful of cases according to the Zfinx spec).

jordancarlin (approved):
I don't think this functionally changes anything since the fp load instructions don't exist with Zfinx, but this is still more correct, so LGTM.

PR #1849, opened 2026-07-29
Switch to an opam cache instead of container for Rocq CI

bacam:
Also starts using the new-style rocq- package names for dependencies.

### DIFF
--- a/.github/workflows/rocq.yml
+++ b/.github/workflows/rocq.yml
@@ -4,16 +4,7 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-    env:
-      OPAMROOT: /home/rocq/.opam
-      # See root comment below
-      OPAMROOTISOK: true
-    container:
-      image: rocq/rocq-prover:9.0.1
-      # github requires root to access some resources from outside the container
-      # (e.g. https://github.com/actions/checkout/issues/47)
-      options: --user root
+    runs-on: ubuntu-24.04
     steps:
       - name: Check out repository code
         uses: actions/checkout@v7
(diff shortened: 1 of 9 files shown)


bacam:
I'll rerun the check later this evening when hopefully the rate limiting on the stdpp server will have calmed down.

jordancarlin (changes requested):
I think a lot of this could be simplified by moving it into the `.github/actions/sail-setup/action.yml` composite action. Add an `install-rocq` input (default false). When it is set to true, the existing `opam` logic can be extended to install `rocq`. Also, should we just install `sail` itself using `opam` in this workflow so we don't need to deal with the fake dependency install?

bacam:
> I think a lot of this could be simplified by moving it into the `.github/actions/sail-setup/action.yml` composite action. Add an `install-rocq` input (default false). When it is set to true, the existing `opam` logic can be extended to install `rocq`. Also, should we just install `sail` itself using `opam` in this workflow so we don't need to deal with the fake dependency install?

I'd prefer to keep this self-contained rather than putting things into the sail action that are specific to this workflow, and I'd like to avoid the time/storage cost of another sail build.  (It's also a nice side-benefit to have an example of how to use a fake opam install of Sail.)

bacam:
I've added another download option for asio.  We had thought about coalescing these into one download-c-libraries option, perhaps we should do that now?

.github/workflows/rocq.yml
```diff
@@ -39,15 +30,24 @@ jobs:
         with:
           sail-version: ${{ env.SAIL_VERSION }}
 
+      - name: Restore cached opam
+        id: cache-opam-restore
+        uses: actions/cache/restore@v6
+        with:
+          path: ~/.opam
+          key: ${{ runner.os }}-${{ runner.arch }}-rocq-opam-sail-${{ env.SAIL_VERSION }}
```
pmundkur: Should the cache key also encode the versions for rocq-core=9.1.1 rocq-stdlib=9.0.0 rocq-stdpp-bitvector=1.13.0?

pmundkur:
> I've added another download option for asio. We had thought about coalescing these into one download-c-libraries option, perhaps we should do that now?

Sounds reasonable;  in a separate PR perhaps?

.github/workflows/rocq.yml
bacam: Yes, and I've moved the versions up to the top of the workflow too.

pmundkur (approved):
Great to have a non-container build.

jordancarlin (approved):

PR #1852, opened 2026-07-30
Fix zicfiss shadow stack page permissions

nadime15:
Two bugs in `check_PTE_permission` that fire the moment a shadow stack page (R=0 W=1 X=0) is accessed.

First is a crash. The assert that a writable page is also readable trips on a shadow stack page, even though `pte_is_invalid` already accepts that encoding as valid. I relaxed the assert to allow it while still catching reserved R=0 W=1 X=1.

Second is a wrong fault. After the shadow stack rules approve an access the code fell through to the normal check, which needs R, so loads, SSPOPCHK, and SSAMOSWAP got rejected with a store/AMO page fault while only SSPUSH worked.

### DIFF
--- a/model/sys/vmem_pte.sail
+++ b/model/sys/vmem_pte.sail
@@ -200,6 +200,11 @@ private function check_PTE_permission(
 
     };
     if not(shadow_stack_ok) then return PTE_Check_Failure((), PTE_No_Access());
+    // On a shadow stack page the rules above fully decide the access. The R, W
+    // and X bits do not apply because the encoding is R=0 W=1 X=0, so grant the
+    // access here rather than falling through to the normal permission check,
+    // which would reject any load or atomic because the page is not readable.
+    return PTE_Check_Success(());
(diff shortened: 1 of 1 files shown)


pmundkur (approved):
Good catches!

nadime15:
Yeah, the https://github.com/XUANTIE-RV/damo-rv-priv-ats/tree/main test suite catches a few issues. They pushed a commit today that adds tests for Hypervisor + (Zicfiss and Zicfilp), which turned out to be really useful.

jordancarlin (approved):

jordancarlin:
> Yeah, the https://github.com/XUANTIE-RV/damo-rv-priv-ats/tree/main test suite catches a few issues. They pushed a commit today that adds tests for Hypervisor + (Zicfiss and Zicfilp), which turned out to be really useful.

And ACT4 non-hypervisor Zicfiss and Zicfilp tests are still under development, so I don't think this part of the model has been tested much yet.

PR #1854, opened 2026-07-30
Refactor `init_model` to use the `InitResult` return type.

pmundkur:
This avoids using a fake entry point when RVFI fails to initialize its socket, resulting in the error getting ignored by the RVFI path.

While here, make the `InitResult` returning functions take a `const` options argument.


pmundkur:
@nadime15 @jordancarlin 'This doesn't conflict with the hypervisor PR and is pretty small.

jordancarlin (approved):
Pretty simple and all looks fine to me.

nadime15 (approved):

PR #1855, opened 2026-07-31
Add `initramfs` support to boot Linux to a shell. 

pmundkur:
This support involves:

- Adding an `--initramfs` option to specify a file containing the `initramfs`.

- The specified `initramfs` has to be loaded into an appropriate place in memory after the kernel has been loaded.

  The `initramfs` is simply placed at the top-end of the first memory region with the space to fit it without conflicts with existing loaded files.

- This location has to be communicated in the device-tree.

  If an `initramfs` is specified, the `--print-device-tree` option is handled after the binaries and `initramfs` are placed into memory.  Some initial logging is suppressed in this case to avoid interfering with the print.  Another option could be to print the device tree to a file, but the logging would still be confusing.

New targets are added to `os-boot/Makefile` to download Busybox and prepare a ramfs image, and to boot Linux with it.  The boot now proceeds to the point where input is needed to activate the terminal shell.

### DIFF
--- a/c_emulator/cli_options.cpp
+++ b/c_emulator/cli_options.cpp
@@ -37,6 +37,10 @@ CLIOptions parse_cli(int argc, char **argv) {
   );
 
   app.add_option("--device-tree-blob", opts.dtb_file, "Device tree blob file")
+    ->excludes("--print-device-tree")
+    ->check(CLI::ExistingFile)
+    ->option_text("<file>");
+  app.add_option("--initramfs", opts.initramfs_file, "File containing the initramfs (for Linux boot)")
     ->check(CLI::ExistingFile)
     ->option_text("<file>");
   app.add_option("--terminal-log", opts.term_log, "Terminal log output file")->option_text("<file>");
--- a/c_emulator/riscv_model_impl.h
+++ b/c_emulator/riscv_model_impl.h
@@ -54,6 +54,7 @@ public:
   void set_config_rvfi(bool on);
   void set_config_use_abi_names(bool on);
 
+  void set_initramfs_location(uint64_t ramfs_start, uint64_t ramfs_size);
   void set_elf_symbols(std::map<uint64_t, std::string> symbols);
   void set_term_fd(int fd);
   void set_trace_log(FILE *log);
@@ -181,6 +182,9 @@ private:
   bool get_config_rvfi(unit) override;
   bool get_config_use_abi_names(unit) override;
 
+  mach_bits get_initramfs_base(unit) override;
+  mach_bits get_initramfs_size(unit) override;
+
   bool m_config_print_instr = false;
   bool m_config_print_clint = false;
   bool m_config_print_exception = false;
@@ -197,8 +201,9 @@ private:
 
   // Initialization.
   uint64_t m_elf_entry = 0;
-  std::string m_config_file = {};
-  std::optional<uint64_t> m_htif_tohost_address = {};
+  std::string m_config_file;
+  std::optional<uint64_t> m_htif_tohost_address;
+  std::optional<MemoryRegion> m_initramfs_region;
 
   std::map<uint64_t, std::string> m_symbols;
   int m_term_fd = 1;
--- a/model/postlude/device_tree.sail
+++ b/model/postlude/device_tree.sail
@@ -153,6 +153,33 @@ function generate_soc_clint() -> option(string) = {
   } else None()
 }
 
+val get_initramfs_base = pure {
+    cpp: "get_initramfs_base"
+} : unit -> bits(64)
+
+val get_initramfs_size = pure {
+    cpp: "get_initramfs_size"
+} : unit -> bits(64)
+
+private function have_initrd() -> bool =
+  get_initramfs_size() != zeros()
+
+private function generate_chosen_initrd() -> option((string, string)) = {
+  let size = get_initramfs_size();
+  if size == zeros() then return None();
+
+  let base = get_initramfs_base();
+  let end_ = base + size;
+
+  let base_hi = unsigned(base >> 32);
+  let base_lo = unsigned(base[31 .. 0]);
+
+  let end_hi  = unsigned(end_ >> 32);
+  let end_lo  = unsigned(end_[31 .. 0]);
+  Some("linux,initrd-start = <" ^ hex_str(base_hi) ^ " " ^ hex_str(base_lo) ^ ">",
+       "linux,initrd-end   = <" ^ hex_str(end_hi) ^ " " ^ hex_str(end_lo) ^ ">")
+}
+
 function generate_dts() -> string = {
   let clock_freq : nat = config platform.clock_frequency;
 
@@ -168,7 +195,13 @@ function generate_dts() -> string = {
 ^ "  compatible = \"ucbbar,spike-bare-dev\";\n"
 ^ "  model = \"ucbbar,spike-bare\";\n"
 ^ "  chosen {\n"
-^ "    bootargs = \"console=hvc0 earlycon=sbi\";\n"
+^ "    bootargs = \"" ^ (if have_initrd() then "root=/dev/ram " else "") ^ "console=hvc0 earlycon=sbi\";\n"
+^ (match generate_chosen_initrd() {
+     None()     => "",
+     Some(s, e) =>
+  "    " ^ s ^ ";\n"
+^ "    " ^ e ^ ";\n"
+   })
 ^ "  };\n"
 ^ "  cpus {\n"
 ^ "    #address-cells = <1>;\n"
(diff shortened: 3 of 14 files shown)


os-boot/linux/Makefile
```diff
@@ -70,10 +112,14 @@ distclean:
...
 
 # Run with the Sail emulator from this repo.
-.PHONY: sail
+.PHONY: sail sail_ramfs
 sail: build/sail.dtb $(elf) $(SAIL_SIM)
 	$(SAIL_SIM) --show-times --inst-limit $(INSTRUCTION_LIMIT) --device-tree-blob build/sail.dtb $(elf)
 
+# Run with initramfs
+sail_ramfs: build/sail_ramfs.dtb $(elf) $(ramfs) $(SAIL_SIM)
+	$(SAIL_SIM) --initramfs $(ramfs) --device-tree-blob build/sail_ramfs.dtb $(elf)
```
nadime15, line 121: I think the initramfs address is computed twice? I think the problem is that nothing checks that the two agree. 

If you pass a different --config to only one of them the DTB says one address and the initramfs is written to another and Linux gets a wrong initrd pointer without any error. 

Could we make it a config value instead, something like memory.initramfs_address next to the existing memory.dtb_address? I think this can simplify the rest of the code too.

pmundkur: That would still leave the possibility of inconsistency: the config could have a value of `initramfs_address`  that is incompatible with the load address of the specified kernel image.  Perhaps a better way would be to skip the `--print-device-tree` step, and have `sail_riscv_sim` directly call `dtc` to generate the blob and paste it in, in a single step?  I think that's how Spike does it too.  If `dtc` is not found, we could have a cli option to specify where it is?

nadime15: Yeah collisions are still possible, but the idea was to catch collisions early rather than silently accept it and run a simulation that potentially takes a long time and then crashes.

I think calling dtc makes sense however calling dtc only covers the generated device tree. When someone passes --device-tree-blob with a hand written one, the two addresses can mismatch again. Could we check the initrd addresses for a provided dtb against where the image was actually placed and warn if they disagree?

pmundkur: We could use `libfdt` to parse the blob and check, but there's a trade-off here: keeping the code simple and requiring the user to follow the two steps of the Makefile, or use `libfdt` to add checks.  For now, I'm leaning towards the first.  I can add the two steps to the README.  We could add more checks later using `libfdt` (it's one of the oldest issues, see #2!)

nadime15: Ok yeah sounds good. We should also add some printfs during boot, similar to what we currently do for the `Entry point:` and `HTIF located at`, so the user can see the actual initramfs addresses being used.

pmundkur:
This is a draft.   I'll split this out into smaller PRs.  #1854 was the first.

c_emulator/cli_options.cpp
```diff
@@ -37,6 +37,10 @@ CLIOptions parse_cli(int argc, char **argv) {
   );
 
   app.add_option("--device-tree-blob", opts.dtb_file, "Device tree blob file")
+    ->excludes("--print-device-tree")
```
jordancarlin, line 40: Why does `--device-tree-blob` need to exclude `--print-device-tree`? It is unlikely that they are used together, but I don't think it is technically incompatible. This is alst _technically_ a breaking change if we keep this.

os-boot/Makefile
```diff
@@ -6,17 +6,30 @@ RISCV_COMPILER_PREFIX := riscv64-buildroot-linux-gnu-
...
 payload_xvisor := build/xvisor/build/vmm.bin
 
+# Built binaries.
+linux_elf  := build/fw_payload_linux.elf
+xvisor_elf := build/fw_payload_xvisor.elf
+
+# Initramfs filesystem image.
+ramfs := build/rootfs.cpio
+
 .PHONY: all linux xvisor
```
jordancarlin:
```suggestion
.PHONY: all linux xvisor ramfs
```

os-boot/Makefile
```diff
@@ -6,17 +6,30 @@ RISCV_COMPILER_PREFIX := riscv64-buildroot-linux-gnu-
...
+# Generic rules to download tools.
 
-# Generic rule to download tools.
 downloads/%.tar.xz: %.url
 	mkdir -p downloads
 	curl --location '$(shell cat $<)' --output $@
 
+downloads/%.tar.bz2: %.url
+	mkdir -p downloads
+	curl --location '$(shell cat $<)' --output $@
```
jordancarlin: I believe you can combine these into a single rule and just list multiple targets before the colon.

os-boot/Makefile
```diff
@@ -62,9 +79,31 @@ build/fw_payload_%.elf: build/opensbi/Makefile
...
+# Prepare busybox install
+build/busybox/_install/bin/busybox: build/busybox/.config
+	CROSS_COMPILE=$(CROSS_COMPILE) $(MAKE) -C build/busybox
+	CROSS_COMPILE=$(CROSS_COMPILE) $(MAKE) -C build/busybox install
+
+# Prepare rootfs cpio.
+$(ramfs): init-busybox build/busybox/_install/bin/busybox
+	cp init-busybox build/busybox/_install/init
+	cd build/busybox/_install && find . | cpio --create --format=newc --file ../rootfs.cpio
+	cp build/busybox/rootfs.cpio $(ramfs)
```
jordancarlin, line 96: Any reason not to do this?
```suggestion
	cd build/busybox/_install && find . | cpio --create --format=newc --file $(ramfs)
```

model/postlude/device_tree.sail
```diff
@@ -153,6 +153,33 @@ function generate_soc_clint() -> option(string) = {
...
+val get_initramfs_size = pure {
+    cpp: "get_initramfs_size"
+} : unit -> bits(64)
+
+private function have_initrd() -> bool =
+  get_initramfs_size() != zeros()
+
+private function generate_chosen_initrd() -> option((string, string)) = {
+  let size = get_initramfs_size();
+  if size == zeros() then return None();
```
jordancarlin, line 169: Sort of redundant with `have_initrd`.

c_emulator/riscv_model_impl.h
```diff
@@ -197,8 +201,9 @@ class ModelImpl final : private hart::Model {
 
   // Initialization.
   uint64_t m_elf_entry = 0;
-  std::string m_config_file = {};
-  std::optional<uint64_t> m_htif_tohost_address = {};
+  std::string m_config_file;
+  std::optional<uint64_t> m_htif_tohost_address;
+  std::optional<MemoryRegion> m_initramfs_region;
```
jordancarlin, line 206: Why are the initial values of m_config_file and m_htif_tohost_address dropped?

c_emulator/cli_options.cpp
pmundkur: It's just pointless to load an unused DTB into memory if all you're doing is printing the device tree and exiting.  But I could remove it if you insist (I'd prefer just documenting this change; I don't think this breaking change is that important).

c_emulator/riscv_model_impl.h
pmundkur: They are not needed; C++ default initializes them anyway without them.

model/postlude/device_tree.sail
pmundkur: True, it just seemed simpler to differentiate the two in `generate_dts()`.

os-boot/Makefile
pmundkur: I'll try that...

os-boot/Makefile
pmundkur: No reason; this seems better, thanks!

pmundkur: Actually, I think due to the `cd` earlier `$(ramfs)` won't be where it's supposed to be and the `cpio` will fail?

c_emulator/cli_options.cpp
jordancarlin: Fine with me to leave it, just wanted to at least point it out.

os-boot/Makefile
jordancarlin: How about dropping the `cd` and just changing the path that is passed to `find`?

pmundkur: I don't think that works, `cpio` is creating the root filesystem, and so `find` needs to be run at the root of the prospective filesystem so that `cpio` gets the right paths.

jordancarlin: Good point. I guess we can leave it as is then.

jordancarlin (approved):

nadime15 (approved):
Looks good, we should keep https://github.com/riscv/sail-riscv/pull/1855#discussion_r3692434640 in mind.

pmundkur:
> Looks good, we should keep [#1855 (comment)](https://github.com/riscv/sail-riscv/pull/1855#discussion_r3692434640) in mind.

I added a caveat about this to the README.

Issue #1788, opened 2026-07-01
Support RVVI-TEXT as trace output format


jordancarlin:
There have been discussions about this in a variety of other issues and PRs, but it hasn't had its own issue to track its status.

ACT4 uses RVVI-Text as its input trace format for coverage tracking. Currently, there is a hacky Python script that converts the Sail model trace to RVVI-Text, but it would be ideal for the model to natively support an RVVI-Text output instead. Furthermore, the existing RVVI-Text converter script does not handle virtual memory signals. Rather than adding them to that Python script, we should add them straight to the native RVVI-Text output from the model.

The RVVI-Text spec is available here: https://github.com/riscv-verification/RVVI/blob/master/RVVI-TEXT/README.md.
A Python script that validates an RVVI-Text file is available here: https://github.com/riscv-verification/RVVI/blob/master/source/host/rvvi/rvviTextChecker.py.

PR #1857, opened 2026-08-01
add support of RVVI_TEXT output

KotorinMinami:
for #1788

### DIFF
--- a/c_emulator/riscv_callbacks_rvvi_text.cpp
+++ b/c_emulator/riscv_callbacks_rvvi_text.cpp
@@ -1,0 +1,308 @@
+#include "riscv_callbacks_rvvi_text.h"
+
+#include "riscv_model_impl.h"
+
+#include <cinttypes>
+#include <gmp.h>
+#include <inttypes.h>
+
+rvvi_text_callbacks::rvvi_text_callbacks(FILE *trace_log, uint64_t xlen, uint64_t flen, uint64_t vlen) :
+    m_trace_log(trace_log),
+    m_xlen(xlen),
+    m_flen(flen),
+    m_vlen(vlen) {
+  emit_header();
+}
+
+std::string rvvi_text_callbacks::hex_value(const sbits &value) {
+  const unsigned nibbles = static_cast<unsigned>((value.len + 3) / 4);
+  char buf[32];
+  snprintf(buf, sizeof(buf), "%0*" PRIX64, nibbles, value.bits);
+  return std::string(buf);
+}
+
+std::string rvvi_text_callbacks::hex_value_lbits(const lbits &value) {
+  const unsigned nibbles = static_cast<unsigned>((value.len + 3) / 4);
+  char *raw = mpz_get_str(nullptr, 16, *value.bits);
+  std::string s = raw ? raw : "";
+  free(raw);
+  if (s.length() < nibbles) {
+    s = std::string(nibbles - s.length(), '0') + s;
+  }
+  for (char &c : s) {
+    if (c >= 'a' && c <= 'f') {
+      c = c - 'a' + 'A';
+    }
+  }
+  return s;
+}
+
+std::string rvvi_text_callbacks::page_type_letter(int64_t level) {
+  switch (level) {
+  case 0:
+    return "K";
+  case 1:
+    return "M";
+  case 2:
+    return "G";
+  case 3:
+    return "T";
+  case 4:
+    return "P";
+  default:
+    return "K";
+  }
+}
+
+void rvvi_text_callbacks::emit_header() {
+  if (m_trace_log == nullptr || m_header_emitted) {
+    return;
+  }
+  m_header_emitted = true;
+
+  // RVVI-TEXT spec version 0.5
+  fprintf(m_trace_log, "VERSION 0 5\n");
+  // Vendor identification
+  fprintf(m_trace_log, "VENDOR \"sail_riscv\" 0 1\n");
+
+  // PARAMS: ILEN is always 32 for RISC-V. XLEN/FLEN/VLEN from model.
+  // NHART=1 and RETIRE=1 for the single-hart in-order emulator.
+  int param_count = 6; // ILEN, XLEN, FLEN, VLEN, NHART, RETIRE
+  fprintf(
+    m_trace_log,
+    "PARAMS %d ILEN 32 XLEN %llu FLEN %llu VLEN %llu NHART 1 RETIRE 1\n",
+    param_count,
+    static_cast<unsigned long long>(m_xlen),
+    static_cast<unsigned long long>(m_flen),
+    static_cast<unsigned long long>(m_vlen)
+  );
+}
+
+void rvvi_text_callbacks::reset_instruction_buffer() {
+  m_pending_pc = 0;
+  m_pending_inst = 0;
+  m_have_inst = false;
+  m_pending_trap = false;
+  m_reg_changes.clear();
+  m_mem_accesses.clear();
+  m_ptw_success = false;
+  m_last_pte = 0;
+  m_last_pte_ppn = 0;
+  m_ptw_success_level = -1;
+}
+
+void rvvi_text_callbacks::fetch_callback(ModelImpl &model, sbits opcode) {
+  // fetch fires before tick_pc(); capture PC+opcode now.
+  // A second fetch with no intervening instret means the previous one
+  // trapped or was abandoned.
+  if (m_have_inst) {
+    reset_instruction_buffer();
+  }
+  m_pending_pc = model.pc();
+  m_pending_inst = opcode.bits;
+  m_have_inst = true;
+}
+
+void rvvi_text_callbacks::vmem_access_callback(
+  ModelImpl &model,
+  sbits vaddr,
+  sbits paddr,
+  bool is_fetch,
+  bool is_write
+) {
+  // Only real accesses (post-translateAddr) fire here, so PTW reads are
+  // excluded. Attach the most recent PTW outcome so the trace can emit
+  // PTE/PT for this access.
+  (void)model;
+  MemAccess access;
+  access.is_fetch = is_fetch;
+  access.is_write = is_write;
+  access.vaddr = vaddr.bits;
+  access.paddr = paddr.bits;
+  if (m_ptw_success) {
+    access.has_pte = true;
+    access.pte = m_last_pte;
+    access.ppn = m_last_pte_ppn;
+    access.page_level = m_ptw_success_level;
+  }
+  m_mem_accesses.push_back(access);
+}
+
+void rvvi_text_callbacks::xreg_full_write_callback(
+  ModelImpl &model,
+  const_sail_string abi_name,
+  sbits reg,
+  sbits value
+) {
+  (void)model;
+  (void)abi_name;
+  RegChange change;
+  change.kind = 'X';
+  change.index = reg.bits;
+  change.value = hex_value(value);
+  m_reg_changes.push_back(std::move(change));
+}
+
+void rvvi_text_callbacks::freg_write_callback(ModelImpl &model, unsigned reg, sbits value) {
+  (void)model;
+  if (m_flen == 0) {
+    return;
+  }
+  RegChange change;
+  change.kind = 'F';
+  change.index = reg;
+  change.value = hex_value(value);
+  m_reg_changes.push_back(std::move(change));
+}
+
+void rvvi_text_callbacks::csr_full_write_callback(
+  ModelImpl &model,
+  const_sail_string csr_name,
+  unsigned reg,
+  sbits value
+) {
+  (void)model;
+  (void)csr_name;
+  RegChange change;
+  change.kind = 'C';
+  change.index = reg;
+  change.value = hex_value(value);
+  m_reg_changes.push_back(std::move(change));
+}
+
+void rvvi_text_callbacks::vreg_write_callback(ModelImpl &model, unsigned reg, lbits value) {
+  (void)model;
+  if (m_vlen == 0) {
+    return;
+  }
+  RegChange change;
+  change.kind = 'V';
+  change.index = reg;
+  change.value = hex_value_lbits(value);
+  m_reg_changes.push_back(std::move(change));
+}
+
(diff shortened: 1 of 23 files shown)


jordancarlin:
@KotorinMinami Thanks for working on this!

Before reviewing, I decided to check this branch out to see what the output looks like, and it doesn't seem to follow the RVVI-TEXT specification linked in the issue (https://github.com/riscv-verification/RVVI/blob/master/RVVI-TEXT/README.md). It uses different key names (e.g. INSN and PC instead of RET), is missing the 0x prefixes before hex numbers, is missing the initial VERSION, VENDOR, and PARAMS lines, etc. Can you update this to follow the actual RVVI-TEXT specification? What is this version based on?

KotorinMinami:
Okay, I will start making the modifications. I developed based on the links and scripts you provided in the issue, but I may have misunderstood some of the content

jordancarlin:
@KotorinMinami any update on this? If not, I am probably going to start working on this myself pretty soon.

KotorinMinami:
Wait a moment, I need to do some verification.

test/first_party/src/common/sail_to_rvvi_spec.py
jordancarlin, line 1: This testing framework seems overly complicated. I think we can simplify to just running the output of a test through the `rvviTextChecker.py` script. If you want to do something more extensive than that, how about checking in one "golden" output from a test and diffing the output against that?

jordancarlin (changes requested):
This is getting closer, but there still seem to be significant issues in the MEM objects. I haven't looked through the implementation yet, but looking at the output I noticed the following issues:
- All MEM records claim to be 4 byte accesses, even for larger accesses like `ld`
- Only loads are reported, stores are missing
- All instruction fetch memory accesses are missing
- Multiple MEM elements are not emitted for instructions that cause multiple memory accesses (for example, misaligned accesses that are broken apart)
- Part of the page table entry information is missing (the page type is present but the rest of the fields are missing)

There might be other issues, but these are the ones that jumped out at me immediately. @KotorinMinami how have you been validating the output of this implementation?

jordancarlin:
@KotorinMinami any chance of getting the issues I described above fixed in time for the 0.14.1 release?

KotorinMinami:
@jordancarlin Sorry for the delay. I've updated the branch to follow RVVI-TEXT 0.5.

 The first version was based on the ACT sail_to_rvvi.py dialect (PC/INSN, no 0x), which is why it didn't match the spec you linked. This version uses
 RET/TRAP, 0x prefixes, and VERSION/VENDOR/PARAMS.

 On the MEM issues:

 - Width now comes from the model (lb/lh/lw/ld and stores/AMOs), not hardcoded 4.
 - Stores are reported.
 - Fetches are MEM I (2-byte compressed fetches and 4-byte fetches when the model does Ziccif).
 - Split accesses: a page-straddling load emits two MEM D records. Sail does not otherwise break a single completed translation into pieces, so one MEM
   per access is what the model actually does.
 - PTE records now have both the PTE bits and PT (K/M/G/T/P). GPTE/GPT/GPADDR are omitted for now (no two-stage / H yet).

 Validation is rvviTextChecker.py, a small semantic linter, and golden diffs of first-party tests (widths, load/store, AMO/SC, Sv39 PTE, fetch, trap,
 vector). I dropped the extra Python spec harness as you suggested.

 Two remaining quirks, in case they may matter for ACT:

 - An AMO currently emits two identical MEM D records (the model's load then the store). I can coalesce those if you want.
 - Vector/FP ops can repeat the same V/C register on one line. Last-write-wins at sample time matches the ACT testbench.

jordancarlin:
@pmundkur @nadime15 the general output of the tracer is looking a lot better, so we should be good to start actually reviewing the implementation now.

pmundkur:
It would be good to rebase to fix the conflicts...

model/core/callbacks.sail
```diff
@@ -22,6 +22,11 @@ function mem_read_callback(_) = ()
 val mem_exception_callback = pure {cpp: "mem_exception_callback"} : (/* addr */ physaddrbits, /* exception code */ exc_code) -> unit
 function mem_exception_callback(_) = ()
 
+/// Fires after a successful v2p translation for a load/store/fetch and
+/// before the corresponding mem_read_callback / mem_write_callback.
+val vmem_access_callback = pure {cpp: "vmem_access_callback"} : forall 'n, 0 < 'n <= max_mem_access . (/* vaddr */ xlenbits, /* paddr */ physaddrbits, /* is_fetch */ bool, /* is_write */ bool, /* width */ int('n)) -> unit
```
pmundkur, line 27: It would be nicer to re-use `access` and then define helpers (either in C++ or Sail) to compute `is_write` and `is_fetch` for the actual trace.
```suggestion
val vmem_access_callback = pure {cpp: "vmem_access_callback"} : forall 'n, 0 < 'n <= max_mem_access . (/* vaddr */ xlenbits, /* paddr */ physaddrbits, /* access */ MemoryAccessType(mem_payload), /* width */ int('n)) -> unit
```

c_emulator/riscv_callbacks_rvvi_text.cpp
```diff
@@ -0,0 +1,264 @@
...
+  sbits value
+) {
+  (void)model;
+  (void)csr_name;
+  record_reg('C', reg, hex_value(value));
+}
+
+void rvvi_text_callbacks::vreg_write_callback(ModelImpl &model, unsigned reg, lbits value) {
+  (void)model;
+  if (m_vlen == 0) {
```
pmundkur, line 150: I don't think this is the right check.  You could have `extensions.V.support_level` set to `Disabled` and still have a non-zero `extensions.V.vlen_exp`.  You could do something like this in `ModelImpl::init_platform_constants()`
```
 std::string vector_support = get_config_string({"extensions", "V", "support_level"});
 m_has_vector_registers = vector_support != "Disabled";
 m_has_float_registers = get_config_bool({"extensions", "F", "supported"});
```
and use `m_has_vector_registers` and `m_has_float_registers` instead to check whether vector and float registers exist.

c_emulator/riscv_callbacks_rvvi_text.cpp
```diff
@@ -0,0 +1,264 @@
...
+  sbits value
+) {
+  (void)model;
+  (void)abi_name;
+  record_reg('X', reg.bits, hex_value(value));
+}
+
+void rvvi_text_callbacks::freg_write_callback(ModelImpl &model, unsigned reg, sbits value) {
+  (void)model;
+  if (m_flen == 0) {
```
pmundkur, line 131: I don't think is the right check.  See below for the same problem with vector registers.

c_emulator/riscv_callbacks_rvvi_text.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+#include <cctype>
+#include <gmp.h>
+#include <inttypes.h>
+
+rvvi_text_callbacks::rvvi_text_callbacks(FILE *trace_log, uint64_t xlen, uint64_t flen, uint64_t vlen) :
+    m_trace_log(trace_log),
+    m_xlen(xlen),
+    m_flen(flen),
+    m_vlen(vlen) {
```
pmundkur, line 13: Instead of passing these lengths in the constructor, pass bools indicatingwhether the register files exist, and then get the `flen` and `vlen` from `model` only if they do.

c_emulator/riscv_sim.cpp
```diff
@@ -716,6 +717,19 @@ InitResult init_model(
...
 
+  if (opts.config_print_rvvi_text) {
+    // Fall back to stdout if no trace-output file was specified.
+    FILE *rvvi_log = run_info.trace_log != stdout ? run_info.trace_log : stdout;
+    model.register_callback(
+      std::make_shared<rvvi_text_callbacks>(
+        rvvi_log,
+        static_cast<uint64_t>(model.xlen()),
+        static_cast<uint64_t>(model.flen()),
+        static_cast<uint64_t>(model.vlen())
```
pmundkur, line 728: See comment above about changing this constructor to use bools indicating whether floating point and vector registers exist.

pmundkur (changes requested):
Some initial comments.

c_emulator/riscv_callbacks_rvvi_text.cpp
pmundkur: It might be simpler if this is  rebased on something like #1957.

pmundkur:
@KotorinMinami you'll need to rebase now that #1957 is merged to reuse the APIs from there instead of adding them in this PR.

pmundkur:
@KotorinMinami you'll need to sign your commits to pass the DCO check.

pmundkur:
@KotorinMinami it looks like the first party tests you added for this are not passing.  There are some significant differences with the golden files the tests are comparing with.

KotorinMinami:
@pmundkur Probably a compiler/toolchain difference: the old tests diffed the whole trace byte-for-byte and
   those files carried absolute PCs with crt0 included, so a different clang/gcc or link layout shifts
   every line starting from the first one.

   This revision compares per-instruction fields inside the `main` window instead. I re-checked it from
   a clean pull with a fresh build and it passes now (gcc and clang). Could you take another look?

pmundkur:
@KotorinMinami thanks, they're passing locally now.

c_emulator/riscv_callbacks_rvvi_text.cpp
```diff
@@ -0,0 +1,271 @@
...
+  }
+}
+
+void rvvi_text_callbacks::ptw_step_callback(ModelImpl &model, int64_t level, sbits pte_addr, uint64_t pte) {
+  (void)model;
+  (void)pte_addr;
+  (void)level;
+  m_last_pte = pte;
+}
+
```
nadime15, line 151: I am not really familiar with RVVI_TEXT, but I guess that since we added Hypervisor support, you need to take this into account somehow now? The ptw_step_callback probably also needs to know the current stage (VS-Stage, G-Stage, S-Stage, etc.)? Maybe it would be easier to do that in a follow up PR? I am not sure how difficult it is to make this PR Hypervisor aware.


I found this table also, and I think GPTE, GPT and GPADDR are not factored in yet:

| Key    | Value | Meaning                                               |
|--------|-------|-------------------------------------------------------|
| PTE    | hex   | Page table entry (VS-stage when Hypervisor is active) |
| GPTE   | hex   | Page table entry (G-stage when Hypervisor is active)  |
| PT     | string | Page type (VS-stage when Hypervisor is active)       |
| GPT    | string | Page type (G-stage when Hypervisor is inactive)      |
| GPADDR | hex   | Guest physical address (when Hypervisor is active)    |

pmundkur (commented):
@KotorinMinami I've simplified this a bit 
https://github.com/pmundkur/sail-riscv/commit/142ac374e82a47e40c977284fdfd2bac91389a9a
and fixed some `clang-tidy` issues
https://github.com/pmundkur/sail-riscv/commit/573c4ae292dcde46021ef999cb6f4b10c992bbd0
If they are okay with you, I could push them to your branch.
This is just fixing code and compilation issues, I haven't reviewed the trace format in detail yet.

c_emulator/riscv_model_impl.cpp
```diff
@@ -398,6 +405,11 @@ void ModelImpl::init_platform_constants() {
   m_supports_hypervisor = get_config_bool({"extensions", "H", "supported"});
   m_supports_D_extension = get_config_bool({"extensions", "D", "supported"});
   m_has_float_registers = get_config_bool({"extensions", "F", "supported"});
+  std::string vector_support = get_config_string({"extensions", "V", "support_level"});
+  if (vector_support != "Disabled") {
+    uint64_t vlen_exp = get_config_uint64({"extensions", "V", "vlen_exp"});
+    m_vlen = 0x1U << vlen_exp;
+  }
```
nadime15: ModelImpl already has vlen() returning zvlen and has_vector_registers() reading zvector_support_level, and emit_header already calls model.vlen(). So m_vlen is written here and never read, and m_has_vector_registers is never assigned at all. Can you drop this block as well as `m_vlen` and `m_has_vector_registers`

```suggestion

```

c_emulator/riscv_sim.cpp
```diff
@@ -802,6 +803,19 @@ InitResult init_model(
...
+    FILE *rvvi_log = run_info.trace_log != stdout ? run_info.trace_log : stdout;
+    model.register_callback(
+      std::make_shared<rvvi_text_callbacks>(
+        rvvi_log,
+        static_cast<uint64_t>(model.xlen()),
+        model.has_float_registers(),
+        model.has_vector_registers()
+      )
+    );
+  }
```
nadime15, line 823: RVVI-TEXT currently goes to the same trace_log as the other --trace-* options, so using --trace-rvvi-text and --trace-mem mixes log lines and causes the checker to fail.

Does it make sense to give it its own output, something like `--rvvi-text-output <file>`?

nadime15 (commented):
I’ll keep reviewing it. I’m gonna leave comments from time to time.

c_emulator/riscv_model_impl.h
```diff
@@ -112,6 +112,9 @@ class ModelImpl final : private hart::Model {
   uint64_t pc() const;
   uint64_t fcsr() const;
 
+  uint64_t cur_privilege_mode() const;
```
pmundkur, line 115: This is fine for now; it's the same issue with the register accessors like `xreg()` that are not const for the same reason.
```suggestion
  uint64_t cur_privilege_mode();
```

c_emulator/riscv_model_impl.cpp
```diff
@@ -563,6 +570,16 @@ int64_t ModelImpl::physaddrbits_len() const {
   return zphysaddrbits_len;
 }
 
+uint64_t ModelImpl::cur_privilege_mode() const {
+  // Generated accessor is non-const but only reads state.
+  auto *mut_this = const_cast<ModelImpl *>(this);
+  return mut_this->zprivLevel_to_bits(zcur_privilege);
+}
```
pmundkur: We can fix the const issues later more generally.

```suggestion
uint64_t ModelImpl::cur_privilege_mode() {
  return zprivLevel_to_bits(zcur_privilege);
}
```

PR #1769, opened 2026-06-19
Add support for the Hypervisor (H) extension

nadime15:
This PR adds support for the Hypervisor (H) extension to the Sail RISC-V Model. It builds on the prior hypervisor work in #612, #1419 and #1624, and incorporates parts of #1110, #1545 and #1513.

It implements all mandatory parts of the extension:

* Two-stage address translation (VS-stage + G-stage)
* Trap entry/exit and two-stage delegation
* Hypervisor instructions (`HLV`/`HLVX`/`HSV`, `HFENCE.VVMA`/`HFENCE.GVMA`)
* Hypervisor and VS CSRs
* Virtual-instruction exceptions

It also covers the H interaction with F, V, Zicbom, Zicboz, Zicntr, Pointer Masking, Smstateen, Ssstateen, Svadu, Sstc, Svpbmt, Svinval, Sscofpmf, Zicfilp and Zicfiss.

The following test suites were used for testing:

* riscv-tests — https://github.com/riscv-software-src/riscv-tests
* riscv-hext-xvisor-linux — https://github.com/nadime15/riscv-hext-xvisor-linux (Xvisor + Linux guest boot)
* xvisor — https://github.com/xvisor/xvisor (Xvisor boot)
* riscv-hypervisor-tests — https://github.com/nadime15/riscv-hypervisor-tests
* riescue — https://github.com/nadime15/riescue
* damo-rv-priv-ats - https://github.com/XUANTIE-RV/damo-rv-priv-ats

PR #1860, opened 2026-08-06
Model the PTE A/D-bit update as an atomic read-check-write

zeldovich:
The privileged spec requires the hardware update of the PTE A/D bits to be atomic:

  "The PTE update must be atomic with respect to other accesses to the
   PTE, and must atomically perform all tablewalk checks for that leaf
   PTE as part of, and before, conditionally updating the PTE value."

and step 7 of the virtual address translation process specifies the update as a compare-and-swap:

  "Compare pte to the value of the PTE at address a+va.vpn[i]*PTESIZE.
   If the values match, set pte.a to 1 and, if the original memory
   access is a store, also set pte.d to 1.  If the comparison fails,
   return to step 2."

update_and_write_pte() previously computed the new A/D bits on the value the walk had read (or on the copy cached in the TLB) and wrote the result back with a plain write.  This can overwrite concurrent updates to the PTE: if another hart updates the PTE (e.g., unmapping a page), the A/D write-back will resurrect the PTE's old value.

This change models the update as the atomic read-check-write that the spec requires: re-read the PTE with an exclusive read, re-perform all tablewalk checks for the leaf PTE on the fresh value (a failure produces the same page fault that "return to step 2" would), and conditionally write back the fresh value with A/D set.

Sequential execution is unchanged: the re-read returns the value the walk just read, the re-checks pass exactly as they did during the walk, and the written value is identical.  The change is visible only in the memory-interface events: the update is now an exclusive read/conditional write pair, with the written value derived from the fresh read, mirroring how the AMO semantics present their accesses.

The one deliberate sequentially-visible difference is the TLB-hit path: if the in-memory PTE was modified after the TLB entry was filled (without an SFENCE.VMA), an A/D update now re-checks the in-memory PTE and raises a page fault instead of writing the stale TLB copy of the PTE back over the modified PTE, matching the privileged spec's compare-and-swap.

This issue came up during a project on verification of a concurrent OS kernel on top of the Sail RISC-V model, using generative AI tools, and this change was co-authored in collaboration with generative AI tools (Claude code).

### DIFF
--- a/model/sys/mem.sail
+++ b/model/sys/mem.sail
@@ -100,10 +100,10 @@ private function pmaCheck forall 'n, 0 < 'n <= max_mem_access .
     InstructionFetch()     => attributes.executable,
     Load(Data)             => {assert(not(res_or_con)); attributes.readable},
     Load(Vector)           => {assert(not(res_or_con)); attributes.readable},
-    Load(PageTableEntry)   => {assert(not(res_or_con)); attributes.supports_pte_read},
+    Load(PageTableEntry)   => attributes.supports_pte_read,
     Store(Data)            => {assert(not(res_or_con)); attributes.writable},
     Store(Vector)          => {assert(not(res_or_con)); attributes.writable},
-    Store(PageTableEntry)  => {assert(not(res_or_con)); attributes.supports_pte_write},
+    Store(PageTableEntry)  => attributes.supports_pte_write,
 
     // For LR/SC, we currently don't differentiate between the presence (RsrvEventual)
     // or absence (RsrvNonEventual) of an eventual success guarantee.
--- a/model/sys/vmem.sail
+++ b/model/sys/vmem.sail
@@ -48,6 +48,15 @@ private function write_pte forall 'n, 'n in {4, 8} . (
 ) -> MemoryOpResult(bool) =
   mem_write_value_priv(paddr, pte_size, pte, Supervisor, Store(PageTableEntry), PBMT_PMA, false, false, false)
 
+// Write a Page Table Entry as the write half of an atomic read-modify-write
+// of that PTE.
+private function write_pte_conditional forall 'n, 'n in {4, 8} . (
+  paddr    : physaddr,
+  pte_size : int('n),
+  pte      : bits('n * 8),
+) -> MemoryOpResult(bool) =
+  mem_write_value_priv(paddr, pte_size, pte, Supervisor, Store(PageTableEntry), PBMT_PMA, false, false, true)
+
 // Sscptr: Main memory supports page table read
 function clause currentlyEnabled(Ext_Ssccptr) = hartSupports(Ext_Ssccptr) & (currentlyEnabled(Ext_Sv32) | currentlyEnabled(Ext_Sv39))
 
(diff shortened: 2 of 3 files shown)


nadime15:
Hi @zeldovich, thanks for finding this!

A few observations: Step 9 says to compare the whole PTE against memory and write only if it matches. You rerun the tablewalk checks instead, but a different PTE can still pass all of them, and then the model returns the old ppn/pbmt while writing the new PTE. Additionally on a stale TLB entry you page fault, but the spec says walk again, and that walk can succeed. Lastly comparing the value is simpler too, since you dont need level/vpn/permissions at the update site.

You can find my alternative implementation here (which I have not explicitly tested but it compiles and passes CI):

https://github.com/riscv/sail-riscv/compare/master...nadime15:sail-riscv:fix-pte-ad-atomic-cas

But this PR has to wait until #1769 is merged (the version above may also be easier to integrate with #1769).

zeldovich:
Your version looks quite a bit cleaner, and you're right that we should compare equality rather than redo the check.  Thanks for catching that!

One possible issue: I think your version might not correctly handle a race during the re-validation.  You throw an internal error ("stale PTE after page table walk"), but I think the conditional write can legitimately fail, in a weak-memory setting where another hart modified the PTE between the read_reserved and the write_conditional.  The spec says to keep looping until the validation fails or the writeback succeeds.  (The comment about "only reachable via a TLB hit" is also incomplete: I think it's also reachable due to a race.)

As a caveat, I should say that I'm just starting to explore weak memory with the Sail RISC-V model, so maybe I don't fully understand how the Sail spec for a single hart is intended to link to a weak-memory model here.

As a minor nit, the PMA check now requires reservability for page tables; is there some weird platform that doesn't satisfy this (where previously this would have been fine)?  Maybe especially if they only support Svade and not Svadu.

pmundkur:
It would be interesting to capture the keep-looping criterion for the formal backends, since they require a termination measure.  This is getting into system-level concurrency models that I know @PeterSewell et al have been thinking of looking at for RISC-V.

nadime15:
>  I think your version might not correctly handle a race during the re-validation. You throw an internal error ("stale PTE after page table walk"), but I think the conditional write can legitimately fail, in a weak-memory setting where another hart modified the PTE between the read_reserved and the write_conditional. The spec says to keep looping until the validation fails or the writeback succeeds. 

Yeah, the PTE_AD_Stale() case in translate_TLB_miss() is an internal_error because I wasn't sure how we want to model it. You are right that the spec says to loop until the validation fails or the writeback succeeds, but I would rather we agree on whether to implement that before I do. The abort at least makes the case noticeable rather than silently wrong.
 
(Btw. the TLB hit path already rewalks, it returns None() and falls through to translate_TLB_miss. Its only a stale detected inside translate_TLB_miss that aborts).

> (The comment about "only reachable via a TLB hit" is also incomplete: I think it's also reachable due to a race.) 

Good point, I updated the comment.

> As a minor nit, the PMA check now requires reservability for page tables; is there some weird platform that doesn't satisfy this (where previously this would have been fine)? Maybe especially if they only support Svade and not Svadu.

Yes I checked, I believe it does not change anything in practice. In the default config the RsrvNone regions already have supports_pte_write: false, so PTE writes there are refused anyway. It would only matter for a config with supports_pte_write: true and reservability: RsrvNone, and that combination can not support hardware A/D updates anyway, since the spec requires them to be atomic.

Svade only is actually the one case that can not hit it at all. update_and_write_pte returns PTW_PTE_Needs_Update() before any memory access, so it never does a reserved read or conditional write.

However I did find something else. With neither Svadu nor Svade the model still updates A/D, but validate_config skipped the check because it only looked at Svadu.supported. I tightened it to Svadu.supported | not(Svade.supported).

zeldovich:
It looks like the H-extension PR is now merged.  Would it be useful for me to port my change to the latest version of the Sail model?  Nadime's version looks cleaner, FWIW.

pmundkur:
I'd rather have this done as part of a more formal model of system level concurrency, as was done with the user-level weak memory model with axiomatic and operational models.  It might be that the spec might need to be changed once those are developed.

zeldovich:
Ok, that seems like a reasonable plan -- happy to integrate with that when it lands, and I'll keep using my private fork of the Sail model for the time being.  Thank you!

PR #1861, opened 2026-08-06
cmake: allow replacing a formal backend's config per arch

dtumad:
Rewritten (2026-08-19) down to the one thing that was actually missing, after #1879.

A project that consumes the formal output may need a given architecture generated from its own
configuration rather than the generated default, while keeping the target and output names it
already depends on. `SAIL_FORMAL_CONFIG_<ARCH>` replaces `config_file` for that architecture only:

```
cmake -B build -DSAIL_FORMAL_CONFIG_RV64D=/abs/path/to/config.json
cmake --build build --target generated_lean_rv64d
```

The `foreach (xlen …)` loop already funnels every formal backend through a single `config_file`, so
this covers SMT, rmem, Rocq, Lean and Lem uniformly. It adds no targets, renames nothing, and is
inert unless set — configuring without it produces byte-identical build rules.

Verified locally: without the flag the generated rules are unchanged; with it, the `generated_lean_rv64d`
rule becomes `sail … --config /abs/path/to/config.json --lean … -o Lean_RV64D`.

This is orthogonal to #1879's refactor and applies unchanged on top of it — the override sits in the
default loop, not in the custom family. Happy to fold it into #1879 instead if that is preferred.

The `${config_file}` `DEPENDS` additions that used to be in this PR are now #1885.

### DIFF
--- a/model/CMakeLists.txt
+++ b/model/CMakeLists.txt
@@ -233,6 +233,17 @@ foreach (xlen IN ITEMS 32 64)
     # for backwards compatibility.
     set(arch "rv${xlen}d")
 
+    # Allow a project that consumes the formal output to generate this architecture from
+    # its own configuration instead of the default one, keeping every target and output
+    # name unchanged. Set to an absolute path.
+    string(TOUPPER ${arch} arch_uc)
+    set(SAIL_FORMAL_CONFIG_${arch_uc} "" CACHE FILEPATH
+        "Configuration JSON the ${arch} formal backends are generated from instead of the default.")
+    if (SAIL_FORMAL_CONFIG_${arch_uc})
+        set(config_file "${SAIL_FORMAL_CONFIG_${arch_uc}}")
+        message(STATUS "Formal backends: ${arch} uses ${config_file}")
+    endif()
+
     ############# SMT #############
 
     # Generate smtlib2 code from the Sail code.


pmundkur (commented):
Looks reasonable to me; deferring to @bacam @ineol @Alasdair

model/CMakeLists.txt
```diff
@@ -225,10 +225,21 @@ add_dependencies(generated_sail_riscv_docs generated_html_tgz)
...
 # Theorem prover backends require compile time configuration.
 foreach (xlen IN ITEMS 32 64)
-    # Use ELEN=XLEN.
-    set(config_file "${CMAKE_BINARY_DIR}/config/rv${xlen}d_v256_e${xlen}.json")
+    if (SAIL_FORMAL_CONFIG)
+        set(config_file "${SAIL_FORMAL_CONFIG}")
+    else()
+        # Use ELEN=XLEN.
+        set(config_file "${CMAKE_BINARY_DIR}/config/rv${xlen}d_v256_e${xlen}.json")
+    endif()
```
jordancarlin: We should probably not even loop over `xlen` when `SAIL_FORMAL_CONFIG` is set.

dtumad: I've tried to make this change with some help from Claude. It seems reasonable semantically and works in my testing for `sp1-lean` but I am not super familiar with `cmake` in general so I'd appreciate any feedback.

.github/workflows/compile-lean.yml
```diff
@@ -34,6 +34,15 @@ jobs:
...
           cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -G Ninja
+          sed '/^[[:space:]]*\/\//d' build/config/rv64d_v256_e64.json > build/formal_rv64d.json
+          cmake -S . -B build \
+            -DSAIL_FORMAL_CONFIG="${{ github.workspace }}/build/formal_rv64d.json"
+          ninja -C build -t targets all > build/targets.txt
+          grep -q '^generated_lean_rv64d:' build/targets.txt
+          if grep -q '^generated_lean_rv32d:' build/targets.txt; then
+            echo "RV32 formal targets must not be generated for an RV64 custom config"
+            exit 1
+          fi
```
bacam: I don't think this extra test is worthwhile, especially as it looks like it would interfere with the normal build.

model/CMakeLists.txt
```diff
@@ -606,4 +610,21 @@ foreach (xlen IN ITEMS 32 64)
...
+# so only instantiate the corresponding target family instead of applying it to both architectures.
+if (SAIL_FORMAL_CONFIG)
+    # Use XLEN to retain the existing target names while only generating the requested model.
+    # The config is otherwise validated by Sail when the target is built.
+    file(READ "${SAIL_FORMAL_CONFIG}" formal_config_contents)
+    string(JSON formal_config_xlen GET "${formal_config_contents}" base xlen)
+
+    set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${SAIL_FORMAL_CONFIG}")
+    add_formal_targets("${formal_config_xlen}" "${SAIL_FORMAL_CONFIG}")
+else()
```
bacam: This is quite a bit of work to figure out xlen, and I think it's only used to pick the output filename.  I suggest using a different filename for custom configurations and only include xlen if you're really keen.

test/CMakeLists.txt
```diff
@@ -210,6 +210,12 @@ endif()
 # Download and add Lean emulator tests.
 option(ENABLE_LEAN_EMULATOR_TESTS "Enable Lean emulator tests" OFF)
 
+if(ENABLE_LEAN_EMULATOR_TESTS AND NOT TARGET build_lean_emulator)
+    message(FATAL_ERROR
+        "Lean emulator tests require RV64 formal targets; the selected SAIL_FORMAL_CONFIG is RV32."
+    )
+endif()
+
```
bacam: The message doesn't seem to match the if guard, and appears to be a different issue anyway (I suppose Claude found it during testing?).

CMakeLists.txt
```diff
@@ -215,7 +215,9 @@ add_subdirectory("model")
 add_subdirectory("c_emulator")
 
 # Lean emulator
-add_subdirectory("lean_emulator")
+if (TARGET generated_lean_executable_rv64d)
+    add_subdirectory("lean_emulator")
+endif()
```
bacam: Similarly, I'm not sure about this, but it looks unrelated to the issue at hand anyway.

arichardson:
I haven't had a chance to review this but I was wondering if we could also add support for building a specialized simulator that only supports one config but therefore runs a lot faster?

dtumad:
I've tried to scale back the changes a good amount, just modifying directly what is looped over, and pulling the variable setting outside the loop. This does mean a custom run needs to set two variables but I don't think that's a major issue(?)

jordancarlin:
> I haven't had a chance to review this but I was wondering if we could also add support for building a specialized simulator that only supports one config but therefore runs a lot faster?

I'd be curious how much faster it actually runs if we do that.

model/CMakeLists.txt
```diff
@@ -225,13 +225,29 @@ add_dependencies(generated_sail_riscv_docs generated_html_tgz)
...
-    # for backwards compatibility.
-    set(arch "rv${xlen}d")
+# Theorem prover backends require compile time configuration. By default the targets are
+# generated for rv32d and rv64d against the generated default configs. Setting
+# SAIL_FORMAL_CONFIG generates a single family of targets from the given config instead,
+# named after SAIL_FORMAL_CONFIG_NAME (e.g. `generated_lean_custom`).
+set(SAIL_FORMAL_CONFIG "" CACHE FILEPATH
+    "Configuration JSON the formal backends are generated from instead of the generated default configs.")
+set(SAIL_FORMAL_CONFIG_NAME "custom" CACHE STRING
+    "Name used for the formal targets generated from SAIL_FORMAL_CONFIG, e.g. generated_lean_<name>.")
```
jordancarlin: Could we simplify this to just use the filename with the extension removed instead of adding an additional option?

model/CMakeLists.txt
```diff
@@ -225,13 +225,29 @@ add_dependencies(generated_sail_riscv_docs generated_html_tgz)
...
+    # A custom config generates a single family of formal targets instead of one per XLEN.
+    set(formal_archs "${SAIL_FORMAL_CONFIG_NAME}")
+    set(formal_configs "${SAIL_FORMAL_CONFIG}")
+else()
+    # The arch names are used in a lot of places so they exclude the `_v256_e{xlen}`
+    # suffix for backwards compatibility. The default configs use ELEN=XLEN.
+    set(formal_archs rv32d rv64d)
+    set(formal_configs
+        "${CMAKE_BINARY_DIR}/config/rv32d_v256_e32.json"
+        "${CMAKE_BINARY_DIR}/config/rv64d_v256_e64.json")
```
jordancarlin, line 247: I wonder if we should just switch to using the full name with VLEN and ELEN in it at this point.

model/CMakeLists.txt
dtumad: For my use case at least it's not a problem, I just want to generate a single version with an SP1 compatible config, don't really care much about the particular name. I guess just a question of if others care about being able to separate them.

jordancarlin: I'd be in favor of using the filename for now since it simplifies things and we can add the extra option later if people request it.

pmundkur:
@dtumad would #1879 do what you want?

dtumad:
Thanks — #1879 is the better refactor. I've reduced this to the one piece we still need; it sits in
the default loop, so it applies unchanged on top of #1879.

We're not adding an architecture: we build the same `rv64d` model under a different platform config
and substitute it for the stock one. Lake identifies packages by name and `riscv-lean` requires
`Lean_RV64D`, so renaming the output means Lake resolves a second copy of the model alongside ours.
`CUSTOM_LEAN_ARCH` renames it by construction; `RV64D` gets past the case-sensitive guard but then
collides with the default family's rule.

Since the loop funnels every backend through one `config_file`, that case is a few lines: if
`SAIL_FORMAL_CONFIG_<ARCH>` is set, use it instead of the generated default and log it at configure
time. No new targets, no renaming, inert unless set. Happy to fold it into #1879 instead if you'd
rather.

Separately opened #1885 for the `${config_file}` `DEPENDS` additions on the SMT/rmem/Rocq rules.

pmundkur:
> Lake identifies packages by name and `riscv-lean` requires `Lean_RV64D`, so renaming the output means Lake resolves a second copy of the model alongside ours.

I'm not a Lean person, so I'm not familiar with `riscv-lean`.  Did you mean `lean-riscv` under `lean_emulator`, or something else?

dtumad:
> > Lake identifies packages by name and `riscv-lean` requires `Lean_RV64D`, so renaming the output means Lake resolves a second copy of the model alongside ours.
> 
> I'm not a Lean person, so I'm not familiar with `riscv-lean`. Did you mean `lean-riscv` under `lean_emulator`, or something else?

This library: https://github.com/opencompl/riscv-lean

pmundkur:
> We're not adding an architecture: we build the same `rv64d` model under a different platform config and substitute it for the stock one. Lake identifies packages by name and `riscv-lean` requires `Lean_RV64D`, so renaming the output means Lake resolves a second copy of the model alongside ours. `CUSTOM_LEAN_ARCH` renames it by construction; `RV64D` gets past the case-sensitive guard but then collides with the default family's rule.

I'm not familiar with the Lean package system.  Does the above mean that you can't really use `CUSTOM_LEAN_ARCH` and `CUSTOM_LEAN_CONFIG` from #1879?

dtumad:
Yes, we can use `CUSTOM_LEAN_ARCH`/`CUSTOM_LEAN_CONFIG` to generate the desired semantics under a new Lean model name. What it does not directly support is our preferred integration: generating the custom configuration under the existing `Lean_RV64D`/`LeanRV64D` identity.

Concretely, [`riscv-lean`](https://github.com/opencompl/riscv-lean/blob/main/lakefile.toml) declares:

```toml
[[require]]
name = "Lean_RV64D"
git = "https://github.com/opencompl/sail-riscv-lean.git"
```

and its sources contain imports and namespace references such as:

```lean
import LeanRV64D.Defs
import LeanRV64D

open LeanRV64D.Functions
open LeanRV64D.Defs
```

There are also many fully qualified references such as `LeanRV64D.Functions.xlen`; for example, see [`RISCV/SailPure.lean`](https://github.com/opencompl/riscv-lean/blob/main/RISCV/SailPure.lean).

In #1879, the architecture name determines that generated interface:

```cmake
string(TOUPPER ${arch} arch_uppercase)
-o "Lean_${arch_uppercase}"

OUTPUT "Lean_${arch_uppercase}/Lean${arch_uppercase}.lean"
add_custom_target(
    generated_lean_${arch}
    DEPENDS "Lean_${arch_uppercase}/Lean${arch_uppercase}.lean"
)
```

Thus `CUSTOM_LEAN_ARCH=SP1` generates `Lean_SP1/LeanSP1.lean` and modules under `LeanSP1`, while `riscv-lean` expects the package `Lean_RV64D` and modules under `LeanRV64D`. That gives us an additional model rather than a transparent replacement for the model consumed by `riscv-lean`.

Keeping the custom model identity as `Lean_RV64D`/`LeanRV64D` should let us substitute it without a large mechanical rename throughout `riscv-lean`. We would still need to check whether any proofs depend on semantics changed by the custom configuration; preserving the names only gives us the source-compatible integration path.

The reason we cannot achieve that merely by removing the `rv64d` check in #1879 is that the default loop has already effectively called:

```cmake
add_lean_targets(rv64d, <default-config>)
```

Calling it again for the custom configuration would redefine both `generated_lean_rv64d` and `Lean_RV64D/LeanRV64D.lean`. Using `RV64D` changes the CMake target case, but `string(TOUPPER ...)` still maps both invocations to the same output path.

#1879 could instead skip the default RV64D invocation whenever a custom RV64D configuration is supplied. This PR implements that replacement behavior at the earlier and smaller decision point: it substitutes the custom `config_file` before the existing RV64D targets and outputs are created.

pmundkur:
Thanks for the explanation, it makes more sense to me now!

> #1879 could instead skip the default RV64D invocation whenever a custom RV64D configuration is supplied. This PR implements that replacement behavior at the earlier and smaller decision point: it substitutes the custom `config_file` before the existing RV64D targets and outputs are created.

I'm still not clear though on whether we need to support the `CUSTOM_LEAN*` variables in #1879 once we merge this PR.  i.e. if we remove those `CUSTOM_` variables from #1879, then merge this PR after that, will your use case be satisfied?  I'd rather not add CUSTOM variables if they will not be used.

jordancarlin:
It seems like a more robust solution would be for `riscv-lean` to require an interface named `Lean_RV` (or something similar) and make the generated interface name agnostic to the config name. We could still generate different configs in different directories, but keep the interface name consistent.

dtumad:
> It seems like a more robust solution would be for `riscv-lean` to require an interface named `Lean_RV` (or something similar) and make the generated interface name agnostic to the config name. We could still generate different configs in different directories, but keep the interface name consistent.

I guess cc @tobiasgrosser @luisacicolini @alexanderlhicks. I'm fine if things also change there rather than here, I'd just like changing the platform config to be compatible cleanly with that existing Lean work without manual `sed` scripting.

PR #1863, opened 2026-08-07
Use Bare translation when S-mode is unsupported

michalchm:
The model cannot run code below M-mode on a hart without S-mode. For non-M-mode
accesses, `translationMode` uses `mstatus.SXL`, which is zero when S-mode is
unsupported, causing an internal error.

Reproduction on current master:

Save this configuration override as `/tmp/no-s.json`:
```
    {
      "extensions": {
        "S": {"supported": false},
        "Sv39": {"supported": false},
        "Sv48": {"supported": false},
        "Sv57": {"supported": false},
        "Svrsw60t59b": {"supported": false},
        "Ssccptr": {"supported": false},
        "Svade": {"supported": false},
        "Svadu": {"supported": false},
        "Svpbmt": {"supported": false},
        "Svnapot": {"supported": false},
        "Svvptc": {"supported": false},
        "Sstvala": {"supported": false}
      }
    }
```
Then
```
$ sail_riscv_sim --config-override /tmp/no-s.json rv64ui-p-simple  
Internal error: core/types.sail:62: architecture(0b00) is invalid
```
A hart without S-mode has no satp and no page tables, so every implemented
privilege uses Bare translation. The fix adds the S-less case to the existing
M-mode check, after which the command above succeeds.

### DIFF
--- a/model/sys/vmem.sail
+++ b/model/sys/vmem.sail
@@ -338,7 +338,8 @@ private function xatp_to_ppn(stage : TranslationStage) -> bits(if xlen == 32 the
 
 // Return the active address translation mode for S-Stage translation.
 private function satp_mode(priv : Privilege) -> SATPMode = {
-  if priv == Machine then Bare
+  // Without S-mode enabled, satp is inactive and translation is Bare at every privilege.
+  if priv == Machine | not(currentlyEnabled(Ext_S)) then Bare
   else {
     let arch = architecture(Supervisor);
     let mbits : satp_mode = match arch {


model/sys/vmem.sail
```diff
@@ -246,7 +246,8 @@ function satp_to_ppn(satp_val) =
 
 // Compute address translation mode from SATP register
 function translationMode(priv : Privilege) -> SATPMode = {
-  if priv == Machine then Bare
+  // A hart without S-mode has no SATP, so translation is Bare at every privilege.
+  if priv == Machine | not(hartSupports(Ext_S)) then Bare
```
jordancarlin: This should apply when S is disabled at runtime too (by setting `misa.S` = 0).

```suggestion
  if priv == Machine | not(currentlyEnabled(Ext_S)) then Bare
```

jordancarlin (changes requested):
I'm surprised no one has run into this. Good catch!

nadime15:
Hi @michalchm, thanks for your contribution! I took a look and the fix makes sense (with @jordancarlin's addition).

We are currently trying to get #1769 into this release, and it refactors a large part of the vmem.sail code. Because of that, I would prefer to hold off on approving this PR for now (the fix itself looks correct!) just to avoid unnecessary rebase pain.

Once #1769 lands, you will need to rebase, but the actual fix should stay the same.

model/sys/vmem.sail
michalchm: Agreed, currentlyEnabled(Ext_S) is the correct check. I'll apply it.

michalchm:
Thanks @jordancarlin and @nadime15!

I'm happy to rebase #1863 after #1769 lands, and I'm not asking for it to merge ahead of that work.

Some scheduling context. I found this fix while preparing the Sail model for the Smpmpind, Smcfiss, Smucfiss and Sspmpss ratification package. I have two more branches ready to open as separate PRs. One covers Smpmpind, Smcfiss and Smucfiss and carries #601 and #956. The other builds on the first, covers Sspmpss and carries #1457. In both, the dependency work is kept as separate commits with the original authors preserved.

For the ratification schedule, the Sail work needs to be submitted, but it does not need to be merged first. These PRs will not be ready to merge until their dependencies land.

I was planning to open both as drafts, with their dependencies clearly marked, and rebase or refactor them as #1769 and the dependency PRs land. If you'd prefer a particular way of marking or organizing them so they don't add review load in the meantime, let me know and I'll set them up that way.

nadime15:
Sounds good drafts with the dependencies listed works. No rush on rebasing.

model/sys/vmem.sail
michalchm: Applied in 84220339. Thanks.

jordancarlin:
@michalchm now that #1769 is merged, can you rebase this PR? Once you do, we should be able to get it merged fairly quickly.

jordancarlin (approved):

pmundkur (approved):

PR #1864, opened 2026-08-08
Add Smpmpind indirect PMP CSR access

michalchm:
This PR covers Smpmpind only. Smcfiss and Smucfiss are being submitted separately.

Adds experimental support for Smpmpind. It exposes PMP address, configuration and per entry E bits through miselect, mireg and mireg2, and defines the E field in the indirect pmpcfg view. Smpmpind itself defines no valid XWR encoding when E is set, so accesses to such a region raise an access fault unless another standard extension defines that encoding.

Specification: https://github.com/ved-rivos/riscv-isa-manual/blob/smpmpss/src/smcfiss.adoc

A preparatory commit exposes existing PMP helpers so the extensions can reuse them. It changes no behavior.

Dependencies:

- #601, carried as separate commits
- #956, represented by a temporary port

The dependency work remains in separate commits with its source authors preserved. It will be removed or reconciled when those PRs merge.

Smpmpind requires --enable-experimental-extensions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ An experimental emulator in the Lean language is available, see its
 - Machine, Supervisor, and User modes
 - Smcntrpmf extension for cycle and instret privilege mode filtering, v1.0
 - Smstateen/Ssstateen extensions for fine-grained privileged state access control, v1.0
+- Smepmp extension for pmp enhancements for memory access and execution prevention on Machine mode, v1.0
 - Ssccptr extension for Main memory supports hardware page table reads, v1.0
 - Sscounterenw extension for writable enables for any supported counter, v1.0
 - Sscofpmf extension for Count Overflow and Mode-Based Filtering, v1.0
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -780,6 +780,12 @@
     "Smepmp": {
       "supported": true
     },
+    "Smcsrind": {
+      "supported": true
+    },
+    "Sscsrind": {
+      "supported": true
+    },
     "Sstc": {
       "supported": true
     },
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -777,6 +777,9 @@
     "Sscounterenw": {
       "supported": true
     },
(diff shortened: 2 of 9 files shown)


nadime15:
I recommend having a dedicated PR for each individual extension. It would make the changes easier to review and also keep the Git commit history cleaner. I understand that these extensions share some code, but I think it would still be better to separate them if possible and just duplicate code in the meanwhile.

jordancarlin:
It might also be possible to use the new GitHub stacked PRs feature: https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests.

nadime15:
> It might also be possible to use the new GitHub stacked PRs feature: https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests.

This looks awesome! This would be fine too. At the same time, we can learn how this new feature works 😄

michalchm:
That does look useful. However, from what I understand, GitHub does not support [cross fork stacks](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs#where-can-you-use-stacked-pull-requests). My branches are in my fork, so I do not think I can set one up here.

I have been working on the split locally, with the dependency commits kept separate and one commit per extension. Smcfiss builds on Smpmpind, and Smucfiss builds on Smcfiss.

If cross fork stacks are not possible, would three normal draft PRs against master be okay? Their diffs against master would initially be cumulative, but the dependency work and each extension would remain in separate commits. I would rebase the upper PRs as the lower ones merge.

nadime15:
> If cross fork stacks are not possible, would three normal draft PRs against master be okay? Their diffs against master would initially be cumulative, but the dependency work and each extension would remain in separate commits. I would rebase the upper PRs as the lower ones merge.

Yeah sure this works too.

michalchm:
Split as agreed. This PR is now Smpmpind only. Smcfiss is #1865 and Smucfiss is #1866.

The diffs are cumulative against master, so the later PRs also show the extensions they depend on.

PR #1869, opened 2026-08-11
Refresh timer pending bits when `menvcfg` is written.

nadime15:
Follow up to #1853. mip[STI] is derived from stimecmp, gated by menvcfg[STCE], so setting STCE changes the timer condition the same way a stimecmp write does. Refresh it there too rather than leaving it stale until the next tick (which is legal, just inconsistent).

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -469,9 +469,14 @@ function clause read_CSR(0x30A) = menvcfg.bits[xlen - 1 .. 0]
 function clause read_CSR(0x31A if xlen == 32) = menvcfg.bits[63 .. 32]
 function clause read_CSR(0x10A) = read_senvcfg().bits[xlen - 1 .. 0]
 
-function clause write_CSR((0x30A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, menvcfg.bits[63 .. 32] @ value); Ok(menvcfg.bits[31 .. 0]) }
-function clause write_CSR((0x30A, value) if xlen == 64) = { menvcfg = legalize_menvcfg(menvcfg, value); Ok(menvcfg.bits) }
-function clause write_CSR((0x31A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, value @ menvcfg.bits[31 .. 0]); Ok(menvcfg.bits[63 .. 32]) }
+// Defined in sys/platform.sail, which comes after this file in the compile order.
+val clint_dispatch : bool -> unit
+
+// menvcfg[STCE] decides whether mip[STI] is derived from stimecmp, so writing
+// it changes the timer interrupt condition just like a stimecmp write does.
+function clause write_CSR((0x30A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, menvcfg.bits[63 .. 32] @ value); clint_dispatch(false); Ok(menvcfg.bits[31 .. 0]) }
+function clause write_CSR((0x30A, value) if xlen == 64) = { menvcfg = legalize_menvcfg(menvcfg, value); clint_dispatch(false); Ok(menvcfg.bits) }
+function clause write_CSR((0x31A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, value @ menvcfg.bits[31 .. 0]); clint_dispatch(false); Ok(menvcfg.bits[63 .. 32]) }
 function clause write_CSR(0x10A, value) = { senvcfg = legalize_senvcfg(senvcfg, zero_extend(value)); Ok(read_senvcfg().bits[xlen - 1 .. 0]) }
 
 // Return whether or not FIOM is currently active, based on the current


model/core/sys_regs.sail
```diff
@@ -469,9 +469,14 @@ function clause read_CSR(0x30A) = menvcfg.bits[xlen - 1 .. 0]
...
-function clause write_CSR((0x30A, value) if xlen == 64) = { menvcfg = legalize_menvcfg(menvcfg, value); Ok(menvcfg.bits) }
-function clause write_CSR((0x31A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, value @ menvcfg.bits[31 .. 0]); Ok(menvcfg.bits[63 .. 32]) }
+// Defined in sys/platform.sail, which comes after this file in the compile order.
+val clint_dispatch : bool -> unit
+
+// menvcfg[STCE] decides whether mip[STI] is derived from stimecmp, so writing
+// it changes the timer interrupt condition just like a stimecmp write does.
+function clause write_CSR((0x30A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, menvcfg.bits[63 .. 32] @ value); clint_dispatch(false); Ok(menvcfg.bits[31 .. 0]) }
+function clause write_CSR((0x30A, value) if xlen == 64) = { menvcfg = legalize_menvcfg(menvcfg, value); clint_dispatch(false); Ok(menvcfg.bits) }
+function clause write_CSR((0x31A, value) if xlen == 32) = { menvcfg = legalize_menvcfg(menvcfg, value @ menvcfg.bits[31 .. 0]); clint_dispatch(false); Ok(menvcfg.bits[63 .. 32]) }
```
pmundkur, line 479: I actually wonder if it would be simpler to avoid having the `clint_dispatch()` invoked eagerly on every single CSR write that affects the timer interrupts, and instead have timer handling all done just once when the timer is ticked.  That would also avoid needing the forward declaration.  Since the lazy approach is legal, any reason to do this, other than consistency with `stimecmp`?  We could revert the `stimecmp` eager update and things could then be consistent with the lazy approach? `stimecmp` write handling does not have to be consistent with `mtimecmp` since `stimecmp` is a CSR and `mtimecmp` is not.

nadime15: I agree I am not a fan of this either because I needed to add multiple clint_dispatch calls to the Hypervisor PR as well. 

But I am not so sure about the approach. There are two things:

a) Dealing with `instructions_per_tick`, tick_clock() only runs every N instructions, so after a csr write the pending bit stays wrong for up to N instructions. the spec says the timer comparison only has to show up eventually so its legal, but if someone sets N high the window gets quite big (imagine setting instructions_per_tick to 10000, i would drive an external interrupt but it would fire much later).

b) It doesnt cover everything. For example, hip.VSTIP is hvip.VSTIP | timer and only the timer part is allowed to be late. if we delay the hvip part to the next tick thats wrong, so hvip writes stay eager either way and we dont fully get rid of the problem.

I think what we actually wanna do is computing those bits on read instead at every tick. Basically what we already do with read_mip(IncludePlatformInterrupts) which does that for MEIP/SEIP.

pmundkur: Good point; it makes sense to treat `hvip` as a special case, but I don't think any of the others are really needed?  For a), I think external interrupts are only guaranteed eventual delivery, so firing later is also legal?

nadime15: > Good point; it makes sense to treat hvip as a special case, but I don't think any of the others are really needed? For a), I think external interrupts are only guaranteed eventual delivery, so firing later is also legal?

The same applies to timer interrupts too, but yeah, you are right that it is legal. I think it comes down to what is easier to implement. If the lazy approach is equally difficult to implement as the on read approach, then I would rather to go with the on read approach, because if we add the lazy approach, we would then have 2–3 different approaches. The lazy one, the MEIP/SEIP approach, which is updated on read, and then the hvip case (which is probably also on read, but I am not sure). Whereas if we implemented everything on read, that would cover all cases simultaneously and be more future proof, without the need for additional special cases later.

Btw. I think (I am not 100% sure about that), that changing `menvcfg[STCE]` or `henvcfg[STCE]` (part of Sstc), also needs to be reflected immediately. I think the sentence about guaranteed eventual delivery only applies to the comparison results (e.g. `bool_to_bit(stimecmp <=_u mtime)`).

We already have the read_mip() function that is called after each instruction, so we would only have to move the timer bits into read_mip(), then all the clint_dispatch calls go away, including the one in tick_clock(). it would work like external_interrupts_pending() already does.

nadime15: > Btw. I think (I am not 100% sure about that), that changing menvcfg[STCE] or henvcfg[STCE] (part of Sstc), also needs to be reflected immediately. I think the sentence about guaranteed eventual delivery only applies to the comparison results (e.g. bool_to_bit(stimecmp <=_u mtime)).

Ok I started to dig a bit deeper and it seems that CSR field modulation rule applies (see here riscv/riscv-isa-manual#1636). It essentially says that the value of mip.STIP is unspecified when STCE changes (in both directions). So lazy is fine here too and hvip is the only case left.

nadime15:
Regarding my comment [here](https://github.com/riscv/sail-riscv/pull/1869#discussion_r3767225109), There is the following example of a reachable (legal) path in our code that we should be aware of. If menvcfg.STCE goes from 1 to 0 while mip.STI is set, mip.STI gets stuck at 1.

For example:

a) M-mode sets menvcfg.STCE = 1.
b) We then write stimecmp such that mip[STI] = bool_to_bit(stimecmp <=_u mtime) = 1.
c) We clear menvcfg.STCE = 0.
d) tick_clock calls clint_dispatch, but it can no longer update mip[STI] because of the following condition:
```
mip[STI] = bool_to_bit(stimecmp <=_u mtime);
};
```

So once menvcfg.STCE is cleared, mip[STI] remains set and is never updated. If mie.STIE = 1 and mideleg.STI = 1, the trap is delegated to S-mode, which will not be able to clear mip.STI, causing the trap to fire indefinitely but I guess if software clears menvcfg.STCE it needs to clear mip.STI too.

model/core/sys_regs.sail
pmundkur: Yeah, I think moving the timer bits into `read_mip` seems to be the more consistent approach.  I suggest we do this in a separate PR after the hypervisor merge.

nadime15: Ok sounds good. What about this PR, should we close it or merge it?

pmundkur: Is it ok with you if we close this, and revert the `clint_dispatch` calls you added to the hypervisor (except in the case of `hvip`), and instead adding a TODO for the `read_mip` approach in the hypervisor PR?  I think the TODO could go with the `clint_dispatch()` for `hvip`.

nadime15: Yeah sure, will do.

nadime15:
Closing this, since we agreed on a different [approach](https://github.com/riscv/sail-riscv/pull/1869#discussion_r3766656868) for the future.

PR #1871, opened 2026-08-12
Port to v2 of the Sail concurrency interface

febyeji:
This ports the RISC-V physical memory interface to version 2 of the Sail concurrency interface:

- Select version 2 of the Sail concurrency interface.
- Classify the existing RISC-V memory access kinds for the v2 interface, including instruction fetches, acquire/release accesses, and LR/SC.
- Use the same `Mem_request` structure for both reads and writes.
- Exchange memory values as little-endian byte sequences, as required by the v2 API.

The change is based on bacam/archsem2 commit 2feb36fb and rebased onto the current master.


model/core/phys_mem_interface.sail
```diff
@@ -40,6 +41,62 @@ enum read_kind = {
   Read_RISCV_reserved_strong_acquire,
 }
 
+union access_kind = {
+  AK_Write : write_kind,
+  AK_Read : read_kind,
+}
+
+// TODO: We currently do not get enough information here to distinguish explicit accesses,
+// translation accesses, and atomic accesses.
```
pmundkur: I think you could get this information if  `access : MemoryAccessType(mem_payload)` was passed to `write_ram()` and `read_ram()`?

pmundkur (commented):
Is there any way to implement `write_ram_ea()` for this interface?  See #1761.

model/core/phys_mem_interface.sail
febyeji: Right, I think that should work. That should let the V2 predicates distinguish explicit, translation-table-walk, and atomic accesses correctly.

febyeji:
For `write_ram_ea()`, I think we can construct the same `Mem_request` as in `write_ram()` and pass it to `sail_mem_address_announce`. I missed that when doing the initial port. I’ll update it.

model/core/phys_mem_interface.sail
pmundkur: FYI, we plan to merge #1769 relatively soon, which adds another `LoadExecute(Data)` payload type for hypervisor.

febyeji: Thanks for the pointer. The latest CI run failed because of an interface generated by Sail itself, so Sail needs to be updated first. I’ll fix Sail, and once #1769 is merged, I’ll incorporate those changes as well.

bacam:
Another small update we should have for Isabelle and HOL4 output: https://github.com/riscv/sail-riscv/commit/5a33810d216b43126819c62815b86a41ecf2d0c8

pmundkur:
@febyeji you'll need to add your signing key to Github to get past the "Commits must have verified signatures" check. See here for instructions if you need them:
Then you'll need to push these commits again.

febyeji:
Thanks for the guidance! I’ve added my SSH signing key and re-signed the five commits. I’m currently looking into the Sail and lean-sail build issues.

bacam:
That lean CI failure is just a missing `apt update` in the action; I'll push a fix.

model/CMakeLists.txt
```diff
@@ -420,6 +420,9 @@ foreach (xlen IN ITEMS 32 64)
     set(lean_sail_common
         --config ${config_file}
         --lean
+        # G-stage translation cannot recursively translate PTE addresses again.
+        # Specialize its constant stage argument to break the VS/G-stage cycle.
+        -const_prop_mutrec translate_g_stage
```
pmundkur, line 425: Could you explain how this works? I still see `translate_g_stage()` being called in `translate_PTE_addr()` in the generated Lean.

pmundkur (commented):
It's great to see the Lean build fixed!

model/CMakeLists.txt
febyeji: Yes, when a VS-stage walk accesses a PTE, `translate_PTE_addr` calls `translate_g_stage` to translate the PTE’s GPA. This starts a G-stage walk, whose `pt_walk` calls a specialized copy of `translate_PTE_addr` with stage fixed to `G_Stage` (generated as `translate_PTE_addr_mutrec_vv_zG_Stagev` in Lean). This function does not call `translate_g_state` again, so the mutual recursion terminates.

bacam:
I've added a commit to use the constant propagation for the other theorem prover targets, which makes the termination measures other than the one for `pt_walk` unnecessary.  @febyeji Feel free to rework this if you want/need to.

PR #1874, opened 2026-08-14
Add indirect CSR access framework (Smcsrind/Sscsrind)

christian-herber-nxp:
This is a prerequisite to AIA.
Coding done by Opus 4.8
Architected and reviewed by me.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -857,6 +857,14 @@
     },
     "Svvptc": {
       "supported": true
+    },
+    // Machine-level indirect CSR access (miselect/mireg). Required by Smaia.
+    "Smcsrind": {
+      "supported": true
+    },
+    // Supervisor-level indirect CSR access (siselect/sireg). Required by Ssaia.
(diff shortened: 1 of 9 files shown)


nadime15:
There is already a PR for Smcsrind/Sscsrind, #956, which has also been used as the basis for #1864 and #1866. If you need these extensions for AIA, I think it would make sense to use #956 for the AIA work. It is fairly complete (includes the Hypervisor related parts), and has gone through several rounds of review.

christian-herber-nxp:
> There is already a PR for Smcsrind/Sscsrind, #956, which has also been used as the basis for #1864 and #1866. If you need these extensions for AIA, I think it would make sense to use #956 for the AIA work. It is fairly complete (includes the Hypervisor related parts), and has gone through several rounds of review.

Thanks for the pointer, I should have checked for that before. After inspecting the PR, I feel like the architecture in this PR is better, and I have provided my comment in #956

PR #1875, opened 2026-08-14
Correct cache key reference in Rocq CI

bacam:
Fix a typo that we didn't notice because we only store caches on the master branch, so it didn't show up until after merging the PR.

### DIFF
--- a/.github/workflows/rocq.yml
+++ b/.github/workflows/rocq.yml
@@ -37,7 +37,7 @@ jobs:
           sail-version: ${{ env.SAIL_VERSION }}
 
       - name: Restore cached opam
-        id: cache-opam-restore
+        id: cache-rocq-opam-restore
         uses: actions/cache/restore@v6
         with:
           path: ~/.opam


jordancarlin (approved):

jordancarlin:
Seems like the Lean failure is unrelated:

error: LeanRV64D/RiscvExtras.lean:14:5: unknown namespace `LeanRV64D.Defs`
    error: LeanRV64D/RiscvExtras.lean:31:32: Function expected at
      SailM
    but this term has type
      ?m.1
    
    Note: Expected a function because this term is being applied to the argument
      Unit
    
    Hint: The identifier `SailM` is unknown, and Lean's `autoImplicit` option causes an unknown identifier to be treated as an implicitly bound variable with an unknown type. However, the unknown type cannot be a function, and a function is what Lean expects here. This is often the result of a typo or a missing `import` or `open` statement.
    error: LeanRV64D/RiscvExtras.lean:32:30: Function expected at
      SailM
    but this term has type
      ?m.1
    
    Note: Expected a function because this term is being applied to the argument
      String
    
    Hint: The identifier `SailM` is unknown, and Lean's `autoImplicit` option causes an unknown identifier to be treated as an implicitly bound variable with an unknown type. However, the unknown type cannot be a function, and a function is what Lean expects here. This is often the result of a typo or a missing `import` or `open` statement.
    error: LeanRV64D/RiscvExtras.lean:35:41: Function expected at
      SailM
    but this term has type
      ?m.1
    
    Note: Expected a function because this term is being applied to the argument
      Unit
    
    Hint: The identifier `SailM` is unknown, and Lean's `autoImplicit` option causes an unknown identifier to be treated as an implicitly bound variable with an unknown type. However, the unknown type cannot be a function, and a function is what Lean expects here. This is often the result of a typo or a missing `import` or `open` statement.
    error: LeanRV64D/RiscvExtras.lean:37:34: Function expected at
      SailM
    but this term has type
      ?m.1
    
    Note: Expected a function because this term is being applied to the argument
      Unit
    
    Hint: The identifier `SailM` is unknown, and Lean's `autoImplicit` option causes an unknown identifier to be treated as an implicitly bound variable with an unknown type. However, the unknown type cannot be a function, and a function is what Lean expects here. This is often the result of a typo or a missing `import` or `open` statement.
    error: LeanRV64D/RiscvExtras.lean:40:34: Function expected at
      SailM
    but this term has type
      ?m.1
    
    Note: Expected a function because this term is being applied to the argument
      (BitVec 16)
    
    Hint: The identifier `SailM` is unknown, and Lean's `autoImplicit` option causes an unknown identifier to be treated as an implicitly bound variable with an unknown type. However, the unknown type cannot be a function, and a function is what Lean expects here. This is often the result of a typo or a missing `import` or `open` statement.
    error: LeanRV64D/RiscvExtras.lean:116:12: Invalid field notation: Field projection operates on types of the form `C ...` where C is a constant. The expression
      extension
    has type `Sort ?u.2733` which does not have the necessary form.
    Some required targets logged failures:
    - LeanRV64D.RiscvExtras

bacam:
The Lean failure is an upstream change that needs a tweak to the handwritten file, and I'll put in a separate pull request for that.

pmundkur (approved):

PR #1877, opened 2026-08-14
Add bounds checks when loading ELF files.

pmundkur:
This adds two checks:

- each ELF segment is required to fit into a defined memory region

- ELF segments are required to not overlap

The second check is more important when multiple ELF files are loaded.

### DIFF
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -129,7 +129,8 @@ void deep_merge_json(jsoncons::json &base, const jsoncons::json &json_override, 
   base = json_override;
 }
 
-void write_dtb_to_rom(ModelImpl &model, const std::vector<uint8_t> &dtb) {
+void write_dtb_to_rom(ModelImpl &model, const std::string &filename, elf_info &elf_info) {
+  std::vector<uint8_t> dtb = read_file(filename);
   uint64_t addr = get_config_uint64({"memory", "dtb_address"});
   uint64_t size = static_cast<uint64_t>(dtb.size());
 
@@ -144,9 +145,10 @@ void write_dtb_to_rom(ModelImpl &model, const std::vector<uint8_t> &dtb) {
   if (!model.dtb_within_configured_pma_memory(addr, size)) {
     fprintf(
       stderr,
-      "DTB does not fit in any configured PMA memory region: "
+      "DTB in '%s' does not fit in any configured PMA memory region: "
       "addr=0x%0" PRIx64 ", size=0x%0" PRIx64 " (end=0x%0" PRIx64 ")\n"
       "Hint: adjust memory.dtb_address or memory.regions in the config.\n",
+      filename.c_str(),
       addr,
       size,
       end
@@ -154,9 +156,24 @@ void write_dtb_to_rom(ModelImpl &model, const std::vector<uint8_t> &dtb) {
     exit(EXIT_FAILURE);
   }
 
+  // Check for conflicts with already loaded regions.
+  const auto &loaded_region_ptr =
+    std::find_if(elf_info.loaded_regions.begin(), elf_info.loaded_regions.end(), [addr, size](const auto &ent) {
+      return addr + size > ent.offset && ent.offset + ent.length > addr;
+    });
+  if (loaded_region_ptr != elf_info.loaded_regions.end()) {
+    std::ostringstream msg;
+    msg << "DTB in " << filename << " of size 0x" << std::hex << size << " @0x" << addr
+        << " conflicts with already loaded segment from file " << loaded_region_ptr->filename << " of size 0x"
+        << loaded_region_ptr->length << " @0x" << loaded_region_ptr->offset << std::endl;
+    throw std::runtime_error(msg.str());
+  }
+
   for (uint8_t d : dtb) {
     write_mem(addr++, d);
   }
+
+  elf_info.loaded_regions.emplace_back(filename, addr, size);
 }
 
 void write_signature(const std::string &file, unsigned signature_granularity, const elf_info &elf_info) {
@@ -696,7 +713,7 @@ InitResult init_model(
 
   if (!opts.dtb_file.empty()) {
     fprintf(stderr, "using %s as DTB file.\n", opts.dtb_file.c_str());
-    write_dtb_to_rom(model, read_file(opts.dtb_file));
+    write_dtb_to_rom(model, opts.dtb_file, elf_info);
   }
 
   entry = run_info.rvfi.has_value() ? rvfi_handler::get_entry()
--- a/c_emulator/riscv_sim.cpp
+++ b/c_emulator/riscv_sim.cpp
@@ -265,13 +265,45 @@ uint64_t load_sail(ModelImpl &model, const std::string &filename, bool main_file
     break;
   }
 
+  auto memory_regions = model.memory_regions();
+
   // Load into memory.
-  elf.load([](uint64_t address, const uint8_t *data, uint64_t length) {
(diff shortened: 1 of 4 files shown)


pmundkur:
Extracted from #1855.

c_emulator/riscv_sim.cpp
```diff
@@ -265,13 +265,45 @@ uint64_t load_sail(ModelImpl &model, const std::string &filename, bool main_file
...
+    // technically be contiguous and compatible for the ELF segment,
+    // but this is unlikely.
+    const auto &memory_region_ptr =
+      std::find_if(memory_regions.begin(), memory_regions.end(), [address, length](const auto &ent) {
+        return address >= ent.base && address + length <= ent.base + ent.size;
+      });
+    if (memory_region_ptr == memory_regions.end()) {
+      std::ostringstream msg;
+      msg << "Cannot load segment from " << filename << " of size 0x" << std::hex << length << " @0x" << address
+          << " since it does not lie within any defined main memory region." << std::endl;
```
nadime15: The for loop in memory_regions() seems to include all memory regions, not only the main memory?

c_emulator/riscv_sim.cpp
```diff
@@ -265,13 +265,45 @@ uint64_t load_sail(ModelImpl &model, const std::string &filename, bool main_file
...
-  elf.load([](uint64_t address, const uint8_t *data, uint64_t length) {
+  elf.load([&filename, &elf_info, &memory_regions](uint64_t address, const uint8_t *data, uint64_t length) {
+    // Ensure these ELFs are loaded into defined memory regions. Each
+    // ELF segment is required to fit into a single memory
+    // region. This is a conservative check: the memory regions could
+    // technically be contiguous and compatible for the ELF segment,
+    // but this is unlikely.
+    const auto &memory_region_ptr =
+      std::find_if(memory_regions.begin(), memory_regions.end(), [address, length](const auto &ent) {
+        return address >= ent.base && address + length <= ent.base + ent.size;
```
nadime15: Seems like address + length can overflow, and the same applies to the code at line 291.

c_emulator/riscv_sim.cpp
```diff
@@ -265,13 +265,45 @@ uint64_t load_sail(ModelImpl &model, const std::string &filename, bool main_file
...
+          << loaded_region_ptr->length << " @0x" << loaded_region_ptr->offset << std::endl;
+      throw std::runtime_error(msg.str());
+    }
+
     // TODO: We could definitely improve on rts.c's memory implementation
     // (which is O(N^2)) and writing one byte at a time here.
     for (uint64_t i = 0; i < length; ++i) {
       write_mem(address + i, data[i]);
     }
+    elf_info.loaded_regions.emplace_back(filename, address, length);
```
nadime15, line 328: I think an elf segment can still overwrite the dtb written earlier by write_dtb_to_rom, since that write is not recorded in loaded_regions. With the default dtb_address of 0x1000 a segment there would pass both checks.

c_emulator/riscv_sim.cpp
pmundkur: Right, these new checks are just checking for overlaps, not overflows.  We could add overflow checks in a subsequent PR perhaps? (I don't think they are as critical, since it indicates a malformed ELF file, which this is not guarding against).

c_emulator/riscv_sim.cpp
pmundkur: Fixed the log message.

c_emulator/riscv_sim.cpp
pmundkur: True; this requires a change in the API for `write_dtb_to_rom` to add to the book-keeping.  I could do that in a subsequent PR.

c_emulator/riscv_sim.cpp
nadime15: What if a config has a region ending exactly at 2^64? Then ent.base + ent.size overflows to 0 and the check rejects every segment, so a valid elf would be rejected.

c_emulator/riscv_sim.cpp
nadime15: Yes that works.

c_emulator/riscv_sim.cpp
pmundkur: Good point, added handling for this case (I think!).

c_emulator/riscv_sim.cpp
pmundkur: It wasn't as big of a change as I'd imagined, so I added this as well.

c_emulator/riscv_sim.cpp
```diff
@@ -144,19 +145,35 @@ void write_dtb_to_rom(ModelImpl &model, const std::vector<uint8_t> &dtb) {
...
+    msg << "DTB in " << filename << " of size 0x" << std::hex << size << " @0x" << addr
+        << " conflicts with already loaded segment from file " << loaded_region_ptr->filename << " of size 0x"
+        << loaded_region_ptr->length << " @0x" << loaded_region_ptr->offset << std::endl;
+    throw std::runtime_error(msg.str());
+  }
+
   for (uint8_t d : dtb) {
     write_mem(addr++, d);
   }
+
```
nadime15, line 176:
```suggestion
 for (uint64_t i = 0; i < size; ++i) {
   write_mem(addr + i, dtb[i]);
  }

```
Similar to what we do on line 272, otherwise, you pass the wrong address.

c_emulator/riscv_sim.cpp
nadime15: Yeah I think this works.

c_emulator/riscv_sim.cpp
pmundkur: Good catch; I  used a simpler approach.

c_emulator/riscv_sim.cpp
```diff
@@ -265,13 +283,49 @@ uint64_t load_sail(ModelImpl &model, const std::string &filename, bool main_file
     break;
   }
 
+  auto memory_regions = model.memory_regions();
+
   // Load into memory.
-  elf.load([](uint64_t address, const uint8_t *data, uint64_t length) {
+  elf.load([&filename, &elf_info, &memory_regions](uint64_t address, const uint8_t *data, uint64_t length) {
+    // Ensure these ELFs are loaded into defined memory regions. Each
```
nadime15, line 290: Just one last thing, segments can have 0 bytes (see https://man7.org/linux/man-pages/man5/elf.5.html, p_filesz), so I think we should early check the size and ignore these cases otherwise we might risk rejecting these.
```suggestion
  elf.load([&filename, &elf_info, &memory_regions](uint64_t address, const uint8_t *data, uint64_t length) {
  if (length == 0) {
      return;
    }
    // Ensure these ELFs are loaded into defined memory regions. Each
```

pmundkur: It turns out that this check is actually insufficient; `elf_loader.cpp:ELF::load()` passes the file size (`filesz`) but then zeros the memory size (`memsz`), which could go outside the bounds.  I'll fix this.  (So `filesz` could be 0, but `memsz` almost always is not.)

pmundkur: Spoke too soon, the `memsz` is handled subsequently and will be bounds checked.  I don't think an early check is needed; we're already seeing `filesz == 0` for `htif` regions in our first party tests.

nadime15: Ah I see, makes sense.

nadime15 (approved):

PR #1887, opened 2026-08-20
Add `platform.msi_supported` config for cores without MSI.

edevolder-tt:
MSI may not be supported on some implementations. Default true preserves current behaviour. When false, mip.MSIP and mie.MSIE are read-only zero for CSR access and interrupt dispatch, sie/sip MSI is not writable via delegation, and the Simple Interrupt Generator cannot set MSIP.

Validated against ACT Interrupt suites for cores that do not implement MSI (Sail goldens aligned with Whisper/RTL mideleg WARL of 0x2222).

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -421,6 +421,10 @@
       "base": 33554432,
       "size": 786432
     },
+    // Whether machine software interrupts are implemented (mip.MSIP / mie.MSIE).
+    // When false, those bits are read-only zero for CSR access and interrupt
+    // dispatch, and the Simple Interrupt Generator cannot set MSIP.
+    "msi_supported": true,
     // Very simple MMIO device to generate interrupts for testing
     // purposes. See docs/SimpleInterruptGenerator.md for details.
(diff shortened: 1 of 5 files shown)


jordancarlin:
All three of the machine mode interrupts are optional (MSI, MTI, and MEI). It would be good to add config options for all three together so that they are consistently handled. Would you be willing to extend this PR to handle all of them?

edevolder-tt:
Sure. The current patch now has conflicts due to hypervisor drop....so I've
work to do.
Would you prefer each MSI MTI MEI as a separate patch, or all in one?
Eric

jordancarlin:
I imagine there will be a lot of similarity, so one patch is probably fine. If you feel like it makes more sense to break them up while you are working on it though, then go ahead.

Thanks for updating now that Hypervisor is in and sorry about the churn!

edevolder-tt:
I'm going to close this pull request as I've conflicts with hypervisor changes to resolve, and I'm going to apply the same strategy to all MTI MEI and MSI bits.

PR #1889, opened 2026-08-20
Allow mhpmevent VSINH/VUINH when H is enabled

nadime15:
They were hardcoded to zero with a "Hypervisor not supported" comment. Smcntrpmf already gates the same bits on Ext_H.


nadime15:
I completely missed this, but @WeiShen1011 opened a PR on my fork a few weeks ago (https://github.com/nadime15/sail-riscv/pull/11).

jordancarlin (approved):
Would be good to grep through the repo for any other instances of "Hypervisor not supported" (or similar) that might be left.

pmundkur (approved):

nadime15:
> Would be good to grep through the repo for any other instances of "Hypervisor not supported" (or similar) that might be left.

Looks like this was the last case.

PR #1891, opened 2026-08-21
Add platform msi_supported, mti_supported, and mei_supported config.

edevolder-tt:
Machine software, timer, and external interrupts may not be supported on some implementations. Default true preserves current behaviour. When false, the corresponding mip/mie bits are read-only zero for CSR access and interrupt dispatch, sie/sip cannot write them via delegation, and the Simple Interrupt Generator cannot set MSIP.

Validated against ACT4 Interrupt suites for cores that do not implement these interrupts.

### DIFF
--- a/model/core/interrupt_regs.sail
+++ b/model/core/interrupt_regs.sail
@@ -54,6 +54,13 @@ bitfield Minterrupts : xlenbits = {
 
 register hvip : Minterrupts
 
+private function apply_mxi_policy(m : Minterrupts) -> Minterrupts =
+  [m with
+    MSI = if plat_msi_supported then m[MSI] else 0b0,
+    MTI = if plat_mti_supported then m[MTI] else 0b0,
+    MEI = if plat_mei_supported then m[MEI] else 0b0,
+  ]
+
 private function legalize_mip(o : Minterrupts, v : xlenbits) -> Minterrupts = {
   // The only writable bits are the S-mode bits.
   let v = Mk_Minterrupts(v);
@@ -89,9 +96,9 @@ private function legalize_mie(o : Minterrupts, v : xlenbits) -> Minterrupts = {
   let v = Mk_Minterrupts(v);
   [o with
     LCOFI = if currentlyEnabled(Ext_Sscofpmf) then v[LCOFI] else 0b0,
-    MEI  = v[MEI],
-    MTI  = v[MTI],
-    MSI  = v[MSI],
+    MEI  = if plat_mei_supported then v[MEI] else 0b0,
+    MTI  = if plat_mti_supported then v[MTI] else 0b0,
+    MSI  = if plat_msi_supported then v[MSI] else 0b0,
     SEI  = if currentlyEnabled(Ext_S) then v[SEI] else 0b0,
     STI  = if currentlyEnabled(Ext_S) then v[STI] else 0b0,
     SSI  = if currentlyEnabled(Ext_S) then v[SSI] else 0b0,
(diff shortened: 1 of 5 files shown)


model/postlude/validate_config.sail
```diff
@@ -426,6 +426,27 @@ private function check_mem_layout() -> bool =
...
+    print_endline("platform.mti_supported is false but base.mideleg.delegatable_bits includes MTI (bit 7).");
+    false
+  } else true
+
+private function check_mei_config() -> bool =
+  if plat_mei_supported then true
+  else if Mk_Minterrupts(plat_mideleg_delegatable_bits)[MEI] == 0b1 then {
+    print_endline("platform.mei_supported is false but base.mideleg.delegatable_bits includes MEI (bit 11).");
+    false
+  } else true
```
jordancarlin: Let's combine this into a single `check_interrupts` function (or something similar). Then it can do `Mk_Minterrupts` once and check all three interrupts.

model/core/interrupt_regs.sail
```diff
@@ -213,16 +222,22 @@ enum XipReadType = {
...
+      )
+    )
+  )
 
 function write_mip(value : xlenbits) -> unit = {
   mip = legalize_mip(mip, value);
 }
 
-function clause read_CSR(0x304) = mie.bits
+function clause read_CSR(0x304) = apply_mei_policy(apply_mti_policy(apply_msi_policy(mie))).bits
```
jordancarlin: I don't think this is necessary because `legalize_mie` will zero out the needed bits when `mie` is written.

model/core/interrupt_regs.sail
jordancarlin, line 485: I don't think these changes are needed because `mideleg` (which is passed as `d`) should already have these bits as 0 if the interrupt is not supported.

model/core/interrupt_regs.sail
```diff
@@ -54,6 +54,15 @@ bitfield Minterrupts : xlenbits = {
...
 register hvip : Minterrupts
 
+private function apply_msi_policy(m : Minterrupts) -> Minterrupts =
+  if plat_msi_supported then m else [m with MSI = 0b0]
+
+private function apply_mti_policy(m : Minterrupts) -> Minterrupts =
+  if plat_mti_supported then m else [m with MTI = 0b0]
+
+private function apply_mei_policy(m : Minterrupts) -> Minterrupts =
+  if plat_mei_supported then m else [m with MEI = 0b0]
```
jordancarlin: It probably makes sense to make this one function since it is only ever called with all three chained together.

jordancarlin (changes requested):
Thanks for working on this. I think there are a handful of simplifications that can be made. I've commented on a few of them, but there are likely others.

model/postlude/validate_config.sail
edevolder-tt: I've combined into check_m_interrupts()

model/core/interrupt_regs.sail
edevolder-tt: legalize_mie() only forces unsupported enable bits to 0 on mie CSR writes. That does cover interrupt dispatch: getPendingSet uses read_mip(...) & mie.bits, so with mie.MSI/MTI/MEI stuck at 0 those causes never fire.

What it does not cover is mip:

- M-mode pending bits are set by the platform (CLINT / timer / external_interrupts_pending()), not by legalize_mie.
- legalize_mip() does not clear MEI/MTI/MSI.
- apply_*_policy runs in read_mip(), which is what CSR reads, write-back values, and callbacks use.

Without that, unsupported MSIP/MTIP/MEIP can still read as 1 even though they never take.

Is my undedrstanding correct?

model/core/interrupt_regs.sail
edevolder-tt: ok. my local testing confirms.

model/core/interrupt_regs.sail
edevolder-tt: ok

edevolder-tt:
I've updated the branch with the changes so far...

pmundkur:
FYI, there's some ongoing discussion on a policy for Gen AI assisted PRs for this repo.   I hope to have a resolution soon, but I don't know yet when that will be.

edevolder-tt:
I see this is blocked, but slated for 0.14. Is there something I need to do?

pmundkur:
We will most likely release 0.14 without this, but hope to fix the underlying issue soon after.

pmundkur:
@edevolder-tt  I implemented a solution in #1934 and added you as co-author while we get the AI stance sorted out.

pmundkur:
closing since the functionality was added in #1934

PR #1892, opened 2026-08-21
Add Xvisor boot

nadime15:
Add Xvisor to the os-boot Makefile so that both Linux and Xvisor can be built. The Makefile has been generalized to support building Linux, Xvisor, and in the future, additional operating systems.

Additionally, a new CI job has been added alongside the existing Linux job to build and boot Xvisor. The Linux job is updated, since the Makefile moved from os-boot/linux/ to os-boot/.

Xvisor currently boots up to the Xvisor prompt after ~21000000 instructions:
```
XVisor#
```
The `xvisor_spike` and `xvisor_qemu` targets do not work for now, which is noted in the README.

This is only the first part of the boot process. The next step is to extend the boot flow to actually launch a guest OS under Xvisor.


pmundkur (commented):
I think if you did a `git mv os-boot/linux/Makefile os-boot/Makefile`, the diff should just show the changes to the Makefile, as opposed to the current diff.

nadime15:
I tested this, this does not work. The file is almost the same but what I could do is split it into two commits, one git mv and one with the changes, then the second commit shows just the Makefile edits.

nadime15:
See https://github.com/riscv/sail-riscv/pull/1892/changes/e7af538fe75a080fdd8f907edc15a781c1aa2d3e

.github/workflows/ci.yml
```diff
@@ -128,37 +128,82 @@ jobs:
         uses: actions/cache/restore@v6
         id: linux-cache-restore
         with:
-          path: os-boot/linux/build/fw_payload.elf
-          key: linux-image-${{ hashFiles('os-boot/linux/*.url', 'os-boot/linux/Makefile') }}
+          path: os-boot/build/fw_payload_linux.elf
+          key: linux-image-${{ hashFiles('os-boot/*.url', 'os-boot/Makefile') }}
```
pmundkur: Perhaps use the specific files here, otherwise changes to xvisor will affect the pure linux cache.
```suggestion
          key: linux-image-${{ hashFiles('os-boot/linux.url', 'os-boot/opensbi.url', 'os-boot/gcc.url', 'os-boot/Makefile') }}
```

.github/workflows/ci.yml
```diff
@@ -128,37 +128,82 @@ jobs:
...
+        run: |
+          cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_LTO=TRUE
+          ninja -C build
+
+      - name: Restore cache (xvisor image)
+        uses: actions/cache/restore@v6
+        id: xvisor-cache-restore
+        with:
+          path: os-boot/build/fw_payload_xvisor.elf
+          key: xvisor-image-${{ hashFiles('os-boot/*.url', 'os-boot/Makefile') }}
```
pmundkur: Same here with hashing only the relevant files, especially if we add other OS targets.

pmundkur:
Hmm, perhaps it only works for single commit PRs.  No worries, thanks for trying it.

os-boot/Makefile
```diff
@@ -0,0 +1,124 @@
...
+#
+# Linux: the image has no userspace, so there is no 'init' and it crashes there.
+# Xvisor: the prompt appears but that's as far as it gets, the model can't read
+# from stdin so there is no way to type anything.
+limit_linux  := 20000000
+limit_xvisor := 21000000
+
+# Run with the Sail emulator from this repo.
+%_sail: build/sail.dtb build/fw_payload_%.elf $(SAIL_SIM)
+	$(SAIL_SIM) --show-times --inst-limit $(limit_$*) --device-tree-blob build/sail.dtb build/fw_payload_$*.elf
```
pmundkur: Slightly shorter
```suggestion
%_sail: build/fw_payload_%.elf build/sail.dtb $(SAIL_SIM)
	$(SAIL_SIM) --show-times --inst-limit $(limit_$*) --device-tree-blob build/sail.dtb $<
```

os-boot/Makefile
```diff
@@ -0,0 +1,124 @@
...
+#     ./configure
+#     make -j4
+#     sudo make install
+#
+#     go install github.com/google/pprof@latest
+#
+%_sail_profile: build/sail.dtb build/fw_payload_%.elf $(SAIL_SIM)
+	rm -f build/prof_$*.out
+	CPUPROFILE=build/prof_$*.out LD_PRELOAD=/usr/local/lib/libtcmalloc_and_profiler.so $(SAIL_SIM) --show-times --inst-limit $(limit_$*) --device-tree-blob build/sail.dtb build/fw_payload_$*.elf
+	pprof -http : build/prof_$*.out
```
pmundkur: Same here, slightly shorter
```suggestion
%_sail_profile: build/fw_payload_%.elf build/sail.dtb $(SAIL_SIM)
	rm -f build/prof_$*.out
	CPUPROFILE=build/prof_$*.out LD_PRELOAD=/usr/local/lib/libtcmalloc_and_profiler.so $(SAIL_SIM) --show-times --inst-limit $(limit_$*) --device-tree-blob build/sail.dtb $<
	pprof -http : build/prof_$*.out
```

os-boot/Makefile
```diff
@@ -0,0 +1,124 @@
...
+
+# Run with Spike: https://github.com/riscv-software-src/riscv-isa-sim
+%_spike: build/fw_payload_%.elf
+	spike --instructions=$(limit_$*) build/fw_payload_$*.elf
```
pmundkur:
```suggestion
	spike --instructions=$(limit_$*) $<
```

os-boot/Makefile
```diff
@@ -0,0 +1,124 @@
...
+
+# Run with QEMU.
+%_qemu: build/fw_payload_%.elf
+	qemu-system-riscv64 -M virt -m 256M -nographic -bios build/fw_payload_$*.elf
```
pmundkur:
```suggestion
	qemu-system-riscv64 -M virt -m 256M -nographic -bios $<
```

pmundkur (commented):
Minor suggestions, but otherwise LGTM.

nadime15:
I addressed your comments from above, but I noticed that Xvisor does not work on Spike/QEMU. I managed to figure out and fix the issue with Spike. We need to provide the generated DTB, since Spike enables extensions based on which ones are available in the device tree. H is disabled by default (and some other mismatches between Spikes default DTB and ours).

QEMU issue, seems a bit more complicated and non trivial. I am not sure what is causing the problem with QEMU at this point. It would take some time to debug.

So the question is now how to move forward? Should I at least commit the Spike patch and exclude the xvisor_qemu rule? Or find the QEMU issue before we merge this and provide a fix for QEMU/Spike? Or maybe none of that and?

pmundkur:
The Spike patch seems good for now.  We could document the QEMU issue in the README but leave the rule in the Makefile.  I doubt these targets were used much.

jordancarlin:
> We need to provide the generated DTB, since Spike enables extensions based on which ones are available in the device tree.

Could you just specify the `--isa` flag to Spike instead? That seems like a simpler fix.

Agreed that we shouldn’t block this on the QEMU issue since that is really just a convenience target.

nadime15:
That's unfortunately not enough, because Spike's default device tree is different from ours. I'm not exactly sure what's causing the issue. I would need to debug this as well. Even during the hypervisor work, I didn't look into it because it was enough to simply provide the Sail DTB. I think it might be related to the chosen node and its bootargs, or something interrupt-related.

jordancarlin:
Strange. Fine with me to use the Sail DTB then.

________________________________

os-boot/README.md
```diff
@@ -32,3 +32,16 @@ since console input is not supported and there is no way to type anything.
...
 and so on.
+
+### Known issues
+
+`xvisor_spike` and `xvisor_qemu` do not currently work.
+
+Spike's default device tree does not match the Sail platform, and passing the
+model's one with `--dtb` instead fails because Spike reads the ISA from it and
+rejects the extensions it does not implement, starting with `zic64b`. Limiting
+the extensions in the device tree to what Spike actually supports makes it work.
```
pmundkur, line 43: This is an example of #993.

pmundkur (approved):

nadime15:
Ok, I realized that passing sails default DTB to spike does not work either because we support more extensions than spike does, so spike will reject the DTB. It would be fairly easy to define a new rule that trims our DTS and removes the unsupported extensions, but for now I just added a note to the README saying that both rules (xvisor_spike and xvisor_qemu) dont work at the moment.

os-boot/README.md
nadime15: Ah yeah, good point.

nadime15:
Can we merge this one?

PR #1890, opened 2026-08-20
Add the XUANTIE-RV/damo-rv-priv-ats hypervisor tests

nadime15:
Update to the latest test suite release (2026-08-20), which includes the new XUANTIE-RV/damo-rv-priv-ats test suite.

Sail RISC-V Tests - 2026-08-20

Built with:
riscv-software-src/sail-riscv-tests: dcb4a38
riscv-tests: 1eb47d946c55f55cab8653c224c2993acc0276bd
riscv-vector-tests: f76bff121dce91b6b23e1d37be77a5e44af914c3
riscv-arch-test: db2b9fd7aa71a6e669417f711379988541f1464c
damo-rv-priv-ats: 3d7b83b79799f57f812da13b2a21b055e79b6c27
Spike: 20feb9c2bf2a7deab964d8190b0cbd4b4131bec3
Toolchain release: 2026.07.15
Sail RISC-V release: 0.13.1

```
$ ./compare-releases.py -p releases/2026-06-10 -c releases/2026-08-20
riscv-tests has 667 test files, with 0 added to and 0 removed from the previous release.`
riscv-arch-tests has 3815 test files, with 1813 added to and 78 removed from the previous release.
damo-tests with 43 test files was added.
riscv-vector-tests-v64x32 has 1941 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v64x64 has 3025 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v128x32 has 1944 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v128x64 has 3042 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v256x32 has 1945 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v256x64 has 3043 test files, with 0 added to and 0 removed from the previous release.
riscv-vector-tests-v512x32 is not present in both releases.
riscv-vector-tests-v512x64 has 3043 test files, with 0 added to and 0 removed from the previous release.
```

PR #1899, opened 2026-08-24
Move to a pull approach for interrupt detection.

pmundkur:
This removes `clint_dispatch()`, which pushed (i.e. wrote) updates to `mip` when memory-mapped registers were updated. This is replaced by `update_mip()`, which updates `mip` to incorporate changes from (i.e. reads) these registers before pending interrupts are computed for dispatch.  

This defines an interface for interrupt controllers in `core/interrupt_interface.sail`, and implements this interface for the CLINT and the simple interrupt generator (SIG).

This required moving the `{s,vs}timecmp` CSRs to `interrupt_regs`, and explicitly introducing the CLINT's `msip`, which was previously implicit in the push-oriented approach.  The `mtime` and `mtimecmp` registers are kept private to the CLINT, to make it easier to isolate later as a separate component. Similar `sig_msip` and `sig_ssip` internal state registers were introduced for the SIG.

This aligns the handling of CLINT's interrupts with the handling of those from the `simple_interrupt_generator`.  The handling of external interrupts using the `simple_interrupt_generator` is unchanged.

In addition,
- the Zicntr extension now requires a source of time, for which the only provider currently is the CLINT
- the CLINT is announced in the device tree only if it is enabled
- the CLINT register arrays are indexed by the `mhartid`, instead of always using the 0'th entry.


pmundkur:
This should wait for #1890 and #1892 to be merged for adequate testing.

model/sys/sys_control.sail
```diff
@@ -80,6 +80,50 @@ function exception_delegatee(e : ExceptionType, p : Privilege) -> Privilege = {
   then deleg else p
 }
 
+private function update_mip() -> unit = {
```
nadime15: I haven’t fully looked into the code yet, but why isn’t all of the code below part of read_mip()?

model/sys/sys_control.sail
nadime15, line 172: Need to add update_mip() here too?
```suggestion
function shouldWakeForInterrupt() -> bool = {
update_mip();
(read_mip(IncludePlatformInterrupts).bits & mie.bits) != zeros()
```

model/sys/sys_control.sail
pmundkur: The `{m,s,vs}timecmp` CSRs are currently not visible to `read_mip`.   They will eventually need to move out of `interrupt_regs` anyway and depending on how we refactor, `update_mip` could then be merged into `read_mip`.

model/sys/sys_control.sail
pmundkur: Good catch, added.

model/sys/sys_control.sail
pmundkur: Actually, only `mtimecmp` (and `msip`) will need to move out.

pmundkur:
It looks like this broke a couple of DAMO tests.  I'll investigate ...

model/sys/sys_control.sail
```diff
@@ -80,6 +80,50 @@ function exception_delegatee(e : ExceptionType, p : Privilege) -> Privilege = {
...
+  // CLINT's `msip` register.  `msip` is zero if the CLINT is not enabled.
+  mip[MSI] = msip[0];
+  mip[MTI] = bool_to_bit(mtimecmp <=_u mtime);
+
+  if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then {
+    mip[STI] = bool_to_bit(stimecmp <=_u mtime);
+  };
+
+  if currentlyEnabled(Ext_H) then {
+    mip[VSSI] = hvip[VSSI];
```
nadime15: I think the problem is the following line, can you try to remove it and run it again?
```suggestion

```
`hip.VSSIP`, `vsip.SSIP` and `hvip.VSSIP` are three names for the same bit. The bit is stored in `mip[VSSI]`. 

Writes through `hip` and `vsip` only update `mip[VSSI]`, they never write `hvip`. So `hvip[VSSI]` stays 0, and `update_mip()` copies that 0 over the 1 before the interrupt is checked.

pmundkur: Indeed, that fixed it, thanks! I'll refactor this a bit tomorrow.

model/sys/platform.sail
```diff
@@ -54,16 +54,33 @@ private function within_htif_readable forall 'n, 0 < 'n <= max_mem_access . (add
...
-register vstimecmp : bits(64)
+// Each hart has a memory-mapped `mtimecmp` register.
+private register mtimecmp : bits(64)
+
+// This is a platform-wide memory-mapped register that counts the
+// number of cycles from the system clock.
+private register mtime : bits(64)
+
+function get_mtime() -> bits(64) =
+  if plat_have_clint then mtime else zeros()
```
nadime15: Thats not needed and would also return 0 if mtime is read directly without clint.
```suggestion
function get_mtime() -> bits(64) = mtime
```

model/core/interrupt_regs.sail
```diff
@@ -218,6 +202,60 @@ function read_mip(read_type : XipReadType) -> Minterrupts =
...
+
+  let cur_mtime = get_mtime();
+  let hart_mtimecmp = get_hart_mtimecmp();
+
+  mip[MSI] = get_machine_software_interrupt_pending();
+  mip[MTI] = bool_to_bit(platform_has_timer & (hart_mtimecmp <=_u cur_mtime));
+
+  if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then {
+    mip[STI] = bool_to_bit(platform_has_timer & (stimecmp <=_u cur_mtime));
+  };
```
```suggestion
   if platform_has_timer then {
    mip[MTI] = bool_to_bit(hart_mtimecmp <=_u cur_mtime);
    if currentlyEnabled(Ext_Sstc) & menvcfg[STCE] == 0b1 then {
      mip[STI] = bool_to_bit(stimecmp <=_u cur_mtime);
    };
  };
```

model/core/interrupt_regs.sail
```diff
@@ -218,6 +202,60 @@ function read_mip(read_type : XipReadType) -> Minterrupts =
...
+    // NOTE: If the result of this comparison changes, it is guaranteed to be reflected in VSTIP eventually,
+    // but not necessarily immediately. The interrupt remains posted until vstimecmp becomes greater
+    // than (time + htimedelta), typically as a result of writing vstimecmp. The interrupt will be
+    // taken based on the standard interrupt enable and delegation rules while V=1.
+    var VSTI : bits(1) = hvip[VSTI];
+
+    if currentlyEnabled(Ext_Sstc) & read_henvcfg()[STCE] == 0b1 then {
+      VSTI = VSTI | bool_to_bit(platform_has_timer & (vstimecmp <=_u (cur_mtime + htimedelta)));
+    };
+    mip[VSTI] = VSTI;
```
nadime15: Maybe like this (up to you)? 
```suggestion
    mip[VSTI] = hvip[VSTI] | bool_to_bit(
        currentlyEnabled(Ext_Sstc) & read_henvcfg()[STCE] == 0b1
        & platform_has_timer & (vstimecmp <=_u (cur_mtime + htimedelta))
     );
```

model/sys/platform.sail
```diff
@@ -141,81 +158,44 @@ function clint_load(access, paddr, width) = {
...
 function clint_store(paddr, width, data) = {
   let Physaddr(addr) = paddr;
   let addr = addr - plat_clint_base;
   if addr == MSIP_BASE & ('n == 8 | 'n == 4) then {
     if   get_config_print_clint()
-    then print_log("clint[" ^ bits_str(addr) ^ "] <- " ^ bits_str(data) ^ " (mip.MSI <- " ^ bits_str(data[0..0]) ^ ")");
-    mip[MSI] = [data[0]];
-    clint_dispatch(true);
+    then print_log("clint[" ^ bits_str(addr) ^ "] <- " ^ bits_str(data) ^ " (msip <- " ^ bits_str(data) ^ ")");
+    msip = data[31 .. 0];
```
nadime15: Shouldn this be the following?
```suggestion
    then print_log("clint[" ^ bits_str(addr) ^ "] <- " ^ bits_str(data) ^ " (msip <- " ^ bits_str(data) ^ ")");
    msip[0] = data[0];
```

Only the least significant bit is writable.

model/core/interrupt_regs.sail
```diff
@@ -218,6 +202,60 @@ function read_mip(read_type : XipReadType) -> Minterrupts =
...
+  // so without a CLINT, no non-virtual timer interrupts can fire.
+  // This may need to change when we support other or multiple
+  // interrupt controllers.
+
+  let platform_has_timer = plat_have_clint;
+
+  let cur_mtime = get_mtime();
+  let hart_mtimecmp = get_hart_mtimecmp();
+
+  mip[MSI] = get_machine_software_interrupt_pending();
```
nadime15: The write to mip[MSI] is lost. The SIG writes to it directly and the line below resets it constantly.

I think the generator should handle this similarly to `external_interrupts_pending()` and MEI.

So we would define the following (sorry, I’m just describing it here instead of at the corresponding lines):

`private register sig_msip : bits(1) = 0b0`

add `MSI = sig_msip` to external_interrupts_pending(), and then `update sig_store` from:

`if data[MSI] == 0b1 then mip[MSI] = value;`

to:

`if data[MSI] == 0b1 then sig_msip = value;`

model/sys/platform.sail
pmundkur: `tick_clock()` increments `mtime` unconditionally, and so the `plat_have_clint` check needs to be there.  I think it makes the CLINT module more self-contained to have this check here, in addition to the `platform_has_timer` in `interrupt_regs`, even though the two in combination are redundant.  Another option is to have `tick_clock()` be called only for interrupt controllers that are enabled; either way seems reasonable to me.

model/core/interrupt_regs.sail
pmundkur: Hmm, I prefer the original version.

model/core/interrupt_regs.sail
pmundkur: Good point, so we already have to deal with multiple controllers generating the same interrupt.  I added the `sig_msip`, and an `_implementation` file that accumulates across the two controllers.

model/sys/platform.sail
nadime15: What I mean is there is a mtime clause in Zicntr which reads the `time` CSR through `get_mtime()`, so with `platform.clint.supported = false` and Zicntr enabled, `time` reads zero forever while `clint_tick_clock()` keeps incrementing `mtime`.

Before this PR `read_CSR(0xC01)` read `mtime` directly and counted either way, and nothing in `validate_config` requires a CLINT for Zicntr.

> Another option is to have tick_clock() be called only for interrupt controllers that are enabled; either way seems reasonable to me.

I may be misunderstanding this, but this would not work if `tick_clock()` only runs for enabled controllers, since mtime would never advance at all.

pmundkur: So one perhaps approach is to require `Zicntr` to depend on a source of time, currently the CLINT?

nadime15: Ok so this mean we add a check in validate_config that Zicntr requires platform.clint.supported, until there is another source of time?

pmundkur: Yes.  I think this is making explicit what was already implicit (i.e. this was a missed check in `validate_config`), and I think a source of time is implicitly required by the spec:
> The RDTIME pseudoinstruction reads the low XLEN bits of the time CSR, which counts wall-clock real time that has passed from an arbitrary start time in the past.

model/sys/simple_interrupt_generator.sail
```diff
@@ -47,7 +47,7 @@ function sig_store(addr, width, data) = {
 
       if data[MEI] == 0b1 then sig_meip = value;
       if data[SEI] == 0b1 then sig_seip = value;
-      if data[MSI] == 0b1 then mip[MSI] = value;
+      if data[MSI] == 0b1 then sig_msip = value;
       // SSI is read-only zero if the hart does not support supervisor mode.
       if data[SSI] == 0b1 & currentlyEnabled(Ext_S) then mip[SSI] = value;
```
jordancarlin: Should SSI be changed too so that we aren't writing directly to `mip` anywhere from the SIG?

model/core/interrupt_regs.sail
```diff
@@ -218,6 +202,63 @@ function read_mip(read_type : XipReadType) -> Minterrupts =
...
+    // NOTE: If the result of this comparison changes, it is guaranteed to be reflected in VSTIP eventually,
+    // but not necessarily immediately. The interrupt remains posted until vstimecmp becomes greater
+    // than (time + htimedelta), typically as a result of writing vstimecmp. The interrupt will be
+    // taken based on the standard interrupt enable and delegation rules while V=1.
+    var VSTI : bits(1) = hvip[VSTI];
+
+    if currentlyEnabled(Ext_Sstc) & read_henvcfg()[STCE] == 0b1 then {
+      VSTI = VSTI | bool_to_bit(platform_has_timer & (vstimecmp <=_u (cur_mtime + htimedelta)));
+    };
+    mip[VSTI] = VSTI;
```
jordancarlin: I'd rather avoid the `var` here. What about something like this:

```suggestion
    if currentlyEnabled(Ext_Sstc) & read_henvcfg()[STCE] == 0b1 then {
      let VSTI_comparator = bool_to_bit(platform_has_timer & (vstimecmp <=_u (cur_mtime + htimedelta)));
    }
    else {
      let VSTI_comparator = 0b0
    };
    mip[VSTI] = VSTI_comparator | hvip[VSTI];
```

model/sys/platform.sail
```diff
@@ -54,27 +56,53 @@ private function within_htif_readable forall 'n, 0 < 'n <= max_mem_access . (add
...
+// Each `msip` register is a 32-bit wide register where the upper 31
+// bits are tied to 0. The least significant bit is reflected in the
+// `MSIP` bit of the `mip` CSR. Other bits in the `msip` registers are
+// hardwired to zero. On reset, each `msip` register is cleared to zero.
+private register msip : bits(32) = zeros()
 
-// vstimecmp is a real CSR; not memory mapped.
-register vstimecmp : bits(64)
+// Each hart has a memory-mapped `mtimecmp` register.
+private register mtimecmp : bits(64)
```
jordancarlin, line 75: Since we are refactoring this anyway and plan to eventually pull it into a separate model, how about moving a lot of this over to a new `clint.sail` file (still part of the `sys` module for now)?

model/sys/platform.sail
```diff
@@ -54,27 +56,53 @@ private function within_htif_readable forall 'n, 0 < 'n <= max_mem_access . (add
...
+
+// This is a platform-wide memory-mapped register that counts the
+// number of cycles from the system clock.
+private register mtime : bits(64)
+
+function clint_get_mtime() -> bits(64) =
+  if plat_have_clint then mtime else zeros()
```
jordancarlin: Seems like it would make more sense to only call this function if `plat_have_clint` is true and have it unconditionally return `mtime`.

jordancarlin: Applies to all of the clint_get functions.

model/sys/simple_interrupt_generator.sail
pmundkur: Good point, I missed that.

model/sys/platform.sail
pmundkur: I left that for later, to reduce the size of the diff for this PR.

model/core/interrupt_regs.sail
nadime15: or https://github.com/riscv/sail-riscv/pull/1899#discussion_r3881413639 :)

model/sys/platform.sail
jordancarlin: Fine with me. Seemed like enough of this code was being touched that it was worth considering, but happy to leave it for another PR if you prefer.

model/core/interrupt_regs.sail
jordancarlin: Missed that comment. That approach would work too (and is more concise).

pmundkur: That doesn't work; the `let` bindings are scoped to inside the `if`/`else` blocks, and are lost after the `if/else` statement.  I find the `var` cleaner for incrementally updating a value, like in `valid` in `validate_config()`.

jordancarlin: If the `let` approach doesn't work, I would prefer @nadime15's approach. We aren't really incrementally updating a value here. We are just combining two different sources together.

pmundkur: I think the `var` is easier to align with the spec:

> VSTIP is read-only in hip, and is the logical-OR of hvip.⁠VSTIP and, when the Sstc extension is implemented, the timer interrupt signal resulting from vstimecmp.

It's easier to read the `and, when the Sstc extension is implemented` as a separate `if`.  @nadime15's approach buries a big expression inside a `bool_to_bit()` which is arguably harder to read.

jordancarlin: I still prefer the other approach, but I'm ok with this if you feel strongly about it and would rather merge this as is.

jordancarlin (approved):

model/core/interrupt_regs.sail
nadime15: Yeah I also prefer https://github.com/riscv/sail-riscv/pull/1899#discussion_r3881413639 (which I think is easier to read and understand)

model/postlude/validate_config.sail
```diff
@@ -752,10 +752,25 @@ private function check_stateen_config() -> bool = {
...
+      valid = false;
+      print_endline("Zicntr is enabled but there is no source of time: a source of time (currently only `platform.clint.supported`) must be enabled.");
+    };
+  };
+  if (config platform.clint.supported : bool) then {
+    let hartid : int = config platform.hartid;
+    if hartid >= 4095 then {
+      // This is a borderline case: the hart will not receive interrupts, but otherwise the configuration is valid,
+      // so print a warning.
+      print_endline("Warning: platform.mhartid is " ^ dec_str(hartid) ^ ", which is out-of-range for the CLINT, which serves harts 0 through 4094; the hart will not receive interrupts from the CLINT.");
```
```suggestion
      print_endline("Warning: platform.hartid is " ^ dec_str(hartid) ^ ", which is out-of-range for the CLINT, which serves harts 0 through 4094; the hart will not receive interrupts from the CLINT.");
```

nadime15 (commented):
Looks good to me! I think that doc/SimpleInterruptGenerator.md needs to be updated but I can do this too later if you like. 

It says things like 

> MSI updates `mip[MSI]` directly

 or 

> Note that if the target hart does not support supervisor mode then mip[SSI] and mip[SEI] must be read-only zero. Attempts to set mip[SSI] will be ignored. 

 (which is not true anymore because its only masked and visible again once S-Mode is renabled )

pmundkur:
> Looks good to me! I think that doc/SimpleInterruptGenerator.md needs to be updated but I can do this too later if you like.
> 

No worries,, I can do it, the reading guide needs to be updated for the new files as well.

pmundkur:
I think I had implemented SSI incorrectly; I fixed it.  I'll need to rebase and sign to fix DCO.

model/core/interrupt_regs.sail
```diff
@@ -219,6 +202,67 @@ function read_mip(read_type : XipReadType) -> Minterrupts =
...
+  let platform_has_timer = plat_have_clint;
+
+  let cur_mtime = get_mtime();
+  let hart_mtimecmp = get_hart_mtimecmp();
+
+  mip[MSI] = machine_software_interrupt_pending();
+
+  if currentlyEnabled(Ext_S) then {
+    mip[SSI] = supervisor_software_interrupt_pending();
+  };
```
```suggestion

```

model/sys/simple_interrupt_generator_regs.sail
```diff
@@ -7,12 +7,20 @@
...
 private register sig_meip : bits(1) = 0b0
 private register sig_seip : bits(1) = 0b0
+private register sig_ssip : bits(1) = 0b0
 
-function external_interrupts_pending() -> Minterrupts =
+function sig_external_interrupts_pending() -> Minterrupts =
   [Mk_Minterrupts(zeros()) with
     MEI = sig_meip,
     // SEIP is read-only zero if the hart does not support supervisor mode.
     SEI = if currentlyEnabled(Ext_S) then sig_seip else 0b0,
```
nadime15, line 19:
```suggestion
    SEI = if currentlyEnabled(Ext_S) then sig_seip else 0b0,
    SSI = if currentlyEnabled(Ext_S) then sig_ssip else 0b0,
```

model/sys/simple_interrupt_generator_regs.sail
```diff
@@ -7,12 +7,20 @@
...
   ]
+
+function sig_machine_software_interrupt_pending() -> bits(1) =
+  sig_msip
+
+function sig_supervisor_software_interrupt_pending() -> bits(1) =
```
nadime15, line 25:
```suggestion

```

nadime15 (commented):
Commit f8fb157162c3fb888e74688cf0043443886e7d58 leads to a regression... This needs to be undone, mip.MSI cannot be written by software, whereas mip.SSI can (this is the same issue we found earlier with mip[VSSI] = hvip[VSSI]).

This approach only works for read-only bits, you need to have them ORed in on read.

model/core/interrupt_regs.sail
nadime15: This has not been addressed? Seems like that Jordan and I actually prefer the alternative https://github.com/riscv/sail-riscv/pull/1899#discussion_r3881413639

model/core/interrupt_regs.sail
pmundkur: Good point.  I was preserving existing functionality, which seems to have been wrong:
https://github.com/riscv/sail-riscv/blob/803f192b5906e39eeff2d31269d2e88a3baf7326/model/sys/simple_interrupt_generator.sail#L52

model/sys/simple_interrupt_generator_regs.sail
pmundkur: I'd rather not do this, since adding SSI to a function called `sig_external_inerrupts_pending()` seems misleading.  I'll keep the additional function and OR it in.

model/core/interrupt_regs.sail
pmundkur: What do you think of https://github.com/riscv/sail-riscv/commit/e6521cc26f550995a9d628778eeb983f0f0a144d

model/core/interrupt_regs.sail
```diff
@@ -219,6 +202,65 @@ function read_mip(read_type : XipReadType) -> Minterrupts =
...
+  let platform_has_timer = plat_have_clint;
+
+  let cur_mtime = get_mtime();
+  let hart_mtimecmp = get_hart_mtimecmp();
+
+  mip[MSI] = machine_software_interrupt_pending();
+
+  if currentlyEnabled(Ext_S) then {
+    mip[SSI] = mip[SSI] | supervisor_software_interrupt_pending();
+  };
```
nadime15: This does not work, you wont be able to clear the bit anymore.

If supervisor_software_interrupt_pending() returns 1, mip[SSI] is 1 now. Next cycle you write 0 into the SIG and now supervisor_software_interrupt_pending() returns 0 but because mip[SSI] is 1 (from the previous cycle) you keep setting mip[SSI] and now its pretty much deadlocked.

```suggestion
```

model/core/interrupt_regs.sail
nadime15: Maybe something like this?
```suggestion
     IncludePlatformInterrupts => {
        let ssi = [Mk_Minterrupts(zeros()) with
          SSI = if currentlyEnabled(Ext_S) then supervisor_software_interrupt_pending() else 0b0
        ];
        Mk_Minterrupts(mip.bits | external_interrupts_pending().bits | ssi.bits)
      },
```

pmundkur: Yeah, this sounds right.

pmundkur: See https://github.com/riscv/sail-riscv/pull/1899/commits/db732523d6fb6fbf0b27c2c9dddd92a04982f676

nadime15 (approved):

pmundkur:
@jordancarlin there were some changes since you approved, so perhaps you'd like to review again?

pmundkur:
I'll merge this soon (in around 30 minutes) since I think #1931 and #1934 would need to slightly update to adapt.

jordancarlin (approved):

PR #1905, opened 2026-08-25
Adapt TLB trace log to whether the hypervisor has been configured.

pmundkur:
This just puts the TLB printing into two separate functions for the two cases.

### DIFF
--- a/c_emulator/riscv_callbacks_log.cpp
+++ b/c_emulator/riscv_callbacks_log.cpp
@@ -166,7 +166,7 @@ void log_callbacks::ptw_fail_callback(
 
 namespace {
 
-void print_tlb(
+void print_hypervisor_tlb(
   FILE *trace_log,
   ModelImpl &model,
   ModelImpl::TLB tlb,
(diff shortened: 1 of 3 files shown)


c_emulator/riscv_callbacks_log.cpp
```diff
@@ -244,6 +298,10 @@ void log_callbacks::tlb_flush_callback(ModelImpl &, uint64_t index) {
 
 void log_callbacks::tlb_flush_end_callback(ModelImpl &model, ModelImpl::TLB tlb) {
   if (trace_log != nullptr && config_print_tlb && !pending_flush_indices.empty()) {
-    print_tlb(trace_log, model, tlb, pending_flush_indices, true);
+    if (model.supports_hypervisor()) {
+      print_hypervisor_tlb(trace_log, model, tlb, pending_flush_indices, true);
+    } else {
+      print_tlb(trace_log, tlb, pending_flush_indices, true);
+    }
```
nadime15: Just as an idea (I don’t mind either way), you could define print_tlb as follows:

```
void print_tlb(
    FILE *trace_log,
    ModelImpl &model,
    ModelImpl::TLB tlb,
    const std::vector<uint64_t> &indices,
    bool is_flush
  ) {
    if (model.supports_hypervisor()) {
      print_hypervisor_tlb(trace_log, model, tlb, indices, is_flush);
    } else {
      print_supervisor_tlb(trace_log, tlb, indices, is_flush);
    }
  }
```

Which basically just dispatches the calls (avoiding duplication of the dispatch logic), while keeping `tlb_flush_end_callback` and `tlb_add_callback` as they are.

nadime15 (approved):

jordancarlin (approved):

PR #1894, opened 2026-08-21
Make extensions.Svbare.supported actually take effect

nadime15:
currentlyEnabled(Ext_Svbare) ignored hartSupports, so satp always accepted Bare. Also reject a config with supervisor mode but no satp mode.

PR #1906, opened 2026-08-25
Add the Shvsatpa extension and vsatp mode config options

nadime15:
Adds Shvsatpa along with `extensions.H.vsatp_modes`, mirroring the `hgatp` mode options from #1893.

We have `satp` (Svnn), `vsatp` (Svnn), and `hgatp` (SvXnn).

The translation modes for satp enforce a dependency. For example, supporting Sv57 without supporting Sv48 is not possible. I can't find anywhere in the spec that explicitly says whether this dependency also applies to `vsatp`, but I'm pretty sure the same rules apply there.

`hgatp` however, is a bit strange in this regard, because it is allowed to support Sv57x4 without supporting Sv48x4, etc.

On top of that, both `vsatp` and `hgatp` need to support Bare, since that is their reset value and they are WARL.

So `vsatp` gets the same ordering checks as `satp`, and Bare stays unconditional for `vsatp`. Keeping Bare unconditional also fixes a regression from #1894. `legalize_satp` is shared between `satp` and `vsatp`, so gating Bare on Svbare applied to `vsatp` too, which according to the spec is incorrect since `vsatp.MODE` (and `hgatp.MODE`) resets to 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ An experimental emulator in the Lean language is available, see its
 - Shcounterenw extension for writable `hcounteren` enables for any supported counter, v1.0
 - Shgatpa extension for `hgatp` mode support matching `satp`, v1.0
 - Shtvala extension for Trap Value Reporting, v1.0
+- Shvsatpa extension for `vsatp` supporting all `satp` translation modes, v1.0
 - Shvstvala extension for Trap Value Reporting, v1.0
 - Shvstvecd extension for Direct Trap Vectoring, v1.0
 - Physical Memory Protection (PMP)
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ An experimental emulator in the Lean language is available, see its
 - Shcounterenw extension for writable `hcounteren` enables for any supported counter, v1.0
 - Shgatpa extension for `hgatp` mode support matching `satp`, v1.0
 - Shtvala extension for Trap Value Reporting, v1.0
-- Shvsatpa extension for `vsatp` supporting all `satp` translation modes, v1.0
+- Shvsatpa extension for Translation Mode Support, v1.0
 - Shvstvala extension for Trap Value Reporting, v1.0
 - Shvstvecd extension for Direct Trap Vectoring, v1.0
 - Physical Memory Protection (PMP)
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -1116,6 +1116,8 @@ function legalize_satp(
     let s = if is_vsatp then s else [s with
       PPN = zero_extend(s[PPN][min(physaddr_bits - pagesize_bits, 44) - 1 .. 0]),
     ];
+    // NOTE: Sv32 is not listed because VSXLEN=XLEN here, hstatus.VSXL is
+    // read-only. It could be added if changing VSXL is supported.
     let sv39_supported = currentlyEnabled(Ext_Sv39) & (not(is_vsatp) | sys_vsatp_sv39_supported);
     let sv48_supported = currentlyEnabled(Ext_Sv48) & (not(is_vsatp) | sys_vsatp_sv48_supported);
     let sv57_supported = currentlyEnabled(Ext_Sv57) & (not(is_vsatp) | sys_vsatp_sv57_supported);
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -96,6 +96,14 @@ let sys_hgatp_sv39x4_supported : bool = config extensions.H.hgatp_modes.Sv39x4
 let sys_hgatp_sv48x4_supported : bool = config extensions.H.hgatp_modes.Sv48x4
 let sys_hgatp_sv57x4_supported : bool = config extensions.H.hgatp_modes.Sv57x4
 
+// Which VS-stage translation modes vsatp supports. Bare is always supported.
+// Shvsatpa requires vsatp to support every mode supported in satp, which
+// validate_config() checks.
+let sys_vsatp_sv32_supported : bool = config extensions.H.vsatp_modes.Sv32
+let sys_vsatp_sv39_supported : bool = config extensions.H.vsatp_modes.Sv39
+let sys_vsatp_sv48_supported : bool = config extensions.H.vsatp_modes.Sv48
+let sys_vsatp_sv57_supported : bool = config extensions.H.vsatp_modes.Sv57
+
 // Whether a guest-page fault writes the faulting guest physical address to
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -221,6 +221,7 @@ An experimental emulator in the Lean language is available, see its
 - Svvptc extension for Obviating Memory-management Instructions after Marking PTEs valid, v1.0
 - Smmpm, Smnpm, Ssnpm, Sspm and Supm extensions for pointer masking, v1.0
 - H extension for Hypervisor Support, v1.0
+- Shvsatpa extension for `vsatp` supporting all `satp` translation modes, v1.0
```
pmundkur: The language of the spec is
```
Shvsatpa Extension for Translation Mode Support
```

model/postlude/validate_config.sail
```diff
@@ -78,6 +86,26 @@ private function check_mmu_config() -> bool = {
...
+      if sys_vsatp_sv48_supported & not(sys_vsatp_sv39_supported)
+      then {
+        valid = false;
+        print_endline("Sv48 is enabled for `vsatp` but Sv39 is disabled: supporting Sv48 requires supporting Sv39.");
+      };
+      if sys_vsatp_sv32_supported
+      then {
+        valid = false;
+        print_endline("Sv32 is enabled for `vsatp`: Sv32 is not supported on RV64.");
+      };
```
pmundkur, line 130: Perhaps leave a note here saying this would need changing to support RV32 guests on RV64.

model/core/sys_regs.sail
```diff
@@ -1082,11 +1090,13 @@ function legalize_satp(
...
       PPN = zero_extend(s[PPN][physaddr_bits - pagesize_bits - 1 .. 0]),
     ];
+    let sv32_supported = currentlyEnabled(Ext_Sv32) & (not(is_vsatp) | sys_vsatp_sv32_supported);
     match satpMode_of_bits(arch, 0b000 @ s[Mode]) {
       None()  => prev_value,
       Some(Sv_mode) => match Sv_mode {
-        Bare if currentlyEnabled(Ext_Svbare) => s.bits,
-        Sv32 if currentlyEnabled(Ext_Sv32) => s.bits,
+        // vsatp.MODE resets to 0, so Bare is always supported there.
+        Bare if is_vsatp | currentlyEnabled(Ext_Svbare) => s.bits,
```
pmundkur, line 1107: I'm missing this in the spec: is `Bare` always supported in `vsatp`?

model/core/sys_regs.sail
```diff
@@ -1097,13 +1107,17 @@ function legalize_satp(
...
       Some(Sv_mode) =>  match Sv_mode {
-        Bare if currentlyEnabled(Ext_Svbare) => s.bits,
-        Sv39 if currentlyEnabled(Ext_Sv39) => s.bits,
-        Sv48 if currentlyEnabled(Ext_Sv48) => s.bits,
-        Sv57 if currentlyEnabled(Ext_Sv57) => s.bits,
+        // vsatp.MODE resets to 0, so Bare is always supported there.
+        Bare if is_vsatp | currentlyEnabled(Ext_Svbare) => s.bits,
+        Sv39 if sv39_supported => s.bits,
+        Sv48 if sv48_supported => s.bits,
+        Sv57 if sv57_supported => s.bits,
```
pmundkur, line 1131: Perhaps leave a note here saying if changing `hstatus.VSXL` is supported, this could include Sv32.

model/core/sys_regs.sail
nadime15: That’s my conclusion. Otherwise, you couldn’t reset vsatp.MODE (and hgatp.MODE) to 0, even though this is mandatory. They are also WARL fields, which are not allowed to contain illegal values.

I also think the reason why satp.MODE is allowed to not support Bare is that a chip could force supervisor code to always go through address translation. This is already somewhat enforced by two-stage address translation, since a guest always has to go through G-stage translation, meaning that HS-code can enforce translation.

Also, omitting vsatp.MODE = Bare would create unnecessary complexity. If a guest writes vsatp.MODE = Bare because it does not want address translation, you would suddenly have to trap every access to it (making use of hstatus.VTVM = 1, which denies VS-mode access to satp). You would then have to create an identity mapping so that GVA = GPA and point vsatp at that instead.

So you would pay all these costs for essentially nothing (unless I miss something here).

model/postlude/validate_config.sail
```diff
@@ -99,6 +107,28 @@ private function check_mmu_config() -> bool = {
...
+      then {
+        valid = false;
+        print_endline("Sv48 is enabled for `vsatp` but Sv39 is disabled: supporting Sv48 requires supporting Sv39.");
+      };
+      // NOTE: Sv32 for `vsatp` needs VSXLEN=32, which needs a writable
+      // `hstatus.VSXL`. Revisit this if RV32 guests on RV64 are supported.
+      if sys_vsatp_sv32_supported
+      then {
+        valid = false;
+        print_endline("Sv32 is enabled for `vsatp`: Sv32 is not supported on RV64.");
```
pmundkur: Perhaps we should say this to be more accurate:
```suggestion
        print_endline("Sv32 is enabled for `vsatp`: Sv32 is not currently supported on RV64.");
```

pmundkur (approved):

model/core/sys_regs.sail
```diff
@@ -1103,13 +1113,19 @@ function legalize_satp(
     let s = if is_vsatp then s else [s with
       PPN = zero_extend(s[PPN][min(physaddr_bits - pagesize_bits, 44) - 1 .. 0]),
     ];
+    // NOTE: Sv32 is not listed because VSXLEN=XLEN here, hstatus.VSXL is
+    // read-only. It could be added if changing VSXL is supported.
+    let sv39_supported = currentlyEnabled(Ext_Sv39) & (not(is_vsatp) | sys_vsatp_sv39_supported);
```
jordancarlin: Unless I am getting my parentheses mixed up, this doesn't seem right. This currently requires always requires Sv39 to be supported, but I don't think that is required for `vsatp`. Shouldn't it be the following instead:

```suggestion
    let sv39_supported = (currentlyEnabled(Ext_Sv39) & not(is_vsatp)) | sys_vsatp_sv39_supported;
```

nadime15: Ups yeah you are right, thanks. I over complicated this quite a bit. I simplified this, should also be easier to read now.

jordancarlin (approved):

jordancarlin:
Looks like this needs a rebase.

PR #1907, opened 2026-08-25
Restore `pmlen_of_mode` to be a function instead of a mapping.

pmundkur:
The mapping was potentially misleading since it was not a complete map.


nadime15 (approved):

jordancarlin (commented):
We could use the `forwards` keyword to keep this as a mapping and still catch `PMM_Reserved`, but this approach is fine too.

PR #1910, opened 2026-08-26
Use an explicit (effective) privilege in some memory utility functions.

pmundkur:
This reduces the impedance mismatch between the handling of instructions such as the hypervisor load/store instructions that operate at a different privilege than the current privilege, and the functions in `vmem_utils`.

The `mem.sail` wrappers that used an implicit privilege have been removed or now require an explicit (effective) privilege.

A subsequent refactor will align the exception contexts, and require an explicit privilege to be passed to the functions in `vmem_utils`.

### DIFF
--- a/model/extensions/Zicbop/zicbop_insts.sail
+++ b/model/extensions/Zicbop/zicbop_insts.sail
@@ -40,6 +40,7 @@ function clause execute ZICBOP(cbop, rs1, offset) = {
   let cache_block_addr = addr & ~(zero_extend(ones(plat_cache_block_size_exp)));
   let offset = cache_block_addr - rs1_val;
   let access : MemoryAccessType(mem_payload) = CacheAccess(CB_prefetch(cbop));
+  let eff_priv = effectivePrivilege(access, mstatus, cur_privilege);
 
   // TODO: This is incorrect since CHERI only requires at least one byte
   // to be in bounds here, whereas `get_transformed_data_addr()` checks that all bytes
@@ -47,7 +48,7 @@ function clause execute ZICBOP(cbop, rs1, offset) = {
   match get_transformed_data_addr(rs1, offset, access, cache_block_size) {
     Ext_DataAddr_Error(_) => (), // No access to cache; no-op.
 
-    Ext_DataAddr_OK(vaddr) => match translateAddr(vaddr, access) {
+    Ext_DataAddr_OK(vaddr) => match translateAddr(vaddr, access, eff_priv) {
       Ok(paddr, pbmt, _) => {
           // Check PMA and PMP access controls.
           let ep = effectivePrivilege(access, mstatus, cur_privilege);
--- a/model/extensions/Zicbop/zicbop_insts.sail
+++ b/model/extensions/Zicbop/zicbop_insts.sail
@@ -51,8 +51,7 @@ function clause execute ZICBOP(cbop, rs1, offset) = {
     Ext_DataAddr_OK(vaddr) => match translateAddr(vaddr, access, eff_priv) {
       Ok(paddr, pbmt, _) => {
           // Check PMA and PMP access controls.
-          let ep = effectivePrivilege(access, mstatus, cur_privilege);
-          match phys_access_check(access, pbmt, ep, paddr, cache_block_size, false) {
+          match phys_access_check(access, pbmt, eff_priv, paddr, cache_block_size, false) {
             Err(_e)  => (), // No access to cache; no-op.
             Ok(_)    => (), // The model has no caches so there's no action required.
           }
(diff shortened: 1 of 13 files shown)


jordancarlin (approved):

model/sys/vmem_utils.sail
```diff
@@ -135,10 +135,10 @@ function vmem_read_addr(vaddr, width, access, aq, rl, res) = {
...
   // virtual memory is in effect, and (b) it crosses a page boundary.
   // If we do split, perform the split access in the configured order.
-  let effPriv = effectivePrivilege(access, mstatus, cur_privilege);
-  let vmem_active : bool = if privLevel_is_virtual(cur_privilege)
-  then vsatp_mode() != Bare | hgatp_mode() != HBare
-  else satp_mode(cur_privilege) != Bare;
+  let eff_priv = effectivePrivilege(access, mstatus, cur_privilege);
+  let vmem_active : bool = if privLevel_is_virtual(eff_priv)
+                           then vsatp_mode() != Bare | hgatp_mode() != HBare
+                           else satp_mode(eff_priv) != Bare;
```
nadime15, line 144: Is this an intended change? You now pass eff_priv instead of cur_privilege in these cases. I realized that this also seems to fix a bug, I think.

pmundkur: Yes, this was intended.  I was thinking that these functions need to get the `eff_priv` as an argument, so they shouldn't be touching `cur_privilege` anyway.  I think that should work, but I'm not yet sure...

nadime15: Yes, I think so, because the old code did not split accesses even though it had to in some cases.

`vmem_active` used `cur_privilege` and basically ignored `mstatus.MPRV` (in effectivePrivilege). In M-mode `privLevel_is_virtual(Machine)` is false, so it falls back to `satp_mode(Machine)`, which always returns Bare. `vmem_active` is false now and we never split, even though `translateAddr` got the effective privilege (say S)
and did translate. The whole 8 bytes were then read from the first page's translation.

model/sys/sys_control.sail
```diff
@@ -9,7 +9,7 @@
 // Machine-mode and supervisor-mode functionality.
 
 function effectivePrivilege(access : MemoryAccessType(mem_payload), m : Mstatus, priv : Privilege) -> Privilege =
-  if   access != InstructionFetch() & m[MPRV] == 0b1
+  if   not(is_implicit_access(access)) & m[MPRV] == 0b1
```
nadime15, line 12: Is this intentional? You are no longer only covering InstructionFetch, you are also covering Load/Store(PageTableEntry).

model/extensions/Zicbop/zicbop_insts.sail
```diff
@@ -40,14 +40,15 @@ function clause execute ZICBOP(cbop, rs1, offset) = {
...
   // to be in bounds here, whereas `get_transformed_data_addr()` checks that all bytes
   // are in bounds. We will need to add a new function, parameter or access type.
   match get_transformed_data_addr(rs1, offset, access, cache_block_size) {
     Ext_DataAddr_Error(_) => (), // No access to cache; no-op.
 
-    Ext_DataAddr_OK(vaddr) => match translateAddr(vaddr, access) {
+    Ext_DataAddr_OK(vaddr) => match translateAddr(vaddr, access, eff_priv) {
       Ok(paddr, pbmt, _) => {
           // Check PMA and PMP access controls.
           let ep = effectivePrivilege(access, mstatus, cur_privilege);
```
nadime15:
```suggestion
```

model/sys/sys_control.sail
pmundkur: Yes, it is intentional, to capture the relevant line in the spec:

>  effective privilege mode, i.e., the privilege level at which explicit memory accesses execute. 

I'm not sure about the change to `fetch.sail` though; we could still use `cur_privilege` there.  If we just use `cur_privilege` there, then this change is just for completeness, but is redundant, since all implicit accesses, including for 2-stage address translation, are using the right privilege anyway (I think?).

nadime15: If you mean keeping your changes to `effectivePrivilege`, I think that makes sense and we can keep them. Then we do not need to worry about something accidentally passing `PageTableEntry` while still using `effectivePrivilege`.

pmundkur: Yeah, let's keep the change to `effectivePrivilege`.   I meant, should we keep the original code in `fetch.sail` and `fetch_rvfi.sail`, or use the change added in this PR?  I'm undecided, and could go either way.

nadime15: Ahh yeah, I am not sure either haha. What do you think @jordancarlin?

jordancarlin: I definitely think we should keep this change in `sys_control.sail` to use `is_implicit_access`. In terms of `fetch.sail`, I don't have a strong opinion one way or another, but I think using `effectivePrivilege` is more likely to catch bugs in the model, so I would lean towards that. In theory, they should be identical.

nadime15 (approved):

PR #1911, opened 2026-08-26
Expose AK_ifetch / AK_ttw at the concurrency interface

zeldovich:
The concurrency interface defines AK_ifetch and AK_ttw, but the current Sail model does not carry the information about the reason for a memory load through to the MemRead event.  This change propagates the memory read type all the way through to the MemRead event.  This makes it possible to implement memory models where instruction or page-table reads are treated differently (e.g., RISC-V's non-coherent icache).

This patch is the result of a verification project on top of the Sail model using Claude Code, and the patch itself was also co-developed together with Claude Code.

### DIFF
--- a/model/core/phys_mem_interface.sail
+++ b/model/core/phys_mem_interface.sail
@@ -30,6 +30,7 @@ enum write_kind = {
 enum read_kind = {
   Read_plain,
   Read_ifetch,
+  Read_ttw,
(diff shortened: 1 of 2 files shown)


pmundkur:
I think this is handled in #1871, which is moving to a new version of this interface.

zeldovich:
Thanks for the pointer -- I had not seen that work.  I'll wait for that to land, and try again with the new concurrency interface.

Issue #1896, opened 2026-08-22
Additional configuration needed for illegal writes to WARL fields


jordancarlin:
This has been discussed many times, but it looks like we don't actually have an open issue tracking it. Right now, we assume that illegal writes to WARL CSR fields are ignored. This is not mandated by the spec and is not the case for all DUTs. Another common choice will be to hardwire some of the bits and make the other bits writable.

This specifically came up in the context of `senvcfg.PMM` in ACT4. If 0b11 (PMLEN=16) is written to `senvcfg.PMM` when `supported_pmlen_16` is false, it is legal (but not currently supported by the model) to record that as `0b10` (PMLEN=7).

We should probably figure out a more general way of expressing this so we don't need something ad-hoc for every WARL field.

jordancarlin:
The ACTs are going to work around this a different way. I am closing this issue for now. Eventually this is something that we will need to resolve, but let's defer to the parameters SIG.

PR #1915, opened 2026-08-27
Do not ignore hgatp writes with a reserved MODE encoding

nadime15:
Per `norm:hgatp_mode_warl` such a write is not ignored like it is for `satp`, the fields are WARL, so VMID and PPN take the written values and an unsupported MODE keeps its previous value (one of a few possible legal values, we could also have set it to HBare).

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -1008,8 +1008,9 @@ function legalize_hgatp(
   written_value : xlenbits,
 ) -> xlenbits = {
   // Unlike satp, a write to hgatp with an unsupported MODE value is not ignored.
-  // Instead, the fields are WARL in the normal way: an unsupported MODE falls
-  // back to HBare while preserving the other fields (VMID, PPN).
+  // Instead, the fields are WARL in the normal way: VMID and PPN take the written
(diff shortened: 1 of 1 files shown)


nadime15:
So this PR sets `hgatp`.MODE to Bare, while Spike keeps the previous  value of `hgatp`.MODE. Both are legal, WARL only says the read has to return a legal value (This is ties back to the issue @jordancarlin opened (or closed) #1896).

(The PR itself fixes a bug where VMID and PPN should be written no matter if the write to `hgatp`.MODE was legal or not.)

The question is whether we want a config option so the user can choose, for example to match Spike. The whole function already assumes Bare is always supported, even before this PR we return Bare on an illegal write. Or we could just switch to Spike's behavior.

So there are 3 options, keep it as it is, switch to Spikes default behaviour or introduce a config option that allows BARE OR the prev. value in case of an illegal write.

pmundkur:
I suggest just following Spike for now and keeping the previous value.  We can add options for other behavior later if needed.

pmundkur (approved):

jordancarlin (approved):

PR #1916, opened 2026-08-27
Allow ignoring of writes of reserved values to `xtvec.MODE` and `xenvcfg.CBIE`.

### LINKED ISSUE #1912: add "Xenvcfg_Ignore" option for xenvcfg_cbie
Some implementation may ignore writing in xenvcfg.CBIE 10b value that is reserved according to specification. However, sail model now can only be configurated for clearing these bits or terminating if these bits are set by value 10b. I guess it would be reasonable to add option for ignoring writing this value to CBIE.

pmundkur:
This adds additional `Ignore` option values for the corresponding config parameters.

Make some of the documentation comments in the config file a bit clearer.

Fixes #1912.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -159,11 +159,12 @@
       "reserved_exceptions": false
     },
     "reserved_behavior": {
-      // The configuration option determines how to handle the reserved behavior `amocas_odd_registers`.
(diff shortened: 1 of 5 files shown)


jordancarlin (approved):
I think we're going to need a more general solution for all of this, but seems fine for now.

PR #1919, opened 2026-08-28
Add the config schema to the release artifacts.

### LINKED ISSUE #1851: Publish config schema
We should publish the JSON schema for the config so that users can specify the schema in their configs and get automatic schema validation in IDEs (like VSCode). At a minimum, we should publish the schema as part of the release (as an individual artifact) so that config files can point directly at that URL. It might also be worth publishing it to [SchemaStore](https://www.schemastore.org/) for better automatic IDE integration.

pmundkur:
Fixes #1851.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,6 +94,7 @@ jobs:
         with:
           subject-path: |
             build/sail-riscv-*
+            build/sail_riscv_config_schema.json
             build/model/sail_riscv_model.json
             build/model/sail_riscv_html.tgz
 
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):

jordancarlin:
For release builds, how about also updating the default configs that we ship so that the schema key points at this published schema? That would make it easier for users who copy those schemas to still get IDE support.

pmundkur:
Sounds good; could you create an issue for that, and we could do it in the different PR?

PR #1920, opened 2026-08-28
Add `memory.misaligned.misaligned_has_priority` config.

edevolder-tt:
The new option is `memory.misaligned.misaligned_has_priority`. The default `true` preserves existing Sail behaviour: a configured VA-misaligned exception (access-fault or address-misaligned) is taken without performing address translation. Setting it to `false` translates first and reports a page/guest-page fault if translation fails.

This exists because the privileged ISA leaves that ordering implementation-defined. Beneath Table "Synchronous exception priority in decreasing priority order":

  The relative priority of load/store/AMO address-misaligned and
  page-fault exceptions is implementation-defined to flexibly cater to
  two design points. Implementations that never support misaligned
  accesses can unconditionally raise the misaligned-address exception
  without performing address translation or protection checks.
  Implementations that support misaligned accesses only to some physical
  addresses must translate and check the address before determining
  whether the misaligned access may proceed, in which case raising the
  page-fault exception or access is more appropriate.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -230,14 +230,21 @@
       "na4_supported": true,
       "napot_supported": true
     },
-    // These settings control global support for misaligned access so they are
-    // checked before address translation. `misaligned_fault` in
-    // `regions` is checked after address translation.
+    // These settings control global support for misaligned access.
+    // `misaligned_has_priority` chooses whether a configured VA-misaligned
+    // exception is taken before address translation, or whether
+    // translation is attempted first. PMA `misaligned_exceptions` in
+    // `regions` are always checked after address translation.
     "misaligned": {
       // `exceptions` specifies the treatment of misaligned accesses
-      // before address translation.  If a fault is specified for an access,
-      // it will have a higher priority than memory-protection faults arising from
-      // that access if it were allowed to proceed.
+      // at the virtual-address level.  If a fault is specified for an
+      // access, `misaligned_has_priority` below controls whether that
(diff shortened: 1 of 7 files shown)


nadime15:
I think we do support this already, @pmundkur? Setting memory.misaligned.exceptions.X to None means nothing is raised before translation, so if translation fails you get a page fault instead, and the regions PMA misaligned exceptions decides afterwards.

pmundkur:
Yes, the model already supports this.  Perhaps this is a sign that our configuration documentation needs improving.

edevolder-tt:
The misaligned exceptions.X = None allows the VA to be misaligned and translated; translate-then-PMA for ordinary loads/stores/amo. So one would obtain PageFault ahead of Misaligned/Access.

However, that mechanism is not available to LR/SC.

The ISA-manual issue #2642 (misaligned SC/AMO plus translation fault) is explicit: the spec does not distinguish LR/SC from other ops for the priority choice.

This PR applies the same strategy for lr/sc (whilst making it uniform for all load/stores). Perhaps the implementation needs to reflect more like the existing misaligned.exception.X = { "Some": "..." }?

jordancarlin:
Support for configuring misaligned LR/SC  to throw a page fault was added in #1900. It will be included in the next release.

jordancarlin:
@edevolder-tt I’m closing this since the same functionality was already added in #1900, but feel free to reopen it if that does not meet your needs.

PR #1923, opened 2026-08-29
Improve per-HART CLINT address matching.

zeldovich:
Previously, the Sail model assumed the running HART was always #0, for the purposes of determining if a memory read/write was accessing the CLINT's MMIO registers.  As a result, accesses to the CLINT registers were modeled incorrectly: e.g., if a non-0 HART accessed HART 0's CLINT, it would actually get its own CLINT.  And if a non-0 HART accessed its own CLINT, it wouldn't get anything.

This patch improves the situation: it correctly decodes this HART's CLINT addresses, based on mhartid.  For any uses of the model where mhartid=0, this doesn't change anything.  Addresses of other HARTs are still not supported, but at least in multi-HART settings, this now allows each HART to access its own CLINT.

This patch is the result of a verification project on top of the Sail model using Claude Code, and the patch itself was also co-developed together with Claude Code.

### DIFF
--- a/model/sys/platform.sail
+++ b/model/sys/platform.sail
@@ -55,8 +55,9 @@ private function within_htif_readable forall 'n, 0 < 'n <= max_mem_access . (add
 // CLINT (Core Local Interruptor), based on Spike.
 
 // Each hart has a memory-mapped mtimecmp register. Typically these are
-// exposed as an array in CLINT. The CLINT implementation here is currently
-// hard-coded to use the mtimecmp for hart 0.
+// exposed as an array in CLINT. This model instance is a single hart, so the
+// register below is that hart's element of the array, and the memory-mapped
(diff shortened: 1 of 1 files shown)


pmundkur:
We haven't really dealt with multi-hart situations, though we have discussed them a few times, including fairly recently.  I'm currently refactoring the interrupt controller interface (see #1899) to eventually make it easier to isolate platform components like interrupt controllers that are shared across multiple harts into their own Sail components.  Would it be okay with you if I folded a fix based on this PR for the multi-hart CLINT into #1899?

zeldovich:
Sure, that sounds like a fine plan; thank you!  No rush from my end -- for the project I'm working on, I'm using a separate fork of the Sail model anyway.

pmundkur:
Closing since this functionality was added in #1899

PR #1928, opened 2026-09-01
Fix CLINT allocated size

nadime15:
Change `platform.clint.size` from 0xC0000 (786432) to 0xC000 (49152). We needed to allocate space for MSIP (4095 * 4 = 16380 Byte), 4 Byte reserved, MTIMECMP (4095 * 8 = 32760 Byte) and MTIME (8 Byte), in total 49152 bytes.


pmundkur (approved):
Good catch!

pmundkur (commented):
We should also fix this in `test/first_party/src/common/link.ld`.

jordancarlin (approved):

PR #1931, opened 2026-09-02
Add guest external interrupts to the simple interrupt generator 

nadime15:
Adds a guest register at offset 8 that sets or clears one hgeip bit. mip.SGEIP and hip.VSEIP are derived from hgeip on read. Additionally the simple interrupt generator version is bumped to 1.1 to reflect this new addition.

The added first-party test covers both SGEIP and VSEIP.

SGEIP fires when `hgeip & hgeie` is nonzero. It is always delegated to HS-mode and cannot be delegated further, so
it always stops at the hypervisor. It tells the hypervisor that one of its guests has a pending interrupt, but not
which one, so the handler reads `hgeip` to find out.

VSEIP tells the currently running guest that it has a pending interrupt. It can be set by writing `hvip`, or externally
like in this test, by setting `hstatus.VGEIN` to 1 and then driving `hgeip[1]` through the SIG. `selected_hgeip()` returns
`hgeip[hstatus.VGEIN]`, so it checks whether the selected guest has one pending.

The test takes the interrupts in descending privilege order, first SGEIP in HS-mode and then VSEIP in VS-mode. This is
not strictly necessary, we could have skipped SGEIP and gone straight to VSEIP, but this way both bits are covered in one test.


nadime15:
The first commit can be ignored. This PR relies on #1927.

model/sys/simple_interrupt_generator.sail
```diff
@@ -53,6 +56,28 @@ function sig_store(addr, width, data) = {
...
+      // Reserved bits must currently be written with zeros.
+      if data[30 .. 6] != zeros()
+      then return Err(addr, E_SAMO_Access_Fault());
+
+      // hgeip bit 0 is always zero, so anything outside 1..GEILEN does not
+      // name a guest interrupt file.
+      if guest == 0 | guest > geilen
+      then return Err(addr, E_SAMO_Access_Fault());
+
+      hgeip[guest] = value;
```
pmundkur: Instead of writing interrupt registers directly, I think we should move to the pull approach of #1899; I think this should be straightforward to port to that approach?  So perhaps merge that first, and then rebase this and port?

nadime15: Does this look okay? https://github.com/riscv/sail-riscv/pull/1931/changes/93e7d96ebd971f182d3ff59307f57c32ba6d7756

doc/ChangeLog.md
```diff
@@ -56,6 +56,11 @@
 - Xvisor boot is now tested in CI. The `os-boot` Makefile has been
   generalized to build both the Linux and Xvisor images.
 
+- The simple interrupt generator can now drive guest external interrupts.
+  A new `guest` register at offset 8 sets or clears one `hgeip` bit, which
+  makes `mip.SGEIP` and `hip.VSEIP` reachable. The device version is bumped
```
pmundkur:
```suggestion
  allows triggering `mip.SGEIP` and `hip.VSEIP`. The device version is bumped
```

model/core/interrupt_regs.sail
```diff
@@ -21,9 +21,22 @@ private function legalize_hgeie(v : xlenbits) -> xlenbits = v & hgei_mask
...
 register hgeie : xlenbits = zeros()
 
-// hgeip stays zero because no interrupt controller in the model asserts
-// guest external interrupts, which also makes mip.SGEIP permanently zero.
-register hgeip : xlenbits = zeros()
+// hgeip is read-only to software and has no state of its own. Its bits are
+// driven by the interrupt controller, so like the other platform interrupts
+// they are pulled in on read.
+// Defined in sys/simple_interrupt_generator_regs.sail, which comes after this
+// file in the compile order.
```
pmundkur:
```suggestion
// Defined in sys/simple_interrupt_generator_regs.sail.
```

model/core/interrupt_regs.sail
```diff
@@ -21,9 +21,22 @@ private function legalize_hgeie(v : xlenbits) -> xlenbits = v & hgei_mask
...
+val guest_external_interrupts_pending : unit -> xlenbits
+
+function read_hgeip() -> xlenbits = guest_external_interrupts_pending() & hgei_mask
+
+function selected_hgeip() -> bits(1) = {
```
pmundkur: This seems easier to understand to me
```suggestion
function selected_vgein_in_hgeip() -> bits(1) = {
```

pmundkur (commented):
Minor suggestions but otherwise LGTM.

doc/SimpleInterruptGenerator.md
```diff
@@ -69,6 +85,16 @@ void set_meip(volatile SimpleInterruptGenerator* sig) {
...
+// Assert the guest external interrupt for guest interrupt file `guest`,
+// i.e. set hgeip[guest]. Requires version 1.1.
+void set_hgeip(volatile SimpleInterruptGenerator* sig, uint32_t guest) {
+    uint32_t version = sig->version;
+    uint32_t minor = version & 0xFFFF;
+    uint32_t major = version >> 16;
+    assert(major == 1 && minor >= 1);
+    sig->guest = SIG_SET | guest;
+}
+
```
jordancarlin, line 99: Might be worth adding an assembly example of this so that we keep the C and assembly examples in sync.

nadime15: Agreed. If that’s okay with you, could we do this in a follow-up?

pmundkur (approved):

Issue #1913, opened 2026-08-27
Supporting page-straddling accesses for `hlv*` and `hsv*` instructions


pmundkur:
This is a known issue, but needs to be fixed before the next release, so I'm creating this to track it.

Some refactoring is needed to have those accesses go through  the `vmem_utils` functions.  #1910 is a start on that, but will need some more followup.

jordancarlin:
This should not have been closed. #1946 puts the necessary features in place, but does not actually fix this.

PR #1932, opened 2026-09-03
Refactor the `vmem_utils` public API to require the effective privilege and pointer masking width. 

pmundkur:
This further reduces the API mismatch between `vmem_utils` and `hext_insts`, en route  to fixing #1913.

### DIFF
--- a/model/extensions/H/hext_insts.sail
+++ b/model/extensions/H/hext_insts.sail
@@ -54,10 +54,11 @@ function clause execute HLVTYPE(width, op, is_unsigned, rs1, rd) = {
   };
   assert(width <= xlen_bytes);
   let eff_priv : Privilege = privLevel_bits(zero_extend(hstatus[SPVP]), 0b1);
+  let pmlen = get_hlsv_pmlen(access, eff_priv);
   match ext_data_get_addr(rs1, zeros(), access, width) {
     Ext_DataAddr_Error(e)  => { Ext_DataAddr_Check_Failure(e) },
     Ext_DataAddr_OK(vaddr) => {
-      let vaddr = transform_hlsv_address(vaddr, access, eff_priv);
+      let vaddr = transform_effective_address(vaddr, eff_priv, pmlen);
       // TODO: HLV and HSV emulate ordinary VS-mode accesses, so a misaligned
       // one that crosses a page boundary should be split and translated per
       // page, the way vmem_read_addr and vmem_write_addr do. That needs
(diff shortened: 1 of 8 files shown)


model/extensions/H/hext_insts.sail
```diff
@@ -54,10 +54,11 @@ function clause execute HLVTYPE(width, op, is_unsigned, rs1, rd) = {
   };
   assert(width <= xlen_bytes);
   let eff_priv : Privilege = privLevel_bits(zero_extend(hstatus[SPVP]), 0b1);
+  let pmlen = get_hlsv_pmlen(access, eff_priv);
   match ext_data_get_addr(rs1, zeros(), access, width) {
     Ext_DataAddr_Error(e)  => { Ext_DataAddr_Check_Failure(e) },
     Ext_DataAddr_OK(vaddr) => {
-      let vaddr = transform_hlsv_address(vaddr, access, eff_priv);
+      let vaddr = transform_effective_address(vaddr, eff_priv, pmlen);
```
nadime15, line 61: Not really an issue (for now) but just one comment so we are aware of that. `vmem_read_addr` and `vmem_write_addr` now take eff_priv, but the exception path still goes through `access_is_gva` which computes its own privilege from effectivePrivilege(..., cur_privilege). When hlv/hsv start using these helpers, they will report gva=0 because the hart is in HS-mode, but the spec says gva=1.  The hlv/hsv instructions avoid this with their by using `hlsv_memory_exception` which hardcodes gva to 1

model/extensions/H/hext_insts.sail
nadime15: This comments needs to be updated, since `transform_hlsv_address` is about to be removed.

model/extensions/H/hext_insts.sail
```diff
@@ -54,10 +54,11 @@ function clause execute HLVTYPE(width, op, is_unsigned, rs1, rd) = {
...
       // TODO: HLV and HSV emulate ordinary VS-mode accesses, so a misaligned
       // one that crosses a page boundary should be split and translated per
       // page, the way vmem_read_addr and vmem_write_addr do. That needs
```
nadime15, line 64: This comment and the one at line ~120 can be modified too, since we know have privilege aware version of those helpers. But you can also keep them and clean everything up in the final PR.

model/extensions/H/hext_insts.sail
pmundkur: Indeed, I plan to address the exception handling differences in the next refactor on top of this.  I think that is the only difference left?

model/extensions/H/hext_insts.sail
pmundkur: Yeah, I'll remove the TODOs in the final PR.

model/extensions/H/hext_insts.sail
nadime15: Yeah, I think so.

nadime15 (approved):

model/riscv.sail_project
```diff
@@ -447,7 +447,7 @@ extensions {
   }
 
   Zicboz {
-    requires core, sys, exceptions
+    requires core, sys, exceptions, PM_utils
```
jordancarlin, line 450: Slightly unfortunate that PM_utils is now required by so many other modules, but seems ok for now. Eventually we might want to restructure this a bit.

jordancarlin (approved):

model/riscv.sail_project
pmundkur: I actually thought that the making the dependency explicit was a good thing, since the interaction with pointer masking with these extensions is discussed in the spec.

jordancarlin: That's fair as well, but these extensions do not depend on pointer masking, so I would prefer to see it wrapped in some kind of `mem` module or something. Either way, not something to address in this PR.

PR #1935, opened 2026-09-03
Allow configuring the `mcountinhibit` CSR.

### LINKED ISSUE #1933: mcountinhibit doesn't have to be accessible
As norm:mcountinhibit_not_impl states:
> If the mcountinhibit register is not implemented, the implementation behaves as though the register were set to zero.

Sail currently doesn't give an option to make `mcountinhibit` non-accessible, nor does it make it possible to mark only subset of bits writeable. Both of those properties are already available via UDB.

pmundkur:
The configuration allows specifying whether the CSR is implemented, and if so, which bits are writable.  The latter configuration decouples it from `base.writable_hpm_counters` which was used to control the writable bits.

Decoupling the two now loosens the constraints, making it possible to configure the inhibition of an unimplemented counter.

Fixes #1933.

### DIFF
--- a/config/config.json.in
+++ b/config/config.json.in
@@ -60,6 +60,17 @@
       "len": 32,
       "value": "0xFFFF_FFFF"
     },
+    "mcountinhibit" : {
+      // Whether the `mcountinhibit` CSR is implemented.
+      "supported": true,
+      // This bits in this value control whether the corresponding
+      // bits of the CSR are writable.  Note that bit 1 cannot be  written to
+      // as it is read-only zero.
+      "writable_bits": {
+        "len": 32,
+        "value": "0xFFFF_FFFD"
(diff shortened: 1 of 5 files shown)


jordancarlin (approved):

nadime15 (approved):

nadime15 (commented):
I just found this:

> if `mcountinhibit` is not implemented, the model should behave as if the register was zero.

I think we should at least initialize it to 0 at declaration or reset it to 0, just to be on the save side (I think the C++ back end will set it to 0, but you never know).

nadime15 (approved):

pmundkur:
Let's merge this after a bug-fix release, since it changes the config file.  I removed the `will-be-merged` label.

Issue #1862, opened 2026-08-07
Priv spec 1.11 exception code 16 through 23 should all be reserved


farazahmad-10xe:
[2010 ACT ISSUE](https://github.com/riscv/riscv-arch-test/issues/2010) 

The target's sail.json selects Privileged_ISA_1_11. Privileged ISA 1.11 reserves exception codes 16 through 23 but sail added unsupported mCause in the test.
 
priv spec 1.11 exception code 16 through 23 should all be reserved. It should not be possible to set MCAUSE to 19 in the sail  model if you have version 1.11 of priv specs.

PR #1939, opened 2026-09-08
Let scause/vscause hold any Exception Code 0-31

### LINKED ISSUE #1938: `scause` does not support values 0-31
The spec says the following:

> The Interrupt bit in the scause register is set if the trap was caused by an interrupt. The Exception Code field contains a code identifying the last exception or interrupt. [Table 115](https://riscv.github.io/riscv-isa-manual/snapshot/spec/#scauses) lists the possible exception codes for the current supervisor ISAs. The Exception Code is a WLRL field. It is required to hold the values 0–31 (i.e., bits 4–0 must be implemented), but otherwise it is only guaranteed to hold supported exception codes.

The `scause` register rejects Hypervisor exception and interrupt codes when H is not supported, which does not comply with the spec.

nadime15:
The WLRL Exception Code field must be able to hold any value in 0-31, interrupt or exception, only mcause is exempt.

Fixes #1938.

### DIFF
--- a/model/core/sys_regs.sail
+++ b/model/core/sys_regs.sail
@@ -659,9 +659,14 @@ bitfield Mcause : xlenbits = {
 
 // The Exception Code is a WLRL field of the `{ms}cause` CSRs.  The
 // read-legal is implemented by preserving the previous value.
-function legalize_cause(o : Mcause, v : xlenbits) -> Mcause = {
+//
+// scause/vscause must also hold any Exception Code in 0-31,
+// interrupt or exception alike, so a code that fits in 5 bits
+// is kept as written. mcause has no such requirement.
+function legalize_cause(o : Mcause, v : xlenbits, is_scause : bool) -> Mcause = {
   let v = Mk_Mcause(v);
   let code = v[Cause];
+  if is_scause & unsigned(code) <= 31 then return v;
   if v[IsInterrupt] == 0b1 then {
     // No platform interrupts are currently supported.
     let is_platform_interrupt = (code & sign_extend(0xF0)) != zeros();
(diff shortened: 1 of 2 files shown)


model/core/sys_regs.sail
```diff
@@ -659,9 +659,14 @@ bitfield Mcause : xlenbits = {
 
 // The Exception Code is a WLRL field of the `{ms}cause` CSRs.  The
 // read-legal is implemented by preserving the previous value.
-function legalize_cause(o : Mcause, v : xlenbits) -> Mcause = {
+//
+// scause/vscause must also hold any Exception Code in 0-31,
+// interrupt or exception alike, so a code that fits in 5 bits
+// is kept as written. mcause has no such requirement.
+function legalize_cause(o : Mcause, v : xlenbits, is_scause : bool) -> Mcause = {
```
jordancarlin, line 666: WLRL technically allows anything to be read back after writing an illegal value, so we don't necessarily have to special-case `mcause`. @pmundkur, thoughts?

nadime15: Ironically if we don't special case `mcause`, we could shrink `legalize_cause()` to:

```
function legalize_cause(o : Mcause, v : xlenbits) -> Mcause = {
    let v = Mk_Mcause(v);
    if unsigned(v[Cause]) <= 31 then v else o;
  }
```

jordancarlin: Eventually we might need to add some kind of configuration for this, but for now, I am fine with either approach.

pmundkur: This code was added to fix #1862, so we should probably keep the checks for `mcause`?

jordancarlin: We are going to change the tests to avoid writing reserved values to `mcause` since the result is nondeterministic, but seems fine to me to keep the check.

jordancarlin (approved):

pmundkur (approved):

pmundkur:
~@nadime~ @nadime15  it looks there's now a verified signature check for this repo that's blocking this from being merged.  

Perhaps these are the relevant instructions?  https://github.com/riscv/riscv-isa-manual/blob/main/CONTRIBUTING.md#signing-git-commits-with-your-signing-key 

I forgot how I set it up.

nadime15:
Does this work now, or do we have to wait until all the jobs are finished?

pmundkur:
I've set this to auto-merge, so it should get merged once the ci finishes if the signing check passes

PR #1940, opened 2026-09-08
Improve CI latency.

pmundkur:
The main CI workflow is split into two matrices: a build matrix to range over the different build combinations, and a test matrix that executes test groups for each build.

Tests are grouped using CMake test labels, and roughly equalized by test group execution time.

The tests downloaded by the default CI run are cached, though this does not significantly affect the overall CI duration.

Although the test runs are now concurrent, the build and test jobs still need to be serialized; so a slow build could still slow down the whole workflow by delaying all the test runs.  Another cause of slowdown is the wait for the increased number of test runners to be available.  But this workflow should pay off once more test suites are added.

The build and test matrices are generated by a script; this should provide some flexibility as more test suites are added.  The trade-off is more book-keeping since labels would need to be added for the tests, and updated in the matrix generation script.

The `test-results` workflow was dropped since it wasn't really useful.


pmundkur:
This took around 2hrs 10 mins in my repo, versus around 2hrs 20 mins of the last few  CI runs in this repo; so not much in savings.  The runs in my repo had significant waits for macos runners; perhaps that wait will be less in this repo.

nadime15:
Hmm, I am not sure if this is the right approach. You added a comment saying that it is not always optimal to put each test suite into a different runner since we have so many different build variants. I am not sure how much of a difference this actually makes in the end in terms of the total time we have to wait, although this also depends on how many runners we can spawn.

We would spawn a lot more runners, but at least the individual test cases would fail much faster, so you would get feedback much sooner. The average runtime of the whole check should probably be faster when splitting (again, depends on the number of runners). I think with the current approach, you are ultimately just pushing the problem further down the road by regrouping the tests into Vector/GeneralTest and Hypervisor which may need to be regrouped at some time again if one of the classes ballons in numbers of test cases.

Also, I am not sure this approach will work long term. Either ACT4 eventually adds the Hypervisor tests and the Alibaba test suite gets abandoned, or ACT4 adds the Hypervisor tests while the Alibaba suite remains separate. In the latter case, how would you extract the Hypervisor tests from ACT4? And even if you could do that, you would end up having to sort and regroup the tests again anyway (haven't they added Vector tests?). 

I also don't think we need to run each test-suites twice, for 2 different versions of CMake and test-suites like "riscv-tests" are fairly small so the job should be done within a few minutes.

pmundkur:
The labels are placeholders; we can adjust them as needed. I was initially planning 'test-bucket1', 'test-bucket-2', etc.  So the question of extracting hypervisor tests from ACT doesn't arise; we can use whatever labels we need.

Also, the last two successful runs in this repo have clocked around 1hr 15 mins (runs #3288 and #3289 in https://github.com/riscv/sail-riscv/actions), so it is a _full hour faster_ than the current approach.  As I expected, it was just slower in my personal repo, since Github understandably doesn't prioritize those repos.

I had tried a runner per suite (as the comment indicated, it needed 40 runners) which took around 2 1/2 hours (I think).  It might be faster in this repo, but it would be a real pain to test our branches in our personal repos before submitting them as PRs here.

> Also, I am not sure this approach will work long term. Either ACT4 eventually adds the Hypervisor tests and the Alibaba test suite gets abandoned, or ACT4 adds the Hypervisor tests while the Alibaba suite remains separate. In the latter case, how would you extract the Hypervisor tests from ACT4? And even if you could do that, you would end up having to sort and regroup the tests again anyway (haven't they added Vector tests?).

That's why I like the label approach.  I picked it precisely because I expect the ACT suite to grow bigger, and eventually taking too long with a single runner.  The label approach allows us to split ACT into different labels using CMake string/regex matching on the directory/folder/name of the test.

pmundkur:
Interestingly, the corresponding two runs in my personal repo took around 1 1/2 hours (runs 1099 and 1100 in https://github.com/pmundkur/sail-riscv/actions), so it's also faster in the personal repos.

.github/workflows/ci.yml
```diff
@@ -3,39 +3,31 @@ name: CI
 on: [push, pull_request, workflow_dispatch, merge_group]
 
 jobs:
```
nadime15, line 5: I am just reading up on [concurrency](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency), which could be really handy. Instead of running CI for every commit pushed to a given PR, it will cancel the CI run for an older commit and only keep the CI run for the latest commit.

```suggestion
on: [push, pull_request, workflow_dispatch, merge_group]

concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}

jobs:
```

pmundkur: I think this is unrelated to what this PR is trying to do.  Also see #1445, where it seems to interact with merge queues in some weird way.  It would certainly be good to fix, but perhaps in #1445 or a new PR?

nadime15:
> I had tried a runner per suite (as the comment indicated, it needed 40 runners) which took around 2 1/2 hours (I think). It might be faster in this repo, but it would be a real pain to test our branches in our personal repos before submitting them as PRs here.

Do you have a link to this specific run? I’d like to see how long each step took.

.github/workflows/ci.yml
nadime15: Yeah, I tried to sneak this in :D But I wasn’t aware of #1445! But yeah, that’s fine then.

pmundkur:
> > I had tried a runner per suite (as the comment indicated, it needed 40 runners) which took around 2 1/2 hours (I think). It might be faster in this repo, but it would be a real pain to test our branches in our personal repos before submitting them as PRs here.
> 
> Do you have a link to this specific run? I’d like to see how long each step took.

https://github.com/pmundkur/sail-riscv/actions/runs/34265537632/usage

It looks like it took 2hrs 3 mins, which is slower than the latest version of the PR in my repo (CI run #1100) which took 1hr 15 mins.

nadime15:
Ok, that’s interesting. The issue is that we run each test twice for two different cmake versions (which I think is somewhat unnecessary) and in turn, limits the number of runners available. Just cutting the number of runners in half would, I thin, close the gap and bring the per test suite approach closer to one hour as well.

I also noticed that the test jobs don’t need to install the Sail Compiler for macOS, which adds quite a bit of overhead to the overall runtime. For example, in this job: https://github.com/pmundkur/sail-riscv/actions/runs/34265537632/job/102205225718, removing the Sail build would probably bring the runtime down from around 12 minutes to 2 to 3 minutes and free up the runner much sooner (for this specific job, since installing Sail takes ~10 minutes but running the tests takes 1 minute).

So if we remove the Sail build from each individual test suite job for the macOS builds, we could save 5 to 10 minutes per macOS runner, which seems to be the bottleneck right now. I think it’s important that, when testing the per test suite/build approach, we keep the number of runners to the bare minimum and reduce the runtime of each runner as much as possible.

pmundkur:
I do think we should really test every build, even across different CMake versions; after all, the tests use CTest, which is part of CMake, so we should ensure the tests work across these CMake versions.  [Separately, we should think about incrementing these versions...]

The test jobs run CMake to generate the test plan, and CMake currently checks that Sail is installed.  We could have an `ONLY_TEST` option to avoid this check, but I think the CMake dependency logic will get quite complicated for it to generate a test plan without a key dependency present for the tests.  

I tried to avoid the whole CMake step on the test runners, and ship the whole build directory.  That doesn't work, since CMake puts full paths into its generated build files, and the downloaded build directory on the test runners had absolute paths pointing to things that didn't exist (especially on the container build).   It just doesn't work; there is even a FAQ for it:
https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#why-does-cmake-use-full-paths-or-can-i-copy-my-build-tree

As it stands now, the CI times with this PR are already roughly back to the pre-hypervisor CI times (see CI runs 2990 and earlier).

.github/workflows/ci.yml
jordancarlin: I forgot about #1445. I got that working in riscv-arch-test and riscv-unified-db, so it should be possible here too. I will circle back to #1445 and update it soon.

jordancarlin:
I wonder if long term, the better solution is to migrate away from ctest for our testing. Seems like a fairly basic Python script would be able to handle this in a much simpler manner and would avoid the messy dependency issues.

In terms of the number of available runners, ACT CI uses close to 70 runners in parallel and never has issues with there being enough available runners. The riscv GitHub org is an enterprise account, so it gets access to a lot more runners in parallel.

pmundkur:
Yeah, moving away from CTest might be better eventually.

I'm hesitant to require resources (like a lot of runners) in this repo that are not readily accessible to a personal account.  It would get in the way of contributors testing their branches before submitting them as PRs.  I don't think we're there yet, fortunately.

nadime15:
What is actually the point of these tests and what do we want to test? Do we want to test the build process or the actual model? Probably 85%+ of all changes in PRs touch the model, while neither the CMake files nor the workflow files are changed. The build process is changed very rarely and is mostly static, while the model itself is constantly changing.

So most of the time, what we actually want to test is the model and make sure that a PR did not break the model itself and not the build process. There is simply no need to run the Hypervisor, Vector, etc. tests twice for the same Rocky Linux (Ubuntu ARM) build just because we use two different CMake versions (since the actual tests, take up most of the time). And it is not as if Vector test A would suddenly stop working for the Rocky Linux (Ubuntu ARM) build with CMake version X while still working with CMake version Y, right?

If we actually want to test the build process, I think we should have a dedicated workflow for that instead of running the build process multiple times and consequently running the same tests twice for each build.

Besides that, I spent some time working on an approach where each suite is run once for each build. Here is an example run:

https://github.com/nadime15/sail-riscv/actions/runs/34386711902

It took around 1h20. On master, this should be even quicker since the `run_tests` matrix does not really run fully in parallel on my fork. In theory, we could make this even quicker with a bit more work. For example, we do not really need to wait for all runners in the build matrix to finish. As soon as a specific build is ready, we could immediately spawn all the runners for that build.

jordancarlin:
I think it is worth running a brief smoke test on each build (either the first party tests or riscv-tests since those are both short), but I agree that beyond that it seems unnecessary to run the full suite for every build. One full run per architecture/OS is probably more than sufficient.

nadime15:
> I think it is worth running a brief smoke test on each build (either the first party tests or riscv-tests since those are both short)...

Yeah this works too.

pmundkur:
While we could improve/simplify the workflow to avoid CMake versions, could we at least get this basic flow in since it improves the CI by _one full hour_ (almost _halving_ it)?

> Besides that, I spent some time working on an approach where each suite is run once for each build. Here is an example run:
>
>https://github.com/nadime15/sail-riscv/actions/runs/34386711902

ACT 4.1 almost doubles the number of tests from what we currently have, so we'd need to start splitting it to balance test runner execution durations.  I think this might be easier to do with the label approach, instead of running a test suite per runner.

jordancarlin:
> While we could improve/simplify the workflow to avoid CMake versions, could we at least get this basic flow in since it improves the CI by _one full hour_ (almost _halving_ it)?

Fully agreed that this can and should be done in follow up PRs (along with possibly moving away from ctest entirely).

nadime15:
Ok sure, then let’s move forward with this approach for now.

pmundkur:
It would be good to get this reviewed, approved and merged soon.

test/CMakeLists.txt
```diff
@@ -1,6 +1,59 @@
...
 set(TEST_DOWNLOAD_URL "https://github.com/riscv-software-src/sail-riscv-tests/releases/download" CACHE STRING "Base URL to download precompiled riscv-tests and riscv-vector-tests from")
-set(TEST_DOWNLOAD_VERSION "2026-08-20" CACHE STRING "Version of precompiled tests to download")
+set(TEST_DOWNLOAD_VERSION "${SAIL_RISCV_TESTS_VERSION}" CACHE STRING "Version of precompiled tests to download")
+
+# Simulator to use for tests
+set(CACHE{SAIL_RISCV_SIM}
+    TYPE FILEPATH
+    HELP "sail_riscv_sim binary to use for tests"
+    VALUE ""
+)
```
nadime15: The signature was added with CMake 4.2, so this wont work in our case (see https://cmake.org/cmake/help/latest/command/set.html)
```suggestion

```

Good that just talked about CMake yesterday! 😄

.github/workflows/ci.yml
nadime15:
```suggestion
    needs: [define-matrix, build, run_tests, linux, xvisor]
if: "${{ always() }}"
steps:
- name: exit failure
if: "${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}"
run: exit 1
- name: exit success
```

.github/workflows/ci.yml
```diff
@@ -3,39 +3,31 @@ name: CI
...
+      - name: Define build matrix
+        id: define_build_matrix
+        run: .github/scripts/ci-matrix.py --build
+
+      - name: Define test matrix
+        id: define_test_matrix
+        run: .github/scripts/ci-matrix.py --test
+
   build:
     name: "Build and test model (OS: ${{ matrix.os }}${{ matrix.container && format(', container: {0}', matrix.container) || '' }}, CMake Version: ${{ matrix.cmake_version }})"
```
jordancarlin: Name should be changed. This no longer does any testing.

.github/workflows/ci.yml
```diff
@@ -88,25 +92,140 @@ jobs:
...
-          ctest --test-dir build --output-junit tests.xml --output-on-failure
+          ctest --test-dir build -L TestAtBuildTime --output-junit tests.xml --output-on-failure
 
       - name: Upload test results
-        if: ${{ matrix.run_all_steps}}
+        if: ${{ matrix.run_all_steps }}
         uses: actions/upload-artifact@v7
         with:
-          name: tests.xml
+          name: first-party-tests-results.xml
```
nadime15: You renamed `tests.xml`, `event.json` and added `${{ matrix.label }}-results.xml` but .github/workflows/test-results.yml still relies on the old names...

.github/workflows/ci.yml
```diff
@@ -88,25 +92,140 @@ jobs:
...
+        uses: actions/cache/restore@v6
+        id: sail-riscv-tests-cache-restore
+        # On the master branch, we could set `fail-on-cache-miss:` to
+        # 'true', but while the PR with this workflow is not merged
+        # into master, the cache won't get created, so don't set it.
+        # Similarly, the same issue will arise for the PR that updates
+        # the version of the tests.
+        with:
+          path: build/test/${{ env.SAIL_RISCV_TESTS_VERSION }}
+          key: ci-sail-riscv-tests-${{ env.SAIL_RISCV_TESTS_VERSION }}
```
jordancarlin, line 193: Does this mean no tests are run when the cache is empty? If so, how do we validate PRs that change the test version?

.github/workflows/ci.yml
```diff
@@ -88,25 +92,140 @@ jobs:
...
+        # only the minimal required flags are specified.
+        run: |
+          cmake -S . -B build -GNinja \
+            -DSAIL_RISCV_SIM=built_c_emulator/sail_riscv_sim \
+            -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+            -DENABLE_RISCV_TESTS=TRUE \
+            -DENABLE_RISCV_ARCH_TESTS=TRUE \
+            -DENABLE_DAMO_TESTS=TRUE \
+            -DENABLE_RISCV_VECTOR_TESTS_V64_E64=TRUE \
+            -DENABLE_RISCV_VECTOR_TESTS_V128_E32=TRUE
```
jordancarlin, line 207: This can probably be further optimized later so that we don't download test suites we don't need, but fine for now.

.github/scripts/ci-matrix.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def define_build_matrix_entries() -> list[dict]:
+    entries: list[dict] = []
+
+    entries.append(
+        {
+            "os": "ubuntu-22.04",
+            "cmake_version": "3.20.0",
+            "first_party_tests": "true",
```
nadime15:
```suggestion
            "first_party_tests": True,
```

And the changes below too

.pre-commit-config.yaml
```diff
@@ -39,3 +39,12 @@ repos:
     hooks:
       - id: markdown-link-check
         args: [-q, -c .markdown-link-check.config]
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
+    rev: v0.16.6
+    hooks:
+      # Run the linter.
+      - id: ruff-check
+      #  args: [--fix]
```
jordancarlin: Why is `--fix` commented out? All of our other pre-commit hooks directly modify the file.

.github/scripts/ci-matrix.py
```diff
@@ -0,0 +1,159 @@
...
+# 'riscv-vector-tests'), there would be 40 runners needed for the test
+# matrix.  The smaller runtime for each suite is often neutralized by
+# the time waiting for more runners to be available.
+#
+# Instead, tests are grouped using labels (see test/CMakeLists.txt),
+# and a test runner is used per label per build.
+
+
+def sail_riscv_test_labels() -> list[str]:
+    return ["GeneralTest", "VectorTest", "HypervisorTest"]
```
jordancarlin: It might make more sense to just make things like this a constant rather than using a function.

.github/scripts/ci-matrix.py
```diff
@@ -0,0 +1,159 @@
...
+        parser.print_help()
+        sys.exit(0)
+
+    opts = parser.parse_args(cliopts)
+    if opts.build:
+        show_build_matrix(opts)
+    elif opts.test:
+        show_test_matrix(opts)
+
+    return 0
```
jordancarlin, line 149:
```suggestion
    parser = cli_parser()
    opts = parser.parse_args()

    if opts.build:
        show_build_matrix(opts)
    elif opts.test:
        show_test_matrix(opts)
    else:
        parser.print_help()

    return 0
```

.github/scripts/ci-matrix.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def cli_parser():
+    parser = argparse.ArgumentParser(description="Generate CI matrix entries")
+    parser.add_argument(
+        "-b", "--build", action="store_const", const=True, help="generate build matrix"
+    )
+    parser.add_argument(
+        "-t", "--test", action="store_const", const=True, help="generate test matrix"
+    )
```
jordancarlin, line 137: How about using `add_mutually_exclusive_group` so that the parser enforces that only one of these is passed? If you also set `required=True` for the group, you can drop the check for 0 arguments below.

.pre-commit-config.yaml
jordancarlin, line 1: Since you typed all of the Python code, it might be worth adding a type checker too. `pyrefly` seems like a good choice these days.

jordancarlin (changes requested):

.github/workflows/ci.yml
jordancarlin: I've been questioning the value of the `test-results.yml` workflow. We don't really care how many tests failed (just that tests failed), which we already get from the CI results. What about just dropping that workflow?

test/CMakeLists.txt
pmundkur: That's weird, I have CMake 3.28 locally, and this works with it.  And it worked in CI with the CMake versions we need to work.  If really needed, instead of dropping it, I could just use the older syntax.

.pre-commit-config.yaml
pmundkur: The `forbid-tabs` check doesn't fix the file.  I'm new to ruff, and was reluctant to let it 'fix' things until I have more experience with it.  But if you're confident about it, then we could let it fix.

.pre-commit-config.yaml
pmundkur: I'll take a look, thanks!

.github/workflows/ci.yml
pmundkur: No, if the cache is empty (e.g. when the test version is changed), the tests are downloaded by the 'cmake -S' in the next step.

.github/workflows/ci.yml
pmundkur: Agreed, I doubt this is very useful, and we should drop it.

.github/workflows/ci.yml
pmundkur: It does run the 'TestAtBuildTime' tests, see the "Run build-time tests" step.

.github/scripts/ci-matrix.py
pmundkur: Not sure this is needed; Python is not evaluating these values, just dumping these strings to JSON.

.pre-commit-config.yaml
jordancarlin: `forbid-tabs` doesn't fix the files because there is not an obvious fix (how many spaces should it replace each tab with?). Most of the other pre-commit hooks do. `ruff --fix` only auto-fixes things with clear fixes. For anything else, it just throws an error and let's the user handle it. I recommend enabling it and have had success with it in several other projects.

.github/workflows/ci.yml
jordancarlin: Oops. Of course it does.

.github/workflows/ci.yml
jordancarlin: True. I still think the name is a little misleading, but will leave it up to you.

.github/scripts/ci-matrix.py
nadime15: json.dumps("false") generates a quoted string  and github will evaluate non empty strings to true. 

See https://docs.github.com/en/actions/reference/workflows-and-actions/expressions

> Note that in conditionals, falsy values (false, 0, -0, "", '', null) are coerced to false and truthy (true and other non-falsy values) are coerced to true.

.github/scripts/ci-matrix.py
```diff
@@ -0,0 +1,155 @@
...
+    parser = cli_parser()
+    opts = parser.parse_args()
+    if opts.build:
+        show_build_matrix(opts)
+    elif opts.test:
+        show_test_matrix(opts)
+    else:
+        parser.print_help()
+
+    return 0
```
jordancarlin, line 149:
```suggestion
def cli_parser():
    parser = argparse.ArgumentParser(description="Generate CI matrix entries")
    group = parser.add_mutually_exclusive_group(required=True)
    group.add_argument(
        "-b", "--build", action="store_const", const=True, help="generate build matrix"
    )
    group.add_argument(
        "-t", "--test", action="store_const", const=True, help="generate test matrix"
    )
    return parser


def main() -> int:
    parser = cli_parser()
    opts = parser.parse_args()
    if opts.build:
        show_build_matrix(opts)
    elif opts.test:
        show_test_matrix(opts)

    return 0
```

.github/scripts/ci-matrix.py
```diff
@@ -0,0 +1,155 @@
...
+
+import argparse
+import copy
+import json
+import os
+import pathlib
+import sys
+
+
+def define_build_matrix_entries() -> list[dict]:
```
jordancarlin, line 11: Might make more sense for this to be a constant too, but either way is fine.

.github/workflows/ci.yml
```diff
@@ -88,25 +92,140 @@ jobs:
...
+        run: |
+          ctest --test-dir build -L ${{ matrix.label }} --output-junit tests.xml --output-on-failure
+
+      - name: Upload test results
+        if: ${{ matrix.run_all_steps }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: ${{ matrix.label }}-results.xml
+          path: build/tests.xml
+          if-no-files-found: error
```
jordancarlin: I don't think there is any reason to upload the xml files (or even produce them for that matter) if we are no longer posting the comment that consumed them.

jordancarlin (approved):

.github/workflows/ci.yml
```diff
@@ -88,25 +92,124 @@ jobs:
             $CLANG_FLAG
           ninja -C build all generated_sail_riscv_docs generated_smt_rv64d generated_smt_rv32d
 
-      - name: Run tests
+      - name: Cache downloaded tests
+        uses: actions/cache/save@v6
+        # Update cache only on the master branch, and from only one of the jobs in the matrix.
+        if: ${{ matrix.run_all_steps == 'true' && steps.sail-riscv-tests-cache-restore.outputs.cache-hit != 'true' && github.ref == 'refs/heads/master' }}
```
nadime15: Mismatched types, lets simplify this.
```suggestion
        if: ${{ matrix.run_all_steps && steps.sail-riscv-tests-cache-restore.outputs.cache-hit != 'true' && github.ref == 'refs/heads/master' }}
```

nadime15 (commented):
Looks good to me. Once the last comment is addressed, it’s good to go.

nadime15 (approved):

PR #1942, opened 2026-09-09
Fix incomplete masking of `mip` and `mie` for the configured supported interrupts.

edevolder-tt:
The recently introduced configuration to specify supported interrupts did not correctly mask all sources of updates to `mip`, e.g. `mip.MSIP`, and all updates via aliases to `mip`, e.g. `sip`.

This extends the mask to the above sources, and also similarly extends the masking for `mie`.


model/sys/interrupt_implementation.sail
```diff
@@ -14,8 +14,11 @@ function external_interrupts_pending(hgeie_val : xlenbits) -> Minterrupts =
   if plat_have_sig then sig_external_interrupts_pending(hgeie_val) else Mk_Minterrupts(zeros())
 
 function machine_software_interrupt_pending() -> bits(1) =
-  (if plat_have_clint then clint_machine_software_interrupt_pending() else 0b0)
-  | (if plat_have_sig then sig_machine_software_interrupt_pending() else 0b0)
+  if not(plat_has_msi) then 0b0
+  else (
+    (if plat_have_clint then clint_machine_software_interrupt_pending() else 0b0)
+    | (if plat_have_sig then sig_machine_software_interrupt_pending() else 0b0)
+  )
```
pmundkur, line 21: I don't think this is needed; this masking is already done here:
https://github.com/riscv/sail-riscv/blob/eb58856d32eed2e3aae4c89a7115c3831299550a/model/core/interrupt_regs.sail#L230
using
https://github.com/riscv/sail-riscv/blob/eb58856d32eed2e3aae4c89a7115c3831299550a/model/core/interrupt_regs.sail#L58

pmundkur (commented):
Do you have a test showing that `mip.MSIP` is being driven even though it is supposed to be disabled by `plat_has_msi`?

model/sys/interrupt_implementation.sail
pmundkur: Hmm, I think this is unprotected:
https://github.com/riscv/sail-riscv/blob/eb58856d32eed2e3aae4c89a7115c3831299550a/model/core/interrupt_regs.sail#L256

pmundkur: I need to leave for the day, but a better way to fix this would be to create a `new_mip` in `update_mip()` instead of updating `mip` directly, then assign `mip = supported_interrupts_mask(new_mip);`  before the `if old_mip != mip` check.

pmundkur: Something like this (untested):
https://github.com/pmundkur/sail-riscv/commit/851c201f2f8b5b7d668bdd11169b9660b7438c6f

edevolder-tt: I did forget to mention that this found with ACT4, and verified with ACT4.
I'll look at it again as you suggest tomorrow.

nadime15:
I haven’t checked it in detail, but isn’t this what we discussed here https://github.com/riscv/sail-riscv/pull/1934#pullrequestreview-5106209540 and here https://github.com/riscv/sail-riscv/pull/1934#pullrequestreview-5115612605?

model/sys/interrupt_implementation.sail
nadime15: What about the following solution? Its pretty straightforward: https://github.com/riscv/sail-riscv/commit/b8cd92a5a158b2bb429082687990464fc772c6f0

pmundkur: > What about the following solution? Its pretty straightforward: [b8cd92a](https://github.com/riscv/sail-riscv/commit/b8cd92a5a158b2bb429082687990464fc772c6f0)

I'd prefer to use the `var new_mip = mip; ... /* build new_mip */ ...;. mip = supported_mask(new_mip)` instead of `mip = supportedMask(mip);` since I think it's clearer for the reader, but could go either way.  

If `mip` is always masked, then only `platform_interrupts` need to be masked in `read_mip()`, masking `mip` again there is redundant.

nadime15: > If mip is always masked....

No its not, `update_mip`() writes some bits directly in update_mip(), and lift_sip() also writes mip.SSI unmasked when mideleg[SSI]=1 with plat_has_ssi=false, so masking in update_mip() alone does not keep mip legal. 

Writing sip would also result in a csr callback showing the dirty bit, even though it is supposed to be masked. If someone runs verification and catches this, it would be a bug.

Masking the full mip platform_interrupts in read_mip() is the only thing that guarantees a read never returns an unsupported pending bit and would be future proof too.

pmundkur: I'd prefer to have the `mip` value always legal according to `supported_interrupts_mask`; it would be confusing in a trace when a write to `sip` shows a bit set, but then a subsequent `read_mip()` shows it clear.
How about this:
https://github.com/pmundkur/sail-riscv/commit/8277a9fe48715575e52b9265eef1dd58186fc950

jordancarlin: Good point about the trace. I don't think we ever want the trace showing an illegal value in a CSR.

nadime15: Hmm, I don't follow. Just to be clear, with the read_mip() masking, the trace is already clean (my commit https://github.com/riscv/sail-riscv/commit/b8cd92a5a158b2bb429082687990464fc772c6f0). The value passed to csr_write_callback on a sip write goes through the masked read_mip(), so it shows SSI=0. The illegal bit only stays in the internal mip register for one instruction and is never passed to any callback.

So if the concern is the callback trace, this already covers it. If you mean a full csr state comparison at every step, then that's a different issue.

https://github.com/pmundkur/sail-riscv/commit/8277a9fe48715575e52b9265eef1dd58186fc950 is not (yet) future proof, if a similar pattern appears we would run into the same issues, that's why i am in favor of applying the mask like I did in read_mip(), then it should be fine because this would never allow illegal bits into the interrupt register (if this is what we want).

edevolder-tt: I've tested both https://github.com/riscv/sail-riscv/commit/b8cd92a5a158b2bb429082687990464fc772c6f0 and https://github.com/pmundkur/sail-riscv/commit/8277a9fe48715575e52b9265eef1dd58186fc950.

Both solve the problem that I observed (ie the InterruptsXX tests pass). Below is an AI-assisted write-up of the problem in detail. This is InterruptsS-00 test.

==========
Failing coverpoint: cp_priority_mip_s_m_nodeleg_m_011
DUT PC: 0x8000b03c (nops after GOTO_SMODE, InterruptsS-00.S:8954)
Checker: M-mode trap, XIP mismatch: Sail 0x88 (MTIP+MSIP) vs Whisper 0x80 (MTIP only). Both took MTI (cause 7).

The previous coverpoint nodeleg_m_010 also took MTI at 0x8000ae7c and matched. That case does not call SET_MSW. nodeleg_m_011 does, then programs the timer and enters S-mode.

What SET_MSW actually is

On the DUT ELF, rvtest_set_msw_int_m is a single ret:

80067e4c:  jalr x0, 0(x1)
RVMODEL_SET_MSW_INT is empty in config/rpc-tt/rvmodel_macros.h (TT ACLINT has no sw_offset). Whisper mie/mip masks also have no bit 3, and there is no MSWI device, so DUT mip.MSIP stays 0.

The Sail signature ELF is compiled with sail_macros.h, which overrides that to:

sw 1, (SAIL_CLINT_BASE + 0)   /* MSIP */
So Sail’s CLINT at 0x42180000+0 gets MSIP; Whisper’s ACLINT does not.

sail.json has platform.interrupts.machine.software.supported = false, so Sail will not take MSI (mie.MSIE is not enabled). mip.MSIP can still be pending. Both sides therefore take MTI when the timer is set, and the checker compares mip: Sail 0x88, DUT 0x80.

pmundkur: Yeah, it would be good to actually cover CSR comparison.  I'd rather avoid illegal bit values in internal register state.  I'm not sure yet whether GDB need to access `mip` directly or use `read_mip()`; I suspect the former.  So lets just keep it always legal, and remember to mask everywhere that is needed.

nadime15: Ok sounds good!

edevolder-tt:
Is there anything I need to do/help?

pmundkur:
@edevolder-tt is it okay if I push my recommended changes to your branch?

edevolder-tt:
sure!

pmundkur:
I've pushed the changes in a single commit.  If they break anything, you could undo that commit.

edevolder-tt:
I see the CI mostly passed; is there something I need to do yet?

pmundkur:
> I see the CI mostly passed; is there something I need to do yet?

Does it break any of your ACT tests?  If not, I'll update the PR title and message to reflect the additional changes and we can merge it if it passes review.

edevolder-tt:
With https://github.com/riscv/sail-riscv/pull/1942/commits/3b3e40c47f0e7e02c5ac49eb7aa72082b314b2ec (cherry-picked on top of master HEAD), our ACT4 tests pass the various Interrupt tests, where we have marked MSI as unavailable. Thank you!

jordancarlin (approved):

nadime15 (approved):

pmundkur:
@edevolder-tt you'll need to add your signing key to Github to get past the "Commits must have verified signatures" check. See here for instructions if you need them:
Then you'll need to push these commits again.

edevolder-tt:
My apologies on the sign off. It's pushed.

pmundkur (approved):

PR #1944, opened 2026-09-10
Fix c.lui assembly to use unsigned immediate

### LINKED ISSUE #1032: c.lui assembly incorrectly uses signed immediates
```
mapping clause assembly = C_LUI(imm, rd)
  <-> "c.lui" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_signed_6(imm)
  when rd != zreg & rd != sp & imm != 0b000000
```

generates assembly like

```
c.lui x23,-0x1d
```

However GAS only accepts an unsigned immediate. LLVM gives this more detailed error:

```
test.s:1:11: error: immediate must be in [0xfffe0, 0xfffff] or [1, 31]
```

JyoshnaKarri:
Fixes #1032. c.lui's assembly mapping used hex_bits_signed_6, producing signed hex output (e.g. -0x1d) that GAS/LLVM reject, since c.lui's immediate is expected in unsigned form. Changed to hex_bits_6 to match GAS/LLVM's expected unsigned encoding, consistent with existing unsigned usage elsewhere in the same file (c.srli, c.srai, c.slli).

Written with AI assistance (Claude) per CONTRIBUTING.md disclosure requirement.

### DIFF
--- a/model/extensions/C/zca_insts.sail
+++ b/model/extensions/C/zca_insts.sail
@@ -211,7 +211,7 @@ function clause execute C_LUI(imm, rd) = {
 }
 
 mapping clause assembly = C_LUI(imm, rd)
-  <-> "c.lui" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_signed_6(imm)
+  <-> "c.lui" ^ spc() ^ reg_name(rd) ^ sep() ^ hex_bits_6(imm)
   when rd != sp & imm != 0b000000
 
 // *****************************************************************


jrtc27:
That's not what LLVM accepts.

jrtc27:
Nor is it what GNU as accepts, which seems to match LLVM.

JyoshnaKarri:
Thanks for catching that — could you point me toward what the correct expected format is? I tested with a single value (31) which happened to assemble correctly, but I may have missed the full valid range this needs to handle.

PR #1946, opened 2026-09-10
Refactor `memory_exception` to take additional information about an access.

### LINKED ISSUE #1913: Supporting page-straddling accesses for `hlv*` and `hsv*` instructions

pmundkur:
This allows differentiating exceptions for HLV/HSV from other cases.

The guest-access flag is bundled with effective privilege and pointer-masking length into an `AccessInfo` structure that is used to access the `vmem_utils` functions.

The HLV/HSV instructions are not yet changed to use the `vmem_utils` API, but this should be the last refactor needed before they can use it to fix #1913.

### DIFF
--- a/model/extensions/H/hext_insts.sail
+++ b/model/extensions/H/hext_insts.sail
@@ -53,12 +53,18 @@ function clause execute HLVTYPE(width, op, is_unsigned, rs1, rd) = {
     HLVX => LoadExecute(Data),
   };
   assert(width <= xlen_bytes);
+
   let eff_priv : Privilege = privLevel_bits(zero_extend(hstatus[SPVP]), 0b1);
-  let pmlen = get_hlsv_pmlen(access, eff_priv);
+  let access_info : AccessInfo = struct {
+    eff_privilege = eff_priv,
+    pmlen = get_hlsv_pmlen(access, eff_priv),
+    as_guest = true,
+  };
+
   match ext_data_get_addr(rs1, zeros(), access, width) {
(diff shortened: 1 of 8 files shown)


model/extensions/cfi/zicfiss_insts.sail
```diff
@@ -27,10 +27,10 @@ function clause execute SSPUSH(rs2) = {
 
   let vaddr_bits = ssp - xlen_bytes;
   let access = Store(ShadowStack);
-  let eff_priv = effectivePrivilege(access, mstatus, cur_privilege);
-  let pmlen = get_pmlen(access, eff_priv);
+  let access_info = default_access_info(access);
+  let eff_priv = access_info.eff_privilege;
```
nadime15: Is this intentional? `eff_priv` is defined but not used.
```suggestion
```

nadime15 (commented):
Looks good. I’ll give it one final check tomorrow!

jordancarlin (approved):

model/extensions/H/hext_insts.sail
nadime15: Can we remove these functions or do you still need them?
```suggestion
```

pmundkur: I was thinking of removing them in the final PR, but might as well remove them now.

nadime15 (approved):

PR #1952, opened 2026-09-15
Add experimental support for the RISC-V P extension (RVP 020)

gglangg:
Implement the packed-SIMD / DSP (P) extension for RV32 and RV64, following draft 020 of the proposed specification
(https://www.jhauser.us/RISCV/ext-P/).

The model is organized under model/extensions/P/:
  - pext_common.sail:  instructions shared between RV32 and RV64
  - pext_32_only.sail: RV32-only instructions
  - pext_64_only.sail: RV64-only instructions
  - pext_types.sail, pext_utils.sail, pext_regs.sail: shared types, helpers and state

The extension is unratified, so it is gated behind --enable-experimental-extensions in addition to its config knob (extensions.P.supported), and every P encdec/assembly clause is guarded by currentlyEnabled(Ext_P).

The P extension reuses the PACK encoding from Zbkb, so that encdec clause is enabled when either extension is present.

Saturating instructions record overflow in the sticky vxsat flag, as required by the in-progress specification (riscv/riscv-p-spec). The vxsat CSR is therefore accessible on harts that implement P without any vector support, in which case writes do not touch mstatus.VS; when the vector context does exist, access continues to require mstatus.VS to be on. The stateen0.VXSAT bit and the VS-off illegal-instruction rule for the P instructions themselves are left as follow-up work.

Encodings were cross-checked against the P-aware binutils (riscv-binutils p-dev) and against the draft 020 encoding tables. The riscv-tests regression passes (664/664) with the extension present, and a riscv_ctg-generated suite of 764 P tests (446 RV32, 318 RV64) runs cleanly on the model.

Generative AI (Claude Code) was used to draft parts of this change and the commit message.


nadime15 (commented):
This will take some time to review, but I had a quick look at it and noticed that you have not added the P bit to `misa`. The spec says it is tentatively reserved, but I could not find anything about `misa` in https://github.com/riscv/riscv-p-spec.

So, in case you still want to add it, you would need to add it to `reset_misa()` in `sys_control.sail` and `legalize_misa()` in `sys_regs.sail`. Otherwise reading misa would indicate P is not supported even though it is.

gglangg:
@nadime15  Fixed in the latest push: misa.P is set in reset_misa(), writable in legalize_misa(), and currentlyEnabled(Ext_P) requires it, same as M/B/V. Thanks!

trdthg:
> a riscv_ctg-generated suite of 764 P tests

Just out of curiosity, where are the tests generated by riscv-ctg? Are they available yet? Also, since you mentioned 'riscv-ctg', does this mean you are still updating the test cases for the older ACT framework, rather than working on ACT 4.0+?

gglangg:
@trdthg Just made them public.

Tests + cgf files at https://github.com/nthu-pllab/rvp-test-suite, and the runner, env headers, patched riscv-ctg and a baseline run at https://github.com/nthu-pllab/rvp-verify.

And yes, this is the older ctg flow, not ACT 4.0.

PR #1956, opened 2026-09-16
Add missed extension dependency checks on F.

pmundkur:


pmundkur:
I found these by accident; it might be good now to invest in a more declarative scheme.

jordancarlin:
> I found these by accident; it might be good now to invest in a more declarative scheme.

Agreed. We might want to think about how we can utilize the fairy extensive extension dependency information from UDB so that we don't need to maintain this in so many different places.

model/postlude/validate_config.sail
```diff
@@ -555,6 +555,32 @@ private function check_misc_extension_dependencies() -> bool = {
...
+      print_endline("The Zfbfmin extension is enabled but F is disabled: supporting Zfbfmin requires F.");
+    };
+    if (config extensions.Zcf.supported : bool) then {
+      valid = false;
+      print_endline("The Zcf extension is enabled but F is disabled: supporting Zcf requires F.");
+    };
+    if (config extensions.Zcd.supported : bool) & not(constraint(ext_d_supported)) then {
+      valid = false;
+      print_endline("The Zcd extension is enabled but D is disabled: supporting Zcd requires D.");
+    };
```
jordancarlin: This doesn't seem right. It is inside the if not F block, so would only be reached if F is not supported, in which case D will never be supported.

nadime15 (approved):

jordancarlin (approved):

PR #1957, opened 2026-09-16
Extend the C++ model API to get information on the F and V extensions.

pmundkur:
While here, strengthen a few asserts.


c_emulator/riscv_model_impl.cpp
```diff
@@ -583,29 +596,35 @@ bool ModelImpl::had_exception() const {
...
+  assert(0 <= reg && reg < 32);
+
   const sbits val = zrX(reg);
   return val.bits;
 }
 
 uint64_t ModelImpl::freg(int64_t reg) {
   // For the E base ISA, this assert should use 16.
-  assert(reg < 32);
+  assert(0 <= reg && reg < 32);
```
jordancarlin, line 606: Are we better off making `reg` a `uint64_t` so we don't need to assert that it is greater than 0? Global comment.

pmundkur: Unfortunately the model uses `int64_t` for register indices (see the signature for `zrX()` for e.g.), so unless this is changed in the Sail backend, we might have to live with it.

jordancarlin: Unfortunate. Fine for now then, but we might want to look at updating that on the Sail side later.

jordancarlin (approved):

c_emulator/riscv_model_impl.cpp
```diff
@@ -396,6 +396,15 @@ void ModelImpl::init_platform_constants() {
...
   m_supports_hypervisor = get_config_bool({"extensions", "H", "supported"});
+  m_supports_D_extension = get_config_bool({"extensions", "D", "supported"});
+  m_has_float_registers = get_config_bool({"extensions", "F", "supported"});
+
+  std::string vector_support = get_config_string({"extensions", "V", "support_level"});
+  m_has_vector_registers = vector_support != "Disabled";
+  if (m_has_vector_registers) {
+    uint64_t vlen_exp = get_config_uint64({"extensions", "V", "vlen_exp"});
+    m_vlen = 0x1U << vlen_exp;
+  }
```
nadime15:
```suggestion
```

c_emulator/riscv_model_impl.h
```diff
@@ -85,6 +85,18 @@ class ModelImpl final : private hart::Model {
...
+  }
+  bool has_float_registers() const {
+    return m_has_float_registers;
+  }
+  bool has_vector_registers() const {
+    return m_has_vector_registers;
+  }
+  uint64_t vlen() const {
+    return m_vlen;
+  }
```
nadime15, line 97: This seems complicated, is any of this actually needed?

Instead of defining m_vlen, just read zvlen, the code already does the same for flen() and xlen() with zxlen and zflen and you could also make use of zvector_support_level.
```suggestion
  bool has_vector_registers() const {
    return zvector_support_level != zDisabled;
  }
  uint64_t vlen() const {
    assert(has_vector_registers());
    return zvlen;
  }
```

nadime15 (approved):

PR #1959, opened 2026-09-18
PMP: ignore grain bits in both TOR bounds

### LINKED ISSUE #1951: PMP: TOR lower bound retains NAPOT readback bits
[`pmpCheck`](https://github.com/riscv/sail-riscv/blob/22fad389cb0ca92ef8f3cfff91fe258f14ee8de8/model/pmp/pmp_control.sail#L138-L142) gets a TOR entry's lower bound through:

```sail
let prev_pmpaddr = if i > 0 then pmpReadAddrReg(i - 1) else zeros();
```

[`pmpReadAddrReg`](https://github.com/riscv/sail-riscv/blob/22fad389cb0ca92ef8f3cfff91fe258f14ee8de8/model/pmp/pmp_regs.sail#L70-L89) applies readback masking according to the preceding entry’s address mode. at G=1, a NAPOT predecessor retains raw pmpaddr[i-1][0], shifting the following TOR lower bound by four bytes.

## Spec.

The spec. says the low G bits do not affect TOR matching, including when a register supplies the following entry’s lower bound:
- [ratified specification](https://github.com/riscv/riscv-isa-manual/blob/be8cb008612ce2ac1712e4ba3f09fc2109c0d462/src/priv/machine.adoc#L3496-L3528)
- [riscv-isa-manual issue #884](https://github.com/riscv/riscv-isa-manual/issues/884#issuecomment-3631302703)

## example

With `G=1`:

- Entry 0 is NAPOT with `pmpaddr[0] = 0x5`, encoding `[0x10, 0x20)`
- Entry 1 is TOR with `pmpaddr[1] = 0x10`
- spec. derived entry-1 range is `[0x10, 0x40)`
- current model entry-1 range: `[0x14, 0x40)`

this changes the per-entry match relation. It doesn't change the final `pmpCheck result in the checked domain: entry 0 returns Match or PartialMatch first and therefore decides before entry 1.

The same source-level mismatch exists for G>1, where NAPOT readback sets low bits that TOR matching should ignore. The formal evidence below covers G=1 only.

## Suggested change

Construct a TOR lower bound by clearing pmpaddr[G-1:0] independently of the preceding entry’s mode. Keep pmpReadAddrReg for architectural CSR readback.

I can PR with the change and a regression if this direction is okay or preferred.

## Evidence

The discrepancy and its shadowing were formally checked on Sail-RISC-V 0.14 and reproduced on 0.12:

- [Protocol and result summary](https://github.com/codeadpool/cva6-priv-sva/blob/3ac9039dc33ed80fcb90b9c9c9140382b93318fc/evidence/sail/TIER1_PROTOCOL.md#2026-09-14-section-31-results)
- [Archived logs and witnesses](https://github.com/codeadpool/cva6-priv-sva/tree/3ac9039dc33ed80fcb90b9c9c9140382b93318fc/evidence/sail/leaf)

issue was exposed while comparing generated Sail PMP logic with CVA6 after the independently authored [CVA6 PR #3490](https://github.com/openhwgroup/cva6/pull/3490).

codeadpool:
For TOR entry `i`, `pmpCheck` currently uses `pmpReadAddrReg(i-1)` as its lower bound. When entry `i-1` is NAPOT, that readback can retain or set low grain bits. The spec. states: “Bits `pmpaddr_i[G-1:0]` do not affect the TOR address-matching logic.”  [riscv-isa-manual#884](https://github.com/riscv/riscv-isa-manual/issues/884#issuecomment-3631302703) clarifies that this applies when the register supplies either TOR bound.

This change clears the low `G` bits of both TOR bounds inside `pmpMatchAddr`. CSR readback behavior is left unchanged; this makes per-entry region correct regardless of how the bounds were obtained, rather than relying on the readback path to pre-mask them.

#### Testing

A native Sail `$[test]` covers `G=0`, `G=1`, and `G=2`, lives beside the private helper and is collected by the existing `unit_tests` target. The target passes, and I verified that the test fails if `pmpTORBound` stops clearing the low bits.

No architectural/ACT test is included. In the formally checked `G=1` domain, whenever the unmasked lower bound changes the following TOR entry’s leaf match,  `pmpCheck` has already returned at the preceding NAPOT entry or an earlier entry.  The final decision is therefore unchanged. [Formal evidence](https://github.com/codeadpool/cva6-priv-sva/blob/3ac9039dc33ed80fcb90b9c9c9140382b93318fc/evidence/sail/TIER1_PROTOCOL.md#2026-09-14-section-31-results).

I used multiple models to get info on repo's testing conventions and inline test usage.

Fixes #1951.


pmundkur:
@codeadpool you'll need to add your signing key to Github to get past the "Commits must have verified signatures" check. See here for instructions if you need them:
Then you'll need to push this commit again.

codeadpool:
Thanks, fixed.

pmundkur (approved):

jordancarlin (approved):

model/pmp/pmp_control.sail
```diff
@@ -73,6 +73,22 @@ private function pmpRangeMatch(
...
   else    PMP_PartialMatch
 
+// Bits pmpaddr[G-1..0] do not affect TOR address matching, regardless of
+// whether the register supplies the lower or upper bound.
+private function pmpTORBound(addr : xlenbits, grain : range(0, 63)) -> xlenbits =
+  if grain == 0 then addr
+  else {
+    let mask : xlenbits = zero_extend(ones(min(grain, xlen)));
+    addr & ~(mask)
+  }
```
nadime15: Small nit, but you could simplify the code to:
```suggestion
private function pmpTORBound(addr : xlenbits, grain : range(0, 63)) -> xlenbits = {
  let mask : xlenbits = zero_extend(ones(min(grain, xlen)));
  addr & ~(mask)
}
```

since addr & ~0 = addr

codeadpool: Thanks, applied.

nadime15 (approved):

PR #1960, opened 2026-09-18
Lean update, including termination measures

bacam:
This is sufficient to get the Lean emulator to build and pass the tests.  However, this is a draft because it needs a nasty hack to tell Lean that `get_root_level` returns a value between one and four.  (It's particularly nasty because Sail _does_ know that, and it will optimise away a simpler hack.)  I'm still looking for a better way to do that.


model/sys/vmem.sail
```diff
@@ -87,7 +87,7 @@ private function translate_PTE_addr(
   }
 }
 
-termination_measure translate_PTE_addr(_, _, _, _) = 1000
+termination_measure translate_PTE_addr(_, _, stage, _) = if stage == VS_Stage then 10 else 0
```
pmundkur, line 90: It would be good to have some comments somewhere on how the constants (e.g. 10 here) for these measures are derived.

bacam:
Closing because #1871 now covers this.

